PR #1, opened 2019-09-23
tree-wide: use a standard project layout

teg:
Keep main.go files under cmd/ and internal libraries under internal/.
This will allow us to add more exutables under cmd/ (whereas only one
was possible when main.go was kept in the root).


msehnout:
lgtm

PR #3, opened 2019-09-24
create jobs queue for scheduling new builds

msehnout:


internal/weldr/api.go
```diff
@@ -10,10 +10,12 @@ import (
 	"github.com/julienschmidt/httprouter"
 
 	"osbuild-composer/internal/rpmmd"
+	"osbuild-composer/queue"
```
teg: Libraries should be in either the `pkg` or `internal` subdirectory. For now, let's put everything in `internal`, and only if we want something to be public we put it in `pkg`.

queue/manifest.go
```diff
@@ -0,0 +1,63 @@
...
+	"os"
+)
+
+// Manifest contains additional metadata attached do a pipeline that are necessary for workers
+type Manifest struct {
+	destination string
+}
+
+// SaveImage saves "src" into provided destination
+func (m *Manifest) SaveImage(src string) error {
```
teg: Drop for now?

queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+	WorkerID string `json:"worker-id"`
+	Build    Build  `json:"build"`
+}
+
+// JobQueue contains already running jobs waiting for
+type JobQueue struct {
+	sync.Mutex
+	incomingBuilds chan Build
+	queue          []Job
+	timeoutSeconds int
```
teg: I suggest dropping any timeout handling for now, and we can figure out how to deal with it in a future PR. I imagine reusing the PATCH call with "status": "running" at regular intervals from the worker may do the trick.

queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+// Build is a request waiting for a worker
+type Build struct {
+	Pipeline string `json:"pipeline"`
+	Manifest string `json:"manifest"`
+}
+
+// Job is an image build already in progress
+type Job struct {
+	UUID     string `json:"uuid"`
+	WorkerID string `json:"worker-id"`
```
teg: Our model currently does not contain a worker objcet/ID, so I suggest dropping this for now. We almost certainly want to introduce it, but we should do it at the same time in the API I think, once we have discussed how it should look like.

teg:
Great stuff! Thanks! I'll merge this for now, and work on top of it rather than give more feedback here (usually we'd work together on the PR, but since you'll be away for a while and this is all new code anyway, I figure that is fine).

PR #5, opened 2019-09-24
Add blueprint workspace diff

jkozol:
The list of changes between the committed blueprint and its workspace are returned when a user requests the diff. Each change includes the new(added) or the old(removed) package.

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -453,13 +453,72 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
 	})
 }
 
-func (api *API) blueprintsDiffHandler(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
-	var reply struct {
-		Diff []interface{} `json:"diff"`
+func (api *API) blueprintsDiffHandler(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
+	type pack struct {
+		Package blueprintPackage `json:"Package"`
 	}
 
-	reply.Diff = make([]interface{}, 0)
-	json.NewEncoder(writer).Encode(reply)
+	type diff struct {
+		New *pack `json:"new"`
+		Old *pack `json:"old"`
+	}
+
+	type reply struct {
+		Diffs []diff `json:"diff"`
+	}
+
+	name := params.ByName("blueprint")
+	if len(name) == 0 {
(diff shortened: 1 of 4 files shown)


internal/weldr/api.go
```diff
@@ -453,13 +453,72 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
...
+		Diffs []diff `json:"diff"`
+	}
+
+	name := params.ByName("blueprint")
+	if len(name) == 0 {
+		statusResponseError(writer, http.StatusNotFound, "no blueprint name given")
+		return
+	}
+	fromCommit := params.ByName("from")
+	if len(fromCommit) == 0 {
```
larskarlitski: I would merge the lower check for `NEWEST` with this one. Go is chatty, but we can try to keep it down a bit. Don't worry that the error message will be the same.

Same goes for `to`.

internal/weldr/api.go
```diff
@@ -453,13 +453,72 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
...
+		}
+	} else {
+		statusResponseError(writer, http.StatusNotFound, "invalid to commit ID given")
+		return
+	}
+
+	diffs := []diff{}
+
+	for _, oldPack := range oldBlueprint.Packages {
+		if !packageListContains(newBlueprint.Packages, oldPack) {
```
larskarlitski: This is O(n²) :cry: 

I'd copy one of the blueprints into a map, with names as keys, and the other one to a list. Then, iterate over the list  and for each package, check if it's in the map. If it is, delete it from both. At then end, the map will contain all added packages and the list all removed packages.

jkozol:
> The documentation says that the route also returns differences in Name, Description, and Version. Does cockpit-composer look at those as well?

Thank you for pointing this out. I missed the version change. A blueprints name can't currently be changed and a change of description is not used by the UI.

internal/weldr/api.go
jkozol: These should be combined now.

internal/weldr/api.go
jkozol: I updated this per your advice except instead of dropping from the list I just added those packages directly to the diffs slice. I believe it runs in O(nlogn) now.

jkozol:
> I'd squash this commit with the other, so we always test the code as it is added.

The commits have been squashed

PR #6, opened 2019-09-25
Fix package name-version for wildcard version

jkozol:
If a package had version "*" its name-version would be "name-*". This would match all packages containing "name" in their name. Instead the name-version is "name-*-*" so it matches any version of the package.

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -443,11 +443,14 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
 		specs := make([]string, len(blueprint.Packages))
 		for i, pkg := range blueprint.Packages {
 			specs[i] = pkg.Name
-			if pkg.Version != "" {
+			// If a package has version "*" the package name suffix must be equal to "-*-*.*"
+			// Using just "-*" would find any other package containing the package name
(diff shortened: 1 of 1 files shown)


internal/weldr/api.go
```diff
@@ -443,11 +443,14 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
 		specs := make([]string, len(blueprint.Packages))
 		for i, pkg := range blueprint.Packages {
 			specs[i] = pkg.Name
-			if pkg.Version != "" {
+			// If a package has version "*" the package name suffix must be equal to "-*-*"
+			// Using just "-*" would find any other package containing the package name
+			if pkg.Version != "" && pkg.Version != "*" {
 				specs[i] += "-" + pkg.Version
+			} else if pkg.Version == "*" {
+				specs[i] += "-*-*"
```
larskarlitski: Sorry, I noticed that this falls down if there's more than a single dash in the package name. It's more robust to include an asterisk for the architecture as well, which is separated by a dot. So `specs[i] += "-*-*.*"`

For example, `python3-*-*` lists a lot of python packages, but `python3-*-*.*` only lists all versions of `python3`.

jkozol: Thanks for catching this. I updated the string.

PR #12, opened 2019-09-30
Update api and store for parity with lorax

jkozol:
Add parameters to the api responses that while not displayed by cockpit-composer are used in props validation or blueprint updates. These updates bring osbuild-composer's api closer to lorax-composer's.

### DIFF
--- a/internal/rpmmd/repository.go
+++ b/internal/rpmmd/repository.go
@@ -32,7 +32,7 @@ type Package struct {
 
 type PackageSpec struct {
 	Name    string `json:"name"`
-	Epoch   uint   `json:"epoch,omitempty"`
+	Epoch   uint   `json:"epoch"`
 	Version string `json:"version,omitempty"`
 	Release string `json:"release,omitempty"`
 	Arch    string `json:"arch,omitempty"`
--- a/internal/weldr/store.go
+++ b/internal/weldr/store.go
@@ -26,10 +26,11 @@ type store struct {
 
 type blueprint struct {
 	Name        string             `json:"name"`
-	Description string             `json:"description,omitempty"`
+	Description string             `json:"description"`
 	Version     string             `json:"version,omitempty"`
 	Packages    []blueprintPackage `json:"packages"`
 	Modules     []blueprintPackage `json:"modules"`
+	Groups      []blueprintPackage `json:"groups"`
 }
 
 type blueprintPackage struct {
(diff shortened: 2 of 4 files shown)


internal/rpmmd/repository.go
```diff
@@ -32,7 +32,7 @@ type Package struct {
 
 type PackageSpec struct {
 	Name    string `json:"name"`
-	Epoch   uint   `json:"epoch,omitempty"`
+	Epoch   uint   `json:"epoch"`
```
larskarlitski: The `omitempty` was for not sending anything when the `Epoch` was unset (i.e., 0). Why did you have to remove it? Maybe this points to a deeper problem somewhere where we forget to set it?

internal/weldr/store.go
```diff
@@ -148,6 +149,12 @@ func (s *store) getBlueprint(name string, bp *blueprint, changed *bool) bool {
 	if bp.Modules == nil {
 		bp.Modules = []blueprintPackage{}
 	}
+	if bp.Groups == nil {
+		bp.Groups = []blueprintPackage{}
+	}
+	if bp.Version == "" {
+		bp.Version = "0"
```
larskarlitski: I've seen lorax-composer complain about version that don't follow "semantic versioning". Does cockpit-composer depend on that as well? if so, it might make sense to put `0.0.0` here and below.

jkozol: Cockpit-composer does not depend on semantic versioning. It just requires the version to be a string. I can change it to semantic versioning to be in line with lorax composer.

internal/rpmmd/repository.go
jkozol: lorax-composer seems to always send an epoch of 0 if there is no epoch defined. This seems to follow the behaviour outlined [here](https://weldr.io/RPM-Dependencies/) which states:

> If the epoch is not specified, it’s treated as 0.

If I run `dnf info 2ping` I see that 2ping does not have an epoch defined. When I call lorax-composer's /projects/info/2ping its build contains "epoch: 0". This seems to come from the python3-dnf plugin lorax uses which will return 0 for the epoch of a package with no defined epoch. Epoch is not displayed by cockpit-composer but is checked for in the props validation. For equivalence with lorax we should default to an epoch of 0 but not returning an epoch does not break anything in the UI.

internal/weldr/store.go
jkozol: The version is semantic now

PR #14, opened 2019-10-01
Integration of osbuild composer with systemd

msehnout:
This PR contains basic configuration for systemd. It uses socket activation for the osbuild-composer service and regular service for osbuild-worker.


msehnout:
It should work after invocation of `make build && make install` in the checkout root and `make sd-install && make sd-start` in the `distribution` directory.

Makefile
```diff
@@ -0,0 +1,12 @@
+build:
+	go build -o osbuild-composer ./cmd/osbuild-composer/
+	go build -o osbuild-worker ./cmd/osbuild-worker/
+
+install:
```
teg: This should depend on `build`, shouldn't it?

Makefile
```diff
@@ -0,0 +1,12 @@
...
+	- mkdir -p /usr/lib/osbuild-composer
+	cp osbuild-composer /usr/lib/osbuild-composer/
+	cp osbuild-worker /usr/lib/osbuild-composer/
+	cp dnf-json /usr/lib/osbuild-composer/
+
+run-socket:
```
teg: Same as above.

distribution/Makefile
```diff
@@ -0,0 +1,21 @@
+sd-install:
+	cp *.service /etc/systemd/system/
+	cp *.socket /etc/systemd/system/
+	systemctl daemon-reload
+	
+sd-start:
+	systemctl enable osbuild-composer.socket
```
teg: Very minor nit: maybe we don't need "enable" here? Or just make it a separate target for convenience?

distribution/osbuild-composer.conf
```diff
@@ -0,0 +1 @@
+u   _osbuild-composer   -   "OSBuild Composer user"
```
teg: Why the underscore? I'm probably unaware of some policy :)

distribution/osbuild-composer.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=OSBuild Composer
+After=multi-user.target
+# Running osbuild-composer without the worker is useless
```
teg: Actually, in the future we don't want to require the workers to be local, so I would drop the requriement. But rather pull in the worker in the install section, so that by default enabling composer enables a worker (but you can disable it if you don't want it).

distribution/osbuild-composer.service
```diff
@@ -0,0 +1,22 @@
...
+# therefore it is a hard requirement
+Requires=osbuild-worker.service osbuild-composer.socket
+
+[Service]
+Type=simple
+ExecStart=/usr/lib/osbuild-composer/osbuild-composer
+# /run/..
```
teg: I'd drop these comments throughout, people can check the manpage ;)

distribution/osbuild-composer.service
```diff
@@ -0,0 +1,22 @@
...
+RuntimeDirectory=osbuild-composer weldr
+# /var/lib/..
+StateDirectory=osbuild-composer
+WorkingDirectory=/usr/lib/osbuild-composer/
+User=_osbuild-composer
+Restart=on-failure
+
+# Worker will contact composer immediately after start-up
```
teg: Hm, I'd drop the comment. We don't know about the worker (when they come online), but most probably we do want to start this eagerly so it can download its metadata etc before someone tries to use it from cockpit (so I agree with the install section).

distribution/osbuild-worker.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=OSBuild Composer Worker
+After=multi-user.target osbuild-composer.socket
+
+[Service]
+Type=simple
+ExecStart=/usr/lib/osbuild-composer/osbuild-worker
+# /var/cache/..
```
teg: I'd drop this too.

distribution/osbuild-worker.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
teg: There is already a service file in the init/ directory. Please delete that as this one replaces it. I would copy some of that logic though, it is a service templace, so people can enable as many or as few workers as they want (but by default only one is enabled when you call "systemctl enable osbuild-worker@.service").

Makefile
msehnout: Yes, in theory. But in that case `build` should probably depend on the source files so we don't rebuild it every time even without changing it. But I don't know it sth like `$(shell find . -name '*.go')` is appropriate for go packages.

distribution/osbuild-composer.conf
msehnout: > It is strongly recommended to pick user and group names that are unlikely to clash with normal users created by the administrator. A good scheme to guarantee this is by prefixing all system and group names with the underscore, and avoiding too generic names.
Source: https://www.freedesktop.org/software/systemd/man/sysusers.d.html#

Makefile
teg: Fair point. Leave it as is.

distribution/osbuild-composer.conf
teg: Ah, sounds good. Thanks!

PR #19, opened 2019-10-03
Translate blueprints to pipelines

teg:
Add basic infrastructure for blueprint to pipeline translations.

### DIFF
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -1,0 +1,28 @@
+// Package blueprint contains primitives for representing weldr blueprints and
+// translating them to OSBuild pipelines
+package blueprint
+
+import "osbuild-composer/internal/pipeline"
+
+// A Blueprint is a high-level description of an image.
+type Blueprint struct {
+	Name        string    `json:"name"`
+	Description string    `json:"description"`
+	Version     string    `json:"version,omitempty"`
+	Packages    []Package `json:"packages"`
+	Modules     []Package `json:"modules"`
+	Groups      []Package `json:"groups"`
+}
+
+// A Package specifies an RPM package.
+type Package struct {
+	Name    string `json:"name"`
+	Version string `json:"version,omitempty"`
+}
+
+// TranslateToPipeline converts the blueprint to a pipeline for a given output format.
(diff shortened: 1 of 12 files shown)


internal/blueprint/blueprint.go
```diff
@@ -0,0 +1,28 @@
+// Package blueprint contains primites for representing weldr blueprints and
```
larskarlitski:
```suggestion
// Package blueprint contains primitives for representing weldr blueprints and
```

internal/blueprint/blueprint.go
```diff
@@ -0,0 +1,28 @@
...
+import "osbuild-composer/internal/pipeline"
+
+// A Blueprint is a high-level description of an image.
+type Blueprint struct {
+	Name        string    `json:"name"`
+	Description string    `json:"description"`
+	Version     string    `json:"version,omitempty"`
+	Packages    []Package `json:"packages"`
+	Modules     []Package `json:"modules"`
+	Groups      []Package `json:"groups"`
```
larskarlitski: Maybe we should only have one thing here? Weldr has all three, but puts packages in each of them. We'd need to support this still in the API, but could combine them here.

internal/blueprint/blueprint.go
```diff
@@ -20,9 +23,34 @@ type Package struct {
 	Version string `json:"version,omitempty"`
 }
 
+type translator interface {
+	translate(b *Blueprint) *pipeline.Pipeline
+}
```
larskarlitski: This type is called `translator`, but the files are `XXX_output.go`. Does it make sense to give those consistent names?

teg: Yeah, I'll rename them all "outputs".

internal/blueprint/blueprint.go
teg: Hm, let's address that in a follow-up? (I would have to look into how it works, this currently just preserves the behaviour).

teg:
Pushed an update, also added accessors for filename and mimetype per output type. This in order to avoid hard-coding those things in more than one place.

internal/blueprint/blueprint.go
larskarlitski: Yeah, I was just thinking out loud.

PR #21, opened 2019-10-04
test: run api tests against lorax

jkozol:
osbuild-composer's tests can be run against lorax-composer's api in order to ensure parity with lorax. This can be accomplished by passing the environment variable LORAX=true into the tests with lorax-composer running on the api socket. Without the environment variable the tests will run as usual against osbuild-composer's api.

### DIFF
--- a/internal/weldr/api_test.go
+++ b/internal/weldr/api_test.go
@@ -2,10 +2,13 @@ package weldr_test
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


jkozol:
Thanks for the feedback. I've updated the naming.

internal/weldr/api_test.go
```diff
@@ -26,16 +29,48 @@ var packages = rpmmd.PackageList{
...
 
-	if resp.Code != expectedStatus {
-		t.Errorf("%s: expected status %v, but got %v", path, expectedStatus, resp.Code)
+	return resp.Result()
+}
+
+func testRoute(t *testing.T, api *weldr.API, method, path, body string, expectedStatus int, expectedJSON string) {
+	var resp *http.Response
+
+	if os.Getenv("EXTERNAL") == "true" {
```
larskarlitski: Hm sorry to be nitpicky, but could you please namespace the environment variable? Something like `OSBUILD_COMPOSER_TEST_EXTERNAL`. Also, don't check for `true` specifically, but only if it is "set", i.e., a non-empty string.

internal/weldr/api_test.go
```diff
@@ -26,16 +29,48 @@ var packages = rpmmd.PackageList{
...
+func externalRequest(method, path, body string) *http.Response {
+	client := http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", "/run/weldr/api.socket")
+			},
+		},
+	}
+
+	req, _ := http.NewRequest(method, "http://localhost"+path, bytes.NewReader([]byte(body)))
```
larskarlitski: When running the test without composer running, I get a panic somewhere further down. It's fine to use `panic` in test code, but please do so here with a descriptive message (whatever the error from this method is as string is fine). Same a couple of lines down.

internal/weldr/api_test.go
jkozol: Oh no thanks for the nitpicks! This has been fixed.

internal/weldr/api_test.go
jkozol: Okay. To make sure I understand properly, you're saying I shouldn't ignore possible errors when creating the request and when sending it. Instead, I can use `panic(error)` to quit the execution and print out the error. This should be fixed now.

internal/weldr/api_test.go
larskarlitski: Thanks! You were checking for existence now, but unsetting a variable like this
```
$ OSBUILD_COMPOSER_TEST_EXTERNAL= go test ./...
```
is quite common. I've changed it to `len(os.Getenv("OSBUILD_COMPOSER_TEST_EXTERNAL")) > 0`.

internal/weldr/api_test.go
larskarlitski: Indeed. Error is now perfectly understandable:
```
panic: Get http://localhost/api/status: dial unix /run/weldr/api.socket: connect: permission denied [recovered]
	panic: Get http://localhost/api/status: dial unix /run/weldr/api.socket: connect: permission denied
```

PR #22, opened 2019-10-04
Introduce tests for image-info

larskarlitski:
See commits. The basic idea of two kinds of tests is there with an example for each.

We might want to only run those when the commit touches `/tools/image-info`. What do you think?


.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+      env: GO111MODULE=on # needed even for Go 1.12
+      script: go test -v ./...
+
+    - language: python
+      python: 3.7
+      env: PYTHONUNBUFFERED=1
+      script:
+        # ubuntu's rpm package sets dbpath to ~/.rpmdb, which makes rpm fail...
+        - sudo sh -c 'mkdir /etc/rpm; echo "%_dbpath /var/lib/rpm" > /etc/rpm/macros'
+        - sudo env "PATH=$PATH" "OSBUILD_TEST_BUILD_PIPELINE=tools/test_image_info/base-from-yum.json" python3 -m unittest tools.test_image_info
```
teg: I think this should be build-from-yum.json. The point being that that is specific to the host, and might be swapped out on a different CI or locally, so any requirements of the build image for the real build should be in each of the pipelines we test.

tools/test_image_info/base-from-yum.json
```diff
@@ -0,0 +1,45 @@
...
+          "packages": [
+            "dnf",
+            "systemd",
+            "gnupg"
+          ]
+        }
+      }
+    ]
+  },
+  "stages": [
```
teg: This should be redundant, no?

.travis.yml
larskarlitski: I had trouble understanding which is which. Thanks :)

PR #27, opened 2019-10-09
build-sys: rework makefiles

teg:
Unify into one make file, with three targets:

build:
  This unconditionally builds the binaries that wolud be shipped in the package,
  this is marked as a phony target, and relies on the go compiler to know what
  is up-to-date and what needs to be rebuilt.

install:
  This installs binaries and config files into the file system and runs any
  scripts to set things up correctly, as if the rpm had been installed.

restart:
  This starts or restarts the services, as if the system had been rebooted.

This drops the old `run` target, which could still be done from the commandline,
but we want to be able to rely on systemd features, so it is not something we
want to test or support.

To run osbuild-composer from a checkout one should now do:
```
$ make build
$ sudo make install
$ sudo make restart
```


larskarlitski:
> Looks good :) but I'd appreciate alternative for `make sd-stop`. I use it when I want to run the services by hand or separately.

I think we should either include `restart` and something like Martin suggests here, or not include any. Either we add convenience rules, or we don't.

I lean towards not including either. Everyone's workflow is different and will need different rules/tools/scripts. Let's not add them all here.

teg:
Yeah, fair enough, I think we probably won't agree on the perfect set of convenience scripts (and adding every combination would create a mess), so I've removed them all, and we'll just have to maintain whatever workflow we have locally. Thanks for the reviews!

PR #32, opened 2019-10-14
dnf-json: Use /var/cache as dnf cachedir

ondrejbudai:
When osbuild-composer is run as systemd service, we don't want to write
anything into working directory. Currently, we write dnf cache into it.
Instead, use absolute path to /var/cache as dnf cache.

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -24,7 +24,6 @@ except IndexError:
 
 base = dnf.Base()
 
-base.conf.cachedir = "./dnf-cache"
 base.conf.substitutions["releasever"] = "30"
 base.conf.substitutions["basearch"] = "x86_64"
 


dnf-json
```diff
@@ -24,7 +24,7 @@ except IndexError:
 
 base = dnf.Base()
 
-base.conf.cachedir = "./dnf-cache"
+base.conf.cachedir = "/var/cache/osbuild-composer/dnf-cache"
```
larskarlitski: Let's just use the systems' normal cache.

ondrejbudai: What do you mean by systems' normal cache?

larskarlitski: DNF's cache is usually in `/var/cache/dnf` (but can be configured, of course). I'd say let's just use whatever is the configured dnf cache on the system instead of having our own. This way, it can be shared.

In other words: please delete that line :)

ondrejbudai: The thing is that `/var/cache/dnf` is owned by root, but we're running osbuild-composer as non-root user.

ondrejbudai: I was wrong, it works just right, I guess libdnf makes some kind of rights escalation... I've updated the PR.

PR #36, opened 2019-10-18
blueprint/output/qcow2: add unit tests

teg:
Use the autogenerated test framework from VSCode (with minor fixes to
make them compile).

As we do not yet support customizations in blueprint, all the tests are
trivial.

I was not entirely sure about the best way to encode the wanted pipeline
output. I currently represent it as a JSON string and unmarshal it into
the object to compare with.

### DIFF
--- a/internal/blueprint/qcow2_output_test.go
+++ b/internal/blueprint/qcow2_output_test.go
@@ -1,0 +1,78 @@
+package blueprint
+
+import (
+	"encoding/json"
+	"reflect"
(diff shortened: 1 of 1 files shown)


teg:
> lgtm, but I'd probably consider creating a directory with expected pipelines as separate files and maybe few utility functions to load them (not necessary for this PR)

I like this approach, as we can then reuse the same json files to generate and test the images.

PR #40, opened 2019-10-21
org.osbuild.dnf → org.osbuild.rpm

larskarlitski:


internal/blueprint/f30_helpers.go
```diff
@@ -2,54 +2,88 @@ package blueprint
...
+		"kernel",
+		"selinux-policy-targeted",
+		"grub2-pc",
+		"spice-vdagent",
+		"qemu-guest-agent",
+		"xen-libs",
+		"langpacks-en",
+	}...)
+	for _, pkg := range deps {
+		options.AddPackage(pkg.URL, pkg.Checksum)
```
msehnout: We can't do the depsolv here because customization might add additional packages which would be incompatible with those resolved here. We need to resolve only after we have a final set of packages. Does that make sense to you?

larskarlitski: Yeah, we have to depsolve after all packages are added. I'll look into how you're adding them in the other output types / customizations and then rebase to include those.

larskarlitski:
This is out of date now, because we've reworked the way we generate pipelines. I'd still like to make this change, but it will need a different approach. Closing for now.

PR #41, opened 2019-10-22
Customizations

ondrejbudai:
Code for simpler customizations seems to be done, we need to discuss:
- how to do customization of "initial stages" (grub stage for kernel args customizations, dnf/rpm stage for additional packages)
- how to do testing - we have n output types and m customizations, together n*m options. That's too much. Also, we need some kind of framework, current test solution would be too much messy for testing customizations imo.

### DIFF
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -20,12 +20,13 @@ func (e *InvalidOutputFormatError) Error() string {
 
 // A Blueprint is a high-level description of an image.
 type Blueprint struct {
-	Name        string    `json:"name"`
-	Description string    `json:"description"`
-	Version     string    `json:"version,omitempty"`
-	Packages    []Package `json:"packages"`
-	Modules     []Package `json:"modules"`
-	Groups      []Package `json:"groups"`
+	Name           string          `json:"name"`
+	Description    string          `json:"description"`
+	Version        string          `json:"version,omitempty"`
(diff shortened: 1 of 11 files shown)


internal/blueprint/blueprint.go
```diff
@@ -66,7 +67,14 @@ func ListOutputFormats() []string {
 // ToPipeline converts the blueprint to a pipeline for a given output format.
 func (b *Blueprint) ToPipeline(outputFormat string) (*pipeline.Pipeline, error) {
 	if output, exists := outputs[outputFormat]; exists {
-		return output.translate(b), nil
+		p := output.translate(b)
+		if b.Customizations != nil {
```
teg: I think you should leave this to each translation function to deal with. I don't think you need to change this function at all. For two reasons: the stages you append should perhaps not go last (SELinux probably should). and it might be that a customization changes some stage, rather than just append to it. As this is going to be in conflict with Martin's work, I suggest simply leaving it out for now, and then adding the same call `addToPipeline()` to each of the translations as he finishes them in follow-up PRs. If it turns out that we need to split addToPipeline() into smaller helper functions for some output  formats that's totally fine of course, but could also be done in follow-ups.

ondrejbudai:
@teg I've removed the integration part for now. Also, I've refactored the methods a bit. It makes a tiny bit more sense to me now, but I can revert the changes if you prefered the previous style.

PR #53, opened 2019-10-25
PoC: Integration test

teg:
This shows how we could generate images to make integration tests. First build the rpms and create a repository, then run a pipeline against that and boot the output image.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -33,7 +33,8 @@ Requires: osbuild
 %{common_description}
 
 %prep
-%forgeautosetup -p1
+%autosetup -p1
+#forgeautosetup -p1
 
 %build
 %gobuildroot
--- a/test.json
+++ b/test.json
@@ -1,0 +1,130 @@
+{
+  "build": {
+    "stages": [
+      {
+        "name": "org.osbuild.dnf",
+        "options": {
+          "repos": [
+            {
+              "metalink": "https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch",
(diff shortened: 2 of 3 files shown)


ondrejbudai:
I am not sure about the spec file change but I think it's not blocking this PR. I can revisit it later when this PR is merged.

In short words, I've reviewed the change in spec file and I'm approving it. :)

msehnout:
I really like the idea of writing more integration tests :) and I have few questions regarding this PoC:
 * What would be the advantage over building the images in the testing farm and installing them directly to the VM? 
 * If we boot this image, do we have a CLI client that we can use to trigger image builds remotely (from the host VM to the guest running osbuild-composer)? For example the lorax-composer CLI tool

Finally, I'd prefer to write these tests in Python over Bash, if possible. I like Bash, because it is very easy to write it and in this case it is much shorter than the same script would be in Python, but if we are going to write more code like this, those extra lines are worth the benefits (far less traps in Python).

teg:
> If we boot this image, do we have a CLI client that we can use to trigger image builds remotely (from the host VM to the guest running osbuild-composer)? For example the lorax-composer CLI tool

Once the VM is booted, you can either access it using cockpit (in the browser), or ssh into it and use the cockpit cli tool. In the latter case, make sure the right package is installed in the image (we only include cockpit-composer for now).

> What would be the advantage over building the images in the testing farm and installing them directly to the VM?

Hm, I'm not sure what exactly you are suggesting. What do you mean to do differently by "build the images in the testing farm", and what do you mean to do differently by "installing them directly into the VM"? (I want to change this, just not sure we have the same thing in mind).

> Finally, I'd prefer to write these tests in Python over Bash.

Agreed :)

My thoughts on how this should all look in the end:

We should (for the time being) use mock to build the rpms (to make it as close as possible to how the rpms wil be produced in the distros). In order for this to work independently of the CI host, I think we could simply use osbuild with the right build pipeline, by introducing a new mock assembler. The  build environment should be the distro/version we are targeting (even though mock supports cross-version builds) and the produced rpms should be dropped into the output directory. An alternative to all this would be to use the rpms from packit, but I don't know if we can build RHEL rpms there, and I don't know how that would work locally/using other infrastructure, and lastly I don't know if we can use the output of one CI as input to another.

Once we have the RPMs I think we should install them onto the image as part of the osbuild run. Either as part of the initial dnf stage (my preference) or as a separate dnf/rpm stage. I don't think we should first create an image, then boot it and install packages into it. Our whole model is about assembling images in one step and not modifying them again before using them, so we should follow that here as a way of dogfooding. The important property is that a pipeline uniquely specifies an image, so there is no way of having two images with the same pipeline where one passes the test and the other doesn't (if we installed rpms in a separate step, this would be possible, as we could (accidentally) have installed different rpms on the different images).

msehnout:
> Hm, I'm not sure what exactly you are suggesting. What do you mean to do differently by "build the images in the testing farm", and what do you mean to do differently by "installing them directly into the VM"? (I want to change this, just not sure we have the same thing in mind).

I have three approaches in my mind (I'm not saying it would work).

 1. Starting with Travis CI:
 * Run mock, build osbuild and osbuild-composer RPMs
 * Run osbuild with Ubuntu build pipeline and produce Fedora image containing both RPMs
 * Boot the image
 * Run cockpit-composer cli over SSH to verify osbuild-composer works as expected

 2. Starting with [Testing Farm](https://packit.dev/testing-farm/)
 * Run mock, build osbuild and osbuild-composer RPMs
 * Run osbuild with regular build pipeline and produce Fedora image containing both RPMs
 * Boot the image
 * Run cockpit-composer cli over SSH to verify osbuild-composer works as expected

 2. Starting with [Testing Farm](https://packit.dev/testing-farm/)
 * Build osbuild and osbuild-composer RPMs using plain rpmbuild
 * Install both RPMs to the host VM (the one we acquired from the Testing Farm)
 * Run cockpit-composer cli directly to verify osbuild-composer works as expected

teg:
This is no longer up-to-date, and the work has been continued in other places.

PR #55, opened 2019-10-28
blueprint/test: express test-cases in json

teg:
A test case is now expressed purely by dropping in a json document in
`tools/test_image_info/pipelines`. It should contain the input compose
(output format and blueprint) as well as the expected pipeline. All the
existing tests are moved over to this format.

This shares the same infrastructure as the image tests, ideally we want
to run the blueprint tests and the image tests against the same pipelines.

For now, test cases are skipped from the blueprint tests if they do not
contain a 'compose' section, and from the image tests if they do not
cotain an 'expected' section. In the future we may want to make both
mandatory.


ondrejbudai:
@teg I love it!

PR #56, opened 2019-10-28
image-info: add expected image info

teg:
This is the output of disk info ran against the images produced
by the specified pipelines.


larskarlitski:
This commit is included in https://github.com/osbuild/osbuild-composer/pull/57.

PR #57, opened 2019-10-29
Move Pipelines for ext4, partitioned-disk, tar, and vmdk towards lorax parity

larskarlitski:
Includes @teg's commit from https://github.com/osbuild/osbuild-composer/pull/56 and updates expected pipelines and image-info outputs accordingly.


internal/blueprint/ext4_output.go
```diff
@@ -5,14 +5,28 @@ import "github.com/osbuild/osbuild-composer/internal/pipeline"
...
+		"langpacks-en",
+	}
+	excludedPackages := [...]string{
+		"dracut-config-rescue",
+	}
+	p := getCustomF30PackageSet(packages[:], excludedPackages[:])
+	addF30LocaleStage(p)
+	addF30GRUB2Stage(p, b.getKernelCustomization())
 	addF30SELinuxStage(p)
+	addF30FixBlsStage(p)
```
teg: This should be fine, but I would move (in a follow-up) SELinux last, always do the fix bls stuff right after DNF.

larskarlitski: Nice catch! Fixed this one and added a commit which also flips them for openstack, vhd, and qcow2.

msehnout: Sorry for this one, I wasn't sure about proper usage of `fix-bls` stage.

PR #76, opened 2019-11-05
Additional sources tests for adding and removing a source

jkozol:
I added tests to add and then remove a source. These currently work against lorax-composer and against the basic routes I added to the api. I chose the fish repo because that is what cockpit-composer uses to test adding and removing a source.


jkozol:
Closed because #77 implements these tests better.

PR #77, opened 2019-11-05
Add user defined sources

jkozol:
Users can now add their own sources alongside the base repo that osbuild-composer is initialized with. I added CRUD functionality to the store and added routes to the api that add and delete sources. I also updated the list and info routes to include the base repo as well as the user defined sources.

This PR does not use the sources to fetch the package list. The functionality added only effects the source objects.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -25,6 +25,7 @@ type Store struct {
 	Blueprints map[string]blueprint.Blueprint `json:"blueprints"`
 	Workspace  map[string]blueprint.Blueprint `json:"workspace"`
 	Composes   map[uuid.UUID]Compose          `json:"composes"`
+	Sources    map[string]SourceConfig        `json:"sources"`
 
 	mu           sync.RWMutex // protects all fields
 	pendingJobs  chan Job
@@ -56,6 +57,15 @@ type Image struct {
 	Name string
 	Mime string
 	Size int64
+}
+
+type SourceConfig struct {
+	Name     string `json:"name"`
+	Type     string `json:"type"`
+	URL      string `json:"url"`
+	CheckGPG bool   `json:"check_gpg"`
+	CheckSSL bool   `json:"check_ssl"`
+	System   bool   `json:"system"`
 }
 
 type NotFoundError struct {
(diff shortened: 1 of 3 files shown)


internal/store/store.go
```diff
@@ -487,3 +500,49 @@ func (s *Store) GetImage(composeID uuid.UUID) (*Image, error) {
...
+
+	return true
+}
+
+func (s *Store) GetAllSources() map[string]SourceConfig {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.Sources
+}
```
jkozol: I am uncertain about what to do with this function. This will return a reference which means that the store's sources can be modified anywhere this gets called. Do you have any recommendations on how to guarantee isolation for the store's sources?

teg: Return copies. It may feel less efficient, but you are right, if we return a reference then the lock must be held while the reference is live, which we don't want to expose to the caller.

internal/store/store.go
```diff
@@ -487,3 +500,49 @@ func (s *Store) GetImage(composeID uuid.UUID) (*Image, error) {
...
+	names := make([]string, 0, len(s.Sources))
+	for name := range s.Sources {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	return names
+}
+
+func (s *Store) GetSource(name string, source *SourceConfig) bool {
```
teg: Hm, why doesn't this just return (a copy of) the SourceConfig or `nil`?

jkozol: I based this function off of the GetBlueprint function which sets the blueprint through a pointer and returns a bool if it is found. Would you recommend I change this to a copy so we don't pass a reference to the store?

jkozol:
@teg  I pushed some updates to the store's sources functions per your advice. If you approve of these changes I can squash the commits.

internal/store/store.go
teg: Yeah, I had preferred that. But let's hear with @larskarlitski (assuming he did the original?), he might see something that I miss.

teg:
This looks all good to me! Great work!

PR #78, opened 2019-11-06
blueprint: move pipeline generation into its own package

larskarlitski:
**NOTE** I called the new package `os`, but that clashes with the `os` from the standard library. I noticed that when using the new API in composer itself. I've aliased the name were this happens, but would prefer a different name, I think. Any suggestions?

@ondrejbudai I moved the customizations themselves into the new `fedora30` package, because they might be OS-dependant, but kept the types in the blueprints package. Do you think that makes sense?

---

Introduce the `os` package, which contains an interface for OS
implementations. Its main purpose is to convert a blueprint to a
OS-specific pipeline.

Also introduce the `os/fedora30` package. It is the first implementation
of the OS interface. Most of its code has been copied with minimal
modifications from the blueprint package.

The `blueprint` package is now back to serving a single purpose:
representing a weldr blueprint. It does not depend on the `pipeline`
package anymore.

Change osbuild-composer and osbuild-pipeline to use the new API,
hard-coding "fedora-30". This looks a bit weird now, but is the same
behavior as before.

All test cases now also take an "os" key in the "compose" object.


ondrejbudai:
> @ondrejbudai I moved the customizations themselves into the new `fedora30` package, because they might be OS-dependant, but kept the types in the blueprints package. Do you think that makes sense?

Ya, definitely makes sense.

internal/os/os.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (e *InvalidOutputFormatError) Error() string {
+	return e.Format
+}
+
+var registered = map[string]OS{}
+
+func New(name string) OS {
+	return registered[name]
```
ondrejbudai: I would rather put existence check with panic here.

larskarlitski: I thought panicking is a bit harsh, in case a caller would like to find out whether an OS exists. However, I never check for `nil` anywhere this is being used. Also, if we do have such a dynamically added OSs, we'd also need a function to list them.

Since this is not public API, we can think about all that later :) I'll add a panic now. Thanks for calling this out!

larskarlitski:
> Ya, definitely makes sense.

Thanks!

>  I've aliased the name were this happens, but would prefer a different name, I think. Any suggestions?

@teg suggested `distro`. I like that.

internal/os/os.go
ondrejbudai: Oh, it returns `nil` when it doesn't exist. I have to remember that interface is always pointer! :) This is kinda nonintuitive in Go. I guess returning `nil` is good. :)

ondrejbudai:
> @teg suggested `distro`. I like that.

:+1: from me, this also came to my mind.

internal/os/os.go
larskarlitski: I've added the `panic` anyway, if only to cause less confusion in the near future :)

larskarlitski:
Update to rename `os` to `distro` and add the `panic` @ondrejbudai was asking for.

PR #81, opened 2019-11-08
Local boot test - qcow2

msehnout:
This is a 2nd PR because the previous one didn't execute Travis, but it works here.

The test uses hard coded pipeline, but a blueprint is also provided. It should be used to produce the expected pipeline, but that does not work yet so I'd include it as part of another PR. Then `osbuild` is executed with the pipeline and `qemu` is used to boot it. SSH connection is attempted many times with a timeout because in environment without KVM or HVF it takes ages to boot ...


test/run
```diff
@@ -68,10 +105,61 @@ def run_test(case, store):
...
+
+    print("ssh test failure")
+    return 1
+
+
 def main(args):
+    # New part, this should be better organized
+    try:
+        flag: str = args[0]
+        if flag.startswith("--local-boot-test"):
```
msehnout: Maybe we should introduce some cmd line argument parsing, but we can do it as part of following PR.

internal/distro/distro_test.go
```diff
@@ -20,8 +21,11 @@ func TestDistro_Pipeline(t *testing.T) {
 		t.Errorf("Could not read pipelines directory '%s': %v", pipelinePath, err)
 	}
 	for _, fileInfo := range fileInfos {
+		if !strings.HasSuffix(fileInfo.Name(), "empty_blueprint.json") {
```
teg: Hm, this will break as soon as we introduce other blueprint tests. If the given blueprint doesn't currently produce the expected pipeline, then I'd just remove it (maybe put it in the commit message for documentation purposes?). The blueprint tests should ignore any files with missing composes.

test/run
```diff
@@ -68,10 +105,61 @@ def run_test(case, store):
...
+
+            return ret
+    except IndexError:
+        pass
+
+    # Lars's original code, needed for backwards compatibility
     failed = False
     with osbuild_test_store() as store:
-        for filename in args or glob.glob(f"{TEST_DIR}/cases/*.json"):
+        for filename in args or glob.glob(f"{TEST_DIR}/cases/*empty_blueprint.json"):
```
teg: This should be unecessary, no?

test/run
```diff
@@ -68,10 +105,61 @@ def run_test(case, store):
...
+                    _, output_id = run_osbuild(pl_dict, store)
+                    print("osbuild success")
+                    with qemu_boot_image(f"{store}/refs/{output_id}/{image_fname}"):
+                        ret = run_ssh_test(f"{keydir}/id_rsa")
+
+            return ret
+    except IndexError:
+        pass
+
+    # Lars's original code, needed for backwards compatibility
```
teg: Please reorganize so you don't need to have a "backwards compatible" part.

internal/distro/distro_test.go
msehnout: fixed by renaming the json object containing metadata about the test

test/run
msehnout: fixed using argparse

test/run
msehnout: fixed using argparse

msehnout:
@teg I've made all requested changes.

PR #82, opened 2019-11-09
distro: add DetectHost

larskarlitski:
lorax-composer uses the host's repositories for building images. This is
prone to accidental configuration errors and duplicates functionality
(adding custom repositories via the source API is much more explicit).

However, blueprints don't specify the distribution they're based on.
This is something they should do in the future to enable cross-distro
image builds. Until then, read `/etc/os-release` to detect the host OS
and use that as the base distro for a blueprint.

Detection works by concatenating the ID field with a `-` and the
VERSION_ID field. This mandates how additional distributions should
register themselves to the `distro` package.

Use this detection everywhere that fedora-30 was hard-coded before.


larskarlitski:
Note: please don't merge this yet (forgot to click "draft", sorry), because it will break everyone who's on Fedora 31.

larskarlitski:
Tests failed on Ubuntu, because we can't build that yet. I've changed the host detection to fall back to `fedora-30` when the host is unknown.

Also added a commit that fails the `jobqueue` test properly in such cases (it was hanging).

larskarlitski:
Another update: forgot to add a `_test.go` file.

ondrejbudai:
> Tests failed on Ubuntu, because we can't build that yet.

:rofl:

PR #83, opened 2019-11-10
Test dnf-json

larskarlitski:
I accidentally broke depsolving in the UI with my last changes to `dnf-json`. Sorry again @jkozol and thanks again @ondrejbudai for finding and fixing that issue (https://github.com/osbuild/osbuild-composer/pull/80)!

Here are some sanity tests that would have caught this issue immediately. The idea is not to check that `dnf` returns the right stuff, but only that `dnf-json` works and passes its arguments along correctly.


teg:
Hm, we should merge the fixes. But the tests I'm not sure about. dnf is not shipped on our CI, so how to test rpmmd I'm not sure about. Maybe split it out into a second PR, and let's discuss that there?

larskarlitski:
> Hm, we should merge the fixes.

I've separated them into https://github.com/osbuild/osbuild-composer/pull/88. I'll rebase this one once that's merged.

> But the tests I'm not sure about. dnf is not shipped on our CI, so how to test rpmmd I'm not sure about.

Probably we'll need to test in a container. This is not urgent though. Let's do it after we get the basics of osbuild-composer done.

larskarlitski:
Note: changing `dnf-json` to be executable introduces https://github.com/osbuild/osbuild-composer/issues/95. It probably doesn't need to be in `PATH` anyway. And we should add integration tests before doing a change like this.

larskarlitski:
This is stale now. These kinds of tests are covered by the tests in `test/cases`.

PR #90, opened 2019-11-11
Sort fetched package list

jkozol:
When dnf-json dumps the packages from the repos passed to it, it does not sort the packages. In order to properly list and search the packages, the package list is now sorted before being returned by the FetchPackageList function.


ondrejbudai:
Just small nit: Can you please use rpmmd instead of dnfadapter in the commit message? We've discarded the idea of renaming in #91

jkozol:
@ondrejbudai Renamed

PR #91, opened 2019-11-12
Projects API improvements & dnf mocking

ondrejbudai:
This PR introduces support for mocking dnf-json, adds error reporting to dnf-json and implements one missing route for projects.

I've renamed rpmmd to dnfadapter. I don't like the name rpmmd - it isn't very descriptive and it was hard to work with it when I introduced the mocking part. I'm not sure if dnfadapter is the right name. Maybe dnfwrapper?

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -38,7 +38,9 @@ func main() {
 		Metalink: "https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=x86_64",
 	}
 
-	packages, err := rpmmd.FetchPackageList([]rpmmd.RepoConfig{repo})
+	rpm := rpmmd.NewRPMMD()
+
+	packages, err := rpm.FetchPackageList([]rpmmd.RepoConfig{repo})
 	if err != nil {
 		panic(err)
 	}
@@ -56,7 +58,7 @@ func main() {
 	store := store.New(&stateFile)
 
 	jobAPI := jobqueue.New(logger, store)
-	weldrAPI := weldr.New(repo, packages, logger, store)
+	weldrAPI := weldr.New(rpm, repo, packages, logger, store)
 
 	go jobAPI.Serve(jobListener)
 	weldrAPI.Serve(weldrListener)
--- a/dnf-json
+++ b/dnf-json
@@ -71,14 +71,11 @@ if command == "dump":
 elif command == "depsolve":
     base = create_base(arguments.get("repos", {}))
     errors = []
-    for pkgspec in arguments["package-specs"]:
-        try:
-            base.install_specs([pkgspec])
-        except dnf.exceptions.MarkingError as e:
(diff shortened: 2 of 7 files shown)


cmd/osbuild-composer/main.go
```diff
@@ -5,8 +5,8 @@ import (
 	"log"
 	"os"
 
+	"github.com/osbuild/osbuild-composer/internal/dnfadapter"
```
larskarlitski: The name `rpmmd` comes from [RPM Metadata](https://en.opensuse.org/openSUSE:Standards_Rpm_Metadata), the spec which defines how repository metadata is stored. The idea was that the package to provide an API to download and query such metadata, but it grew into a bit of a mess indeed.

I agree that `rpmmd` is clunky to say and write, but it describes *what* the package does. `dnfadapter` describes *how* it's being done. I think the former is preferable, because it allows switching the *how*.

ondrejbudai: I chose the approach of mocking `dnf-json` because it also tests the marshal/unmarshal bit, which seemed a bit fragile to me - passing data from one language to another can be always tricky. Also with mocks in json format we have quite nice documentation of possible outputs of `dnf-json`. About the added complexity - I find both solutions very similar in terms of complexity.

About the `rpmmd` - I know what you mean with "*how* it's being done". Although, instead of `DNFAdapter` calling the type `RPMMD` (all caps, I guess?) sounds weird to me. Calling method `Depsolve` on object of name "metadata" is weird, metadata sounds like static blob, not something which can do depsolving.

DNF is a package manager. What about calling the struct PackageManager? It explains "what it does" and sounds like something able to depsolve and dump packages.

ondrejbudai: The package name can still be rpmmd, I haven't been able to come up with something better.

teg: Let's leave the name rpmmd for now, and we can discuss renaming separately.

teg: It is true that you'll get more testing by mocking the backend rather than the golang package. However, I think we should avoid this sort of accidental/secondary testing, it leads to unclear tests in the long run. I'd agree with Lars that mocking the golang package is the way to go, and let these tests only be about testing the weldr API.

This is, by the way, the same reason we try to only have one API test at a time, rather than combining several. It means the code is more verbose, but I think it makes it clearer what exactly it is we are testing and make it easier to know what we have and haven't covered and how to easily add to the tests in the future.

ondrejbudai:
I've pushed a new version of the PR with all discussion points solved:
- no more renaming rpmmd
- mocking of rpmmd is now done on Go level

ondrejbudai:
Don't merge this. I've figured out there's an error in dnf-json with depsolve MarkingError, will fix it after lunch and push a new version.

ondrejbudai:
@teg @larskarlitski I fixed the error and repushed, can you approve this PR?

PR #93, opened 2019-11-12
api/tests: print all differences between expected and actual values

ondrejbudai:
After this change one can immediately see all the differences when developing
or running the tests.


ondrejbudai:
Sorry, that commit was stupid. The new one should be good. This is just small non-important change but it makes my life easier. When I use these tests against lorax API and lorax returns different status code I don't get the diff for response body. In this case I need to fix the status code in the test and rerun it to get also the response body. After this commit I will get both status code and response body nicely printed next to each other.

PR #94, opened 2019-11-12
Blueprint changes

jkozol:
Every update to a blueprint creates a change. Each change contains a commit which is a hex string based off of an sha1 hash of the current time, a message describing the change, a revision which will usually be null, a timestamp, and the blueprint at the time of the change. Even when a blueprint is deleted the list of changes remains and will be updated if a blueprint with the same name is created again.


jkozol:
I squashed the api commit and test commit together. On my branch the tests commit came after the api one but because of the timestamp github has the test commit before the api functionality.

PR #106, opened 2019-11-14
Add blueprint undo route

jkozol:
Add a route to set a blueprint back to its state at a particular change. The route `blueprints/undo/:blueprint/:commit` requires the blueprint name and the commit hash for the change that the blueprint should be reverted too. Also, the commit message for the change created when a
blueprint is pushed is now passed from the api to the store's PushBlueprint funtion.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -368,6 +368,17 @@ func (s *Store) GetBlueprintCommitted(name string, bp *blueprint.Blueprint) bool
 	return true
 }
 
+func (s *Store) GetBlueprintChange(name string, commit string) *blueprint.Change {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	change, ok := s.BlueprintsChanges[name][commit]
+	if !ok {
+		return nil
+	}
+	return &change
+}
+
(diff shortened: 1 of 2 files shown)


internal/store/store.go
```diff
@@ -381,7 +392,7 @@ func (s *Store) GetBlueprintChanges(name string) []blueprint.Change {
 	return changes
 }
 
-func (s *Store) PushBlueprint(bp blueprint.Blueprint) {
```
teg: Won't this break the caller of this API?

internal/weldr/api.go
```diff
@@ -877,7 +878,8 @@ func (api *API) blueprintsNewHandler(writer http.ResponseWriter, request *http.R
 		return
 	}
 
-	api.store.PushBlueprint(blueprint)
+	commitMsg := "Recipe " + blueprint.Name + ", version " + blueprint.Version + " saved."
+	api.store.PushBlueprint(blueprint, commitMsg)
```
teg: This hunk belongs with the previous commit, but that's not a big deal. Just keep it in mind for next time.

PR #112, opened 2019-11-19
Azure image upload

msehnout:
Implementation of image upload functionality for Azure. It is a separate CLI tool for now.

### DIFF
--- a/cmd/osbuild-upload-azure/main.go
+++ b/cmd/osbuild-upload-azure/main.go
@@ -1,0 +1,148 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net/url"
+	"os"
+	"path"
+	"sync"
+
+	"github.com/Azure/azure-storage-blob-go/azblob"
+)
+
+func handleErrors(err error) {
+	if err != nil {
+		if serr, ok := err.(azblob.StorageError); ok { // This error is a Service-specific
+			switch serr.ServiceCode() { // Compare serviceCode to ServiceCodeXxx constants
+			case azblob.ServiceCodeContainerAlreadyExists:
+				// This error is not fatal
+				fmt.Println("Received 409. Container already exists")
+				return
+			}
+		}
+		// All other error causes the program to exit
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+type azureCredentials struct {
+	storageAccount   string
+	storageAccessKey string
+}
+
+type azureImageMetadata struct {
+	containerName string
+	imageName     string
+}
+
+func azureUploadImage(credentials azureCredentials, metadata azureImageMetadata, fileName string, threads int) {
+	// Create a default request pipeline using your storage account name and account key.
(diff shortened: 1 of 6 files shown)


cmd/osbuild-upload-azure/main.go
```diff
@@ -0,0 +1,147 @@
...
+func azureUploadImage(credentials azureCredentials, metadata azureImageMetadata, fileName string, threads int) {
+	// Create a default request pipeline using your storage account name and account key.
+	credential, err := azblob.NewSharedKeyCredential(credentials.storageAccount, credentials.storageAccessKey)
+	handleErrors(err)
+	p := azblob.NewPipeline(credential, azblob.PipelineOptions{})
+
+	// get storage account blob service URL endpoint.
+	URL, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net/%s", credentials.storageAccount, metadata.containerName))
+
+	// Create a ContainerURL object that wraps the container URL and a request
```
teg: I think we should (at least at first) leave it to the user to create the container (assuming this is what corresponds to an S3 bucket, basically a folder for storing objects). Creating it if it does not exist should be a follow-on feature (this was in some design document @jgiardino shared which I don't seem to have the link for at the moment).

cmd/osbuild-upload-azure/main.go
```diff
@@ -0,0 +1,147 @@
...
+	blobURL := containerURL.NewPageBlobURL(metadata.imageName)
+	_, err = blobURL.Create(ctx, stat.Size(), 0, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{})
+	handleErrors(err)
+
+	// Create control variables
+	// This channel simulates behavior of a semaphore and bounds the number of parallel threads
+	var semaphore = make(chan int, threads)
+	// The page size on Azure is 512, but it was too slow to upload page by page, this is the number
+	// of pages uploaded in single call to "blobURL.UploadPages"
+	var magic int64 = 512
```
teg: Why not use the max number of bytes (`PageBlobMaxUploadPagesBytes` (4MB)) rather than talk about pages at all?

cmd/osbuild-upload-azure/main.go
jgiardino: There was a question about this in the document I created to capture feedback. The [one reply to this question](https://docs.google.com/document/d/1JppnE3Ai0BETjJf5cLruSFNsLeOAWI_Mj1uDqPtjTi4/edit?disco=AAAADvSoa7Y) (in the context of AWS), was to support both, which eventually would be nice. However, the level of complexity this raises in the UI is significant and requires a lot of time to handle all potential error cases correctly in the UI. I had raised some concerns about this with @teg but didn't do a good job at communicating this with everyone else.

I don't have concerns about supporting the ability to create new containers in the backend tooling if the API can enable the front-end UI to only use a method for specifying existing containers (and not try to create a new container, for example, if the value specified does not exist).

teg: Thanks @jgiardino, then I remembered correctly. For now I'd just drop the create container functionality (it is currently unconditional in this patch), and then we can consider adding it back once we have figured out how it should look in the UI.

teg:
Great stuff! Thanks!

PR #113, opened 2019-11-19
upload/aws: add a sample AWS upload client

teg:
This commandline tools uploads a file to S3, as a proof of concept.

All options are mandatory. Credentials are only read from the
commandline and not from the environment or configuration files.

The tool will be refactored into a package that will be used both
by our tests and by the worker.

The next step is to add support for importing from S3 to EC2,
currently the images we produce cannot be imported as-is, so this
requires more research.

To try this out: create an S3 bucket, get your credentials and
call the tool, passing any value as `key`. Note that if the key
already exists, it will be overwritten.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/osbuild/osbuild-composer
 go 1.12
 
 require (
+	github.com/aws/aws-sdk-go v1.25.37
 	github.com/Azure/azure-storage-blob-go v0.8.0
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f
 	github.com/gobwas/glob v0.2.3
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/osbuild/osbuild-composer
 go 1.12
 
 require (
+	github.com/aws/aws-sdk-go v1.25.37
```
msehnout: Any change in `go.mod` without corresponding change in the spec file will result in broken rpm build. I think we should add it as part of this PR even though we don't yet have RPM build as part of CI.

teg: Ah, good catch! Thanks, and I agree, we should update the spec file in the same patch. I was first thinking that this tool is not going to be shipped, so we should find a way of not requiring the dep in the rpm build, but the same dependency will anyway soon land in the worker, so that is not going to be worth the effort. I'll update.

PR #115, opened 2019-11-22
api: Add stubs for compose and upload routes supported by lorax-composer

ondrejbudai:
This commit doesn't add much functionality but provides reference what we
need to implement in future. Imho it's good idea to have the routes defined
in code to make syncs across the team easier.


ondrejbudai:
Do not merge this PR, it needs more investigation.

ondrejbudai:
superseded by #116

PR #122, opened 2019-11-26
Fixes to the worker

teg:
Make sure we don't panic, but correctly return FAILED state instead. Also, set the correct AMI name for the aws target type.


internal/jobqueue/job.go
```diff
@@ -87,7 +87,7 @@ func (job *Job) Run() error {
 			}
 
 			/* TODO: communicate back the AMI */
-			_, err = a.Register(result.OutputID, options.Bucket, options.Key)
+			_, err = a.Register(options.ImageName, options.Bucket, options.Key)
```
ondrejbudai: I would prefer `ImageName` to be part of `Target` struct to make things easier. API requires each upload to have its own name, so it isn't anything special to AWS.

internal/jobqueue/job.go
```diff
@@ -21,28 +22,28 @@ type JobStatus struct {
 	Status string `json:"status"`
 }
 
-func (job *Job) Run() error {
+func (job *Job) Run() (error, []error) {
```
ondrejbudai: What about []error as a single return type? Returning `nil` if no error happened, `[]error{err}` when one error happened, `r` in other cases.

internal/jobqueue/job.go
```diff
@@ -52,50 +53,58 @@ func (job *Job) Run() error {
...
+		return err, nil
 	}
 
 	err = cmd.Wait()
 	if err != nil {
-		return err
+		return err, nil
 	}
 
+	r := []error(nil)
```
ondrejbudai: `var r []error` is imho easier to write but this is just small nit, feel free to use your approach. :)

internal/jobqueue/job.go
teg: Good point!

internal/jobqueue/job.go
teg: Yeah, good point.

internal/jobqueue/job.go
teg: I like having the array matching the targets array, so that eventually it becomes eaiser to return per-target status info to osbuild-composer and match it to the individual targets. Would be fine to ignore the first one, but I prefer to make it explicit.

PR #128, opened 2019-11-28
Provides: osbuild-composer in .spec

atodorov:
this is a small helper to enable easy switch between backends
in the existing test suite where we reference everything by name.

The current working variant is:

	yum install $BACKEND
	systemctl enable $BACKEND.socket

where BACKEND is either lorax-composer or osbuild-composer!

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -31,6 +31,8 @@ BuildRequires:  golang(github.com/gobwas/glob)
 Requires: systemd
 Requires: osbuild
 
+Provides: osbuild-composer
+
 %description
 %{common_description}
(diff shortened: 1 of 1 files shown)


atodorov:
> %package     -n %osbuild-composer

this will change the RPM name if I'm not mistaken. And I don't know whether or not we want to do this. Golang packages have a certain naming convention so maybe not a good idea to break that.

atodorov:
@teg can you cherry-pick this onto the rhel8.2-branch or we need to go though the standard process and open another PR for this ?

teg:
No need to open another PR. We rebase on top of master and take care of keeping the branches in sync.

PR #129, opened 2019-11-28
Introduce Azure upload test

msehnout:
The test is very simple, it first upload a file with random content and then downloads it back and compare hashes.


msehnout:
This fails because the environment variables in CI are not available to forks.

msehnout:
It is ready to be merged. I made the test optional in case in has non of the configuration variables (like in case of running in Travis from a fork). It will run once it is merged into our main repo.

PR #134, opened 2019-12-02
add makefile targets for rpm build

msehnout:
it makes it possible to use the same workflow as with osbuild


Makefile
```diff
@@ -1,3 +1,6 @@
+PACKAGE_NAME = osbuild-composer
+VERSION = 3
```
ondrejbudai: Current version is 4, can you please fix it?

msehnout: Sure, I parse it from spec file now.

PR #135, opened 2019-12-02
weldr: add support for toml blueprints

larskarlitski:


internal/weldr/api.go
```diff
@@ -701,11 +704,44 @@ func (api *API) blueprintsInfoHandler(writer http.ResponseWriter, request *http.
...
+	if format == "json" || format == "" {
+		json.NewEncoder(writer).Encode(reply{
+			Blueprints: blueprints,
+			Changes:    changes,
+			Errors:     []responseError{},
+		})
+	} else if format == "toml" {
+		// lorax concatenates multiple blueprints with `\n\n` here,
+		// which is never useful. Deviate by only returning the first
+		// blueprint.
```
teg: Wouldn't it be just as easy to just concatenate them all. Not that I think anyone relies on this, but I could imagine tools asking for many and considering only the first, for instance.

internal/weldr/api.go
```diff
@@ -1030,21 +1066,29 @@ func (api *API) blueprintsNewHandler(writer http.ResponseWriter, request *http.R
 	}
 
 	contentType := request.Header["Content-Type"]
-	if len(contentType) != 1 || contentType[0] != "application/json" {
+	if len(contentType) != 1 {
```
teg: Would it make sense to make a helper out of this patch, that takes a `request` and decodes it into a generic type?

internal/weldr/api.go
bcl: Nothing should be relying on that, it was basically a hack to be able to support both output formats with multiple blueprints listed. I agree that an error here is the right thing to do.

PR #136, opened 2019-12-03
api: fix compose routes returning null instead of empty array

ondrejbudai:
Cockpit-composer really doesn't like that


ondrejbudai:
I accidentally broke it in #131, so the answer is: It didn't work. It often confuses me slices can be both empty and nil. :(

PR #139, opened 2019-12-04
weldr: bump blueprint version

larskarlitski:
According to the same rules as lorax-compposer:

  https://weldr.io/lorax/lorax-composer.html#blueprints

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -6,11 +6,14 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/store/store.go
```diff
@@ -330,6 +333,21 @@ func (s *Store) GetBlueprintChanges(name string) []blueprint.Change {
 	return changes
 }
 
+func bumpVersion(str string) string {
+	v := [3]uint64{}
+	fields := strings.Split(str, ".")
+	for i := 0; i < len(fields); i++ {
+		if n, err := strconv.ParseUint(fields[i], 10, 64); err == nil {
+			v[i] = n
```
ondrejbudai: This will panic if `len(fields) > 3`. It isn't likely but we are not doing any version validation on new blueprints therefore it's possible.

larskarlitski: You're completely right and it's a big deal. I've pushed a fix that ignores such version strings and added a test.

internal/store/store.go
```diff
@@ -330,6 +333,21 @@ func (s *Store) GetBlueprintChanges(name string) []blueprint.Change {
...
 
+func bumpVersion(str string) string {
+	v := [3]uint64{}
+	fields := strings.SplitN(str, ".", 3)
+	for i := 0; i < len(fields); i++ {
+		if n, err := strconv.ParseUint(fields[i], 10, 64); err == nil {
+			v[i] = n
+		} else {
+			// don't touch strings with invalid versions
+			return str
```
bcl: In lorax-composer it runs the versions through semver.Version() in order to check their validity, raising an error if they are not compatible. I'm worried that by passing through invalid versions that they may end up in the saved blueprint and cause problems in the future.

internal/store/store.go
```diff
@@ -352,6 +370,12 @@ func (s *Store) PushBlueprint(bp blueprint.Blueprint, commitMsg string) {
 			s.BlueprintsChanges[bp.Name] = make(map[string]blueprint.Change)
 		}
 		s.BlueprintsChanges[bp.Name][commit] = change
+
```
bcl: I think that this isn't validating new blueprint versions, just passing them through. This could result in problems for anything else expecting a valid semver version in the blueprint.

internal/store/store.go
larskarlitski: It does indeed! Thanks for the pointer. I think we should do the same here (filed as https://github.com/osbuild/osbuild-composer/issues/141). Since this PR doesn't regress on that, but rather moves a step in the right direction (being more compatible with lorax-composer by bumping the version), and it unblocks Red Hat QE, I'll merge it anyway.

PR #140, opened 2019-12-04
Produce correct .toml files

larskarlitski:


larskarlitski:
> Shouldn't we `omitempty` them?

Good point. Fixed and added a comment why we're not setting `omitempty` for json. Thanks!

larskarlitski:
Forgot updating the test, which assumed empty arrays there. Pushed a fix.

PR #150, opened 2019-12-07
distro: set the repository checksum dynamically and use latest RHEL nightly

larskarlitski:
See commits.


internal/rpmmd/repository.go
```diff
@@ -172,26 +171,32 @@ func NewRPMMD() RPMMD {
...
 	var arguments = struct {
 		PackageSpecs []string     `json:"package-specs"`
 		Repos        []RepoConfig `json:"repos"`
 	}{specs, repos}
-	var dependencies []PackageSpec
-	err := runDNF("depsolve", arguments, &dependencies)
-	return dependencies, err
+	var reply struct {
+		Checksums    map[string]string `json:"checksums"`
+		Dependencies []PackageSpec     `json:"dependencies"`
```
ondrejbudai: You named this field "packages" in `dnf-json`, this won't work.

larskarlitski: oh wow thanks. Good catch!

Our tests didn't detect this, because that only breaks the UI. We probably want to gate on those tests in the future.

ondrejbudai: Well, we don't have tests for /blueprint/depsolve which was broken by this change. Adding them would solve it. :)

larskarlitski: You're completely right of course. I've added a test :)

test/cases/ami_empty_blueprint.json
```diff
@@ -18,9 +21,9 @@
...
-                  "checksum": "sha256:9f596e18f585bee30ac41c11fb11a83ed6b11d5b341c1cb56ca4015d7717cb97"
+                  "checksum": "test:foo"
```
ondrejbudai: I figured out this change broke the tests, it probably needs a revert. :)

larskarlitski: Actually a rebase to @teg's test cleanup broke the tests. We didn't seem to be running those through osbuild before, so the actual checksum didn't matter. I put a distinctive one to easily spot that it's passed through correctly.

I've pushed a fix that uses the correct checksums. This means though that we cannot run the tests on changing repositories :/

PR #152, opened 2019-12-09
Compose/delete

ondrejbudai:
see invididual commits


internal/jobqueue/job.go
```diff
@@ -99,10 +99,13 @@ func (job *Job) Run(d distro.Distro) (*store.Image, error, []error) {
...
-			cp.Stderr = os.Stderr
-			cp.Stdout = os.Stdout
-			err = cp.Run()
+			err = runCommand("cp", "-a", "-L", "/var/cache/osbuild-composer/store/refs/"+result.OutputID+"/.", options.Location)
+			if err != nil {
+				r = append(r, err)
+				continue
+			}
+
+			err = runCommand("chown", "-R", "_osbuild-composer:_osbuild-composer", options.Location)
```
teg: This is correct, though in a follow-up, we may want to simply move the directory creation to composer, rather than the worker.

PR #154, opened 2019-12-09
image-info: read initrds and emit kmod list

gicmo:
Iterate over all the found boot entries, parse the initramfs files and report the contained kernel modules. This is important info because depending on the included kmods images might not be able to boot on certain systems (i.e. missing block device driver kmod).
The output will be something like contain the filetype, and kmods per found initrd, a la:

```
"initrd": [
  {
    "initramfs-5.0.9-301.fc30.aarch64.img": {
      "filetype": "gzip compressed data, max compression, ...",
      "kmods": [
        "8250_exar.ko.xz",
        "8250_moxa.ko.xz",
        ...
	]}
  }]
```


gicmo:
@larskarlitski and I figured out the hacky whacky way to list the kernel modules via `cpio` contained in the initrd. It is not pretty...

msehnout:
Awesome! I do this by hand so I'll be glad to see it as part of image-info. :-)

larskarlitski:
Force pushed to add `initrd` output to all test cases and rebase master to get @teg's test cleanups.

larskarlitski:
Hm, the list of returned modules is different on my machine and here on travis. I assume because we're still running dracut in host-only mode?

teg:
Should we, at least for now. Unconditionally useg `dracut-config-generic` everywhere?

larskarlitski:
> Should we, at least for now, unconditionally useg `dracut-config-generic` everywhere?

We should indeed. Now that your work adding this has landed (https://github.com/osbuild/osbuild-composer/pull/170), I rebased this branch. I've run the `qcow2` test locally and the modules seem to be correct for that. Letting another computer check the rest though :wink:

larskarlitski:
Update: add an additional commit which adds `dracut-config-generic` to `fedora30` as well.

larskarlitski:

PR #155, opened 2019-12-10
composer: add provisional multi-arch support

teg:
The pipeline generation now takes the architecture as an argument.
Currently only x86_64 is supported. The architecture is detected
at start-up, and passed down to each pipeline translation.

For osbuild-pipeline we now requrie the architecture to be passed
in.

### DIFF
--- a/distribution/osbuild-worker@.service
+++ b/distribution/osbuild-worker@.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=OSBuild Composer Worker (%i)
+Requires=osbuild-composer.socket
 After=multi-user.target osbuild-composer.socket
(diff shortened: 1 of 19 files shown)


distribution/osbuild-worker@.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=OSBuild Composer Worker (%i)
+Requires=osbuild-composer.socket
```
larskarlitski: Is it ok to depend on both sockets here?

teg: We could split it up, but I don't see why it would not be ok?

PR #171, opened 2019-12-14
test/run: print more helpful output

teg:
Make sure we catch expected errors and fail gracefully. Also, make
sure the output id is printed on successufl osbuild run so it can
be introspected from outside the test suite.

Handle the case where osbuild suceeds but does not return an
output_id.

### DIFF
--- a/test/run
+++ b/test/run
@@ -116,10 +116,12 @@ def nspawn_boot_image(image_file, name):
 @contextlib.contextmanager
 def nspawn_boot_archive(image_file, name):
     with tempfile.TemporaryDirectory(dir="/var/tmp") as dir:
-        subprocess.run(["tar", "xf", image_file], cwd=dir)
-        cmd = ["systemd-nspawn", "--boot", "--register=no", "-M", name, "--directory", "."]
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -116,10 +116,12 @@ def nspawn_boot_image(image_file, name):
 @contextlib.contextmanager
 def nspawn_boot_archive(image_file, name):
     with tempfile.TemporaryDirectory(dir="/var/tmp") as dir:
-        subprocess.run(["tar", "xf", image_file], cwd=dir)
-        cmd = ["systemd-nspawn", "--boot", "--register=no", "-M", name, "--directory", "."]
+        cmd = ["tar", "xf", image_file]
+        print(f"extracting image: {' '.join(cmd)}")
+        subprocess.run(cmd, cmd=dir)
```
larskarlitski: Tests are failing: this should be `cwd=dir`.

larskarlitski:
This, including the fix I mentioned in my last comment, was included in https://github.com/osbuild/osbuild-composer/pull/157.

PR #180, opened 2019-12-17
weldr/api: add missing compose/cancel route

teg:
This was accidentally left out. It is the same as the delete route,
except that it works in the WAITING or RUNNING states rather than
FINISHED or FAILED, and it takes a single UUID, rather than a list.

Resolves rhbz#1783991.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -461,6 +461,24 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 	return nil
 }
 
+func (s *Store) CancelCompose(id uuid.UUID) error {
+	return s.change(func() error {
+		compose, exists := s.Composes[id]
+
+		if !exists {
+			return &NotFoundError{}
+		}
+
+		if compose.QueueStatus != "WAITING" && compose.QueueStatus != "RUNNING" {
+			return &InvalidRequestError{fmt.Sprintf("Compose %s is not in WAITING or RUNNING.", id)}
+		}
+
+		delete(s.Composes, id)
+
+		return nil
+	})
+}
+
 func (s *Store) DeleteCompose(id uuid.UUID) error {
 	return s.change(func() error {
 		compose, exists := s.Composes[id]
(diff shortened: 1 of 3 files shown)


internal/weldr/api.go
```diff
@@ -1280,6 +1281,68 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
...
+				})
+			}
+		}
+	}
+
+	if len(errors) > 0 {
+		reply := struct {
+			Errors []composeCancelError `json:"errors"`
+		}{errors}
+		json.NewEncoder(writer).Encode(reply)
```
ondrejbudai: Lorax sets BadRequest status code in case of errors.

internal/weldr/api.go
```diff
@@ -1280,6 +1281,68 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
...
+	}
+
+	type composeCancelError struct {
+		ID  string `json:"id"`
+		Msg string `json:"msg"`
+	}
+
+	uuidString := params.ByName("uuid")
+
+	errors := []composeCancelError{}
```
ondrejbudai: Do we need to use array of errors when we can have maximum of 1 error?

internal/store/store.go
```diff
@@ -461,6 +461,24 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
+
+		if !exists {
+			return &NotFoundError{}
+		}
+
+		if compose.QueueStatus != "WAITING" && compose.QueueStatus != "RUNNING" {
+			return &InvalidRequestError{fmt.Sprintf("Compose %s is not in WAITING or RUNNING.", id)}
+		}
+
+		delete(s.Composes, id)
```
ondrejbudai: After many layers of abstraction, this will crash the worker when it tries to update the status of non-existing compose. However, it will be restarted by systemd. From osbuild-worker@.service:
```
Restart=on-failure
RestartSec=10s
```

I just wanna double check whether we are OK with depending on systemd restarting our stuff when user cancels a job. It surely works but looks a bit hacky to me.

internal/weldr/api.go
teg: Ugh, good catch, thanks!

internal/weldr/api.go
teg: We could make it a pointer instead. Probably should :)

internal/store/store.go
teg: Hmmmm..... the intention was that it should restart, but not crash. I'll fix this.

atodorov:
@teg the missing `compose cancel` is affecting composer-cli tests with the new backend.

teg:
I'm planning to redo this as soon as the prerequisites are in.

teg:
@larskarlitski is working on a replacement for this.

PR #184, opened 2019-12-20
Allow for user defined image size

jkozol:
When starting a compose users can pass in an image size. The resulting image will then be of this size. If no image size is passed a default size will be used.

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -58,7 +58,7 @@ func main() {
 		panic(err.Error())
 	}
 
-	pipeline, err := d.Pipeline(blueprint, nil, checksums, archArg, format)
+	pipeline, err := d.Pipeline(blueprint, nil, checksums, archArg, format, 0)
 	if err != nil {
 		panic(err.Error())
 	}
--- a/internal/distro/fedora30/distro.go
+++ b/internal/distro/fedora30/distro.go
@@ -36,12 +36,15 @@ type output struct {
 	DisabledServices []string
 	KernelOptions    string
 	Bootable         bool
-	Assembler        func(uefi bool) *pipeline.Assembler
+	DefaultSize      uint64
+	Assembler        func(uefi bool, size uint64) *pipeline.Assembler
 }
 
 const Name = "fedora-30"
 
 func New(confPaths []string) *Fedora30 {
+	const GigaByte = 1024 * 1024 * 1024
+
 	r := Fedora30{
 		arches:  map[string]arch{},
 		outputs: map[string]output{},
@@ -110,7 +113,10 @@ func New(confPaths []string) *Fedora30 {
 		},
 		KernelOptions: "ro no_timer_check console=ttyS0,115200n8 console=tty1 biosdevname=0 net.ifnames=0 console=ttyS0,115200",
 		Bootable:      true,
-		Assembler:     func(uefi bool) *pipeline.Assembler { return r.qemuAssembler("raw.xz", "image.raw.xz", uefi) },
+		DefaultSize:   6 * GigaByte,
+		Assembler: func(uefi bool, size uint64) *pipeline.Assembler {
+			return r.qemuAssembler("raw.xz", "image.raw.xz", uefi, size)
+		},
 	}
 
 	r.outputs["ext4-filesystem"] = output{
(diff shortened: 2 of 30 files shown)


internal/distro/fedora30/distro.go
```diff
@@ -129,7 +135,8 @@ func New(confPaths []string) *Fedora30 {
 		},
 		KernelOptions: "ro biosdevname=0 net.ifnames=0",
 		Bootable:      false,
-		Assembler:     func(uefi bool) *pipeline.Assembler { return r.rawFSAssembler("filesystem.img") },
+		DefaultSize:   3221225472,
```
teg: `3 * GigaByte` ?

internal/distro/rhel82/distro.go
```diff
@@ -187,7 +189,8 @@ func New(confPaths []string) *RHEL82 {
 		},
 		Bootable:      false,
 		KernelOptions: "ro net.ifnames=0",
-		Assembler:     func(uefi bool) *pipeline.Assembler { return r.rawFSAssembler("filesystem.img") },
+		DefaultSize:   3221225472,
```
teg: `3 * GigaByte` ?

internal/distro/fedora30/distro.go
teg: Don't know how to comment on the commit message, so leaving this here: please flesh it out to make it clear what you are adding (allowing size to be passed in) and changing (per output-type default sizes (as opposed to?), for when no size is passed in).

cmd/osbuild-pipeline/main.go
```diff
@@ -58,7 +58,7 @@ func main() {
 		panic(err.Error())
 	}
 
-	pipeline, err := d.Pipeline(blueprint, nil, checksums, archArg, format)
+	pipeline, err := d.Pipeline(blueprint, nil, checksums, archArg, format, 0)
```
teg: This needs to be part of the previous commit, or that will not compile, right?

internal/pipeline/tar_assembler.go
```diff
@@ -6,16 +6,18 @@ package pipeline
...
+	Size        uint64 `json:"size"`
 	Compression string `json:"compression,omitempty"`
 }
 
 func (TarAssemblerOptions) isAssemblerOptions() {}
 
 // NewTarAssemblerOptions creates a new TarAssemblerOptions object, with the
 // mandatory options set.
-func NewTarAssemblerOptions(filename string) *TarAssemblerOptions {
+func NewTarAssemblerOptions(filename string, size uint64) *TarAssemblerOptions {
```
teg: Hm, this won't compile, will it? You need to update the callers at the same time (ideally without any functional change).

internal/store/store.go
```diff
@@ -463,7 +463,7 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 		repos = append(repos, source.RepoConfig())
 	}
 
-	pipeline, err := s.distro.Pipeline(bp, repos, checksums, arch, composeType)
+	pipeline, err := s.distro.Pipeline(bp, repos, checksums, arch, composeType, size)
```
teg: I think you should change this to pass 0 in the commit that changes the interface, then update in this commit (introducing the size parameter to PushCompose()) to pass on the real size.

internal/store/store.go
```diff
@@ -443,7 +443,7 @@ func (s *Store) GetComposeResult(id uuid.UUID) (io.ReadCloser, error) {
 	return os.Open(*s.stateDir + "/outputs/" + id.String() + "/result.json")
 }
 
-func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checksums map[string]string, arch, composeType string, uploadTarget *target.Target) error {
+func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checksums map[string]string, arch, composeType string, size uint64, uploadTarget *target.Target) error {
```
teg: You need to update the callers at the same time (ideally just passing in zero), to make sure things compile.

internal/weldr/api.go
```diff
@@ -1246,6 +1248,11 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
 
 	bp := api.store.GetBlueprintCommitted(cr.BlueprintName)
 
+	size := cr.Size
+	if size%MegaByte != 0 {
+		size = (size/MegaByte + 1) * MegaByte
```
teg: This needs some discussion / comment. The alternative is to refuse the compose if we don't like what is passed (we do that further down the stack). I think you are right that this is what we want, but would need some discussion as to why at this point?

test/cases/ami_empty_blueprint.json
```diff
@@ -130,7 +130,7 @@
       "options": {
         "format": "raw.xz",
         "filename": "image.raw.xz",
-        "size": 3222274048,
+        "size": 6442450944,
```
teg: Why are the sizes changing? Should be some argument why the new sizes are right wherever the change was introduced. Also, this needs to be squashed with the commit that introduce the change to make sure the tests keep passing.

internal/distro/fedora30/distro.go
larskarlitski: > `3 * GigaByte` ?

This seems to be a go-ism. I've seen this done for time and kind of liked the idea: https://gobyexample.com/timeouts

PR #185, opened 2019-12-23
spec: support both Fedora and RHEL

teg:


golang-github-osbuild-composer.spec
```diff
@@ -39,9 +45,19 @@ Provides: lorax-composer
 %{common_description}
 
 %prep
-%goprep
+%forgeautosetup -p1
 
 %build
+GO_BUILD_PATH=$PWD/_build
```
msehnout: wow, this is considerably more complicated, can we have some comments explaining the approach here?

larskarlitski: I agree it should be documented, but this is only moving code from the `rhel-8.2` branch over here.

PR #186, opened 2019-12-28
integration tests: add subpackage with integration tests

teg:
The overall idea is to ship integration tests as a subpackage to be installed and run on the target machine, not requiring any additional dependencies or framework to be installed on the target, and allowing the test suite to be triggered form the outside.

See individual commits for details.


ondrejbudai:
`673ad7d6` - Imho we should merge this asap (also it's kinda unrelated), can you cherrypick it on a different branch.

golang-github-osbuild-composer.spec
```diff
@@ -100,16 +108,17 @@ export GOPATH=$PWD/_build:%{gopath}
 %license LICENSE
 %doc README.md
 %{_libexecdir}/osbuild-composer/
+%{_datadir}/osbuild-composer/
 %{_unitdir}/*.{service,socket}
 %{_sysusersdir}/osbuild-composer.conf
 
-%changelog
```
ondrejbudai: Please, don't kill the changelog. :)

internal/rpmmd/repository.go
```diff
@@ -106,6 +106,34 @@ func (err *DNFError) Error() string {
...
+	path := "/repositories/" + distro + ".json"
+
+	for _, confPath := range confPaths {
+		f, err = os.Open(confPath + path)
+		if os.IsNotExist(err) {
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+	}
```
larskarlitski: I think this loop is wrong. It doesn't stop when it finds the first repo file. Did you mean to add an uncondontional `break` at the end?

repositories/fedora-30.json
```diff
@@ -0,0 +1,18 @@
+{
+    "x86_64": [
+        {
+            "id": "fedora",
+            "name": "Fedora 30",
```
larskarlitski: Unrelated to this patch, but why are we including the `name` everywhere? It's not really used, is it?

cmd/osbuild-tests/main.go
```diff
@@ -0,0 +1,249 @@
...
+	err := json.Unmarshal(rawReply, &reply)
+	if err != nil {
+		log.Fatalf("Unexpected reply: " + err.Error())
+	}
+	if reply.Status != true {
+		log.Fatalf("Unexpected status %v", reply.Status)
+	}
+}
+
+func runCommand(quiet bool, command ...string) json.RawMessage {
```
larskarlitski: Do you mind naming this `runComposerCLI` or something more descriptive?

cmd/osbuild-tests/main.go
```diff
@@ -0,0 +1,249 @@
...
+	command = append([]string{"--json"}, command...)
```
larskarlitski: @bcl mentioned that `--json` doesn't apply to all commands and is not guaranteed to stay the same. We should probably clarify before depending on it to heavily.

cmd/osbuild-tests/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	contents, err := ioutil.ReadAll(stdout)
+	if err != nil {
+		log.Fatalf("Could not read stdout from command: " + err.Error())
+	}
+
+	if len(contents) != 0 {
+		err = json.Unmarshal(contents, &result)
+		if err != nil {
+			// We did not get JSON, try interpreting it as TOML
```
larskarlitski: There are some commands that output neither `json` nor `toml`, like `status show`. We might need to make the output format an explicit parameter to this function at some point (no need to do it now, though).

golang-github-osbuild-composer.spec
larskarlitski: Hehe :) What's the use of maintaining a changelog upstream if most entries are `New upstream release`?

internal/distro/distro.go
```diff
@@ -40,25 +40,37 @@ type Distro interface {
...
-var registered map[string]Distro
+type DistroRegister struct {
+	distros map[string]Distro
+}
 
-func init() {
-	registered = map[string]Distro{
-		fedora30.Name: fedora30.New(),
-		rhel82.Name:   rhel82.New(),
+func NewDistroRegister(confPaths []string) *DistroRegister {
```
larskarlitski: I really prefer `NewRegistry` (`Register` → `Registry` and no stuttering), but really don't worry changing it everywhere if that's too much hassle. It's just a preference, after all.

internal/distro/distro_test.go
```diff
@@ -42,7 +42,8 @@ func TestDistro_Pipeline(t *testing.T) {
 			continue
 		}
 		t.Run(tt.Compose.OutputFormat, func(t *testing.T) {
-			d := distro.New(tt.Compose.Distro)
+			distros := distro.NewDistroRegister([]string{"../.."})
+			d := distros.GetDistro(tt.Compose.Distro)
```
larskarlitski: We're now creating the distro objects every time they're being accessed. I think that's idempotent right now, so we are fine, but we might want to move the register object into something that's passed around.

golang-github-osbuild-composer.spec
msehnout: Please kill the changelog :D

msehnout:
I like this PR.

One question though: I'd like to see the tests as part of our CI, do you have any plan for this?

repositories/fedora-30.json
teg: Yeah, I was going to ask you the same question ;) Should look into that (separately).

cmd/osbuild-tests/main.go
teg: Good point! Done.

cmd/osbuild-tests/main.go
teg: Yeah, I thought I'd start off trying to make everything JSON and generalize if and when we run into issues. I expect this needs to be refactored eventually, but seems to get us pretty far for now.

internal/distro/distro.go
teg: Ah, you are totally right. Will fix.

internal/distro/distro_test.go
teg: Agreed, for now this tries to minimize the changes, but should be followed by a clean-up.

teg:
@msehnout yeah, the aim is to get this into CI. I'm not 100% sure how to hook it all up. I have some local scripts that do it how I imagine it should be as a proof of concept, but still need to figure out how to make that a reality. Will be the next step :)

cmd/osbuild-tests/main.go
teg: Yeah, see below.

internal/rpmmd/repository.go
teg: Ehhh... Right. I want to say I tested this, but obviously I can't have.

teg:
Addressed the comments, except for the changelog, that's still to be resolved.

teg:
Reverted the changelog. We can discuss that separately.

cmd/osbuild-tests/main.go
larskarlitski: Yeah. This was (as most of these comments) just me thinking out loud :wink:

PR #189, opened 2020-01-03
update RHEL8.2 repo to a new one (the old one is gone)

msehnout:


ondrejbudai:
What about using `http://download-ipv4.eng.brq.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.2.0/`?

msehnout:
> What about using `http://download-ipv4.eng.brq.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.2.0/`?

That was my first idea as well, but I think we don't like repositories that change content which would be the case for this one.

ondrejbudai:
As it's only nightly repository and we are using it only for development, I don't care using changing repository. :)

msehnout:
> As it's only nightly repository and we are using it only for development, I don't care using changing repository. :)

You are probably right. As opposed to Fedora, we don't have any test that would compare output images with known image-info, so it should be fine even though it changes.

@teg WDYT?

larskarlitski:
@teg made overriding repositories from the file system possible as part of #186: https://github.com/osbuild/osbuild-composer/pull/186/commits/32c6a40dd8219fe13e993c86b7f1cfdf46319256

He also changed it to the latest nightly (the symlink). I think we don't need this one.

Sorry!

PR #191, opened 2020-01-09
upload/aws: Add basic S3 upload test

ondrejbudai:
The test is very simple, it just uploads object to S3, downloads it back and
compares them.


msehnout:
One suggestion: I know we don't usually do this, but I'd create a new branch in this repository so that the test actually runs before we merge this. You can easily delete the branch afterwards.

ondrejbudai:
Branch created and the tests are passing. Thank you @msehnout for the tip!

ondrejbudai:
Done! I wasn't sure if I want to test only the 3rd party API or also our code. I guess the latter makes more sense. :-)

PR #197, opened 2020-01-13
Rcm api

### LINKED ISSUE #192: RCM API
Introduce a limited API for RCM to use to push composes to osbuild-composer.

Acceptance criteria:

- [x] Introduce a dedicated TCP socket for this API.
- [x] Implement a method to push a new compose, specifying the architectures to build for and the repositories to use. It would be ok to silently drop requests for unsupported (non-native) architecture, until we have support for more. For the time being assume the qcow2 output type and the local upload target.
- [x] Implement a method to get the status of a compose (waiting/running/finished/failed)
- [x] Make the new API optional (ship the systemd socket unit in a sub-package and gracefully handle when not installed, possibly needing further discussion?)
- [ ] Add support for the koji upload target, and use that exclusively in stead of the local one. This requires changes both to the input to the endpoint pushing a new compose, and the output of the status endpoint in finished state.
- [ ] Return per-job status from GET /v1/compose/

Optional:
- [ ] Support kerberos authentication in the same way brew etc does.
- [ ] Extend the status API to optionally block until the compose is no longer waiting/running.

msehnout:
This issue fixes #192

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/osbuild/osbuild-composer/internal/distro"
 	"github.com/osbuild/osbuild-composer/internal/jobqueue"
+	"github.com/osbuild/osbuild-composer/internal/rcm"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/store"
 	"github.com/osbuild/osbuild-composer/internal/weldr"
@@ -41,8 +42,8 @@ func main() {
 		log.Fatalf("Could not get listening sockets: " + err.Error())
 	}
 
-	if len(listeners) != 2 {
-		log.Fatalf("Unexpected number of listening sockets (%d), expected 2", len(listeners))
+	if len(listeners) != 2 || len(listeners) != 3 {
+		log.Fatalf("Unexpected number of listening sockets (%d), expected 2 or 3", len(listeners))
 	}
 
 	weldrListener := listeners[0]
@@ -67,5 +68,13 @@ func main() {
 	weldrAPI := weldr.New(rpm, currentArch(), distribution, logger, store)
 
 	go jobAPI.Serve(jobListener)
+
+	// Optionally run RCM API as well as Weldr API
+	if len(listeners) == 3 {
+		rcmListener := listeners[2]
+		rcmAPI := rcm.New(logger, store)
+		go rcmAPI.Serve(rcmListener)
+	}
+
 	weldrAPI.Serve(weldrListener)
 }
--- a/internal/rcm/api.go
+++ b/internal/rcm/api.go
@@ -1,0 +1,183 @@
+// Package rcm provides alternative HTTP API to Weldr.
+// It's primary use case is for the RCM team. As such it is driven solely by their requirements.
+package rcm
+
+import (
+	"encoding/json"
+	"log"
+	"net"
+	"net/http"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/julienschmidt/httprouter"
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/store"
+)
+
+// API encapsulates RCM-specific API that is exposed over a separate TCP socket
+type API struct {
+	logger *log.Logger
+	store  *store.Store
+	router *httprouter.Router
+}
+
+// New creates new RCM API
+func New(logger *log.Logger, store *store.Store) *API {
+	api := &API{
+		logger: logger,
+		store:  store,
+		router: httprouter.New(),
+	}
+
+	api.router.RedirectTrailingSlash = false
+	api.router.RedirectFixedPath = false
+	api.router.MethodNotAllowed = http.HandlerFunc(methodNotAllowedHandler)
+	api.router.NotFound = http.HandlerFunc(notFoundHandler)
+
+	api.router.POST("/v1/compose", api.submit)
+	api.router.GET("/v1/compose/:uuid", api.status)
+
+	return api
+}
+
+// Serve serves the RCM API over the provided listener socket
+func (api *API) Serve(listener net.Listener) error {
+	server := http.Server{Handler: api}
+
+	err := server.Serve(listener)
+	if err != nil && err != http.ErrServerClosed {
+		return err
+	}
+
+	return nil
+}
+
+// ServeHTTP logs the request, sets content-type, and forwards the request to appropriate handler
+func (api *API) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	if api.logger != nil {
+		log.Println(request.Method, request.URL.Path)
+	}
+
+	writer.Header().Set("Content-Type", "application/json; charset=utf-8")
+	api.router.ServeHTTP(writer, request)
+}
+
+func methodNotAllowedHandler(writer http.ResponseWriter, request *http.Request) {
+	writer.WriteHeader(http.StatusMethodNotAllowed)
+}
+
+func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
+	writer.WriteHeader(http.StatusNotFound)
+}
+
+func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
+	// Check some basic HTTP parameters
+	contentType := request.Header["Content-Type"]
+	if len(contentType) != 1 || contentType[0] != "application/json" {
+		writer.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	type Repository struct {
+		URL      string `json:"url"`
+		Checksum string `json:"checksum"`
+	}
+
+	// JSON structure expected from the client
+	var composeRequest struct {
+		Distribution  string       `json:"distribution"`
+		ImageTypes    []string     `json:"image_types"`
+		Architectures []string     `json:"architectures"`
+		Repositories  []Repository `json:"repositories"`
+	}
+	// JSON structure with error message
+	var errorReason struct {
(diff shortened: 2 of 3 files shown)


internal/rcm/api.go
```diff
@@ -0,0 +1,148 @@
...
+		store:  store,
+		router: httprouter.New(),
+	}
+
+	api.router.RedirectTrailingSlash = false
+	api.router.RedirectFixedPath = false
+	api.router.MethodNotAllowed = http.HandlerFunc(methodNotAllowedHandler)
+	api.router.NotFound = http.HandlerFunc(notFoundHandler)
+
+	api.router.POST("/submit", api.submit)
```
ondrejbudai: I recommend using
```
POST /compose
GET /compose/:uuid
```
This should more in line with general REST principles.

internal/rcm/api.go
```diff
@@ -0,0 +1,148 @@
...
+	api.router.GET("/status/:uuid", api.status)
```
ondrejbudai: Do we want versioning? `GET /v1/compose/:uuid`?

internal/rcm/api.go
```diff
@@ -0,0 +1,148 @@
...
+		Architectures []string `json:"architectures"`
+		Repositories  []string `json:"repositories"`
+	}
+	// Parse and verify the structure
+	var composeRequest ComposeRequest
+	decoder := json.NewDecoder(request.Body)
+	decoder.DisallowUnknownFields()
+	err := decoder.Decode(&composeRequest)
+	if err != nil || composeRequest.Distribution == "" || len(composeRequest.Architectures) == 0 || len(composeRequest.Repositories) == 0 {
+		writer.WriteHeader(http.StatusBadRequest)
```
ondrejbudai: It would be cool to provide more information about what was wrong with the request.

internal/rcm/api_test.go
```diff
@@ -0,0 +1,129 @@
...
+		{"GET", "/submit", ``, "", http.StatusMethodNotAllowed, ``},
+		{"POST", "/submit", `{"status":"RUNNING"}`, "application/json", http.StatusBadRequest, ``},
+		{"POST", "/submit", `{"status":"RUNNING"}`, "text/plain", http.StatusBadRequest, ``},
+		{"POST", "/submit", `{"distribution": "fedora-30", "architectures":["x86_64"], "repositories": ["aaa"]}`, "application/json", http.StatusOK, `{"compose_id":".*"}`},
+		{"POST", "/status", `{"status":"RUNNING"}`, "application/json", http.StatusNotFound, ``},
+		{"POST", "/status/111-222-333", `{"status":"RUNNING"}`, "application/json", http.StatusMethodNotAllowed, ``},
+		{"GET", "/status/7802c476-9cd1-41b7-ba81-43c1906bce73", `{"status":"RUNNING"}`, "application/json", http.StatusBadRequest, `{"error_reason":"Compose UUID does not exist"}`},
+	}
+
+	ln, err := net.Listen("tcp", socket)
```
ondrejbudai: I would prefer these tests not to rely on TCP - I think their main purpose is to test the routes, not the transport layer. We can reuse the code from weldr API tests.

It totally makes sense to test also if the TCP listening works as expected but I think it's more suitable to test it in integration tests.

internal/rcm/api.go
msehnout: Sound reasonable to me.

internal/rcm/api.go
msehnout: I don't have any strong opinion about this. I decided not to include it yet because I'm not sure it is necessary, but I'm fine with including it.

internal/rcm/api.go
msehnout: Should I return a JSON with an error message?

internal/rcm/api_test.go
msehnout: This seemed like the easiest way to implement this, but I can check the helpers you wrote.

internal/rcm/api.go
msehnout: I don't have any strong opinion about this. I decided not to include it yet because I'm not sure it is necessary, but I'm fine with including it.

internal/rcm/api_test.go
ondrejbudai: Even without the helpers it might be easier. :)

internal/rcm/api.go
ondrejbudai: I think it will be easier for RCM guys to integrate with our API if we have sane error messages. We might consider using something like this: https://github.com/go-playground/validator

internal/rcm/api.go
ondrejbudai: I don't have it neither. Better introduce it sooner than later?

internal/rcm/api.go
teg: Agreed.

internal/rcm/api.go
teg: I think I agree. The alternative would be to introduce a new socket/listening port if ever we wanted a second version of the API. I think I prefer the option of just changing the version in the endpoint, but I think @larskarlitski also had some opinions on this when I introduced the jobqueue API (we should probably do the same in both, but happy to change the jobqueue if we agree that that's not right).

internal/rcm/api.go
teg: Some structured JSON errors makes sense to me. I don't know the project @ondrejbudai mentions, but on first glance something like that looks very useful (we could land the first version without though, if it is not clear which solution to pick).

cmd/osbuild-composer/main.go
```diff
@@ -67,5 +68,13 @@ func main() {
 	weldrAPI := weldr.New(rpm, currentArch(), distribution, logger, store)
 
 	go jobAPI.Serve(jobListener)
+
+	// Optionally run RCM API as well as Weldr API
+	if len(listeners) == 3 {
```
teg: This is a fine first version, but is there a way to get more information out of systemd about the passed sockets so we don't have to make assumptions based purely on the number of sockets passed?

internal/rcm/api.go
```diff
@@ -0,0 +1,148 @@
...
+		return
+	}
+
+	// Push the requested compose to the store
+	composeUUID := uuid.New()
+	err = api.store.PushCompose(composeUUID, &blueprint.Blueprint{}, make(map[string]string), "test_arch", "test_output", &target.Target{})
```
teg: I"d make this somehow work, even when the other features are missing:
 - refuse requests that don't have exactly one architecture
 - pass the one architecture given instead of `test_arch`
 - pass `qcow2` instead of `test_output` (we should discuss if this should be configurable as well, but let's leave it like this for now)
 - set the target to a LocalTarget, the same way we do in weldr, this will be replaced by koji, but at least then we can test

internal/rcm/api.go
```diff
@@ -0,0 +1,148 @@
...
+		json.NewEncoder(writer).Encode(errorReason)
+		return
+	}
+
+	// JSON structure with success response
+	var reply struct {
+		Status string `json:"status"`
+	}
+
+	reply.Status = compose.QueueStatus
```
teg: I agree with this, and this is what we discussed. But also thinking: we may want to make the `QueueStatus` a struct with per-job information. Just like you now got per-arch information in koji when building an RPM for several architectures (not for now, just something to think about for the future).

teg:
Very nice work! Thanks! I left some minor requests.

cmd/osbuild-composer/main.go
msehnout: I'll investigate.

internal/rcm/api.go
msehnout: I cannot hardcode `qcow2`, because the package `test_distro` only works with `test_output`, otherwise I'll fix it.

internal/rcm/api.go
msehnout: Yeah, definitely. I decided to go this way, because we need the changes in store, but I could leave a note in the code.

internal/rcm/api.go
msehnout: `api.store.PushCompose` creates the `LocalTarget` implicitly so there is no need to pass it

msehnout: Also note the "checksums" dictionary. It is supposed to be a checksum of repositories, but we want to pass the repositories as an argument, so what we really need is a new/modified argument completely.

cmd/osbuild-composer/main.go
msehnout: I think it is only possible to assign a name to all sockets in one socket unit:
https://www.freedesktop.org/software/systemd/man/systemd.socket.html#FileDescriptorName=

msehnout: You can also take a look here:
https://github.com/coreos/go-systemd/blob/master/activation/listeners.go#L42

msehnout: I might have misunderstood the use-case here. Do we want to create a 2nd socket file specifically for RCM? If so, we can distinguish the sockets by the unit in which they were defined. (I think)

teg: I think having one unit file per socket would be nice. Then people can disable/enable/mask to their heart's content, and get exactly the API's they need. And also, as you say, we get to distinguish between them.

internal/rcm/api.go
teg: Yeah, you are right this all needs a bit of a rework. Feel free to leave this as is for the first version.

cmd/osbuild-composer/main.go
teg: But we can do that in a follow-up, let's get this in as-is first.

bcl: systemd sets some environmental variables when it starts the service, I found mention of $LISTEN_FDNAMES here - https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html which may be useful. It doesn't seem to be very well documented though.

PR #198, opened 2020-01-15
Add tcp and tls support to worker API

ondrejbudai:
Several decision I made which we might discuss:
- configuration is passed in composer and worker as command line arguments. This is the easiest options but there might be some security concerns.
- socket activation is used even with tcp (systemd supports it). I'm not sure if that's what we want.

TODO:
- tests - we need some kind of integration tests for worker/composer.
- disabling local target in the case of a remote worker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,29 @@ install:
 	cp distribution/*.socket /etc/systemd/system/
 	systemctl daemon-reload
 
+.PHONY: ca
+ca:
+ifneq (/etc/osbuild-composer/ca-key.pem/etc/osbuild-composer/ca-crt.pem,$(wildcard /etc/osbuild-composer/ca-key.pem)$(wildcard /etc/osbuild-composer/ca-crt.pem))
+	@echo CA key or certificate file is missing, generating a new pair...
+	- mkdir -p /etc/osbuild-composer
+	openssl req -new -nodes -x509 -days 365 -keyout /etc/osbuild-composer/ca-key.pem -out /etc/osbuild-composer/ca-crt.pem -subj "/CN=osbuild.org"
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -20,11 +20,36 @@ install:
 	- mkdir -p /etc/sysusers.d/
 	cp distribution/osbuild-composer.conf /etc/sysusers.d/
 	systemd-sysusers osbuild-composer.conf
+	- mkdir -p /var/lib/osbuild-composer
+	- chown _osbuild-composer:_osbuild-composer /var/lib/osbuild-composer
```
teg: Should this also be added to the specfile?

teg:
Sorry for not getting to this sooner. Overall, I think this is great. I think we should not block on the CI, but should merge this after testing manually. We need proper CI for this and all other current functionality.

For the local target, I wonder if we should just bite the bullet, and replace the `cp` call with a new endpoints on the jobqueue API where logs and images can be uploaded. This way the local target and the remote workers are orthogonal, and we avoid ever using the filesystem to communicate between worker/composer.

ondrejbudai:
No worries, the current status is that I'm working on passing logs between worker/composer through the API. I'm a bit stuck with the testing part but it should be ready soon. I have some thoughts about how the image passing should work but I need some time to think it through.

Makefile
ondrejbudai: This was a mistake - `/var/lib/osbuild-composer` is created by systemd when starting `osbuild-composer.service`.

ondrejbudai:
I added more detailed commit messages and successfully retested the whole PR on RHEL 8.2.

PR #202, opened 2020-01-22
Documentation and new types for the internal package

msehnout:
TODO
- [x] Implement MarshalJSON, UnmarshalJSON for ImageType
- [x] Implement MarshalJSON, UnmarshalJSON for Distribution
- [x] Implement MarshalJSON, UnmarshalJSON for UploadTarget

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,3 +22,36 @@ or using a browser: `http://localhost:9090`
 
 Please refer to the [lorax-composer](https://github.com/weldr/lorax)'s documenation as osbuild-composer is a drop-in replacement.
 
+## High-level overview
+
+![overview](docs/osbuild-composer.svg)
+
+### Frontends
+
+`osbuild-composer` is meant to be used with 2 different front-ends. The primary one, which is meant for general use, is cockpit-composer. It is part of the Cockpit project and unless you have a strong reason not to use it, you should use it. `composer-cli` is a command line tool that can be used with `osbuild-composer`.
+
+### Compose
+* Compose is what the user submits over one of the frontends
+* It contains of one or more **image builds**
+* It contains zero or more **upload actions**
+
+### Image build
+* The resulting *image* has a *type*: https://github.com/osbuild/osbuild-composer/blob/master/internal/distro/fedora30/distro.go#L19
+* Running build in osbuild-composer is referred to as a "job" (internal terminology, not related to end-user experience)
+
+### Job
+* What composer submits to a worker
+* Is a unit of work performed by `osbuild` (internally it is a single execution of `osbuild`)
+* Consists of **one** image build and **zero or more** Upload actions
+
+### Image type
+* In the cockpit-composer, for examples these are image types:
+  * Openstack
+  * Azure
+  * AWS
+* As of now, we name them internally by their file format: vhd, ami, etc.
+* You can see a list of types by executing: `composer-cli compose types`
+
+### Upload action
+* Each image can be, but does not have to be, uploaded to a remote location
+* One image can be uploaded to multiple locations
--- a/internal/common/types.go
+++ b/internal/common/types.go
@@ -1,0 +1,212 @@
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/google/uuid"
+)
+
+// CustomJsonConversionError is thrown when parsing strings into enumerations
+type CustomJsonConversionError struct {
+	reason string
+}
+
+func (err *CustomJsonConversionError) Error() string {
+	return err.reason
+}
+
+func unmarshalHelper(data []byte, jsonError, typeConversionError string, mapping map[string]int) (int, error) {
+	var stringInput string
+	err := json.Unmarshal(data, &stringInput)
+	if err != nil {
+		return 0, &CustomJsonConversionError{string(data) + jsonError}
+	}
+	value, exists := mapping[stringInput]
+	if !exists {
+		return 0, &CustomJsonConversionError{stringInput + typeConversionError}
+	}
+	return value, nil
+}
+
+func marshalHelper(input int, mapping map[string]int, errorMessage string) ([]byte, error) {
+	for k, v := range mapping {
+		if v == input {
+			return json.Marshal(k)
+		}
+	}
+	return nil, &CustomJsonConversionError{fmt.Sprintf("%d %s", input, errorMessage)}
+}
+
+// Architecture represents one of the supported CPU architectures available for images
+// produced by osbuild-composer. It is represented as an integer because if it
+// was a string it would unmarshal from JSON just fine even in case that the architecture
+// was unknown.
+type Architecture int
+
+// A list of supported architectures. As the comment above suggests the type system does
+// not allow to create a type with a custom set of values, so it is possible to use e.g.
+// 56 instead of an architecture, but as opposed to a string it should be obvious that
+// hardcoding a number instead of an architecture is just wrong.
+//
+// NOTE: If you want to add more constants here, don't forget to add a mapping below
+const (
+	X86_64 Architecture = iota
+	Aarch64
+	Armv7hl
+	I686
+	Ppc64le
+	S390x
+)
+
+// getArchMapping is a helper function that defines the conversion from JSON string value
+// to Architecture.
+func getArchMapping() map[string]int {
+	mapping := map[string]int{
+		"x86_64":  int(X86_64),
+		"aarch64": int(Aarch64),
+		"armv7hl": int(Armv7hl),
+		"i686":    int(I686),
+		"ppc64le": int(Ppc64le),
+		"s390x":   int(S390x),
+	}
+	return mapping
+}
+
+// UnmarshalJSON is a custom unmarshaling function to limit the set of allowed values
+// in case the input is JSON.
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -22,3 +22,35 @@ or using a browser: `http://localhost:9090`
 
 Please refer to the [lorax-composer](https://github.com/weldr/lorax)'s documenation as osbuild-composer is a drop-in replacement.
 
+## High-level overview
+
+![overview](docs/osbuild-composer.svg)
+
+### Frontends
+
+`osbuild-composer` is meant to be used with 3 different front-ends. The primary one, which is meant for general use, is cockpit-composer. It is part of the Cockpit project and unless you have a strong reason not to use it, you should use it. `composer-cli` is a command line tool that can be used for non-interactive usage. Finally, RCM client (TBD) is a special-purpose client that is not meant for general usage.
```
bcl: `composer-cli` is *not* for non-interactive usage. It is for humans and makes no guarantees about its output. All non-interactive usage outside of tests that are tightly coupled to the code repository should be using the API directly.

README.md
```diff
@@ -22,3 +22,35 @@ or using a browser: `http://localhost:9090`
...
+
+### Compose
+* Compose is what the user submits over one of the frontends
+* It contains of one or more **image builds**
+* It contains zero or more **upload actions**
+
+### Image build
+* The resulting *image* has a *type*: https://github.com/osbuild/osbuild-composer/blob/master/internal/distro/fedora30/distro.go#L19
```
bcl: That link doesn't tell me anything about supported types. Maybe it should reference `composer-cli compose types` instead?

internal/common/types.go
```diff
@@ -0,0 +1,105 @@
...
+		*arch = S390x
+	default:
+		return &ParseCustomTypeError{archString + " is not a valid architecture"}
+	}
+	return nil
+}
+
+type ImageType int
+
+const (
```
bcl: I think this and Distribution should also have unmarshal functions, depending on int values for the inputs is going to be fragile and difficult to keep synchrnized with the front ends when they are extended.

internal/crypt/crypt_impl.go
```diff
@@ -1,3 +1,5 @@
+// +build !macos
```
bcl: What is `+build !macos`?

internal/crypt/crypt_impl_macos.go
```diff
@@ -0,0 +1,7 @@
+// +build macos
+
+package crypt
+
+func crypt(pass, salt string) (string, error) {
+	return "", nil
```
bcl: Why is this here? This seems like the kind of thing that can come back to haunt you in the future...

README.md
msehnout: thanks, I didn't know that

README.md
msehnout: I think about this README file as a file for developers, but it is worth mentioning.

internal/common/types.go
msehnout: Sure, I'll implement Unmarshal as well. Would you suggest to use something else than `int`?

internal/crypt/crypt_impl.go
msehnout: Conditional compilation for all non-macos systems.

internal/crypt/crypt_impl_macos.go
msehnout: yeah, @ondrejbudai advised me to use panic instead, which is very, very resonable :)

msehnout: And it's here to make compilation on macos possible, because glibc does not support macos so the implementation in `crypt_impl` cannot be compiled...

README.md
```diff
@@ -22,3 +22,36 @@ or using a browser: `http://localhost:9090`
 
 Please refer to the [lorax-composer](https://github.com/weldr/lorax)'s documenation as osbuild-composer is a drop-in replacement.
 
+## High-level overview
+
+![overview](docs/osbuild-composer.svg)
+
+### Frontends
+
+`osbuild-composer` is meant to be used with 3 different front-ends. The primary one, which is meant for general use, is cockpit-composer. It is part of the Cockpit project and unless you have a strong reason not to use it, you should use it. `composer-cli` is a command line tool that can be used with `osbuild-composer`.
```
teg: Maybe change to "`osbuild-composer` has two frontends:" (rather than 3, no need to talk about RCM yet).

internal/common/types.go
```diff
@@ -0,0 +1,126 @@
...
+type ParseCustomTypeError struct {
+	reason string
+}
+
+func (err *ParseCustomTypeError) Error() string {
+	return err.reason
+}
+
+// Architecture represents one of the supported CPU architectures available for images
+// produced by osbuild-composer. I decided to represent it as an integer because if it
```
teg: Use the passive voice in comments: "It is represented as an integer because...". The same for the rest of the comments. This doesn't make a difference to the meaning of course, just keeps the style consistent.

internal/common/types.go
```diff
@@ -0,0 +1,126 @@
...
+// was a string it would unmarshal from JSON just fine even in case that the architecture
+// was unknown (yeah, thanks Go for doing the same mistakes C did almost 50 years ago)
```
teg: I don't disagree, but drop the comments in the parenthesis to make the tone of the documentation more neutral.

internal/common/types.go
```diff
@@ -0,0 +1,126 @@
...
+//
+// I considered some alternatives like a struct with private fields, but struct cannot
+// be const, so that does not help either. Eventually I think this is the "get s**t done"
+// solution and next time please use a language with decent type system.
```
teg: This comment would be better suited to the commit message ("why I did what I did", as opposed to "how it works"). And comments about Go would be better suited to Twitter ;)

internal/store/store.go
```diff
@@ -489,6 +489,11 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 	return nil
 }
 
+func (s *Store) PushRcmCompose(composeRequest ComposeRequest) error {
```
teg: This looks like it belongs in a different commit.

README.md
bcl: Even as a developer that line didn't help. I see that variable is populated later with the types but it's not immediately clear when looking at that line.

internal/common/types.go
bcl: I think using int constants internally is fine.

internal/common/types.go
msehnout: Sure. It's just that all the grammar-checking tools nowadays complain about overuse of passive voice, that I'm trying to avoid it from time to time.

internal/common/types.go
msehnout: Hehe, I don't use Twitter or any other social network, but I'll drop it if you insist!

internal/store/store.go
msehnout: You are right.

README.md
msehnout: Sure, but I still think it is a pointer in the right direction. There is a whole section about "image type" below this one.

internal/common/types.go
```diff
@@ -0,0 +1,105 @@
...
+
+const (
+	Fedora30 Distribution = iota
+	RHEL82
+)
+
+type UploadTarget int
+
+const (
+	S3           UploadTarget = iota
```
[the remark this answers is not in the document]
msehnout: done

README.md
```diff
@@ -22,3 +22,35 @@ or using a browser: `http://localhost:9090`
...
+![overview](docs/osbuild-composer.svg)
+
+### Frontends
+
+`osbuild-composer` is meant to be used with 3 different front-ends. The primary one, which is meant for general use, is cockpit-composer. It is part of the Cockpit project and unless you have a strong reason not to use it, you should use it. `composer-cli` is a command line tool that can be used for non-interactive usage. Finally, RCM client (TBD) is a special-purpose client that is not meant for general usage.
+
+### Compose
+* Compose is what the user submits over one of the frontends
+* It contains of one or more **image builds**
+* It contains zero or more **upload actions**
```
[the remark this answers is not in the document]
msehnout: done

README.md
```diff
@@ -22,3 +22,35 @@ or using a browser: `http://localhost:9090`
...
+
+### Image build
+* The resulting *image* has a *type*: https://github.com/osbuild/osbuild-composer/blob/master/internal/distro/fedora30/distro.go#L19
+* Running build in osbuild-composer is referred to as a "job" (internal terminology, not related to end-user experience)
+
+### Job
+* What composer submits to a worker
+* Is a unit of work performed by `osbuild` (internally it is a single execution of `osbuild`)
+* Consists of **one** image build and **one or more** Upload actions
```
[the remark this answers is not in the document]
msehnout: fixed

internal/common/types.go
```diff
@@ -0,0 +1,212 @@
+package store
```
ondrejbudai: Is this right? I'm not sure if this belongs to the `store` package.

msehnout: Thanks for catching this! I guess `package common` was in the commit I removed...

PR #203, opened 2020-01-22
Add support for the Fedora 32 distribution

bcl:


bcl:
> When adding support for new distros/output types, please add tests as well. However, we may hold off on this as we plan to rework some of the pipeline generation.

This includes changes to internal/distro/fedora32/distro_test.go can you point me to an example of other distro related tests?

jkozol:
@bcl This PR fell behind master and given the recent changes I found it easiest to use master as a starting point. I believe this PR can be closed as long as #241 looks good to you.

PR #206, opened 2020-01-24
Add coveralls.io coverage reports to the unit tests

bcl:
This uses the goveralls project to run the tests and push the results to
coveralls.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,19 +13,22 @@ matrix:
       language: go
       go: 1.12.x
       env: GO111MODULE=on # needed even for Go 1.12
-      script: go test -v ./...
+      before_install: go get github.com/mattn/goveralls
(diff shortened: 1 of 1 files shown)


bcl:
I think an admin needs to add this repo to coveralls.io for this to work completely.

bcl:
Turned on more verbose output, looks like it is running all the tests so I'm not sure why that coveralls report only showed coverage for crypt.

msehnout:
We need to discuss enabling of coveralls for this organization first. I enabled it for a while, but other didn't agree, so I denied the access again. But I'm definitely interested in having code coverage tool.

gicmo:
I switched one of the projects (G-Node/nix) from https://coveralls.io to https://codecov.io because I personally preferred that (better config options, etcpp).

bcl:
See PR #225 for a simpler coverage report.

PR #209, opened 2020-01-27
osbuild-pipeline: Search also the current dir for repository configs

ondrejbudai:
…nfigs


ondrejbudai:
Makes sense. I've fixed it and added a note about this decision into the commit messsage.

PR #212, opened 2020-01-28
distro: use fedora-30 as hostname for all fedora versions

jkozol:
Osbuild-composer will not run on fedora versions other than fedora 30. For instance, osbuild-composer will get fedora 31's distro name as fedora-31 and will crash because it is an unkown distro. But, fedora 31 is able to build fedora 30 images. So, I changed the FromHost function so all versions of fedora will use the distro name of fedora-30 and osbuild-composer will be able to run on them and build fedora 30 images.


jkozol:
Yes that's fine

Issue #143, opened 2019-12-05
Composes should contain frozen blueprint


ondrejbudai:
Currently, we just copy the blueprint to the new compose. This means they can contain wildcard package versions.

However, Lorax does a different thing: Takes all the extra packages and groups, depsolves them and replaces the wildcard versions with specific ones. This process is known in lorax's codebase as blueprint freezing.

This has visible outcomes in some APIs like `/compose/info` or `/compose/metadata`. Generally it's a good feature for reproducibility of composes.

teg:
This was closed by #214

PR #214, opened 2020-01-30
Fix /blueprints/freeze results

bcl:
The dependencies are not sorted, so depending on what order they were
returned in the freeze route would or would not return the correct
results (exhibited by the version being the original glob instead of the
EVRA).

This also fixes the tests so that the depsolve results are slightly
unsorted by adding a dep-package3 to the start of the list.


internal/blueprint/blueprint_test.go
```diff
@@ -0,0 +1,32 @@
...
+		Name:        "deepcopy-test",
+		Description: "Testing DeepCopy function",
+		Version:     "0.0.1",
+		Packages: []Package{
+			{Name: "dep-package1", Version: "*"}},
+		Modules: []Package{
+			{Name: "dep-package2", Version: "*"}},
+	}
+
+	bpCopy, err := bpOrig.DeepCopy()
```
ondrejbudai: I think we should have an equality test of the original and the copy here.

internal/weldr/api.go
```diff
@@ -848,6 +848,29 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
 	})
 }
 
+// setPkgEVRA replaces the version globs in the blueprint with their EVRA values from the dependencies
+//
+// The dependencies must be pre-sorted for this function to work properly
+// It will return an error if it cannot find a package in the dependencies
+func setPkgEVRA(dependencies []rpmmd.PackageSpec, packages []blueprint.Package) error {
```
ondrejbudai: Shouldn't we use this function also in `composeHandler`? I think lorax does freeze the blueprint before making a compose. See https://github.com/osbuild/osbuild-composer/issues/143

internal/blueprint/blueprint_test.go
bcl: Good idea, I added cmp.Diff so we can see what went wrong.

internal/weldr/api.go
bcl: I think that's future work, the way (AFAIK) osbuild works doesn't support passing the version into the pipeline, it seems to depend on the state of the metadata at the time of the depsolve.

teg: You can pass the versions into the pipeline, the packages provided are passed as-is to dnf. Though might make more sense to wait for the dnf/rpm rework to land.

bcl: Right. I think that's a separate issue from this PR.

internal/blueprint/blueprint.go
```diff
@@ -31,6 +35,22 @@ type Group struct {
 	Name string `json:"name" toml:"name"`
 }
 
+// DeepCopy returns a deep copy of the blueprint
+// This uses json.Marshal and Unmarshal which are not very efficient
+func (b *Blueprint) DeepCopy() (Blueprint, error) {
```
teg: This does the job, but it feels a bit odd to me. Maybe consider (for a follow-up) to either do a proper deep-copy implementation for all the fields (and see if anything can be turned from pointers to embedded members to make it simpler), or simply construct a new Blueprint based on the old and the additional information.

bcl: Yeah, it feels a bit dirty to me as well, but the alternative is a huge pile of code for all the fields, and given the number of pointers involved I guarantee that's going to have bugs.
I was hoping someone would suggest a well-known deepcopy library -- but this will work for now :)

PR #215, opened 2020-01-30
Store virtual image size

jkozol:
Currently, osbuild-composer returns the file size of an image in its compose status/info response. We should instead return the image size that is defined either by the user when creating the compose or is the default value if the user does not define a size. The user defined size is passed to the store when the compose is created but the default image size is only stored in the pipeline's assembler options. So, in order to access the image size even if a user does not define one, I added a function to the assembler options' interface which will return the size specified for the assembler.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -484,6 +484,8 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 		repos = append(repos, source.RepoConfig())
 	}
 
+	size = s.distro.GetSizeForOutputType(composeType, size)
+
 	pipeline, err := s.distro.Pipeline(bp, repos, checksums, arch, composeType, size)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


internal/store/store.go
```diff
@@ -468,6 +468,11 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 	if err != nil {
 		return err
 	}
+
+	if pipeline.Assembler != nil {
+		size = pipeline.Assembler.Options.GetImageSize()
```
teg: Rather than reading this out from the Pipeline, how about introducing a `distro.GetSizeForOutputType(size uint64)` function? This could then also deal with rounding (as is done for vhd in the weldr API), and if you pass it 0 it uses the default value. Then you could store that in the compose, and also pass it back to the pipeline translation.

In general, I'd rather avoid adding any logic to the pipeline package apart from its current role as a thin wrapper around the json objects.

PR #220, opened 2020-02-02
Prepare to switch from dnf to rpm in our pipelines

teg:
These preliminary commits adds the needed low-level features to be able to use the rpm rather than the dnf stage. The switchover to change the produced pipelines is not yet done, so this can be merged independently of the required changes in osbuild.

### DIFF
--- a/internal/distro/fedora30/distro.go
+++ b/internal/distro/fedora30/distro.go
@@ -43,6 +43,7 @@ type output struct {
 }
 
 const Distro = common.Fedora30
+const ModulePlatformID = "platform:f30"
 
 func New(confPaths []string) *Fedora30 {
 	const GigaByte = 1024 * 1024 * 1024
@@ -311,6 +312,10 @@ func (r *Fedora30) Distribution() common.Distribution {
 	return Distro
 }
 
+func (r *Fedora30) ModulePlatformID() string {
+	return ModulePlatformID
+}
+
 func (r *Fedora30) Repositories(arch string) []rpmmd.RepoConfig {
 	return r.arches[arch].Repositories
 }
@@ -448,7 +453,7 @@ func (r *Fedora30) dnfStageOptions(arch arch, additionalRepos []rpmmd.RepoConfig
 	options := &osbuild.DNFStageOptions{
 		ReleaseVersion:   "30",
 		BaseArchitecture: arch.Name,
-		ModulePlatformId: "platform:f30",
+		ModulePlatformId: ModulePlatformID,
 	}
 
 	for _, repo := range append(arch.Repositories, additionalRepos...) {
--- a/internal/distro/fedora30/distro.go
+++ b/internal/distro/fedora30/distro.go
@@ -2,16 +2,17 @@ package fedora30
 
 import (
 	"errors"
-	"github.com/osbuild/osbuild-composer/internal/common"
 	"log"
 	"sort"
 	"strconv"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/osbuild"
 
 	"github.com/google/uuid"
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/crypt"
-	"github.com/osbuild/osbuild-composer/internal/pipeline"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 )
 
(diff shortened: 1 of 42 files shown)


internal/weldr/api.go
```diff
@@ -1757,7 +1757,14 @@ func (api *API) fetchPackageList() (rpmmd.PackageList, error) {
 	return packages, err
 }
 
-func (api *API) depsolveBlueprint(bp *blueprint.Blueprint, clean bool) ([]rpmmd.PackageSpec, map[string]string, error) {
+func (api *API) depsolveBlueprint(bp *blueprint.Blueprint, outputType, arch string, clean bool) ([]rpmmd.PackageSpec, []rpmmd.PackageSpec, map[string]string, error) {
```
msehnout: I'd prefer if new code used the types from `internal/common` and variable names reflected what is written in the README file. So I'd suggest to replace `outputType` with `imageType`.

internal/weldr/api.go
```diff
@@ -1769,16 +1776,33 @@ func (api *API) depsolveBlueprint(bp *blueprint.Blueprint, clean bool) ([]rpmmd.
 			specs[i] += "-*-*.*"
 		}
 	}
+	excludeSpecs := []string{}
+	if outputType != "" {
```
msehnout: How is it possible to reach this function without an image type? I thought depsolv without the set of base packages does not make any sense.

internal/weldr/api.go
```diff
@@ -1769,16 +1776,33 @@ func (api *API) depsolveBlueprint(bp *blueprint.Blueprint, clean bool) ([]rpmmd.
...
+		return nil, nil, nil, err
 	}
-	for _, source := range api.store.GetAllSources() {
-		repos = append(repos, source.RepoConfig())
+
+	buildPackages := []rpmmd.PackageSpec{}
+	if outputType != "" {
+		buildSpecs, err := api.distro.BuildPackages(arch)
+		if err != nil {
+			return nil, nil, nil, err
```
msehnout: Do we have a reasonable way to tell the user what happened here? Was it possible to have such errors in Weldr?

internal/osbuild/source.go
```diff
@@ -0,0 +1,43 @@
...
+		return err
+	}
+	*sources = make(map[string]Source)
+	for name, rawSource := range rawSources {
+		var source Source
+		switch name {
+		case "org.osbuild.files":
+			source = new(FilesSource)
+		default:
+			return errors.New("unexpected suorce name" + name)
```
msehnout: Just a small typo :)

internal/distro/fedora30/distro.go
```diff
@@ -436,16 +455,11 @@ func (r *Fedora30) Runner() string {
...
-		"e2fsprogs",
-		"policycoreutils",
-		"qemu-img",
-		"systemd",
-		"tar",
-	}
-	packages = append(packages, arch.BuildPackages...)
+	packages, err := r.BuildPackages(arch.Name)
+	if err != nil {
+		panic("impossibly invalid arch")
```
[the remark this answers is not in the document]
jkozol: I agree that this is not a widely used phrase and "invalid arch" or "impossible and invalid arch" might be technically better but I think it explains the error fine and I personally enjoy the phrase "impossibly invalid arch".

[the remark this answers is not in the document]
teg: I spent way too long trying to decide what to do with this feedback ;) In the end: yes, the phrase has a nice sound to it, but I'll make it grammatically correct.

internal/weldr/api.go
teg: I decided to leave this in for now, and rather clean up all the naming in one commit/PR.

internal/weldr/api.go
teg: This would be an programming error on our side. We should `panic()` or at most return `InternalServerError`.

teg:
Rebased on master, fixed according to the reviews and dropped the final commit that changes the output of osbuild-pipeline. This should be done after osbuild itself has been adjusted.

internal/distro/test/distro.go
```diff
@@ -0,0 +1,60 @@
+package test
```
msehnout: Is this package used anywhere? I replaced all usages with the fedoratest packages if I remember correctly.

internal/osbuild/source.go
msehnout: Still there ;-)

msehnout:
As other were fine with this PR already, I'm going to merge this now. Thanks @teg !

PR #221, opened 2020-02-03
Rcm api continued

msehnout:
see individual commits for detailed description


internal/store/store.go
```diff
@@ -509,6 +510,59 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
+		// TODO: handle custom upload targets
+		//targets := append(imgRequest.UpTarget, appendTargets)
+		// TODO: implment support for no checksums
+		// TODO: implement compatibility translation from image type or refactor distro packages
+		pipeline, err := s.distro.Pipeline(&request.Blueprint, request.Repositories, nil, arch, "qcow2", 0)
+		if err != nil {
+			return err
+		}
+		// TODO: refactor compose so that it can contain multiple image builds with different types and upload targets
+		// TODO: per image build
```
teg: Overall, I like this very much. It is ok to leave missing functionality out as TODO, but in this case I think a couple of these TODOs must be implemented in earlier commits, because as it stands this would lead to invalid behaviour. Firstly, I think you need to resolve the checksum (as mentioned in previous comments) and pass them in. Having a `nil` checksum would lead to an invalid `pipeline`. Secondly, I think you need to either refuse composes with more than one request, or implement multiple job support in a previous commit. You are right that these things are now indexed by `ComposeID` and that's not enough by itself. I think we should introduce a separate JobID, which is just a random UUID in the same way that `ComposeID` is (so from the workers this makes no difference), and then have a mapping back from `JobID`s to `ComposeID`s internally.
In order to get the rest of this merged, I think it would be ok to leave that as a TODO and instead just error out in case we are passed more than one request at once.

internal/rcm/api.go
```diff
@@ -81,7 +81,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 
 	type Repository struct {
 		URL      string `json:"url"`
-		Checksum string `json:"checksum"`
+		Checksum string `json:"checksum,omitempty"`
```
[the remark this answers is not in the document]
msehnout: My thinking was: Make it possible to use RCM API at least with it (because the codebase currently require it), but sure it makes sense to remove it completely.

internal/common/types.go
```diff
@@ -238,9 +266,10 @@ type ImageRequest struct {
...
-	ComposeID       uuid.UUID      `json:"uuid"`
-	Distro          Distribution   `json:"distro"`
-	Arch            Architecture   `json:"arch"`
-	RequestedImages []ImageRequest `json:"requested_images"`
+	Blueprint       blueprint.Blueprint `json:"blueprint"`
+	ComposeID       uuid.UUID           `json:"uuid"`
+	Distro          Distribution        `json:"distro"`
+	Arch            Architecture        `json:"arch"`
+	Repositories    []rpmmd.RepoConfig  `json:"repositories"`
+	RequestedImages []ImageRequest      `json:"requested_images"`
```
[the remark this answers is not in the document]
msehnout: I'm basically still working on the structure. Previously I thought it would be enough to have the name, but since we want to move out the blueprint storage it has to take the blueprint as a structure. But sure, I'll try to explain it in a commit message.

msehnout:
Nice, thanks for the review :)

internal/rcm/api.go
```diff
@@ -134,7 +134,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 		if api.logger != nil {
 			api.logger.Println("RCM API failed to push compose:", err)
 		}
-		writer.WriteHeader(http.StatusInternalServerError)
+		writer.WriteHeader(http.StatusBadRequest)
```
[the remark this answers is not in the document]
ondrejbudai: The error handling here sadly isn't ideal - in fact if `err == nil` it can be both because of bad request or internal server error.

internal/store/store.go
```diff
@@ -509,6 +510,59 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
+func (s *Store) PushComposeRequest(request common.ComposeRequest) error {
+	// Create a local target that will be appended to the requested targets
+	appendTargets := []*target.Target{}
+	if s.stateDir != nil {
+		appendTargets = append(appendTargets, target.NewLocalTarget(
+			&target.LocalTargetOptions{
+				Location: *s.stateDir + "/outputs/" + request.ComposeID.String(),
+			},
+		))
+	}
```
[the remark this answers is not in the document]
ondrejbudai: :+1: I very much agree.

internal/common/types.go
```diff
@@ -54,6 +56,20 @@ func existsHelper(mapping map[string]int, testedValue string) bool {
 	return false
 }
 
+func toStringHelper(mapping map[string]int, tag int) (string, bool) {
```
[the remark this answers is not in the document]
msehnout: I decided to create a comment in the code.

internal/store/store.go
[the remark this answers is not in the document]
msehnout: ok, removed

internal/store/store.go
msehnout: I introduced a panic in the beginning of this function to prevent it from being used. I think we should not implement support for no checksum as part of this PR because it is huge already and the cost of rebasing it is far greater then the benefits of not merging it now. The RCM API is not yet ready anyway and this decision is not going to affect this fact. If we merge this without support for no checksums it will enable us to focus on implementing it instead of keeping the changes synchronized with master branch.

internal/store/store.go
```diff
@@ -484,35 +489,110 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
+			panic("fatal error, image type should exist but it does not")
+		}
+		pipelineStruct, err := distroStruct.Pipeline(&request.Blueprint, request.Repositories, nil, arch, imgTypeCompatStr, 0)
+		if err != nil {
+			return err
+		}
+
+		s.pendingJobs <- Job{
+			ComposeID: request.ComposeID,
+			Distro:    s.distro.Name(),
```
ondrejbudai: Why not use the `request.Distro` value here? I think we should get rid of `s.distro` altogether because we will want to make cross-distro builds in the future.

internal/store/store.go
```diff
@@ -484,35 +489,110 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
-			Targets:     targets,
-			JobCreated:  time.Now(),
-			Size:        size,
+	// FIXME: handle or comment this possible error
+	_ = s.change(func() error {
+		s.Composes[composeID] = compose.Compose{
+			Blueprint: bp,
+			ImageBuilds: []compose.ImageBuild{
+				{
+					QueueStatus: common.IBWaiting,
```
ondrejbudai: Why is there no `Distro` field? I think it might confuse us in the future.

msehnout: Because this is the compatibility version, which is supposed to replace previous pushcompose without any changes.

ondrejbudai: Makes sense, we can clean-up in the follow-up.

internal/store/store.go
msehnout: Thanks, this is a clear mistake.

msehnout: Fixed

internal/store/store.go
```diff
@@ -521,46 +601,83 @@ func (s *Store) DeleteCompose(id uuid.UUID) error {
...
 
-		if compose.QueueStatus != "FINISHED" && compose.QueueStatus != "FAILED" {
+		// If any of the image builds have build artifacts, remove them
+		invalidRequest := true
+		for _, imageBuild := range compose.ImageBuilds {
+			if imageBuild.QueueStatus == common.IBFinished || imageBuild.QueueStatus == common.IBFailed {
+				invalidRequest = false
+			}
+		}
+		if invalidRequest {
```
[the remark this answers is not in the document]
msehnout: The compose could still be running while some of the image builds would already be finished.

internal/compose/compose.go
```diff
@@ -0,0 +1,159 @@
...
+		if !fn(i) {
+			acc = false
+		}
+	}
+	return acc
+}
+
+// GetState returns a state of the whole compose which is derived from the states of
+// individual image builds inside the compose
+func (c *Compose) GetState() common.ComposeState {
```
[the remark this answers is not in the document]
msehnout: done

internal/store/store.go
```diff
@@ -484,35 +489,110 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
...
+		imgTypeCompatStr, exists := imageRequest.ImgType.ToCompatString()
+		if !exists {
+			panic("fatal error, image type should exist but it does not")
+		}
+		pipelineStruct, err := distroStruct.Pipeline(&request.Blueprint, request.Repositories, nil, arch, imgTypeCompatStr, 0)
+		if err != nil {
+			return err
+		}
+		
+		s.pendingJobs <- Job{
```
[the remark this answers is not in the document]
msehnout: You are right, fixed.

internal/store/store.go
```diff
@@ -521,46 +601,83 @@ func (s *Store) DeleteCompose(id uuid.UUID) error {
...
+		// Check that the compose exists
+		currentCompose, exists := s.Composes[composeID]
 		if !exists {
 			return &NotFoundError{"compose does not exist"}
 		}
-		if compose.QueueStatus == "WAITING" {
+		// Check that the image build was waiting
+		if currentCompose.ImageBuilds[imageBuildID].QueueStatus == common.IBWaiting {
 			return &NotPendingError{"compose has not been popped"}
 		}
```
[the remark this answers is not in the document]
msehnout: yeah, I missed that one, but I decided to revert the change because RCM API will not use local target and Weldr API will not use multiple image builds per compose, so it was useless anyway

msehnout:
> The current blocker for me is that this PR breaks the weldr API (tested on RHEL 8.2). For some reason when job succeeds or fails, this doesn't change the job state, so in cockpit-composer it is still shown as running.

Good catch! I didn't notice, because the resulting image was there as usual. It should work now or at least it works for my on my hybrid Fedora 31-30 machine :D.

cmd/osbuild-composer/main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 		logger = log.New(os.Stdout, "", 0)
 	}
 
-	store := store.New(&stateDir, distribution)
+	store := store.New(&stateDir, distribution, *distros)
```
teg: This should probably be part of the commit that changed the `New()` API.

internal/common/states.go
```diff
@@ -0,0 +1,79 @@
+package common
```
teg: For this and the similar commits later on: Please explain in the commit message how (if?) this differs from the original code, if there are any behavioural changes.

internal/store/store.go
```diff
@@ -484,35 +488,119 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, checks
 		repos = append(repos, source.RepoConfig())
 	}
 
-	size = s.distro.GetSizeForOutputType(composeType, size)
```
ondrejbudai: It seems you accidentally removed this line - probably while rebasing. The removal of this line breaks the rhel 8.2 for me - without it the default image size isn't set, which leads to images with size of 0.

ondrejbudai: With this fix I was able to succesfully build an image! :slightly_smiling_face:

internal/weldr/api.go
```diff
@@ -1482,14 +1490,16 @@ func (api *API) composeInfoHandler(writer http.ResponseWriter, request *http.Req
...
-	reply.ComposeType = compose.OutputType
-	reply.QueueStatus = compose.QueueStatus
-	if compose.Image != nil {
-		reply.ImageSize = compose.Size
+	// Weldr API assumes only one image build per compose, that's why only the
+	// 1st build is considered
+	reply.ComposeType, _ = compose.ImageBuilds[0].ImageType.ToCompatString()
+	reply.QueueStatus = compose.GetState().ToString()
+	if compose.ImageBuilds[0].Image != nil {
+		reply.ImageSize = compose.ImageBuilds[0].Image.Size
```
ondrejbudai: Oh, one more bad rebase bug. The line should be:
```
reply.ImageSize = compose.ImageBuilds[0].Size
```

*Note: actually the condition doesn't make much sense but please don't remove it because it's a bit of unrelated change and I already did it in one of my yet-to-be-posted PR.*

internal/weldr/compose.go
```diff
@@ -2,69 +2,70 @@ package weldr
...
+	case common.IBRunning:
+		composeEntry.JobCreated = float64(compose.ImageBuilds[0].JobCreated.UnixNano()) / 1000000000
+		composeEntry.JobStarted = float64(compose.ImageBuilds[0].JobStarted.UnixNano()) / 1000000000
 
-	case "FINISHED":
-		if compose.Image != nil {
-			composeEntry.ImageSize = compose.Size
+	case common.IBFinished:
+		if compose.ImageBuilds[0].Image != nil {
+			composeEntry.ImageSize = compose.ImageBuilds[0].Image.Size
```
ondrejbudai: The same thing as before.

internal/rcm/api.go
```diff
@@ -122,7 +121,8 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 			errors = append(errors, "input must specify repositories")
 		}
 		errorReason.Error = strings.Join(errors, ", ")
-		json.NewEncoder(writer).Encode(errorReason)
+		// TODO: handle error
```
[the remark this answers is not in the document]
msehnout: It is explicit "I don't care". What we had until now was the same, but implicit. I don't think we should leave any errors without explicit handling.

internal/rcm/api_test.go
```diff
@@ -42,18 +43,19 @@ func TestBasicRcmAPI(t *testing.T) {
 		{"GET", "/v1/compose", ``, "", http.StatusMethodNotAllowed, ``},
 		{"POST", "/v1/compose", `{"status":"RUNNING"}`, "application/json", http.StatusBadRequest, ``},
 		{"POST", "/v1/compose", `{"status":"RUNNING"}`, "text/plain", http.StatusBadRequest, ``},
-		{"POST", "/v1/compose", `{"distribution": "fedora-30", "image_types": ["test_output"], "architectures":["test_arch"], "repositories": [{"url": "aaa", "checksum": "bbb"}]}`, "application/json", http.StatusOK, `{"compose_id":".*"}`},
+		{"POST", "/v1/compose", `{"distribution": "fedora-30", "image_types": ["qcow2"], "architectures":["x86_64"], "repositories": [{"url": "aaa"}]}`, "application/json", http.StatusOK, `{"compose_id":".*"}`},
```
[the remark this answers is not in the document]
msehnout: Yeah, it would be nice. I need to learn how to do this with git, because I still do it manually.

internal/rcm/api.go
```diff
@@ -178,6 +182,7 @@ func (api *API) status(writer http.ResponseWriter, request *http.Request, params
 	}
 
 	// TODO: return per-job status like Koji does (requires changes in the store)
-	reply.Status = compose.QueueStatus
-	json.NewEncoder(writer).Encode(reply)
+	reply.Status = compose.GetState().ToString()
```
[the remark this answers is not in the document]
msehnout: Because there is no field "QueueStatus" in the compose any more. It must be derived from the Image Builds states in a method call.

internal/rcm/api_test.go
```diff
@@ -42,18 +43,19 @@ func TestBasicRcmAPI(t *testing.T) {
...
 		{"POST", "/v1/compose/111-222-333", `{"status":"RUNNING"}`, "application/json", http.StatusMethodNotAllowed, ``},
 		{"GET", "/v1/compose/7802c476-9cd1-41b7-ba81-43c1906bce73", `{"status":"RUNNING"}`, "application/json", http.StatusBadRequest, `{"error_reason":"Compose UUID does not exist"}`},
 	}
 
-	distro := test_distro.New()
-	api := rcm.New(nil, store.New(nil, distro))
+	registry := distro.NewRegistry([]string{"."})
+	distroStruct := fedoratest.New()
+	api := rcm.New(nil, store.New(nil, distroStruct, *registry))
```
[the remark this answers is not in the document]
msehnout: ok, I'll add a comment. See my comment above, it's the same case.

internal/jobqueue/api.go
```diff
@@ -89,15 +89,17 @@ func (api *API) addJobHandler(writer http.ResponseWriter, request *http.Request,
...
+	nextJob := api.store.PopJob()
 
 	writer.WriteHeader(http.StatusCreated)
-	json.NewEncoder(writer).Encode(Job{
-		ID:         nextJob.ComposeID,
-		Distro:     nextJob.Distro,
-		Pipeline:   nextJob.Pipeline,
-		Targets:    nextJob.Targets,
-		OutputType: nextJob.OutputType,
+	// FIXME: handle or comment this possible error
```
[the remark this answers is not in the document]
msehnout: I guess. This side has nothing to do with user interface so it makes sense. I'd like to work on these in a following PR.

internal/store/store.go
```diff
@@ -457,11 +448,24 @@ func (s *Store) GetComposeResult(id uuid.UUID) (io.ReadCloser, error) {
 	return os.Open(*s.stateDir + "/outputs/" + id.String() + "/result.json")
 }
 
+func (s *Store) getImageLocationForLocalTarget(composeID uuid.UUID) string {
+	// This might result in conflicts because one compose can have multiple images, but (!) the Weldr API
+	// does not support multiple images per build and the RCM API does not support Local target, so it does
+	// not really matter any more.
```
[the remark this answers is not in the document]
msehnout: It depends. We might want to rethink the way I currently create Image Build ID.

internal/common/states.go
msehnout: ok, done

internal/weldr/api.go
msehnout: thanks!

internal/weldr/compose.go
msehnout: thanks!

internal/store/store.go
msehnout: thanks!

msehnout:
All remaining errors in journal have been fixed! (except for modularity which is not part of this PR :) )

PR #222, opened 2020-02-04
Transfer logs from worker to composer using API instead of filesystem

ondrejbudai:
This is needed in order to implement remote workers.


teg:
Nice stuff! Thanks!

PR #223, opened 2020-02-04
Enforce semantic versions in new blueprints

bcl:


bcl:
Fixed, tests pass with PR #246 added.

PR #225, opened 2020-02-10
Add coveralls.io code coverage github-action

bcl:


.travis.yml
```diff
@@ -13,19 +13,25 @@ matrix:
       language: go
       go: 1.12.x
       env: GO111MODULE=on # needed even for Go 1.12
-      script: go test -v ./...
+      script:
+        go test -coverprofile=coverage.out -v ./...
+        go tool cover -func=coverage.out
```
dvdhrm: This ends up as a multiline script, not as two separate entries. I think you need to prefix the lines with a dash.

At least that is how I interpret the travis logs.

bcl: Yeah. I love yaml :)

bcl:
> Can you elaborate on the intention here? Is this meant to amend coverage information to the CI output, or will the coverage data just be printed to the CI logs?

Well, I'd prefer to push things to a service like coveralls.io but nobody liked that patch :)
I think it's useful to be able to see the coverage levels for submitted PRs, even if it isn't easy to compare it to the parent commit.

teg:
I didn't have a chance to look into your other proposal (yet). But I think having some service would be more useful. I think that just lacked some justification/discussion. What does it give us, how does it work, etc.

bcl:
You can take a look at lorax and pykickstart as good examples of how coveralls.io looks over time:

* https://coveralls.io/github/weldr/lorax
* https://coveralls.io/github/pykickstart/pykickstart

It is linked to the github repo the same way travis-ci is, and when the upload runs from inside travis it includes the credentials to link to the correct repo without having to do a bunch of extra setup.

I'm sure there are other services out there, this is just the one I'm most familiar with, and we have been using it for years with minimal issues with it.

PR #229, opened 2020-02-12
weldr: add filtering to compose route

jkozol:
When requesting the compose status, a user may want to filter the list of composes by blueprint name, compose status, and/or compose type. These filters can now be set in the /compose/status route's url as the queries blueprint, status, and type.


internal/weldr/api.go
```diff
@@ -1436,6 +1436,36 @@ func (api *API) composeStatusHandler(writer http.ResponseWriter, request *http.R
...
+			if filterBlueprint != "" && compose.Blueprint.Name != filterBlueprint {
+				delete(composes, uuid)
+			}
+			if filterStatus != "" && compose.ImageBuilds[0].QueueStatus.ToString() != filterStatus {
+				delete(composes, uuid)
+			}
+			imageType, found := compose.ImageBuilds[0].ImageType.ToCompatString()
+			if filterImageType != "" && found && imageType != filterImageType {
+				delete(composes, uuid)
+			}
```
dvdhrm: I would move the `...ToCompatString()` up and then use `else if`. Your version is not wrong, but it might feel weird calling `delete` many times.

PR #230, opened 2020-02-13
Upload images from worker to compose using jobqueue API

ondrejbudai:
This PR grew bigger than I originally though but honestly, **I love it**. Things changed:

1) Images from worker to composer are now transported using jobqueue API. This enables us to implement remote workers (should be just fairly simple rebase of my existing PR).
2) `LocalTarget.Location` field is no more. The store is now responsible for managing all the build artifacts and keeping track of their locations. The LocalTarget's now only denoting if there's an image for an image build.
3) `Compose.Image` is no more. We can calculate its values when they are needed, no need to store them redundantly.
4) osbuild results and images are now stored per image build instead of per compose. This was TODO from #221 and it made sense to me to implement it in this PR.

### DIFF
--- a/internal/jobqueue/api_test.go
+++ b/internal/jobqueue/api_test.go
@@ -16,20 +16,20 @@ import (
 
 func TestBasic(t *testing.T) {
 	var cases = []struct {
-		Method         string
-		Path           string
-		Body           string
-		ExpectedStatus int
-		ExpectedJSON   string
+		Method           string
+		Path             string
+		Body             string
+		ExpectedStatus   int
+		ExpectedResponse string
 	}{
 		// Create job with invalid body
-		{"POST", "/job-queue/v1/jobs", ``, http.StatusBadRequest, ``},
+		{"POST", "/job-queue/v1/jobs", ``, http.StatusBadRequest, `invalid request: EOF`},
 		// Update job with invalid ID
-		{"PATCH", "/job-queue/v1/jobs/foo", `{"status":"RUNNING"}`, http.StatusBadRequest, ``},
+		{"PATCH", "/job-queue/v1/jobs/foo", `{"status":"RUNNING"}`, http.StatusBadRequest, `invalid compose id: invalid UUID length: 3`},
 		// Update job that does not exist, with invalid body
-		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", ``, http.StatusBadRequest, ``},
+		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", ``, http.StatusBadRequest, `invalid status: EOF`},
 		// Update job that does not exist
-		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", `{"image_build_id": 0, "status":"RUNNING"}`, http.StatusNotFound, ``},
+		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa", `{"image_build_id": 0, "status":"RUNNING"}`, http.StatusNotFound, `compose does not exist`},
 	}
 
 	for _, c := range cases {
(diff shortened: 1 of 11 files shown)


internal/jobqueue/api_test.go
```diff
@@ -37,7 +37,7 @@ func TestBasic(t *testing.T) {
 		registry := distro.NewRegistry([]string{"."})
 		api := jobqueue.New(nil, store.New(nil, distroStruct, *registry))
 
-		test.TestRoute(t, api, false, c.Method, c.Path, c.Body, c.ExpectedStatus, c.ExpectedJSON)
+		test.TestNonJsonRoute(t, api, false, c.Method, c.Path, c.Body, c.ExpectedStatus, &c.ExpectedJSON)
```
teg: Why did this and the next call change from `TestRoute` to `TestNonJsonRoute`? What is the difference in behaviour?

For the commit message: how/why are we now ignoring the response in some cases?

internal/store/store.go
```diff
@@ -445,18 +445,20 @@ func (s *Store) GetAllComposes() map[uuid.UUID]compose.Compose {
 	return composes
```
teg: General comment for the patch:
Very nice, but commit message should explain a bit more. Comment on (and in the future, even better if you had split the patch apart): which part is just a refactoring/renaming, what is the functional change (renaming the directory) and why is the image creation now necessary. All sort of obvious, but stating it in the commit message makes reviewing much easier.

internal/store/store.go
```diff
@@ -734,6 +742,42 @@ func (s *Store) UpdateImageBuildInCompose(composeID uuid.UUID, imageBuildID int,
...
+	currentCompose, exists := s.Composes[composeID]
+	if !exists {
+		return &NotFoundError{"compose does not exist"}
+	}
+
+	imageBuild := currentCompose.ImageBuilds[imageBuildID]
+	localTarget := imageBuild.GetLocalTarget()
+
+	if localTarget == nil {
+		return &NoLocalTargetError{fmt.Sprintf("image upload requested for compse %s and image build %d but it has no local target", composeID.String(), imageBuildID)}
```
teg: I like that you make sure to error out in this case, but a thought: wouldn't it be nice to treat this error the same way as if the endpoint was invalid? That way, if in the future we dropped the local target completely, the behaviour from the point of view of a worker that never received jobs with the local target set would be exactly the same.

internal/jobqueue/job.go
```diff
@@ -120,14 +120,19 @@ func (job *Job) Run() (*common.ComposeResult, error) {
...
 		switch options := t.Options.(type) {
 		case *target.LocalTargetOptions:
+			filename, _, err := d.FilenameFromType(job.OutputType)
+			if err != nil {
+				r = append(r, err)
+				continue
+			}
 
-			err = runCommand("cp", "-a", "-L", tmpStore+"/refs/"+result.OutputID+"/.", options.Location)
+			f, err := os.Open(tmpStore + "/refs/" + result.OutputID + "/" + filename)
```
teg: You mention that this may be slower, would be nice to do a before/after sanity test for how long this takes. My assumption would be that in both cases this would be barely noticeable.

internal/jobqueue/api_test.go
ondrejbudai: I've removed this commit and instead properly implemented the tests in the next commit.

internal/store/store.go
ondrejbudai: I think the size of this commit is kinda low and the renamings are obvious, therefore I think it's not worth splitting it up. I totally agree with the commit message, I think I got tired and omitted it, because as you said it's "all sort of obvious" :smile: I added a better commit message, feel free to propose more changes. :slightly_smiling_face:

internal/store/store.go
ondrejbudai: So what do you propose? I'm probably missing out what is a better solution from your point of view.

internal/jobqueue/job.go
ondrejbudai: I've reworded the commit message and added information that everything is good after this commit. :)

internal/store/store.go
teg: My idea was to make sure this would return 404. Let's discuss that separately, this is good to go :)

PR #231, opened 2020-02-13
RCM API: use the new PushComposeRequest method

msehnout:
see individual commits

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -72,7 +72,7 @@ func main() {
 	// Optionally run RCM API as well as Weldr API
 	if len(listeners) == 3 {
 		rcmListener := listeners[2]
-		rcmAPI := rcm.New(logger, store)
+		rcmAPI := rcm.New(logger, store, rpmmd.NewRPMMD())
 		go rcmAPI.Serve(rcmListener)
 	}
 
--- a/cmd/osbuild-dnf-json-tests/main.go
+++ b/cmd/osbuild-dnf-json-tests/main.go
@@ -1,0 +1,66 @@
+// This package contains tests related to dnf-json and rpmmd package.
+package main
+
+import (
+	"fmt"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path"
+)
+
+func main() {
+	// Tests that the package wrapping dnf-json works as expected
+	dir, err := setUpTemporaryRepository()
+	if err != nil {
+		log.Fatal("Failed to set up temporary repository:", err)
+	}
+	TestFetchChecksum(false, dir)
+	err = tearDownTemporaryRepository(dir)
+	if err != nil {
+		log.Print("Warning: failed to clean up temporary repository.")
+	}
+}
+
+func setUpTemporaryRepository() (string, error) {
+	dir, err := ioutil.TempDir("/tmp", "osbuild-composer-test-")
+	if err != nil {
+		return "", err
+	}
+	cmd := exec.Command("createrepo_c", path.Join(dir))
+	err = cmd.Start()
+	if err != nil {
+		return "", err
+	}
+	err = cmd.Wait()
+	if err != nil {
+		return "", err
+	}
+	return dir, nil
+}
+
+func tearDownTemporaryRepository(dir string) error {
+	return os.RemoveAll(dir)
+}
+
+func TestFetchChecksum(quiet bool, dir string) {
+	repoCfg := rpmmd.RepoConfig{
+		Id:        "repo",
+		Name:      "repo",
+		BaseURL:   fmt.Sprintf("file://%s", dir),
+		IgnoreSSL: true,
+	}
+	if !quiet {
+		log.Println("Running TestFetchChecksum on:", dir)
+	}
+	c, err := repoCfg.FetchChecksum()
+	if err != nil {
+		log.Fatal("Failed to fetch checksum:", err)
+	}
+	if c == "" {
+		log.Fatal("The checksum is empty")
+	}
+	log.Println("SUCCESS")
+}
--- a/cmd/osbuild-dnf-json-tests/main.go
+++ b/cmd/osbuild-dnf-json-tests/main.go
@@ -57,11 +57,12 @@ func TestFetchChecksum(quiet bool, dir string) {
 	if !quiet {
 		log.Println("Running TestFetchChecksum on:", dir)
 	}
-	c, err := repoCfg.FetchChecksum()
+	rpmMetadata := rpmmd.NewRPMMD()
+	_, c, err := rpmMetadata.FetchMetadata([]rpmmd.RepoConfig{repoCfg}, "platform:f31")
 	if err != nil {
 		log.Panic("Failed to fetch checksum:", err)
 	}
-	if c == "" {
+	if c["repo"] == "" {
 		log.Panic("The checksum is empty")
 	}
 	if !quiet {
--- a/cmd/osbuild-dnf-json-tests/main.go
+++ b/cmd/osbuild-dnf-json-tests/main.go
@@ -14,14 +14,16 @@ import (
 func main() {
 	// Tests that the package wrapping dnf-json works as expected
 	dir, err := setUpTemporaryRepository()
+	defer func(dir string) {
+		err := tearDownTemporaryRepository(dir)
+		if err != nil {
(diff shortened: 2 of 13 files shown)


internal/rpmmd/repository_test.go
```diff
@@ -0,0 +1,48 @@
+package rpmmd
```
teg: Very nice! Just as mentioned in the chat, I'd make this a part of the integration test, by creating a new binary that is shipped in the -test rpm subpackage.

internal/rpmmd/repository.go
```diff
@@ -282,3 +290,27 @@ func (pkg *PackageInfo) FillDependencies(rpmmd RPMMD, repos []RepoConfig) (err e
 	pkg.Dependencies, _, err = rpmmd.Depsolve([]string{pkg.Name}, repos, false)
 	return
 }
+
+// FetchChecksum returns a repository checksum for given repo config
```
teg: Hm, why cannot the current helpers be used for the same, the way weldr does? I'm probably missing something.

internal/rcm/api.go
```diff
@@ -121,16 +119,26 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
...
 		errorReason.Error = strings.Join(errors, ", ")
-		// TODO: handle error
-		_ = json.NewEncoder(writer).Encode(errorReason)
+		err = json.NewEncoder(writer).Encode(errorReason)
+		if err != nil {
+			// JSON encoding is clearly our fault, log it and fail. Unfortunately there is no sane way to return
+			// HTTP 500 here because the "writer" requires call to "WriteHeader" before the call to "Write" but
+			// there is no way to know the "Write" operation failed before we try it, but once we try it there is no
+			// way to change the return code.
+			if api.logger != nil {
```
teg: I'd just `panic()` in this case.

internal/rpmmd/repository.go
msehnout: Maybe I'm missing something as well, let me again check those other functions.

internal/rcm/api.go
msehnout: ok, sounds reasonable

internal/rpmmd/repository.go
msehnout: Well, there is a function `FetchPackageList`, but it will be super hard to read the code if I use `FetchPackageList` to get a checksum.

internal/rpmmd/repository.go
```diff
@@ -287,3 +295,27 @@ func (pkg *PackageInfo) FillDependencies(rpmmd RPMMD, repos []RepoConfig, module
 	pkg.Dependencies, _, err = rpmmd.Depsolve([]string{pkg.Name}, nil, repos, modulePlatformID, false)
 	return
 }
+
+// FetchChecksum returns a repository checksum for given repo config
+func (rc *RepoConfig) FetchChecksum() (string, error) {
```
ondrejbudai: Why isn't this a member function of `rpmmdImpl`? This way it's not mockable. I think we should rename `FetchPackageList` to something like `FetchRepositoriesMetadata` and not have this method. This way we can reduce the duplicate code and have mocking.

internal/rcm/api_test.go
```diff
@@ -68,61 +66,4 @@ func TestBasicRcmAPI(t *testing.T) {
 			t.Errorf("The response to %s request to %s should match this regex %s but returns %s", c.Method, c.Path, c.ExpectedBodyRegex, response_body)
 		}
 	}
-}
-
-func TestSubmitCompose(t *testing.T) {
```
ondrejbudai: If `FetchPackageList` is used instead of `FetchChecksum`, then we can keep this test due to the mocking support in `FetchPackageList`.

Also if we were to delete this test, I would prefer doing it in the same commit which adds the dnf dependency. However, mocking is the best way I think.

cmd/osbuild-dnf-json-tests/main.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func main() {
+	// Tests that the package wrapping dnf-json works as expected
+	dir, err := setUpTemporaryRepository()
+	if err != nil {
+		log.Fatal("Failed to set up temporary repository:", err)
+	}
+	TestFetchChecksum(false, dir)
+	err = tearDownTemporaryRepository(dir)
```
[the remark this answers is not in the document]
msehnout: Good point, thanks!

internal/rpmmd/repository.go
msehnout: Thanks for the suggestion. It is now implemented this way.

internal/rcm/api_test.go
msehnout: The test was reintroduced.

internal/common/types.go
```diff
@@ -250,6 +261,20 @@ func getDistributionMapping() map[string]int {
 	return mapping
 }
 
+func (dist *Distribution) ModulePlatformID() (string, error) {
+	mapping := map[Distribution]string{
+		// TODO: this could be refactored so we don't have these strings hard coded in two places
+		Fedora30: "platform:f30",
+		RHEL82:   "platform:el8",
```
ondrejbudai: f31 and f32 were merged, therefore we should add them here.

cmd/osbuild-dnf-json-tests/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"path"
+)
+
+func main() {
+	// Tests that the package wrapping dnf-json works as expected
+	dir, err := setUpTemporaryRepository()
+	defer func(dir string) {
```
ondrejbudai: This just a small nit - I don't think the `dir` parameter is needed, I'm fine with the function being closure.

internal/common/types.go
```diff
@@ -242,6 +252,7 @@ const (
 // getArchMapping is a helper function that defines the conversion from JSON string value
 // to Distribution.
 func getDistributionMapping() map[string]int {
+	// Don't forget to add module-platform-id mapping below
 	mapping := map[string]int{
 		"fedora-30": int(Fedora30),
 		"fedora-31": int(Fedora31),
```
ondrejbudai: F32 should go there as well

internal/common/types.go
```diff
@@ -250,6 +261,20 @@ func getDistributionMapping() map[string]int {
...
 
+func (dist *Distribution) ModulePlatformID() (string, error) {
+	mapping := map[Distribution]string{
+		// TODO: this could be refactored so we don't have these strings hard coded in two places
+		Fedora30: "platform:f30",
+		RHEL82:   "platform:el8",
+	}
+	id, exists := mapping[*dist]
+	if !exists {
+		return "", &CustomTypeError{reason: "Distribution does not have a module platform ID assigned"}
```
ondrejbudai: I think we should just panic here, this is clearly our mistake.

cmd/osbuild-composer/main.go
```diff
@@ -72,7 +72,7 @@ func main() {
 	// Optionally run RCM API as well as Weldr API
 	if len(listeners) == 3 {
 		rcmListener := listeners[2]
-		rcmAPI := rcm.New(logger, store)
+		rcmAPI := rcm.New(logger, store, rpmmd.NewRPMMD())
```
ondrejbudai: `rpm` variable can be reused here instead of creating new `RPMMD` instance

msehnout: yeah, I was thinking about it. Is it safe to share it?

internal/common/types.go
msehnout: Makes sense!

internal/common/types.go
msehnout: oh no, more merge conflicts :)

cmd/osbuild-composer/main.go
ondrejbudai: `type rpmmdImpl struct{}` - it doesn't have much state. :-D

internal/common/types.go
msehnout: Thinking about it again, I think we should keep it like this because we have the same logic in other transformation functions in this package.

internal/common/types.go
msehnout: It's there as part of other PRs.

msehnout: => it's already there because I rebased to the latest master :)

internal/common/types.go
teg: I would have preferred to look up the distro and query that, but let's not block on that. Not a fan of hardcoding strings more than once.

PR #234, opened 2020-02-14
Vagrantfile for integration tests

msehnout:
Requires #231 to work, otherwise it is described in the file itself.

Ignore all the commits except for the last one! The previous commits belong to the other PR. You can also safely ignore the CI failure, because the last commit is not related to CI.


msehnout:
See #255 for follow-up discussion.

PR #237, opened 2020-02-14
Vendor everywhere

larskarlitski:
See individual commits. The most important change is that it's including the `vendor` directory upstream.

We can remove the `rhel-8.2` branch after merging this.


larskarlitski:
> wow, there must be a different way, this is crazy.

What about it is crazy? This seems common practice in go upstreams and we had already included it in the `rhel-8.2` branch.

> Why not to run `go mod vendor` in `make srpm` and include `vendor` in .gitignore?

We're not using `make srpm` in all cases. Also, it depends on `tarball`, which just `git archive`s all things in git. Could be done, but seems more complex to me than just keeping the dependencies in tree.

msehnout:
> What about it is crazy? This seems common practice in go upstreams and we had already included it in the `rhel-8.2` branch.

Well, statically linking all the dependencies it one thing, but including all dependencies as source code directly in the repository is the crazy part. Just compare the the number of lines of code, this PR adds 260k lines to a project that has 10k lines. 

> We're not using make srpm in all cases. Also, it depends on tarball, which just git archives all things in git. Could be done, but seems more complex to me than just keeping the dependencies in tree.

Sure, but what is the reason to include all the dependencies. Is it the way RPMs are built in brew?

.travis.yml
```diff
@@ -8,6 +8,16 @@ addons:
...
+    - name: source-test
+      script:
+        - |
+          set -e
+          ./tools/prepare-source.sh
+          if [ ! $(git diff --quiet --exit-code) ]; then
+            echo
+            echo "Please include these changes in your branch: "
+            git diff
+          fi
```
dvdhrm: You can do multiline scripts in Travis without `sh -c`? I did not know this works.

larskarlitski: Welcome to the wonderful world of yaml. The `|` means everything on that level of indentation is interpreted as a [multi-line string](https://stackoverflow.com/questions/15540635/what-is-the-use-of-the-pipe-symbol-in-yaml).

PR #238, opened 2020-02-17
WIP: Execute integration tests in Travis CI

atodorov:


ondrejbudai:
@atodorov: Currently, @major is working on this, you might want to get in touch with him.

teg:
I don't think we can run the integration tests like this. Travis is running some old Ubuntu system, we would at least need to install the right dependencies, and even then, it is not really what we want to test, is it? We need to install our tests into the right target OS and run them there. I'll be looking into this next and coordinate with @major and @dvdhrm.

atodorov:
@teg right, I was playing with this to figure out things. Closing this PR.

PR #240, opened 2020-02-17
distro: add fedora 31 support

jkozol:
Images can be built for fedora 31. The pipeline generation and distro tests are based off of the fedora 30 ones. Repository information has also been added for the fedora 31 repos.


major:
FWIW, I was able to build the RPMs on Fedora 31 and use `composer-cli` to start building a compose.

msehnout:
Hurray, support for Fedora 31! :D 

I have one question though: We will soon have support for 4 different versions of Fedora. Did you consider creating some common code or do we want to avoid it on purpose?

PR #241, opened 2020-02-17
 distro: add fedora 32 support

jkozol:
Images can be built for fedora 32. The pipeline generation and distro tests are based off of the fedora 30 ones. Repository information has also been added for the fedora 32 repos.


jkozol:
@ondrejbudai Thanks for the review. I rebased.

PR #247, opened 2020-02-19
dnf-json: always specify a cache directory

teg:
dnf does not behave well when we don't specify a cache directory. Use /var/cache/osbuild-composer/rpmmd, as we did at one point in the past. Let systemd create the cache directory and set up the right user/group permissions for us.

This should avoid lots of stale cache directories around as reported by @msehnout.

I cherry-picked a patch from #198, to (hopefully) avoid a merge conflict.

### DIFF
--- a/distribution/osbuild-worker@.service
+++ b/distribution/osbuild-worker@.service
@@ -7,7 +7,6 @@ After=multi-user.target osbuild-composer.socket
 Type=simple
 PrivateTmp=true
 ExecStart=/usr/libexec/osbuild-composer/osbuild-worker
-CacheDirectory=osbuild-composer
 Restart=on-failure
 RestartSec=10s
 CPUSchedulingPolicy=batch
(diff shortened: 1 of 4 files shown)


bcl:
How much do we want to depend on external things like systemd and rpm for setting up things like directories? I guess it probably doesn't matter at this point with it already being so tightly dependent on systemd, but I tend to like it when things can stand on their own.

teg:
At the moment we are completely dependent on systemd to be able to run as non-root.

distribution/osbuild-worker@.service
```diff
@@ -7,7 +7,7 @@ After=multi-user.target osbuild-composer.socket
 Type=simple
 PrivateTmp=true
 ExecStart=/usr/libexec/osbuild-composer/osbuild-worker
-CacheDirectory=osbuild-composer
+CacheDirectory=osbuild-worker
```
msehnout: I'd prefer not to have it here if we don't use it. It seems to me that it is just confusing and the effort to remove it and reintroduce it is minimal.

teg: Yeah, I don't mind. I'm happy to merge a PR that drops it. I just did the minimal change for now to move it out of the way (sharing it caused problems with the wrong owner/group).

teg: Correction, I'll drop this. I have to rebase on top of your RCM PR anyway.

PR #254, opened 2020-02-24
Build RPMs using GitHub actions

major:


.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
+name: RPM Build
+
+on: [pull_request]
```
dvdhrm: I think you want `push` as well, to make sure updated PRs run this as well:

`on: [push, pull_request]`

.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
...
+          echo "fastestmirror=1" >> /etc/dnf/dnf.conf
+          echo "install_weak_deps=0" >> /etc/dnf/dnf.conf
+          dnf -y upgrade
+          dnf -y install dnf-plugins-core git make rpm-build
+          mkdir rpms
+
+      - uses: actions/checkout@v2
+
+      - name: Checkout submodules
+        shell: bash
```
dvdhrm: Isn't that the default?

.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive
```
dvdhrm: What is the reason for the extraheader? Or requiring protocol-version-2?

.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Install RPM build dependencies
+        run: |
+          dnf -y builddep \
+            osbuild-composer/golang-github-osbuild-composer.spec \
+            osbuild/osbuild.spec
```
dvdhrm: `actions/checkout@v2` clones into the working-dir, so you need to drop `osbuild-composer/` from the first path, right?

.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
+name: RPM Build
+
+on: [pull_request]
+
+jobs:
+  rpm_build:
+    name: "Build RPMs"
+    runs-on: ubuntu-latest
+    container: registry.gitlab.com/majorhayden/containers/rpm-builder:latest
```
msehnout: Is there a reason to use private container instead of the official fedora one? I can see you install dnf/git/rpmbuild/etc. below...

.github/workflows/build_rpms.yml
```diff
@@ -0,0 +1,50 @@
...
+            osbuild-composer/golang-github-osbuild-composer.spec \
+            osbuild/osbuild.spec
+
+      - name: Build RPMs
+        run: |
+          pushd osbuild-composer
+            make rpm
+          popd
+          pushd osbuild
+            make rpm
```
teg: I think it should be sufficient to build osbuild-composer, no? @ondrejbudai is working on dropping the submodule and move all the tests that use it to proper integration tests.

ondrejbudai:
This looks very outdated, we now have rpm build in the github workflow. Also, it doesn't contain any emojis. Closing, feel free to reopen if I'm wrong.

PR #256, opened 2020-02-25
Download images inside osbuild-tests

atodorov:
For every image built we download it via `composer-cli compose image` then delete the file using globs to match only on the UUID. 

Globbing doesn't seem t produce errors so maybe I'm not using it correctly.

### DIFF
--- a/cmd/osbuild-tests/main.go
+++ b/cmd/osbuild-tests/main.go
@@ -207,7 +207,7 @@ func runComposerCLI(quiet bool, command ...string) json.RawMessage {
 	cmd := exec.Command("composer-cli", command...)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		log.Fatalf("Colud not create command: " + err.Error())
+		log.Fatalf("Could not create command: " + err.Error())
 	}
 
 	if !quiet {
--- a/cmd/osbuild-tests/main.go
+++ b/cmd/osbuild-tests/main.go
@@ -61,13 +61,6 @@ func main() {
 
 	// Full integration tests
 	testCompose("ami")
-	testCompose("ext4-filesystem")
-	testCompose("openstack")
-	testCompose("partitioned-disk")
-	testCompose("qcow2")
-	testCompose("tar")
-	testCompose("vhd")
-	testCompose("vmdk")
 }
 
 func testCompose(outputType string) {
@@ -93,6 +86,8 @@ func testCompose(outputType string) {
 	if status != "FINISHED" {
 		log.Fatalf("Unexpected compose result: %s", status)
 	}
+
+	runComposerCLI(false, "compose", "image", uuid.String())
 }
 
 func startCompose(name, outputType string) uuid.UUID {


teg:
Looking good, but I don't understand this sentence: "Globbing doesn't seem t produce errors so maybe I'm not using it correctly."

atodorov:
> Looking good, but I don't understand this sentence: "Globbing doesn't seem t produce errors so maybe I'm not using it correctly."

I mean that if I change my code to look like:

`files, err := filepath.Glob("non-existing" + uuid.String() + "*")`

then the list of files is empty and there are no errors returned. Which means if there's some issue this will be leaving files laying around on the filesystem and not cleaning them up. It's not a big concern but I was wondering why I don't get any errors.

atodorov:
@teg I was looking into making `testCompose()` take an optional parameter so we can download only once (not for every outputType) to save time but golang doesn't support optional params and parameter default values. 

OTOH if I create a second function it won't work b/c `testCompose()` deletes the actual compose after it is done. I'm not sure how to design a function which will build, download & cleanup after itself without copying the entire body of `testCompose()`.

atodorov:
@teg, @ondrejbudai, @larskarlitski can I get +/- on this PR?

cmd/osbuild-tests/main.go
```diff
@@ -84,6 +85,23 @@ func testCompose(outputType string) {
...
 	}
+
+	if download {
+		runComposerCLI(false, "compose", "image", uuid.String())
+		defer deleteImage(uuid)
+	}
+}
+
+func deleteImage(uuid uuid.UUID) {
+	files, err := filepath.Glob(uuid.String() + "*")
```
larskarlitski: We ought to know the right filename, no? I'm wary of deleting files based on a glob.

atodorov:
> Hm I'm missing why this is necessary and if it is, why it is only necessary for AMI images.

@larskarlitski - that's part of the sanity testing done here. Just download the file to exercise this functionality. 

Because this operation takes time and depends on underlying storage speed and size of the actual image I wanted to perform the download action just once. Hence it's done only for AMI (can be any other output type, ami is just the first one).

cmd/osbuild-tests/main.go
atodorov: I guess I can try `image-info` to get the exact file name. Didn't know how to do that originally and globbing was the fastest.

larskarlitski: That won't work, because you need a file name to pass to `image-info`.

It seems to me like `composer-cli` doesn't have a way to output the image name. However, I think that globbing in a directory we know nothing about might lead to bugs later. I've proposed https://github.com/osbuild/osbuild-composer/pull/289, which creates a new temporary directory to work in. If that's accepted, please you use that functionality here as well.

larskarlitski:
> Because this operation takes time and depends on underlying storage speed and size of the actual image I wanted to perform the download action just once. Hence it's done only for AMI (can be any other output type, ami is just the first one).

That makes sense, thanks! Please include this reasoning in the commit message for future reference.

larskarlitski:
I've changed my mind on this, thanks to a conversation with @teg. There is no need to call `testCompose` for each output type. We're testing that that works in different tests. These are about testing that `composer-cli` can talk to `osbuild-composer` in the way it expects. It is enough to try one output type.

Do you mind changing this? Then you can drop the `bool` as well and just unconditionally download.

atodorov:
Rebased & updated

cmd/osbuild-tests/main.go
```diff
@@ -60,14 +60,7 @@ func main() {
...
 	// Full integration tests
-	testCompose("ami")
-	testCompose("ext4-filesystem")
-	testCompose("openstack")
-	testCompose("partitioned-disk")
-	testCompose("qcow2")
-	testCompose("tar")
-	testCompose("vhd")
-	testCompose("vmdk")
+	testCompose("ami", true)
```
larskarlitski: `testCompose` doesn't take the second argument anymore.

PR #259, opened 2020-02-26
First part of the porting the current ./test/run test suite to Go

ondrejbudai:
Image info tests are running, therefore I think it's time for reviewing. Please note that this is meant to be a part of the integration test suite (therefore run on a pristine Fedora VM, not on Travis CI).

Changes from the Python version:
1) osbuild is called from `PATH`, not from the submodule.
2) `--arches` argument is gone. Instead only the tests for the current arch are run.
  Go's `flag` package doesn't support arguments with multiple values, therefore I just dropped it for now. We can revisit this later.
3) `--distros` argument is gone. Instead only the tests for the host distro are run.
  The same reason as the previous point, we can also revisit this later.
4) `--cases` argument is now a positional argument. (`osbuild-image-tests CASES... `), when no `CASES` are given, all the tests are run
5) For diffing the actual output from `image-info` and the expected output Go's `cmp` library is now used. I think this approach is much simpler (no need for two temporary files) and I even like its output more.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -68,6 +68,7 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 %gobuild -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
 %gobuild -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
+%gobuild -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
 
 %install
 install -m 0755 -vd                                         %{buildroot}%{_libexecdir}/osbuild-composer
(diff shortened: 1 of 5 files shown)


golang-github-osbuild-composer.spec
```diff
@@ -68,6 +68,7 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 %gobuild -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
 %gobuild -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
+%gobuild -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
```
atodorov: Can you add this to `make build` as well, would be useful for running only the tests from a git checkout.

ondrejbudai:
> I think it is ok, however, to start out mimicking the current behaviour and then build on that to move it over to using the weldr API.

Yeah, that was exactly the plan, start by re-implementing the suite without touching the testcases and then switch from osbuild to osbuild-composer. I think this approach is much less error-prone. :-)

PR #261, opened 2020-02-26
osbuild weldr test

bcl:
This is a rewrite of the osbuild-test code that was running composer-cli from inside Go. It includes a client library, tests for the library, and integration tests (which I called checks to prevent confusion with the unit tests in _test.go files).

It includes blueprint integration tests, 2 of which fail on osbuild-composer (I'll work on fixing them).

It doesn't currently include a test for the blueprint diff, the osbuild-composer implementation is incomplete and I am unsure how useful diff really is in practice so it currently prints 'SKIP' for it.

Output looks like this with osbuild-composer:

```
[root@composer-rawhide osbuild-composer]# go build -o osbuild-test ./cmd/osbuild-tests/ && ./osbuild-test
2020/02/26 17:55:06 Running API check on /run/weldr/api.socket
2020/02/26 17:55:06 OK: status request
2020/02/26 17:55:06 INFO: Running tests against: osbuild-composer devel server using V1 API
2020/02/26 17:55:06 INFO: Running API V0 checks
2020/02/26 17:55:06 OK: List blueprint changes was successful
2020/02/26 17:55:07 OK: Depsolve a blueprint was successful
2020/02/26 17:55:07 SKIP: Diff of blueprint changes was skipped, needs to be implemented
2020/02/26 17:55:08 OK: Freeze a blueprint was successful
2020/02/26 17:55:08 FAIL: Tag a blueprint Tag failed: ClientError: unexpected end of JSON input
2020/02/26 17:55:08 OK: Bump Blueprint Version number was successful
2020/02/26 17:55:08 OK: DELETE blueprint was successful
2020/02/26 17:55:08 OK: DELETE blueprint changes from workspace was successful
2020/02/26 17:55:08 OK: DELETE new blueprint from workspace was successful
2020/02/26 17:55:08 OK: Get JSON Blueprint was successful
2020/02/26 17:55:08 OK: Get TOML Blueprint was successful
2020/02/26 17:55:08 OK: List blueprints was successful
2020/02/26 17:55:08 OK: POST of a JSON blueprint was successful
2020/02/26 17:55:08 OK: POST JSON blueprint to workspace was successful
2020/02/26 17:55:08 OK: POST of a TOML blueprint was successful
2020/02/26 17:55:08 OK: POST TOML blueprint to workspace was successful
2020/02/26 17:55:08 FAIL: Undo blueprint changes failed to undo change: Wrong package in the blueprint: {bash 0.5.*}
2020/02/26 17:55:08 FAIL: One or more V0 API checks failed
2020/02/26 17:55:08 INFO: Running API V1 checks
2020/02/26 17:55:08 OK: ALL V1 API checks were successful

```

### DIFF
--- a/internal/client/blueprints.go
+++ b/internal/client/blueprints.go
@@ -10,158 +10,158 @@ import (
 
 // PostTOMLBlueprintV0 sends a TOML blueprint string to the API
 // and returns an APIResponse
-func PostTOMLBlueprintV0(socket, blueprint string) *APIResponse {
-	body, err := PostTOML(socket, "/api/v0/blueprints/new", blueprint)
-	if err != nil {
-		return err
+func PostTOMLBlueprintV0(socket, blueprint string) (*APIResponse, error) {
+	body, resp, err := PostTOML(socket, "/api/v0/blueprints/new", blueprint)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // PostTOMLWorkspaceV0 sends a TOML blueprint string to the API
 // and returns an APIResponse
-func PostTOMLWorkspaceV0(socket, blueprint string) *APIResponse {
-	body, err := PostTOML(socket, "/api/v0/blueprints/workspace", blueprint)
-	if err != nil {
-		return err
+func PostTOMLWorkspaceV0(socket, blueprint string) (*APIResponse, error) {
+	body, resp, err := PostTOML(socket, "/api/v0/blueprints/workspace", blueprint)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // PostJSONBlueprintV0 sends a JSON blueprint string to the API
 // and returns an APIResponse
-func PostJSONBlueprintV0(socket, blueprint string) *APIResponse {
-	body, err := PostJSON(socket, "/api/v0/blueprints/new", blueprint)
-	if err != nil {
-		return err
+func PostJSONBlueprintV0(socket, blueprint string) (*APIResponse, error) {
+	body, resp, err := PostJSON(socket, "/api/v0/blueprints/new", blueprint)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // PostJSONWorkspaceV0 sends a JSON blueprint string to the API
 // and returns an APIResponse
-func PostJSONWorkspaceV0(socket, blueprint string) *APIResponse {
-	body, err := PostJSON(socket, "/api/v0/blueprints/workspace", blueprint)
-	if err != nil {
-		return err
+func PostJSONWorkspaceV0(socket, blueprint string) (*APIResponse, error) {
+	body, resp, err := PostJSON(socket, "/api/v0/blueprints/workspace", blueprint)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // DeleteBlueprintV0 deletes the named blueprint and returns an APIResponse
-func DeleteBlueprintV0(socket, bpName string) *APIResponse {
-	body, err := DeleteRaw(socket, "/api/v0/blueprints/delete/"+bpName)
-	if err != nil {
-		return err
+func DeleteBlueprintV0(socket, bpName string) (*APIResponse, error) {
+	body, resp, err := DeleteRaw(socket, "/api/v0/blueprints/delete/"+bpName)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // DeleteWorkspaceV0 deletes the named blueprint's workspace and returns an APIResponse
-func DeleteWorkspaceV0(socket, bpName string) *APIResponse {
-	body, err := DeleteRaw(socket, "/api/v0/blueprints/workspace/"+bpName)
-	if err != nil {
-		return err
+func DeleteWorkspaceV0(socket, bpName string) (*APIResponse, error) {
+	body, resp, err := DeleteRaw(socket, "/api/v0/blueprints/workspace/"+bpName)
+	if resp != nil || err != nil {
+		return resp, err
 	}
 	return NewAPIResponse(body)
 }
 
 // ListBlueprintsV0 returns a list of blueprint names
-func ListBlueprintsV0(socket string) ([]string, *APIResponse) {
-	body, err := GetJSONAll(socket, "/api/v0/blueprints/list")
+func ListBlueprintsV0(socket string) ([]string, *APIResponse, error) {
+	body, resp, err := GetJSONAll(socket, "/api/v0/blueprints/list")
+	if resp != nil || err != nil {
+		return nil, resp, err
+	}
+	var list BlueprintsListV0
+	err = json.Unmarshal(body, &list)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	var resp BlueprintsListV0
-	jerr := json.Unmarshal(body, &resp)
-	if jerr != nil {
-		return nil, clientError(err)
-	}
-	return resp.Blueprints, nil
+	return list.Blueprints, nil, nil
 }
 
 // GetBlueprintInfoTOMLV0 returns the requested blueprint as a TOML string
-func GetBlueprintInfoTOMLV0(socket, bpName string) (string, *APIResponse) {
-	body, err := GetRaw(socket, "GET", "/api/v0/blueprints/info/"+bpName+"?format=toml")
-	if err != nil {
-		return "", err
+func GetBlueprintInfoTOMLV0(socket, bpName string) (string, *APIResponse, error) {
+	body, resp, err := GetRaw(socket, "GET", "/api/v0/blueprints/info/"+bpName+"?format=toml")
+	if resp != nil || err != nil {
+		return "", resp, err
 	}
-	return string(body), nil
+	return string(body), nil, nil
 }
 
 // GetBlueprintsInfoJSONV0 returns the requested blueprints and their changed state
-func GetBlueprintsInfoJSONV0(socket, bpName string) (BlueprintsInfoV0, *APIResponse) {
-	body, err := GetRaw(socket, "GET", "/api/v0/blueprints/info/"+bpName)
+func GetBlueprintsInfoJSONV0(socket, bpName string) (BlueprintsInfoV0, *APIResponse, error) {
+	body, resp, err := GetRaw(socket, "GET", "/api/v0/blueprints/info/"+bpName)
+	if resp != nil || err != nil {
+		return BlueprintsInfoV0{}, resp, err
+	}
+	var info BlueprintsInfoV0
+	err = json.Unmarshal(body, &info)
 	if err != nil {
-		return BlueprintsInfoV0{}, err
+		return BlueprintsInfoV0{}, nil, err
 	}
-	var resp BlueprintsInfoV0
-	jerr := json.Unmarshal(body, &resp)
-	if jerr != nil {
-		return BlueprintsInfoV0{}, clientError(err)
-	}
-	return resp, nil
+	return info, nil, nil
 }
 
 // GetBlueprintsChangesV0 returns the changes to the listed blueprints
-func GetBlueprintsChangesV0(socket string, bpNames []string) (BlueprintsChangesV0, *APIResponse) {
+func GetBlueprintsChangesV0(socket string, bpNames []string) (BlueprintsChangesV0, *APIResponse, error) {
 	names := strings.Join(bpNames, ",")
-	body, err := GetRaw(socket, "GET", "/api/v0/blueprints/changes/"+names)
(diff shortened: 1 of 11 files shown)


jikortus:
@bcl I think we shouldn't drop the tests that actually use `composer-cli` (maybe just reduce them to some reasonable subset?). IMHO we could make use of both approaches (testing via API + also retain testing via `composer` cli, even though those tests would rather cover the CLI tool rather than `osbuild-composer`).

pkg/client/client.go
```diff
@@ -0,0 +1,232 @@
...
+// String returns the error id and message as a string
+func (r *APIErrorMsg) String() string {
+	return fmt.Sprintf("%s: %s", r.ID, r.Msg)
+}
+
+// APIResponse is returned with status code 400 and may contain a list of errors
+// If Status is true the Errors list will not be included or will be empty.
+// When Status is false it will include at least one APIErrorMsg with details about the error.
+// It also implements the error interface so that it can be used in place of error
+type APIResponse struct {
```
ondrejbudai: I think the `APIReponse` structure is confusing. It's fulfilling the `error` interface but if `response.Status == true` then it's not an error.

I think we should stick to idiomatic as Go as possible. If a function can return an error, it should be the last return value in the form of `error` interface. If no error happened, that value should be nil. Variables with name `err` should always be typed as `error` and not as a type implementing `error`.

Imho all the functions which currently have `*APIResponse` struct as the last return type should return `error` instead. We should have several different Error structs denoting where the error happened (e.g. StatusCodeError, TransportError, JSONError).

What do you think? I'm very open to a discussion, this is imo a really interesting topic.

pkg/client/client.go
```diff
@@ -0,0 +1,232 @@
...
+// GetRaw returns raw data from a GET request
+// Errors from the client and from the API are returned as an APIResponse
+func GetRaw(socket, method, path string) ([]byte, *APIResponse) {
+	resp, err := Request(socket, method, path, "", map[string]string{})
+	if err != nil {
+		return nil, clientError(err)
+	}
+
+	// Convert the API's JSON error response to an error type and return it
+	if resp.StatusCode == 400 {
```
ondrejbudai: The caller of the `GetRaw` function can distinguish if the status code was 400 or not. Is it enough? We might also want to test other status codes.

pkg/client/client.go
bcl: I'm really not sure there is a right way to handle this. We have (at least) 2 completely different kinds of errors. API 'errors' aren't realy errors in the go sense, they are error responses from the API call which can in some cases have multiple errors included.

I did conflate things a bit though by converting go errors to 'ClientError' to make handling things a bit less complicated.

I don't want to add a bunch of error structs that have to be kept in-sync with the API errors, that's why they include an error ID and a message, so that the client doesn't need to know exactly what could fail.

I could split things out though and return go errors as a separate 'error' from the API response.

pkg/client/client.go
bcl: Error 400 is specific to the API responses, it indicates that the body should be a JSON API error response. All other errors, like 404, should be handled by the caller.

bcl:
> * Please keep the library in `internal/` for now (exposing this should be a separate discussion, so far we intentionally kept everything internal)
> * Please leave the existing `composer-cli` tests as-is for now and introduce this as a new test-binary (feel free to rename the existing one for clarity). We should have a discussion about what belongs where and the direction we want to go in. I agree that we probably want to keep some minimal tests using `composer-cli` to verify that there are no quirks there, but that the bulk of it should be using this library to test the API directly, as that is what we actually expose and support.

Ok, that's why I kept them as a separate commit, I wasn't sure how well that would be received. I do think it is important to have a public library that can be used. Otherwise we will continue to use composer-cli hacks, and like I've said, it wasn't really written for that.

I'll move this to a new binary for now. I still think composer-cli has no business being called from inside go, parsing its --json output is a complete hack.

You may want to hold any other reviews until I get things moved around and address @ondrejbudai comments about errors.

jikortus:
> 
> > I'll move this to a new binary for now. I still think composer-cli has no business being called from inside go, parsing its --json output is a complete hack.
> 
> Well, the point is to test that composer-cli keeps working, I agree that except for actually testing composer-cli, it would make no sense.

I'd say testing composer-cli definitely makes sense. It's not testing osbuild-composer per se, but rather the integration with composer-cli, which we also need to cover. Other question is whether we want to test composer-cli with `--json`. I think the answer is yes and no :) - we should use such an option again only for testing that composer-cli works properly and outputs the expected json, but not use it as a source of a json output that would be subsequently parsed and evaluated for some specific purpose. Just my two cents :).

teg:
> I think the answer is yes and no :) - we should use such an option again only for testing that composer-cli works properly and outputs the expected json, but not use it as a source of a json output that would be subsequently parsed and evaluated for some specific purpose. Just my two cents :).

I agree, that's a nice way of putting it!

internal/weldrcheck/check.go
```diff
@@ -0,0 +1,108 @@
...
+	if apiVersion > 0 {
+		log.Println("INFO: Running API V1 checks")
+		passV1 := runV1Checks(socket)
+
+		pass = pass && passV1
+	}
+
+	if !pass {
+		os.Exit(1)
+	}
```
atodorov: @bcl why aren't we using any sort of assert library here ?

internal/weldrcheck/blueprints.go
```diff
@@ -0,0 +1,1109 @@
...
+		name="root"
+		password="qweqweqwe"
+		`
+	resp, err := client.PostTOMLBlueprintV0(c.socket, bp)
+	if err != nil {
+		log.Printf("FAIL: %s failed with a client error: %s", name, err)
+		return false
+	}
+	if resp.Status != true {
+		log.Printf("FAIL: %s failed: %s", name, resp)
```
atodorov: Same question here: wouldn't an assert library make this easier ?

internal/client/blueprints.go
```diff
@@ -0,0 +1,167 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// PostTOMLBlueprintV0 sends a TOML blueprint string to the API
+// and returns an APIResponse
+func PostTOMLBlueprintV0(socket, blueprint string) *APIResponse {
```
teg: Suggestion for future improvement:
It would be nice to pass in an already opened socket here, rather than a string, and also to pass in an Blueprint type and let the lower layers marshal it.

internal/client/blueprints.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// ListBlueprintsV0 returns a list of blueprint names
+func ListBlueprintsV0(socket string) ([]string, *APIResponse) {
+	body, err := GetJSONAll(socket, "/api/v0/blueprints/list")
+	if err != nil {
+		return nil, err
+	}
+	var resp BlueprintsListV0
+	jerr := json.Unmarshal(body, &resp)
```
teg: I think it would make sense to pass an `interface{}` type to `GetJSONAll()` and let that deal with the `Unmarshal()`, similarly to my suggestion above for letting the lower layers do the marshalling.

internal/client/client.go
```diff
@@ -0,0 +1,232 @@
...
+// Request handles sending the request, handling errors, returning the response
+// socket is the path to a Unix Domain socket
+// path is the full URL path, including query strings
+// body is the data to send with POST
+// headers is a map of header:value to add to the request
+//
+// If it is successful a http.Response will be returned. If there is an error, the response will be
+// nil and error will be returned.
+func Request(socket, method, path, body string, headers map[string]string) (*http.Response, error) {
+	client := http.Client{
```
teg: I think this should be passed in from the highest level, making it more flexible and type-safe.

internal/client/client.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, clientError(err)
+	}
+
+	return body, nil
+}
+
+// GetJSONAll returns all JSON results from a GET request using offset/limit
+// This function makes 2 requests, the first with limit=0 to get the total number of results,
+// and then with limit=TOTAL to fetch all of the results.
```
teg: Is this due to some limitation in the weldr API? Is there no way to get this in one call? What happens if you drop the limit argument, or set it to 9999999999?

internal/client/client.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, clientError(err)
+	}
+
+	return responseBody, nil
+}
+
+// PostTOML sends a POST with TOML data and the Content-Type header set to "text/x-toml"
+// It returns the raw response data or errors as an APIResponse
+func PostTOML(socket, path, body string) ([]byte, *APIResponse) {
```
teg: Would be nice to pass the body as `interface{}` here and do the marshalling to TOML in this function (same for JSON below).

internal/client/blueprints.go
```diff
@@ -10,158 +10,158 @@ import (
 
 // PostTOMLBlueprintV0 sends a TOML blueprint string to the API
 // and returns an APIResponse
-func PostTOMLBlueprintV0(socket, blueprint string) *APIResponse {
-	body, err := PostTOML(socket, "/api/v0/blueprints/new", blueprint)
-	if err != nil {
-		return err
+func PostTOMLBlueprintV0(socket, blueprint string) (*APIResponse, error) {
+	body, resp, err := PostTOML(socket, "/api/v0/blueprints/new", blueprint)
+	if resp != nil || err != nil {
```
teg: I believe the distinction between client errors and API errors are not usually made in Golang. Would it be possible to treat them all as `error` return values and switch over their types from the caller if we want to make the distinction instead?

internal/weldrcheck/check.go
bcl: I'm not sure how using a library would help. The final return code indicates whether or not all the tests passed, and the individual failures are in the output. I don't want to fail early, I want to continue running all of the tests.

internal/client/blueprints.go
bcl: I hadn't considered an open socket. I'm not sure how much benefit that would have, or if it even really matters one way or the other.

As for passing in the blueprint, that could be done at a higher layer. I was trying to keep this as generic as I could, only depending on the return types.

internal/client/blueprints.go
bcl: I really don't like interface{}, it should be used sparingly. It's too easy to lean on it and lose type safety. We know what the response type should be at this point, so it should be enforced.

internal/client/client.go
bcl: I think I need more context here. Pass what in?

internal/client/client.go
bcl: Yeah, the output from the api defaults to 20 items so you have to set a higher limit to get them all. Setting an arbitrary number will work, until it doesn't, so it's safer to do it this way.

internal/client/blueprints.go
bcl: Well, that's kinda how I had it originally but @ondrejbudai wanted separate errors so I changed it. I had an API Error that implemented the error interface, but has the extra functions to get multiple errors from it, and was converting client errors to a new error type id and message.

internal/client/blueprints.go
teg: In general, we try to open the sockets at the highest possible level, and not in the libraries. That way they can be more flexible (maybe for writing tests?). In practice it would probably not matter for this, would just be about consistency and typesafety.

As for the Blueprint type: if you use that rather than `string` and do the marshalling internally, then you enforce both that this sends actual blueprints, and that the marshalling is TOML, reflecting the name of the function.

internal/client/blueprints.go
teg: My point was just to push the `json.Unmarshal()` into `GetJSONAll()`, which would encapsulate the JSON unmarshalling. The fact that the type is `interface{}` is just because that's what `json.Unmarshal()` takes.

internal/client/client.go
teg: The `http.Client{}` object. Rather than the `socket` as a string.

internal/weldrcheck/check.go
atodorov: > I'm not sure how using a library would help. The final return code indicates whether or not all the tests passed, and the individual failures are in the output

That's exactly what go test is doing as well. I don't see a reason for us not to follow the standard way of writing tests in this language. Even though `if !pass { os.Exit(1) }` isn't a big deal why do we have to do this everywhere when a simple assert will do the same job.

I'll see if I can do a POC so we can evaluate the differences between the two approaches. 

```
go test -failfast 
    Do not start new tests after the first test failure
```
If you don't want to fail early then we simply don't use this flag (which I think isn't on by default) so this looks like a non issue.

bcl: But these are not standard tests. It is an integration test that requires that the server (whichever one you want to run it against) is already running. We don't want it to run when running go test ...

internal/client/client.go
bcl: Ah! right. That would also allow for the timeout to be adjusted at setup time.

internal/weldrcheck/check.go
atodorov: Yes, they are not. This discussion is now in https://github.com/osbuild/osbuild-composer/pull/282, if you want to follow.

PR #263, opened 2020-02-28
Add a linter to this project

ondrejbudai:
I would like to introduce linting of our Go code in the project. I think this could help us to detect some errors before the merge button is pressed. I chose golangci-lint as our linter as it's a popular project used by many big projects (podman, openshift), and it aggregates multiple Go linters so more errors can be caught. To demonstrate its functionality, I run it on the current master. Then, I chose one example for each error type. The results are here:

### errcheck (45 cases)
```
errcheck in cmd/osbuild-tests/main.go:
Error return value of `cmd.Start` is not checked:
	cmd.Start()
```

There are 45 cases of unchecked errors in our code. It's surely annoying to do the check every time (in some cases god bless exceptions) but without any checking of errors we will never know about them. I think we should always react - if it's clearly programming mistake, let's panic. If it's something minor, pass it to the caller or log it. However, never let it escape!

### gosimple/S1002 (4 cases)
```
gosimple in cmd/osbuild-tests/main.go:
S1002: should omit comparison to bool constant, can be simplified to `!reply.Status`:
	if reply.Status != true {
```

A stylistic thing, I think it makes reading code easier.

### gosimple/S1004 (4 cases)
```
gosimple in internal/osbuild/source_test.go:
S1004: should use !bytes.Equal(gotBytes, tt.args.data) instead:
			if bytes.Compare(gotBytes, tt.args.data) != 0 {
```

I think this is very reasonable.

### gosimple/S1005 (3 cases)
```
gosimple in internal/common/types.go:
S1005: should omit value from range; this loop is equivalent to `for k := range ...`:
	for k, _ := range mapping {
```

This might be a bit annoying, I'm not sure about it but I decided to fix it, there are just 3 cases.

### gosimple/S1011 (2 cases)
```
gosimple in internal/weldr/api.go:
S1011: should replace loop with `repos = append(repos, api.distro.Repositories(api.arch)...)`:
	for _, repo := range api.distro.Repositories(api.arch) {
```

Very reasonable.

### gosimple/S1028 (1 case)
```
gosimple in cmd/osbuild-worker/main.go:
S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)):
		return nil, errors.New(fmt.Sprintf("couldn't create job, got %d: %s", response.StatusCode, r))
```

Nice catch, gosimple is trying to teach you how to use Go more efficiently.

### govet/composites (22 cases)
```
govet in internal/distro/fedora30/distro.go:
composites: `github.com/osbuild/osbuild-composer/internal/osbuild.ChronyStageOptions` composite literal uses unkeyed fields:
		p.AddStage(osbuild.NewChronyStage(&osbuild.ChronyStageOptions{ntpServers}))
```

According to govet it should be:
```
        p.AddStage(osbuild.NewChronyStage(&osbuild.ChronyStageOptions{Timeservers: ntpServers}))
```
This rule is about not specifying the names of fields when initializing structs. I decided to **turn it off**, I think in a lot of cases it's actually better to omit a field name, so the code can be shorter and easier to read.

### ineffassign (1 case)
```
ineffassign in internal/weldr/api.go:
ineffectual assignment to `err`:
	packages, checksums, err := api.rpmmd.Depsolve(...)
	if err != nil {
		return nil, nil, nil, err
	}

	buildPackages := []rpmmd.PackageSpec{}
	if outputType != "" {
		buildSpecs, err := api.distro.BuildPackages(arch)
		if err != nil {
			return nil, nil, nil, err
		}
		buildPackages, _, err = api.rpmmd.Depsolve(...) <<< HERE IS THE ERROR
	}

	return packages, buildPackages, checksums, err
```

This is a rather interesting catch. One might think that the error from the last depsolve gets returned by the last return statement. However, that's not true! `buildSpecs, err := api.distro.BuildPackage(arch)` creates a new `err` variable which shadows the one from outer scope. Therefore, the depsolve error is just thrown away when the if block is left, and `err` with the value of `nil` is returned.

Another example from the past (I fixed it when firstly playing with linting):

```
ineffassign in internal/distro/fedora30/distro.go:
ineffectual assignment to `enabledServices`:
        enabledServices = append(enabledServices, s.Enabled...)
        enabledServices = append(disabledServices, s.Disabled...)
```

Yeah, this is a clear bug!

### deadcode (1 case)
```
deadcode in internal/jobqueue/job.go:
`runCommand` is unused (deadcode):
    func runCommand(command string, params ...string) error {
```

It's nice not to have any dead code, right?

## Full log
https://gist.github.com/ondrejbudai/5697c4ace57ec5024d48f1a30c55bfc4

## Can you lint the past?
Yes, you can! I decided it would be awesome to lint not only the current master but every commit in the master branch. Surprisingly for me, the code in many commits cannot be built (and therefore linted). Fair amount of PRs have broken code prior the last commit. Is that bad? I don't know.

Overall, the results were better than expected - I didn't find any other kind of errors than those currently in master. There were about 8 cases of `ineffassign`, many of them clear bugs. Now, we will be able to detect them with proper linting on CI! Otherwise, nothing major appeared, apparently we are great coders!

## Outcome of this PR
- I created Github workflow with golangci-lint
- I fixed all the errors or turn off the violated rule

  In the end I decided only to turn off the `govet/composites` rule because I find it a bit annoying.

  The error handling - yeah, it's not fun, but I think it's worth catching as many errors as possible and not letting them escape. With the `common.PanicError()` helper it's not that bad.

I'm open to discussions! Also, we may want to squash all those fix commits, I just want to be able to revert a fix simply if requested.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,32 @@
+name: Lint
+
+# NOTE(mhayden): Restricting branches prevents jobs from being doubled since
+# a push to a pull request triggers two events.
+on:
+  pull_request:
+    branches:
+      - "*"
+  push:
+    branches:
+      - master
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Install golangci-lint
+        run: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.23.7
+
+      - name: Run golangci-lint
+        run: $(go env GOPATH)/bin/golangci-lint run
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,9 @@
+linters-settings:
+  govet:
+    disable:
+      - shadow # default value recommended by golangci
+      - composites
+
 issues:
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
   max-issues-per-linter: 0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,6 @@
+issues:
+  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  max-issues-per-linter: 0
+
+  # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
+  max-same-issues: 0
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -91,7 +91,10 @@ func main() {
 	jobAPI := jobqueue.New(logger, store)
 	weldrAPI := weldr.New(rpm, common.CurrentArch(), distribution, logger, store)
 
-	go jobAPI.Serve(jobListener)
+	go func() {
+		err := jobAPI.Serve(jobListener)
+		common.PanicOnError(err)
+	}()
 
 	// Optionally run RCM API as well as Weldr API
 	if rcmApiListeners, exists := listeners["osbuild-rcm.socket"]; exists {
@@ -124,9 +127,16 @@ func main() {
 			}
 
 			listener := tls.NewListener(listener, tlsConfig)
-			go jobAPI.Serve(listener)
+			go func() {
+				err := jobAPI.Serve(listener)
+				common.PanicOnError(err)
+			}()
 		}
 	}
 
-	weldrAPI.Serve(weldrListener)
+	go func() {
+		err := weldrAPI.Serve(weldrListener)
+		common.PanicOnError(err)
+	}()
+
 }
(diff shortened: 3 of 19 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Install golangci-lint
+        run: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.23.7
```
teg: Is there a way to install this using a package manager?

ondrejbudai: I think it isn't. However, is it a better option? `curl` into pipe is a bit ugly and can be dangerous but not on a CI environment which just gets deleted after finished. Additionally, versioning of the lint tool is much easier (you know the story about Fedora and keeping just two versions of packages in the repos). I think I don't mind the current installation method.

There's option to use a pre-built container from dockerhub. This might a bit better but I kinda disliked the idea of downloading a whole container to do linting.

teg: I think the most important thing is to pin the version, which this seems to do, so I'm convinced.

PR #264, opened 2020-03-02
spec: move worker-related service units to worker sub-package

ondrejbudai:
Prior this commit installing the worker sub-packages shows the following
warning:

Failed to preset unit: Unit file osbuild-worker@.service does not exist.

Moving the unit file to the sub-package fixes it.


msehnout:
Just curious: do we want to have the ability to install worker separately?

ondrejbudai:
Yes, in the case of remote workers it doesn't make much sense to install the whole osbuild-composer on the target machine.

PR #268, opened 2020-03-02
Minor cleanup in distro package

larskarlitski:
See individual commits. Tried to follow golang conventions a bit more: renaming a function, returning errors instead of panicking in a library, etc.

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -70,15 +70,18 @@ type Registry struct {
 	distros map[common.Distribution]Distro
 }
 
-func WithSingleDistro(dist Distro) *Registry {
+func NewRegistry(distros ...Distro) *Registry {
 	reg := &Registry{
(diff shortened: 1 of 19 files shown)


internal/distro/distro.go
```diff
@@ -74,51 +74,45 @@ func NewRegistry(distros ...Distro) *Registry {
...
-		panic("a distro with this name already exists: " + distro.Name())
-	}
-	r.distros[distroTag] = distro
+	return NewRegistry([]Distro{
+		f30,
+		f31,
+		f32,
+		el81,
+		el82,
+	}...,)
```
dvdhrm: Is that syntax intentional? Why not:

```go
return NewRegistry(f30, f31, f32, el81, el82)
```

larskarlitski: Oh dear. There's no excuse I'm afraid.

Thanks for catching that. Fixed it.

PR #269, opened 2020-03-03
distribution: set CACHE_DIRECTORY in the .service file

larskarlitski:
systemd >= 240 sets this variable to `/var/cache/` + the value of
CacheDirectory. osbuild-composer must run on earlier versions though
(specifically RHEL 8.2).

### DIFF
--- a/distribution/osbuild-composer.service
+++ b/distribution/osbuild-composer.service
@@ -13,5 +13,8 @@ WorkingDirectory=/usr/libexec/osbuild-composer/
 User=_osbuild-composer
 Restart=on-failure
 
+# systemd >= 240 sets this, but osbuild-composer runs on earlier versions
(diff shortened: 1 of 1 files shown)


distribution/osbuild-composer.service
```diff
@@ -13,5 +13,8 @@ WorkingDirectory=/usr/libexec/osbuild-composer/
 User=_osbuild-composer
 Restart=on-failure
 
+# systemd >= 240 sets this, but osbuild-composer runs on earlier versions
+Environment="CACHE_DIRECTORY=/var/cache/osbuild-composer"
```
msehnout: Is it ok to leave both this and the `CacheDirectory` directive?

larskarlitski: Yes. systemd overrides `CACHE_DIRECTORY` when `CacheDirectory=` is set.

PR #270, opened 2020-03-03
Fix blueprint commit ordering and add blueprint tag support

bcl:
In the process of adding tests I discovered that the tag route was unimplemented, and that the blueprint commits were un-ordered. This PR builds on my test PR #261 and should not be merged until that one is.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -400,6 +400,44 @@ func (s *Store) DeleteBlueprintFromWorkspace(name string) {
 	})
 }
 
+// TagBlueprint will tag the most recent commit
+// It will return an error if the blueprint doesn't exist
+func (s *Store) TagBlueprint(name string) error {
+	return s.change(func() error {
+		_, ok := s.Blueprints[name]
+		if !ok {
+			return errors.New("Unknown blueprint")
+		}
+
+		// Get the latest revision for this blueprint, and the most recent commit
+		var revision int
+		var timestamp string
+		var commit blueprint.Change
+		for _, c := range s.BlueprintsChanges[name] {
+			if c.Revision != nil && *c.Revision > revision {
+				revision = *c.Revision
+			}
+			if c.Timestamp > timestamp {
+				timestamp = c.Timestamp
+				commit = c
+			}
+		}
+		if len(timestamp) == 0 {
+			return errors.New("No commits for blueprint")
+		}
+
+		// The most recent commit already has a revision, don't bump it
+		if commit.Revision != nil {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


internal/store/store.go
```diff
@@ -400,6 +400,44 @@ func (s *Store) DeleteBlueprintFromWorkspace(name string) {
...
+
+		// Get the latest revision for this blueprint, and the most recent commit
+		var revision int
+		var timestamp string
+		var commit blueprint.Change
+		for _, c := range s.BlueprintsChanges[name] {
+			if c.Revision != nil && *c.Revision > revision {
+				revision = *c.Revision
+			}
+			if c.Timestamp > timestamp {
```
teg: As mentioned in a later commit, timestamps may not be unique. Would it make sense to order by version as a secondary key in case the timestamps match?

internal/store/store.go
```diff
@@ -193,6 +195,49 @@ func New(stateDir *string, distroArg distro.Distro, distroRegistryArg distro.Reg
 	if s.BlueprintsChanges == nil {
 		s.BlueprintsChanges = make(map[string]map[string]blueprint.Change)
 	}
+	if s.BlueprintsCommits == nil {
+		s.BlueprintsCommits = make(map[string][]string)
+	}
+
+	// Populate BlueprintsCommits for existing blueprints without commit history
```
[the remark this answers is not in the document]
bcl: We can't add anything to blueprint.Change since it is also used for the API response.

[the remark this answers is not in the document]
teg: Make it a new type and translate?

internal/store/store.go
bcl: That would probably work well enough. It would only happen if the timestamps were equal, which may results in unexpected ordering already. It would only fail if the store was expecting a lower version in the same second, like a push followed by an undo.

teg: Yeah, not ideal, but until we have a stable release maybe acceptable?

teg: Thinking more about this, I would leave the sorting by version out. It is possibly improving an edge case, but not necessarily, and it can even make it worse. So adding it in here would be misleading.

bcl:
After pondering this a bit more I can't see any way to avoid this change. I don't think it needs a separate update utility since it is a new addition to the store, not a change in behavior for one of the existing fields -- that case is going to be much more difficult to handle.

If you don't populate the new commit map you then need to have two code paths when getting the commits, one of which isn't really correct and which may return different results each time since iterating go maps isn't guaranteed to follow the same order. By forcing a conversion to the new system you at least have a consistent result.

PR #271, opened 2020-03-03
osbuild-pipeline: validate distro when fetching it

larskarlitski:
We were verifying two things: if the passed distroArg exists in the
distribution mapping in common/types.go and if the it is an actually
registered distro. Since you cannot have distros registered that don't
correspond to a type, the first test is unnecessary.

Merge the two tests by moving the (much better) error message down into
the second test. This makes DistributionExists redundant, because
Registry.GetDistro() checks this implicitly.

Also, move ListDistributions() to the Registry object, because we want
to show distributions that are actually registered.

Add a test which checks that Registry.List() works and that all included
distributions register correctly.

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -41,15 +41,6 @@ func main() {
 		return
 	}
 
-	// Validate distribution
-	if !common.DistributionExists(distroArg) {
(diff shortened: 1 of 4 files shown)


larskarlitski:
Rebased to include @ondrejbudai's linter fixes. Thanks!

PR #273, opened 2020-03-03
📦 Add RPM builds via github actions 

major:
Unify the github actions workflows under `tests.yml` and add an RPM build
job to match the one for osbuild.


larskarlitski:
The tests are failing because `cmd/osbuild-weldr-tests/main.go` was missing in https://github.com/osbuild/osbuild-composer/pull/261.

The fix is on https://github.com/osbuild/osbuild-composer/pull/274. Please rebase once that lands.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,96 @@
...
+  pull_request:
+    branches:
+      - "*"
+  push:
+    branches:
+      - master
+
+jobs:
+  lint:
+    name: Lint
```
larskarlitski: Is there a benefit from having it in one file or just a matter of style, because we have the other workflows in her as well?

larskarlitski:
> The fix is on #274. Please rebase once that lands.

Landed.

.github/workflows/tests.yml
major: We can make the workflows dependent on each other. There's no need to build RPMs if unit tests or linting fails. ;)

major:
@larskarlitski Rebased and Github Actions are green. ;)

PR #280, opened 2020-03-04
Makefile: simplify `make rpm`

larskarlitski:
Prior to this patch, `make rpm` would produce rpms that have the latest
tag as their versions. This was confusing, because one could never know
which contents are in a locally built rpm.

Change this so that the is version always based on the commit hash of
HEAD. This is easy: the golang macros read a `%commit` macro when it
exists and do this for us.

To simplify more, only define `%_topdir` to ./rpmbuild and use
rpmbuild's known directory structure (SPEC, SOURCES, RPMS, ...)
otherwise, to make it easier to find build results.

Build the specfile, tarball, source rpms, and rpms with `make rpm`,
without separate sub-targets. We can reintroduce them if they're needed
somewhere.

Also remove the `check-working-directory` target. It should be clear
from the output that only the currently-committed files are included,
because the resulting tarball and rpms contain the commit hash. Without
the check, one can work on the Makefile without having to commit all the
time, for example ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,4 @@ __pycache__
 /*.tar.gz
 /*.src.rpm
 
-/build
 /rpmbuild
-/output
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKAGE_NAME = osbuild-composer
-VERSION = $(shell grep Version golang-github-osbuild-composer.spec | awk '{gsub(/[^0-9]/,"")}1')
+COMMIT=$(shell git rev-parse HEAD)
 
 .PHONY: build
 build:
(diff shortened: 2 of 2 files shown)


msehnout:
Please don't remove Makefile targets without prior notice. I use them. This commit broke my workflow because the checkout directory lives on NFS drive and I cannot execute `make rpm` there. That's why I use `make srpm && rpmbuild --rebuild <srpm>` because the rebuild takes place in `~/rpmbuild` which is not NFS drive.

I also consider removing the `check-working-directory` an unfortunate choice because it prevents me from errors when testing new features. I don't think it was necessary to remove it.

teg:
I think we should have some common understanding of what kind of make targets we have and why. Splitting `make rpm` back into two seems reasonable to me. But most of the other make targets I'd like to drop (exactly because we don't know how and why they are used, so updating them becomes cumbersome, at least the usecases should be documented). My take is that private helpers should not live in the repo, but we should write this up and get consensus on what we want.

ondrejbudai:
I am for bringing back the `srpm` target as well. All out other repos have it as well (lorax, cockpit-composer) and it makes sense because sometimes you just need a srpm. Especially, when building go packages takes a while.

msehnout:
> I think we should have some common understanding of what kind of make targets we have and why. Splitting `make rpm` back into two seems reasonable to me. But most of the other make targets I'd like to drop (exactly because we don't know how and why they are used, so updating them becomes cumbersome, at least the usecases should be documented). My take is that private helpers should not live in the repo, but we should write this up and get consensus on what we want.

That's what I tried to address with the common development environment, because these issues are a result of the fact that each of us have a completely different workflow. In general osbuild-composer, when compared to other open source software projects, is very unfriendly to get started with. The development process in cumbersome because we remove all private/convenience make targets, helper scripts etc. 

I agree that we need to keep the repo and code organized and, if possible, small but we shouldn't sacrifice pleasant development for few lines of code because unpleasant and cumbersome development process will result in less/worse contributions.

larskarlitski:
@msehnout Sorry for breaking your workflow. I should have requested a review from everyone.

I'll add the `srpm` target back today.

> That's what I tried to address with the common development environment, because these issues are a result of the fact that each of us have a completely different workflow.

We will always have different workflows. For example, the `check-working-directory` target was breaking mine. Is there a different way we can support preventing you "from errors when testing new features."?

> In general osbuild-composer, when compared to other open source software projects, is very unfriendly to get started with. The development process in cumbersome because we remove all private/convenience make targets, helper scripts etc.

Which convenience targets have we removed? I still see `make build`, `make install` (which is questionable, because it might get out of sync with the package), `make ca` and friends, which are purely convenience. Also, which helper scripts do you mean?

It is very hard to respond to such a general comment. What exactly do you think is cumbersome?

msehnout:
> We will always have different workflows. For example, the check-working-directory target was breaking mine. Is there a different way we can support preventing you "from errors when testing new features."?

I thought we agreed on this one pretty long time ago. More specifically it was introduced to `osbuild` on Jul 25, 2019.  I introduced the same so that I can keep the same workflow for both.

Yes, there is a different way, I think we should agree on a development process, write it down to `CONTRIBUTING.md` and maintain it. Nobody is forced to use it, but it would be the single "official" development workflow.

> Which convenience targets have we removed? I still see make build, make install (which is questionable, because it might get out of sync with the package), make ca and friends, which are purely convenience. Also, which helper scripts do you mean?

hmm, I didn't even know we have these because we agreed it would get out of sync with the package. I probably confused `osbuild-composer` with `osbuild`.

> Also, which helper scripts do you mean?

For example `make run-socket`, here:
```
$ git log -p Makefile | tail -3
+
+run-socket:
+	systemd-socket-activate -l /run/weldr/api.socket -l /run/osbuild-composer/job.socket ./osbuild-composer
```

> It is very hard to respond to such a general comment. What exactly do you think is cumbersome?

I always think about the workflow this way: If I was completely new to this project what would I need to do to build, run, and debug this project? 

In case of osbuild-composer I always did checkout => rpm build (because I didn't know we still have make install) => systemctl start, then you need to attach gdb. In my opinion, this workflow is cumbersome. Ideal workflow (again mine, not necessarily for others) is to open an IDE, such as GoLand or VSCode, set up a target like `go run osbuild-composer` or `go run osbuild-composer-dev` * and click the green button (run or debug).

\* I proposed to create alternative to osbuild-composer executable which wouldn't use the systemd socket activation so that I could easily use graphical debugger.

larskarlitski:
> I'll add the `srpm` target back today.

Opened a PR: https://github.com/osbuild/osbuild-composer/pull/283

> Yes, there is a different way, I think we should agree on a development process, write it down to `CONTRIBUTING.md` and maintain it. Nobody is forced to use it, but it would be the single "official" development workflow.

Sorry, I meant if there's a different way to support both your and my workflow. They seem to be in conflict with each other.

> For example `make run-socket`, here:

Its removal has a good explanation: https://github.com/osbuild/osbuild-composer/commit/029dd0a81bdc405225b6e00b55e74a74eabbd42d. `systemd-socket-activate` isn't enough when using more advanced systemd features, like setting up `/var/lib`, `/var/cache`, and temporary directories. It's really not beginner-friendly to keep a target that doesn't work.

> I always think about the workflow this way: If I was completely new to this project what would I need to do to build, run, and debug this project? [...]

Right, I'm happy to improve on the workflow, especially via documentation, but also through things like Tom's test-vm effort, which aims to streamline this.

I don't see the point in reimplementing systemd features only for development, though. There are probably other solutions that are less error-prone (running it in a container?).

PR #281, opened 2020-03-05
spec: remove lorax-composer provide

larskarlitski:
We're not fully compatible with lorax-composer's API yet, and we don't
provide lorax-composer's systemd unit files.

As a result, cockpit-composer's integration tests fail, because `dnf
install lorax-composer` on Fedora can result in installing
osbuild-composer in some cases.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -41,7 +41,6 @@ Requires: systemd
 Requires: osbuild >= 7
 
 Provides: osbuild-composer
-Provides: lorax-composer
 
 %description
 %{common_description}


larskarlitski:
> I will end up merging this when it turns green. So lets make sure I don't.

Thanks. I just talked to @marusak and he explained the situations (thanks!). The problem is not that `dnf install lorax-composer` pulls in `osbuild-composer`, but that `cockpit-composer` requiring it does. The fix remains the same.

ondrejbudai:
In the meantime I found this article:
https://fedoraproject.org/wiki/Packaging:WeakDependencies#Real_life_example

I'm not sure but it might be possible to fix this issue by adding `Suggests: lorax-composer` to `cockpit-composer`. Then, `lorax-composer` should have a higher priority. However, I'm not sure if this will work correctly, because one provider (`lorax-composer` ) is the real provider and one provider (`osbuild-composer`) is just a virtual one, so dnf might treat this case differently.

ondrejbudai:
OK, I think that for a proper fix we need to do something like this:

lorax-composer.spec
```
Provides: weldr
```

osbuild-composer.spec
```
Provides: weldr
```

cockpit-composer.spec
```
Requires: weldr
Suggests: lorax-composer
```

Another solution might be this one:

cockpit-composer.spec
```
Requires: (golang-github-osbuild-composer OR lorax-composer)
Suggests: lorax-composer
```

In both cases `Provides: lorax-composer` must be removed from `osbuild-composer.spec`.

PR #282, opened 2020-03-05
[WIP] POC: Rewrite dnf-json integration tests for go test

atodorov:
Rewrites the file so that it can be discovered by `go test` and so we can use asserts as well. This makes it more common to what is there in the golang eco system wrt writing test suites.

We can still produce a binary that can be executed to run the test. We can also discover/execute tests on the fly via go test (currently fails b/c it can't find the dnf-json command). 

I've git a snag when building the binary for inclusion in RPM though:
- There is a %gotest macro for RPM but it doesn't accept any options so I can't pass (-c (compile, don't execute) and -o)
- With the current workaround rpmbuild failed with 
```
error: Missing build-id in /root/osbuild-composer/rpmbuild/BUILDROOT/golang-github-osbuild-composer-6-1.20200305git46ea9cc.fc31.x86_64/usr/libexec/tests/osbuild-composer/osbuild-dnf-json-tests
```

@teg, @msehnout can you please take a look.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build:
 	go build -o osbuild-upload-aws ./cmd/osbuild-upload-aws/
 	go build -o osbuild-tests ./cmd/osbuild-tests/
 	go build -o osbuild-weldr-tests ./cmd/osbuild-weldr-tests/
-	go build -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/
+	go test -c -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/main_test.go
 	go build -o osbuild-rcm-tests ./cmd/osbuild-rcm-tests/
 
 .PHONY: install
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 // This package contains tests related to dnf-json and rpmmd package.
 package main
 
@@ -5,26 +7,11 @@ import (
 	"fmt"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
+	"testing"
 	"path"
 )
-
-func main() {
-	// Tests that the package wrapping dnf-json works as expected
-	dir, err := setUpTemporaryRepository()
-	defer func(dir string) {
-		err := tearDownTemporaryRepository(dir)
-		if err != nil {
-			log.Print("Warning: failed to clean up temporary repository.")
-		}
-	}(dir)
-	if err != nil {
-		log.Panic("Failed to set up temporary repository:", err)
-	}
-	TestFetchChecksum(false, dir)
-}
 
 func setUpTemporaryRepository() (string, error) {
 	dir, err := ioutil.TempDir("/tmp", "osbuild-composer-test-")
@@ -47,25 +34,30 @@ func tearDownTemporaryRepository(dir string) error {
 	return os.RemoveAll(dir)
 }
 
-func TestFetchChecksum(quiet bool, dir string) {
+func TestFetchChecksum(t *testing.T) {
+	dir, err := setUpTemporaryRepository()
+	defer func(dir string) {
+		err := tearDownTemporaryRepository(dir)
+		if err != nil {
+			t.Errorf("Warning: failed to clean up temporary repository.")
+		}
+	}(dir)
+	if err != nil {
+		t.Errorf("Failed to set up temporary repository: %s", err)
+	}
+
 	repoCfg := rpmmd.RepoConfig{
 		Id:        "repo",
 		Name:      "repo",
 		BaseURL:   fmt.Sprintf("file://%s", dir),
 		IgnoreSSL: true,
 	}
-	if !quiet {
-		log.Println("Running TestFetchChecksum on:", dir)
-	}
 	rpmMetadata := rpmmd.NewRPMMD(path.Join(dir, "rpmmd"))
 	_, c, err := rpmMetadata.FetchMetadata([]rpmmd.RepoConfig{repoCfg}, "platform:f31")
 	if err != nil {
-		log.Panic("Failed to fetch checksum:", err)
+		t.Errorf("Failed to fetch checksum: %s", err)
 	}
 	if c["repo"] == "" {
-		log.Panic("The checksum is empty")
-	}
-	if !quiet {
-		log.Println("TestFetchChecksum: SUCCESS")
+		t.Errorf("The checksum is empty")
 	}
 }
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ __pycache__
 
 /osbuild-composer
 /osbuild-worker
+/osbuild-weldr-tests
 /osbuild-pipeline
 /osbuild-upload-azure
 /osbuild-upload-aws
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/julienschmidt/httprouter v1.2.0
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/stretchr/testify v1.4.0 // indirect
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/net v0.0.0-20200202094626-16171245cfb2 // indirect
 	golang.org/x/sys v0.0.0-20200212091648-12a6c2dcc1e4 // indirect
 )
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -1,4 +1,6 @@
 %global goipath         github.com/osbuild/osbuild-composer
+# workaround for https://pagure.io/go-rpm-macros/pull-request/23
+%define gotest_co(c:o:) GO111MODULE=off go test %{gotestflags} -ldflags "${LDFLAGS:-}%{?currentgoldflags} -extldflags '%{gotestextldflags}'" %{?**};
 
 Version:        7
 
@@ -68,7 +70,7 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 %gobuild -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
 %gobuild -o _bin/osbuild-weldr-tests %{goipath}/cmd/osbuild-weldr-tests
-%gobuild -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
+%gotest_co -c -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
 %gobuild -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
 
 %install


cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return "", err
+	}
+	return dir, nil
+}
+
+func TestFetchChecksum(t *testing.T) {
+	defer filet.CleanUp(t)
+	
+	dir := filet.TmpDir(t, "/tmp")
```
msehnout: What was the reason to use this go module?

Makefile
```diff
@@ -10,7 +10,7 @@ build:
 	go build -o osbuild-upload-aws ./cmd/osbuild-upload-aws/
 	go build -o osbuild-tests ./cmd/osbuild-tests/
 	go build -o osbuild-weldr-tests ./cmd/osbuild-weldr-tests/
-	go build -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/
+	go test -c -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/
```
msehnout: Why to change only osbuild-dnf-json-tests to use something different then the other testing executables in `/cmd`?

ondrejbudai:
I'm not sure if this is something we want. `go test` is great but I think it breaks our current integration testing strategy - to have test binaries in the tests rpm subpackage. The subpackage is then installed on a test machine and all the binaries are run.

cmd/osbuild-dnf-json-tests/main_test.go
atodorov: I'm not really sure. I was seeing some errors initially which may not have been related to the temporary directory at all. I will revert this change and try again.

Makefile
atodorov: With this change `osbuild-dnf-json-tests` is now relying on Go's testing module (notice there is no main function). 

That's POC b/c this is the smallest file. If the approach is well received all of the other executables should be converted in this way as well.

atodorov:
> I'm not sure if this is something we want. go test is great but I think it breaks our current integration testing strategy - to have test binaries in the tests rpm subpackage. The subpackage is then installed on a test machine and all the binaries are run.

It doesn't break that strategy per se. It is still possible to do that (minus the part where I can't get rpm-build to cooperate and ATM I'm not sure how much effort it would take for that).

teg:
I'm not sure I fully understand the reasoning here. So let me just throw out some comments in no particular order.

 - I think using some internal go testing framework makes sense for unit tests (and we do).
 - I think integration test binary should be compiled and shipped just like any other binary, and the fact that it is go-code internally should not matter to the caller. A priori, I don't think unit test framework is helpful for integration tests, but I may very well be wrong on this.
 - Integration tests may depend on a correctly set-up host system, so running them as part of `go test ./...` would be unexpected, and quite possibly unhelpful (could have test failures that are just caused by a miss-configured system). If the tests could be run as part of the unit-tests, then we should do that unconditionally and not ship them in the test RPM at all.
 - I have not yet any strong opinion about whether asserts or anything else makes the most sense, my understanding was that QE would like some machine-readable output of the test suite, so I think that should be our main focus, whether assert helpers help this or not I don't know.

What is the aim of this PR?

atodorov:
@teg - valid comments, I am providing more info below

> What is the aim of this PR?

To give the `go test` approach a try and figure out the following:

* what does it take to rewrite existing testing files in the way go test expects them to be structured - definitely doable, it took me couple of hours without prior experience to produce this PR
* can we compile them as binares and execute them independently outside of the `go test` workflow - YES
* Find out if this will be causing any problems - ATM issues with building the RPM described above. Unknown whether it can be resolved easily.
* Figure out how everyone else feels about the subject - the discussion we're having now.


>    I think using some internal go testing framework makes sense for unit tests (and we do).

This isn't in conflict with the proposed changes. For unit tests we may or may not built the internal framework on top of popular Go frameworks/modules (e.g. testing or testify). I have not yet looked into the unit tests to have a more informed opinion, my main concern ATM is the integration test suite.

>    I think integration test binary should be compiled and shipped just like any other binary, and the fact that it is go-code internally should not matter to the caller. 

This is not with conflict with the proposed changes. As seen in this PR you can have the binaries produced by `go test` laying around if you wish. It is also possible to compile these binaries without executing them during the process.  This is how I've been executing them (no change to existing process).

> A priori, I don't think unit test framework is helpful for integration tests, but I may very well be wrong on this.

TBH I've not seen a right or wrong way here but I have seen preferred way (e.g. generally considered best practice).

In all of the other languages I've worked with testing frameworks are generally built around the underlying xUnit framework. That gives you the test discovery, test runner, results reporter, setup/teardown hooks and asserts mechanisms out of the box. The ones I've see also support test runner plugins e.g. you can hook into the entire process and extract information out of it. In the case of Python pretty much everything builds on top of Python's unittest module. 

The fact that a framework was designed for unit tests doesn't prevent it from being used for higher level testing. You still have the need for setup/teardown, discovery (or specify) which test modules/methods should be executed, actually executing them and consuming the results. A prime example of this is https://github.com/weldr/lorax/blob/master/test/composertest.py - builds on top of unittest but spins up VMs from the get go and uses ssh to actually execute the testing code


>    Integration tests may depend on a correctly set-up host system, so running them as part of go test ./... would be unexpected, and quite possibly unhelpful (could have test failures that are just caused by a miss-configured system).

I don't question this, fully agree. In this PR I am not advocating we start running integration tests as part of `go test`, only using `go test` to compile the testing binaries themselves.

> If the tests could be run as part of the unit-tests, then we should do that unconditionally and not ship them in the test RPM at all.

I have not investigated this yet. I'm pretty sure we can do this but probably isn't feasible. Can't say how much effort will be to have the CI system appropriately configured. OTOH having the -test RPM will be very useful for internal downstream qualification. The other option (which QE uses now) is to git clone the repository and kick-off the integration test suite in that way.

My preference would be to still have the -test RPM containing the testing binaries.

>    I have not yet any strong opinion about whether asserts or anything else makes the most sense, 

From what I've read in the Go eco-system using or not using asserts is an opinionated practice. Granted the only thing you need is if condition + exit/return code everywhere else I've seen asserts are integral part of any testing code, also assert methods are usually part of the test framework itself. In Go and Node.js for example use external modules for assertions.

The question for me is why aren't we using something that is readily available instead of DYI ? 

> my understanding was that QE would like some machine-readable output of the test suite, so I think that should be our main focus, whether assert helpers help this or not I don't know.

This is a separate topic and IMO not of such high priority (I personally would classify it as medium or nice to have).  

From what I can see the output of `go test` is machine readable. I can't find lots of info but the format spec is shown here https://golang.org/pkg/cmd/go/internal/test/. Looks to me a lot like what is called TAP (test anything protocol).

You also have tools like go-junit-report which can take that and convert into a more standard format which can be consume by other tooling down the line. Quick search reveals other tools which can augment this output (I have no idea what their level of quality & usability is though).

Again I will ask the same question here: why not use something that is readily available instead of DYI ?

larskarlitski:
As I understand it, the aim of these patches is to use `go test` for integration tests in addition to unit tests. This makes sense to me, because it makes these two kinds of tests more consistent.

However, I think integration tests should be run entirely separate from the unit tests, because having two ways of running them (from source with `go test ./...` and when they're installed via the `-tests` package) adds unnecessary complexity. Excluding the integration tests from `go test ./...` can be solved by build constraints (`+build integration` seems to be a common way to do this).

That means the main benefit of this the use of `testing.T` and using the existing naming conventions for test functions. I think it's still worth doing just for that.

We haven't used an assert library in any of the other tests. Please leave that out of this patch to keep consistency with the unit tests. I'm happy to talk about that independently, but moving all tests over seems to like too much trouble for the benefit.

atodorov:
Updated & rebased. Changes related to asserts have been removed. 

I had redefined the %gotest macro so I can pass -c -o during RPM build time. When I try the resulting command by hand it appears to resolve the original problem with missing build-id from the binary files. However now I get a different error from rpmbuild:
```
+ GO111MODULE=off
+ go test -buildmode pie -compiler gc -ldflags '-X github.com/osbuild/osbuild-composer/version.commit=652f3e10827e3d339685b81a9576fa56acf7e667 -X github.com/osbuild/osbuild-composer/version=7 -extldflags '\''-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '\''' -c -o _bin/osbuild-dnf-json-tests github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests
can't load package: package github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests: cannot find package "github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests" in any of:
	/usr/lib/golang/src/github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests (from $GOROOT)
	/root/go/src/github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests (from $GOPATH)
error: Bad exit status from /var/tmp/rpm-tmp.zO3U1r (%build)
```

atodorov:
> +build integration seems to be a common way to do this).

@larskarlitski that is supposed to be added somewhere but I'm not sure where. Can you elaborate ?

ondrejbudai:
> Updated & rebased. Changes related to asserts have been removed.
> 
> I had redefined the %gotest macro so I can pass -c -o during RPM build time. When I try the resulting command by hand it appears to resolve the original problem with missing build-id from the binary files. However now I get a different error from rpmbuild:
> 
> ```
> + GO111MODULE=off
> + go test -buildmode pie -compiler gc -ldflags '-X github.com/osbuild/osbuild-composer/version.commit=652f3e10827e3d339685b81a9576fa56acf7e667 -X github.com/osbuild/osbuild-composer/version=7 -extldflags '\''-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '\''' -c -o _bin/osbuild-dnf-json-tests github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests
> can't load package: package github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests: cannot find package "github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests" in any of:
> 	/usr/lib/golang/src/github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests (from $GOROOT)
> 	/root/go/src/github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests (from $GOPATH)
> error: Bad exit status from /var/tmp/rpm-tmp.zO3U1r (%build)
> ```

I think `%gotest` doesn't have the right paths set when used in `%build` stage.

larskarlitski:
> Updated & rebased. Changes related to asserts have been removed.

Thanks!

> @larskarlitski that is supposed to be added somewhere but I'm not sure where. Can you elaborate?

Yes. At the top of the source file, above the `package` directive but below the documentation for the package. It's explained in the [documentation for the build package](https://golang.org/pkg/go/build/#hdr-Build_Constraints).

larskarlitski:
I've filed https://github.com/osbuild/osbuild-composer/pull/301 and changed it back use `go test -c` directly. As [mentioned by nim](https://pagure.io/go-rpm-macros/pull-request/23#) upstream, the `%gotest` macro is really meant to be used in a `%check` section.

Please see the other PR for additional changes.

bcl:
I think I have a better idea of where this is going now, thanks. If we can exclude these tests from a normal go test ./... run then that's a big step in the right direction, but I think there is another issue. As far as I can see there is no way to pass information from the framework setup into the test functions. eg. If you look at weldrcheck/check.go I am passing in a struct with the http.Client, with the testing framework you have to put things like that into globals, which is fine for simple things but integration tests tend to get more complex.

larskarlitski:
> As far as I can see there is no way to pass information from the framework setup into the test functions. eg. If you look at weldrcheck/check.go I am passing in a struct with the http.Client, with the testing framework you have to put things like that into globals, which is fine for simple things but integration tests tend to get more complex.

We should strive towards sharing no state, but creating it fresh for every test. This is simple for the client example, because it's cheap to construct and can be factored into a helper function.

When things get more complex, i.e., things that take a long time to set up, I'd be fine with just using a global variable. Not perfect, but I think the benefits that `go test` brings are worth it.

teg:
Can this be closed in favour of #301 ?

atodorov:
> Can this be closed in favour of #301 ?

yes.

PR #286, opened 2020-03-05
🐣 Use new 'make rpm' in GitHub Actions

major:
Now that `make rpm` does what we need, let's simplify GitHub Actions workflows and use that instead. 

Also, let's add emojis. *Why?* Don't ask why. **Just love the emojis.** 🥰

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   lint:
-    name: Lint
+    name: "🛃 Lint"
     runs-on: ubuntu-latest
     steps:
 
@@ -32,9 +32,7 @@ jobs:
         run: $(go env GOPATH)/bin/golangci-lint run
 
   rpm_build:
-    name: "RPM build"
-    needs:
-      - lint
+    name: "📦 RPM"
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -32,9 +32,9 @@ jobs:
         run: $(go env GOPATH)/bin/golangci-lint run
 
   rpm_build:
-    name: "RPM build"
-    needs:
-      - lint
+    name: "📦 RPM"
+    # needs:
+    #   - lint
```
larskarlitski: Is this commented out by mistake?

.github/workflows/tests.yml
```diff
@@ -43,54 +43,33 @@ jobs:
...
-          rpmbuild -bb \
-            --define "_sourcedir `pwd`/sources" \
-            --define "_specdir `pwd`/spec" \
-            --define "_builddir `pwd`/rpmbuild" \
-            --define "_srcrpmdir `pwd`/srpms" \
-            --define "_rpmdir `pwd`/rpms" \
-            --define "_buildrootdir `pwd`/build" \
-            spec/golang-github-osbuild-composer.spec
+          mkdir rpms
+          cp -av rpmbuild/RPMS/*/*.rpm rpms/
```
larskarlitski: Please also copy the `rpmbuild/SRPMS/*.src.rpm` to get the srpm as an artifact as well.

major: Oh, good idea.

.github/workflows/tests.yml
major: Yeah, I plan to just remove it so the RPM builds will finish sooner.

PR #289, opened 2020-03-06
osbuild-tests: run compose tests in a temporary directory

larskarlitski:
Some `composer-cli` commands operate on the current directory, for
example saving blueprints or images. Add the `TemporaryWorkDir` type,
which helps changing to a temporary working directory and cleaning
everything up after.

The alternative would have been to pass a working directory to all calls
of `runComposerCLI`, but that seemed like it would require a lot of
change, which I didn't deem worth for testing code.

### DIFF
--- a/cmd/osbuild-tests/main.go
+++ b/cmd/osbuild-tests/main.go
@@ -82,6 +82,11 @@ func testCompose(outputType string) {
 	pushBlueprint(&bp)
 	defer deleteBlueprint(&bp)
 
+	runComposerCLI(false, "blueprints", "save", "empty")
+	if _, err := os.Stat("empty.toml"); os.IsNotExist(err) {
+		log.Fatalf("empty.toml doesn not exist: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


larskarlitski:
Triggered by this PR: https://github.com/osbuild/osbuild-composer/pull/256

cmd/osbuild-tests/main.go
```diff
@@ -71,13 +71,22 @@ func main() {
...
+	defer tmpdir.Close()
+
 	bp := blueprint.Blueprint{
 		Name:        "empty",
 		Description: "Test empty blueprint in toml format",
 	}
 	pushBlueprint(&bp)
 	defer deleteBlueprint(&bp)
 
+	runComposerCLI(false, "blueprints", "save", "empty")
```
teg: Hm, this hunk is unrelated to this patch, no?

larskarlitski: Indeed. I've split it into a separate commit.

PR #292, opened 2020-03-06
image-types: start describing image types

larskarlitski:
This is the start to document the image types that osbuild-composer
produces on a higher level than the code in the `distro` package.

The idea is to document the purpose for each of the image types as well
as the reasons why deviating from a standard install is necessary.

This is a draft. Want to get feedback if this is generally going in the right direction before fleshing it out further. That's why I requested reviews from *all the people* :tada:

### DIFF
--- a/image-types/README.md
+++ b/image-types/README.md
@@ -1,0 +1,8 @@
+# OSBuild Composer Image Types
+
+This directory contains high-level descriptions of all image types that
+*osbuild-composer* generates, grouped by operating system.
+
+Each of them contains a short description of the purpose and intended use case
+of the image type, and any peculiarities that make this image type differ from
+a standard installation.
--- a/image-types/rhel8/README.md
+++ b/image-types/rhel8/README.md
@@ -1,0 +1,21 @@
+# Red Hat Enterprise Linux 8
+
+Owner: *None*
+
+*osbuild-composer* can create [Red Hat Enterprise Linux 8][rhel] images.
+
+## Architectures
+
+RHEL supports `x86_64`, `aarch64`, `ppc64le`, and `s390x`.
+
+## Partitioning and Booting
+
+The default file system is `xfs` on one bootable partition.
+
+## Packages
+
+By default, these images are created by installing the `@Core` group and the
+`kernel` package.
+
+
+[rhel]: https://access.redhat.com/products/red-hat-enterprise-linux
--- a/image-types/rhel8/amazon-ec2.md
+++ b/image-types/rhel8/amazon-ec2.md
@@ -1,0 +1,48 @@
+# Amazon EC2 Image
+
+Owner: *None*
+
+This image is meant to be used in [Amazon Elastic Compute Cloud (EC2)][ec2], a
+popular cloud computing platform. It has to follow Amazon’s
+[guidelines][guidelines] and [requirements][requirements] for shared AMIs.
+
+## File Format
+
+EC2 uses Amazon Machine Images (AMIs) internally, which can only be created
+inside EC2. An image in a standard format (ova, vmdk, vhd/x, or raw) must be
+imported from S3 storage. *osbuild-composer* generates this image type in the
+VHDX format, because it is fairly modern, widely used, and uses less space than
+RAW images.
+
+The default image size is 6 GB.
+
+## Architectures
+
+Limited to `x86_64` and `aarch64`, because those are available on EC2.
+
+## Partitioning and Booting
+
+The image specifies partitions with a MBR partition table, because that is
+required by EC2. It contains one bootable partition formatted with RHEL's
+default file system `xfs`.
+
+## Packages
+
+`cloud-init` is included in this image, because it is required by EC2.
+
+EC2 doesn't require any specialized firmware. Thus, in order to keep the
+resulting image size small, all `-firmware` packages are excluded. An exception
+is the `linux-firmware` package, which cannot be excluded because the `kernel`
+package depends on it.
+
+`dracut-config-rescue` is excluded from the image, because the boot loader
+entry it provides is broken.
+
+## Kernel Command Line
+
+`ro console=ttyS0,115200n8 console=tty0 net.ifnames=0 rd.blacklist=nouveau nvme_core.io_timeout=4294967295 crashkernel=auto`
+
+
+[ec2]: https://aws.amazon.com/ec2
+[guidelines]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/building-shared-amis.html
+[requirements]: https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html
--- a/image-types/rhel8/openstack.md
+++ b/image-types/rhel8/openstack.md
@@ -1,0 +1,31 @@
+# OpenStack Image
+
+Owner: *None*
+
+This image is meant to be run in an OpenStack environment. It has to conform to
+OpenStack’s [image requirements][requirements].
+
+## File Format
+
+This image is OpenStack's native file format is qcow2.
+
+The default image size is 2 GB.
+
+## Partitioning and Booting
+
+The generated images contain one bootable partition formatted with RHEL's
+default file system `xfs`.
+
+## Software
+
+`cloud-init` is included in this image, because it is required by OpenStack.
+
+`dracut-config-rescue` is excluded from the image, because the boot loader
+entry it provides is broken.
+
+## Kernel Command Line
+
+`ro net.ifnames=0`
+
+
+[requirements]: https://docs.openstack.org/image-guide/openstack-images.html


teg:
I love the general approach! How do you imagine the process for reviewing and maintaining and keeping these things up to date should be?

msehnout:
I like the general idea, because this "formal" definition of an image is something I miss a lot. But I have basically the same question as @teg : how to we keep this up-to-date?

What do you think about generating these from the source code? Could we agree on a specific comment format and write a script that would generate the document from the code. Especially the list of packages and kernel parameters could be generated.

Example (three slashes and a markdown format):
```go
	/// # OpenStack Image
	/// This image is meant to be run in an OpenStack environment. It has to conform to
	/// OpenStack’s [image requirements][requirements].
	/// 
	/// ## Architectures
	/// 
	/// Only `x86_64` images are generated.
	r.outputs["openstack"] = output{
		Name:     "disk.qcow2",
		MimeType: "application/x-qemu-disk",
	        /// ## Additional Software
	        /// 
	        /// The following packages are installed in addition to a base install (`@Core` and `kernel`):
		Packages: []string{
			"@Core",
			"chrony",
			"kernel",
			"selinux-policy-targeted",
			"spice-vdagent",
			"qemu-guest-agent",
			"xen-libs",
			"langpacks-en",
			"cloud-init",
			"libdrm",
		},
		ExcludedPackages: []string{
			"dracut-config-rescue",
		},
		KernelOptions: "ro biosdevname=0 net.ifnames=0",
		Bootable:      true,
		DefaultSize:   2 * GigaByte,
                // ## File Format
                // This image is OpenStack's native file format is qcow2.
		Assembler: func(uefi bool, size uint64) *osbuild.Assembler {
			return r.qemuAssembler("qcow2", "disk.qcow2", uefi, size)
		},
	}

```

image-types/README.md
```diff
@@ -0,0 +1,8 @@
+# OSBuild Composer Image Types
+
+This directory contains high-level descriptions of all image types that
+**osbuild-composer** generates, grouped by operating system.
```
jgiardino: Why would we group them by OS? And are the definitions different across the different operating systems? If the definition is mostly the same, then having one place where the definition is defined would help avoid duplication. And the definition can include what's unique per OS, which might make it easier to review and maintain the definitions.

image-types/rhel/amazon-ec2.md
```diff
@@ -0,0 +1,94 @@
...
+format, because it is fairly modern, widely used, and uses less space than RAW
+images.
+
+## Architectures
+
+Only `x86_64` images are generated.
+
+## Size
+
+The default image size is 6 GB.
```
jgiardino: Are there a maximum or minimum size? If there are output types that would, then the definition should either state what those values are or state that there is no max or min size (so that it's obvious it wasn't just forgotten)

Ideally, this information would be passed to the UI so that we can avoid image sizes that we know won't work.

image-types/README.md
msehnout: Yes, different operating systems can have different packages, kernel parameters, initramdisk modules etc. If I am not mistaken, that is the case in Azure. But otherwise it should be mostly the same.

msehnout:
> Or to go one step further, we can place a simple json distro definition file somewhere, which can contain more than just package-lists, but also other input we need for distros. Note that none of this needs to be ABI (unless we want to). So the format can be changed anytime.

We discussed something similar during the hackfest. I basically proposed to have the definition of translation between empty blueprint and osbuild manifest as a JSON file so it becomes easier for others to see how we define what e.g. "Fedora 30 is". But we decided not to have this declarative format, because we already have blueprints and manifests and we didn't want to introduce another intermediate configuration file.

Or do you propose to have only the "static data" like packages, kernel parameters, etc. in the JSON document and then we include it directly in the binary?

jgiardino:
@msehnout Your idea of auto-generating the definitions is really cool, but Lars and I have discussed a workflow for how these definitions are maintained independently of the source code, so that the definition is the source of truth against which the source code is implemented. To support this workflow, we would identify owners/stakeholders who understand what requirements the image types need to support and would own the definitions. 

So for example, if the formal definition is updated to include different requirements by the image owner, then after we review and commit the updated definition, we would then update the image type in the source code to match. Or alternatively, if the source code somehow gets out of sync, then it's easy enough to check the definition and know what it's supposed to be.

Along with adding the existing definitions, we would also want to include a description of this process for updating or introducing new image types.

larskarlitski:
Thanks for all the feedback! I've reworded things a bit and added an additional `README.md` in the rhel8 directory, which is meant to document the base install.

I hope I'm answering all questions below. Please let me know if I forgot something :)

@teg
> How do you imagine the process for reviewing and maintaining and keeping these things up to date should be?

@jgiardino and I have some ideas, but nothing concrete yet. It will probably revolve around having points-of-contact for each product and image type. I thought a good first step is to document what we currently have in this PR, to figure out the format that we want.

@msehnout 
> What do you think about generating these from the source code?

I'd like this to be a bit more high-level. More like a requirements document which states the reasons behind the decisions. I've made this more clear by not including full package lists, but saying things like "no -firmware packages". This (hopefully) makes it clear that we'd exclude the next firmware package that shows up as well.

Also, as @jgiardino mentioned, we'd like people who don't know the code to contribute to these documents. It will be harder for them to edit comments.

@dvdhrm 
> However, rather than auto-generating documentation, how about an authoritative package-list somewhere in the source tree, which is human-readable and sourced by our code to determine the packages to install (at compile-time should be sufficient).

I would like this a lot, if it is possible. As mentioned above, I think this is not in scope for this particular patch. Could you and @msehnout take this to a separate issue?

image-types/README.md
larskarlitski: There might be some overlap indeed, but in general they're different for each operating system. Most importantly, the reasoning might be. Also, different operating systems call their images different things (Fedora Cloud Base vs RHEL KVM Guest Image, for example).

These documents will also grow as new releases for these come out. I didn't want to make them too complex by conflating different releases and different operating systems in the same document.

image-types/rhel/amazon-ec2.md
larskarlitski: I've clarified a bit by saying that they can be larger when the packages don't fit. In general, we should treat it as: minimum size is the size of the installed packages (+ 10% or so) and there is no maximum size other than what's inherent in the file formats (which is ridiculously huge).

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,79 @@
...
+from S3 storage. **osbuild-composer** generates this image type in the VHDX
+format, because it is fairly modern, widely used, and uses less space than RAW
+images.
+
+The default image size is 6 GB, unless it has to be larger due to installed
+packages.
+
+## Architectures
+
+Only `x86_64` images are generated.
```
atodorov: I've verified that lorax-composer can also build arch64 AMIs and we can boot them in AWS. Brian made some patches in this area.

Do you mean that only x86_64 is supported b/c osbuild-composer is lacking multi-arch support or simply b/c this was not tested ?

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,79 @@
+# Amazon EC2 Image
+
+This image is meant to be used in [Amazon Elastic Compute Cloud (EC2)][ec2], a
+popular cloud computing platform. It has to follow Amazon’s
+[guidelines][guidelines] and [requirements][requirements] for shared AMIs.
+
+## File Format
+
+EC2 uses Amazon Machine Images (AMIs) internally, which only be created inside
```
atodorov: Small typo:
`which only be created`
`can` ^^^

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,79 @@
...
+
+The image specifies partitions with a MBR partition table, because that is
+required by EC2. It contains one bootable partition formatted with RHEL's
+default file system XFS.
+
+## Software
+
+`cloud-init` is included in this image, because it is required by EC2.
+
+These packages are also included, for unknown reasons:
```
atodorov: If reasons are unknown how safe it is to remove these packages?   Or are they included due to depsolving ?

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,79 @@
...
+
+EC2 doesn't require any specialized firmware. Thus, in order to keep the
+resulting image size small, all `-firmware` packages are excluded. An exception
+is the `linux-firmware` package, which cannot be excluded because the `kernel`
+package depends on it.
+
+`dracut-config-rescue` is excluded from the image, because the boot loader
+entry it provides is broken.
+
+These packages are also excluded, for unknown reasons:
```
atodorov: How do we get to find out the reason ?

image-types/rhel8/amazon-ec2.md
larskarlitski: These packages are removed in the official amazon image. We don't know why, yet.

The idea behind these documents is to find all the reasoning behind these decisions.

image-types/rhel8/amazon-ec2.md
larskarlitski: Thanks!

image-types/rhel8/amazon-ec2.md
larskarlitski: Oops, osbuild-composer actually creates `aarch64` images as well. Thanks!

larskarlitski:
Some changes after talking to @teg (thanks!):
* I removed the awkward "unknown reasons" sections. If this PR is accepted, I'll open issues that point out the differences.
* I added an "Owner: *None*" to each of the documents. This is meant to signal that these are in draft state until we have identified someone who owns the requirements for this image type. This way, we can merge these two now.

image-types/rhel8/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+*osbuild-composer* can create [Red Hat Enterprise Linux 8][rhel] images.
+
+## Architectures
+
+RHEL supports `x86_64`, `aarch64`, `ppc64le`, and `s390x`.
+
+## File System
+
+The default file system is `xfs`.
```
teg: Implicit: only one partition, right?

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,49 @@
+# Amazon EC2 Image
+
+Owner: *None*
+
+This image is meant to be used in [Amazon Elastic Compute Cloud (EC2)][ec2], a
+popular cloud computing platform. It has to follow Amazon’s
+[guidelines][guidelines] and [requirements][requirements] for shared AMIs.
```
teg: Is the intention that we duplicate the parts of the guidelines/requirements that we choose to adopt in this doc, so that if they develop over time, we would make the explicit choice to copy the changes here?

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,49 @@
...
+
+## File Format
+
+EC2 uses Amazon Machine Images (AMIs) internally, which can only be created
+inside EC2. An image in a standard format (ova, vmdk, vhd/x, or raw) must be
+imported from S3 storage. *osbuild-composer* generates this image type in the
+VHDX format, because it is fairly modern, widely used, and uses less space than
+RAW images.
+
+The default image size is 6 GB, unless it has to be larger due to installed
```
teg: Hm, I don't understand. The deafult image size is fixed at 6GB, we never automatically make it larger do we?

ondrejbudai: I was wondering about it too. Lorax does that, I guess we should also implement it.

image-types/rhel8/README.md
larskarlitski: I had that on both the specific documents (ec2 and openstack), but yeah, probably makes sense to add here as well as a default.

image-types/rhel8/amazon-ec2.md
larskarlitski: I think so, yes. I don't think it makes sense to copy everything over straight away. There's a lot of leeway in those documents, though. We should document the choices we're making and the reasons for them here.

image-types/rhel8/amazon-ec2.md
larskarlitski: This was aspirational :wink: 

Changed it until we actually implement that, of course.

larskarlitski:
> This looks good to me. Why not assign the owners now?

We don't have them yet :) Working on getting them as soon as we can.

jgiardino:
This is amazing!

image-types/rhel8/amazon-ec2.md
```diff
@@ -0,0 +1,48 @@
...
+inside EC2. An image in a standard format (ova, vmdk, vhd/x, or raw) must be
+imported from S3 storage. *osbuild-composer* generates this image type in the
+VHDX format, because it is fairly modern, widely used, and uses less space than
+RAW images.
+
+The default image size is 6 GB.
+
+## Architectures
+
+Limited to `x86_64` and `aarch64`, because those are available on EC2.
```
andreasn: bike-shedding, but not only are they available, but they are _the only ones_ available

larskarlitski: You are right! All these beautiful reviews will get dismissed when I push again, so I won't do that and fix it in the next time I touch this file.

larskarlitski:
Thanks a lot everyone!

PR #295, opened 2020-03-07
client: Use http.Client instead of a string for the socket

bcl:
This converts the client and weldrcheck functions to use http.Client for
connections instead of passing around the socket path and opening it for
each test.

### DIFF
--- a/internal/client/blueprints.go
+++ b/internal/client/blueprints.go
@@ -5,12 +5,13 @@ package client
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"strings"
 )
 
 // PostTOMLBlueprintV0 sends a TOML blueprint string to the API
 // and returns an APIResponse
-func PostTOMLBlueprintV0(socket, blueprint string) (*APIResponse, error) {
+func PostTOMLBlueprintV0(socket *http.Client, blueprint string) (*APIResponse, error) {
 	body, resp, err := PostTOML(socket, "/api/v0/blueprints/new", blueprint)
 	if resp != nil || err != nil {
 		return resp, err
@@ -20,7 +21,7 @@ func PostTOMLBlueprintV0(socket, blueprint string) (*APIResponse, error) {
 
 // PostTOMLWorkspaceV0 sends a TOML blueprint string to the API
 // and returns an APIResponse
-func PostTOMLWorkspaceV0(socket, blueprint string) (*APIResponse, error) {
+func PostTOMLWorkspaceV0(socket *http.Client, blueprint string) (*APIResponse, error) {
 	body, resp, err := PostTOML(socket, "/api/v0/blueprints/workspace", blueprint)
 	if resp != nil || err != nil {
 		return resp, err
@@ -30,7 +31,7 @@ func PostTOMLWorkspaceV0(socket, blueprint string) (*APIResponse, error) {
 
 // PostJSONBlueprintV0 sends a JSON blueprint string to the API
 // and returns an APIResponse
-func PostJSONBlueprintV0(socket, blueprint string) (*APIResponse, error) {
+func PostJSONBlueprintV0(socket *http.Client, blueprint string) (*APIResponse, error) {
 	body, resp, err := PostJSON(socket, "/api/v0/blueprints/new", blueprint)
 	if resp != nil || err != nil {
 		return resp, err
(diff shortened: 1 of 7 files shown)


internal/client/blueprints.go
```diff
@@ -1,16 +1,17 @@
-// Package client - blueprints contains functions for the blueprint API
+// Package socket - blueprints contains functions for the blueprint API
```
teg: This seems like a search and replace error ;)

internal/client/client.go
```diff
@@ -1,17 +1,14 @@
-// Package client contains functions for communicating with the API server
+// Package socket contains functions for communicating with the API server
```
teg: Same.

internal/client/client.go
```diff
@@ -111,8 +98,8 @@ func apiError(resp *http.Response) (*APIResponse, error) {
 }
 
 // GetRaw returns raw data from a GET request
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func GetRaw(socket, method, path string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client.go
```diff
@@ -137,8 +124,8 @@ func GetRaw(socket, method, path string) ([]byte, *APIResponse, error) {
 // This function makes 2 requests, the first with limit=0 to get the total number of results,
 // and then with limit=TOTAL to fetch all of the results.
 // The path passed to GetJSONAll should not include the limit or offset query parameters
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func GetJSONAll(socket, path string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client.go
```diff
@@ -166,8 +153,8 @@ func GetJSONAll(socket, path string) ([]byte, *APIResponse, error) {
 }
 
 // PostRaw sends a POST with raw data and returns the raw response body
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func PostRaw(socket, path, body string, headers map[string]string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client.go
```diff
@@ -189,22 +176,22 @@ func PostRaw(socket, path, body string, headers map[string]string) ([]byte, *API
 }
 
 // PostTOML sends a POST with TOML data and the Content-Type header set to "text/x-toml"
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func PostTOML(socket, path, body string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client.go
```diff
@@ -189,22 +176,22 @@ func PostRaw(socket, path, body string, headers map[string]string) ([]byte, *API
...
+func PostTOML(socket *http.Client, path, body string) ([]byte, *APIResponse, error) {
 	headers := map[string]string{"Content-Type": "text/x-toml"}
 	return PostRaw(socket, path, body, headers)
 }
 
 // PostJSON sends a POST with JSON data and the Content-Type header set to "application/json"
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func PostJSON(socket, path, body string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client.go
```diff
@@ -189,22 +176,22 @@ func PostRaw(socket, path, body string, headers map[string]string) ([]byte, *API
...
+func PostJSON(socket *http.Client, path, body string) ([]byte, *APIResponse, error) {
 	headers := map[string]string{"Content-Type": "application/json"}
 	return PostRaw(socket, path, body, headers)
 }
 
 // DeleteRaw sends a DELETE request
-// Errors from the API are returned as an APIResponse, client errors are returned as error
-func DeleteRaw(socket, path string) ([]byte, *APIResponse, error) {
+// Errors from the API are returned as an APIResponse, socket errors are returned as error
```
teg: Same.

internal/client/client_test.go
```diff
@@ -21,9 +23,12 @@ import (
 // socketPath is the path to the temporary unix domain socket
 var socketPath string
 
+// socket is the client connection to the server on socketPath
+var socket *http.Client
```
teg: Hm, is there no way to avoid having this as a global variable and rather pass it down into the test? If that's not possible, feel free to leave as is.

bcl: Reading the testing docs I don't see any way to pass things into the individual tests.

PR #301, opened 2020-03-09
tests: use go's test framework in osbuild-dnf-json-tests

larskarlitski:
This is based on @atodorov's https://github.com/osbuild/osbuild-composer/pull/282. It contains the same commits, but changes the spec file to not use the `gotest` macro for building the binary. It also squashes the last two commits, because the intermediate state doesn't work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build:
 	go build -o osbuild-upload-aws ./cmd/osbuild-upload-aws/
 	go build -o osbuild-tests ./cmd/osbuild-tests/
 	go build -o osbuild-weldr-tests ./cmd/osbuild-weldr-tests/
-	go build -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/
+	go test -c -tags=integration -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/main_test.go
 	go build -o osbuild-rcm-tests ./cmd/osbuild-rcm-tests/
 
 .PHONY: install
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -1,30 +1,18 @@
 // This package contains tests related to dnf-json and rpmmd package.
+
+// +build integration
+
 package main
 
 import (
 	"fmt"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
+	"testing"
 	"path"
 )
-
-func main() {
-	// Tests that the package wrapping dnf-json works as expected
-	dir, err := setUpTemporaryRepository()
-	defer func(dir string) {
-		err := tearDownTemporaryRepository(dir)
-		if err != nil {
-			log.Print("Warning: failed to clean up temporary repository.")
-		}
-	}(dir)
-	if err != nil {
-		log.Panic("Failed to set up temporary repository:", err)
-	}
-	TestFetchChecksum(false, dir)
-}
 
 func setUpTemporaryRepository() (string, error) {
 	dir, err := ioutil.TempDir("/tmp", "osbuild-composer-test-")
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ build:
 	go build -o osbuild-upload-aws ./cmd/osbuild-upload-aws/
 	go build -o osbuild-tests ./cmd/osbuild-tests/
 	go build -o osbuild-weldr-tests ./cmd/osbuild-weldr-tests/
-	go build -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/
+	go test -c -o osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/main_test.go
```
teg: Does this work without the tag? What's the purpose of this make target?

larskarlitski: It does not. I don't usually use this target, that's why I missed this.

I've fixed it to include the tag here. Whether to remove the `make` target is a separate discussion, I'd say.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -47,25 +35,30 @@ func tearDownTemporaryRepository(dir string) error {
...
+func TestFetchChecksum(t *testing.T) {
+	dir, err := setUpTemporaryRepository()
+	defer func(dir string) {
+		err := tearDownTemporaryRepository(dir)
+		if err != nil {
+			t.Errorf("Warning: failed to clean up temporary repository.")
+		}
+	}(dir)
+	if err != nil {
+		t.Errorf("Failed to set up temporary repository: %s", err)
```
ondrejbudai: I would prefer `t.Fatalf` because it doesn't make much sense to continue in the test when the repo is not set up.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/julienschmidt/httprouter v1.2.0
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/stretchr/testify v1.4.0 // indirect
+	github.com/stretchr/testify v1.4.0
```
ondrejbudai: I fail to see why this line changed... We are not using `testify` directly, are we?

golang-github-osbuild-composer.spec
```diff
@@ -68,9 +68,17 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 %gobuild -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
 %gobuild -o _bin/osbuild-weldr-tests %{goipath}/cmd/osbuild-weldr-tests
-%gobuild -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
 %gobuild -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
 
+# Build test binaries with `go test -c`, so that they can take advantage of
```
ondrejbudai: Have you tried this on RHEL? Also, I'm a bit worried what happens when `%gobuild` definition changes in future Fedora/RHEL versions.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -47,25 +35,30 @@ func tearDownTemporaryRepository(dir string) error {
...
 		IgnoreSSL: true,
 	}
-	if !quiet {
-		log.Println("Running TestFetchChecksum on:", dir)
-	}
 	rpmMetadata := rpmmd.NewRPMMD(path.Join(dir, "rpmmd"))
 	_, c, err := rpmMetadata.FetchMetadata([]rpmmd.RepoConfig{repoCfg}, "platform:f31")
 	if err != nil {
-		log.Panic("Failed to fetch checksum:", err)
+		t.Errorf("Failed to fetch checksum: %s", err)
```
ondrejbudai: `t.Fatalf` might be worth considering here as well.

Makefile
atodorov: @teg it looks like some folks are using this target when working on new test binaries, myself included.

atodorov:
@larskarlitski thanks for picking up on this. One additional question:

Do we want the rename like it is now:
`cmd/osbuild-dnf-json-tests/main.go → cmd/osbuild-dnf-json-tests/main_test.go`

or

`cmd/osbuild-dnf-json-tests/main.go → cmd/osbuild-dnf-json_test.go` and having all test binaries as files under the cmd/ directory ?

teg:
The workaround is indeed a bit fragile-looking. However, all said, I think it may be worth it. I suggest dropping all the flags that are not necessary to make it build (may get out of sync anyway, and how this this thing is built should not be important for us). To make it all sustainable, we should probably file a RFE to add some macro we could use to the official RPM macros (even if that will take time to trickle down). I did not yet try building on RHEL.

ondrejbudai:
I was able to successfully build the RPMs and run `osbuild-dnf-json-tests` on rhel 8.2.

cmd/osbuild-dnf-json-tests/main_test.go
larskarlitski: Good point. Thanks!

go.mod
larskarlitski: Must have been from a previous run. Fixed by running `go mod tidy`.

`/tools/prepare-source.sh` should have found this, and it did, but it's hooked up wrong so we didn't notice. Filed https://github.com/osbuild/osbuild-composer/pull/307 to fix that.

cmd/osbuild-dnf-json-tests/main_test.go
larskarlitski: Thanks again :)

golang-github-osbuild-composer.spec
larskarlitski: Yeah, it is a bit hacky indeed. I've changed it like @teg suggested. Building it in such a way that still works without the internal variables. I do hope it works everywhere: I tried Fedora 31, but didn't have time for RHEL yet. Will try that in the morning.

teg: @ondrejbudai noted below that he checked that this works.

PR #303, opened 2020-03-10
proposal: osbuild composer alternative suitable for interactive debugging

msehnout:
As I've already mentioned in the document about becoming a more welcoming open-source project, I'd love to have the possibility to run osbuild-composer in debugger. If it was possible with zero configuration, it'd be the best.

This PR contains an alternative to `cmd/osbuild-composer` which is independent of systemd and thus I can just click the bug button in IDE and I can easily debug the code in our web APIs. See `cmd/osbuild-composer/debug_main.go` for the implementation.

It uses conditional compilation so it does not affect the production binary in any way. Most of the code in the file is just creating sockets and setting environment variables which are set by systemd. Neither of these is a subject to frequent change, so it shouldn't go out of sync often.

_Note: I've requested review from everybody who I think might be interested. Sorry if I missed you!_

### DIFF
--- a/cmd/osbuild-composer/configuration.go
+++ b/cmd/osbuild-composer/configuration.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"github.com/osbuild/osbuild-composer/internal/distro"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+)
+
+type connectionConfig struct {
+	CACertFile     string
+	ServerKeyFile  string
+	ServerCertFile string
+}
+
+func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
+	caCertPEM, err := ioutil.ReadFile(c.CACertFile)
+	if err != nil {
+		return nil, err
+	}
+
+	roots := x509.NewCertPool()
+	ok := roots.AppendCertsFromPEM(caCertPEM)
+	if !ok {
+		panic("failed to parse root certificate")
+	}
+
+	cert, err := tls.LoadX509KeyPair(c.ServerCertFile, c.ServerKeyFile)
+	if err != nil {
+		return nil, err
+	}
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		ClientAuth:   tls.RequireAndVerifyClientCert,
+		ClientCAs:    roots,
+	}, nil
+}
+
+func createDistroConfiguration(registryPaths []string) (rpmmd.RPMMD, distro.Distro, *distro.Registry) {
+	cacheDirectory, ok := os.LookupEnv("CACHE_DIRECTORY")
+	if !ok {
+		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
+	}
+
+	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"))
+
+	distros, err := distro.NewDefaultRegistry(registryPaths)
+	if err != nil {
+		log.Fatalf("Error loading distros: %v", err)
+	}
+
+	distribution, err := distros.FromHost()
+	if err != nil {
+		log.Fatalf("Could not determine distro from host: " + err.Error())
+	}
+
+	return rpm, distribution, distros
+}
--- a/cmd/osbuild-composer/debug_main.go
+++ b/cmd/osbuild-composer/debug_main.go
@@ -1,0 +1,86 @@
+// +build debug
+
+package main
+
+import (
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/jobqueue"
+	"github.com/osbuild/osbuild-composer/internal/rcm"
+	"github.com/osbuild/osbuild-composer/internal/store"
+	"github.com/osbuild/osbuild-composer/internal/weldr"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"path/filepath"
+)
+
+func setUpUDSListener(path string) net.Listener {
+	if err := os.RemoveAll(path); err != nil {
+		log.Panic(err)
+	}
+	l, err := net.Listen("unix", path)
+	if err != nil {
+		log.Panic("listen error:", err)
+	}
+	return l
+}
+
+func setUpTCPListener(socket string) net.Listener {
+	l, err := net.Listen("tcp", socket)
+	if err != nil {
+		log.Panic("listen error:", err)
+	}
+	return l
+}
+
+func main() {
+	// Set up temp directory
+	dir, err := ioutil.TempDir("", "osbuild-composer-debug-")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	// Set environment variable which are set by systemd in production
+	stateDir := filepath.Join(dir, "state")
+	_ = os.Mkdir(stateDir, 0700)
+
+	cacheDir := filepath.Join(dir, "cache")
+	_ = os.Mkdir(cacheDir, 0700)
+
+	_ = os.Setenv("STATE_DIRECTORY", stateDir)
+	_ = os.Setenv("CACHE_DIRECTORY", stateDir)
+
+	// Create all necessary distro configurations
+	rpm, distribution, distros := createDistroConfiguration([]string{"."})
+	store := store.New(&stateDir, distribution, *distros)
+
+	// Run all the APIs
+	logger := log.New(os.Stdout, "", 0)
+
+	// UDS only Job API
+	jobListener := setUpUDSListener(filepath.Join(dir, "job.socket"))
+	defer jobListener.Close()
+	jobAPI := jobqueue.New(logger, store)
+	go func() {
+		err := jobAPI.Serve(jobListener)
+		common.PanicOnError(err)
+	}()
+
+	// RCM API running on localhost
+	rcmListener := setUpTCPListener("127.0.0.1:8080")
+	defer rcmListener.Close()
+	rcmAPI := rcm.New(logger, store, rpm)
+	go func() {
+		err := rcmAPI.Serve(rcmListener)
+		// If the RCM API fails, take down the whole process, not just a single gorutine
+		log.Fatal("RCM API failed: ", err)
+	}()
+
+	weldrListener := setUpUDSListener(filepath.Join(dir, "weldr.socket"))
+	defer weldrListener.Close()
+	weldrAPI := weldr.New(rpm, common.CurrentArch(), distribution, logger, store)
+	err = weldrAPI.Serve(weldrListener)
+	common.PanicOnError(err)
+}
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -1,61 +1,28 @@
+// +build !debug
+
 package main
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"flag"
-	"github.com/osbuild/osbuild-composer/internal/rcm"
-	"io/ioutil"
-	"log"
-	"path"
-	"os"
-
 	"github.com/osbuild/osbuild-composer/internal/common"
-	"github.com/osbuild/osbuild-composer/internal/distro"
 	"github.com/osbuild/osbuild-composer/internal/jobqueue"
-	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"github.com/osbuild/osbuild-composer/internal/rcm"
 	"github.com/osbuild/osbuild-composer/internal/store"
 	"github.com/osbuild/osbuild-composer/internal/weldr"
+	"log"
+	"os"
 
 	"github.com/coreos/go-systemd/activation"
 )
 
-type connectionConfig struct {
-	CACertFile     string
-	ServerKeyFile  string
-	ServerCertFile string
-}
-
-func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
-	caCertPEM, err := ioutil.ReadFile(c.CACertFile)
-	if err != nil {
-		return nil, err
-	}
-
-	roots := x509.NewCertPool()
-	ok := roots.AppendCertsFromPEM(caCertPEM)
-	if !ok {
-		panic("failed to parse root certificate")
-	}
-
-	cert, err := tls.LoadX509KeyPair(c.ServerCertFile, c.ServerKeyFile)
-	if err != nil {
-		return nil, err
-	}
-	return &tls.Config{
-		Certificates: []tls.Certificate{cert},
-		ClientAuth:   tls.RequireAndVerifyClientCert,
-		ClientCAs:    roots,
-	}, nil
-}
-
 func main() {
+	// Parse command line arguments
 	var verbose bool
 	flag.BoolVar(&verbose, "v", false, "Print access log")
 	flag.Parse()
 
-	stateDir := "/var/lib/osbuild-composer"
-
+	// Set up sockets
 	listeners, err := activation.ListenersWithNames()
 	if err != nil {
 		log.Fatalf("Could not get listening sockets: " + err.Error())
@@ -70,32 +37,18 @@ func main() {
 	if len(composerListeners) != 2 && len(composerListeners) != 3 {
 		log.Fatalf("Unexpected number of listening sockets (%d), expected 2 or 3", len(composerListeners))
 	}
-
 	weldrListener := composerListeners[0]
 	jobListener := composerListeners[1]
 
-	cacheDirectory, ok := os.LookupEnv("CACHE_DIRECTORY")
-	if !ok {
-		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
-	}
-
-	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"))
-
-	distros, err := distro.NewDefaultRegistry([]string{"/etc/osbuild-composer", "/usr/share/osbuild-composer"})
-	if err != nil {
-		log.Fatalf("Error loading distros: %v", err)
-	}
-
-	distribution, err := distros.FromHost()
-	if err != nil {
-		log.Fatalf("Could not determine distro from host: " + err.Error())
-	}
+	// Set up distribution
+	rpm, distribution, distros := createDistroConfiguration([]string{"/etc/osbuild-composer", "/usr/share/osbuild-composer"})
 
 	var logger *log.Logger
 	if verbose {
 		logger = log.New(os.Stdout, "", 0)
 	}
 
+	stateDir := "/var/lib/osbuild-composer"
 	store := store.New(&stateDir, distribution, *distros)
 
 	jobAPI := jobqueue.New(logger, store)
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -114,9 +114,9 @@ func New(stateDir *string, distroArg distro.Distro, distroRegistryArg distro.Reg
 	var s Store
 
 	if stateDir != nil {
-		err := os.Mkdir(*stateDir+"/"+"outputs", 0700)
+		err := os.Mkdir(filepath.Join(*stateDir, "outputs"), 0700)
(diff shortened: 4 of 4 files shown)


teg:
Thanks for writing this up so we have something concrete to discuss.

Happy to merge the first patch if you file that separately, by the way.

My gut reaction is that this should not be needed. My reasoning being that all our library code should be hooked up with proper unit-tests, so that one can step through them in a debugger. The thing you don't get from that is the integration parts in `./cmd/*`, but as this is replacing that, we now end up with code-paths that are not the ones we are actually ending up shipping.

I do not doubt the use for a debugger, but debugging against a live system (with arbitrary local state), sounds to me as something we should not encourage. Rather, we should make it so that this is not necessary. Of course, people are welcome to develop in whatever way they prefer, but the stuff in the repo sholud in some way be blessed as making sense (and not encourage misleading debugging methods, which I think this does).

bcl:
Are there situations where you would want to do a debug build without the debug main? A separate command would make sense if that were the case.

It would be nice if configuration.go has unit tests, now that those are separate functions :)

I think this is a good idea, it's not something that will be used in production, and sometimes having a debugger option is much faster than adding prints -- I know I sure miss using an IDE some days. It looks like the main setup, while not the same code as production main, is small enough that it shouldn't be an issue, assuming developer only use of this.

I'd also like to see a howto for using this with delve (which seem to be the go debugger most recommended), since I've never actually used a debugger with Go :)

msehnout:
> My gut reaction is that this should not be needed.

I understand the gut feeling, but production software always needs debugging at some point.

> My reasoning being that all our library code should be hooked up with proper unit-tests, so that one can step through them in a debugger.

Yes, that would be very nice but given our very low unit-test coverage, it is not the case. I know the "correct" way would be to write more unit tests, but the world is not perfect and sometimes temporary workaround live for a long time before we can finally get rid of them (looking at you libdir hack).

> The thing you don't get from that is the integration parts in ./cmd/*, but as this is replacing that, we now end up with code-paths that are not the ones we are actually ending up shipping.

That is correct (pedantically speaking, in practise I think it is almost the same), but when developing code I tend to run it frequently to see what happens. That is something I cannot do right now and I think it is better to run the code from different entrypoint than not to run it at all. We, of course, don't encourage people to run this instead of the primary main, that's why it is hidden behind the `debug` flag.

Also, I spent years debugging network daemons with various tracing tools because it was almost impossible to run them in debugger and at that time I would have really appreciated something like this.

> I do not doubt the use for a debugger, but debugging against a live system (with arbitrary local state), sounds to me as something we should not encourage.

As I said earlier I think it is better to run it than not to run it, even if running against live system with arbitrary local state. At least we can test more code paths and possibly get more useful error messages.

The other thing is that modern IDEs support remote run/debug and I use it very often because I switch between different Linux and Mac machines. That means I can easily run it in fresh container every time (because we have remote workers :) )

> Rather, we should make it so that this is not necessary.

IMO this would be possible if we had really strong static-analysis. Unit tests and linters are not enough because their possibilities are always limited due to the flexibility the programming language provides.

> Of course, people are welcome to develop in whatever way they prefer, but the stuff in the repo sholud in some way be blessed as making sense (and not encourage misleading debugging methods, which I think this does).

As mentioned before, it is primarily for development cycle (write -> run in debugger -> repeat), but if you encounter a bug in our production code and you are able to reproduce it even with this code, it can make the debugging much faster. So I would not call it misleading, rather supplementary.

msehnout:
> Are there situations where you would want to do a debug build without the debug main? A separate command would make sense if that were the case.

I wanted to avoid separate `cmd` because then it would look like something we could ship (but we don't want to)

> It would be nice if configuration.go has unit tests, now that those are separate functions :)

Yes, but let's not rush things too much, we will see if we drop this or not.

> I think this is a good idea, it's not something that will be used in production, and sometimes having a debugger option is much faster than adding prints -- I know I sure miss using an IDE some days. It looks like the main setup, while not the same code as production main, is small enough that it shouldn't be an issue, assuming developer only use of this.

Exactly, I really thing the divergence from the production main is very limited, basically just getting rid of systemd. As Tom mentioned, bigger problem is probably running the code on arbitrary system, but on the other hand, it should be our responsibility to make osbuild indepentent of the host system as much as possible.

> I'd also like to see a howto for using this with delve (which seem to be the go debugger most recommended), since I've never actually used a debugger with Go :)

That is very easy:
```
$ go build -tags debug ./cmd/osbuild-composer
$ dlv exec ./osbuild-composer
Type 'help' for list of commands.
(dlv) help
The following commands are available:
    args ------------------------ Print function arguments.
    break (alias: b) ------------ Sets a breakpoint.
    breakpoints (alias: bp) ----- Print out info for active breakpoints.
    call ------------------------ Resumes process, injecting a function call (EXPERIMENTAL!!!)
    clear ----------------------- Deletes breakpoint.
    clearall -------------------- Deletes multiple breakpoints.
    condition (alias: cond) ----- Set breakpoint condition.
    config ---------------------- Changes configuration parameters.
    continue (alias: c) --------- Run until breakpoint or program termination.
    deferred -------------------- Executes command in the context of a deferred call.
    disassemble (alias: disass) - Disassembler.
    down ------------------------ Move the current frame down.
    edit (alias: ed) ------------ Open where you are in $DELVE_EDITOR or $EDITOR
    exit (alias: quit | q) ------ Exit the debugger.
    frame ----------------------- Set the current frame, or execute command on a different frame.
    funcs ----------------------- Print list of functions.
    goroutine ------------------- Shows or changes current goroutine
    goroutines ------------------ List program goroutines.
    help (alias: h) ------------- Prints the help message.
    list (alias: ls | l) -------- Show source code.
    locals ---------------------- Print local variables.
    next (alias: n) ------------- Step over to next source line.
    on -------------------------- Executes a command when a breakpoint is hit.
    print (alias: p) ------------ Evaluate an expression.
    regs ------------------------ Print contents of CPU registers.
    restart (alias: r) ---------- Restart process.
    set ------------------------- Changes the value of a variable.
    source ---------------------- Executes a file containing a list of delve commands
    sources --------------------- Print list of source files.
    stack (alias: bt) ----------- Print stack trace.
    step (alias: s) ------------- Single step through program.
    step-instruction (alias: si)  Single step a single cpu instruction.
    stepout --------------------- Step out of the current function.
    thread (alias: tr) ---------- Switch to the specified thread.
    threads --------------------- Print out info for every traced thread.
    trace (alias: t) ------------ Set tracepoint.
    types ----------------------- Print list of types
    up -------------------------- Move the current frame up.
    vars ------------------------ Print package variables.
    whatis ---------------------- Prints type of an expression.
Type help followed by a command for full documentation.
(dlv) b rcm.submit
Breakpoint 1 set at 0x7cd9cb for github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit() ./internal/rcm/api.go:81
(dlv) c
Running the debug build, do NOT use this in production!
```
In separate terminal run: `curl -X POST -v http://127.0.0.1:8080/v1/compose`
```
2020/03/11 10:01:46 POST /v1/compose
> github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit() ./internal/rcm/api.go:81 (hits goroutine(9):1 total:1) (PC: 0x7cd9cb)
Warning: debugging optimized function
    76:	
    77:	func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
    78:		writer.WriteHeader(http.StatusNotFound)
    79:	}
    80:	
=>  81:	func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
    82:		// Check some basic HTTP parameters
    83:		contentType := request.Header["Content-Type"]
    84:		if len(contentType) != 1 || contentType[0] != "application/json" {
    85:			writer.WriteHeader(http.StatusBadRequest)
    86:			return
(dlv) bt
0  0x00000000007cd9cb in github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit
   at ./internal/rcm/api.go:81
1  0x00000000007cf406 in github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit-fm
   at ./internal/rcm/api.go:81
2  0x0000000000706d2e in github.com/julienschmidt/httprouter.(*Router).ServeHTTP
   at /home/msehnout/go/pkg/mod/github.com/julienschmidt/httprouter@v1.2.0/router.go:334
3  0x00000000007cd841 in github.com/osbuild/osbuild-composer/internal/rcm.(*API).ServeHTTP
   at ./internal/rcm/api.go:70
4  0x00000000006d0c34 in net/http.serverHandler.ServeHTTP
   at /usr/lib/golang/src/net/http/server.go:2802
5  0x00000000006cd5a5 in net/http.(*conn).serve
   at /usr/lib/golang/src/net/http/server.go:1890
6  0x000000000045c3c1 in runtime.goexit
   at /usr/lib/golang/src/runtime/asm_amd64.s:1357
(dlv) locals
(no locals)
(dlv) s
> github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit() ./internal/rcm/api.go:83 (PC: 0x7cd9e2)
Warning: debugging optimized function
    78:		writer.WriteHeader(http.StatusNotFound)
    79:	}
    80:	
    81:	func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
    82:		// Check some basic HTTP parameters
=>  83:		contentType := request.Header["Content-Type"]
    84:		if len(contentType) != 1 || contentType[0] != "application/json" {
    85:			writer.WriteHeader(http.StatusBadRequest)
    86:			return
    87:		}
    88:	
(dlv) 
> github.com/osbuild/osbuild-composer/internal/rcm.(*API).submit() ./internal/rcm/api.go:84 (PC: 0x7cda25)
Warning: debugging optimized function
    79:	}
    80:	
    81:	func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
    82:		// Check some basic HTTP parameters
    83:		contentType := request.Header["Content-Type"]
=>  84:		if len(contentType) != 1 || contentType[0] != "application/json" {
    85:			writer.WriteHeader(http.StatusBadRequest)
    86:			return
    87:		}
    88:	
    89:		type Repository struct {
(dlv) locals
contentType = (unreadable read out of bounds)

```

dvdhrm:
> Yes, that would be very nice but given our very low unit-test coverage, it is not the case. I know the "correct" way would be to write more unit tests, but the world is not perfect [...]

But what is holding us back? What is the reason to add this, when you say the _correct way_ is to test via unit-tests?

In the end, I agree that we want to be able to run services in a debugger. But what is wrong with attaching delve to a running process? Why not optimize there and make sure this can be done easily? I would be much more inclined to add scripts into `tools/` which look for a running composer instance and attaches to it. So all you do is `make rpm-install && systemctl restart foo && ./tools/debug-composer`, or something like it.

I am all for making sure our code-base allows for a wide variety of development work-flows. What I am missing though is a reasoning what exactly is not working? What is this PR trying to solve? What is the problem description? The only justification in PR-summary and commit message I found is this:

    `I'd love to have the possibility to run osbuild-composer in debugger. If it was possible with zero configuration, it'd be the best.`

But why does the current code not allow to run in a debugger? And what is the reasoning for the approach taken in this PR? What alternatives exist, what is their drawback? Is an intrusive approach the only viable option? What is the justification for this change?

ondrejbudai:
I agree with @dvdhrm . Having a helper script which runs delve on a running osbuild-composer sounds very reasonable and it doesn't introduce any new code paths. @msehnout, do you see any issues/cons in this approach?

bcl:
I have been able to use delve attach <PID> successfully with it running from systemd, so unless I'm missing something it looks like the real benefit to this is that the IDE could launch it for debugging.

msehnout:
> But what is holding us back? What is the reason to add this, when you say the _correct way_ is to test via unit-tests?

Our testing strategy is still unclear to me, so I just stated the fact.

> In the end, I agree that we want to be able to run services in a debugger. But what is wrong with attaching delve to a running process? Why not optimize there and make sure this can be done easily? I would be much more inclined to add scripts into tools/ which look for a running composer instance and attaches to it. So all you do is make rpm-install && systemctl restart foo && ./tools/debug-composer, or something like it.

It is much more convenient this way.

> I am all for making sure our code-base allows for a wide variety of development work-flows. What I am missing though is a reasoning what exactly is not working? What is this PR trying to solve? What is the problem description? The only justification in PR-summary and commit message I found is this:

It was mostly discussed in google documents, but in general this is part of my effort to make the development more convenient.

> I agree with @dvdhrm . Having a helper script which runs delve on a running osbuild-composer sounds very reasonable and it doesn't introduce any new code paths. @msehnout, do you see any issues/cons in this approach?

I don't see the need for helper script. If we expect developers to run delve from cmd line we can as well don't do anything

> I have been able to use delve attach successfully with it running from systemd, so unless I'm missing something it looks like the real benefit to this is that the IDE could launch it for debugging.

Exactly.

Imagine the workflow we are proposing here since we want to get rid of `make install` as well:
 1. change the code
 2. build RPM (this will take a while, since we are running tests in the rpmbuild as well)
 3. install it
 4. run systemd start
 5. run systemd status to get the PID
 6. attach delve to the PID

What I propose:
 1. click the bug button

If I am wrong about the workflow, please let me know. Also I think an example of a helper script would help to move this discussion forward, anyone wants to work on it or should I do it?

ondrejbudai:
However, I think there some differences in the outcomes of the steps you described:

If you use the "click the bug button" method, you don't have a worker running. To run a worker you need to perform an additional step. Also, another tinkering is required to access the weldr socket from cockpit-composer.

If you use the rpm method, everything is set up correctly. You are right that it takes a while but you can automate it.

msehnout:
Yeah, thanks for the note. I added a note about it into the previous comment.

> If you use the rpm method, everything is set up correctly. You are right that it takes a while but you can automate it.

That is correct, but running RPM build whenever you want to test your code changes sounds like a no go solution for me. Do we really want to use this as the preffered method of development?

bcl:
> Imagine the workflow we are proposing here since we want to get rid of `make install` as well:

Not me. I used make build && make install. My workflow looks like this:

* Setup vm with current rpm builds of lorax-composer and osbuild-composer.
* edit files on workstation with vim and lots of plugins
* rsync code tree w/o .git over to VM
* make build && make install (mixed with systemctl stop osbuild-composer depending on state)
* restart service

so I depend on the rpms for initial setup, but for actual work I'm using the make targets. Adding a rpm build/install step is just going to slow things down. It should be *available* but not the only way.

dvdhrm:
> > But what is holding us back? What is the reason to add this, when you say the _correct way_ is to test via unit-tests?
> 
> Our testing strategy is still unclear to me, so I just stated the fact.

Sorry, but this does not answer my question. What is holding us back? What is the reason for this change, when you say yourself that there is a "correct way"?

> > In the end, I agree that we want to be able to run services in a debugger. But what is wrong with attaching delve to a running process? Why not optimize there and make sure this can be done easily? I would be much more inclined to add scripts into tools/ which look for a running composer instance and attaches to it. So all you do is make rpm-install && systemctl restart foo && ./tools/debug-composer, or something like it.
> 
> It is much more convenient this way.

You are proposing a controversial change. But where is the evaluation of alternatives, a justification why the approach you take is considered superior? What are the drawbacks, what are the advantages? I already asked this before:

```
[...] what is the reasoning for the approach taken in this PR?
What alternatives exist, what is their drawback? Is an
intrusive approach the only viable option? What is the
justification for this change?
```

I don't see how `It is much more convenient this way.` answers these questions.

> > I am all for making sure our code-base allows for a wide variety of development work-flows. What I am missing though is a reasoning what exactly is not working? What is this PR trying to solve? What is the problem description? The only justification in PR-summary and commit message I found is this:
> 
> It was mostly discussed in google documents, but in general this is part of my effort to make the development more convenient.

If things were discussed elsewhere, please copy relevant information over into commit-messages, so they are self-contained and descriptive.

Your change is very intrusive. It changes the resulting binary for **everyone** if compiled with debugging enabled. Hence, I want this to be well evaluated and documented. And I recommend alternatives, if they achieve the same, just to be out of the hot-path.

msehnout:
> Sorry, but this does not answer my question. What is holding us back? What is the reason for this change, when you say yourself that there is a "correct way"?

First of all it takes time to write unit-tests and in general, we are in a hurry. Second, it is not always possible to write unit-tests because we interact with dnf, rpm, etc. and we agreed not to require `dnf`, `createrepo_c`, etc. to run the unit test and leave this for the integration tests, but the integration tests are running against an actual osbuild-composer, so it does not provide the possibility to use it like I propose.

So the "correct way" is more theoretical than actually possible.

> where is the evaluation of alternatives

Alternative is not to do anything, but that is what we currently do.

> What are the drawbacks, what are the advantages?

The drawback are ~100 lines of code which is not used in production, advantage is the possibility to run it without systemd as a standalone binary. I don't think there is more to it, just this.

>  I don't see how It is much more convenient this way. answers these questions.

By stating the only advantage. The disadvantage is clear from the surrounding discussion.

> If things were discussed elsewhere, please copy relevant information over into commit-messages, so they are self-contained and descriptive.

It is a draft PR, @teg asked me to create it so that we have a place to have discussion about it and also to have a real example. The commit messages are not supposed to be final.

> Your change is very intrusive. It changes the resulting binary for everyone if compiled with debugging enabled. Hence, I want this to be well evaluated and documented. And I recommend alternatives, if they achieve the same, just to be out of the hot-path.

This is very manipulative, of course it changes the binary if you change the compiler setting, what would be the point otherwise? As I mentioned above, the alternative is not to do anything.

teg:
It seems there is not much appetite for this change. I suggest we do not go down this road.

I few comments still:

> First of all it takes time to write unit-tests and in general, we are in a hurry.

I don't think this is, or at least should be, the case. We should never develop things in such a hurry that we do not have the time to add unit-tests in order to better debug whatever we are doing. It often happens that code slips in without enough tests, and that is something we should improve on. But the reason should never be that we are in too much of a hurry, then we need to assign more time to issues or make issues smaller.

> Second, it is not always possible to write unit-tests because we interact with dnf, rpm, etc. and we agreed not to require `dnf`, `createrepo_c`, etc. to run the unit test and leave this for the integration tests, but the integration tests are running against an actual osbuild-composer, so it does not provide the possibility to use it like I propose.

This is, I believe the point. When we can write unit tests (possibly with mocked out versions of `rpmmd` and friends), that is what we should do to help with our debugging (which means we can run things through a debugger until we have written enough unit tests to identify some issue. When we cannot write unit tests for some issue, then it means that the problem is in the integration with the surrounding system, which is exactly the part this PR is replacing, so I think it isn't the right solution there either.

> Alternative is not to do anything, but that is what we currently do.

I don't agree with this. The alternative, as discussed above is to work on unit tests and document how to use a debugger against our integration tests.

> The drawback are ~100 lines of code which is not used in production, advantage is the possibility to run it without systemd as a standalone binary. I don't think there is more to it, just this.

I think that's not capturing the discussion correctly.

The advantage is to be able to quickly run the project through a debugger to step through the code and identify issues that would otherwise have to be identified by first writing more extensive unit-test coverage. So in the short-term it is a time-saving measure. It could also be able to sometimes identify bugs in our integration that happens to be the same between the devel version and the real version, though this I think is much more hit-and-miss and in no way something we could rely upon.

The disadvantage is that using this tool when unit-tests were the answer means that the debugging is not reproducible, and next time a similar/same issue crops up we have to start from scratch, rather than rely on already written tests. Also, a downside is that we may have both false positives and negatives when trying to reproduce bugs in debug mode vs. the shipped binary, as the code-paths are really different in the cases where unit-tests do not apply. We are trying to do our integration tests (and unit tests) in perfectly reproducible environments, and this sort of setup would go in the other direction.

Overall, I think it is perfectly fine for someone to carry this sort of patch locally if that helps them in the day-to-day development, but before we take it upstream and make it the recommended way to develop, it needs a much stronger consensus that this is the way to go, and also a it needs a consistent story that this fits into the overall way we want to do testing and development, where the my main concern is about reproducibility and predictability.

Lastly, I do think we still have work to do to make it easier in particular to run and debug integration tests, even though I don't think this particular PR solves the problem in a way I'd be with merging.

> It is a draft PR, @teg asked me to create it so that we have a place to have discussion about it and also to have a real example. The commit messages are not supposed to be final.

Just a general note on that: I think it is important to write commits and commit messages in a way that is designed to explain the context and why something is done, especially on a PR that is meant to be the basis for discussion, even if it just a draft (doesn't mean they have to be perfect on the first attempt, just saying what the goal should be).

msehnout:
> It seems there is not much appetite for this change. I suggest we do not go down this road.

Fair enough, closing this for now.

PR #304, opened 2020-03-10
osbuild-composer: don't use hardcoded state directory

msehnout:
Use $STATE_DIRECTORY environment variable which is set by systemd
because we use: StateDirectory=osbuild-composer in the service unit.

_Note: I haven't yet had time to run integration tests on this => draft._

### DIFF
--- a/distribution/osbuild-composer.service
+++ b/distribution/osbuild-composer.service
@@ -15,6 +15,7 @@ Restart=on-failure
 
 # systemd >= 240 sets this, but osbuild-composer runs on earlier versions
 Environment="CACHE_DIRECTORY=/var/cache/osbuild-composer"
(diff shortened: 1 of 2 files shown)


distribution/osbuild-composer.service
```diff
@@ -15,6 +15,7 @@ Restart=on-failure
 
 # systemd >= 240 sets this, but osbuild-composer runs on earlier versions
 Environment="CACHE_DIRECTORY=/var/cache/osbuild-composer"
+Environment="STATE_DIRECTORY=/var/lib/osbuild-composer"
```
larskarlitski: Heh, I almost wanted to say that setting `Environment` doesn't work, but of course it does: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment=

PR #306, opened 2020-03-10
Finish rewriting the image tests to Go

ondrejbudai:
This PR makes the osbuild-image-tests binary doing the same set of tests
like the old test/run script.

Changes from test/run:
- qemu/nspawn are now killed gracefully. Firstly, SIGTERM is sent. If the process doesn't exit till the timeout, SIGKILL is sent. I changed this because nspawn leaves some artifacts behind when killed by SIGKILL.
- the unsharing of network namespace now works differently because of systemd/systemd#15079

### DIFF
--- a/cmd/osbuild-image-tests/osbuild-image-tests.go
+++ b/cmd/osbuild-image-tests/osbuild-image-tests.go
@@ -19,11 +19,11 @@ import (
 )
 
 type testcaseStruct struct {
-	Compose struct {
+	ComposeRequest struct {
 		Distro   string
 		Arch     string
 		Filename string
-	}
+	} `json:"compose-request"`
 	Manifest  json.RawMessage
 	ImageInfo json.RawMessage `json:"image-info"`
 }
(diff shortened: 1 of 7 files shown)


cmd/osbuild-image-tests/osbuild-image-tests.go
```diff
@@ -197,52 +519,54 @@ func getAllCases() ([]string, error) {
...
 
 		currentArch := common.CurrentArch()
-		if testcase.Compose.Arch != currentArch {
-			log.Printf("%s: skipping, the required arch is %s, the current arch is %s", path, testcase.Compose.Arch, currentArch)
+		if testcase.ComposeRequest.Arch != currentArch {
+			log.Printf("%s: skipping, the required arch is %s, the current arch is %s", path, testcase.ComposeRequest.Arch, currentArch)
 			continue
 		}
 
 		hostDistroName, err := distro.GetHostDistroName()
```
teg: This is unrelated to this PR, but in general I think we may want to test all the distros we can, not just the host. We want to be able to cross-build. Maybe that can be in a follow-up once we actually add support for that though.

ondrejbudai:
This *should* be now ready to be merged. I will mark it as Ready for review when the tests finish succesfully on my machine.

PR #308, opened 2020-03-10
Move more tests from travis to github actions

teg:
This started out looking into fixing the source test, but as Travis is so slow, I decided to move over the tests we can.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,18 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v1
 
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -25,6 +25,18 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v1
 
+      - name: Check that source has been perpared
```
larskarlitski:
```suggestion
      - name: Check that source has been prepared
```

teg: Fixed! Thanks!

PR #319, opened 2020-03-12
store: move ComposeRequest to the store package

larskarlitski:
A ComposeRequest is data used to submit a compose to the store, so it
should live in that package.

Remove the json marshalling test, because ComposeRequest is never
marshalled to JSON.

This will allow to use types from `distro` in the ComposeRequest struct.


internal/store/store.go
```diff
@@ -645,8 +656,8 @@ func (s *Store) PushCompose(composeID uuid.UUID, bp *blueprint.Blueprint, packag
 
 // PushComposeRequest is an alternative to PushCompose which does not assume a pre-defined distro, as such it is better
 // suited for RCM API and possible future API that would respect the fact that we can build any distro and any arch
-func (s *Store) PushComposeRequest(request common.ComposeRequest) error {
-	// This should never happen and once distro.Manifest is refactored this check will go away
+func (s *Store) PushComposeRequest(request ComposeRequest) error {
+	// This should never happen and once distro.Pipeline is refactored this check will go away
```
teg: Probably a left-over? `distro.Pipeline` is no longer a thing.

larskarlitski: Yeah, I wrote the patch before the pipeline → manifest switchover and messed up the rebase. Thanks for the catch! Fixed.

Issue #291, opened 2020-03-06
Clarify and automate test case generation


jkozol:
**Problem**
Our test/cases can verify the validity of a pipeline and image created from a specified blueprint, distro, and boot method. Currently, they are manually written. This is suboptimal since each additional distro or boot method will require several additional tests and with the move from dnf to rpm the cases will be tedious to manually write.

**Solution**
A python script will be able to generate test cases for each new distro and boot method. This script can be called by a developer with certain parameters and it will output json test cases. This will require clarifying which test cases are needed. Testing each combination of distro, output type, boot method, and blueprint customization will create too many test cases. Instead, the combinations that need to be tested should be clarified so as to reduce unnecessary test cases. For the script, it will need to generate pipelines and the resulting image's info. 

**Achievables**
- [x] Add a flag to `osbuild-pipeline` so it can output the results of rpmmd instead of the pipeline manifest
- [x] Update `distro/distro_test.go` to accept both package checksums and package specs
- [x] Create python script that accepts a `boot` and a `compose-request` key as input and uses `osbuild-pipeline` and `image-info` to generate test cases
- [x] Generate tests for combinations of boot methods and image types on x86_64

teg:
This was implemnted! Thanks @jkozol.

PR #320, opened 2020-03-12
osbuild-pipeline: make blueprints optional and read them from stdin

teg:
This fits with how our other tools work: passing JSON on stdin/stdout and makes it easier to use this from other tools. In particular will simplify #291.

Additionally, clean up a bit @jkozol patch and add some test-case clean-ups on top. This moves towards test-cases being generated, but there are still cases that would not workout out-of-the-box.


tools/generate-test-cases
```diff
@@ -0,0 +1,84 @@
...
+        "openstack": "qemu",
+        "qcow2": "qemu",
+        "vhd": "qemu",        
+        "vmdk": "qemu",
+        "ami": "qemu-extract",
+    }
+    return format_boot_dict[output_format]
+
+
+def main(compose_request, store):
```
ondrejbudai: The name `compose_request` can create confusion because of the `store.ComposeRequest` struct we have. We may consider using something like `image_request`.

tools/generate-test-cases
```diff
@@ -0,0 +1,84 @@
...
+                    subprocess.run(["xz", "--decompress", "--stdout", image_file], stdout=f)
+                image_file = fn
+        test_case["image-info"] = json.loads(subprocess.check_output(["tools/image-info", image_file]))
+
+    return test_case
+
+
+if __name__ == '__main__':
+    args = json.load(sys.stdin)
+    test_case = main(args, ".osbuild")
```
ondrejbudai: I'm not sure if I like hardcoded path here. I know it's just development tool but if the path was passed in as an argument, the developer knows for sure where to find the image without looking into this script's source code.

cmd/osbuild-pipeline/main.go
```diff
@@ -23,12 +23,12 @@ type rpmMD struct {
...
 func main() {
 	var imageType string
-	var blueprintArg string
+	var blueprintArg bool
 	var archArg string
 	var distroArg string
 	var rpmmdArg bool
 	flag.StringVar(&imageType, "image-type", "", "image type, e.g. qcow2 or ami")
-	flag.StringVar(&blueprintArg, "blueprint", "", "path to a JSON file containing a blueprint to translate")
+	flag.BoolVar(&blueprintArg, "blueprint", false, "read a blueprint from stdin")
```
ondrejbudai: Hmm, I think it might be still useful to pass blueprint both from a file and stdin. Many tools can use both filename or `-` as "the main input" (`gcc` for example). It shouldn't make the code much more complicated and I can imagine some cases when it would be more convenient.

tools/generate-test-cases
teg: Hm, @larskarlitski do you remember our discussion around this?
If we should rename this, I suggest doing that in a separate PR, as this is just following the naming of the test case structs.

tools/generate-test-cases
teg: Hehe, yeah, this was meant to be made configurable. An oversight. Thanks!

cmd/osbuild-pipeline/main.go
teg: Fair point. Fixed!

tools/generate-test-cases
ondrejbudai: Alright, this isn't imo a big issue.

teg:
Ah, yes. I forgot to comment on that. I think we should do that in a separate commit after all the things are autogenerated. Before moving to rpm. Just to make it easier to follow.

PR #321, opened 2020-03-13
how swagger generation could look like

msehnout:
We currently reference Weldr API documentation as a source of information for our own API. As we will build our own API (for the hosted version) it would be nice to already have a tool for documenting it. In this PR I created a small PoC how go-swagger could be used to generate OpenAPI 2.0 docs for the RCM API. 

It is not complete, just to give an idea how it could look like.

More info:
https://github.com/go-swagger/go-swagger
https://swagger.io/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,3 +86,7 @@ rpm: $(RPM_SPECFILE) $(RPM_TARBALL)
 	rpmbuild -bb \
 		--define "_topdir $(CURDIR)/rpmbuild" \
 		$(RPM_SPECFILE)
+
+.PHONY: swagger-specs
+swagger-specs:
+	swagger generate spec --scan-models --output=docs/rcm-api.yaml
--- a/docs/rcm-api.yaml
+++ b/docs/rcm-api.yaml
@@ -1,0 +1,91 @@
+definitions:
+  UUID:
+    description: |-
+      A UUID is a 128 bit (16 byte) Universal Unique IDentifier as defined in RFC
+      4122.
+    items:
+      format: uint8
+      type: integer
+    type: array
+    x-go-package: github.com/google/uuid
+  errorReason:
+    description: errorReason represents the response in case of any error
+    properties:
+      error_reason:
+        description: a string describing what happened
+        type: string
+        x-go-name: Error
+    required:
+    - error_reason
+    type: object
+    x-go-package: github.com/osbuild/osbuild-composer/internal/rcm
+  statusReply:
+    description: statusReply represents the response when getting a status of a running
+      compose
+    properties:
+      status:
+        description: the status of running compose
+        type: string
+        x-go-name: Status
+    required:
+    - status
+    type: object
+    x-go-package: github.com/osbuild/osbuild-composer/internal/rcm
+  submitReply:
+    description: submitReply represents the response after submitting a compose
+    properties:
+      compose_id:
+        $ref: '#/definitions/UUID'
+    required:
+    - compose_id
+    type: object
+    x-go-package: github.com/osbuild/osbuild-composer/internal/rcm
+info:
+  description: It's primary use case is for the RCM team. As such it is driven solely
+    by their requirements.
+  title: RCM API for osbuild-composer
+  version: "1"
+paths:
+  /v1/compose:
+    post:
+      consumes:
+      - application/json
+      operationId: submit
+      produces:
+      - application/json
+      responses:
+        "200":
+          description: submitReply
+          schema:
+            $ref: '#/definitions/submitReply'
+        default:
+          description: errorReason
+          schema:
+            $ref: '#/definitions/errorReason'
+      schemes:
+      - http
+      summary: Submit requests for composes.
+  /v1/compose/{uuid}:
+    get:
+      operationId: status
+      parameters:
+      - description: uuid to display status for
+        in: path
+        name: uuid
+        required: true
+        type: string
+      produces:
+      - application/json
+      responses:
+        "200":
+          description: status of a running compose
+          schema:
+            items:
+              $ref: '#/definitions/statusReply'
+            type: object
+        default:
+          description: unexpected error
+          schema:
+            $ref: '#/definitions/errorReason'
+      summary: Check current status of a running compose.
+swagger: "2.0"
--- a/internal/rcm/api.go
+++ b/internal/rcm/api.go
@@ -1,5 +1,10 @@
-// Package rcm provides alternative HTTP API to Weldr.
+// Package rcm RCM API for osbuild-composer
+//
 // It's primary use case is for the RCM team. As such it is driven solely by their requirements.
+//
+// Version: 1
+//
+// swagger:meta
 package rcm
 
 import (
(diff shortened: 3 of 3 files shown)


teg:
Interesting. Can we see the sample output anywhere?

msehnout:
> Interesting. Can we see the sample output anywhere?

Yes, it's the `swagger.yaml` or do you mean the generated website?

teg:
I meant the website :)

msehnout:
I'm pretty sure there are plenty of options, but the easiest on is to copy-paste the yaml here:
https://editor.swagger.io/

teg:
Thanks!

teg:
At first glance this looks pretty nice. Maybe others have other ideas or alternatives, but it seems to me that we do want something along these lines.

teg:
By the way, @msehnout there appeared to be some error in the yaml, at least the page you linked to complained.

msehnout:
I don't yet understand the first error but I expect it to be sth like missing info about the whole API. The second one is just complaining that `$ref: '#/responses/errorReason'` does not point to an object definition.

My primary intention here was to showcase as soon as possible and discuss before I go further.

teg:
Makes sense!

ondrejbudai:
I think swagger is widely accepted these days and it looks very nice, so I will gladly accept it. :)

bcl:
Nice! How does this look in go doc output? Does it recognize the markup?

msehnout:
> Nice! How does this look in go doc output? Does it recognize the markup?

It displays the comment as usual:
```
$ go doc ./internal/rcm
package rcm // import "github.com/osbuild/osbuild-composer/internal/rcm"

Package rcm RCM API for osbuild-composer

It's primary use case is for the RCM team. As such it is driven solely by
their requirements.

Version: 1

swagger:meta

type API struct{ ... }
    func New(logger *log.Logger, store *store.Store, rpmMetadata rpmmd.RPMMD) *API
```

msehnout:
> By the way, @msehnout there appeared to be some error in the yaml, at least the page you linked to complained.

I've fixed those errors, it was fairly straight forward.

msehnout:
I made the PR ready for review. The commit now includes full specification of the RCM API.

msehnout:
> I don't know much about `swagger` yet. I have a couple of preliminary questions:
> 
>     1. swagger seems to support generating the OpenAPI specs from source, as well as generating source from OpenAPI specs. Why did you choose the former?

Because the code is already written. We can try to generate the code for our next API if we want to or we can experiment more with RCM API. Anyway both should go to separate PR.

> 
>     2. Why is the spec committed when it is generated from the source?

At first it was because I wanted to show it to you, then I thought it would be useful for website generation is osbuild.github.io repo, but it can as well be created there.

> 
>     3. Why is this generating OpenAPI 2.0 specs, when 3.0 seems to be the current version.
> 

It seems go-swagger is not going to support it due to the specs being too complicated: https://github.com/go-swagger/go-swagger/issues/1122 , but 2.0 seems to be sufficient for our current needs.

>     4. What is the best way to obtain the `swagger` binary?

To download their container and run it with `podman`.
https://github.com/swagger-api/swagger-codegen#public-pre-built-docker-images

larskarlitski:
> > 1. swagger seems to support generating the OpenAPI specs from source, as well as generating source from OpenAPI specs. Why did you choose the former?
> 
> Because the code is already written. We can try to generate the code for our next API if we want to or we can experiment more with RCM API. Anyway both should go to separate PR.

Heh, I see. What would be the ideal model, though? It is hard to assess the value of this PR without knowing the direction this effort is going to.

> > 2. Why is the spec committed when it is generated from the source?
> 
> At first it was because I wanted to show it to you, then I thought it would be useful for website generation is osbuild.github.io repo, but it can as well be created there.

In general, I'd say not to include anything that cannot be recreated again. If there is a reason to commit it, we need to make sure that it stays in sync, like we do with go vendoring in [tools/prepare-source.sh](https://github.com/osbuild/osbuild-composer/blob/master/tools/prepare-source.sh).

> > 4. What is the best way to obtain the `swagger` binary?
> 
> To download their container and run it with `podman`.
> https://github.com/swagger-api/swagger-codegen#public-pre-built-docker-images

These are not working for me at all when I run them like you do in the Makefile here:
```
$ podman run --rm swaggerapi/swagger-codegen-cli generate spec --scan-models --output=docs/rcm-api.yaml
Exception in thread "main" io.airlift.airline.ParseArgumentsUnexpectedException: Found unexpected parameters: [spec, --scan-models]
	at io.airlift.airline.Cli.validate(Cli.java:148)
	at io.airlift.airline.Cli.parse(Cli.java:116)
	at io.airlift.airline.Cli.parse(Cli.java:97)
	at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:35)

$ podman run --rm swaggerapi/swagger-codegen-cli 
usage: swagger-codegen-cli <command> [<args>]

The most commonly used swagger-codegen-cli commands are:
    config-help   Config help for chosen lang
    generate      Generate code with chosen lang
    help          Display help information
    langs         Shows available langs
    meta          MetaGenerator. Generator for creating a new template set and configuration for Codegen.  The output will be based on the language you specify, and includes default templates to include.
    validate      Validate specification
    version       Show version information

See 'swagger-codegen-cli help <command>' for more information on a specific
command.

$ podman run --rm swaggerapi/swagger-codegen-cli help spec 
Unknown command spec
```

I think there are too many open questions to take this as it is.

In addition, there are some nits in the code itself: the `Makefile` target should depend on the file and the rule seems useless to me if the official way to run this is through the container. Also, the documentation comments use a different format.

msehnout:
> Heh, I see. What would be the ideal model, though? It is hard to assess the value of this PR without knowing the direction this effort is going to.

The ideal model is the one we agree on. Some prefer writing code, some prefer generating it and I don't think there is a single truth here.

I, personally, prefer to write code and generate docs from it.

> In general, I'd say not to include anything that cannot be recreated again. If there is a reason to commit it, we need to make sure that it stays in sync, like we do with go vendoring in tools/prepare-source.sh.

Sure, that's not the most important thing to discuss here. Let's focus on the other questions.

> These are not working for me at all when I run them like you do in the Makefile here:

Sorry I cannot reproduce :-(, I'm running on mac and my Linux machine is in the office.

> In addition, there are some nits in the code itself: the Makefile target should depend on the file 

I agree

> and the rule seems useless to me if the official way to run this is through the container. 

You can create an alias, but again not super important at this point. We can as well just create a CI to generate it for us and not care about client machines at  all.

> Also, the documentation comments use a different format.

Yeah, that is indeed unfortunate and hard to believe, but that's what the documentation suggests.

Before we dive further into this discussion, what do we agree on?
 1. We want documentation for our REST API
 2. We want to use OpenAPI specification (be it 2.0 or 3.0)
 3. Go-swagger fits our needs
 4. Let's find a different generator or write our own

internal/rcm/api.go
```diff
@@ -42,7 +79,47 @@ func New(logger *log.Logger, store *store.Store, rpmMetadata rpmmd.RPMMD) *API {
...
+	//   '200':
+	//     description: status of a running compose
+	//     schema:
+	//       type: object
+	//       items:
+	//         "$ref": "#/definitions/statusReply"
+	//   default:
+	//     description: unexpected error
+	//     schema:
+	//       "$ref": "#/definitions/errorReason"
```
dvdhrm: Are we sure we want to put this comment here rather than in front of the `status` function definition?

docs/rcm-api.yaml
```diff
@@ -0,0 +1,91 @@
+definitions:
```
dvdhrm: I think we should not commit this. I don't see the advantage in having this in the repo.

Makefile
```diff
@@ -86,3 +86,7 @@ rpm: $(RPM_SPECFILE) $(RPM_TARBALL)
 	rpmbuild -bb \
 		--define "_topdir $(CURDIR)/rpmbuild" \
 		$(RPM_SPECFILE)
+
+.PHONY: swagger-specs
+swagger-specs:
+	swagger generate spec --scan-models --output=docs/rcm-api.yaml
```
dvdhrm: What is the requirement here? Where does the `swagger` tool come from? What is the recommended way to run this?

I think it is ok to make this a run-always target. A dependency list would easily get out-of-sync, and we cannot magically know which files the swagger tool inspects. So I wouldn't worry about making this more fine-grained.

internal/rcm/api.go
msehnout: It does not matter, it is just a personal preference.

docs/rcm-api.yaml
msehnout: Agreed.

Makefile
msehnout: In your case probably AUR. For fedora it should run in a container as mentioned above, but seems to fail. I used homebrew to install it and it ran just fine, it's sad that the container fails.

msehnout:
Closing this in favor of #345

PR #323, opened 2020-03-13
ci: check also the unstaged files

ondrejbudai:
Prior this commit missing files in the vendor directory would be ignored by the CI.


teg:
Looks reasonable. Would mean that if your run the script locally and have random files around it would complain, but I guess that's ok. @larskarlitski ?

ondrejbudai:
I changed only the CI config yaml, nothing run locally changed. :slightly_smiling_face:

PR #324, opened 2020-03-13
test/cases: move all the tests over to being autogenerated ones

teg:
Also rework travis integration to no longer have any knowledge about the test cases, and to be ready for test cases for other distros.


teg:
This has been replaced by #327

PR #325, opened 2020-03-14
Switch weldrcheck to testing

bcl:
The one remaining thing to do here is to figure out how to make the linter ignore the 'deadcode' in utils.go, or how to organize it so that other tests can use the code w/o triggering lint errors.


golang-github-osbuild-composer.spec
```diff
@@ -77,6 +77,7 @@ export GOFLAGS=-mod=vendor
 TEST_LDFLAGS="${LDFLAGS:-} -B 0x$(od -N 20 -An -tx1 -w100 /dev/urandom | tr -d ' ')"
 
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
+go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-weldr-tests %{goipath}/cmd/osbuild-weldr-tests
```
atodorov: Path mismatch between here and Makefile

internal/weldrcheck/blueprints_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+// +build integration
+
+package weldrcheck
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	//	"strings"
```
atodorov: Looks like this is not used. Should it be removed instead ?

internal/weldrcheck/blueprints_test.go
```diff
@@ -0,0 +1,754 @@
...
+	"time"
+
+	"github.com/BurntSushi/toml"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/osbuild-composer/internal/client"
+)
+
+// Hold test state to share between tests
+var testState *TestState
```
atodorov: The comment  here triggered my spider sense. There shouldn't be any state shared between the tests (as much as possible) and in fact from what I can see only the socket is shared.  Maybe this just needs better wording. OTOH if we need only the socket why not make a global variable called `testSocket` or similar ?

internal/weldrcheck/blueprints_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+// Setup the socket to use for running the tests
+// Also makes sure there is a running server to test against
+func TestMain(m *testing.M) {
+	var err error
+	testState, err = setupTestState("/run/weldr/api.socket", 60*time.Second)
+	if err != nil {
+		fmt.Printf("ERROR: Test setup failed: %s", err)
+		os.Exit(1)
+	}
```
atodorov: Why not use `require.Nilf` here and abort immediately in case of error ?

internal/weldrcheck/blueprints_test.go
```diff
@@ -0,0 +1,754 @@
...
+	bp := `{
+		"name": "test-invalid-json-blueprint-v0",
+		"version": "0.0.1",
+		"modules": [{"name: "util-linux", "version": "*"}],
+	}`
+
+	resp, err := client.PostJSONBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "failed with a client error")
+	require.False(t, resp.Status, "did not return an error")
+}
```
atodorov: From what I see the invalid yaml/json scenarios are using payload which isn't valid YAML/JSON. A different scenario is when the payload is a valid file format but the contents inside are invalid, e.g. missing a required field. 

Would also help if you add a pointer/comment as to exactly which part of the string isn't valid. In this case I believe it is the comma on the last line which will choke the JSON parser but that is hard to spot.

internal/weldrcheck/blueprints_test.go
```diff
@@ -0,0 +1,754 @@
...
+	}`
+
+	resp, err = client.PostJSONWorkspaceV0(testState.socket, bp)
+	require.NoError(t, err, "POST workspace failed with a client error")
+	require.True(t, resp.Status, "POST workspace failed: %#v", resp)
+
+	// Get the blueprint, make sure it is the modified one and that changes = true
+	info, api, err := client.GetBlueprintsInfoJSONV0(testState.socket, "test-delete-blueprint-changes-ws-v0")
+	require.NoError(t, err, "GET blueprint failed with a client error")
+	require.Nil(t, api, "GET blueprint request failed: %#v", api)
```
atodorov: Shouldn't this be `require.Nilf` or am I not understanding something ?

golang-github-osbuild-composer.spec
bcl: Good catch!

internal/weldrcheck/blueprints_test.go
bcl: Correct.

internal/weldrcheck/blueprints_test.go
bcl: So the alternative is to setup and tear down the connection in each test, which is slower and lots more code. There is no way to pass this from the main test function to the test functions so it has to be a global, and it has to be in each of the test files. It also contains the API version so that the tests can be extended to run API v1 specific tests at some point soon.

I also want to share this setup/state code with other tests (there are going to be sources_test.go, compose_test.go, etc. at some point).

internal/weldrcheck/blueprints_test.go
bcl: I haven't figured out what the 'f' suffix is for. The docs, and in practice, indicate that all of them support formatted error messages. So I'm not sure what their purpose is.

internal/weldrcheck/blueprints_test.go
bcl: This isn't inside a test, it's inside the main setup so you can't use the library.

internal/weldrcheck/blueprints_test.go
bcl: TOML not YAML :) In these tests I am using either an empty blueprint or one that has a formatting problem. Added comments to the Invalid tests, thanks!

internal/weldrcheck/blueprints_test.go
atodorov: hmm, my impression was that -f methods supported string formatting and the others didn't. At least this is how I've used them in my limited experience. But if you say it works and I am not going to argue.

ondrejbudai:
@bcl I think I was able to solve the `deadcode` linter issue in #336. I would love your opinion on it. :slightly_smiling_face:

bcl:
> @bcl I think I was able to solve the `deadcode` linter issue in #336. I would love your opinion on it.

Actually I solved it by adding the integration tag to the file, as well as adding deadcode exceptions, take a look at the latest version of the file.

PR #327, opened 2020-03-14
distro: switch over to RPM-based pipelines

teg:
Also contains lots of clean-ups throughout. Replaces #324.


teg:
> You're adding the s390 travis runs and removing them in a separate commit...

Well, the first commit duplicates the current logic for every distro. The second commit removes s390 from every distro. So conceptually they are indepndent, but you are right, would have been nicer to flip the order.

> Passing `packageSpecs` and the `bluprint` to `distro.Manifest` is redundant in some way, no?

Not the way it now at least. `packageSpecs` are resolved from the `blueprint`, we don't call rpmmd from `distro`.

larskarlitski:
> Not the way it now at least. `packageSpecs` are resolved from the `blueprint`, we don't call rpmmd from `distro`.

Yeah sorry, I know. Just wanted to mention that it feels weird.

teg:
I agree, something to think about.

I force pushed to fix an oversight in the commit dropping the checksum from `distro.Manifest`, also another commit to fix a nil-pointer dereference I noticed.

PR #329, opened 2020-03-16
ci: migrate image tests on Travis to the Go implementation

ondrejbudai:
After the `dnf` to `rpm` switch this is now much simpler and I love it! :-)

Minor hacks (see `constants.go`) were needed but I think it's worth it, because we now have only one implementation of the image tests.

Migration to Github actions isn't sadly possible, because it doesn't support nested virtualization. :-(


teg:
Very nice! Any chance we can have ARM support too? Then it would be a no brainer.

ondrejbudai:
I will see what's possible. My plan was to drop `test/run` and use the go implementation to ensure they don't diverge. I decided I want to also try the github actions but I'm not sure that's possible.

PR #331, opened 2020-03-16
tests: rewrite RCM tests to use go test framework

jrusz:
This is my attempt at rewriting the current rcm test using the testify assert and require packages. I'm not sure if this will behave in the same way, I might have removed too much.

### DIFF
--- a/cmd/osbuild-rcm-tests/main_test.go
+++ b/cmd/osbuild-rcm-tests/main_test.go
@@ -1,18 +1,20 @@
 // osbuild-rcm-tests run tests against running osbuild-composer instance that was spawned using the
 // osbuild-rcm.socket unit. It defines the expected use cases of the RCM API.
+
+// +build integration
+
 package main
 
 import (
 	"encoding/json"
 	"github.com/google/uuid"
-	"log"
+	"github.com/stretchr/testify/require"
 	"net/http"
-	"os"
 	"strings"
+	"testing"
 )
 
-func main() {
-	failed := false
+func TestRCM(t *testing.T) {
 	// This is the first request the user sends to osbuild.
 	submitBody := `
 		{
@@ -41,40 +43,18 @@ func main() {
 	// Case 1: POST request
 
 	resp, err := http.Post(socket+endpoint, "application/json", strings.NewReader(submitBody))
-	if err != nil {
-		log.Fatal("Failed to submit a compose: ", err.Error())
-	}
-	if resp.StatusCode != 200 {
-		log.Print("Error: the ", endpoint, " returned non 200 status. Full response: ", resp)
-		failed = true
-	} else {
-		err = json.NewDecoder(resp.Body).Decode(&submitResponse)
-		if err != nil {
-			log.Fatal("Failed to decode JSON response from ", endpoint)
-		}
-		log.Print("Success: the ", endpoint, " returned compose UUID: ", submitResponse.UUID)
-	}
+	require.Nilf(t, err, "Failed to submit a compose: %v", err)
+	require.Equalf(t, resp.StatusCode, 200, "Error: the %v returned non 200 status. Full response: %v", endpoint, resp)
+	err = json.NewDecoder(resp.Body).Decode(&submitResponse)
+	require.Nilf(t, err, "Failed to decode JSON response from %v", endpoint)
 
 	// Case 2: GET status
 
 	statusEndpoint := endpoint + "/" + submitResponse.UUID.String()
 	resp, err = http.Get(socket + statusEndpoint)
-	if err != nil {
-		log.Fatal("Failed to get a status: ", err.Error())
-	}
-	if resp.StatusCode != 200 {
-		log.Print("Error: the ", endpoint, " returned non 200 status. Full response: ", resp)
-		failed = true
-	} else {
-		err = json.NewDecoder(resp.Body).Decode(&statusResponse)
-		if err != nil {
-			log.Fatal("Failed to decode JSON response from ", endpoint)
-		}
-		log.Print("Success: the ", statusEndpoint, " returned status: ", statusResponse.Status)
-	}
+	require.Nilf(t, err, "Failed to get a status: %v", err)
+	require.Equalf(t, resp.StatusCode, 200, "Error: the %v returned non 200 status. Full response: %v", endpoint, resp)
+	err = json.NewDecoder(resp.Body).Decode(&statusResponse)
+	require.Nilf(t, err, "Failed to decode JSON response from %v", endpoint)
 
-	// If anything failed return non-zero exit code.
-	if failed {
-		os.Exit(1)
-	}
 }
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -68,7 +68,6 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 %gobuild -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
 %gobuild -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
-%gobuild -o _bin/osbuild-rcm-tests %{goipath}/cmd/osbuild-rcm-tests
 
 # Build test binaries with `go test -c`, so that they can take advantage of
 # golang's testing package. The golang rpm macros don't support building them
@@ -78,6 +77,7 @@ TEST_LDFLAGS="${LDFLAGS:-} -B 0x$(od -N 20 -An -tx1 -w100 /dev/urandom | tr -d '
 
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-weldr-tests %{goipath}/internal/weldrcheck/
+go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-rcm-tests %{goipath}/cmd/osbuild-rcm-tests
 
 %install
 install -m 0755 -vd                                         %{buildroot}%{_libexecdir}/osbuild-composer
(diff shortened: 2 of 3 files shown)


ondrejbudai:
The source code check is failing - to fix it, you need to run `tools/prepare-source.sh`, commit all the changes (including the ones in vendor directory) and push again. :slightly_smiling_face:

Another thing caught my eye: @jrusz added the rcm test into the spec file. When `systemctl start osbuild-composer.socket` is started, rcm socket is not activated, therefore the test will fail. To enable it, you need to run `systemctl start osbuild-composer.socket osbuild-rcm.socket`. This leads to a question: what environment should a test expect? Can we somehow define it? @teg once suggested that a test may have the responsibility to start the required services. Do we want that? I think it sounds reasonable. However, I guess that we want to test multi-arch composes in the future and that means running multiple machines - orchestrating services over many machines seems like a task for someone else than the test executables in `/cmd/*`.

teg:
As these are local tests, I think it would make sense for now to require each test to start whatever sockets/services they require on the local machine. This could be separate to this PR though, and I guess should be made consistent with the other tests.

For the orchestration of multiple tests we should indeed figure something out.

jrusz:
Thanks for the review, I ran the `tools/prepare-source.sh` and rebased on latest master.

cmd/osbuild-rcm-tests/main_test.go
```diff
@@ -41,40 +41,18 @@ func main() {
...
-		err = json.NewDecoder(resp.Body).Decode(&submitResponse)
-		if err != nil {
-			log.Fatal("Failed to decode JSON response from ", endpoint)
-		}
-		log.Print("Success: the ", endpoint, " returned compose UUID: ", submitResponse.UUID)
-	}
+	require.Nilf(t, err, "Failed to submit a compose: %v", err)
+	assert.Equalf(t, resp.StatusCode, 200, "Error: the %v returned non 200 status. Full response: %v", endpoint, resp)
+	err = json.NewDecoder(resp.Body).Decode(&submitResponse)
+	require.Nilf(t, err, "Failed to decode JSON response from %v", endpoint)
```
msehnout: This implementation will attempt to decode the body even in case the HTTP return code wasn't 200. Isn't not a big problem, but I think we could preserve the if statement here so that we don't inform about HTTP failure and JSON failure at the same time. If we preserve it, we can also inform about the success.

cmd/osbuild-rcm-tests/main_test.go
```diff
@@ -41,40 +41,18 @@ func main() {
...
-		err = json.NewDecoder(resp.Body).Decode(&statusResponse)
-		if err != nil {
-			log.Fatal("Failed to decode JSON response from ", endpoint)
-		}
-		log.Print("Success: the ", statusEndpoint, " returned status: ", statusResponse.Status)
-	}
+	require.Nilf(t, err, "Failed to get a status: %v", err)
+	assert.Equalf(t, resp.StatusCode, 200, "Error: the %v returned non 200 status. Full response: %v", endpoint, resp)
+	err = json.NewDecoder(resp.Body).Decode(&statusResponse)
+	require.Nilf(t, err, "Failed to decode JSON response from %v", endpoint)
```
msehnout: Same case as above.

cmd/osbuild-rcm-tests/main_test.go
```diff
@@ -41,40 +41,18 @@ func main() {
...
 
-	// If anything failed return non-zero exit code.
-	if failed {
-		os.Exit(1)
-	}
```
msehnout: How do we report failure here? Does the testing framework return non-zero return code in case any `assert` failed?

cmd/osbuild-rcm-tests/main_test.go
larskarlitski: Shouldn't the body always be valid JSON, even on error codes? If so, I think this is correct, but might warrant a comment.

cmd/osbuild-rcm-tests/main_test.go
larskarlitski: Yes, the idea is that returning non-0 from a test executable means it failed.

cmd/osbuild-rcm-tests/main_test.go
msehnout: I'm not sure, the body can be anything depending on the content-type, can it? I'd expect it to be empty when the code is 400: https://play.golang.org/p/LcPMoym15-7

msehnout: I think it will print the error, check this if I wrote a correct reproducer:
```
~/t/gohttp [1] $ cat server.go

package main

import (
    "log"
    "net/http"
)

func handler(w http.ResponseWriter, r *http.Request) {
    w.WriteHeader(http.StatusBadRequest)
}

func main() {
    http.HandleFunc("/", handler)
    log.Fatal(http.ListenAndServe(":8080", nil))
}
~/t/gohttp $ cat client.go
package main

import (
	"encoding/json"
	"fmt"
	"net/http"
)

func main() {
	resp, err := http.Get("http://127.0.0.1:8080/")
	fmt.Println("err: ", err)
	fmt.Println("resp: ", resp)
	fmt.Println("resp.Body: ", resp.Body)
	var statusResponse struct {
		Status string `json:"status"`
	}
	err = json.NewDecoder(resp.Body).Decode(&statusResponse)
	fmt.Println("err: ", err)
	if err != nil {
		fmt.Println("err != nil")
	}
	fmt.Println("statusResponse: ", statusResponse)
	fmt.Println("statusResponse.Status: ", statusResponse.Status)
}                                                                                                                  
~/t/gohttp $ go run server.go &
~/t/gohttp $ go run client.go
err:  <nil>
resp:  &{400 Bad Request 400 HTTP/1.1 1 1 map[Content-Length:[0] Date:[Tue, 17 Mar 2020 08:55:49 GMT]] {} 0 [] false false map[] 0xc000114000 <nil>}
resp.Body:  {}
err:  EOF
err != nil
statusResponse:  {}
statusResponse.Status:
```

larskarlitski: > I'm not sure, the body can be anything depending on the content-type, can it? I'd expect it to be empty when the code is 400: https://play.golang.org/p/LcPMoym15-7

Right, I mean our API should always return valid JSON, no?

jrusz: I was not sure how to write the asserts and use the if else statement together. Would this make sense? 
```
        if resp.StatusCode != 200 {
￼		require.Nilf(t, err, "Failed to submit a compose: %v", err)
￼	} else {
￼		err = json.NewDecoder(resp.Body).Decode(&submitResponse)
￼	        require.Nilf(t, err, "Failed to decode JSON response from %v", endpoint)
￼		t.Log("Success: the ", endpoint, " returned compose UUID: ", submitResponse.UUID)
￼	}
```

larskarlitski: Yeah this would work, but I think we always want to treat the reply as JSON, as you had before. Unless I'm misunderstanding something. Let's wait for @msehnout's reply.

msehnout: Yes, we don't return anything else than JSON.

cmd/osbuild-rcm-tests/main_test.go
jrusz: Yes, I just verified that when an `assert` fails it exits with `1`

msehnout: ok, in that case it seems to be ready for merging as soon as the CI is happy

jrusz: > ok, in that case it seems to be ready for merging as soon as the CI is happy

Yeah for that we need to figure out an acceptable way to start the rcm socket.

ondrejbudai: I think we should just assume someone does that for us (we do the same thing with `osbuild-composer` service currently) and we can think of a proper solution in a follow-up.

msehnout: I think @jrusz was referring to the CI failure below. Why is it even running in the CI? We don't yet have a support for running the tests in `/cmd`, do we?

ondrejbudai: Oh, I understood that incorrectly. Integration tests are not meant to be run on Github Actions/Travis, so we just use a tag in these tests. When `go test ./...` is run, it ignores all the files which has a tag specified. See https://github.com/osbuild/osbuild-composer/blob/54e27e607ca865d1a7793640ca5f134daea13d45/cmd/osbuild-dnf-json-tests/main_test.go#L3

Just put `// +build integration` in `cmd/osbuild-rcm-tests/main_test.go` and everything should be alright.

cmd/osbuild-rcm-tests/main_test.go
atodorov: `assert.Equalf(t, resp.StatusCode, 200,` -> `require.Equal` and that will short-circuit the test execution not trying to decode JSON in case of HTTP errors.  Wouldn't that be sufficient ?

golang-github-osbuild-composer.spec
```diff
@@ -89,8 +89,8 @@ install -m 0755 -vd                                         %{buildroot}%{_libex
 install -m 0755 -vp _bin/osbuild-tests                      %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 install -m 0755 -vp _bin/osbuild-weldr-tests                %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 install -m 0755 -vp _bin/osbuild-dnf-json-tests             %{buildroot}%{_libexecdir}/tests/osbuild-composer/
-install -m 0755 -vp _bin/osbuild-image-tests                %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 install -m 0755 -vp _bin/osbuild-rcm-tests                  %{buildroot}%{_libexecdir}/tests/osbuild-composer/
+install -m 0755 -vp _bin/osbuild-image-tests                %{buildroot}%{_libexecdir}/tests/osbuild-composer/
```
atodorov: why move this line around? I don't see any changes here.

jrusz: Right, this happened after I rebased onto https://github.com/osbuild/osbuild-composer/pull/333/files I think. I can change it back.

cmd/osbuild-rcm-tests/main_test.go
jrusz: > `assert.Equalf(t, resp.StatusCode, 200,` -> `require.Equal` and that will short-circuit the test execution not trying to decode JSON in case of HTTP errors. Wouldn't that be sufficient ?

Ah yes, I din't think of that, thanks.

Issue #193, opened 2020-01-10
Unify naming, improve readability and code correctness


msehnout:
### Inconsistent naming
The current implementation refers to the same variable with 3 different names:
 1. `composeType` https://github.com/osbuild/osbuild-composer/blob/02a194f6126042aedf9d5ddb3d7bfb8d3d6d3237/internal/store/store.go#L446
 2. `OutputType` https://github.com/osbuild/osbuild-composer/blob/02a194f6126042aedf9d5ddb3d7bfb8d3d6d3237/internal/store/store.go#L475
 3. `outputFormat` https://github.com/osbuild/osbuild-composer/blob/02a194f6126042aedf9d5ddb3d7bfb8d3d6d3237/internal/distro/distro.go#L37

### Undefined terminology
We should start using "type" or "format" consistently across the codebase and also define our terminology, preferably in some guide for contributors:
 * type vs format
 * output vs compose
 * job vs build
 * probably some more

### Weak typing
Also I would prefer to use enumeration for architectures and compose types instead of a string which is error prone. For example:
```go
// Not necessarily valid Go code
type Architecture int

const (
        X86_64 Base = iota
        AARCH64
        PPC64
)

// helper function as an input validation
func ArchitectureFromString(input string) Architecture { 
    // ...
}
```

teg:
I agree. Thanks for writing this up!

Enums over strings seems like a no-brainer to me, so no further comments on that.

Naming:
 - I think we should use the term "Job" over "Build" (just makes sense to me that a "worker" does a "job").
 - The weldr API uses ComposeType, we may want to do the same. Though we should check what the UI does/wants to do, and consider following that instead if different (@jgiardino, thoughts?)? Also, we may end up wanting to have several outputs, with different types for a given compose (azure+aws+...), and then referring to this as ComposeType might be wrong.

jgiardino:
Here is an example from a [design document](https://docs.google.com/presentation/d/1LCvefUG_a3o4M2Cf4xbhtNHEeyJJ2uzfQpmycKEBJHY/edit?usp=sharing) I put together last year. This example shows how we could present the hierarchy of a compose, and all the images/jobs created from that compose, and then the uploads per image.

After doing a few rounds of reviews, the terminology I ended up using is the following. But please let me know if you have suggestions.

- Compose
- Image build
- Upload action

[...]

PR #332, opened 2020-03-16
rcm: merge the two PushCompose functions

larskarlitski:
See individual commits. Lots of cleanup-type commits leading up to be able to merge these two functions.

@msehnout I wonder about the benefit of the common types. There's lots of code to convert between them and strings, all of which can return an error.

### DIFF
--- a/internal/rcm/api.go
+++ b/internal/rcm/api.go
@@ -16,6 +16,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/julienschmidt/httprouter"
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/distro"
 	"github.com/osbuild/osbuild-composer/internal/store"
 )
 
@@ -27,15 +28,17 @@ type API struct {
 	// rpmMetadata is an interface to dnf-json and we include it here so that we can
 	// mock it in the unit tests
 	rpmMetadata rpmmd.RPMMD
+	distros     *distro.Registry
 }
 
 // New creates new RCM API
-func New(logger *log.Logger, store *store.Store, rpmMetadata rpmmd.RPMMD) *API {
+func New(logger *log.Logger, store *store.Store, rpmMetadata rpmmd.RPMMD, distros *distro.Registry) *API {
 	api := &API{
 		logger:      logger,
 		store:       store,
 		router:      httprouter.New(),
 		rpmMetadata: rpmMetadata,
+		distros:     distros,
 	}
 
 	api.router.RedirectTrailingSlash = false
@@ -93,7 +96,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 
 	// JSON structure expected from the client
 	var composeRequest struct {
-		Distribution  common.Distribution   `json:"distribution"`
+		Distribution  string                `json:"distribution"`
 		ImageTypes    []common.ImageType    `json:"image_types"`
 		Architectures []common.Architecture `json:"architectures"`
 		Repositories  []Repository          `json:"repositories"`
@@ -156,17 +159,17 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 		})
 	}
 
-	modulePlatformID, err := composeRequest.Distribution.ModulePlatformID()
-	if err != nil {
+	distro := api.distros.GetDistro(composeRequest.Distribution)
+	if distro == nil {
 		writer.WriteHeader(http.StatusBadRequest)
-		_, err := writer.Write([]byte(err.Error()))
+		_, err := writer.Write([]byte("unknown distro"))
 		if err != nil {
 			panic("Failed to write response")
 		}
 	}
 
 	// map( repo-id => checksum )
-	_, checksums, err := api.rpmMetadata.FetchMetadata(repoConfigs, modulePlatformID)
+	_, checksums, err := api.rpmMetadata.FetchMetadata(repoConfigs, distro.ModulePlatformID())
 	if err != nil {
 		writer.WriteHeader(http.StatusBadRequest)
 		_, err := writer.Write([]byte(err.Error()))
@@ -181,7 +184,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 	err = api.store.PushComposeRequest(store.ComposeRequest{
 		Blueprint:       blueprint.Blueprint{},
 		ComposeID:       composeUUID,
-		Distro:          composeRequest.Distribution,
+		Distro:          distro,
 		Arch:            composeRequest.Architectures[0],
 		Repositories:    repoConfigs,
 		Checksums:       checksums,
--- a/internal/rcm/api.go
+++ b/internal/rcm/api.go
@@ -84,18 +84,8 @@ func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
 
 // Depsolves packages and build packages for building an image for a given
 // distro, in the given architecture
-func depsolve(rpmmd rpmmd.RPMMD, distro distro.Distro, repos []rpmmd.RepoConfig, imageType common.ImageType, arch common.Architecture) ([]rpmmd.PackageSpec, []rpmmd.PackageSpec, error) {
-	archString, ok := arch.ToString()
-	if !ok {
-		panic("architecture was validated before")
-	}
-
-	imageTypeString, ok := imageType.ToString()
-	if !ok {
-		panic("image type was validated before")
-	}
-
-	specs, excludeSpecs, err := distro.BasePackages(imageTypeString, archString)
+func depsolve(rpmmd rpmmd.RPMMD, distro distro.Distro, repos []rpmmd.RepoConfig, imageType, arch string) ([]rpmmd.PackageSpec, []rpmmd.PackageSpec, error) {
+	specs, excludeSpecs, err := distro.BasePackages(imageType, arch)
 	if err != nil {
 		return nil, nil, fmt.Errorf("Distro.BasePackages: %v", err)
 	}
@@ -105,7 +95,7 @@ func depsolve(rpmmd rpmmd.RPMMD, distro distro.Distro, repos []rpmmd.RepoConfig,
 		return nil, nil, fmt.Errorf("RPMMD.Depsolve: %v", err)
 	}
 
-	specs, err = distro.BuildPackages(archString)
+	specs, err = distro.BuildPackages(arch)
 	if err != nil {
 		return nil, nil, fmt.Errorf("Distro.BuildPackages: %v", err)
 	}
@@ -184,17 +174,6 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 		})
 	}
 
-	// Image requests are derived from requested image types. All of them are uploaded to Koji, because
-	// this API is only for RCM.
-	requestedImages := []common.ImageRequest{}
-	for _, imgType := range composeRequest.ImageTypes {
-		requestedImages = append(requestedImages, common.ImageRequest{
-			ImgType: imgType,
-			// TODO: use koji upload type as soon as it is available
-			UpTarget: []common.UploadTarget{},
-		})
-	}
-
 	distro := api.distros.GetDistro(composeRequest.Distribution)
 	if distro == nil {
 		writer.WriteHeader(http.StatusBadRequest)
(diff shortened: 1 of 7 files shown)


larskarlitski:
I've manually verified that the rcm integration tests still pass on Fedora (after applying the patch in https://github.com/osbuild/osbuild-composer/pull/333):

```
$ make rpm
$ sudo dnf install rpmbuild/x86_64/RPMS/*.rpm
$ sudo systemctl start osbuild-rcm.socket
$ sudo /usr/libexec/tests/osbuild-composer/osbuild-rcm-tests 
2020/03/17 06:52:21 Success: the v1/compose returned compose UUID: ace94271-418d-4661-818b-74c073759930
2020/03/17 06:52:21 Success: the v1/compose/ace94271-418d-4661-818b-74c073759930 returned status: RUNNING
[vm-user@vm osbuild-composer]$ echo $?
0
```

msehnout:
> @msehnout I wonder about the benefit of the common types. There's lots of code to convert between them and strings, all of which can return an error.

We discussed this when I was working on the prevous refactoring. Most of them will go away once we use ONLY the common types and never strings, because we shouldn't hardcode the same string all around the place.

The other thing is that golang does not provide an enumeration type which leaves us only with type aliases and named constants. This inherently leaves space for errors but it is a language design choice, so there is nothing I can do about it.

internal/rcm/api.go
```diff
@@ -93,7 +96,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
 
 	// JSON structure expected from the client
 	var composeRequest struct {
-		Distribution  common.Distribution   `json:"distribution"`
+		Distribution  string                `json:"distribution"`
```
msehnout: > `ComposeRequest` included a `common.Distribution`, which had to be
resolved in PushComposeRequest. Use a real `distro.Distro` object here,
and push resolving it to the rcm package. That makes more sense, because
the input is validated closer to where it is received now. It also
simplifies `PushComposeRequest`.

This is only partially correct. The fact that you replaced `common.Distribution` here with `string` actually moved the validation further away, because one point of having the common types is to do their validation immediately at user's input and then only operate with them instead of strings.

If you still want to get rid of `common.Distribution` isn't it possible to write this line like this:
```
		Distribution  distro.Distro                `json:"distribution"`
```
That is write a common unmarshaler for `distro.Distro`?

internal/store/store.go
```diff
@@ -705,14 +695,17 @@ func (s *Store) PushComposeRequest(request ComposeRequest) error {
...
 			ImageBuildID: imageBuildID,
-			Distro:       distroString,
+			Distro:       request.Distro.Name(),
 			Manifest:     manifestStruct,
 			Targets:      []*target.Target{},
 			ImageType:    imgTypeCompatStr,
 		})
 
+		// this ought to exist, because we're creating it from an existing distro struct
+		distroTag, _ := common.DistributionFromString(request.Distro.Name())
```
msehnout: I'm not big fan of "ought to". I know it shouldn't happen, but I prefer error handling everywhere.

internal/rcm/api.go
```diff
@@ -82,6 +82,42 @@ func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
 	writer.WriteHeader(http.StatusNotFound)
 }
 
+// Depsolves packages and build packages for building an image for a given
+// distro, in the given architecture
+func depsolve(rpmmd rpmmd.RPMMD, distro distro.Distro, repos []rpmmd.RepoConfig, imageType common.ImageType, arch common.Architecture) ([]rpmmd.PackageSpec, []rpmmd.PackageSpec, error) {
```
msehnout: This looks like something that Weldr and RCM APIs should have in common. Why is it not common?

internal/store/store.go
```diff
@@ -556,8 +554,18 @@ func (s *Store) GetImageBuildImage(composeId uuid.UUID, imageBuildId int) (io.Re
 		return nil, 0, &NoLocalTargetError{"compose does not have local target"}
 	}
 
+	name, ok := imageBuild.Distro.ToString()
+	if !ok {
+		panic("distro name was validated earlier")
```
msehnout: I don't understand the error message here, what does it mean "validated earlier"?

internal/rcm/api.go
```diff
@@ -216,16 +205,7 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
...
-		Blueprint:       blueprint.Blueprint{},
-		ComposeID:       composeUUID,
-		Distro:          distro,
-		Arch:            composeRequest.Architectures[0],
-		Repositories:    repoConfigs,
-		Packages:        packages,
-		BuildPackages:   buildPackages,
-		RequestedImages: requestedImages,
-	})
+	err = api.store.PushCompose(distro, composeUUID, &blueprint.Blueprint{}, packages, buildPackages, arch, imageType, 0, nil)
```
msehnout: The `ComposeRequest` structure was supposed to represent what the user wanted. My intention here was to pass the requirements to the store and the store would do the depsolving etc. This way we are mixing the user input (blueprint) and our internal implementation (packages, build packages) which seems unnecessary to me. The `ComposeRequest` was also nice for decoupling the API logic from the logic in store.

internal/store/store.go
```diff
@@ -662,84 +650,6 @@ func (s *Store) PushCompose(distro distro.Distro, composeID uuid.UUID, bp *bluep
 	return nil
 }
 
-// PushComposeRequest is an alternative to PushCompose which does not assume a pre-defined distro, as such it is better
-// suited for RCM API and possible future API that would respect the fact that we can build any distro and any arch
-func (s *Store) PushComposeRequest(request ComposeRequest) error {
```
msehnout: I wish you said this during the previous refactoring, because my intention was to replace the `PushCompose` with `PushComposeRequest`. Now it feels like we are just replacing code randomly.

internal/rcm/api.go
larskarlitski: I referred to (captial) `ComposeRequest` in that commit message, not (lower-case) `composeRequest`. Sorry for the confusion. I've reworded the commit message to explain this change better.

> [...] because one point of having the common types is to do their validation immediately at user's input and then only operate with them instead of strings.

Right, but because we get the real `Distro` object in the same function now, I ended up with code like this:

```go
name, ok := composeRequest.Distribution.ToString()
// error handling

distro := api.distros.GetDistro(name)
// error handling
```

This effectively validates distro twice, and needlessly converts a `string` to an `int`, just to convert it back to a `string`. It doesn not buy any type-safety either, because go doesn't have enums.

> That is write a common unmarshaler for `distro.Distro`?

That's not possible, because unmarshallers cannot have access to surrounding state (we'd need a `distro.DistroRegistry`.

internal/store/store.go
larskarlitski: But at this point, we're 100% certain that the tag exist. I don't think adding some dead code (`if ok`) is of any use.

(This code was removed in a later commit.)

internal/rcm/api.go
larskarlitski: `weldr` also takes a blueprint as an input, which the RCM API doesn't know anything about. We could move the low-level function to `rpmmd` indeed, but I couldn't come up with a good name yet.

internal/store/store.go
larskarlitski: When creating the `Distro` object. It actually doesn't make sense to have the error here, for the same reason as above: we're 100% certain that this cannot fail.

internal/store/store.go
larskarlitski: > I wish you said this during the previous refactoring, [...]

I'm not sure I understand? Which previous refactoring?

I know that this was your intention, but it didn't happen. Honestly, I don't think we should have merged a function that does something very similar to another function without removing this other function.

My intention here was to merge these two functions, so that it becomes easier to split the queue from the store down the line. It was simpler to keep `PushCompose` than `PushComposeRequest` without sacrificing any existing functionality.

> Now it feels like we are just replacing code randomly.

I thought about this for a very long time. It is not random at all. I'm sorry if I didn't communicate my intention clearly enough in this PR. Hope these comments clear it up a bit.

internal/rcm/api.go
larskarlitski: See below.

larskarlitski:
> We discussed this when I was working on the prevous refactoring. Most of them will go away once we use ONLY the common types and never strings, because we shouldn't hardcode the same string all around the place.
> 
> The other thing is that golang does not provide an enumeration type which leaves us only with type aliases and named constants. This inherently leaves space for errors but it is a language design choice, so there is nothing I can do about it.

I think this is a bigger discussion which we should have separately. I shouldn't have started it on this PR. I'll file an issue.

internal/rcm/api.go
msehnout: If we cannot unmarshal to `distro.Distro` directly, I'd prefer to instead change the `api.distros.GetDistro()` to accept `common.Distribution` instead of a string, it is ready for the change. It just wasn't a priority yet.

So the code will become:
```
distro := api.distros.GetDistro(composeRequest.Distribution)
// error handling
```
And we simply cannot get rid of the error handling here.

internal/store/store.go
msehnout: > (This code was removed in a later commit.)

 👍

internal/rcm/api.go
msehnout: RCM API has a predefined blueprint which is enough. I'd prefer to move this function to `rpmmd`.

internal/store/store.go
msehnout: In this specific case, I'd prefer to change the `GetDistro` signature (same as above) and we can drop this check completely.

For the error message: It might just be my English skill but I'd be very confused if I saw this error message. I'd write something like: "internal error, distro variable contains non-existing value". In general something more specific and longer.

internal/store/store.go
msehnout: > I'm not sure I understand? Which previous refactoring?

https://github.com/osbuild/osbuild-composer/pull/221

> I know that this was your intention, but it didn't happen. Honestly, I don't think we should have merged a function that does something very similar to another function without removing this other function.

This should have been said in #221 and it didn't happen because it wasn't a priority yet. If you consider this an issue with high priority you could just say so during a planning and I would happily work on it.

> My intention here was to merge these two functions, so that it becomes easier to split the queue from the store down the line. It was simpler to keep PushCompose than PushComposeRequest without sacrificing any existing functionality.

This sounds again like something we were discussing but didn't agree on yet. In cases like this we should first create an issue with description.

> I thought about this for a very long time. It is not random at all. I'm sorry if I didn't communicate my intention clearly enough in this PR. Hope these comments clear it up a bit.

As mentioned above, bigger plans like this should start with an issue that describes your intentions.

msehnout:
> I think this is a bigger discussion which we should have separately. I shouldn't have started it on this PR. I'll file an issue.

We had a discussion on this here: https://github.com/osbuild/osbuild-composer/issues/193 and follow up discussions over BlueJeans.

internal/rcm/api.go
larskarlitski: I don't think that is a good idea. It would mean that
1. We validate twice: once the `string` that's coming in from the request, turning it into an int. And the again turning the int in `GetDistro` to a `Distro`. This seems redundant and complex.
2. The function signature would become `GetDistro(distro common.Distribution) Distro`. I think the current one—`GetDistro(name string) Distro`—is much simpler to understand.

internal/store/store.go
larskarlitski: You would never see this error message. I'll remove the check. It is pointless.

internal/store/store.go
larskarlitski: > This should have been said in #221 and it didn't happen because it wasn't a priority yet. If you consider this an issue with high priority you could just say so during a planning and I would happily work on it.

Because I wasn't around when #221 was discussed, I cannot propose changing it now? I think having two functions in the code that do almost the same thing should be merged. I wanted to do it myself, because that was quicker than asking you to do it during planning.

> This sounds again like something we were discussing but didn't agree on yet. In cases like this we should first create an issue with description.

I haven't proposed anything yet. I was giving some background on why I am doing these cleanups now. I think they stand on their own. Please comment on why you think merging them is a bad idea.

internal/rcm/api.go
msehnout: But look at the `GetDistro` implementation. I agree that the signature looks weird, but in that case we should change the name, for example:
```
ToDistroInterface(distributionTag common.Distribution) Distro
```
it's not an awesome example, but represents the idea.

internal/store/store.go
msehnout: >  I wanted to do it myself, because that was quicker than asking you to do it during planning.

I think discussing such things during the planning meeting is the correct way to do it.

> Please comment on why you think merging them is a bad idea.

Without further context (the planning), it goes the opposite direction that I had in mind with these changes.

larskarlitski:
> After discussing the context with Tom, this change makes more sense. I'm happy to merge it now. Thanks!

Thanks a lot :blush:  I've created https://github.com/osbuild/osbuild-composer/issues/342 last night—I'll try doing that earlier next time around.

PR #333, opened 2020-03-17
spec: build and install osbuild-rcm-tests

larskarlitski:


ondrejbudai:
#331 does the same thing. What about we close this one and wait for the other one to prevent conflicts?

larskarlitski:
> #331 does the same thing. What about we close this one and wait for the other one to prevent conflicts?

Ah, I didn't know it did. When is that slated for landing? Looks like it needs to be rebased anyway due to failing tests.

I'd like to run this test regularly as I'm making more changes to the way the store and the RCM API interact (for example, https://github.com/osbuild/osbuild-composer/pull/332).

However, if #331 lands today, I'm happy to juggle this patch on my branches.

PR #335, opened 2020-03-17
tests/image: rewrite to go testing framework

ondrejbudai:
I decided not to use the testing framework in functions which are not directly tight to testing (booting images, running osbuild). I think it's reasonable to use classic error handling there and propagate the errors to places directly tight to testing and then use the testing library. This enables us to reuse the code in different part of projects if needed.

Other changes:
- I removed the hack which assumed that every fedora is Fedora 30. We currently do have tests for other Fedora versions, therefore it's not needed anymore.
- The testing VM has now only 1GB of memory, now it should be able to run in [teg/testvm](https://github.com/teg/test-vm/).

This PR is depending on #336, before it gets merged, the CI will be failing.

### DIFF
--- a/cmd/osbuild-image-tests/main_test.go
+++ b/cmd/osbuild-image-tests/main_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 package main
 
 import (
@@ -13,13 +15,17 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"testing"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/osbuild/osbuild-composer/internal/common"
 	"github.com/osbuild/osbuild-composer/internal/distro"
 )
+
+var distroArg = flag.String("distro", "", "which distro tests to run")
 
 type testcaseStruct struct {
 	ComposeRequest struct {
@@ -43,18 +49,14 @@ func runOsbuild(manifest []byte, store string) (string, error) {
 	var outBuffer bytes.Buffer
 	cmd.Stdout = &outBuffer
 
-	log.Print("[osbuild] running")
 	err := cmd.Run()
 
 	if err != nil {
-		log.Print("[osbuild] failed")
 		if _, ok := err.(*exec.ExitError); ok {
 			return "", fmt.Errorf("running osbuild failed: %s", outBuffer.String())
 		}
 		return "", fmt.Errorf("running osbuild failed from an unexpected reason: %v", err)
 	}
-
-	log.Print("[osbuild] succeeded")
 
 	var result struct {
 		OutputID string `json:"output_id"`
@@ -92,12 +94,10 @@ func splitExtension(path string) (string, string) {
 
 // testImageInfo runs image-info on image specified by imageImage and
 // compares the result with expected image info
-func testImageInfo(imagePath string, rawImageInfoExpected []byte) error {
+func testImageInfo(t *testing.T, imagePath string, rawImageInfoExpected []byte) {
 	var imageInfoExpected interface{}
 	err := json.Unmarshal(rawImageInfoExpected, &imageInfoExpected)
-	if err != nil {
-		return fmt.Errorf("cannot decode expected image info: %v", err)
-	}
+	require.Nilf(t, err, "cannot decode expected image info: %v", err)
 
 	cmd := exec.Command(imageInfoPath, imagePath)
 	cmd.Stderr = os.Stderr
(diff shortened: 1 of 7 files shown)


teg:
Should be ok to rebase on top of master now.

cmd/osbuild-image-tests/main_test.go
```diff
@@ -92,39 +94,27 @@ func splitExtension(path string) (string, string) {
...
 // testImageInfo runs image-info on image specified by imageImage and
 // compares the result with expected image info
-func testImageInfo(imagePath string, rawImageInfoExpected []byte) error {
+func testImageInfo(t *testing.T, imagePath string, rawImageInfoExpected []byte) {
 	var imageInfoExpected interface{}
 	err := json.Unmarshal(rawImageInfoExpected, &imageInfoExpected)
-	if err != nil {
-		return fmt.Errorf("cannot decode expected image info: %v", err)
-	}
+	require.Nil(t, err)
```
atodorov: You can add the error message to the `require.Nil` function and when that fails we'll see a humanly understandable error message instead of that something was expected to be nil and it wasn't.  For example:
```
require.Nilf(t, err, "cannot decode expected image info: %v", err)
```

Note: I personally used `.Nilf` while Brian uses just `.Nil` and both seem to be able to use string formatting for the error message.

cmd/osbuild-image-tests/main_test.go
```diff
@@ -92,39 +94,27 @@ func splitExtension(path string) (string, string) {
...
 	cmd := exec.Command(imageInfoPath, imagePath)
 	cmd.Stderr = os.Stderr
 	reader, writer := io.Pipe()
 	cmd.Stdout = writer
 
 	err = cmd.Start()
-	if err != nil {
-		return fmt.Errorf("image-info cannot start: %v", err)
-	}
+	require.Nil(t, err)
```
atodorov: Same here, looks like you've removed error messages everywhere. IMO it is better for them to stay.

jrusz:
> but what I find weird is that some of the images require Fedora30, is there a reason for that?

Could it be that it's hardcoded in here?
[https://github.com/osbuild/osbuild-composer/blob/master/tools/generate-test-cases#L14](url)

ondrejbudai:
Regarding the error messages: Previously, I added them because it was hard to trace where the actual error happened. However, I think it's not needed anymore because the testify library prints a line when the error happened.

```
err = cmd.Wait()
require.Nilf(t, err, "running image-info failed: %v", err)
}
```

What does this error message provide us? `running-image-info failed` - when I know the exact line where the fail happened, I can derive that `Wait()` call failed and it was called on `image-info` command. This imho gives me the exact same information and as a benefit the code is shorter and I don't have to maintain it.

Nevertheless, there could be cases where more explanation may be useful. WDYT?

> There were some wrong mirrors selected I guess but it seems to be fine in the end

That's a message from `curl` called by `osbuild`. We set `curl` to do multiple attempts and if one attempt fails, `curl` prints that message to the stderr. If all attempts failed, this would of course fail the build, but it didn't happened that time.

> what I find weird is that some of the images require Fedora30, is there a reason for that?

I didn't allow the cross-distro builds before but I guess it should be fine now. However, I think it's out of scope for this PR and we should do it in follow-up if we want to do it.

> Could it be that it's hardcoded in here?
> https://github.com/osbuild/osbuild-composer/blob/master/tools/generate-test-cases#L14

Nah, that's just a comment. The `if` is here:
https://github.com/osbuild/osbuild-composer/blob/e70e6a72a06c150ebbc171294c1ef9393a1801f2/cmd/osbuild-image-tests/main.go#L354

dvdhrm:
> What does this error message provide us? `running-image-info failed` - when I know the exact line where the fail happened, I can derive that `Wait()` call failed and it was called on `image-info` command. This imho gives me the exact same information and as a benefit the code is shorter and I don't have to maintain it.

I agree. I generally don't add error-messages to assertions / programming errors. Either an assertion is obvious, or you should document with code-comments why an assertion exists. I think either option is superior to error messages.

What is the reason in favor of the error messages? In most (if not all) cases they just point out programming mistakes which need to be fixed in the code, rather than mistakes made by the caller/user, right? Hence, a backtrace or file+line information is much more valuable than instructions to the *user*.

atodorov:
> What is the reason in favor of the error messages? 

@dvdhrm from a testers POV, especially when it comes to running an integration type of test I don't necessarily want to look at the source code to figure out what went wrong. A message like "failed to create test dir", "file XXX not found", "network error" will tell me what happened, which could point me to a misconfigured system, not necessarily a problem with the software under test.

In this specific case I favor the error messages b/c they were there in the first place and b/c the asserts framework makes it very easy to actually use them and b/c they will probably save me couple of minutes figuring out what happened when something goes wrong. If there were no messages in the first place I will not insist on adding them everywhere.

That said there is already a mix between the two styles in various parts of the test suite but it looks like we did reuse the existing messages most of the time. This could have been just a coincidence though.

teg:
Makes sense to me to include error messages about things that could be due to a misconfigured system. Ideally we would make no assumptions about system configuration, but at the moment at least we do depend for instance on on the osbulid-composer test having been started externally, in some tests.
I do agree with @ondrejbudai and @dvdhrm that we should not include error messages for purely programmatic errors (which most of the assertions should be about, I think).

ondrejbudai:
Thanks for reviews @jrusz and @atodorov! Sorry, I didn't have time to write here some notes: Yeah, I brought back the message where I think that they are useful. Also I removed the confusing `--distro` argument and also I made all the error messages more verbose by using `%#v` instead of `%v`.

Can I ask for reapprove? I had to rebase.

Issue #340, opened 2020-03-17
Create RPM scriptlets for the rcm subpackage


msehnout:
https://github.com/osbuild/osbuild-composer/blob/54e27e607ca865d1a7793640ca5f134daea13d45/golang-github-osbuild-composer.spec#L124

 - [ ] Investigate what the RPM scriptlets do and if it makes sense, create them for the rcm subpackage as well.

teg:
This makes sense. I believe the policy is that we should do the systemd scriptlets for all unit files we ship, but we should verify that.

msehnout:
Fixed in #344

PR #346, opened 2020-03-18
spec: remove the changelog

ondrejbudai:
The changelog is distribution-specific, therefore it doesn't make sense to have it upstream.

The removal also makes the releasing of new version to Fedora easier (We don't have to sync the changelogs).

However, I could be wrong, feel free to tell me why to keep the changelog in the repository.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -190,12 +190,4 @@ The worker for osbuild-composer
 %systemd_postun_with_restart osbuild-worker@.service osbuild-remote-worker@.service
 
 %changelog
-* Sun Dec 1 11:00:00 CEST 2019 Ondrej Budai <obudai@redhat.com> - 4-1
-- New upstream release.
-* Fri Nov 15 11:00:00 CEST 2019 Ondrej Budai <obudai@redhat.com> - 3-1
-- New upstream release.
-* Wed Nov 13 15:14:00 CEST 2019 Ondrej Budai <obudai@redhat.com> - 2-2
-- Fix specfile according to packaging guidelines.
-* Mon Nov 11 13:23:00 CEST 2019 Tom Gundersen <teg@jklm.no> - 2-1
-- First release.
-
+# the changelog is distribution-specific, therefore it doesn't make sense to have it upstream


msehnout:
Just wondering, now that we don't use Packit, do we have a common workflow for doing Fedora updates? Are the changes like #340 going to synchornized by hand?

teg:
Ideally we would just not have a changelog, but I think we are required to. How do you imagine the workflow should to do a release after this change?

ondrejbudai:
Last time I did it manually and it was OK except for changelog. Merging the upstream and downstream changelogs isn't fun, because the downstream one may contain some other records like mass rebuilds. Also, if we want to have more detailed changelog in the repository, I think it doesn't make sense to have it in spec file. Instead, we should have in a separate file and use it to generate the spec file changelog.

Ideally, we should automate the process somehow, packit can do something in this field, but I need to investigate what.

larskarlitski:
> Also, if we want to have more detailed changelog in the repository, [...]

Dropping this here for reference: https://github.com/osbuild/osbuild/pull/280. I really like @dvdhrm's approach.

msehnout:
> Ideally we would just not have a changelog, but I think we are required to. How do you imagine the workflow should to do a release after this change?

I image it like this:
 1. manually (or using a script) sync the spec file = take this + append what is in the repo
 2. proceed with `fedpkg` workflow as usual

ondrejbudai:
> take this + append what is in the repo

can you explain more in detail? I don't know what is "this" and "what is in the repo".

msehnout:
> > take this + append what is in the repo
> 
> can you explain more in detail? I don't know what is "this" and "what is in the repo".

specfile from the upstream repo and changelog from the specfile in the downstream repo. We need to append the changelog at some point, right?

msehnout:
> > Also, if we want to have more detailed changelog in the repository, [...]
> 
> Dropping this here for reference: [osbuild/osbuild#280](https://github.com/osbuild/osbuild/pull/280). I really like @dvdhrm's approach.

Do we want to describe also the prefered downstream packaging process? I don't remember it from the top of my head, but sth. like `fedpkg new-sources`, `rpm-bumpspec`, `fedpkg build`, etc. Or we could have it in the downstream repo since we maintain it by hand now.

ondrejbudai:
Yeah, that's exactly what I would like to achieve with this PR:

on update:
dist-git spec = src-git spec + new release changelog + "old" changelog from dist-git spec

msehnout:
In that case I'd write a RELEASE.md or Makefile target, if you like a lot of `@echo`s :-). But that can be part of another PR.

msehnout:
I'm happy to approve this and merge it, I would just like others to agree explicitly. Is it ok like this? Should I create a follow-up issue with recommendations on how to use it?

PR #347, opened 2020-03-18
store/jobqueue: remove `distro` field from jobs

larskarlitski:
A job's purpose is to build an osbuild manifest and upload the results
somewhere. It should not know about which distro was used to generate
the pipeline.

Workers depended on the distro package in two ways:

1. To set an osbuild `--build-env`. This is not necessary anymore in new
   versions of osbuild. More importantly, it was wrong: it passed the
   runner from the distro that is being built, instead of one that
   matches the host.

   This patch simply removes that logic.

2. To fetch the output filename with `Distro.FilenameFromType()`. While
   that is useful, I don't think it warrants the dependency.

   This patch uses the fact that all current pipelines output exactly
   one file and uploads that. This should probably be extended in the
   future to upload all output files, or to name them explicitly in the
   upload target.

The worker should now compile to a smaller binary and do less
unnecessary work on startup (like reading repository files).

### DIFF
--- a/internal/jobqueue/api.go
+++ b/internal/jobqueue/api.go
@@ -105,7 +105,6 @@ func (api *API) addJobHandler(writer http.ResponseWriter, request *http.Request,
 	_ = json.NewEncoder(writer).Encode(Job{
 		ID:           nextJob.ComposeID,
 		ImageBuildID: nextJob.ImageBuildID,
-		Distro:       nextJob.Distro,
 		Manifest:     nextJob.Manifest,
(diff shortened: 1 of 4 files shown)


larskarlitski:
Force-puhshed to resolve conflicts.

larskarlitski:
> For a follow up: I think we should pass in the filename rather than expect it to be only one. Alternatively we could make the filename predictable based on the compose ID or something like that.

Agreed. Let's discuss this here: https://github.com/osbuild/osbuild-composer/issues/348

Issue #312, opened 2020-03-11
Rewrite osbuild-tests to use the testing module syntax 


atodorov:
See #301 for example & discussion.  Also see #309 for possible future changes which require the use of the new syntax. 

Benefits in short: 
- automatic test discovery, running & reporting
- machine readable test output with possibility to feed that to additional tools
- common syntax (available in most resources about testing in Go)
- will make some code unnecessary (main function, run function, etc)

teg:
Sounds good to me!

atodorov:
@teg great. @jikortus, @jrusz and myself can work on this. We'll coordinate between ourselves and ping you if we have questions.

teg:
Perfect! Thanks!

PR #352, opened 2020-03-18
Use testing module syntax and asserts in osbuild-tests

jikortus:
This is a WIP (or RFC :) commit covers issue #312 - aiming at using the integrated golang test functionality, together with using asserts instead of loggin functionality for errors. I'd appreciate any comments, as I guess this can be done better, although it looks fine from functional point of view. As this PR is in a WIP state right now, I didn't remove parts of the old code, but just left them commented out for a quick reference, and will remove them once the code is polished.

One of the possible downsides I'm aware of is that I'm not sure if there is any simpler/more elegant way to pass the `*testing.T` struct to all of the subsequently called functions or if I should take some different direction.

Another potential issue I dealt with is when I replaced some condition blocks that logged some text together with content of `err.Error()` when an error occurred, with 'relevant' asserts. However, I didn't manage to find a way to implement the functionality in the exactly same way as before, as in this case I can't use the `err.Error()` value in the formatted assert message anymore, since the test segfaults as there is no existing error when everything behaves fine. Therefore I had to use just `err` together with `%v` in the formatted text, which is different from what we had before. Is there a better way to do that, or my implementation is sufficient for such a use?

Many thanks for all suggestions and comments!


atodorov:
Commit log message:
> This covers issue #312.

-> Fixes #312 and GH will close the related issue automatically upon merge.

cmd/osbuild-tests/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// Smoke tests, until functionality tested fully
+	// that the calls succeed and return valid output
+	runComposerCLI(t, false, "compose", "types")
+	runComposerCLI(t, false, "compose", "status")
+	runComposerCLI(t, false, "compose", "list")
+	runComposerCLI(t, false, "compose", "list", "waiting")
+	runComposerCLI(t, false, "compose", "list", "running")
+	runComposerCLI(t, false, "compose", "list", "finished")
+	runComposerCLI(t, false, "compose", "list", "failed")
+	// runCommand(false, "compose", "log", UUID, [<SIZE>])
```
atodorov: Ideally these will be split into several testing functions, one for each command. In case there are more scenarios for sub-commands then we can split again.  However this can be done as a follow up.

cmd/osbuild-tests/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// runCommand(false, "sources", "delete" SOURCES.TOML)
+	runComposerCLI(t, false, "status", "show")
+
+	// Full integration tests
+	testCompose(t, "ami")
+}
+
+func testCompose(t *testing.T, outputType string) {
+	tmpdir := NewTemporaryWorkDir(t, "osbuild-tests-")
+	log.Println(tmpdir.Path)
```
atodorov: I don't think we need logging here, at least other PRs have removed logging lines when converting the syntax.

cmd/osbuild-tests/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+		Name:        "empty",
+		Description: "Test empty blueprint in toml format",
+	}
+	pushBlueprint(t, &bp)
+	defer deleteBlueprint(t, &bp)
+
+	runComposerCLI(t, false, "blueprints", "save", "empty")
+	//if _, err := os.Stat("empty.toml"); os.IsNotExist(err) {
+	//log.Fatalf("empty.toml doesn not exist: %v", err)
+	//}
```
atodorov: Remove commented out lines

cmd/osbuild-tests/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	_, err := os.Stat("empty.toml")
+	require.Falsef(t, os.IsNotExist(err), "empty.toml doesn not exist: %v", err)
```
atodorov: I think this should abort on any kind of error, can be permissions not only file doesn't exist. If I'm not mistaken we should require.Nil instead, no ?

cmd/osbuild-tests/main_test.go
jikortus: Could you please elaborate here? I'm not sure I've understood the suggestion correctly - if that means replacing `runComposerCLI(t, x, y, z)` with new specialized functions, it would seem more difficult to maintain and understand. Is it so or you meant something different? :)

cmd/osbuild-tests/main_test.go
jikortus: OK, all occurrences of logging functions have been removed.

cmd/osbuild-tests/main_test.go
jikortus: Good point, I agree. I changed that to `require.Nilf` and modified the error message to be more generic.

cmd/osbuild-tests/main_test.go
atodorov: It means not calling `runComposerCLI` for all possible commands/sub-commands inside a single function.  Some of the commented out lines will need their own setup and makes sense for them to be separated. 

However this is not related to the current changes and can be done as a follow up.

atodorov:
@jikortus I'm not sure if you did `git mv` when renaming the file or GitHub isn't handling that correctly. On some PRs the renames are shown nicely but here GH is showing the changes as deleted + newly added files. Makes it harder to compare.

/end rant

cmd/osbuild-tests/main_test.go
jikortus: OK, we can discuss that separately, ideally together with some examples of how it could look like (although I got the idea).

jikortus: I created issue #356 to discuss this further.

jikortus:
Sorry for the mess with file renaming - indeed I used `git mv` for renaming, but either it's a GitHub problem or something else went wrong in the process...

PR #357, opened 2020-03-18
distro: don't expose repositories in distro objects

### LINKED ISSUE #341: RCM API includes official repositories in pipelines
When running a compose through the RCM API, the official repositories for a distro are used in addition to the ones supplied by the API.

This is wrong: the semantics of the API is that only those repos that are provided are used. The resulting image might include packages from the wrong repos.

larskarlitski:
Mixing the way to build a distribution with where to get the source
packages from is wrong: it breaks pre-release repos, local mirrors, and
other use cases. To accommodate those, we introduced
`/etc/osbuild-composer/repositories`.

However, that doesn't work for the RCM API, which receives repository
URLs to use from outside requests. This API has been wrongly using the
`additionalRepos` parameter to inject those repos. That's broken,
because the resulting manifests contained both the installed repos and
the repos from the request.

To fix this, stop exposing repositories from the distros, but require
passing them on every call to `Manifest()`. This makes `additionalRepos`
redundant.

Fixes #341

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -84,7 +84,7 @@ func main() {
 
 	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"))
 
-	distros, err := distro.NewDefaultRegistry([]string{"/etc/osbuild-composer", "/usr/share/osbuild-composer"})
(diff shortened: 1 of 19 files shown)


teg:
This is perfect!. See many more things to clean up as a result! So nice!

PR #360, opened 2020-03-18
Add script to generate multiple test cases

jkozol:
This script generates json test cases for each supported output format. It requires an architecure, distro, osbuild store, and output directory as input arguments. It formats and passes this info to the generate-test-case (previously generate-test-cases) script which creates the individual test cases. Then this script writes these json test cases to files in the specified output directory.


tools/generate-test-case
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/python3
```
teg: I think you did a `git add` rather than `git mv`, which makes both the patches a bit hard to follow. If you first do the rename, then do a second patch to add a boot type to this script (or the other way around), and then add the new script in the final patch, I think it will be clearer.
Very happy about the actual change here though. Love the in-line documentation.

tools/generate-test-cases
```diff
@@ -3,84 +3,75 @@
...
-    return test_case
 
+def main(distro, arch, store, output):
+    test_case = []
+    for output_format, boot in format_boot_dict.items():
+        test_case_request = {
+            "boot": {
+                "type": boot
+            },
+            "compose-request": {
```
teg: I don't think this works. i think you need a map from image type to each test_case_request, as the blueprints will be different. AMI uses an empty blueprint, the others have the ssh key, and some have the openssh-server package explicitly enabled. Running the script should not change the actual test cases, but this does, right? (could add some new fields to rpmmd, but that's not crucial as they are unused).
Overall this patch looks great though!

tools/generate-test-cases
```diff
@@ -3,84 +3,75 @@
...
+            json.dump(test_case, case_file, indent=2)
+    return
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="Generate test cases")
-    parser.add_argument("store", metavar="DIRECTORY", type=os.path.abspath, help="path to the osbuild store")
+    parser.add_argument("--distro", help="distribution for test cases", required=True)
+    parser.add_argument("--arch", help="architecture for test cases", required=True)
+    parser.add_argument("--store", metavar="STORE_DIRECTORY", type=os.path.abspath, help="path to the osbuild store", required=True)
+    parser.add_argument("--output", metavar="OUTPUT_DIRECTORY", type=os.path.abspath, help="path to the osbuild store", required=True)
```
ondrejbudai: `help` parameter seems to be copy-pasted from the previous line

tools/generate-test-case
```diff
@@ -0,0 +1,70 @@
...
+                    subprocess.run(["xz", "--decompress", "--stdout", image_file], stdout=f)
+                image_file = fn
+        test_case["image-info"] = json.loads(subprocess.check_output(["tools/image-info", image_file]))
+
+    return test_case
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Generate test cases")
+    parser.add_argument("store", metavar="DIRECTORY", type=os.path.abspath, help="path to the osbuild store")
```
ondrejbudai: I suggest adding `required=True` for the sake of consistency.

tools/generate-test-case
jkozol: Improved the test case

tools/generate-test-cases
jkozol: I now only use a single mapping between the test case request and the output type. The test case requests are located in the format-request-map.json.

tools/generate-test-cases
jkozol: Thanks for pointing this out. I have updated the help text.

tools/generate-test-case
jkozol: Since store is a positional argument it is required regardless of having required=True. For consistency, it could be nice to have the store assigned to a flag `--store` as it is in generate-test-cases but since the store is the only argument here I think it is fine as a positional argument.

PR #361, opened 2020-03-19
spec: fix building of the tests

ondrejbudai:
Successful Koji builds:
- f31 https://koji.fedoraproject.org/koji/taskinfo?taskID=42617537
- f32 https://koji.fedoraproject.org/koji/taskinfo?taskID=42617539
- f33 https://koji.fedoraproject.org/koji/taskinfo?taskID=42617541

Prior this commit building of the RPM package was failing in mock/koji.
The reason for the failure was that we hadn't set correctly the GOPATH and
GO111MODULE environment variables for the `go test -c` command.

This commit takes a whole new approach to the building of tests. There's now
%gobuildtest macro which should be as close as possible to %gobuild macro.
The only change is that it doesn't call `go build`, but `go test -c`.
Additionally, this approach should make the test building prone to changes
in go-rpm-macros.

Also, see the comments above the %gobuildtest macro.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -26,14 +26,15 @@ BuildRequires:  systemd
 BuildRequires:  systemd-rpm-macros
 BuildRequires:  git
 BuildRequires:  golang(github.com/aws/aws-sdk-go)
-BuildRequires:  golang-github-azure-storage-blob-devel
+BuildRequires:  golang(github.com/Azure/azure-storage-blob-go/azblob)
 BuildRequires:  golang(github.com/BurntSushi/toml)
 BuildRequires:  golang(github.com/coreos/go-semver/semver)
 BuildRequires:  golang(github.com/coreos/go-systemd/activation)
-BuildRequires:  golang(github.com/google/uuid)
-BuildRequires:  golang(github.com/julienschmidt/httprouter)
 BuildRequires:  golang(github.com/gobwas/glob)
 BuildRequires:  golang(github.com/google/go-cmp/cmp)
+BuildRequires:  golang(github.com/google/uuid)
+BuildRequires:  golang(github.com/julienschmidt/httprouter)
+BuildRequires:  golang(github.com/stretchr/testify/assert)
 %endif
 
 Requires: golang-github-osbuild-composer-worker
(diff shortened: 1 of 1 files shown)


golang-github-osbuild-composer.spec
```diff
@@ -68,16 +69,30 @@ export GOFLAGS=-mod=vendor
...
+# `go test -c` has very similar to arguments to `go build`. The following
+# macro uses that fact. It takes gobuild macro and substitutes `go build`
+# with `go test -c` inside the macro. This way we get working `go test -c`
+# command which should be in sync with changes in gobuild macro.
+#
+# Because gobuild macro gets expanded to a string containing apostrophes
+# and shell expansions (which we don't want to expand just yet), heredoc
+# is used to pipe the macro into sed. The apostrophes around HEREDOC
+# put it into non-expansion mode.
+%define gobuildtest(o:) %(sed "s/go build/go test -c/" <<'HEREDOC'
```
msehnout: Oh, this is so awesome!

golang-github-osbuild-composer.spec
```diff
@@ -68,16 +69,30 @@ export GOFLAGS=-mod=vendor
...
-
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-weldr-tests %{goipath}/internal/weldrcheck/
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-rcm-tests %{goipath}/cmd/osbuild-rcm-tests
+# The tests need to be compiled by `go test -c`.
+# Go rpm macros have gotest, but it can be used only in check to run
+# unit tests during rpmbuild.
+#
+# `go test -c` has very similar to arguments to `go build`. The following
```
atodorov: Minor typo. Drop the first occurrence of `to`. Otherwise looks awesome.

golang-github-osbuild-composer.spec
larskarlitski: **very** impressive hack, but it is quite a hack. For example, we're passing command line options to `go test` that are documented for `go build` only.

But wow, it is quite something :smiley:

ondrejbudai: From `go help test`:

> In addition to the build flags, the flags handled by 'go test' itself are:
> ...

ondrejbudai: Also, from the first line of `go help test`:

> usage: go test [build/test flags] [packages] [build/test flags & test binary flags]

atodorov: This is also how I figured out the flags in my original attempts at this.

larskarlitski: Ahhh I overread that. Thanks!

larskarlitski:
I've proposed an alternative suggestion here: https://github.com/osbuild/osbuild-composer/pull/364. My version (more or less) does what the `%gobuild` macro does internally, while @ondrejbudai's here tries to use that macro as a function with a small hack to make the actual command being run a parameter. That is pretty cool, and means that we are more robust against that macro changing. I'm not sure though...

ondrejbudai:
FYI we decided to apply this patch downstream. The discussion for a proper upstream patch was postponed to the next week.

ondrejbudai:
Let's go with #364, this fix has issues on rhel. (`BUILDTAGS` env variable doesn't exist there and it would be hard to set the +integration build tag)

PR #364, opened 2020-03-19
Fix building the tests in mock/koji

larskarlitski:
Explanation in the commits.

Phew.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -72,6 +72,13 @@ export GOFLAGS=-mod=vendor
 # Build test binaries with `go test -c`, so that they can take advantage of
 # golang's testing package. The golang rpm macros don't support building them
 # directly. Thus, do it manually, taking care to also include a build id.
+#
+# On Fedora, also turn off go modules and set the path to the one into which
+# the golang-* packages install source code.
+%if 0%{?fedora}
+export GO111MODULE=off
+export GOPATH=%{gobuilddir}:%{gopath}
+%endif
 
 TEST_LDFLAGS="${LDFLAGS:-} -B 0x$(od -N 20 -An -tx1 -w100 /dev/urandom | tr -d ' ')"
 
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -82,6 +82,13 @@ export GOFLAGS=-mod=vendor
 # Build test binaries with `go test -c`, so that they can take advantage of
 # golang's testing package. The golang rpm macros don't support building them
 # directly. Thus, do it manually, taking care to also include a build id.
+#
+# On Fedora, also turn off go modules and set the path to the one into which
+# the golang-* packages install source code.
+%if 0%{?fedora}
+export GO111MODULE=off
+export GOPATH=%{gobuilddir}:%{gopath}
+%endif
 
 TEST_LDFLAGS="${LDFLAGS:-} -B 0x$(od -N 20 -An -tx1 -w100 /dev/urandom | tr -d ' ')"
 


ondrejbudai:
Nice! It's cool to see a simple, not-so-hacky solution. However, I have some concerns. It will require some research from my side, I will post them tommorow.

larskarlitski:
Note: the travis CI run is green, but somehow didn't report that to GitHub.

golang-github-osbuild-composer.spec
```diff
@@ -26,14 +26,15 @@ BuildRequires:  systemd
 BuildRequires:  systemd-rpm-macros
 BuildRequires:  git
 BuildRequires:  golang(github.com/aws/aws-sdk-go)
-BuildRequires:  golang-github-azure-storage-blob-devel
+BuildRequires:  golang(github.com/Azure/azure-storage-blob-go/azblob)
```
teg: What's actually going on here? This is different from the other `golang()` requires, no?

golang-github-osbuild-composer.spec
```diff
@@ -26,14 +26,15 @@ BuildRequires:  systemd
...
 BuildRequires:  golang(github.com/BurntSushi/toml)
 BuildRequires:  golang(github.com/coreos/go-semver/semver)
 BuildRequires:  golang(github.com/coreos/go-systemd/activation)
 BuildRequires:  golang(github.com/google/uuid)
 BuildRequires:  golang(github.com/julienschmidt/httprouter)
 BuildRequires:  golang(github.com/gobwas/glob)
 BuildRequires:  golang(github.com/google/go-cmp/cmp)
+BuildRequires:  golang(github.com/stretchr/testify/assert)
```
teg: How come this wasn't caught in our tests, this is down to modules being enabled before this PR? Will our tests catch stuff like this in the future, once this has been merged?

golang-github-osbuild-composer.spec
larskarlitski: These golang packages simple include a `Provide` that looks like this. That one package does it slightly differently. I think it's still worth the consistency. I've added an explanation to the commit message.

golang-github-osbuild-composer.spec
larskarlitski: It was only used by test binaries, which were built with `go test` (prior to this patch). `go test` simply downloaded the missing library, because we didn't turn off module support. This patch does that and we should find these kinds of issues in the future.

Probably still better to move to building in `mock`, to be sure.

teg:
Thanks for the updates, answered my questions

ondrejbudai:
I don't know if I like the idea of not using the same set of flags as `%gobuild` uses. `%gobuild` definition itself is a mess, for most platforms it looks like this:

```
%{?gobuilddir:GOPATH="%{gobuilddir}:${GOPATH:+${GOPATH}:}%{?gopath}"} GO111MODULE=off \\
go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-}%{?currentgoldflags} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags %{?__golang_extldflags}'" -a -v -x %{?**};
```

However, for ppc64, the `-buildmode pie` is omitted. There is even a whole different definition for mips, which uses a different `-compiler`, but let's not consider it, because it isn't a primary architecture for Fedora.

Let's run `rpmspec -P` on our spec file to see what `%gobuild` is expanded into:

```
  GOPATH="/home/obudai/rpmbuild/BUILD/osbuild-composer-8/_build:${GOPATH:+${GOPATH}:}/usr/share/gocode" GO111MODULE=off \
  go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-}-X github.com/osbuild/osbuild-composer/version=8 -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n') -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v -x -o _bin/osbuild-composer github.com/osbuild/osbuild-composer/cmd/osbuild-composer;
```

Yeah, a thing of beauty. Compare it with the addition in this PR (also expanded):

```
GO111MODULE=off \
  GOPATH=/home/obudai/rpmbuild/BUILD/osbuild-composer-8/_build:/usr/share/gocode \
  go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-dnf-json-tests github.com/osbuild/osbuild-composer/cmd/osbuild-dnf-json-tests
```

There are a lot of differences between our take on `-ldflags` and the official one. I'm a bit worried by using our own set, because I don't understand the differences. Also, if the authors of go-rpm-macros decide to change something, we won't probably notice it and we will end up using our old set which can have some issues by then.

I know that my proposal looks scary, but I'm happy with the fact that it uses the same flag set as `%gobuild`. I don't know how the flags were chosen, but I'm alright with it, because I trust the people who did the selection for us. Also, in the future it should stay in sync with any changes in `%gobuild`. Feel free to correct me.

Finally, I think I'm ok with merging both PRs, because both are solving the issue. However, I think shipping the tests built by the specfile from **this** proposal to Fedora is questionable, because if someone asks why we did these changes to`-ldflags`, our answer is basically: "We don't know".

golang-github-osbuild-composer.spec
```diff
@@ -73,11 +75,17 @@ export GOFLAGS=-mod=vendor
 
 TEST_LDFLAGS="${LDFLAGS:-} -B 0x$(od -N 20 -An -tx1 -w100 /dev/urandom | tr -d ' ')"
```
ondrejbudai: I think there should be `export`, otherwise this line has no effect.

golang-github-osbuild-composer.spec
```diff
@@ -73,11 +75,17 @@ export GOFLAGS=-mod=vendor
...
 
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-tests %{goipath}/cmd/osbuild-tests
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-dnf-json-tests %{goipath}/cmd/osbuild-dnf-json-tests
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-weldr-tests %{goipath}/internal/weldrcheck/
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-rcm-tests %{goipath}/cmd/osbuild-rcm-tests
-go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
+%define buildtest(o:) %{expand:
+  GO111MODULE=off \\
+  GOPATH=%{gobuilddir}:%{gopath} \\
+  go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" %{-o} %{?*}
```
msehnout: As idea: Should this use the `GOFLAGS` variable defined on line 66?

msehnout:
Does it build with RHEL buildroot?

golang-github-osbuild-composer.spec
larskarlitski: No, this works (and has been working on master), because this variable is only used locally in this script.

golang-github-osbuild-composer.spec
larskarlitski: Yeah, but we probably need to change more to make this work on RHEL.

golang-github-osbuild-composer.spec
ondrejbudai: Right, sorry, I got confused.

ondrejbudai:
On RHEL it fails with `go: GOPATH entry is relative; must be absolute path: "%{gobuilddir}".` Interesting...

larskarlitski:
> I don't know if I like the idea of not using the same set of flags as `%gobuild` uses. `%gobuild` definition itself is a mess, for most platforms it looks like this:
> 
> [...]

Thanks for diving deep on this!

I had included the same flags (well, who knows if I got all of them) in an early version of https://github.com/osbuild/osbuild-composer/pull/301, but we decided on keeping it simple. I agree that that isn't right after all. If we accept this approach, we should copy the `%rpmbuild` macro as closely as possible.

> I know that my proposal looks scary, but I'm happy with the fact that it uses the same flag set as `%gobuild`. I don't know how the flags were chosen, but I'm alright with it, because I trust the people who did the selection for us. Also, in the future it should stay in sync with any changes in `%gobuild`. Feel free to correct me.
>
> Finally, I think I'm ok with merging both PRs, because both are solving the issue. However, I think shipping the tests built by the specfile from **this** proposal to Fedora is questionable, because if someone asks why we did these changes to`-ldflags`, our answer is basically: "We don't know".

Yeah. Like I said on your PR, it comes down to taste: do we want to duplicate the macros (and use internal flags) in our spec file, or do we want your (brilliant) small hack which uses the upstream one directly. Both have pros and cons.

I'm happy to leave the choice to you. Let me know if you want this one—then I'll fix it up to include the right flags and make it work on RHEL.

atodorov:
For what is worth I prefer the solution in #361. It is indeed a hack but is well documented and follows upstream more closely. 

Probably the right solution will be to work with go-rpm-macros upstream and clean up the mess, but I am not qualified to comment on that.

larskarlitski:
Update to only make the change on Fedora. Tested in `fedora-31-x86_64.cfg` and `rhel-8-x86_64` mocks.

I've also simplified a bit. Redefining the macro for RHEL was too cumbersome.

larskarlitski:
Travis is green, but forgot mentioning that again.

PR #366, opened 2020-03-19
distro: require passing a filename in Manifest()

### LINKED ISSUE #348: Local target only supports uploading one file

3. Upload all files

larskarlitski:
1b7cb6c11 changed workers so that they don't have access to the distro
that was used to create a manifest anymore. This means that they also
don't know what the filename of the final image in the store is. That
commit worked around this by picking the only file that resulted in that
build. That was a bit of a hack.

Really, image filenames are a user-facing concept. Internally, we're
better of fixing the name to something well-known.

Add a filename parameter to Distro.Manifest() that is independent of the
output type. Set it to "image" from weldr, osbuild-pipeline, and the rcm
API.

Use that name in the worker as well.

Fixes #348

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -144,7 +144,7 @@ func main() {
 		}
 	} else {
 		size := d.GetSizeForOutputType(imageType, 0)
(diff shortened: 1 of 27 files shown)


larskarlitski:
This turned out making things more complex that I'd like. Also, the `Target` struct has an `ImageName` field already. Embarrassing. Closing in favor of https://github.com/osbuild/osbuild-composer/pull/370.

PR #367, opened 2020-03-20
Add integration tests for projects/source

bcl:


internal/weldrcheck/blueprints_test.go
```diff
@@ -33,12 +33,29 @@ var testState *TestState
...
+		fmt.Printf("ERROR: Test repo setup failed: %s\n", err)
+		os.Exit(1)
+	}
+	testState.repoDir = dir
+
+	// Run the tests
+	rc := m.Run()
+
+	// Cleanup after the tests
+	err = tearDownTemporaryRepository(dir)
```
atodorov: Wouldn't calling via defer this right after set up better?

internal/weldrcheck/source_test.go
```diff
@@ -0,0 +1,190 @@
...
+	source := `{
+		"name": "package-repo-json-v0",
+		"url": "file://REPO-PATH",
+		"type": "yum-baseurl",
+		"proxy": "https://proxy-url/",
+		"check_ssl": true,
+		"check_gpg": true,
+		"gpgkey_urls": ["https://url/path/to/gpg-key"]
+	}`
+	source = strings.Replace(source, "REPO-PATH", testState.repoDir, 1)
```
atodorov: Is `testState` a global variable coming from another file ? I didn't know you can do this.

internal/weldrcheck/utils.go
```diff
@@ -64,3 +69,26 @@ func setupTestState(socketPath string, timeout time.Duration) (*TestState, error
...
+	if err != nil {
+		return "", err
+	}
+	return dir, nil
+}
+
+// Remove the temporary repository
+func tearDownTemporaryRepository(dir string) error {
+	return os.RemoveAll(dir)
+}
```
atodorov: Same functions in `osbuild-dnf-json-tests/main_test.go`. Makes sense to have them in utils.go and then reuse them.

internal/weldrcheck/blueprints_test.go
bcl: That's what I thought originally :) But since it uses os.Exit at the end the defer isn't run so you end up with the temporary repos left behind. I've tested this with success and failure in the tests and it always gets cleaned up.

internal/weldrcheck/source_test.go
bcl: Yes, all of this code is in the weldrcheck package, so it shares global variables. It is setup by the  TestMain function in blueprints_test.go (it may make sense to move that block of code and the global into a separate file soon, but I'll save that for later).

internal/weldrcheck/utils.go
bcl: Right, I was going to look for a good common location for this. internal/test/helpers.go might be the right place.

bcl:
> @larskarlitski we may want to make more use of these weldr-specific serialization types and avoid reusing internal types for the public API, as we just discussed.

My plan is to eventually convert the api code to use them. But for now they are mostly duplicates to make it easier to merge/review.

PR #370, opened 2020-03-20
job: get image name from target

### LINKED ISSUE #348: Local target only supports uploading one file

larskarlitski:
1b7cb6c changed workers so that they don't have access to the distro
that was used to create a manifest anymore. I thought this meant that
they also don't know what the filename of the final image in the store
is. That commit worked around this by picking the only file that
resulted in that build. That was a bit of a hack.

Turns out, the `Target` struct already has an ImageName field.

Fixes #348

### DIFF
--- a/internal/jobqueue/job.go
+++ b/internal/jobqueue/job.go
@@ -97,23 +97,7 @@ func (job *Job) Run(uploader LocalTargetUploader) (*common.ComposeResult, error)
 	for _, t := range job.Targets {
 		switch options := t.Options.(type) {
 		case *target.LocalTargetOptions:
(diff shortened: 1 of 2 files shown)


larskarlitski:
Update: that ImageName was not actually set for the default local target.

teg:
The `ImageName` field is not the name of the file, but the name the caller of the weldr API wants the image to have in the target cloud provider. I filed an alternative PR #374

larskarlitski:
> The `ImageName` field is not the name of the file, but the name the caller of the weldr API wants the image to have in the target cloud provider. I filed an alternative PR #374

I think you're right. Thanks for catching that!

PR #374, opened 2020-03-21
Specify image filenames in upload targets

### LINKED ISSUE #348: Local target only supports uploading one file

teg:
Drop any assumptions about the name or number of output files that osbuild produces from the worker.

Inspired by #370, fixes #348.

### DIFF
--- a/internal/weldr/upload.go
+++ b/internal/weldr/upload.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/distro"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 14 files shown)


internal/weldr/upload.go
```diff
@@ -99,32 +118,26 @@ func TargetsToUploadResponses(targets []*target.Target) []UploadResponse {
 	return uploads
 }
 
-func UploadRequestToTarget(u UploadRequest, filename string) (*target.Target, error) {
```
larskarlitski: Hm, you added the parameter in the previous commit, just to remove it again here. It is never used.

teg: Ah, a rebase gone awry. Thanks! Fixed and pushed two natural follow-ups on top.

PR #375, opened 2020-03-22
Distro interfaces

teg:
Introduce new interfaces to represent a distro at a given architecture, and each image type for a architecture/distro pair. Methods on the ImageType object are used rather than passing in architecture and image type to methods on the distro objects, simplifying the code and making the need for much of the `common` package redundant.

As discussed with @msehnout and @larskarlitski .

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -10,6 +10,11 @@ import (
 	"path"
 
 	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora31"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora32"
+	"github.com/osbuild/osbuild-composer/internal/distro/rhel81"
+	"github.com/osbuild/osbuild-composer/internal/distro/rhel82"
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/distro"
@@ -74,7 +79,7 @@ func main() {
 		}
 	}
 
-	distros, err := distro.NewDefaultRegistry()
+	distros, err := distro.NewRegistry(fedora30.New(), fedora31.New(), fedora32.New(), rhel81.New(), rhel82.New())
 	if err != nil {
 		panic(err)
 	}
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path"
 
-	"github.com/osbuild/osbuild-composer/internal/common"
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora31"
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora32"
@@ -28,12 +27,12 @@ type rpmMD struct {
 }
 
 func main() {
-	var imageType string
+	var imageTypeArg string
 	var blueprintArg string
 	var archArg string
 	var distroArg string
 	var rpmmdArg bool
-	flag.StringVar(&imageType, "image-type", "", "image type, e.g. qcow2 or ami")
+	flag.StringVar(&imageTypeArg, "image-type", "", "image type, e.g. qcow2 or ami")
 	flag.StringVar(&archArg, "arch", "", "architecture to create image for, e.g. x86_64")
 	flag.StringVar(&distroArg, "distro", "", "distribution to create, e.g. fedora-30")
 	flag.BoolVar(&rpmmdArg, "rpmmd", false, "output rpmmd struct instead of pipeline manifest")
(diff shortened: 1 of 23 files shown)


internal/rcm/api.go
```diff
@@ -183,17 +171,36 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
...
-	arch, ok := composeRequest.Architectures[0].ToString()
-	if !ok {
-		panic("architecture was validated before")
+	arch, err := distro.GetArch(composeRequest.Architectures[0])
+	if err != nil {
+		writer.WriteHeader(http.StatusBadRequest)
+		_, err := writer.Write([]byte("unknown architecture for distro"))
+		if err != nil {
+			panic("Failed to write response")
+		}
```
msehnout: Shouldn't this function `return` here?

internal/rcm/api.go
```diff
@@ -183,17 +171,36 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
...
+	}
+
+	imageType, err := arch.GetImageType(composeRequest.ImageTypes[0])
+	if err != nil {
+		writer.WriteHeader(http.StatusBadRequest)
+		_, err := writer.Write([]byte("unknown image type for distro and architecture"))
+		if err != nil {
+			panic("Failed to write response")
+		}
```
msehnout: And here?

internal/store/store.go
```diff
@@ -560,11 +560,11 @@ func (s *Store) getImageBuildDirectory(composeID uuid.UUID, imageBuildID int) st
 	return fmt.Sprintf("%s/%d", s.getComposeDirectory(composeID), imageBuildID)
 }
 
-func (s *Store) PushCompose(distro distro.Distro, composeID uuid.UUID, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, arch, composeType string, size uint64, uploadTarget *target.Target) error {
```
ondrejbudai: The `distro` and `arch` parameters are no longer used, I suggest removing them in this PR.

internal/distro/fedora31/distro.go
```diff
@@ -44,7 +39,104 @@ type output struct {
...
+	arches        map[string]arch
+	outputs       map[string]output
+	buildPackages []string
+}
+
+type Fedora31Arch struct {
+	name   string
+	distro *Fedora31
+	arch   *arch
+}
```
larskarlitski: Having two structs representing architecture in each of the distros is weird. I'd merge them. Can be done in a follow-up, though.

internal/rcm/api.go
teg: Fixed. Thanks!

internal/rcm/api.go
teg: Yup, fixed it all. Thanks!

internal/distro/distro.go
```diff
@@ -54,6 +58,49 @@ type Distro interface {
...
+	// Returns the name of the architecture.
+	Name() string
+
+	// Returns a sorted list of the names of the image types this architecture
+	// supports.
+	ListImageTypes() []string
+
+	// Returns an object representing a given image format for this architecture,
+	// on this distro.
+	GetImageType(imageType string) (ImageType, error)
```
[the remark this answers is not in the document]
teg: I prefixed with `Get` for things that can return an error. For pure accessors I did not.

internal/distro/fedora31/distro.go
teg: Totally, these patches only add things. Have a massive cleanup PR to do on top.

internal/store/store.go
teg: I'd like to do clean-ups separately. This is massive as it is, but was intended to do the minimal amount of changes in order to replace the interfaces. Next I get to do the immensely satisfying task of deleting code. (But yes, I agree, these should be dropped from here).

cmd/osbuild-pipeline/main.go
```diff
@@ -93,6 +83,18 @@ func main() {
...
 
+	arch, err := d.GetArch(archArg)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "The provided architecture (%s) is not supported.: %s\n", archArg, err.Error())
+		return
+	}
+
+	imageType, err := arch.GetImageType(imageTypeArg)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "The provided image type (%s) is not supported.: %s\n", imageTypeArg, err.Error())
```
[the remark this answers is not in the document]
teg: Good point! Fixed!

internal/weldr/upload.go
```diff
@@ -118,14 +118,9 @@ func targetsToUploadResponses(targets []*target.Target) []uploadResponse {
 	return uploads
 }
 
-func uploadRequestToTarget(u uploadRequest, d distro.Distro, imageType string) (*target.Target, error) {
```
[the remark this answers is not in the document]
teg: Agreed, will do in follow-ups though.

teg:
Thanks for the reviews! I addressed most, but left some for future work, just because there is so much clean-ups that can be done anyway.

@msehnout I have a pending patch doing all the renames. I will also delete all the redundant code from each of the distro packages (there is now a lot), so shouldn't add much overall.

internal/distro/distro.go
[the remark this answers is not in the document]
msehnout: I'd love if you wrote such ideas into the CONTRIBUTING file. I think common naming would help a lot in code readability.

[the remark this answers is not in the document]
larskarlitski: I like this.

msehnout:
Travis is green, but forgot to tell Github. Merging it now so we can move forward.

internal/distro/distro.go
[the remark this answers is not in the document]
teg: I agree, we should write more of these things down. Maybe a hackfest activity unless we get to it before. In this case I didn't realise what I'd done before Lars asked ;)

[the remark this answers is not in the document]
msehnout: I'm pretty sure I'd forget it until then, but I'll leave this one up to you :-)

PR #376, opened 2020-03-23
Request specific architecture from dnf

msehnout:
The depsolving currently does not work cross-arch. Also when passing repos with multiple architectures it does not work but we already have the information about required architecture, so we can just pass it to dnf.

See individual commits for further explanation.

### DIFF
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -6,12 +6,18 @@ package main
 
 import (
 	"fmt"
-	"github.com/stretchr/testify/assert"
-	"path"
-	"testing"
-
+	"github.com/osbuild/osbuild-composer/internal/distro"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora31"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora32"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
 )
 
 func TestFetchChecksum(t *testing.T) {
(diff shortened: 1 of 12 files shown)


msehnout:
> Looking very good! Thanks for figuring this out. Maybe worth adding some tests for coss-arch builds to the dnf-json tests?

Good idea, will do :)

> I am not 100% sure what exactly the options you added do internally, this is the best reference I could finde: https://docs.fedoraproject.org/en-US/Fedora/22/html/System_Administrators_Guide/sec-Using_DNF_Variables.html.

I can add more documentation, maybe a comment to dnf-json would be useful?

> They also reference $releasever, is that something we also should set, or does that not have any effect? I suppose we should, if for no other reason than people potentially passing us repositories using these variables in the URLs. That's for a follow-up patch though. I only have one small nit.

I was wondering about that as well, so I went through the code and asked the upstream developers and it does not effect us, because it is only about the repositories dnf uses, but we set those explicitly.

I've created an issue for this and closed it myself: https://github.com/osbuild/osbuild-composer/issues/377

dnf-json
```diff
@@ -37,12 +37,15 @@ def dnfrepo(desc, parent_conf=None):
...
 
 
-def create_base(repos, module_platform_id, persistdir, cachedir):
+def create_base(repos, module_platform_id, persistdir, cachedir, arch):
     base = dnf.Base()
     base.conf.module_platform_id = module_platform_id
     base.conf.config_file_path = "/dev/null"
     base.conf.persistdir = persistdir
     base.conf.cachedir = cachedir
+    if arch:
```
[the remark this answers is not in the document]
msehnout: I was about to suggest the same. Thanks for the tip with ordering 👍

teg:
> I can add more documentation, maybe a comment to dnf-json would be useful?

If you have more info, then at least put it in the commit message, then people can find it if they need. If you think it makes sense, feel free to add comments in the code too of course.

internal/distro/fedora31/distro.go
```diff
@@ -149,6 +149,7 @@ func New() *Fedora31 {
 			"qemu-img",
 			"systemd",
 			"tar",
+			"xz",
```
teg: Hm, how come these tests passed before your PR? Something sounds off...

msehnout: I agree, but I haven't yet come with a good theory why is that.

msehnout: It's because `x86_64` has:
```
				BuildPackages: []string{
					"grub2-pc",
				},
```
where `grub2-pc` requires `dracut` which in turn requires `xz`. `aarch64` does not require anything that requires `xz` therefore the error.

teg: Thanks!

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -54,3 +59,42 @@ func TestFetchChecksum(t *testing.T) {
...
+	// TODO: add RHEL, but don't make it hard requirement because it will fail outside of VPN
+	for _, distroStruct := range []distro.Distro{fedora30.New(), fedora31.New(), fedora32.New()} {
+		repoConfig, err := rpmmd.LoadRepositories([]string{repositories}, distroStruct.Name())
+		assert.Nilf(t, err, "Failed to LoadRepositories from %v for %v: %v", repositories, distroStruct.Name(), err)
+		if err != nil {
+			// There is no point in running the tests without having repositories, but we can still run tests
+			// for the remaining distros
+			continue
+		}
+		// TODO: wait for @teg to implement GetArches() method and then loop over them, meanwhile at least hard code them
```
teg: I filed a PR for this just now. So you can rebase this and drop the TODO, I think.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -54,3 +59,42 @@ func TestFetchChecksum(t *testing.T) {
...
+		for _, archStr := range []string{"x86_64", "aarch64"} {
+			arch, err := distroStruct.GetArch(archStr)
+			assert.Nilf(t, err, "Failed to GetArch from %v structure: %v", distroStruct.Name(), err)
+			// TODO: loop over image types
```
teg: I'd add the loop in this patch, just because if ever we have a distro/arch without ami support (we will), then this will break :) Alternatively, drop the TODO, if you think this is good enough for now.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -54,3 +59,42 @@ func TestFetchChecksum(t *testing.T) {
...
+			imgType, err := arch.GetImageType("ami")
+			assert.Nilf(t, err, "Failed to GetImageType for %v on %v: %v", distroStruct.Name(), arch.Name(), err)
+
+			// TODO: check also base packages
```
teg: Same, drop TODO or add the check, I'd say (or is there a reason not to)?

msehnout: Nope, it wasn't my intention to submit it like this, but I have a bash script that downloads this branch and run the tests and I wanted to run it asap, so I just decided to push unfinished work. Sorry about that. It is finished now.

msehnout:
It should be ready now. I've rebased on top of the latest changes and I'm now manually running the tests on aarch64.

msehnout:
Running the test actually revealed more issues on aarch64, but those seems unrelated to this PR, so I'd merge this as is and continue in another PR.
```
    main_test.go:83:
        	Error Trace:	main_test.go:83
        	Error:      	Expected nil, but got: &rpmmd.DNFError{Kind:"MarkingErrors", Reason:"Error occurred when marking packages for installation: Problems in request:\nmissing packages: open-vm-tools"}
        	Test:       	TestCrossArchDepsolve
        	Messages:   	Failed to Depsolve base packages for fedora-31 vmdk aarch64 image: DNF error occured: MarkingErrors: Error occurred when marking packages for installation: Problems in request:
        	            	missing packages: open-vm-tools
```

cmd/osbuild-dnf-json-tests/main_test.go
teg: Ah, got it. No problem.

Issue #362, opened 2020-03-19
tests: increase coverage of blueprints tests + use asserts


jrusz:
I'd like to try increase the coverage of `internal/blueprint/blueprint.go` and `internal/blueprint/customizations.go` by expanding the already existing `internal/blueprints/blueprint_test.go`. Also I'd like to rewrite a few lines there to use `testify/asserts` and `testify/require`

atodorov:
@jrusz can you be more precise here - in particular which functions/code paths/test scebnarios would you like to cover if there are more than a handful ?  If there are sounds like a good opportunity to split.

If not, are you looking for help/guidance/advise or just stating your intentions with this issue ?

jrusz:
@atodorov to be honest this is the first time I am doing something like this so I am not sure what I'll accomplish here. I looked at the coveralls report and saw which functions are not covered and I want to try covering them by adding test functions. I only want to work with the functions in the files I mentioned. I'll consider splitting this if I find it overwhelming but right now I am just stating my intention and will update this issue if needed. Once I have something I'll create a PR and we can discuss it there.

teg:
Sounds reasonable to me. Happy to review the PR once you got started.

PR #379, opened 2020-03-23
Add coverage to internal/blueprints

jrusz:
I will be adding more commits to this PR. I'll be happy for any comments. This will fix https://github.com/osbuild/osbuild-composer/issues/362 once ready to merge.

### DIFF
--- a/internal/blueprint/blueprint_test.go
+++ b/internal/blueprint/blueprint_test.go
@@ -2,6 +2,7 @@ package blueprint
 
 import (
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
@@ -88,4 +89,21 @@ func TestBumpVersion(t *testing.T) {
 			t.Errorf("BumpVersion(%#v) is expected to return %#v, but instead returned %#v", c.OldVersion, c.ExpectedVersion, bp.Version)
 		}
 	}
+}
+
+func TestGetPackages(t *testing.T) {
+
+	bp := Blueprint{
+		Name:        "packages-test",
+		Description: "Testing GetPackages function",
+		Version:     "0.0.1",
+		Packages: []Package{
+			{Name: "tmux", Version: "1.2"}},
+		Modules: []Package{
+			{Name: "openssh-server", Version: "*"}},
+		Groups: []Group{
+			{Name: "anaconda-tools"}},
+	}
+	Received_packages := bp.GetPackages()
+	assert.ElementsMatch(t, []string{"tmux-1.2", "openssh-server", "@anaconda-tools"}, Received_packages)
 }
--- a/internal/blueprint/blueprint_test.go
+++ b/internal/blueprint/blueprint_test.go
@@ -1,8 +1,8 @@
 package blueprint
 
 import (
-	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"testing"
 )
 
@@ -18,24 +18,16 @@ func TestDeepCopy(t *testing.T) {
 	}
 
 	bpCopy, err := bpOrig.DeepCopy()
-	if err != nil {
-		t.Fatalf("Blueprint.DeepCopy failure: %s", err.Error())
-	}
-	if diff := cmp.Diff(bpOrig, bpCopy); diff != "" {
-		t.Fatalf("Blueprint.DeepCopy is different from original\ndiff: %s", diff)
-	}
+	require.NoError(t, err)
+	require.Equalf(t, bpOrig, bpCopy, "Blueprints.DeepCopy is different from original.")
 
 	// Modify the copy
 	bpCopy.Packages[0].Version = "1.2.3"
-	if bpOrig.Packages[0].Version != "*" {
-		t.Fatalf("Blueprint.DeepCopy failed, original modified")
-	}
+	require.Equalf(t, bpOrig.Packages[0].Version, "*", "Blueprint.DeepCopy failed, original modified")
 
 	// Modify the original
 	bpOrig.Packages[0].Version = "42.0"
-	if bpCopy.Packages[0].Version != "1.2.3" {
-		t.Fatalf("Blueprint.DeepCopy failed, copy modified")
-	}
+	require.Equalf(t, bpCopy.Packages[0].Version, "1.2.3", "Blueprint.DeepCopy failed, copy modified.")
 }
 
 func TestBlueprintInitialize(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/blueprint/blueprint_test.go
```diff
@@ -3,6 +3,7 @@ package blueprint
 import (
 	"github.com/google/go-cmp/cmp"
 	"testing"
+    "github.com/stretchr/testify/assert"
```
atodorov: Looks like tab vs spaces issue. Try `./tools/prepare-source.sh`

internal/blueprint/blueprint_test.go
```diff
@@ -89,3 +90,26 @@ func TestBumpVersion(t *testing.T) {
...
+		Version:     "0.0.1",
+		Packages: []Package{
+			{Name: "tmux", Version: "1.2"}},
+        Modules: []Package{
+            {Name: "openssh-server", Version: "8.*"}},
+        Groups: []Group{
+            {Name: "anaconda-tools"}},
+
+	}
+    out := bp.GetPackages()
```
atodorov: Other than naming this `result` looks good.

internal/blueprint/blueprint_test.go
jrusz: Thanks, that helped. I also added all the things it generated like @ondrejbudai once told me to do. Should also fix the failing CI.

internal/blueprint/blueprint_test.go
jrusz: Sorry, forgot to change this, this was just a placeholder, thanks for noticing.

internal/blueprint/blueprint_test.go
```diff
@@ -89,3 +90,22 @@ func TestBumpVersion(t *testing.T) {
...
+			{Name: "tmux", Version: "1.2"}},
+		Modules: []Package{
+			{Name: "openssh-server", Version: "8.*"}},
+		Groups: []Group{
+			{Name: "anaconda-tools"}},
+	}
+	out := bp.GetPackages()
+	assert.Equal(t, out[0], "tmux-1.2")
+	assert.Equal(t, out[1], "openssh-server-8.*")
+	assert.Equal(t, out[2], "@anaconda-tools")
```
atodorov: The arguments definition for assert functions show that expected value if first, then actual value is second. The order will not affect the assertion but will affect the error message in case of failure.

You may also try `assert.ElementsMatch` which will allow you to compare the arrays directly.

jrusz: Great idea, looks much better this way, thanks.

internal/blueprint/blueprint_test.go
```diff
@@ -18,24 +18,16 @@ func TestDeepCopy(t *testing.T) {
 	}
 
 	bpCopy, err := bpOrig.DeepCopy()
-	if err != nil {
-		t.Fatalf("Blueprint.DeepCopy failure: %s", err.Error())
-	}
-	if diff := cmp.Diff(bpOrig, bpCopy); diff != "" {
-		t.Fatalf("Blueprint.DeepCopy is different from original\ndiff: %s", diff)
-	}
+	require.Nilf(t, err, "Blueprints.DeepCopy failure: %#v", err)
```
bcl: An alternative to this is NoError

jrusz: I tried it like this:
```
    if require.NoError(t, err) {
        require.Equalf(t, bpOrig, bpCopy, "Blueprints.DeepCopy is different from original.")
    }
```
which should be the same as in the documentation but I'm getting this:
```
$ go test -v -race -covermode atomic -coverprofile=profile.cov .
# github.com/osbuild/osbuild-composer/internal/blueprint [github.com/osbuild/osbuild-composer/internal/blueprint.test]
./blueprint_test.go:21:23: require.NoError(t, err) used as value
FAIL    github.com/osbuild/osbuild-composer/internal/blueprint [build failed]
```
What am I doing wrong?

jrusz: Should have used assert, but I used it wrong anyway, should be better now.

internal/blueprint/blueprint_test.go
```diff
@@ -89,3 +90,20 @@ func TestBumpVersion(t *testing.T) {
...
+		Version:     "0.0.1",
+		Packages: []Package{
+			{Name: "tmux", Version: "1.2"}},
+		Modules: []Package{
+			{Name: "openssh-server", Version: "8.*"}},
+		Groups: []Group{
+			{Name: "anaconda-tools"}},
+	}
+	Received_packages := bp.GetPackages()
+    assert.ElementsMatch(t, []string{"tmux-1.2", "openssh-server-8.*", "@anaconda-tools"} , Received_packages)
```
teg: Looks like this needs a `go fmt`?

teg:
The added test looks good (just some white-space error). I think @ondrejbudai wrote the original tests for blueprints, would you like to have a look at that change?

internal/blueprint/blueprint_test.go
jrusz: Ah yes, I got it formatted in the other commit. I'll move it here, thanks.

internal/blueprint/blueprint.go
```diff
@@ -43,13 +43,13 @@ type Group struct {
 func (b *Blueprint) DeepCopy() (Blueprint, error) {
 	bpJSON, err := json.Marshal(b)
 	if err != nil {
-		return Blueprint{}, err
+		panic(err)
```
atodorov: I can't say what the implications of this are, maybe @bcl can. However I'm looking at this statement: 
"The return can't be easily covered the way it's written now"

Can you explain why ? 

What is preventing you from supplying appropriate input data (or mocking json.Marshall so it returns an error) then reading the returned values from this function and comparing the result via asserts, e.g. assert.NotNil(t, returned_err) ?

jrusz: Well it only accepts type `Blueprint` as an input and I can't find a way to give it wrong data. It's defined in a way that it only accepts the specified data types:
```
type Blueprint struct {
	Name           string          `json:"name" toml:"name"`
	Description    string          `json:"description" toml:"description"`
	Version        string          `json:"version,omitempty" toml:"version,omitempty"`
	Packages       []Package       `json:"packages" toml:"packages"`
	Modules        []Package       `json:"modules" toml:"modules"`
	Groups         []Group         `json:"groups" toml:"groups"`
	Customizations *Customizations `json:"customizations,omitempty" toml:"customizations,omitempty"`
}
```
so the `json.Marshal` and `json.Unmarshal` will always get the correct input. If you see a way to make it return an error I would be happy if you can share it.

jrusz:
@atodorov Alright, it's ready now. Thanks for the review. I wanted to cover `customizations.go` as a part of this but I can always make another PR.

PR #381, opened 2020-03-23
Add coverage for *assembler.go

jikortus:
This improves coverage of assembler.go, the other related _assembler.go files still remaining to cover. I'd be glad for any comments, especially I wonder it there is any elegant way to better format the long lines with json strings.

### DIFF
--- a/internal/osbuild/assembler_test.go
+++ b/internal/osbuild/assembler_test.go
@@ -1,0 +1,159 @@
+package osbuild
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAssembler_UnmarshalJSON(t *testing.T) {
+	tests := []struct {
+		name          string
+		assembler     Assembler
+		data          []byte
+		errorExpected bool
+	}{
+		{
+			// invalid JSON - note the missing brace at the end of the string
+			name:          "invalid json",
+			data:          []byte(`{"name":"org.osbuild.tar","options":{"filename":"",size:0}`),
+			errorExpected: true,
+		},
+		{
+			// valid JSON, but with an unknown assembler (org.osbuild.foo)
+			name:          "unknown assembler",
+			data:          []byte(`{"name":"org.osbuild.foo","options":{"bar":null}}`),
+			errorExpected: true,
+		},
+		{
+			name:          "missing options",
+			data:          []byte(`{"name":"org.osbuild.rawfs"`),
+			errorExpected: true,
+		},
+		{
+			name:          "missing name",
+			data:          []byte(`{"options":{"bar":null}}`),
+			errorExpected: true,
+		},
+		{
+			name: "qemu assembler empty",
+			assembler: Assembler{
+				Name:    "org.osbuild.qemu",
+				Options: &QEMUAssemblerOptions{},
+			},
+			data: []byte(`{"name":"org.osbuild.qemu","options":{"format":"","filename":"","size":0,"ptuuid":"","pttype":"","partitions":null}}`),
+		},
+		{
+			name: "qemu assembler full",
+			assembler: Assembler{
+				Name: "org.osbuild.qemu",
+				Options: &QEMUAssemblerOptions{
+					Format:   "qcow2",
+					Filename: "disk.qcow2",
+					Size:     2147483648,
+					PTUUID:   "0x14fc63d2",
+					PTType:   "mbr",
+					Partitions: []QEMUPartition{QEMUPartition{
+						Start:    2048,
+						Bootable: true,
+						Filesystem: QEMUFilesystem{
+							Type:       "ext4",
+							UUID:       "76a22bf4-f153-4541-b6c7-0332c0dfaeac",
+							Label:      "root",
+							Mountpoint: "/",
+						},
+					}},
+				},
+			},
+			data: []byte(`{"name":"org.osbuild.qemu","options":{"format":"qcow2","filename":"disk.qcow2","size":2147483648,"ptuuid":"0x14fc63d2","pttype":"mbr","partitions":[{"start":2048,"bootable":true,"filesystem":{"type":"ext4","uuid":"76a22bf4-f153-4541-b6c7-0332c0dfaeac","label":"root","mountpoint":"/"}}]}}`),
+		},
+		{
+			name: "tar assembler empty",
+			assembler: Assembler{
+				Name:    "org.osbuild.tar",
+				Options: &TarAssemblerOptions{},
+			},
+			data: []byte(`{"name":"org.osbuild.tar","options":{"filename":"","size":0}}`),
+		},
+		{
+			name: "tar assembler full",
+			assembler: Assembler{
+				Name: "org.osbuild.tar",
+				Options: &TarAssemblerOptions{
+					Filename:    "root.tar.xz",
+					Size:        0,
+					Compression: "xz",
+				},
+			},
+			data: []byte(`{"name":"org.osbuild.tar","options":{"filename":"root.tar.xz","size":0,"compression":"xz"}}`),
+		},
+		{
+			name: "rawfs assembler empty",
+			assembler: Assembler{
+				Name:    "org.osbuild.rawfs",
+				Options: &RawFSAssemblerOptions{},
+			},
+			data: []byte(`{"name":"org.osbuild.rawfs","options":{"filename":"","root_fs_uuid":"00000000-0000-0000-0000-000000000000","size":0}}`),
+		},
+		{
+			name: "rawfs assembler full",
+			assembler: Assembler{
+				Name: "org.osbuild.rawfs",
+				Options: &RawFSAssemblerOptions{
+					Filename:           "filesystem.img",
+					RootFilesystemUUID: uuid.MustParse("76a22bf4-f153-4541-b6c7-0332c0dfaeac"),
+					Size:               2147483648,
+				},
+			},
+			data: []byte(`{"name":"org.osbuild.rawfs","options":{"filename":"filesystem.img","root_fs_uuid":"76a22bf4-f153-4541-b6c7-0332c0dfaeac","size":2147483648}}`),
+		},
+	}
+
+	assert := assert.New(t)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assembler := &tt.assembler
+			var gotAssembler Assembler
+			err := gotAssembler.UnmarshalJSON(tt.data)
+			if tt.errorExpected {
+				assert.NotNil(err)
+				return
+			} else {
+				assert.Nil(err)
+			}
+			gotBytes, err := json.Marshal(assembler)
+			assert.Nilf(err, "Could not marshal assembler: %v", err)
+			assert.Equal(tt.data, gotBytes)
+			assert.Equal(&gotAssembler, assembler)
+		})
+	}
+}
+
+func TestNewQEMUAssembler(t *testing.T) {
+	options := &QEMUAssemblerOptions{}
+	expectedAssembler := &Assembler{
+		Name:    "org.osbuild.qemu",
+		Options: &QEMUAssemblerOptions{},
+	}
+	assert.Equal(t, expectedAssembler, NewQEMUAssembler(options))
+}
+
+func TestNewTarAssembler(t *testing.T) {
+	options := &TarAssemblerOptions{}
+	expectedAssembler := &Assembler{
+		Name:    "org.osbuild.tar",
+		Options: &TarAssemblerOptions{},
+	}
+	assert.Equal(t, expectedAssembler, NewTarAssembler(options))
+}
+
+func TestNewRawFSAssembler(t *testing.T) {
+	options := &RawFSAssemblerOptions{}
+	expectedAssembler := &Assembler{
+		Name:    "org.osbuild.rawfs",
+		Options: &RawFSAssemblerOptions{},
+	}
+	assert.Equal(t, expectedAssembler, NewRawFSAssembler(options))
+}


jikortus:
FTR, the unit tests failed because of a typo in `rawfs_assembler.go`, should be fixed by #380.

teg:
This looks very nice! Thanks a lot! I merge #380, so this can be rebased on top now. For the record, you can file a PR as a draft, then you don't need to put [WIP] in the subject, and you can change the status to "ready for review" when you are happy with it. Makes sure we don't accidentally merge things that either isn't ready, or is ready but still has [WIP] in the title by mistake.

I have just let the long json strings be and not broken the lines. Open to suggestions...

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/google/uuid"
+	"reflect"
+	"testing"
+)
+
+func TestAssembler_UnmarshalJSON(t *testing.T) {
+	//nullUUID := uuid.MustParse("00000000-0000-0000-0000-000000000000")
```
atodorov: left over comment

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+	type fields struct {
+		Name    string
+		Options AssemblerOptions
+	}
```
atodorov: this looks like `type Assembler struct` from assembler.go. 

Why does the test have to define its own data type? 
Can't it just reuse the `Assembler` struct instead ?

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+				data: []byte(`{"name":"org.osbuild.rawfs","options":{"filename":"filesystem.img","root_fs_uuid":"76a22bf4-f153-4541-b6c7-0332c0dfaeac","size":2147483648}}`),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assembler := &Assembler{
+				Name:    tt.fields.Name,
+				Options: tt.fields.Options,
+			}
```
atodorov: Related to my questions above - I don't understand why not construct the test objects directly using their native data types instead of proxying via the `tests` array and then initializing the `assembler` variable here ?

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+			var gotAssembler Assembler
+			if err := gotAssembler.UnmarshalJSON(tt.args.data); (err != nil) != tt.wantErr {
```
atodorov: A very similar question for `args`.  Why do we need a proxy data type which encapsulates the `.data` field instead of providing `.data` (or whatever the name) directly inside `tests` ?

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+				t.Errorf("Assembler.UnmarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.wantErr {
+				return
+			}
+			gotBytes, err := json.Marshal(assembler)
+			if err != nil {
```
atodorov: assert.NoError or require.NoError

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,164 @@
...
+				t.Errorf("Could not marshal assembler: %v", err)
+			}
+			if !bytes.Equal(gotBytes, tt.args.data) {
+				t.Errorf("Expected `%v`, got `%v`", string(tt.args.data), string(gotBytes))
+			}
```
atodorov: again, use asserts here

jikortus:
@atodorov, @teg Thanks for the review! I addressed the issues Alex mentioned, but I guess I'll add some more polishing tomorrow. Are the 3 scenarios for tar/qemu/rawfs assemblers fine as they are, or should I rather try to execute them in a similar way as the previous tests (test definitions in a list that is subsequently looped through and evaluated)?

teg:
I think most the test cases we have so far have used a scheme autogenerated by VS Code (see for example https://rms1000watt.github.io/post/testing-your-go-app-get-started-the-easy-way/). I think it makes sense to stick to that style for consistency, but at the end of the day I'd leave it up to you if you think it makes sense to do things slightly differently.

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+				},
+			},
+			args: args{
+				data: []byte(`{"name":"org.osbuild.rawfs","options":{"filename":"filesystem.img","root_fs_uuid":"76a22bf4-f153-4541-b6c7-0332c0dfaeac","size":2147483648}}`),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
```
atodorov: Note: you don't really have to create a new object here. The alternative syntax will be `assert.Nilf(t, ...)` - i.e. passing `t` as the first argument for every assert function.

I don't really know if this is only a syntax difference or there could be more implications (other than consuming a few bytes of memory)

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+			assembler := &tt.assembler
+			var gotAssembler Assembler
+			err := gotAssembler.UnmarshalJSON(tt.args.data)
```
atodorov: ping about args.data, see comment from yesterday.

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+			assert.Equalf((err != nil), tt.wantErr, "Assembler.UnmarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
+			if tt.wantErr {
+				return
+			}
+			gotBytes, err := json.Marshal(assembler)
+			assert.Nilf(err, "Could not marshal assembler: %v", err)
+			assert.Equalf(gotBytes, tt.args.data, "Expected `%v`, got `%v`", string(tt.args.data), string(gotBytes))
```
atodorov: The arguments order is expected, actual and the error message is really redundant. The assert framework should be doing this for you. If you swap the order of arguments then in case of assertion failure the message will be slightly different. 

You can experiment with this but IMO printing the 2 values as an error message is absolutely unnecessary. That's why we have the framework.

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			assembler := &tt.assembler
+			var gotAssembler Assembler
+			err := gotAssembler.UnmarshalJSON(tt.args.data)
+			assert.Equalf((err != nil), tt.wantErr, "Assembler.UnmarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
```
atodorov: If `tt.wantErr` is missing from the data structure (lines 49-57 above)  will it be initialized to False ?

On side note I will probably go with
```golang
if tt.wantErrr {
    assert.Error(t, err)
} else {
   assert.NoError(...)
}
```

I like to have this more explicit so I don't have to think twice what I read.
to make

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+func TestAssembler_UnmarshalJSON(t *testing.T) {
+	type args struct {
+		data []byte
+	}
+	tests := []struct {
+		name      string
+		assembler Assembler
+		args      args
+		wantErr   bool
```
atodorov: I would prefer using a more expressive name like `wantsError` or `shouldError` or even `expectedError` (which will be the actual error type if you want to assert on that).

I don't know what the golang convention is but I don't see a reason for abbreviating variable names when they can be written in full and more expressive way.

atodorov:
> Are the 3 scenarios for tar/qemu/rawfs assemblers fine as they are, or should I rather try to execute them in a similar way as the previous tests (test definitions in a list that is subsequently looped through and evaluated)?

@jikortus to answer this question answer the question "what is the behavior I want to validate with this test". From what I can see your 3 functions `TestNew*Assembler` are trying to verify that the constructor of the respective 3 data types works. Where "works" is defined by:

1) does not crash and returns an object
2) actual object matches a hard coded expected object

The implementation of the 3 methods under test is basically: 
```
return &Assembler{
                Name:    "xxxxxx",
                Options: options,
        }
```

where the name is different. 


So to answer the original question: when validating the behavior of `New` you are validating the expected name. The options are actually irrelevant in this scenario and can be the same (you can also iterate on a data structure if you want and design the test is that way). 

So the answer is yes, the current tests are fine as they are b/c:
- you are validating a very simple behavior and the current implementation is sufficient for that
- this is a unit type test and it shouldn't go outside trying to read definitions from the file system
- nothing else is actually necessary.

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+				},
+			}},
+		},
+	}
+	assert.Equal(t, expectedAssembler, NewQEMUAssembler(options))
+}
+
+func TestNewTarAssembler(t *testing.T) {
+	options := NewTarAssemblerOptions("disk.tar.xz", 0)
+	options.SetCompression("xz")
```
atodorov: This should not be here. It is changing the expected behavior/meaning of this scenario. 

The constructor doesn't call `SetCompression` so you should assert that there is no compression by default and design a separate scenario which will assert that after calling SetCompression the options have been changed (but that is not related to the behavior of New).

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,230 @@
...
+	assert.Equal(t, expectedAssembler, NewTarAssembler(options))
+}
+
+func TestNewRawFSAssembler(t *testing.T) {
+	options := &RawFSAssemblerOptions{
+		Filename:           "disk.img",
+		RootFilesystemUUID: uuid.MustParse("76a22bf4-f153-4541-b6c7-0332c0dfaeac"),
+		Size:               0,
+		FilesystemType:     "ext4",
+	}
```
atodorov: Like mentioned in my previous comment: I think this is not necessary for the current scenario(s). The constructor is passing the options argument unmodified to the Assembler class so it actually should be possible to exercise and validate the functionality of all `New` functions without any kind of options (empty or nil struct even).

internal/osbuild/assembler_test.go
jikortus: The two ways should be equal (https://godoc.org/github.com/stretchr/testify/assert) and that one I used can slightly improve readability, IMHO. But of course I don't insist on using it.

internal/osbuild/assembler_test.go
jikortus: I don't think there is a better way around this. I have to loop through all of the test scenarios and compare the expected and actual results. I could use something like `tests[i].assembler`, but IMO it wouldn't look nicer that what I'm using now (as I found out, the code is partially generated by VS code, and I borrowed the base for the assembler tests from an existing test file).

internal/osbuild/assembler_test.go
jikortus: That's a valid point, so I simplified all the `Equalf`s.

internal/osbuild/assembler_test.go
jikortus: Good point, I made changes for a better readability. IIRC the undefined value of `wantErr` (now `errorExpected`) is fine, since in this case its expected value is `false`, which is assigned implicitly (https://golang.org/ref/spec#The_zero_value).

internal/osbuild/assembler_test.go
jikortus: OK, I changed that to `errorExpected`.

internal/osbuild/assembler_test.go
jikortus: That's true, so I moved the use of `SetCompression` into a separate test.

internal/osbuild/assembler_test.go
atodorov: The original question was:
> Why do we need a proxy data type which encapsulates the .data field instead of providing .data (or whatever the name) directly inside tests ?

So I was asking why do we need to have `tt.args.data` instead of simply `tt.data`, iow why do we need the intermediate `args` struct which contains only 1 attribute inside of it? Why not assign the byte array directly as an attribute in the `tests` struct ?

internal/osbuild/assembler_test.go
atodorov: Yes, both are equivalent. However you are creating a new object inside a loop and that new object always has the same arguments, namely `t`.

It would be different if that was `assert.New(tt)` - you may say that assert errors will get better annotation that points to each particular test data member in the `tests` array. However the code in question is doing the same thing on every iteration. 

If you want to keep the current syntax you can either change the arguments to New() (although I don't think there will be any practical difference) or move this line above the for loop (which will gain a miscule speed improvement).

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,202 @@
...
+				assert.NotNil(err)
+				return
+			} else {
+				assert.Nil(err)
+			}
+			gotBytes, err := json.Marshal(assembler)
+			assert.Nilf(err, "Could not marshal assembler: %v", err)
+			assert.Equal(tt.args.data, gotBytes)
+			assert.Truef(reflect.DeepEqual(&gotAssembler, assembler),
+				"got {%v, %v}, expected {%v, %v}",
```
atodorov: About error message here: assert should print out expected & actual so we can see the difference in case of failure. The error message should be redundant, isn't it ? 

Also isn't assert.Equal here sufficient, you are using it in the test scenarios below so I'm wondering why the need for reflect.DeepEqual ?

internal/osbuild/assembler_test.go
jikortus: Ah, of course. The `assert.New` inside of the loop doesn't really make sense (and it even wasn't an intention). Fixed.

internal/osbuild/assembler_test.go
jikortus: Now I see what you were getting at :), of course all you said makes perfect sense. Fixed as well.

internal/osbuild/assembler_test.go
jikortus: To be honest, this assert made it into the new test file as it was part of the test file I used as the basis for the new one (together with a few other thing that didn't really make sense and you pointed at). Hopefully all of the imperfections have been fixed.

internal/osbuild/assembler_test.go
atodorov: I'm not seeing the difference.

jikortus: The difference is that I moved the instantiation of the assert object outside of the loop, so that now it's instantiated only once, before looping through all the scenarios.

atodorov: Line 112 contradicts that.

jikortus: OMG, sorry for my blindness. Fixed.

internal/osbuild/assembler_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}{
+		{
+			name:          "invalid json",
+			data:          []byte(`{"name":"org.osbuild.foo","options":{"bar":null}`),
+			errorExpected: true,
+		},
+		{
+			name:          "unknown assembler",
+			data:          []byte(`{"name":"org.osbuild.foo","options":{"bar":null}}`),
+			errorExpected: true,
```
atodorov: Either I'm not seeing something or this record is the same as the one above. . It would also help if you add a short comment for invalid data entries which points out exactly where the invalid piece of data is.

jikortus: The two records are indeed distinct - the first one is for an invalid JSON (with a missing brace at the end). I added comments for the first two records to make that obvious.

Issue #342, opened 2020-03-17
Split the job queue from the store


larskarlitski:
The store contains two main things:

1. User state: blueprints (and their history), additional sources
2. The compose queue: past, running, and pending composes

We have two APIs for **osbuild-composer**: `weldr` and `rcm`. `weldr` needs both (1) and (2), but `rcm` only needs (2).

Going forward, we'd like to put the queue into a proper database, so that it can scale to many image builds. I don't think there's a need to put (1) into a database just yet, because of the nature of the `weldr` API (local socket only, not meant for 1000s of blueprints). And even if we did, it would probably have different access policies than the one for the queue.

Thus, I propose splitting (1) and (2). Specifically, I'd like to move the queue into the `jobqueue` package, which already contains the worker-facing part of the queue. It would be an interface with different backends: as a first step, a JSON-file backend, similar to the store (`/var/lib/osbuild-composer/queue.json`) gives us the same functionality that we have right now. Going forward, we could add one that is backed by a database (out of scope for this issue).

A side benefit would be a much simplified `store` package. It has a few too many moving parts for my taste right now­—especially considering the amount of locking and goroutines in there.

@msehnout, @dvdhrm, @teg curious to hear your thoughts about this.

teg:
This makes perfect sense to me! Thanks for writing it up!

ondrejbudai:
I have a thought or two too! :raising_hand_man: 

### Unit testing
Since the beginning of times we have `stateDir *string`. When store is used in the production code, it's set to `STATE_DIRECTORY`. However, in the unit tests it's set to `nil`. Unfortunately, this means that we need to check for nil-ness in several places and this just feels weird. To make things worse, in several places the checks are missing and the behavior is therefore non-consistent (I'm guilty of this offense).

I feel like this weirdness should be removed. I can think of two solutions:

1) Make a very thin wrapper responsible for reading/writing data from/to filesystem. Then, make a mock which does all of this in-memory.

2) Make `stateDir` non-pointer and set up temporary directory in the unit tests. This solution may be a bit easier, however I'm not sure if I like the idea of unit tests writing data to disk.

It may be out of scope of this issue, however I feel like these things are closely connected.

### Store versioning

We discussed that yesterday. We want to start maintaining backwards compatibility soon, so we need to version the store to be able to do migrations (that's the word!) from an older version to a newer one.

I think this is probably out of scope of this issue, but I wanted to mention it here if you have any thoughts and we can create a separate one later.

[...]

PR #382, opened 2020-03-23
Store cleanups

larskarlitski:
Some random cleanups of the store and its interaction with weldr, while I'm continuing to research towards #342.

Hope the commits are self-explanatory.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -554,8 +554,10 @@ func (s *Store) getImageBuildDirectory(composeID uuid.UUID, imageBuildID int) st
 	return fmt.Sprintf("%s/%d", s.getComposeDirectory(composeID), imageBuildID)
 }
 
-func (s *Store) PushCompose(distro distro.Distro, arch distro.Arch, imageType distro.ImageType, composeID uuid.UUID, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, size uint64, uploadTarget *target.Target) error {
-	targets := []*target.Target{}
+func (s *Store) PushCompose(distro distro.Distro, arch distro.Arch, imageType distro.ImageType, composeID uuid.UUID, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, size uint64, targets []*target.Target) error {
+	if targets == nil {
+		targets = []*target.Target{}
+	}
 
 	// Compatibility layer for image types in Weldr API v0
 	imageTypeCommon, exists := common.ImageTypeFromCompatString(imageType.Name())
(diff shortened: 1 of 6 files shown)


internal/store/store.go
```diff
@@ -554,8 +554,10 @@ func (s *Store) getImageBuildDirectory(composeID uuid.UUID, imageBuildID int) st
 	return fmt.Sprintf("%s/%d", s.getComposeDirectory(composeID), imageBuildID)
 }
 
-func (s *Store) PushCompose(distro distro.Distro, arch distro.Arch, imageType distro.ImageType, composeID uuid.UUID, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, size uint64, uploadTarget *target.Target) error {
-	targets := []*target.Target{}
+func (s *Store) PushCompose(distro distro.Distro, arch distro.Arch, imageType distro.ImageType, composeID uuid.UUID, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, size uint64, targets []*target.Target) error {
+	if targets == nil {
```
teg: This doesn't actually drop the appending of the local target, does it?

internal/weldr/api.go
```diff
@@ -1411,8 +1411,25 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
 			statusResponseError(writer, http.StatusBadRequest, errors)
 			return
 		}
+		targets = append(targets, t)
 	}
 
+	filename, _, err := api.distro.FilenameFromType(cr.ComposeType)
```
teg: `distro.FilenameFromType()` is deprecated. Simply use `imageType.Filename()` instead, and you can drop the error handling.

internal/weldr/api.go
```diff
@@ -1411,8 +1411,25 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
...
+	if err != nil {
+		errors := responseError{
+			ID:  "BadCompose",
+			Msg: fmt.Sprintf("internal error: no filename for compose type %s", cr.ComposeType),
+		}
+		statusResponseError(writer, http.StatusBadRequest, errors)
+		return
+	}
+
+	targets = append(targets, target.NewLocalTarget(
```
teg: Before this patch, this was skipped if the store did not have a state directory. I think that's the wrong place to have a conditional (do it when receiving the file, if anything), so this should be ok (that was only conditional for easier testing, so might never matter at all). But worth a comment in the commit message?

internal/store/store.go
larskarlitski: Oh dear, I missed that while rebasing on top of the refactor. Thanks for catching that!

internal/weldr/api.go
larskarlitski: Oh I get to drop the whole error case. Nice!

internal/weldr/api.go
larskarlitski: Good point. Added an explanation.

larskarlitski:
> Love these cleanups! In addition to request to one of the patches, a `go fmt` is needed.

Thanks for the review! Should be good for another round.

larskarlitski:
Update to resolve the conflict.

larskarlitski:
This [passed on travis](https://travis-ci.org/github/larskarlitski/osbuild-composer/builds/666522179), but it failed to send status.

PR #383, opened 2020-03-23
Distro cleanup

teg:
After the recent addition of new interfaces to the distro package, drop dead code and do other mechanical cleanups.

### DIFF
--- a/internal/distro/fedora30/distro_test.go
+++ b/internal/distro/fedora30/distro_test.go
@@ -1,30 +1,10 @@
 package fedora30_test
 
 import (
-	"reflect"
 	"testing"
 
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
 )
-
-func TestListOutputFormats(t *testing.T) {
-	want := []string{
-		"ami",
-		"ext4-filesystem",
-		"openstack",
-		"partitioned-disk",
-		"qcow2",
-		"tar",
-		"vhd",
-		"vmdk",
-	}
-
-	f30 := fedora30.New()
-
-	if got := f30.ListOutputFormats(); !reflect.DeepEqual(got, want) {
-		t.Errorf("ListOutputFormats() = %v, want %v", got, want)
-	}
-}
 
 func TestFilenameFromType(t *testing.T) {
 	type args struct {
--- a/internal/distro/fedora31/distro.go
+++ b/internal/distro/fedora31/distro.go
@@ -5,7 +5,6 @@ import (
 	"sort"
 	"strconv"
 
-	"github.com/osbuild/osbuild-composer/internal/common"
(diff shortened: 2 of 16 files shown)


internal/distro/fedora31/distro.go
```diff
@@ -16,45 +15,45 @@ import (
...
+	excludedPackages []string
+	enabledServices  []string
+	disabledServices []string
+	kernelOptions    string
+	bootable         bool
+	defaultSize      uint64
+	assembler        func(uefi bool, size uint64) *osbuild.Assembler
+}
+
+type fedora31Arch struct {
```
msehnout: What is the reason for having 2 private "arch" struct here? (`fedora31Arch` and `arch`), seems to me that they have name in common, so the only difference is a pointer to the distro struct.

internal/distro/fedora31/distro_test.go
```diff
@@ -1,31 +1,11 @@
 package fedora31_test
 
 import (
-	"reflect"
 	"testing"
 
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora31"
 )
 
-func TestListOutputFormats(t *testing.T) {
```
msehnout: This should still be possible, just more complicated, no? For example a test case testing a matrix of arches and image types.

internal/distro/fedora31/distro.go
teg: This was to make the (already very large) patch a bit smaller. Would be a natural next cleanup. You are right, this is clearly not how it should be.

internal/distro/fedora31/distro_test.go
teg: Yeah, totally. Also need tests for the other newly introduced functions. I'll post a follow up to add this.

internal/distro/fedora30/distro_test.go
```diff
@@ -1,31 +1,11 @@
 package fedora30_test
 
 import (
-	"reflect"
 	"testing"
 
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
 )
 
-func TestListOutputFormats(t *testing.T) {
```
larskarlitski: Why did you drop these?

internal/distro/fedora31/distro.go
larskarlitski: We discussed this on the last PR: https://github.com/osbuild/osbuild-composer/pull/375#discussion_r396374144

Tom's idea is to do this in chunks.

internal/distro/fedora30/distro_test.go
larskarlitski: Sorry, didn't see that you already discussed this with Martin above.

PR #385, opened 2020-03-24
Rename the rpms to osbuild-composer

### LINKED ISSUE #355: Create osbuild-composer package in Fedora
Goal: get rid of `golang-github-` prefix in the RPM package name

We need to start the whole review process again:
https://fedoraproject.org/wiki/Package_Renaming_Process
And retire the old package afterwards. 

We should also do the same for RHEL.

ondrejbudai:
This PR adds the osbuild-composer spec file. The rpm targets in Makefile were aligned with this change. Additonally, `old-rpm` and `old-srpm` targets were added to build the old spec file. The CI now builds both packages: `github-go...` for f31 and `osbuild-composer` for f32 and f33. Both specfiles were synced to be fairly similar.

The old spec file should be dropped when we decide not to release for f31 anymore.

Closes #355


.github/workflows/tests.yml
```diff
@@ -46,13 +46,54 @@ jobs:
...
+        uses: actions/upload-artifact@v1
+        with:
+          name: rpms
+          path: rpms/
+
+      - name: "🔎 Test RPM installation"
+        run: dnf -y install $(ls rpms/*.x86_64.rpm)
+
   rpm_build:
-    name: "📦 RPM"
```
larskarlitski: There's still a mention of `golang-github-osbuild-composer` in the "Install RPM dependencies" section.

(Can't comment on the lines that are not in the diff.)

ondrejbudai: Ah, I missed that, thanks!

ondrejbudai:
I dropped the packit config for now, we can reintroduce it later if we decide to enable it again.

ondrejbudai:
Rebased!

PR #387, opened 2020-03-24
🤦 [WIP] blueprints: drop unsued returns

jrusz:
This drops unsued returns in the function which could not be covered the way it's written now.


jrusz:
I am currently investigating `mock` as an option to overcome this and test it the way it's written now.

bcl:
I can't find the PR with the panic() in it for some reason, I know I saw it earlier though.
I think panic() is fine in this case, since the types guarantee they should always be able to marshal/unmarshal any error is going to be very unusual.
If you use the panic then there is no reason to return an error, so that could be dropped.
(In general you shouldn't use panic, but I think this is one of the exceptions since we likely have no control over whatever could make it actually fail).

jrusz:
This is obsolete now, closing.

Issue #315, opened 2020-03-12
RFC: Testing for repositories


atodorov:
From lorax-composer we have a test_repos.sh file which performs the following scenario: 

1) Start composer with --no-system-repos, that is 0 repositories defined; then start a new compose and expect it to fail. We make sure the failure says "Problem depsolving"

2) Manually define a repo in `/var/lib/lorax/composer/repos.d/`, start composer again (--no-system-repos again), start a compose and wait for it to complete successfully.

I can see osbuild-composer comes with a predefined set of repositories based on the OS type (Fedora/RHEL).  There are also `composer-cli sources` commands which do not appear covered by tests. 

I also tried `composer-cli sources delete fedora` (the default repo) and that didn't crash but also didn't delete the source.

As of now it feels like reproducing the old scenarios verbatim doesn't make sense. It's probably better to add sanity tests for the `sources` command and probably API tests for add/delete behavior. 

Questions: 
* Is there a way to provide additional repos configuration/exclude stock repos for osbuild-composer, other then `composer-cli sources` command ? For example a directory that we're reading from.
* What is the intended behavior for `sources delete` and default sources ? Do we want to allow users to delete absolutely all sources coming from the backend and substitute with their own? I can see how this would be a desirable feature (e.g. do not connect to anything upstream).
* What would be the place where we want this implemented. For the sanity portion I think `osbuild-tests` is the place. Not sure if we want to split this out into a separate binary.

teg:
- It is currently possible to override the built-in repositories, but dropping files in `/etc/osbuild-composer/repositories.d/`. However, I think we should possibly reconsider this behaviour and I don't think it is something worth testing. The intended use-case is to be able to override the stock repositories for the purposes of testing against beta releases where the repositories are net yet in their designated location. As I said, I'm unsure if this is the path forward, so I would not make this the highest priority to test.
 - We should follow the behaviour of the weldr API. That is, if running lorax-composer in the normal mode (without `--no-system-repos`) means that the system repositories cannot be deleted, then we should behave the same way. If, on the other hand, `lorax-composer` allows any repository to be deleted, we should consider doing the same (would require some rework, but we could discuss that in a dedicated issue).
 - I think we the behaviour should be tested as part of `osbuild-test-weldr` (i.e., checking the behavior of deleting of system repos etc). `osbuild-test` could contain a test-case that just checks that the `sources delete` command works at all, but I would not go into verifying the behavior.

@bcl do you have any more thoughts on this?

[...]

PR #388, opened 2020-03-24
tests: Make TestSources distro independent. Refs #315

atodorov:
by loading repositories configuration first and then running `composer-cli sources info <repo>` on the first repository found.

@teg I took a different approach for this. Please review.

### DIFF
--- a/cmd/osbuild-tests/main_test.go
+++ b/cmd/osbuild-tests/main_test.go
@@ -77,10 +77,8 @@ gpgkey_urls = ["https://url/path/to/gpg-key"]
 	require.NoError(t, err)
 
 	runComposerCLI(t, false, "sources", "list")
-	// todo: this will fail on RHEL
-	runComposerCLI(t, false, "sources", "info", "fedora")
-
 	runComposerCLI(t, false, "sources", "add", sources_toml.Name())
+	runComposerCLI(t, false, "sources", "info", "osbuild-test-addon-source")
 	runComposerCLI(t, false, "sources", "change", sources_toml.Name())
 	runComposerCLI(t, false, "sources", "delete", "osbuild-test-addon-source")
 }


atodorov:
@teg +1 for leaving the sanity tests as-is and doing more checking elsewhere. 

My motivation here is to make this piece of code distro independent without having to hard-code the repository names in the actual test code b/c I don't like hard coding it. Other than that there is no difference in the size of this diff regardless which approach we take.

atodorov:
> I think we should do a mapping in the test from distro to expected source name. Then we can test both against lorax composer and osbuild composer to check they are the same.

@teg IIRC lorax composer doesn't care about the names. This proposed scenario doesn't make sense to me. I can't figure out what are we trying to validate here (other than the names are as defined in repositories json files).  However this is a separate discussion.

teg:
I think these tests should not depend on `osbuild-composer` internals, and I do not think we need to verify here that the names of the system sources are as expected. If we want that, we could add that to the `weldr` test.

I still think it makes sense to test that `sources info` works, without getting into the specific behaviour. How about something like this (untested):
```patch
diff --git a/cmd/osbuild-tests/main_test.go b/cmd/osbuild-tests/main_test.go
index a3b5ed8..083a3a4 100644
--- a/cmd/osbuild-tests/main_test.go
+++ b/cmd/osbuild-tests/main_test.go
@@ -77,10 +77,9 @@ gpgkey_urls = ["https://url/path/to/gpg-key"]
        require.NoError(t, err)
 
        runComposerCLI(t, false, "sources", "list")
-       // todo: this will fail on RHEL
-       runComposerCLI(t, false, "sources", "info", "fedora")
 
        runComposerCLI(t, false, "sources", "add", sources_toml.Name())
+       runComposerCLI(t, false, "sources", "info", "osbuild-test-addon-source")
        runComposerCLI(t, false, "sources", "change", sources_toml.Name())
        runComposerCLI(t, false, "sources", "delete", "osbuild-test-addon-source")
 }
```

atodorov:
@teg I like your idea. It makes the change very minimal

teg:
Thanks!

PR #390, opened 2020-03-24
Add integration tests for projects and modules

bcl:


internal/weldrcheck/modules_test.go
```diff
@@ -0,0 +1,78 @@
...
+	require.Nil(t, api, "GetModulesInfo failed: %#v", api)
+	require.True(t, len(modules) == 1, "Not enough modules returned: %#v", modules)
+}
+
+// Get info on two specific modules
+func TestTwoModuleInfoV0(t *testing.T) {
+	modules, api, err := client.GetModulesInfoV0(testState.socket, "bash,tmux")
+	require.NoError(t, err)
+	require.Nil(t, api, "GetModulesInfo failed: %#v", api)
+	require.True(t, len(modules) == 2, "Not enough modules returned: %#v", modules)
```
atodorov: Could have used `require.Equal(t, 2, len(modules))` as well. It is a bit more expressive.

PR #391, opened 2020-03-24
Document when to panic

larskarlitski:
There was a long discussion about when to `panic` today. I've tried capturing the output in `docs/errors.md`. The idea is to add more rules to this document as we make them.

I've also added a couple of commits that change existing code to use these new rules.

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -141,7 +141,7 @@ func main() {
 		}
 		bytes, err = json.Marshal(rpmMDInfo)
 		if err != nil {
-			panic("could not marshal rpmmd struct into JSON")
+			panic(err)
 		}
 	} else {
 		manifest, err := imageType.Manifest(blueprint.Customizations, repos[arch.Name()], packageSpecs, buildPackageSpecs, imageType.Size(0))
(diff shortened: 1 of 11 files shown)


dvdhrm:
Just because I happened to think of it, there is a very nice writeup about error-handling by `Joe Duffy`, a Microsoft researcher who managed the `Midori` project some years ago (one of their research operating systems):

    http://joeduffyblog.com/2016/02/07/the-error-model/

The entire series is available here:

    http://joeduffyblog.com/2015/11/03/blogging-about-midori/

Just in case someone is interested. I found this entire series to be a very great read.

ondrejbudai:
I very much like the document!

However, I just realized there may be a slight issue with panicking when `json.NewEncoder(writer).Encode(reply)` errors.

If `writer` is `http.ResponseWriter`, it's possible that we're actually writing to a tcp stream. If a network fail occurs, this could crash composer. However, I'm not really sure about this - networking is hard.

If we have a public version of osbuild-composer and my thoughts are true, this could lead to osbuild-composer crash everytime a network drop occurs. Or if someone decides to mess with us... 🦹

larskarlitski:
> However, I just realized there may be a slight issue with panicking when `json.NewEncoder(writer).Encode(reply)` errors.

You are completely right of course. Good catch!

I don't know what we want to do in these situations (ignore the error? log?). I've dropped these hunks from this PR, because it is unrelated to this PR.

ondrejbudai:
I think we should log them, just so we know that they happen. Also, we should remove the encoding panics in `weldr/api`.

PR #392, opened 2020-03-25
worker: move ComposerClient to jobqueue and defaults to unit files

larskarlitski:
See individual commits.

More is coming for `jobqueue.Client` if people think this it the right direction. This is a good first chunk that's ready for review.

It changes how `osbuild-worker`'s command line arguments. I don't think we treat those as stable API, right?

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -23,6 +23,7 @@ const RemoteWorkerPort = 8700
 
 type ComposerClient struct {
 	client   *http.Client
+	scheme   string
 	hostname string
 }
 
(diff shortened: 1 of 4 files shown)


larskarlitski:
Rebased because of conflicts I introduced with https://github.com/osbuild/osbuild-composer/pull/391

cmd/osbuild-worker/main.go
```diff
@@ -55,32 +56,33 @@ func createTLSConfig(config *connectionConfig) (*tls.Config, error) {
...
+func NewClient(remoteAddress string, conf *tls.Config) *ComposerClient {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: conf,
+		},
 	}
 
-	// plain non-encrypted connection
-	return net.Dial("unix", "/run/osbuild-composer/job.socket")
+	var scheme string
```
teg: What's the logic here? When are we meant to use http and when https, and why was it ok to hardcode to http before this patch?

larskarlitski: We hard-coded the tls config when connecting before. One of the patches changes that to set the config on Transport and let the library to its magic. It only does that magic for `https`, though. I think it's the more correct thing to do regardless.

Would you like to have this explained in a commit message? Sorry I forgot that.

teg: Ah, yes, of course. Thanks!

PR #393, opened 2020-03-25
tools/image-info: fix running the tool on Fedora 32

ondrejbudai:
see the individual commits


ondrejbudai:
I added two more commits:
- the first one fixes `image-info`, so it doesn't contain duplicate entries in services anymore
- the second one has regenerated test cases

PR #394, opened 2020-03-25
Add source tests for deleting user-defined & system wide repositories

atodorov:
Between the existing sanity tests and this PR the behavior documented in #315 is covered.

### DIFF
--- a/internal/client/source_test.go
+++ b/internal/client/source_test.go
@@ -187,7 +187,7 @@ func TestGetSourceInfoV0(t *testing.T) {
 func UploadUserDefinedSources(t *testing.T, sources []string) {
 	for i := range sources {
 		source := strings.Replace(sources[i], "REPO-PATH", testState.repoDir, 1)
-		resp, err := client.PostJSONSourceV0(testState.socket, source)
+		resp, err := PostJSONSourceV0(testState.socket, source)
 		require.NoError(t, err, "POST source failed with a client error")
 		require.True(t, resp.Status, "POST source failed: %#v", resp)
 	}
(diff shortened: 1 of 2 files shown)


atodorov:
@bcl can you take a look at this ? Should be OK to merge. Thanks.

bcl:
Let's wait until PR #404 is merged, then rebase it and make sure it works as a unit test as well.

internal/client/source_test.go
```diff
@@ -183,3 +187,92 @@ func TestGetSourceInfoV0(t *testing.T) {
...
+func TestDeleteSystemSourcesV0(t *testing.T) {
+	hostDistroName, err := distro.GetHostDistroName()
+	hostArch := common.CurrentArch()
+	require.NoError(t, err)
+
+	// note: this is an integration test and it expects osbuild-composer to be installed
+	// that's why we search for repositories configuration in system wide directories!
+	repoMap, err := rpmmd.LoadRepositories([]string{"/etc/osbuild-composer", "/usr/share/osbuild-composer"}, hostDistroName)
+	require.NoError(t, err, "rpmmd.LoadRepositories: %v", err)
+	require.GreaterOrEqual(t, len(repoMap[hostArch]), 1)
```
atodorov: @bcl after the changes in #404 I'm not sure how these tests are expected to be executed and if repository definitions should be read from the system wide directories. Can you comment on this?

atodorov:
Failures here don't seem related to the proposed changes.

internal/client/source_test.go
atodorov: After the rebase this now fails b/c I have only a "test-id" repository available and not the fedora and updates repositories, although I am trying to load them.

atodorov: @bcl I am starting to think this better become an integration test. What do you think ?

atodorov: Scratch that. I've managed to rework this test and now it tries to delete the repository which comes configured with the API server.

bcl: I'm not seeing that commit, it's still trying to delete "test-system-repo" during the integration test.

atodorov: In my previous comment I meant "test-system-repo" (before that I had a different implementation). 

The latest update tries several well known source names and skips them if they aren't available in the initial sources list. That makes the test fault  tolerant and you can execute it both as unit test and as an integration test. In integration test mode I get a failure which doesn't seem related to the current changes.

PR #395, opened 2020-03-25
distro: define valid image types for the arches

msehnout:
Certain image types are not supported on certain architectures. For
example VMWare works only on x86_64, so it is not possible nor necessary
to produce vmdk images for any other architecture. The same goes to
cloud-provider specific images, we only support architectures they
support, e.g. AWS: x86_64 and aarch64, but not ppc64le or s390x.

This patch move the imageTypes dictionary to the appropriate place which
is in the architecture specific structure. It also merges the two "arch"
structures because it wasn't clear which one to use for this purpose.
It does the same for the imageType structures.

Only Fedora distros are covered in this PR.

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -31,11 +31,6 @@ type Distro interface {
 	// Returns an object representing the given architecture as support
 	// by this distro.
 	GetArch(arch string) (Arch, error)
-
-	// Returns the canonical filename and MIME type for a given output
-	// format. `outputFormat` must be one returned by
-	// TODO: remove when redundant
-	FilenameFromType(outputFormat string) (string, string, error)
 }
 
 // An Arch represents a given distribution's support for a given architecture.
--- a/internal/distro/fedora31/distro.go
+++ b/internal/distro/fedora31/distro.go
@@ -24,6 +24,7 @@ type Fedora31 struct {
 }
 
 type imageType struct {
+	arch             *arch
 	name             string
 	mimeType         string
 	packages         []string
@@ -43,12 +44,6 @@ type arch struct {
 	buildPackages      []string
 	uefi               bool
 	imageTypes         map[string]imageType
-}
-
-type fedora31ImageType struct {
-	name      string
-	arch      *arch
-	imageType *imageType
 }
 
 func (d *Fedora31) ListArchs() []string {
@@ -87,52 +82,48 @@ func (a *arch) GetImageType(imageType string) (distro.ImageType, error) {
 	if !exists {
 		return nil, errors.New("invalid image type: " + imageType)
 	}
-
-	return &fedora31ImageType{
-		name:      imageType,
-		arch:      a,
-		imageType: &t,
-	}, nil
+	t.arch = a
+	return &t, nil
 }
 
-func (t *fedora31ImageType) Name() string {
+func (t *imageType) Name() string {
 	return t.name
 }
 
-func (t *fedora31ImageType) Filename() string {
-	return t.imageType.name
+func (t *imageType) Filename() string {
+	return t.name
 }
 
-func (t *fedora31ImageType) MIMEType() string {
-	return t.imageType.mimeType
+func (t *imageType) MIMEType() string {
+	return t.mimeType
 }
 
-func (t *fedora31ImageType) Size(size uint64) uint64 {
+func (t *imageType) Size(size uint64) uint64 {
 	const MegaByte = 1024 * 1024
 	// Microsoft Azure requires vhd images to be rounded up to the nearest MB
 	if t.name == "vhd" && size%MegaByte != 0 {
 		size = (size/MegaByte + 1) * MegaByte
 	}
 	if size == 0 {
-		size = t.imageType.defaultSize
+		size = t.defaultSize
 	}
 	return size
 }
 
-func (t *fedora31ImageType) BasePackages() ([]string, []string) {
-	packages := t.imageType.packages
-	if t.imageType.bootable {
+func (t *imageType) BasePackages() ([]string, []string) {
+	packages := t.packages
+	if t.bootable {
 		packages = append(packages, t.arch.bootloaderPackages...)
 	}
 
-	return packages, t.imageType.excludedPackages
+	return packages, t.excludedPackages
 }
 
-func (t *fedora31ImageType) BuildPackages() []string {
+func (t *imageType) BuildPackages() []string {
 	return append(t.arch.distro.buildPackages, t.arch.buildPackages...)
 }
 
-func (t *fedora31ImageType) Manifest(c *blueprint.Customizations,
+func (t *imageType) Manifest(c *blueprint.Customizations,
 	repos []rpmmd.RepoConfig,
 	packageSpecs,
 	buildPackageSpecs []rpmmd.PackageSpec,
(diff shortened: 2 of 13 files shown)


internal/distro/distro.go
```diff
@@ -35,7 +35,7 @@ type Distro interface {
 	// Returns the canonical filename and MIME type for a given output
 	// format. `outputFormat` must be one returned by
 	// TODO: remove when redundant
-	FilenameFromType(outputFormat string) (string, string, error)
+	FilenameFromType(arch, outputFormat string) (string, string, error)
```
teg: Hm, if this function takes an `arch` argument, then it means we have all the information to resolve the real `ImageType` object in the caller, and imply call `Filename()` on that. There was a reason I thought we couldn't, but seeing your patch, I think it should be ok. See below.

internal/weldr/api.go
```diff
@@ -1735,7 +1735,7 @@ func (api *API) composeImageHandler(writer http.ResponseWriter, request *http.Re
 
 	imageBuild := compose.ImageBuilds[0]
 	imageType, _ := imageBuild.ImageType.ToCompatString()
-	imageName, imageMime, err := api.distro.FilenameFromType(imageType)
+	imageName, imageMime, err := api.distro.FilenameFromType(api.arch.Name(), imageType)
```
teg: This is the only user left of `FilenameFromType()` and I would very much like to get rid of it ;) So far I did not, as the `imageBuild` struct did not contain an `Architecture` field, so we could not resolve this correctly. However, I think what you are doing here is ok, but I would instead use `api.arch` to resolve to the real `imageType` and then call `Filename` on that. Note that there is a chance this could be confusing: if builds are scheduled via the RCM api, then they may not match the architecture pinned in `api`, so this would either be wrong, or the `GetImageType()` might even fail. I don't think we should block the rest of this patch on that though, as that is anyway a problem to solve for another day.

internal/weldr/api.go
```diff
@@ -1749,16 +1749,20 @@ func (api *API) composeImageHandler(writer http.ResponseWriter, request *http.Re
 
 	imageBuild := compose.ImageBuilds[0]
 	imageType, _ := imageBuild.ImageType.ToCompatString()
-	imageName, imageMime, err := api.distro.FilenameFromType(imageType)
-
+	archStruct, err := api.distro.GetArch(api.arch.Name())
+	common.PanicOnError(err) // This must not fail because api.arch is predefine, not a user input
```
teg: You can just use `api.arch` directly. These sort of "errors that can never happen" are usually a sign that something is wrong.

internal/distro/fedora31/distro.go
```diff
@@ -396,7 +351,61 @@ func New() *Fedora31 {
...
+			"x86_64": {
+				name: "x86_64",
+				bootloaderPackages: []string{
+					"grub2-pc",
+				},
+				buildPackages: []string{
+					"grub2-pc",
+				},
+				imageTypes: map[string]imageType{
+					"ami": amiImgType,
```
teg: I'd use `amiImgType.name` as the key, rather than typing out the string, that way we only define the string in one place, and don't risk typos. Alternatively, add a helper that takes image types as input and returns the map (by doing the `t.Name()` lookup internally to get the keys).

internal/distro/fedora31/distro.go
```diff
@@ -23,7 +23,8 @@ type Fedora31 struct {
 	buildPackages []string
 }
 
-type imageType struct {
```
teg: You could actually keep the name `imageType` (and same for `arch`, i just realised now), that would make it easier to copy into subsequent distros.

internal/distro/fedora31/distro.go
```diff
@@ -366,48 +364,55 @@ func New() *Fedora31 {
...
-					"qcow2": qcow2ImageType,
-					"openstack": openstackImgType,
-					"tar": tarImgType,
-					"vhd": vhdImgType,
-				},
-			},
+		arches: map[string]fedora31Arch{},
+	}
+
+	x8664 := fedora31Arch{
```
teg: Not `x86_64`?

msehnout: Camel case linter is complaining about the underscore.

internal/distro/fedora31/distro.go
```diff
@@ -366,48 +364,55 @@ func New() *Fedora31 {
...
+		name: "x86_64",
+		bootloaderPackages: []string{
+			"grub2-pc",
 		},
+		buildPackages: []string{
+			"grub2-pc",
+		},
+		imageTypes: map[string]fedora31ImageType{},
+	}
+	x8664.imageTypes = map[string]fedora31ImageType{
```
teg: How about introducing a new method
```golang
func(a *fedora31Arch)addImageType(imageType fedora31ImageType) {
    a.imageTypes[imageType.name] = &fedoraImageType{
        arch: a,
        name: imageType.name,
        /* and so on for the rest of the normal copy */
    }
}
```
I was thinking that if really all but the `arch` pointer is shared, we might have wanted to keep the inner imageType as a separate struct, and just point to it, but I think conceptually we may end up with the imageTypes being really different per-arch, so the fact that for now they are just copies could be an implementation detail.

cc: @larskarlitski

teg: One final thought: I think the type of the struct we use to list all the output types i in `New()` should be something just local to that function, and not include an `arch` pointer. In the future, we could imagine `addImageType()` doing more than just coping, and in particular having completely different fields than the structs we use for conveniently writing the definitions down.

internal/distro/fedora31/distro.go
teg: Ah, fair enough.

internal/distro/fedora31/distro.go
msehnout: Alternative solution available here:
https://github.com/msehnout/osbuild-composer/commit/d91846509c33417371d937dea0565f647a485192

msehnout: It's not particularly pleasant to read either, but it might be a bit shorter at least.

msehnout: And also a version with the helpers you proposed:
https://github.com/msehnout/osbuild-composer/commit/4e815057554d76c426f985f972196710d51e027a
But I'm wondering if the version with the wrappers wasn't easier to read. Maybe just rename them to not be so confusing?

ondrejbudai: I like this one: msehnout@357858c. Generally, it's imho best to initialize data structures in one step allowing us to write more immutable code (Go is bad at it, sigh...). However, it's impossible to achieve this in `distro` and friends, because it's basically a tree with reverse links, so it cannot be initialized in one step.

I would change the `addArchList` method a bit:
```golang
func (d *Fedora31) setArchList(arches ...arch) {
	d.arches = map[string]arch{}
	for _, arch := range arches {
		arch.distro = d
		d.arches[arch.name] = arch
	}
}
```

1) rename it to `setArchList`
2) use `...` thingy thing in `arches` argument (I cannot recall the name of `...`)
3) clear the map before adding all the arches

msehnout: Thanks for the suggestion!

I've created one more iteration of the "wrapper struct" idea. It is described in more detail in the commit message:
https://github.com/msehnout/osbuild-composer/commit/20e97b47fc54d6da880034daccb24ee2637210c1

and I'm promise I'm done producing more versions of the same code :D

teg: > I like this one: [msehnout/osbuild-composer@357858c](https://github.com/msehnout/osbuild-composer/commit/357858c).

I also like this the most, and I agree with Ondrej's suggestions.

>     1. rename it to `setArchList`

Or even `setArchs()`


Though one part of it will not work, you are changing the passed in structs, but then reusing them later. You need something like:

```go
func (d *Fedora31) setArchs(arches ...arch) {
	d.arches = map[string]arch{}
	for _, arch := range arches {
		d.arches[arch.name] = &arch{
                    distro: d,
                    name: arch.name,
                    /* etc for all the other fields */
		}
	}
}
```

ondrejbudai: Also, I think we should agree on spelling of `arch` plural. Is it `archs` or `arches`? We have both in the current codebase and that's bad. I think I like `arches` more, it's easier to pronounce.

teg: I did an poll in the channel, and @major, @bcl and @jrusz all agree: "arches" it is (lots of things to change, in that case).

Issue #386, opened 2020-03-24
Error handling in blueprint.go


jrusz:
The returns in this function here: 
https://github.com/osbuild/osbuild-composer/blob/master/internal/blueprint/blueprint.go#L43 
are never called since you can't pass it a wrong `Blueprint` type. It's not possible to cover it with unit tests the way it's written now. 

I want to propose a change where those returns will be remove and `err` will be returned on the last line where it can be handled by the unit test. 

The `err` is used there twice and my first attempt is probably not the right solution to this so I'm looking forward to some comments. I can think of returning `err1` and `err2` but that might be something we don't want.

atodorov:
> The returns in this function here:
https://github.com/osbuild/osbuild-composer/blob/master/internal/blueprint/blueprint.go#L43
are never called since you can't pass it a wrong Blueprint type. It's not possible to cover it with unit tests the way it's written now.

In the case that passing invalid values isn't possible then you are looking at mocking. That is changing behavior of external dependencies. In this case `json.Marshall` is external to `DeepCopy`, which is the subject under test.

Mocking `json.Marshall` to always return an error will cause the traversal of lines 45 + 46. Then you can assert on the returned values.

For some high-level reference on the subject see,
http://atodorov.org/blog/2019/04/05/the-art-of-unit-testing/, section
*What is the difference between a double, stub, mock, fake and spy?*

In this particular context it doesn't make a lot of sense to try hard to cover these couple of lines b/c we're not getting enough information to warrant ROI. However I still think it is a valuable exercise b/c the testing scenario is very simple and will allow for very clean understanding how to mock things in Go.

A quick search revealed these 2 articles but I don't know how accurate or usable they are to you:
https://blog.learngoprogramming.com/how-to-mock-in-your-go-golang-tests-b9eee7d7c266
https://deployeveryday.com/2019/10/08/golang-auth-mock.html

jrusz:
@atodorov this sounds promising, will look trough this, thanks!

teg:
I think, in general, when we have code paths that cannot be reached, and we are confident cannot be reached, the correct solution is to add a
```
if err != nil {
    panic()
}
```
That documents the intentions, and gives us a stack-trace in case our assumption were wrong. In this particular case, the only way for the marshaller to fail is if the passed type cannot be marshalled, which would then mean that the the panic would either always or never be triggered. So if we were to make a mistake and make `Blueprint` unmarshallable, then we would trigger the panic on every unit-test run.

[...]

PR #398, opened 2020-03-25
tests: complete coverage of blueprint.go

jrusz:
With this commit we should have 100% coverage. I would like to hear your opinion on this or suggestions to accomplish the same in a simpler way. This was mainly an exercise for me to try mocking. Thank you.

### DIFF
--- a/internal/blueprint/blueprint_test.go
+++ b/internal/blueprint/blueprint_test.go
@@ -1,6 +1,7 @@
 package blueprint
 
 import (
+	"errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"testing"
@@ -28,6 +29,27 @@ func TestDeepCopy(t *testing.T) {
 	// Modify the original
 	bpOrig.Packages[0].Version = "42.0"
 	require.Equalf(t, bpCopy.Packages[0].Version, "1.2.3", "Blueprint.DeepCopy failed, copy modified.")
+
+	// This simulates json.Marshal returning an error
+	myJsonMarshal := JsonMarshal
+	JsonMarshal = func(interface{}) ([]byte, error) {
+		return []byte("test"), errors.New("test")
+	}
+
+	assert.Panics(t, func() { _, err = bpOrig.DeepCopy() })
+
+	// Return the original function definition
+	JsonMarshal = myJsonMarshal
+
+	// This simulates json.Unmarshal returning an error
+	myJsonUnmarshal := JsonUnmarshal
+	defer func() { JsonUnmarshal = myJsonUnmarshal }()
+	JsonUnmarshal = func([]byte, interface{}) error {
+		return errors.New("test")
+	}
+
+	assert.Panics(t, func() { _, err = bpOrig.DeepCopy() })
+
 }
 
 func TestBlueprintInitialize(t *testing.T) {
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -38,16 +38,19 @@ type Group struct {
 	Name string `json:"name" toml:"name"`
 }
 
+var JsonMarshal = json.Marshal
+var JsonUnmarshal = json.Unmarshal
+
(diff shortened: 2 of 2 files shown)


internal/blueprint/blueprint_test.go
```diff
@@ -28,6 +29,27 @@ func TestDeepCopy(t *testing.T) {
...
 	bpOrig.Packages[0].Version = "42.0"
 	require.Equalf(t, bpCopy.Packages[0].Version, "1.2.3", "Blueprint.DeepCopy failed, copy modified.")
+
+	// This simulates json.Marshal returning an error
+	myJsonMarshal := JsonMarshal
+	JsonMarshal = func(interface{}) ([]byte, error) {
+		return []byte("test"), errors.New("test")
+	}
+
+	assert.Panics(t, func() { _, err = bpOrig.DeepCopy() })
```
jrusz: I don't actually need to get the output of `bpOrig.DeepCopy()` since I know this will return a `panic()` but it makes the linter unhappy if I don't.

teg:
Hm, this is clever, but ultimately I think it is not worth it. The test itself is more complicated than the thing we are testing, and not getting to 100% coverage due to well-understood exceptions, I think is ok. So, I would leave this as it is now that @larskarlitski added `panic()` in these paths.
There are after all plenty of other parts of the code that is far from 100% that we could rather focus on :)

bcl:
I agree with @teg, I don't think it's worth the extra complexity. Any time you have to start poking holes in your running code just to make tests happy is a bad thing, especially when the benefit here is so small.

But thanks for figuring out how to do it!

internal/blueprint/blueprint_test.go
```diff
@@ -28,6 +29,27 @@ func TestDeepCopy(t *testing.T) {
...
+
+	// Return the original function definition
+	JsonMarshal = myJsonMarshal
```
atodorov: FTR this should be `origJsonMarshall` to make it more obvious. The clean-up (teardown) part I think should be deferred immediatelly after the setup & before the assert a few lines above.

internal/blueprint/blueprint_test.go
```diff
@@ -28,6 +29,27 @@ func TestDeepCopy(t *testing.T) {
...
+
+	// This simulates json.Unmarshal returning an error
+	myJsonUnmarshal := JsonUnmarshal
+	defer func() { JsonUnmarshal = myJsonUnmarshal }()
+	JsonUnmarshal = func([]byte, interface{}) error {
+		return errors.New("test")
+	}
+
+	assert.Panics(t, func() { _, err = bpOrig.DeepCopy() })
```
atodorov: This is testing another portion of the code so should be a scenario on its own, that is a separate test function with lots of copy & paste between. Otherwise your testing code becomes more brittle and less independent of one another.

atodorov:
@jrusz also FYI, this "Closes #386". You may want to add that to your commit log.

atodorov:
I agree with @teg and @bcl that it isn't worth it in this case. 100% coverage is nothing special and we don't gain much more confidence when we have it. It really depends on what is being tested and how important/risky it is.

Honestly I am surprised how nicely this dependency injection turned out. The concept is very easy and elegant to apply in interpreted languages like Python and here I thought we would need a lot more abstractions to get this working.

The exercise portion of #386 is now complete so I am going to close this PR and the issue.

jrusz:
> Honestly I am surprised how nicely this dependency injection turned out. The concept is very easy and elegant to apply in interpreted languages like Python and here I thought we would need a lot more abstractions to get this working.

Yes in the end this was quite easy but took me a lot of time to understand how to cover this particular case because almost all of the mock examples I found online were about methods for structures imported from a third party or something like that which indeed looked more complicated because it involved creating interfaces and then mocking each method the interface had. 

I agree with all of you that this is too much to cover just one return that should never happen, my main goal was to figure out how to do this and collect your feedback, which is exactly what happened, thank you :)

PR #400, opened 2020-03-26
tests/image: test ami images on AWS

### LINKED ISSUE #337: tests/image: Test ami images on AWS
Currently, we are only booting the ami images locally using qemu. We should switch the ami test to do a proper test in the real AWS environment. Currently, there's a minor issue with it: The image tests use `osbuild` instead of `osbuild-composer` to do the image build, because it's currently impossible to disable updates repository in composer and we don't want the tested images to change all the time. Therefore, we don't have a simple way to upload the image to AWS. For now, let's just directly use the code from `internal/upload` to do the upload and we can switch to `osbuild-composer` later.

What is needed:
- create a new boot type in image tests
- use the functions in `internal/upload` to do the upload
- create and run a VM from the uploaded image in AWS
- try to connect to it using ssh and get the machine's status from `systemctl --wait is-system-running`
- tear everything down

Also, a workaround/fix for #257 is needed.

ondrejbudai:
Blocked by #460, without this, the s3 objects won't be deleted.

This PR adds a lot of small cleanups to the image tests and the support for testing the ami images on the actual AWS. The AWS tests won't be run if secrets are not set as environment variables, qemu is used as a fallback in this case.

It should be fairly simple to switch the aws upload to be done by osbuild-composer when we decide to use it in the image tests.

To answer @atodorov 's question, in the end, I decided to use a freshly generated key-pair for each AWS upload. I think it's definitely better for security purposes. I understand that we may have a use-case for debugging the running instance and this will make it hard, but I feel this should be done as a follow-up when we know how do we want to implement this.

I decided to use cloud-init's userData to insert the public key. We could also use the EC2 key-pairs, but that would require a bit more setup and teardown. Both methods use cloud-init in the end, so I don't think there's a big difference in these two variants and thus I prefer the simpler one.

The tests are designed to be self-contained - nothing apart from valid credentials and correctly set `vmimport` role is required.

Closes #337

### DIFF
--- a/cmd/osbuild-image-tests/aws.go
+++ b/cmd/osbuild-image-tests/aws.go
@@ -1,0 +1,281 @@
+// +build integration
+
+package main
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/osbuild/osbuild-composer/internal/upload/awsupload"
+)
+
+type awsCredentials struct {
+	AccessKeyId     string
+	SecretAccessKey string
+	Region          string
+	Bucket          string
+}
+
+// getAWSCredentialsFromEnv gets the credentials from environment variables
+// If none of the environment variables is set, it returns nil.
+// If some but not all environment variables are set, it returns an error.
+func getAWSCredentialsFromEnv() (*awsCredentials, error) {
+	accessKeyId, akExists := os.LookupEnv("AWS_ACCESS_KEY_ID")
+	secretAccessKey, sakExists := os.LookupEnv("AWS_SECRET_ACCESS_KEY")
+	region, regionExists := os.LookupEnv("AWS_REGION")
+	bucket, bucketExists := os.LookupEnv("AWS_BUCKET")
+
+	// Workaround Travis security feature. If non of the variables is set, just ignore the test
+	if !akExists && !sakExists && !bucketExists && !regionExists {
+		return nil, nil
+	}
+	// If only one/two of them are not set, then fail
+	if !akExists || !sakExists || !bucketExists || !regionExists {
+		return nil, errors.New("not all required env variables were set")
+	}
+
+	return &awsCredentials{
+		AccessKeyId:     accessKeyId,
+		SecretAccessKey: secretAccessKey,
+		Region:          region,
+		Bucket:          bucket,
+	}, nil
+}
+
+// encodeBase64 encodes string to base64-encoded string
+func encodeBase64(input string) string {
+	return base64.StdEncoding.EncodeToString([]byte(input))
+}
+
+// createUserData creates cloud-init's user-data that contains user redhat with
+// the specified public key
+func createUserData(publicKeyFile string) (string, error) {
+	publicKey, err := ioutil.ReadFile(publicKeyFile)
+	if err != nil {
+		return "", fmt.Errorf("cannot read the public key: %#v", err)
+	}
+
+	userData := fmt.Sprintf(`#cloud-config
+user: redhat
+ssh_authorized_keys:
+  - %s
+`, string(publicKey))
+
+	return userData, nil
+}
+
+// wrapErrorf returns error constructed using fmt.Errorf from format and any
+// other args. If innerError != nil, it's appended at the end of the new
+// error.
+func wrapErrorf(innerError error, format string, a ...interface{}) error {
+	if innerError != nil {
+		a = append(a, innerError)
+		return fmt.Errorf(format+"\n\ninner error: %#s", a...)
+	}
+
+	return fmt.Errorf(format, a...)
+}
+
+// uploadImageToAWS mimics the upload feature of osbuild-composer.
+// It takes an image and an image name and creates an ec2 instance from them.
+// The s3 key is never returned - the same thing is done in osbuild-composer,
+// the user has no way of getting the s3 key.
+func uploadImageToAWS(c *awsCredentials, imagePath string, imageName string) error {
+	uploader, err := awsupload.New(c.Region, c.AccessKeyId, c.SecretAccessKey)
+	if err != nil {
+		return fmt.Errorf("cannot create aws uploader: %#v", err)
+	}
+
+	_, err = uploader.Upload(imagePath, c.Bucket, imageName)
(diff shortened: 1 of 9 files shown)


cmd/osbuild-image-tests/main_test.go
```diff
@@ -96,22 +96,22 @@ func splitExtension(path string) (string, string) {
 func testImageInfo(t *testing.T, imagePath string, rawImageInfoExpected []byte) {
 	var imageInfoExpected interface{}
 	err := json.Unmarshal(rawImageInfoExpected, &imageInfoExpected)
-	require.Nilf(t, err, "cannot decode expected image info: %#v", err)
+	require.NoErrorf(t, err, "cannot decode expected image info: %#v", err)
```
teg: For the uninitiated: what's the difference, why do you make this change?

tools/test-case-generators/format-request-map.json
```diff
@@ -1,7 +1,7 @@
 {
   "ami": {
     "boot": {
-      "type": "qemu-extract"
+      "type": "aws"
```
teg: This works as-is, even if we are still producing `raw.xz` images?

cmd/osbuild-image-tests/aws.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return prefix + id.String(), nil
+}
+
+var userData = `#cloud-config
+user: redhat
+ssh_authorized_keys:
+  - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC61wMCjOSHwbVb4VfVyl5sn497qW4PsdQ7Ty7aD6wDNZ/QjjULkDV/yW5WjDlDQ7UqFH0Sr7vywjqDizUAqK7zM5FsUKsUXWHWwg/ehKg8j9xKcMv11AkFoUoujtfAujnKODkk58XSA9whPr7qcw3vPrmog680pnMSzf9LC7J6kXfs6lkoKfBh9VnlxusCrw2yg0qI1fHAZBLPx7mW6+me71QZsS6sVz8v8KXyrXsKTdnF50FjzHcK9HXDBtSJS5wA3fkcRYymJe0o6WMWNdgSRVpoSiWaHHmFgdMUJaYoCfhXzyl7LtNb3Q+Sveg+tJK7JaRXBLMUllOlJ6ll5Hod
+`
```
atodorov: Where is this key coming from ? 

The script in lorax-composer's repo doesn't actually supply the keys like that, but instead tells RunInstances which KeyName (from the definitions available in AWS) to use. I think that is a bit closer to what happens in reality but I may be mistaken. 

Anyway I think I prefer having the test create/utilize the ssh key from the system executing it. That way a person would be able to ssh into the machine if it boots but something else fails for example.

ondrejbudai:
Still WIP, should be ready for review pretty soon I hope.

cmd/osbuild-image-tests/main_test.go
ondrejbudai: NoError() is more understandable (both in the code and in the output) than Nil() when asserting errors. Also the rest of the tests uses mostly NoError() when asserting errors, this change should lead to a better consistency.

cmd/osbuild-image-tests/aws.go
ondrejbudai: See the updated "PR body".

cmd/osbuild-image-tests/aws.go
```diff
@@ -0,0 +1,286 @@
...
+
+	return fmt.Errorf(format, a...)
+}
+
+// uploadImageToAWS mimics the upload feature of osbuild-composer.
+// It takes an image and an image name and creates an ec2 instance from them.
+// The s3 key is never returned - the same thing is done in osbuild-composer,
+// the user has no way of getting the s3 key.
+func uploadImageToAWS(c *awsCredentials, imagePath string, imageName string) error {
+	s3Key, err := generateRandomString("osbuild-image-tests-s3-")
```
atodorov, line 91: wouldn't it be better to match the UUID of the image ?

cmd/osbuild-image-tests/main_test.go
```diff
@@ -169,10 +180,10 @@ func trySSHOnce(ns netNS) error {
 // testSSH tests the running image using ssh.
 // It tries 20 attempts before giving up. If a major error occurs, it might
 // return earlier.
-func testSSH(t *testing.T, ns netNS) {
+func testSSH(t *testing.T, address string, privateKey string, ns *netNS) {
 	const attempts = 20
 	for i := 0; i < attempts; i++ {
-		err := trySSHOnce(ns)
+		err := trySSHOnce(address, privateKey, ns)
 		if err == nil {
```
atodorov, line 187: you can use require.NoError here as well

cmd/osbuild-image-tests/main_test.go
```diff
@@ -189,39 +200,111 @@ func testSSH(t *testing.T, ns netNS) {
...
+		if err != nil {
+			log.Printf("cannot delete the ec2 image, resources could have been leaked: %#v", err)
+		}
+	}()
+
+	// If the test is failed at this point, there's no point in continuing.
+	// The deferred cleanup will be called.
+	if t.Failed() {
+		t.FailNow()
+	}
```
atodorov, line 272: If you change line 255 to use require instead of assert won't this condition become obsolete ?

cmd/osbuild-image-tests/main_test.go
ondrejbudai: I want to pass the test immediately when `err == nil`, therefore `require.NoError` doesn't make much sense here.

cmd/osbuild-image-tests/aws.go
ondrejbudai: Yeah, let's do it as a follow-up?

cmd/osbuild-image-tests/main_test.go
ondrejbudai: I want to register the deferred function on line 261 and I couldn't find a cleaner way to write it, therefore I chose this a bit cumbersome, but reliable way.

atodorov: I think you should register the deferred function on line 258 before `require.NoError` b/c in case of error that will short-cirtcuit the test and you won'd perform cleanup, right ? Then this if condition will not be necessary.

ondrejbudai: I changed my mind, the error handling was very confusing and over-engineered, I pushed a new much simpler version, which should do basically the same.

cmd/osbuild-image-tests/aws.go
ondrejbudai: Fixed in the latest push.

Issue #402, opened 2020-03-26
Unused code in internal/blueprint/customizations.go


jrusz:
I am currently working on a test for the file in $SUBJ and I believe I came across some lines that will never be run. Please take a look at this: https://github.com/osbuild/osbuild-composer/blob/master/internal/blueprint/customizations.go#L75
Those returns are nearly in all of the functions but I think they'll never be called because these are the methods for `Customizations struct` and when I call them the structures will never be just one `nil`. The fields in the structure will have `nil` values but the mentioned `if` will not be called it would have to be `if c.GetHostname == nil` to actually go through that. If I remove it and leave with just `return c.Hostname` it will return `nil` if it's called upon an empty `Customizations struct`. 

So my suggestion is to remove those returns. I'll add it to my next PR if you agree.

I hope it's clear what I mean, I can demonstrate with some examples if it's not.

larskarlitski:
This seems to have been a refactoring mistake in 7957feff48c0a7f01205adf9615b1ae4a8678945. We were checking if the `Customizations` field on a blueprint was set.

I think it's fine to remove them. @teg?

ondrejbudai:
We cannot remove them, because `blueprint.Customizations` can indeed be true. In Go, it's just fine to call a method with `nil` receiver. See:
- https://tour.golang.org/methods/12
- https://play.golang.org/p/5ZxL80qq6c9

The `c == nil` path is indeed used in the case when user doesn't want any customizations. Therefore, the tests should cover these paths.

Actually, we should refactor the customizations at some point. The code was moved around several times now and a deduplication would be nice. However, I think it isn't particularly easy to do it.

larskarlitski:
Oh, I thought we checked in the distros whether it is set to `nil`, but we don't indeed. Thanks @ondrejbudai and sorry for the confusion.

jrusz:
@ondrejbudai Right I just tested calling the methods on `Blueprint` which has no `Customizations` and it goes that way. Thanks, I'll add it to my PR :)

PR #403, opened 2020-03-26
tests: add coverage for internal/blueprint/customizations.go

jrusz:
This is my WIP on `customizations.go`. I have a few questions I'd like to ask.

1. First one is mentioned here: https://github.com/osbuild/osbuild-composer/issues/402
1. Is it ok to use 1 element in arrays for testing or is it better to put at least two in there? Sometimes there are loops going trough that so maybe it would have some benefit but I'm not sure.
1. In the first function `TestCustomizationsHostname` I have actually put the customization into a `Blueprint` and asserted the call from there as well (L24). Is that something we want or is the assert from `Customizations` enough? (L23)

I would also like some general feedback if it looks ok or if I could do something better.

Thank you.

### DIFF
--- a/internal/blueprint/customizations_test.go
+++ b/internal/blueprint/customizations_test.go
@@ -1,0 +1,267 @@
+package blueprint
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestGetHostname(t *testing.T) {
+
+	var expectedHostname = "Hostname"
+
+	TestCustomizations := Customizations{
+		Hostname: &expectedHostname,
+	}
+
+	retHostname := TestCustomizations.GetHostname()
+	assert.Equal(t, &expectedHostname, retHostname)
+
+}
+
+func TestGetKernel(t *testing.T) {
+
+	expectedKernel := KernelCustomization{
+		Append: "--test",
+	}
+
+	TestCustomizations := Customizations{
+		Kernel: &expectedKernel,
+	}
+
+	retKernel := TestCustomizations.GetKernel()
+
+	assert.Equal(t, &expectedKernel, retKernel)
+}
+
+func TestSSHKey(t *testing.T) {
+
+	expectedSSHKeys := []SSHKeyCustomization{
+		SSHKeyCustomization{
+			User: "test-user",
+			Key:  "test-key",
+		},
+	}
+	TestCustomizations := Customizations{
+		SSHKey: expectedSSHKeys,
+	}
+
+	retUser := TestCustomizations.GetUsers()[0].Name
+	retKey := *TestCustomizations.GetUsers()[0].Key
+
+	assert.Equal(t, expectedSSHKeys[0].User, retUser)
+	assert.Equal(t, expectedSSHKeys[0].Key, retKey)
+
+}
+
+func TestGetUsers(t *testing.T) {
+
+	Desc := "Test descritpion"
+	Pass := "testpass"
+	Key := "testkey"
+	Home := "Home"
+	Shell := "Shell"
+	Groups := []string{
+		"Group",
+	}
+	UID := 123
+	GID := 321
+
+	expectedUsers := []UserCustomization{
+		UserCustomization{
+			Name:        "John",
+			Description: &Desc,
+			Password:    &Pass,
+			Key:         &Key,
+			Home:        &Home,
+			Shell:       &Shell,
+			Groups:      Groups,
+			UID:         &UID,
+			GID:         &GID,
+		},
+	}
+
+	TestCustomizations := Customizations{
+		User: expectedUsers,
+	}
+
+	retUsers := TestCustomizations.GetUsers()
+
+	assert.ElementsMatch(t, expectedUsers, retUsers)
+}
+
+func TestGetGroups(t *testing.T) {
+
+	GID := 1234
+	expectedGroups := []GroupCustomization{
+		GroupCustomization{
+			Name: "TestGroup",
+			GID:  &GID,
+		},
+	}
+
+	TestCustomizations := Customizations{
+		Group: expectedGroups,
+	}
+
+	retGroups := TestCustomizations.GetGroups()
+
+	assert.ElementsMatch(t, expectedGroups, retGroups)
+}
+
+func TestGetTimezoneSettings(t *testing.T) {
+
+	expectedTimezone := "testZONE"
+	expectedNTPServers := []string{
+		"server",
+	}
+
+	expectedTimezoneCustomization := TimezoneCustomization{
+		Timezone:   &expectedTimezone,
+		NTPServers: expectedNTPServers,
+	}
+
+	TestCustomizations := Customizations{
+		Timezone: &expectedTimezoneCustomization,
+	}
+
+	retTimezone, retNTPServers := TestCustomizations.GetTimezoneSettings()
+
+	assert.Equal(t, expectedTimezone, *retTimezone)
+	assert.Equal(t, expectedNTPServers, retNTPServers)
+
+}
+
+func TestGetPrimaryLocale(t *testing.T) {
+
+	expectedLanguages := []string{
+		"enUS",
+	}
+	expectedKeyboard := "en"
+
+	expectedLocaleCustomization := LocaleCustomization{
+		Languages: expectedLanguages,
+		Keyboard:  &expectedKeyboard,
+	}
+
+	TestCustomizations := Customizations{
+		Locale: &expectedLocaleCustomization,
+	}
+
+	retLanguage, retKeyboard := TestCustomizations.GetPrimaryLocale()
+
+	assert.Equal(t, expectedLanguages[0], *retLanguage)
+	assert.Equal(t, expectedKeyboard, *retKeyboard)
+}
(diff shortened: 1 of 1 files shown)


atodorov:
> Is it ok to use 1 element in arrays for testing or is it better to put at least two in there? Sometimes there are loops going trough that so maybe it would have some benefit but I'm not sure.

@jrusz for the number of elements you can use some heuristics to decide how many. I think what is called boundary value analysis (BVA) or an interpretation of it will help you. Like @ondrejbudai says you've got the case where a loop body never executes and jumps right after the loop and another case where you go into the loop body and iterate 1 or more times. The important (lower) boundary here is 1 - >=1 element or <1 element.

@ondrejbudai's comment about line coverage is on point. I will even go ahead and say coverage itself can be a very misguided metric when it comes to testing (see the 101 possible coverages listed here http://kaner.com/pdfs/negligence_and_testing_coverage.pdf). The only information it gives you is whether or not a particular code path has been executed! Once it is executed you don't get more information.

Most language tools that measure code coverage focus on line and branch coverage, sometimes (e.g. Haskell) they also report partial coverage (lazy evaluation). Coveralls itself keeps track of branch coverage with small annotations to the right side of the lines when you look at the web interface.

How many times you want to iterate over a loop which you are trying to test really depends on what is inside the loop body, especially if conditions around break and continue operators. In your case GetUsers and GetGroups will iterate when `if len(c.SSHKey) > 0` so that's also an indication of your boundary value.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
+package blueprint
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestCustomizationsHostname(t *testing.T) {
```
atodorov: I suggest using the name of the function under test as part of the name of the test func, e.g. `TestCustomizationsGetHostName` or even `TestGetHostName` b/c the filename is `blueprins/customizations_test.go` which already points to `blueprints/customizations.go` as the general test subject.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	var myHostname = "MyHostname"
```
atodorov: better name will be expectedHostname

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	bpOrig := Blueprint{
+		Name:           "customizations-test",
+		Description:    "Testing customizations functions",
+		Version:        "0.0.1",
+		Customizations: &TestCustomizations,
+	}
+
+	assert.Equalf(t, &myHostname, TestCustomizations.GetHostname(), "The hostname is different")
```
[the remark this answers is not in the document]
atodorov: +1 here, one of the benefits we get from the assert package is that it tells you what was expected (the 1st argument) and what was the actual value (2nd argument) in case of failure. The error message doesn't really bring out any additional information. 

You can experiment with this by changing the expected hostname after calling GetHostname & before asserting the result to see how it looks like.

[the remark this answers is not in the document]
atodorov: @ondrejbudai it is still not clear to me what the difference between .Equal and .Equalf is. It looks like both accept a format string for error messages. Maybe we need to document this in test/README once we figure it out.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+	assert.Equalf(t, &myHostname, bpOrig.Customizations.GetHostname(), "The hostname is different")
```
atodorov: Question: why do we have 2 asserts which call GetHostName via 2 different paths? I don't understand the scenario here.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+}
+
+func TestKernelCustomization(t *testing.T) {
+
+	myKernel := KernelCustomization{
```
atodorov: Same comment for variable name here (and everywhere else). It would be much better to prefix variable names with `expected`.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+		Append: "--test",
+	}
+
+	TestCustomizations := Customizations{
+		Kernel: &myKernel,
+	}
+
+	assert.Equalf(t, &myKernel, TestCustomizations.GetKernel(), "The kernel is different")
+}
```
atodorov: Like @ondrejbudai says, the nil scenario applies here.

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+			User: "test-user",
+			Key:  "test-key",
+		},
+	}
+	TestCustomizations := Customizations{
+		SSHKey: mySSHKeys,
+	}
+
+	assert.Equal(t, mySSHKeys[0].User, TestCustomizations.GetUsers()[0].Name, "The user names are different")
+	assert.Equal(t, mySSHKeys[0].Key, *TestCustomizations.GetUsers()[0].Key, "The users keys are different")
```
atodorov: IMO assigning the actual value to a temporary variable and asserting on that will be more clear. However I don't understand the following:
- why does the 2nd call to GetUsers uses `*` in front of the result? I'm sure the answer has to do with pointers but from what I can tell both Name and Key fields have the same data type and I don't understand why that applies to one of them and not the other.

atodorov:
> In the first function TestCustomizationsHostname I have actually put the customization into a Blueprint and asserted the call from there as well (L24). Is that something we want or is the assert from Customizations enough? (L23)

I see the answer to one of my questions in the code review comments.  You don't actually need to use the customizations via blueprints for the test. One reason like already mentioned is that customizations are independent. The other is that in unit tests you are trying to exercise the functionality under test as independently as possible. Where a "unit" is defined as a function or a class. Here you don't want possible modifications in how blueprints consume customizations to affect customizations_test.go.  If there is a blueprint+customization behavior that needs testing this should be a separate task.

internal/blueprint/customizations_test.go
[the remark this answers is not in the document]
jrusz: Yes that should be enough, thanks.

internal/blueprint/customizations_test.go
jrusz: That is actually related to my 3rd question in my first comment. I can access the `Customizations struct` on it's own or put in inside a `Blueprint struct` and call the method from there. My question was if there's any practical difference or if that makes sense to check it both ways, but @ondrejbudai has already answered me that it should not bring any benefit.

internal/blueprint/customizations_test.go
jrusz: Well it only works when I use a pointer to that and that seems to be because of how `GetUsers()` function is written.. It actually returns `Users struct` where it appends the `SSHKeys` if they exist. `Name` is just a regular `string` but `Key` is a `*string` so that's probably why.

jrusz:
@atodorov @ondrejbudai Thank you very much for your feedback. I won't reply to all the comments individually but I will address them in the next update :)

internal/blueprint/customizations_test.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	assert.ElementsMatch(t, myUsers, TestCustomizations.GetUsers(), "The users are different")
+}
+
+func TestGroupCustomizations(t *testing.T) {
+
+	GID := 1234
+
+	myGroup := []GroupCustomization{
```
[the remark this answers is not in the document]
atodorov: `expectedGroups`

internal/blueprint/customizations_test.go
atodorov: I see the difference now. Thanks.

internal/blueprint/customizations_test.go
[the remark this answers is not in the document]
jrusz: > @ondrejbudai it is still not clear to me what the difference between .Equal and .Equalf is. It looks like both accept a format string for error messages. Maybe we need to document this in test/README once we figure it out.

I have found one difference. `.Equal` **can** accepted a formatted string but `.Equalf` **requires** it. I don't know what the use case for this would be though.

jrusz:
I pushed an improved version. I assign returned values in a `ret*` variable before assert, it should look a bit more readable. So now I am asserting an `expected*` variable with `ret*` variable. I also changed the function names to reflect the names of the functions that they test. 

I still have to figure out how to decide about going trough the loops and cover a few more uncovered paths.

internal/blueprint/customizations_test.go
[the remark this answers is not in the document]
teg: If you use a format string, you should use `.Equalf`, the reason is that some linters will compalain if you use one in `.Equal`, and that is why `.Equalf` was added (and I suppose the old behaviour was kept for compatibility).

jrusz:
> LGTM now, let's focus on merging this first and then you can work on remaining edge cases where necessary.

Thanks, I pushed my latest changes. Not everything is covered but should be good enough now. Should be ready to merge.

PR #404, opened 2020-03-27
Rearrange the integration tests so they can be shared and run as unit tests

bcl:


ondrejbudai:
Quick note: you need to do the same changes in `osbuild-composer.spec` as you did in `golang-github-osbuild-composer.spec`. Sorry for that, we will drop the old one as soon as we stop doing releases for F31.

bcl:
> Quick note: you need to do the same changes in `osbuild-composer.spec` as you did in `golang-github-osbuild-composer.spec`.

Thanks, hadn't realized there were 2 of them for now.

PR #412, opened 2020-03-27
rcm: rework the structure of the compose requests

teg:
See the commits for more details.


teg:
That was not a well formed sentence, sorry. We don't want a list of architectures and a separate list of image types. That would presumably mean that we want to build images for every combination, but we want to explicitly state which pairs to use.

PR #418, opened 2020-03-30
.github: switch to codecov.io

larskarlitski:
coveralls doesn't work from GitHub actions. Its "github" service type
uses the GITHUB_TOKEN from the action, which only has read access when
invoked from a forked repository.

codecov gets this right: it validates that an uploaded coverage file
originated from a GitHub action run by asking GitHub, and then uses its
OAuth credentials (through the Marketplace App) to comment and set
status.

Also, coveralls requires a third-party package to convert go's coverage
report to a format it understands. codecov detects the format
server-side. It also handles go's coverage format better: it highlights
lines with "some coverage" in yellow (coveralls has no concept of this).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,19 +40,14 @@ jobs:
       - name: Install golangci-lint
         run: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.23.7
 
-      - name: Install goveralls
-        run: GO111MODULE=off go get github.com/mattn/goveralls
-
       - name: Run golangci-lint
         run: $(go env GOPATH)/bin/golangci-lint run
 
       - name: Run unit tests
-        run: go test -v -race -covermode atomic -coverprofile=profile.cov ./...
+        run: go test -v -race -covermode atomic -coverprofile=coverage.txt ./...
 
-      - name: Send coverage to coveralls.io
-        env:
-          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


larskarlitski:
Seems this is working as expected. I'll request some reviews from people that are most affected by this change. Please wait for all of them to chime in before merging.

teg:
Sounds reasonable to me. We should also update the readme.

larskarlitski:
@teg 
> Sounds reasonable to me. We should also update the readme.

Thanks! Pushed an update to change the link.

@jrusz 
> Looks good, it's nice that I can see the report among the other check and get a link to it from there.

It seems to only do that after all CI runs have finished, even though I've told it not to in the config file. Not sure what's going on yet.

jrusz:
> It seems to only do that after all CI runs have finished, even though I've told it not to in the config file. Not sure what's going on yet.

Ah now I see that. It would be better if it could show right away...

jikortus:
The idea looks good to me. I wanted to investigate further how the reports look like, but I found out I can't access the reports for individual files (_File/folder was not found on base commit. No compare to display. _) - will this change with further commits? It also seems that the reported coverage lowered, but I guess that's just a matter of how the coverage is calculated.

larskarlitski:
> The idea looks good to me. I wanted to investigate further how the reports look like, but I found out I can't access the reports for individual files (_File/folder was not found on base commit. No compare to display._ ) - will this change with further commits?

You need to be logged in to see details. Not many are there yet, because there's nothing to compare to yet. Once we have results uploaded for master, the interface will be more useful.

> It also seems that the reported coverage lowered, but I guess that's just a matter of how the coverage is calculated.

Yeah... for example, codecov deals with golang's cover output more correctly by not counting lines that have partial coverage. Not sure what's correct, but it *is* different.

PR #420, opened 2020-03-30
repositories/RHEL: use repositories for the latest releases

teg:
Also include the GPG keys. When we first started testing we were
testing against nightlies, whose packages were not signed. But
when using released content (which is signed) the GPG keys are
required.


teg:
Travis was green, but did not report back.

Issue #419, opened 2020-03-30
Test coverage for internal/common


jrusz:
Add test coverage for the following files in internal/common:

- [x] compose_result
- [ ] helpers
- [ ] states
- [ ] types

jrusz:
Starting work on `compose_result`

teg:
Just a heads-up, some of these things may be re-factored/removed in the not-too-distant future.

jrusz:
Thanks for the info @teg . I guess I'll move along to more permanent pieces of code for now then :)

jrusz:
Some work has been done here and since it's likely to change soon I'm going to close this now.

PR #428, opened 2020-03-31
tests: Add coverage for compose_result

jrusz:
I also changed the structure definition that was there, I hope it still works the same.
Refs #419

### DIFF
--- a/internal/common/compose_result.go
+++ b/internal/common/compose_result.go
@@ -6,32 +6,33 @@ import (
 	"io"
 )
 
+type assembler struct {
+	Name    string          `json:"name"`
+	Options json.RawMessage `json:"options"`
+	Success bool            `json:"success"`
+	Output  string          `json:"output"`
+}
+
+type stage struct {
+	Name    string          `json:"name"`
+	Options json.RawMessage `json:"options"`
+	Success bool            `json:"success"`
+	Output  string          `json:"output"`
+}
+
+type build struct {
+	Stages  []stage `json:"stages"`
+	TreeID  string  `json:"tree_id"`
+	Success bool    `json:"success"`
+}
+
 type ComposeResult struct {
-	TreeID   string `json:"tree_id"`
-	OutputID string `json:"output_id"`
-	Build    *struct {
-		Stages []struct {
-			Name    string          `json:"name"`
-			Options json.RawMessage `json:"options"`
-			Success bool            `json:"success"`
-			Output  string          `json:"output"`
-		} `json:"stages"`
-		TreeID  string `json:"tree_id"`
-		Success bool   `json:"success"`
-	} `json:"build"`
-	Stages []struct {
-		Name    string          `json:"name"`
-		Options json.RawMessage `json:"options"`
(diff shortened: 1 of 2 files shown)


internal/common/compose_result.go
```diff
@@ -6,32 +6,33 @@ import (
...
 )
 
+type Assembler struct {
+	Name    string          `json:"name"`
+	Options json.RawMessage `json:"options"`
+	Success bool            `json:"success"`
+	Output  string          `json:"output"`
+}
+
+type Stages []struct {
```
ondrejbudai: I think I prefer having non-array `Stage` type, it's more common to define non-array types.

internal/common/compose_result_test.go
```diff
@@ -0,0 +1,56 @@
...
+		OutputID:  "OutputID",
+		Build:     &testBuild,
+		Stages:    testStage,
+		Assembler: &testAssembler,
+		Success:   true,
+	}
+
+	var b bytes.Buffer
+	assert.NoError(t, testComposeResult.Write(&b))
+	expectedMessage := "Build pipeline:\nStage testStage\n{\n  \"msg\": \"test\"\n}\n\nOutput:\nFinished\nStages:\nStage: testStage\n{\n  \"msg\": \"test\"\n}\n\nOutput:\nFinished\nAssembler testAssembler:\n{\n  \"msg\": \"test\"\n}\n\nOutput:\nDone\n"
```
ondrejbudai: I suggest using the raw string literal here: https://stackoverflow.com/questions/7933460/how-do-you-write-multiline-strings-in-go/7933487

atodorov:
> I also changed the structure definition that was there,

Why ?

jrusz:
> > I also changed the structure definition that was there,
> 
> Why ?

Well it was a nested structure before and it's kinda hard to initialize and there was no reason to have it that way so for the sake of simplicity I broke it apart.

internal/common/compose_result.go
jrusz: Alright, I'm still kinda confused about these types. I changed it to hopefully a better state.

internal/common/compose_result_test.go
jrusz: I just copied what it gave me and used it for comparison. I was thinking that this will just watch out for potential changes to this in the future. But it does look better with the formatted string, thanks!

internal/common/compose_result.go
```diff
@@ -6,32 +6,33 @@ import (
 	"io"
 )
 
+type Assembler struct {
```
ondrejbudai: I just realized we possibly don't want to have these helper types public. Can you make Assembler`, `Stage` and `Build` private? They are not really meant to be used on their own. Sorry for not pointing earlier.

internal/common/compose_result_test.go
ondrejbudai: I would do the same with copying. :)

internal/common/compose_result.go
jrusz: No worries, changed that. Thanks!

PR #429, opened 2020-03-31
Documentation Improvements

dvdhrm:
This improves a bit on the documentation all over the place. It mostly follows what we did in *osbuild*, as it looks like everyone liked it so far:

 * reformat markdown files in the repo (CONTIRBUTING, README, ..)
 * create new osbuild-composer(7) man-page
 * minor reshuffling in Makefile
 * create a NEWS.md file with changelogs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,104 @@
-COMMIT=$(shell git rev-parse HEAD)
+#
+# Maintenance Helpers
+#
+# This makefile contains targets used for development, as well as helpers to
+# aid automatization of maintenance. Unless a target is documented in
+# `make help`, it is not supported and is only meant to be used by developers
+# to aid their daily development work.
+#
+# All supported targets honor the `SRCDIR` variable to find the source-tree.
+# For most unsupported targets, you are expected to have the source-tree as
+# your working directory. To specify a different source-tree, simply override
+# the variable via `SRCDIR=<path>` on the commandline. By default, the working
+# directory is used for build output, but `BUILDDIR=<path>` allows overriding
+# it.
+#
+
+BUILDDIR ?= .
+SRCDIR ?= .
+
+RST2MAN ?= rst2man
+
+#
+# Automatic Variables
+#
+# This section contains a bunch of automatic variables used all over the place.
+# They mostly try to fetch information from the repository sources to avoid
+# hard-coding them in this makefile.
+#
+# Most of the variables here are pre-fetched so they will only ever be
+# evaluated once. This, however, means they are always executed regardless of
+# which target is run.
+#
+#     COMMIT:
+#         This evaluates to the latest git commit sha. This will not work if
+#         the source is not a git checkout. Hence, this variable is not
+#         pre-fetched but evaluated at time of use.
+#
+
+COMMIT = $(shell (cd "$(SRCDIR)" && git rev-parse HEAD))
+
+#
+# Generic Targets
+#
+# The following is a set of generic targets used across the makefile. The
+# following targets are defined:
+#
+#     help
+#         This target prints all supported targets. It is meant as
+#         documentation of targets we support and might use outside of this
+#         repository.
+#         This is also the default target.
+#
+#     $(BUILDDIR)/
+#     $(BUILDDIR)/%/
+#         This target simply creates the specified directory. It is limited to
+#         the build-dir as a safety measure. Note that this requires you to use
+#         a trailing slash after the directory to not mix it up with regular
+#         files. Lastly, you mostly want this as order-only dependency, since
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,62 +1,82 @@
...
-### Compose
-* Compose is what the user submits over one of the frontends
-* It contains of one or more **image builds**
-* It contains zero or more **upload actions**
+At build-time, the following software is required:
 
-### Image build
-* The resulting *image* has a *type*: https://github.com/osbuild/osbuild-composer/blob/master/internal/distro/fedora30/distro.go#L19
-* Running build in osbuild-composer is referred to as a "job" (internal terminology, not related to end-user experience)
+ * `go >= 1.14`
```
larskarlitski: Did we specify this before? I think we're able to build on older versions as well. Or better: we ought to, because we want to work on RHEL 8.1, which ships go 1.12.

docs/osbuild-composer.7.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Frontends
+---------
+
+*Composer* does not ship with frontends itself. However, several external
+frontends for *Composer* already exist. These include:
+
+**Cockpit Composer**
+    This module for *Cockpit* [#cockpit]_ allows a great level of control of a
+    *Composer* instance running on a cockpit-managed machine.
```
larskarlitski: Do you mind calling this "Image Builder module for Cockpit" or something similar? We're not referring to it as "Composer" in the UI.

Technically, "Cockpit" is only the project name and we always use "Fedora Web Console" in documentation. It's probably fine to keep using "Cockpit" here though, because it's intended for an audience that knows about these things.

README.md
ondrejbudai: Hmm, we should decide on this one, because I just wanted to include some code supported only by 1.13+.

msehnout: And we should change this, if we want to make sure we can run on 1.12:
https://github.com/osbuild/osbuild-composer/blob/master/.github/workflows/tests.yml#L22

Makefile
```diff
@@ -106,3 +212,66 @@ old-rpm: $(OLD_RPM_SPECFILE) $(RPM_TARBALL)
...
+#
+# Releasing
+#
+
+NEXT_VERSION := $(shell expr "$(VERSION)" + 1)
+
+.PHONY: release
+release:
+	@echo
+	@echo "Checklist for release of osbuild-composer-$(NEXT_VERSION):"
```
bcl: I don't think this belongs in the Makefile, RELEASE.md would be the place I'd normally look for instructions like this.

dvdhrm: Can you elaborate? Why does this not belong here? What **does** belong into this makefile? Why is putting this into RELEASE.md preferable? Especially given that this will prevent us from auto-generating the content just as `make release` does.

This target allows running `make release`, returning you generated output that you can easily copy-paste. This is very convenient when prepping a release, as you do not have to edit all the different tags in the commands.

docs/osbuild-composer.7.rst
dvdhrm: Yeah, makes sense.

I am still a bit unsure how to phrase things in different contexts. This needs more experience and trial-and-error I guess.

Regarding Cockpit, do you have a reference to the documentation you talk about? They do have their upstream-project-documentation (eg. [[1]](https://cockpit-project.org/guide/latest/feature-storaged.html)), but this consistently speaks of `Cockpit`. I guess you were referring to RHEL documentation (or at least more downstream docs)? I would like to get a better feel of how they pulled this off. I am still unsure how exactly we will achieve something similar.

[1] https://cockpit-project.org/guide/latest/feature-storaged.html

README.md
dvdhrm: This is a "best effort guess". The alternative is to not specify a version. The advantage of specifying the version is to help document what we do and providing guidance. The disadvantage is that we might be too strict, thus restricting downstream unnecessarily. I am open for suggestions here.

My intention was to document something I am sure works. From now on, we bump it whenever we need something newer, thus keeping this as a live dependency map. No strong feelings either way, though.

docs/osbuild-composer.7.rst
larskarlitski: I think a useful differentiation is upstream vs downstream documentation: upstream it's all "Cockpit" (like in the link you provided), downstream it's "X Web Console".

Not super important to get exactly right. I just wanted to make you aware of it.

Makefile
bcl: Actions belong in the Makefile, instructions belong in textfiles. I'd almost never think to look in the Makefile for the release checklist.

PR #430, opened 2020-03-31
tools: remove accidentally added line in test case generator

ondrejbudai:
The line slipped in f945c505 when testing the changes. Let's remove it,
it breaks the script.


teg:
Ugh! Sorry about this, and thanks for the fix!

PR #432, opened 2020-03-31
tests: add coverage for common/helpers.go

jrusz:
This also changes the helpers.go to enable mocking all the different architectures.  If you have a better idea for the name of the newly created variable I'll be glad to hear it :)  
The `runtime.GOARCH` is actually a constant that is taken from `sys` package https://golang.org/pkg/runtime/#pkg-constants

Refs #419

### DIFF
--- a/internal/common/helpers.go
+++ b/internal/common/helpers.go
@@ -2,14 +2,16 @@ package common
 
 import "runtime"
 
+var RuntimeGOARCH = runtime.GOARCH
+
 func CurrentArch() string {
-	if runtime.GOARCH == "amd64" {
+	if RuntimeGOARCH == "amd64" {
 		return "x86_64"
-	} else if runtime.GOARCH == "arm64" {
+	} else if RuntimeGOARCH == "arm64" {
 		return "aarch64"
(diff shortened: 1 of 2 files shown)


atodorov:
Is `sys.GOARCH` also a constant or a function/variable ? I can't find it's definition. 

I'm wondering if the tests could be overriding that instead?

jrusz:
> Is `sys.GOARCH` also a constant or a function/variable ? I can't find it's definition.
> 
> I'm wondering if the tests could be overriding that instead?

It seems to also be a constant that is configured by during GO installation on the system I suppose?  
You mean that it might be possible to modify that constant somewhere instead of mocking it here like I did? Sounds doable, I can try that if we'd rather prefer it that way.

jrusz:
I can see it running `go env` and it seems to be possible to change using `os.Setenv`.

ondrejbudai:
> Is `sys.GOARCH` also a constant or a function/variable ? I can't find it's definition.

It's constant, thus it cannot be modified. Defined in `/usr/lib/golang/src/runtime/extern.go`.

PR #443, opened 2020-04-01
tests: use asserts in states_test.go

jrusz:
Asserts look nicer and are easier to read. Functionality should be the same.

Refs #419


jrusz:
I rebased to latest master and force pushed to re-run the checks.

PR #450, opened 2020-04-01
tests: increase coverage of internal/crypt

jrusz:
@atodorov could you give me some advice here please? By simply calling the `CryptSHA512` function I get a lot of lines covered which is fine I guess but I am not sure if there's anything else I can do to verify correct functionality of the hash function. I thought about calling it multiple times with the same input and comparing the returned values if they're all unique (each one of them gets random salt). 

But since the actual hashing algorithm is implemented in an external C function I guess simply verifying that this returns a hash value and no errors is enough? What's your opinion on this?

### DIFF
--- a/internal/crypt/crypt_test.go
+++ b/internal/crypt/crypt_test.go
@@ -1,6 +1,7 @@
 package crypt
 
 import (
+	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
@@ -40,4 +41,21 @@ func Test_crypt_PasswordIsCrypted(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestCryptSHA512(t *testing.T) {
+	retPassFirst, err := CryptSHA512("testPass")
+	assert.NoError(t, err)
+	retPassSecond, _ := CryptSHA512("testPass")
+	expectedPassStart := "$6$"
+	assert.Equal(t, expectedPassStart, retPassFirst[0:3])
+	assert.NotEqual(t, retPassFirst, retPassSecond)
+}
+
+func TestGenSalt(t *testing.T) {
+	length := 10
+	retSaltFirst, err := genSalt(length)
+	assert.NoError(t, err)
+	retSaltSecond, _ := genSalt(length)
+	assert.NotEqual(t, retSaltFirst, retSaltSecond)
 }


internal/crypt/crypt_test.go
```diff
@@ -41,3 +42,9 @@ func Test_crypt_PasswordIsCrypted(t *testing.T) {
 		})
 	}
 }
+
+func TestCryptSHA512(t *testing.T) {
+	retPass, err := CryptSHA512("testPass")
+	assert.NoError(t, err)
+	assert.True(t, PasswordIsCrypted(retPass))
```
atodorov, line 49: This last line is mixing the scenarios for the 2 functions a bit. You are using the PasswordIsCrypted function to validate the result of CrytpSHA512

- For `CryptSHA512` you really don't have much to validate b/c it is a very straight forward function. The resulting string should start with `$6$` and nothing else. There should be no error.

- for `genSalt` - the behavior is that it always returns random strings and there is no error. 
- `PasswordIsCrypted` - the behavior is that it recognizes an encrypted password with one of several algorithms.

atodorov:
> since the actual hashing algorithm is implemented in an external C function I guess simply verifying that this returns a hash value and no errors is enough?

Yes, we're not concerned what external software is doing. We assume it to work correctly for the purposes of unit testing. Otherwise we end up with a higher level of testing which ideally should be happening in crypto/rand's upstream.

atodorov:
@ondrejbudai I'm looking at `genSalt` and I'm wondering isn't there a function that provides secure salts/tokens built into the language? Kind of like Python's secrets module ?

internal/crypt/crypt_test.go
jrusz: Thanks for breaking it down like this. 

> This last line is mixing the scenarios for the 2 functions a bit. You are using the PasswordIsCrypted function to validate the result of CrytpSHA512
> 
I thought it would be easy to use that to verify it actually returns an encrypted pass. So it's preferable to verify the output without calling another function from the same `.go` file? I'm not actually trying to test the `PasswordIsCrypted` by that...

For `genSalt` I'm not sure what's the best way to verify that it always returns a random string. Should I just call it ~10 times and see if all the values are different?

jrusz:
I got curious and googled a bit and found out that this seems to be what people use: https://godoc.org/golang.org/x/crypto/bcrypt

Some example of usage can be found here: https://medium.com/@jcox250/password-hash-salt-using-golang-b041dc94cb72

Looks a lot simpler to me than what's being used here now and it seems to have the same functionality.

bcl:
> I got curious and googled a bit and found out that this seems to be what people use: https://godoc.org/golang.org/x/crypto/bcrypt

bcrypt is fairly old at this point, the newest one is scrypt:
https://pkg.go.dev/golang.org/x/crypto/scrypt?tab=doc

But we need to make sure the distribution being built supports it.

ondrejbudai:
The point of using this implementation is that it uses SHA512-based crypt, which is the exact same thing as Lorax does.

There's this package, which can do SHA512 in pure go: https://github.com/GehirnInc/crypt , but I don't trust it (it's not widely used).

We can surely use something different, but I just wanted to point out why I chose this implementation. Sadly, for SHA512 we don't have many options in my opinion.

> bcrypt is fairly old at this point, the newest one is scrypt

scrypt is cool, we just need to make sure that all the distros support it.

internal/crypt/crypt_test.go
teg: > I thought it would be easy to use that to verify it actually returns an encrypted pass. So it's preferable to verify the output without calling another function from the same `.go` file? I'm not actually trying to test the `PasswordIsCrypted` by that...

At least I always prefer to have one test case per function. Sometimes you need to use other functions in order to set-up/tear-down/verify the test case, but at least I like to make it clear which of the functions you are testing in each case, and not mix several into one.

> For `genSalt` I'm not sure what's the best way to verify that it always returns a random string. Should I just call it ~10 times and see if all the values are different?

I'd say verifying the format (length of strings, possibly characters used) and then that two calls give different values (I think you don't get much more value from >2).

atodorov: Also need to point out that if `PasswordIsCrypted` has a bug that will also break the SHA512 test so I prefer for the asserts to be as independent as possible.

jrusz: @atodorov @teg Thanks for the advice and explanation, makes sense to me. I made the changes in the latest commit and also added a separate test function for `genSalt`.

jrusz:
Should be ready for review and merge. I guess we can decide whether we want to use `scrypt` here in a separate issue/PR and merge a test for what we have now...

teg:
Travis was green, but did not report back.

Issue #442, opened 2020-04-01
 Improve test coverage on the distro packages


jikortus:
Based on the info from Tom, it would be helpful to improve the coverage of the distro packages. There is a planned refactoring of the internal helpers, so this is out of scope of this particular effort, with rather focusing on the public methods.

jikortus:
This issue is covered now -> closing.

PR #453, opened 2020-04-01
Improve coverage for distro packages - F30

jikortus:
Just pushing what I currently have for comments. Not sure if it's not overcomplicated (`TestImageType_Name`) or not too little inclusive (`TestImageType_Size`). But maybe I've just been staring into the code for too long :-).

Refs: #442

### DIFF
--- a/internal/distro/fedora30/distro_test.go
+++ b/internal/distro/fedora30/distro_test.go
@@ -1,9 +1,12 @@
 package fedora30_test
 
 import (
-	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
 	"reflect"
 	"testing"
+
+	"github.com/osbuild/osbuild-composer/internal/distro/distro_test_common"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestFilenameFromType(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/distro/fedora30/distro_test.go
```diff
@@ -1,11 +1,38 @@
...
-	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
 	"reflect"
 	"testing"
+
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora30"
+	"github.com/osbuild/osbuild-composer/internal/osbuild"
+	"github.com/stretchr/testify/assert"
 )
 
+type arch struct {
```
teg, line 12: You should not need to redefine the internal structs to test the public functions. Otherwise I think the tests look good.

internal/distro/fedora30/distro_test.go
```diff
@@ -137,3 +164,136 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+			for _, imgName := range mapping.imgNames {
+				imgType, err := arch.GetImageType(imgName)
+				assert.Nil(t, err)
+				if err == nil {
+					assert.Equalf(
+						t,
+						imgName,
+						imgType.Name(),
+						"arch: %s, expected: %s, got: %s",
+						arch, imgName, imgType.Name(),
```
atodorov, line 212: The "expected, got" part of the error message is a bit redundant b/c assertEqual will do this for you in case of failure.

internal/distro/fedora30/distro_test.go
```diff
@@ -137,3 +164,136 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+				"qcow2",
+				"openstack",
+				"tar",
+			},
+		},
+	}
+	for _, mapping := range imgMap {
+		arch, err := distro.GetArch(mapping.arch)
+		assert.Nil(t, err)
+		if err == nil {
```
atodorov, line 202: if you replace assert.Nil with require.Nil then the if condition isn't necessary. 

However I don't quite understand how this is going to work here. If I am on a x86_64 system and the loop iterates over the 2nd element in the array what will happen? 

What if I am on a ppc64le system ?

teg: All this code is independent of the architecture of the machine you are running the tests on, so you can run all the functions on any architecture.

jikortus: @atodorov if I do the replacing as you suggest, it won't be functionally equivalent IMO (failure of one scenario vs. failure of the whole test).

jikortus: I did some more changes/additions, especially I added the `TestDistro_Manifest` function I borrowed from `/internal/distro/distro_test.go` and changed a bit. I wonder if it would make sense to modify it a bit to be more generic (to include scenario file prefixes and distro objects as parameters), move it a bit higher in the tree and then call it from the more specific test files (for F30, .., RHEL-...)?

internal/distro/fedora30/distro_test.go
```diff
@@ -137,3 +149,313 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+			for _, imgName := range mapping.imgNames {
+				imgType, err := arch.GetImageType(imgName)
+				assert.Nil(t, err)
+				if err == nil {
+					assert.Equalf(t, imgName, imgType.Name(), "arch: %s", imgType.Name())
```
atodorov, line 192: Shouldn't the last argument be `arch`, or was it wrong before ?

jikortus:
I guess the coverage increase for F30 is complete, if you think it's OK @atodorov @teg, I'd port it to the remaining distro tests.

I also moved the `TestDistro_Manifest` function to a separate package and made it more versatile, so that it can be used from tests for specific distros. I don't really like it being in a separate directory (and file/package), but I wasn't able to just add it into `.../distro_test` package and then import it from `.../distro/fedora30_test` package. Not sure if I did it wrong, or it's not possible at all?

internal/distro/fedora30/distro_test.go
jikortus: Yes, thanks for noticing this.

atodorov:
@jikortus LGTM, let's merge to avoid possible conflicts if it stays open for too long.

About `TestDistr_Manifest` - I don't think you can import into other files if this is not part of a package but I may be wrong.

jikortus:
Thanks @atodorov! So I will edit the PR title and remove WIP from the commit message, and then I'll focus on porting this to cover other distro tests.

However, before this gets merged, can @teg also please have a look (especially if `TestDistr_Manifest` is OK this way, or it can be somehow integrated as part of `distro_test` package? Thanks!

teg:
ok to test

jikortus:
@teg Thanks! There was a need for rebase, so it should be ready to merge (although it still seems that Schutzbot doesn't like some aspect of the PR :) ).

jikortus:
@msehnout That was definitely not my intention and I have no clue why this happened, sorry for that. Hope it should be OK now.

PR #456, opened 2020-04-02
Test cases for aarch64

msehnout:
This commit includes a support for aarch64 in the `/cmd/osbuild-image-tests` and generated test cases for supported image types for aarch64. All the test cases are missing the "boot" JSON object as well as the "image-info" object, so the only thing it does right now is to build the image.

See commit messages for further explanation.


cmd/osbuild-image-tests/main_test.go
```diff
@@ -232,7 +232,7 @@ func testImage(t *testing.T, testcase testcaseStruct, imagePath, outputID string
 			testImageInfo(t, imagePath, testcase.ImageInfo)
 		})
 	}
-
+
```
teg, line 235: Looks like a mistake?

cmd/osbuild-image-tests/context-managers.go
```diff
@@ -90,7 +91,25 @@ func withBootedQemuImage(image string, ns netNS, f func() error) error {
 			return fmt.Errorf("cannot close temporary cloudinit file: %#v", err)
 		}
 
-		qemuCmd := ns.NamespacedCommand(
+		var qemuCmd *exec.Cmd
+		// This command does not use KVM as I was unable to make it work in Beaker,
+		// once we have machines that can use KVM, enable it to make it faster
```
teg, line 96: Could we skip the test if we detect that kvm does not work?
I think we should keep the test cases unmodified from the output of the generation scripts, or they will get out-of-sync (and we also want to be able to easily regenerate them without having to edit them), so if we need to skip anything, we should do that here, in the test runner.

teg: Another alternative would be to support a commandline option to disable the qemu tests, if auto-detection is not possible.

msehnout: Hmm can you think of a way to detect KVM? I can think of `lsmod | grep kvm`, but I'm not sure how reliable it is.

msehnout:
> As of osbuild-11, the partuuid support exists, so we should be able to include the image-info in the test-case as well. I think we should include the boot struct too, so that the test cases are exactly as autogenerated, or this will drift over time.

Yes, I plan to introduce that as part of a following PR.

msehnout:
Meanwhile I was able to run all those tests and they seem to work except for failing curl in osbuild.

cmd/osbuild-image-tests/context-managers.go
teg: Could we not run `qemu` in a way that forces `kvm` emulation, and if that fails skip the test? Not sure what it allows. @gicmo, @dvdhrm, @larskarlitski: Hilfe!

dvdhrm: If you use `-M accel=kvm`, it will only use KVM and fail otherwise. If you use `-M accel=kvm:hvf:tcg`, it will try KVM first, then HVF, and lastly TCG. TCG should always work, as it is the built-in ISA emulator IIRC.

I am not sure whether you can really catch the error when KVM is not there. Usually, it is enough to use `test -f /dev/kvm`, though. Might wanna give this a try?

teg: `test -f /dev/kvm` sounds ok, but why not just pass `-M accel=kvm:hvf` and skip the test if `qemu` itself fails?

dvdhrm: How do you know whether qemu failed because of missing KVM or because the test failed? Other than some hacky output-string comparison.

larskarlitski: I agree with David. If we assume qemu can fail (by only letting it use kvm), let's have a separate check for it by testing for `/dev/kvm`.

`test -f /dev/kvm` will fail. You need `-c` ("FILE exists and is character special") or `-e` ("FILE exists").

teg: I thought there would be some way to distinguish between "`qemu` failed during setup" and "invalid image file", but if not, then checking for `/dev/kvm` sounds fine.

cmd/osbuild-image-tests/context-managers.go
```diff
@@ -90,7 +91,25 @@ func withBootedQemuImage(image string, ns netNS, f func() error) error {
...
+		qemuAarch64Cmd := ns.NamespacedCommand(
+			"qemu-system-aarch64",
+			"-cpu", "host",
+			"-M", "virt",
+			"-m", "2048",
+			// As opposed to x86_64, aarch64 uses UEFI, this one comes from edk2-aarch64 package on Fedora
```
ondrejbudai, line 102: I think we should add a dependency to that package to the spec file.

cmd/osbuild-image-tests/context-managers.go
```diff
@@ -90,7 +91,25 @@ func withBootedQemuImage(image string, ns netNS, f func() error) error {
...
 
-		qemuCmd := ns.NamespacedCommand(
+		var qemuCmd *exec.Cmd
+		// This command does not use KVM as I was unable to make it work in Beaker,
+		// once we have machines that can use KVM, enable it to make it faster
+		qemuAarch64Cmd := ns.NamespacedCommand(
+			"qemu-system-aarch64",
+			"-cpu", "host",
+			"-M", "virt",
+			"-m", "2048",
```
ondrejbudai, line 101: Is 2 GBs of memory really needed?

msehnout: I don't know, that's what @gicmo suggested :).

cmd/osbuild-image-tests/context-managers.go
msehnout: Is it possible that `/dev/kvm` exists, but `kvm` still does not work? Would `modprobe kvm` be more reliable here?

cmd/osbuild-image-tests/context-managers.go
msehnout: good idea, fixed

cmd/osbuild-image-tests/context-managers.go
teg: Hmm, I think checking for the device should be sufficient. If that exists but doesn't work, then I think the tests should rightly fail. We may not always be able to modprobe (probably in practice we will, but I don't think we should rely on it).

teg: You added the `kvm` detection, so I think the command now should contain `-accel` so that if we do get `kvm` to work it will just work.

cmd/osbuild-image-tests/main_test.go
```diff
@@ -299,6 +299,12 @@ func testBoot(t *testing.T, imagePath string, bootType string, outputID string)
 	}
 }
 
+func kvmAvailable() bool {
+	cmd := exec.Command("modprobe", "kvm")
```
teg, line 303: Any reason you are not just checking for `/dev/kvm`?

cmd/osbuild-image-tests/main_test.go
```diff
@@ -309,6 +315,10 @@ func testImage(t *testing.T, testcase testcaseStruct, imagePath, outputID string
 	}
 
 	if testcase.Boot != nil {
+		if common.CurrentArch() == "aarch64" && !kvmAvailable() {
```
teg, line 318: I think we could/should skip the tests if `kvm` (or equivalent) is not available, independent of the architecture, no? (This could be in a follow-up if this is not obvious).

internal/osbuild/qemu_assembler.go
```diff
@@ -22,7 +22,7 @@ type QEMUPartition struct {
 	Size       uint64         `json:"size,omitempty"`
 	Type       *uuid.UUID     `json:"type,omitempty"`
 	Bootable   bool           `json:"bootable,omitempty"`
-	UUID       *uuid.UUID     `json:"uuid,omitempty"`
```
teg, line 25: Hm, the `Type` two lines above uses `uuid.UUID`, I think the type-safety is nicer to avoid typos. Combined with `MustParse()`, that is. Not a blocker though.

cmd/osbuild-image-tests/context-managers.go
msehnout: ok

cmd/osbuild-image-tests/main_test.go
ondrejbudai: I think I would also move this condition inside the `boot` sub-test and use `t.Skip`, this way the output log would be more consistent I think.

cmd/osbuild-image-tests/main_test.go
msehnout: No, I changed to `/dev/kvm` detection.

cmd/osbuild-image-tests/main_test.go
msehnout: I wanted to avoid breaking other architectures tests so I'd prefer a follow up.

internal/osbuild/qemu_assembler.go
msehnout: Maybe this can be a follow up and also change `PTUUID`?

cmd/osbuild-image-tests/main_test.go
teg: Fair enough!

internal/osbuild/qemu_assembler.go
teg: I agree :)

ondrejbudai:
Travis seems to be unhappy by the changes, I think we should just comment out all the arm jobs for now.

msehnout:
Sorry about that. I screw up merge of `.travis.yml`... I had to do one more respin.

PR #459, opened 2020-04-03
introduce support for partuuid

msehnout:
more info in the commits

### DIFF
--- a/internal/osbuild/qemu_assembler.go
+++ b/internal/osbuild/qemu_assembler.go
@@ -22,6 +22,7 @@ type QEMUPartition struct {
 	Size       uint64         `json:"size,omitempty"`
 	Type       *uuid.UUID     `json:"type,omitempty"`
 	Bootable   bool           `json:"bootable,omitempty"`
+	UUID       *uuid.UUID     `json:"uuid,omitempty"`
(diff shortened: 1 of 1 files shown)


internal/osbuild/qemu_assembler.go
```diff
@@ -22,6 +22,7 @@ type QEMUPartition struct {
 	Size       uint64         `json:"size,omitempty"`
 	Type       *uuid.UUID     `json:"type,omitempty"`
 	Bootable   bool           `json:"bootable,omitempty"`
+	PARTUUID   *uuid.UUID     `json:"uuid,omitempty"`
```
ondrejbudai, line 25: I think I prefer naming this field just `UUID` for the sake of consistency.

msehnout: I don't really have a strong opinion here. The field few lines above is called "PTUUID", so I thought it would be consistent like this.

ondrejbudai: Yeah, but the `PTUUID` field is consistent with its json representation, whereas the `PTUUID` is not.

msehnout: ok, I'll change it

PR #460, opened 2020-04-03
upload/aws: delete the s3 object after the snapshot is imported

ondrejbudai:
Previously the s3 object was just left on s3 after the upload had finished. However, the object is not needed anymore after an ec2 snapshot is imported from the object. I think this is wrong from the customers' perspective. I think they would not anticipate that we leave not needed s3 objects after upload.

The s3 key can be currently used using the weldr api. However, the `key` field is currently not used by cockpit-composer. When the field isn't set, osbuild-composer uses the compose id as a key. That's however not ideal, because we soon will be able to upload more images for one compose.

Let's discuss how lorax does it: Lorax does remove the s3 object after upload. Also, it doesn't the `key` object in its API, the key is always this: `{{ image_name }}-{{ image_stat['stat']['checksum'] }}.ami`.

So I think it totally makes sense to remove the s3 object after the upload.

Possible follow-ups:

1) Drop the key from API. It's imho useless.
2) Find a better auto-generated key.

cc @jgiardino

### DIFF
--- a/internal/upload/awsupload/awsupload.go
+++ b/internal/upload/awsupload/awsupload.go
@@ -9,12 +9,14 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
 type AWS struct {
 	uploader *s3manager.Uploader
 	importer *ec2.EC2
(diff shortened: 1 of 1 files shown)


teg:
This makes sense to me, and I think it is actually the workflow we documented. I'll leave it to Jenn to approve (or not ;) ).

jgiardino:
Yep, sounds good! I put this question out to the rest of the team, and heard no objections to removing the image. Some of the reasons in support of removing it from S3 is that it helps keep the cost to the user low, and that's what the Ansible playbooks do. And if we ever do hear feedback in support of a use case for S3 being an additional upload destination, we can always add that later.

PR #463, opened 2020-04-03
dnf-json: always check for new metadata

teg:
After including the fedora updates repository, we were hitting issues
due to expired metadata. The updates repository requires metadata to
be expired at most every six hours, and the default was to do so every
48h.

On further reflection, it was not clear to me what the right value
here is. The values set in the repo files are the maximums, but
users may want to have the ability to access updated content sooner
than that (especially if they are pushing content to a repo and
immediately pulling into osbuild for testing purposes).

For now, we do like lorax and always check for new metadata, rather
than try to come up with some compromise value or (possibly worse)
expose this to the caller to figure out. I imagine this is something
we eventually will want to revisit.

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -34,6 +34,21 @@ def dnfrepo(desc, parent_conf=None):
     if desc.get("ignoressl", False):
         repo.sslverify = False
 
+    # The default metadata expiration time is 48 hours. This is not enough for
+    # all repositories we want to support (in particular fedora updates needs
+    # it set to at most 6h in order to guarantee the backing rpms are still
+    # available). The time rpms are guaranteed to stay around is not indicated
+    # in the repo metadata, but must be set by the caller. We would really
+    # rather not have to keep knowledge about this, but just as importantly,
+    # we may want to set the expiration timeout to a smaller value, so that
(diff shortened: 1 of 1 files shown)


bcl:
> As discussed, let's treat metadata as never expiring (`-1`) unless specifically overriden by a repository.

That won't work. As soon as the mirror removes the packages you'll get download errors.

This is what I had to do in lorax to make 100% sure new metadata was always used:

https://github.com/bcl/osbuild-composer/commit/82db3eb7861b994b951fcb208ac54bd54637850b

(commented out because I haven't been able to reproduce the problem in a controlled manner).

Of course this means more network traffic. In lorax I set a manual timeout of 6 hours so that depsolving a blueprint wouldn't pull new data. But then I forced an update when depsolving just before a build to make sure it was current -- but this was because Anaconda would do the same thing, so that doesn't really apply here.

teg:
The idea would be to leave it -1 if unset, but set it to the right value for all the repositories we use. Then, when people use repos that keep rpms around, they can skip this (but this is only relevant in the low-level API). I'll post a new PR with a proper proposal.

bcl:
Ah! Thanks for the clarification.

PR #466, opened 2020-04-03
💃 Add dnf integration tests

major:
Run the dnf integration tests with every pull request.


larskarlitski:
Looks like Mr Sander lied to us:
```
_hawkey.RuntimeException: Failed to resolve: Encountered two streams with matching NSVCA testmodule:master:3020190304173846:a5b0195c:aarch64 but differing content
```

I have no immediate idea what this could be. I'll have a look later.

PR #467, opened 2020-04-04
queue: cleanups

larskarlitski:
These commits don't change behavior. They're mostly cleanups and refactors around the `jobqueue` package.


larskarlitski:
Update to run `gofmt`. Sorry.

larskarlitski:
Schutzbot Fedora32 fails, because cockpit-composer failed to build.

codecov/patch failed, because this patch doesn't add tests for all the lines it adds. I think this shouldn't apply here though, because I'm just moving code around.

PR #469, opened 2020-04-06
jobqeue: worker API fixups

larskarlitski:
Make the worker API behave a bit better, especially not returning plain text when advertising `application/json`. See commits messages for details.

### DIFF
--- a/internal/jobqueue/api.go
+++ b/internal/jobqueue/api.go
@@ -66,10 +66,6 @@ func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
 	writer.WriteHeader(http.StatusNotFound)
 }
 
-func statusResponseOK(writer http.ResponseWriter) {
-	writer.WriteHeader(http.StatusOK)
-}
-
 func statusResponseError(writer http.ResponseWriter, code int, errors ...string) {
 	writer.WriteHeader(code)
 
@@ -150,7 +146,6 @@ func (api *API) updateJobHandler(writer http.ResponseWriter, request *http.Reque
 		}
 		return
 	}
-	statusResponseOK(writer)
 }
 
 func (api *API) addJobImageHandler(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
@@ -180,6 +175,4 @@ func (api *API) addJobImageHandler(writer http.ResponseWriter, request *http.Req
 		}
 		return
 	}
-
-	statusResponseOK(writer)
 }
--- a/internal/jobqueue/api.go
+++ b/internal/jobqueue/api.go
@@ -59,34 +59,36 @@ func (api *API) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 	api.router.ServeHTTP(writer, request)
 }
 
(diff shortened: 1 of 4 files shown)


internal/jobqueue/api.go
```diff
@@ -105,59 +107,57 @@ func (api *API) addJobHandler(writer http.ResponseWriter, request *http.Request,
...
-		case *store.NotRunningError:
-			statusResponseError(writer, http.StatusBadRequest, err.Error())
-		case *store.InvalidRequestError:
-			statusResponseError(writer, http.StatusBadRequest, err.Error())
+		case *store.NotFoundError, *store.NotPendingError:
+			jsonErrorf(writer, http.StatusNotFound, "%v", err)
+		case *store.NotRunningError, *store.InvalidRequestError:
+			jsonErrorf(writer, http.StatusBadRequest, "%v", err)
 		default:
-			statusResponseError(writer, http.StatusInternalServerError, err.Error())
```
teg, line 152: What's the difference betwene `err.Error()` and `"%v", err`

internal/jobqueue/api_test.go
```diff
@@ -15,30 +15,28 @@ import (
...
-		// Wrong method
-		{"GET", "/job-queue/v1/jobs", ``, http.StatusMethodNotAllowed, ``},
-		// Update job with invalid ID
-		{"PATCH", "/job-queue/v1/jobs/foo/builds/0", `{"status":"RUNNING"}`, http.StatusBadRequest, `invalid compose id: invalid UUID length: 3`},
-		// Update job that does not exist, with invalid body
-		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa/builds/0", ``, http.StatusBadRequest, `invalid status: EOF`},
-		// Update job that does not exist
-		{"PATCH", "/job-queue/v1/jobs/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa/builds/0", `{"status":"RUNNING"}`, http.StatusNotFound, `compose does not exist`},
+		{"POST", "/job-queue/v1/jobs", ``, http.StatusBadRequest},
+		// // Wrong method
```
teg, line 27: Hm, typo? Double comment?

internal/jobqueue/api.go
larskarlitski: There is none, except that `%v` reads nicer (less characters) and seems idiomatic. According to [the documentation](https://golang.org/pkg/fmt/):

>  If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
> 
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any). 

So even `%s` would work without calling `Error()` explicitly. I've seen `%v` more often, though, and would go with that.

internal/jobqueue/api_test.go
larskarlitski: Oops indeed.

PR #470, opened 2020-04-06
osbuild requirement: bump to version 11

msehnout:
The newest osbuild is needed for support for aarch64. The aarch64 images
are using GPT which requires stable partuuid which was included in the
latest osbuild release (11). This will be used to produce stable
image-info tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ provided infrastructure and services.
 
 The requirements for this project are:
 
- * `osbuild >= 7`
+ * `osbuild >= 11`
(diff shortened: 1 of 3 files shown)


msehnout:
Waiting for https://bodhi.fedoraproject.org/updates/FEDORA-2020-ab8aeef5a3 to land in stable.

msehnout:
As discussed with @dvdhrm and @major , f32 just fails to build. I've tried 3 times, but it's still failing. We don't necessarily need to block on that check.

PR #472, opened 2020-04-06
distro: don't include grub2 stage in ext4 images

msehnout:
@larskarlitski , @jgiardino is it ok to change the ext4 image type like this? I'm really unsure about its definition. The image type is declared non-bootable here:
https://github.com/osbuild/osbuild-composer/blob/7dd09443cf22b4cb3a9f15f6fe3d34eb02d8d3ac/internal/distro/fedora31/distro.go#L226

There is no point in having the grub2 stage in pipelines for image types
that are not bootable. The current version is probably a result of
previous refactoring where the member variable was named `IncludeFSTab`.

Moving the grub2 stage into the conditional branch should also fix test
generation on aarch64.

TODO:
 - [x] Regenerate all ext4 test cases

### DIFF
--- a/internal/distro/fedora30/distro.go
+++ b/internal/distro/fedora30/distro.go
@@ -519,8 +519,8 @@ func (t *imageType) pipeline(c *blueprint.Customizations, repos []rpmmd.RepoConf
 
 	if t.bootable {
 		p.AddStage(osbuild.NewFSTabStage(t.fsTabStageOptions(t.arch.uefi)))
+		p.AddStage(osbuild.NewGRUB2Stage(t.grub2StageOptions(t.kernelOptions, c.GetKernel(), t.arch.uefi)))
 	}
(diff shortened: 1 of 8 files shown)


larskarlitski:
I think the whole `ext4` image type is a bit pointless. There's not really a definition for what it should do.

Unless this is blocking you on something, I'd wait until we drop it completely or have written up a definition for it, which should include its purpose (and thus whether it should boot ;) )

This fix should be ok though. I don't see a reason against it.

teg:
> Unless this is blocking you on something

The aarch64 work is blocked on this. The grub stage depends on the grub2-* package, which is included on x86_64, but only on aarch64 if it is listed as bootable (maybe this is a bug on x86_64?).

> This fix should be ok though. I don't see a reason against it.

I agree, let's fix this to at least pass the tests for now, then revisit once we have image type definitions written up.

PR #473, opened 2020-04-06
dnf-json-tests: split into sub tests

larskarlitski:
While debugging a modularity error on fedora-32, I noticed that it's hard to find out which distro / arch / image type combination failed. This doesn't fix the actual error (@major turned off testing on fedora-32 for now), but might make things simpler in the future.

There's also a commit that fixes `dnf-json` not print python stacktraces in a few additional cases.

See commits for more details.

### DIFF
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -56,34 +56,31 @@ func TestCrossArchDepsolve(t *testing.T) {
 
 	// NOTE: we can add RHEL, but don't make it hard requirement because it will fail outside of VPN
 	for _, distroStruct := range []distro.Distro{fedora30.New(), fedora31.New(), fedora32.New()} {
-		repos, err := rpmmd.LoadRepositories([]string{repoDir}, distroStruct.Name())
-		assert.NoErrorf(t, err, "Failed to LoadRepositories %v", distroStruct.Name())
-		if err != nil {
-			// There is no point in running the tests without having repositories, but we can still run tests
-			// for the remaining distros
-			continue
-		}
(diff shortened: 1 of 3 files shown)


larskarlitski:
> Nice! Thanks! Love the possibility of only running some tests or running them in parallel!

Heh, thanks. Was fun reading up on that.

I'd like to wait for @msehnout's review though. Maybe I'm missing something from why he didn't use sub tests before.

teg:
> I'd like to wait for @msehnout's review though.

Sure :)

larskarlitski:
Added @major's commit from https://github.com/osbuild/osbuild-composer/pull/466/files to get extra testing :100:

major:
@larskarlitski Output is a lot more readable. You have pleased the Schutzbot.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -50,38 +52,35 @@ func TestCrossArchDepsolve(t *testing.T) {
 	rpm := rpmmd.NewRPMMD(dir)
 
 	// Load repositories from the definition we provide in the RPM package
-	repositories := "/usr/share/osbuild-composer"
+	repoDir := "/usr/share/osbuild-composer"
```
bcl, line 55: Are these tests always run from a built and installed package? If they run from the git checkout there is a possibility of not using the correct set of repos.

larskarlitski: Yes, they are meant to be run as integration tests. It has the `integration` build constraint at the top of the file.

I guess we could run it from source, but we'd need to change this path and set `PATH` to find the right `dnf-json`. I don't think it's worth making this test more complex in such a way.

PR #479, opened 2020-04-06
🤏 Little bit more testing: image-tests

major:


major:
schutzbot

larskarlitski:
`rpm` is failing with exit code 254 and printing a lot of lines like this one

	installing package rootfiles-8.1-24.fc30.noarch needs 307MB on the / filesystem

@dvdhrm and @gicmo say this points to the disk being full. Could that be the case here?

dvdhrm:
> `rpm` is failing with exit code 254 and printing a lot of lines like this one
> 
> ```
> installing package rootfiles-8.1-24.fc30.noarch needs 307MB on the / filesystem
> ```
> 
> @dvdhrm and @gicmo say this points to the disk being full. Could that be the case here?

Yeah, I think so. We also accumulate entries in our cache currently. We should just make sure to delete cache-entries again whenever a test-run is over. This should fix all possible disk-space issues.

major:
So the disk space issues seem to be okay on F31 now, but the test does take a long time (which I guess we expect since it's doing a lot of things).

major:
schutzbot

major:
@larskarlitski @teg  This seems to work fine via Jenkins, but our testing time has ballooned out to ~ 50 minutes. There are optimizations that can be made, such as choosing a closer mirror for downloading packages or setting up a small proxy for caching packages downloaded via curl.

teg:
I suppose we want to hold off on this for a bit until we have figured out a way to speed it up a bit? Do we know exactly what's consuming the most time? Would it be possible to try backing our store by a tmpfs to see how that affects things?
@dvdhrm did some work on creating a container hosting a local mirror. That should work nicely for GitHub Actions, is that something we could reuse also for the integration tests?

atodorov:
+1 for local mirror.

From my humble experience network download takes the most time followed by disk i/o when building the images but I haven't done any specific measurement. Disk i/o seems to be faster than lorax-composer though. 

Jenkins log shows that rhel_psi build was 15 mins faster compared to fedora31. From what I can see RPM build time is faster on RHEL and also most of the tests are also faster. The ansible playbooks on Fedora seem to also be doing more setup than on RHEL.

major:
The RHEL bits are very close to the RHEL PSI instance (Boston has the bits, RDU has the RHEL PSI instance). That makes things pretty quick.

Our Fedora instances in DigitalOcean are located in NYC. It probably makes sense to find the closest repo to that location and adjust the JSON to point to those URLs.

major:
The fastest mirrors to DO (according to dnf's fastestmirror plugin) seem to be:

```
[http://fedora.mirrors.tds.net]
ts=1586981802
connectime=0.0012339999999999574

[https://fedora.mirror.constant.com]
ts=1586981802
connectime=0.0025419999999999332

[https://ewr.edge.kernel.org]
ts=1586981802
connectime=0.0038299999999999966

[http://fedora.mirror.constant.com]
ts=1586981802
connectime=0.010171999999999959

[http://mirrors.mit.edu]
ts=1586981802
connectime=0.015890999999999999
```

major:
Oh, also, we're not doing image tests on RHEL right now since we don't have test cases there.

major:
schutzbot

test/cases/fedora_30-aarch64-ami-boot.json
```diff
@@ -8,7 +8,7 @@
     "image-type": "ami",
     "repositories": [
       {
-        "baseurl": "http://dl.fedoraproject.org/pub/fedora/linux/releases/30/Everything/aarch64/os/",
+        "baseurl": "http://fedora.mirrors.tds.net/fedora/releases/30/Everything/aarch64/os/",
```
atodorov, line 11: If you replace with `download.fedoraproject.org` it should automatically point to the closest mirror, no ?
 I prefer not hard-coding a specific local mirror in the sources.

major: @atodorov This was mainly a hack just to see if the job run time improves. I don't want to put it into production like this. :smile:

major:
Still about 52 minutes. About 3-4 minutes per image test.

major:
schutzbot

major:
schutzbot

major:
schutzbot

Issue #317, opened 2020-03-12
RFC: compose types sanity tests


atodorov:
From lorax-composer's test suite we have https://github.com/weldr/lorax/blob/master/tests/cli/test_compose_sanity.sh which does the following things:

* Calls `composer-cli compose types` and asserts the list matches a hard-coded value. This is dependent on the CPU arch we're running on. `compose types` is executed in `osbuild-tests` but the output is not checked. The list of output types can also be checked elsewhere, e.g. in API tests

* Starts a compose, waits for it to actually be running, wait for anaconda process to be present and then cancel the running compose. Expect `compose info $UUID` to return 1; expect `/var/run/anaconda.pid` to not exist. 

* Then start another compose again

^^^ the 2 items above were related to anaconda bug IIRC.

* Wait for compose to FINISH, download the resulting image and then assert internal locations for lorax-composer.

^^^ the download part is already covered in `osbuld-tests`, testCompose() without checking anything else, just making sure we can download successfully.

Questions: 
* do we want to check the list of supported compose types ? In the past this has lead to the need to update the test suite when new output types have been introduced. I don't recall having any errors or bugs reported in this area.  

* Do the cancel & start again scenarios apply for osbuild-composer ? Doesn't seem like they do.

teg:
Thanks for writing this up.
 - I think verifying the output types would be a fine test as part of the welder tests. The benefit of having the tests in the upstream code base is that keeping this up-to-date should be fairly straight-forward, as any time people change them they are forced to also update the test suite.
 - Again, this would be something to potentially test in the weldr tests. Though I think we should restrict ourselves to the behavior of the API, not go looking for what processes are running etc. We currently do not make no guarantees about how long it takes to cancel the actual osbuild run after calling `cancel`, but from the API point of view it should immediately appear to be cancelled.

atodorov:
@bcl I've explored api_test.go to add a test for /compose/types and assert the list of supported compose types.

The addition of a new test function is rather easy but I see that `createWeldrAPI` receives something called a fixture and I can't see how to make it talk to a real running API server instead of a mocked one. 

With `rpmmd_mock.BaseFixture` I only get a qcow2 compose in the list of available types which isn't what I want. Can you give me a hint where to look or if it is even possible to run an API test against a real API server?

bcl:
The api_test.go code is a unit test, it runs against the mocked server code, not a live server.

[...]

Issue #451, opened 2020-04-01
Add compose tests to osbuild-weldr-tests


bcl:
Integration testing of composes. Possibly disabled by a cmdline flag for those times when you don't want long running tests.

teg:
The plan so far was to only have the image test generate all the images. What did you have in mind here? It may make sense to generate one image type just to verify the weldr API works, and testing different image types would still be left to the image tests, right? (Just to try to keep the overall run-time somewhat in check).

bcl:
I was thinking about doing the same thing I did in lorax, passing a ?test=1|2 argument that makes it generate a fake successful or fake failed compose. You can then examine the queue and status responses without needing to wait for an actual compose to finish.

teg:
Ah, ok. I didn't know that was possible. Makes sense.

larskarlitski:
I must say I'm not a fan of adding test code to production. (We could say this is for backwards compat to lorax-composer, though :wink:).

Could we have quick-running composes by passing in a mock store, that makes `PushCompose` be a noop?

bcl:
Yeah, I agree in principle. But if we want to be able to run this against both API servers then it needs them unless we change the tests to run lorax specific compose tests with the arguments, and osbuild tests with a noop image or somesuch.

The goal of the test is to populate the results so that the queue, log, etc. APIs can be tested. I'd prefer to do it without a full compose because that just introduces unrelated places for it to break.

larskarlitski:
>  But if we want to be able to run this against both API servers [...]

Ah I'm sorry, I didn't think about running it against both servers. Ignore my comment then :)

bcl:
PR #503 finishes this.

PR #480, opened 2020-04-07
Add test for compose types. Refs #317

atodorov:
will compare the list of supported image types (per arch) against
a hard-coded list of expected types. This simulates what we have
in lorax-composer.

Needs some more work. @teg, @bcl can you check-out my todo comments and give me a hint ?

### DIFF
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -3,11 +3,15 @@
 package client
 
 import (
+	"encoding/json"
 	"net/http"
 	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
 )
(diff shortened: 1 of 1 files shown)


bcl:
I'm already going to be working on this as part of #451 so you don't need to do anything with compose API tests right now.

For new integration tests they should go into a *_test.go file named after a descriptive part of the API route (eg. compose_test.go in this case). 

client_test.go should be reserved for just the lower level generic API functions in client.go

The structs from the api.go should be copied to internal/weldr/json.go and given a name that indicates the route that uses them and the API version they implement. eg. ModulesInfoV0

atodorov:
Brian suggested taking care of this on the cli side: https://github.com/osbuild/osbuild-composer/pull/503#pullrequestreview-393613980

Here's the PR with related changes for the CLI https://github.com/weldr/lorax/pull/1007

PR #481, opened 2020-04-07
tests: Use the require package

atodorov:
I am seeing an error when executing this as an integration test but I think it is not related to the changes I made here.

With this PR I see:
```
--- FAIL: TestRequest (0.00s)
    client_test.go:26: 
        	Error Trace:	client_test.go:26
        	            	unexpected end of JSON input
        	Test:       	TestRequest
```

with master I see
```
--- FAIL: TestRequest (0.00s)
    client_test.go:35: apiError could not parse the response: unexpected end of JSON input
```

### DIFF
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -4,45 +4,31 @@ package client
 
 import (
 	"net/http"
-	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
 )
 
 func TestRequest(t *testing.T) {
 	// Make a request to the status route
 	resp, err := Request(testState.socket, "GET", "/api/status", "", map[string]string{})
-	if err != nil {
-		t.Fatalf("Request good route failed: %v", err)
-	}
-	if resp.StatusCode != 200 {
-		t.Fatalf("Request good route: %d != 200", resp.StatusCode)
-	}
+	require.NoError(t, err)
+	require.Equal(t, 200, resp.StatusCode)
 
 	// Make a request to a bad route
 	resp, err = Request(testState.socket, "GET", "/invalidroute", "", map[string]string{})
-	if err != nil {
-		t.Fatalf("Request bad route failed: %v", err)
-	}
-	if resp.StatusCode != http.StatusNotFound {
-		t.Fatalf("Request bad route: %d != 404", resp.StatusCode)
-	}
+	require.NoError(t, err)
+	require.Equal(t, http.StatusNotFound, resp.StatusCode)
 
 	// Test that apiError returns an error response
 	_, err = apiError(resp)
-	if err != nil {
-		t.Fatalf("apiError could not parse the response: %s", err)
-	}
+	require.NoError(t, err)
 
 	// Make a request with a bad offset to trigger a JSON response with Status set to 400
 	resp, err = Request(testState.socket, "GET", "/api/v0/blueprints/list?offset=bad", "", map[string]string{})
-	if err != nil {
-		t.Fatalf("Request bad offset failed: %v", err)
-	}
-	if resp.StatusCode != http.StatusBadRequest {
-		t.Fatalf("Request bad offset: %d != 400", resp.StatusCode)
-	}
+	require.NoError(t, err)
+	require.Equal(t, http.StatusBadRequest, resp.StatusCode)
 }
 
 func TestAPIErrorMsg(t *testing.T) {
@@ -54,9 +40,7 @@ func TestAPIErrorMsg(t *testing.T) {
 
 func TestAPIResponse(t *testing.T) {
 	resp := APIResponse{Status: true}
-	if resp.String() != "" {
-		t.Fatalf("Empty APIResponse Errors doesn't return empty string: %v", resp.String())
-	}
+	require.Equal(t, "", resp.String())
 
 	resp = APIResponse{Status: false,
 		Errors: []APIErrorMsg{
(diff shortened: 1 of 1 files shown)


internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
-		t.Fatalf("Status != false: %#v", resp)
-	}
-	if len(resp.AllErrors()) < 1 {
-		t.Fatalf("GetRaw error did not return error message: %#v", resp)
-	} else if resp.Errors[0].ID != "BadLimitOrOffset" {
-		t.Fatalf("GetRaw error ID is not BadLimitOrOffset: %#v", resp)
-	}
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+	require.Equal(t, false, resp.Status)
```
msehnout, line 69: Why not to use https://godoc.org/github.com/stretchr/testify/require#False here?

larskarlitski:
> with master I see

How are you running the tests? They seem to pass no CI.

atodorov:
> How are you running the tests? They seem to pass no CI.

1) make build
2) ./osbuild-weldr-tests -test.v

internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
-	if resp.Status != false {
-		t.Fatalf("Status != false: %#v", resp)
-	}
-	if len(resp.AllErrors()) < 1 {
-		t.Fatalf("GetRaw error did not return error message: %#v", resp)
-	} else if resp.Errors[0].ID != "BadLimitOrOffset" {
-		t.Fatalf("GetRaw error ID is not BadLimitOrOffset: %#v", resp)
-	}
+	require.NoError(t, err)
+	require.NotNil(t, resp)
```
bcl, line 68: Add the error message that prints the value of b to make it easier to debug why it wasn't an error. This will also fix the linter.

internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
 
 	// Run it on a route that doesn't support offset/limit
 	b, resp, err = GetJSONAll(testState.socket, "/api/status")
-	if err == nil {
-		t.Fatalf("GetJSONAll bad route failed: %v", b)
-	}
-	if err.Error() != "Response is missing the total value" {
-		t.Fatalf("GetJSONAll bad route has unexpected total value: %v", resp)
-	}
+	require.EqualError(t, err, "Response is missing the total value")
```
bcl, line 83: Add a NotNil check on err, that prints the value of b.
And add printing the value of resp to the EqualError check.

internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
-	}
-	if resp != nil {
-		t.Fatalf("GetJSONAll request failed: %v", resp)
-	}
-	if len(b) < 100 {
-		t.Fatalf("GetJSONAll response is too short: %#v", b)
-	}
+	require.NoError(t, err)
+	require.Nil(t, resp)
+	require.GreaterOrEqual(t, len(b), 100)
```
bcl, line 79: Add a message that prints the value of b to help debugging.

internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
-		t.Fatalf("PostRaw bad request status != false: %#v", resp)
-	}
-	if len(resp.AllErrors()) < 1 {
-		t.Fatalf("GetRaw error did not return error message: %#v", resp)
-	} else if resp.Errors[0].ID != "BlueprintsError" {
-		t.Fatalf("GetRaw error ID is not BlueprintsError: %#v", resp)
-	}
+	_, resp, err := PostRaw(testState.socket, "/api/v0/blueprints/new", "nobody", nil)
+	require.NoError(t, err)
+	require.NotNil(t, resp)
```
bcl, line 90: Add 'b' back to the response and print it here.

internal/client/client_test.go
```diff
@@ -74,103 +58,57 @@ func TestAPIResponse(t *testing.T) {
...
+	require.False(t, resp.Status)
+	require.GreaterOrEqual(t, len(resp.AllErrors()), 1)
+	require.Equal(t, "BlueprintsError", resp.Errors[0].ID)
```
bcl, line 93: Add error messages to print the value of resp to both of these.

larskarlitski:
> * make build
> * ./osbuild-weldr-tests -test.v

These are integration tests that are meant to be installed. I don't know if this one requires it, but the right way to run then is by installing the osbuild-composer-test rpm.

internal/client/client_test.go
larskarlitski: `require.EqualError` checks for `nil`.

atodorov:
> These are integration tests that are meant to be installed. I don't know if this one requires it, but the right way to run then is by installing the osbuild-composer-test rpm.

What difference does it make ? I have the osbuild-composer.rpm installed and running, doesn't seem to me there is any difference if I run the binary from /usr/libexec/ or the one I have just compiled locally (+ compiling locally & running the binary from the local dir is what we advise in tests/README when working on these integration tests). IMO the problem is not how the binary itself is executed but something else.

larskarlitski:
> What difference does it make ?

Tests might install data which is required to run them. For example, `osbuild-image-tests` has a collection of test cases in JSON files. You can run the binary from the source directory, but it's looking for its data in `/usr/share`.

atodorov:
Updated. 

> In general if the error doesn't show the value I want to add a message to print it to aid in debugging things more quickly

@bcl I'm not sure what do you mean by that. How do I know when I have to add error messages to the assert statements and when not ?

atodorov:
> You can run the binary from the source directory, but it's looking for its data in /usr/share.

@larskarlitski - got it. However in this case osbuild-composer-tests.rpm is already installed along with all of the data that comes with it. I am modifying only the source for the test binary so it should be pretty safe to execute it from the local directory.

bcl:
> @bcl I'm not sure what do you mean by that. How do I know when I have to add error messages to the assert statements and when not ?

You can look at the testify code, or give it a wrong value to check and see what the output looks like. Some of them have decent defaults and print the value being compared, but especially in the cases where an error is expected and there isn't one you actually want to see what's in the normal path's variable. eg. if err is nil instead of an error, print the body to see what it actually returned.

atodorov:
Please checkout the last commit which replaces cmp.Diff with require.Equal or require.ElementsMatch.  If this is accepted I will replace the other occurrences of cmp.Diff in a separate PR.

atodorov:
> Do we need Diff anywhere? Or is the error from testify verbose enough to tell what's different (not important in these simple strings, but very useful when looking at json differences).

Good point. IDK. I'll keep this in mind when I go over the other files.

internal/client/client_test.go
atodorov: updated

internal/client/client_test.go
atodorov: message added back

internal/client/client_test.go
atodorov: This should be good to go

internal/client/client_test.go
atodorov: error message added back

internal/client/client_test.go
atodorov: added

internal/client/client_test.go
atodorov: added

PR #482, opened 2020-04-07
image-info: report errors from subprocess calls

msehnout:
The subproces.check_output function hides the output of the subprocess
it runs, using subprocess.run directly enable us to see the errors.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -75,7 +75,14 @@ def parse_unit_files(s, expected_state):
 
 
 def subprocess_check_output(argv, parse_fn=None):
-    output = subprocess.check_output(argv, encoding="utf-8")
+    try:
+        output = subprocess.check_output(argv, encoding="utf-8")
+    except subprocess.CalledProcessError as e:
+        sys.stderr.write(f"--- Output from {argv}:\n")
+        sys.stderr.write(e.stdout)
+        sys.stderr.write("\n--- End of the output\n")
+        raise
+
     return parse_fn(output) if parse_fn else output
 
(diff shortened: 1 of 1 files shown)


msehnout:
schutzbot

tools/image-info
```diff
@@ -75,8 +75,14 @@ def parse_unit_files(s, expected_state):
 
 
 def subprocess_check_output(argv, parse_fn=None):
-    output = subprocess.check_output(argv, encoding="utf-8")
-    return parse_fn(output) if parse_fn else output
+    sp = subprocess.run(argv, encoding="utf-8", stdout=subprocess.PIPE)
```
larskarlitski, line 78: Pylint usually fails when `check` is not passed. We're not running pylint here, but I think it's good form to pass it.

msehnout:
You are right, using exceptions as a control flow mechanism looks more like the Python way to do things, though I'd still print the output to stderr. That is replacing this line:
```python
     print(e.output) # in the nice way you did
```
with:
```
        sys.stderr.write(f"--- Output from {argv}:\n")
        sys.stderr.write(sp.stdout)
        sys.stderr.write("\n--- End of the output\n")
```
what do you think? It looks bit like an overkill, but seems easier to read for me.

larskarlitski:
> what do you think? The code looks bit like an overkill, but the result seems easier to read (console output, not code) for me.

Sorry yeah, that's what I meant with "in the nice way you did". What you had makes much more sense, especially sending it to `stderr`.

PR #485, opened 2020-04-07
🦥 Try running some integration tests together

major:
We may be able to shorten test time by running some tests
simultaneously.

### DIFF
--- a/jenkins/run_tests.sh
+++ b/jenkins/run_tests.sh
@@ -20,9 +20,25 @@ pushd ansible-osbuild
     playbook.yml
 popd
 
-# Run the integration tests.
-/usr/libexec/tests/osbuild-composer/osbuild-dnf-json-tests -test.v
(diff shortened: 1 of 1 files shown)


major:
schutzbot

major:
schutzbot

major:
@teg Would like your feedback on this if you think it's a good idea. It does save some time.

teg:
In general, I think we can't run tests together (as they will clobber each other). However, I agree that the tests you are running together here are independent and that should work just fine. So if that speeds things up, I'd say go for it!

I have some ideas of how we should try to speed things up long-term:
 - make sure we can enable appropriate caching (without exhausting our disk space)
 - use @dvdhrm's fedmir mirror-in-a-container to test against
 - run osbuild on swap-backed-tmpfs (this is just a matter of setting up the instances correctly with /var/cache/being backed by tmpfs

PR #486, opened 2020-04-07
📦 Use prepared containers

major:
We added this to osbuild but it was never added to osbuild-composer.

[skip ci]


major:
I cancelled the Jenkins builds since this has no effect there.

major:
@teg This was the same change we made to osbuild to use the pre-prepared containers. This should also eliminate our F32 repo issues.

PR #487, opened 2020-04-07
jsondb: introduce a simple JSON database

larskarlitski:
weldr's store is quite complex and handled serialization itself. Move
that part out into a separate package `jsondb`.

This package is more generic than the store needs: it can write an
arbitrary amount of JSON documents while the store only needs one:
state.json. This is in preparation for future work, which introduces a
queue package that builds on top of `jsondb`.

### DIFF
--- a/internal/jsondb/db.go
+++ b/internal/jsondb/db.go
@@ -1,0 +1,106 @@
+// Package jsondb implements a simple database of JSON documents, backed by the
+// file system.
+//
+// It supports two operations: Read() and Write(). Their signatures mirror
+// those of json.Unmarshal() and json.Marshal():
+//
+//     err := db.Write("my-string", "octopus")
+//
+//     var v string
+//     exists, err := db.Read("my-string", &v)
+//
+// The JSON documents are stored in a directory, in the form name.json (name as
+// passed to Read() and Write()). Thus, names may only contain characters that
+// may appear in filenames.
+
+package jsondb
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+type JSONDatabase struct {
+	dir  string
+	perm os.FileMode
+}
+
+// Create a new JSONDatabase in `dir`. Each document that is saved to it will
+// have a file mode of `perm`.
+func New(dir string, perm os.FileMode) *JSONDatabase {
+	return &JSONDatabase{dir, perm}
+}
+
+// Reads the value at `name`. `document` must deserializable from JSON. Returns
+// false if a document with `name` does not exist.
+func (db *JSONDatabase) Read(name string, document interface{}) (bool, error) {
+	f, err := os.Open(path.Join(db.dir, name+".json"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("error accessing db file %s: %w", name, err)
+	}
+	defer f.Close()
+
+	err = json.NewDecoder(f).Decode(&document)
+	if err != nil {
+		return false, fmt.Errorf("error reading db file %s: %w", name, err)
+	}
+
+	return true, nil
+}
+
+// Writes `document` to `name`, overwriting a previous document if it exists.
+// `document` must be serializable to JSON.
+func (db *JSONDatabase) Write(name string, document interface{}) error {
+	return writeFileAtomically(db.dir, name+".json", db.perm, func(f *os.File) error {
+		return json.NewEncoder(f).Encode(document)
+	})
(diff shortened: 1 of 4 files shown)


internal/jsondb/db.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Create a new JSONDatabase in `dir`. Each document that is saved to it will
+// have a file mode of `perm`.
+func New(dir string, perm os.FileMode) *JSONDatabase {
+	return &JSONDatabase{dir, perm}
+}
+
+// Reads the value at `name`. `document` must deserializable from JSON. Returns
+// false if a document with `name` does not exist or an error occured.
+func (db *JSONDatabase) Read(name string, document interface{}) (bool, error) {
```
teg, line 40: Hm, using bool as a special error feels odd.

internal/store/store.go
```diff
@@ -114,28 +114,11 @@ func New(stateDir *string) *Store {
...
-		stateFile := *stateDir + "/state.json"
-		state, err := ioutil.ReadFile(stateFile)
-
-		if err == nil {
-			err := json.Unmarshal(state, &s)
-			if err != nil {
-				log.Fatalf("invalid initial state: %v", err)
-			}
-		} else if !os.IsNotExist(err) {
+		s.db = jsondb.New(*stateDir, 0600)
```
teg, line 117: Should this be passed in so it can be easily mocked?

internal/jsondb/db.go
```diff
@@ -0,0 +1,106 @@
...
+
+	err = tmpfile.Chmod(mode)
+	if err != nil {
+		_ = os.Remove(tmpfile.Name())
+		return fmt.Errorf("error setting permissions on %s: %w", tmpfile.Name(), err)
+	}
+
+	err = writer(tmpfile)
+	if err != nil {
+		_ = os.Remove(tmpfile.Name())
```
ondrejbudai, line 89: I would personally put the `os.Remove()` to deferred function to line 75.

internal/store/store.go
```diff
@@ -114,28 +114,11 @@ func New(stateDir *string) *Store {
...
+		_, err = s.db.Read("state", &s)
```
ondrejbudai, line 118: Should the `"state"` string be extracted to a constant?

internal/jsondb/db.go
larskarlitski: Yeah I was unsure here. I went with what I thought is a common Go conventions, especially the `map` type, which I think is most similar in concept: 
```golang
// maps
value, ok := m["foo"]
// type assertion
i, ok := v.(int)
// os.LookupEnv
v, ok := os.LookupEnv("CACHE_DIRECTORY")
```

The alternative would be to return a `os.ErrNotExist` error and make callers check for that. That would leak the underlying implmentation, no?

Another alternative would be to define a custom error type `jsondb.ErrorNotExist`. (I think I prefer this one if you really don't like the `bool` return.)

internal/store/store.go
larskarlitski: I was wondering about that and decided against it, because it would be a code change that's only beneficial to testing. This is something I'm trying to avoid.

internal/jsondb/db.go
larskarlitski: I think that won't work, because we don't want to remove the file when it is renamed successfully. I've put a comment about that in line 76, but apparently it's not obvious enough :)

Is there anything I can do to make this more apparent?

internal/store/store.go
larskarlitski: It wasn't before, but I agree that that would be better. I'll fix it with the next force-push. Thanks!

internal/jsondb/db.go
ondrejbudai: I'm blind, sorry!

internal/store/store.go
larskarlitski: Ha! I just realized I've used `"state"` once and `"store"` the other time. Case in point!

internal/jsondb/db.go
teg: Ah, I see it now. Fair enough, thanks!

internal/store/store.go
larskarlitski: I've changed my mind on this after talking to @teg. It should be mocked, but it's out of scope for this PR. At least until we get all of the `stateDir` out of `store.go`.

larskarlitski:
Updated with Tom's and Ondrej's suggestions. Thanks!

internal/jsondb/db.go
larskarlitski: I've (hopefully) clarified the doc string.

Issue #476, opened 2020-04-06
dnf-json: allow the metadata_expire to be configured


teg:
Currently, we default to metadata_expire of 48h (as that is what dnf uses), however, the fedora update repositories need it set to 6h, so we need to make this configurable and hook it up correctly.

PR #489, opened 2020-04-08
Properly expire the metadata of Fedora's updates repositories.

teg:
We were using dnf's default of 48h, but that does not work for
updates repositories, as they depend on an expiration time of 6h.

Allow the metadata_expire value to be configured per repository.
If the value is unset, then never expire the metadata. Set the
value to 6h for all the fedora testing repos.

This fixes issue #476.


teg:
ok to test

major:
schutzbot

bcl:
You may want to add a call to base.update_cache() to make sure it is actually reloaded. While exploring the dnf code I found that it doesn't appear to actually check until it has an error. I think.
But even if I am, update_cache() is safe, it checks the timeout before loading new metadata.

larskarlitski:
> You may want to add a call to base.update_cache() to make sure it is actually reloaded. While exploring the dnf code I found that it doesn't appear to actually check until it has an error. I think.
> But even if I am wrong, update_cache() is safe, it checks the timeout before loading new metadata.

This is instead of calling `fill_sack()`? It looks like `update_cache` [calls it](https://github.com/rpm-software-management/dnf/blob/master/dnf/base.py#L313) itself.

bcl:
> This is instead of calling `fill_sack()`? It looks like `update_cache` [calls it](https://github.com/rpm-software-management/dnf/blob/master/dnf/base.py#L313) itself.

yeah, looks like that should work fine. It sets load_system_repo=False which is good.

teg:
I was not able to convince myself of the effect of calling `expire()` on all the repos (which is what `update_cache()` does), and since changing from `fill_sack()` would anyway be a separate patch and @bcl seemed a bit uncertain too, I'll merge this as is for now.

PR #491, opened 2020-04-08
store: use default size in image build

jkozol:
If a user creates a compose with a size of 0, the default image size for the image type should be used. Also, certain image types have requirements for the image size. In order to ensure that the proper image size is stored in the compose object, the compose's ImageBuild object uses ImageType.Size() to get the correct image size for the image type.


larskarlitski:
ok to test

jkozol:
@larskarlitski Yep. When the api returns the compose status/info it uses the size set in the ImageBuild object. I noticed this because `cockpit-composer`, which currently passes in 0 for the image size, was receiving a size of 0 for all composes it was creating even though the images were properly sized when I checked them.

larskarlitski:
Thanks!

PR #497, opened 2020-04-09
store: Set JobStarted time when PopJob is called

bcl:
If you don't set the time it ends up being the default Go time which is
1/1/1 and when you convert that using UnixNano() you get a nice big
negative number (-6795364578871345152) which then eventually shows up in
the queue, finished, and failed output as 'Fri Aug 30 17:47:39 1754'
and since the Time Travel feature is not yet complete this is
impossible.

The fix is to set it when the job is started.


larskarlitski:
ok to test

PR #498, opened 2020-04-10
Use the require and remove cmp.Diff where possible

atodorov:


atodorov:
@bcl to answer your previous question whether or not require.Equal prints the difference for more complex data structures:

I've modifed one of the scenarios in api_test.go  (which calls TestRoute) and the result is:

```
=== RUN   TestBasic
--- FAIL: TestBasic (0.01s)
    helpers.go:131: 
        	Error Trace:	helpers.go:131
        	Error:      	Not equal: 
        	            	expected: map[string]interface {}{"blueprints2":[]interface {}{}, "changes":[]interface {}{}, "errors":[]interface {}{map[string]interface {}{"id":"UnknownBlueprint", "msg":"foo: "}}}
        	            	actual  : map[string]interface {}{"blueprints":[]interface {}{}, "changes":[]interface {}{}, "errors":[]interface {}{map[string]interface {}{"id":"UnknownBlueprint", "msg":"foo: "}}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 (map[string]interface {}) (len=3) {
        	            	- (string) (len=11) "blueprints2": ([]interface {}) {
        	            	+ (string) (len=10) "blueprints": ([]interface {}) {
        	            	  },
        	Test:       	TestBasic
```

I think that's good enough, right ?

internal/test/helpers.go
```diff
@@ -105,66 +105,41 @@ func TestRoute(t *testing.T, api API, external bool, method, path, body string,
...
-		return
-	}
+	require.NoErrorf(t, err, "%s: could not read response body", path)
 
 	if expectedJSON == "" {
-		if len(replyJSON) != 0 {
-			t.Errorf("%s: expected no response body, but got:\n%s", path, replyJSON)
-		}
-		return
+		require.Equalf(t, 0, len(replyJSON), "%s: expected no response body, but got:\n%s", path, replyJSON)
```
ondrejbudai, line 114: `require.Lenf` can be used here, but I don't mind using `Equalf`, just wanted to mention that.

internal/test/helpers.go
```diff
@@ -105,66 +105,41 @@ func TestRoute(t *testing.T, api API, external bool, method, path, body string,
...
 	}
 
 	var reply, expected interface{}
 	err = json.Unmarshal(replyJSON, &reply)
-	if err != nil {
-		t.Errorf("%s: %v\n%s", path, err, string(replyJSON))
-		return
-	}
+	require.NoErrorf(t, err, "%s: json.Unmarshall failed for\n%s", path, string(replyJSON))
```
ondrejbudai, line 119: `s/Unmarshall/Unmarshal`

internal/test/helpers.go
```diff
@@ -105,66 +105,41 @@ func TestRoute(t *testing.T, api API, external bool, method, path, body string,
...
+	require.Equal(t, expected, reply)
 }
 
 func TestNonJsonRoute(t *testing.T, api API, external bool, method, path, body string, expectedStatus int, expectedResponse string) {
 	response := SendHTTP(api, external, method, path, body)
-
-	if response.StatusCode != expectedStatus {
-		t.Errorf("%s: expected status %v, but got %v", path, expectedStatus, response.StatusCode)
-	}
+	require.Equalf(t, expectedStatus, response.StatusCode, "%s: status mismatch", path)
```
ondrejbudai, line 136: `assert` here as well?

internal/test/helpers.go
atodorov: Thanks for pointing this out. I like it better.

PR #500, opened 2020-04-14
upload/aws: double the timeout for snapshot import

ondrejbudai:
us-east-1 seems to very slow these days, some imports can take up to
15 minutes. This commit raises the timeout from 10 to 20 minutes.


ondrejbudai:
Thanks for feedback! I made the commit message clearer:

> This commit raises the number of attempts before we give up. Previously, we did 40 attempts, each delayed by 15 seconds, making the total timeout equal to 10 minutes. Now we do 80 attempts with the same delay, making the total timeout 20 minutes.

ondrejbudai:
> This might be unrelated, but the [documantation for `Waiter`](https://docs.aws.amazon.com/sdk-for-go/api/aws/request/#Waiter) states that
> 
> > This type should not be used directly. The API operations provided in the service packages prefixed with "WaitUntil" should be used instead.

There isn't `WaitUntilSnapshotImported` method in the library, so I guess that's why Tom, that implemented this originally, had to implement the wait method himself.

teg:
Yes, I only implemented it as it was missing

Issue #353, opened 2020-03-18
Extend the test suite with support for alternative architectures


msehnout:
### ARM

 - [x] Request support for `partuuid` in OSBuild because aarch64 requires uefi and image-info tests require this id to be stable. ([Explanation of different UUIDs related to disks](https://unix.stackexchange.com/questions/375548/what-is-uuid-partuuid-and-ptuuid)) [1]
 - [x] As a temporary solution for the unstable `partuuid` remove the image-info part from the test and [only run the boot test](https://github.com/osbuild/osbuild-composer/blob/b0e28689fb1123c05fb2221bd1121a43db6429c0/cmd/osbuild-image-tests/main_test.go#L230)
 - [x] Extend the `/cmd/osbuild-image-tests/context-managers.go` file with support for aarch64, because it is currently [x86_64 specific](https://github.com/osbuild/osbuild-composer/blob/master/cmd/osbuild-image-tests/context-managers.go#L94)
 - [x] Generate test cases for aarch64 using the test case generation script
 - [x] Manually run the `aarch64` tests to make sure it works

### PowerPC and IBM s390x

- [ ] Reimplement support for legacy boot to match the latest update in osbuild: https://github.com/osbuild/osbuild/pull/327
- [ ] Add support for following arches and distros into appropriate `distro` subpackage:

Arch/Distro | `fedora30` | `fedora31`
------------ | ------------- | -------------
IBM `s390x` | □ | □
`ppc64le` | □ | □

- [ ] Extend the test generation script with support for `s390x` and `ppc64le`
- [ ] Manually run the `s390x` and `ppc64le` tests in any available IaaS offering (Beaker, PowerHub?)

Related PRs:
 * #376
 * https://github.com/osbuild/osbuild-composer/pull/395
 * https://github.com/osbuild/osbuild/pull/306 (authored by gicmo)
 * https://github.com/osbuild/osbuild-composer/pull/470
 * https://github.com/osbuild/osbuild-composer/pull/459
 * https://github.com/osbuild/osbuild-composer/pull/471
 * https://github.com/osbuild/osbuild-composer/pull/482
 * https://github.com/osbuild/osbuild-composer/pull/493
 * https://github.com/osbuild/osbuild-composer/pull/472
 * https://github.com/osbuild/osbuild-composer/pull/488
 * https://github.com/osbuild/osbuild-composer/pull/456
 * https://github.com/osbuild/osbuild-composer/pull/501

atodorov:
[...]

PR #501, opened 2020-04-14
osbuild: make grub2 "legacy" parameter a union

msehnout:
The "legacy" entry in grub2 stage from osbuild can be both bool and
string as you can see for example here:
https://github.com/osbuild/osbuild/blob/b53de35b0aebd52b38b979d491731aa589548b03/samples/f30-ppc64le.json#L60
In order to support that we need to make it a union but since go
does not have any unions I made a struct with a tag (IsString) so it is
like the classic tagged union pattern in C except for the union part.

The reason why a choose IsString instead of IsBool is that in case of
IsString the structure defaults to bool because bool defaults to false.
If I choose IsBool instead it would default to string, which would
require more rewriting of the code as code supports only partially
initialized structures.

### DIFF
--- a/internal/distro/fedora30/distro.go
+++ b/internal/distro/fedora30/distro.go
@@ -673,7 +673,7 @@ func (r *imageType) grub2StageOptions(kernelOptions string, kernel *blueprint.Ke
 	return &osbuild.GRUB2StageOptions{
 		RootFilesystemUUID: id,
 		KernelOptions:      kernelOptions,
-		Legacy:             !uefi,
+		Legacy:             osbuild.GRUB2LegacyFromBool(!uefi),
 		UEFI:               uefiOptions,
 	}
 }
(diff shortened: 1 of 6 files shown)


teg:
Hm, this looks like a symptom of something not being right. @gicmo could we change osbuild to only take a string or `None`, and treat `None` the way we treat `False` today? Then we could just have it as a `string,omitempty` in composer.

gicmo:
> Hm, this looks like a symptom of something not being right. @gicmo could we change osbuild to only take a string or `None`, and treat `None` the way we treat `False` today? Then we could just have it as a `string,omitempty` in composer.

So the idea would be to have: no `legacy` key: no legacy support (i.e. same as `legacy: false`). If one wants legacy boot, to specify `i386-pc` for x86/amd64 or `"powerpc-ieee1275"` for PPC64LE (s390x uses zipl not grub2, aarch64 uses grub uefi, so no legacy). If I got that right, then this should all work if we change `legacy = options.get("legacy", True)` to `legacy = options.get("legacy", None)`.

teg:
> If I got that right, then this should all work if we change `legacy = options.get("legacy", True)` to `legacy = options.get("legacy", None)`.

Hehe, I should have read your reply before preparing my patch, would have sped things up. Turns out that's exactly the patch I ended up with. Thanks!

msehnout:
Perfect, I'm closing this PR and I left a note in #353 so that we know what is needed for ppc support.

PR #503, opened 2020-04-14
Add compose tests, client functions, and fix a few minor differences in the API

bcl:


internal/client/compose.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	var composes weldr.ComposeStatusResponseV0
+	err = json.Unmarshal(body, &composes)
+	if err != nil {
+		return []weldr.ComposeEntryV0{}, nil, err
+	}
+	return composes.UUIDs, nil, nil
+}
+
+// GetComposeTypesV0 returns a list of the failed composes
```
atodorov, line 94: Comment looks like a copy/paste error ?

internal/client/compose_test.go
```diff
@@ -0,0 +1,352 @@
...
+	require.NoError(t, err, "failed with a client error")
+	require.Nil(t, resp)
+	require.Greater(t, len(composeTypes), 0)
+	var found bool
+	for _, t := range composeTypes {
+		if t.Name == "qcow2" && t.Enabled == true {
+			found = true
+			break
+		}
+	}
```
atodorov, line 47: How about the scenario where all of the enabled compose types are hard-coded as expected values ? 

Or do we not want to do this anymore ? Can still do it on the composer-cli side though (it's actually already there, just need to switch the backend rpm and possibly leave only the cli tests, not the image building ones).

bcl: I think that's better on the cli side, I wanted to make sure these run against both backends, and as unit tests, without a bunch of 'which test am I running' logic embedded in it. FYI the unit test only has qcow2 mocked.

internal/client/compose.go
bcl: Looks like I pluralized the function name when I didn't mean to.

atodorov: Oh, I meant this part `list of the failed composes`.

PR #510, opened 2020-04-17
Azure image tests

ondrejbudai:
Let's test the vhd images on Azure!

Some details about how this works:

Azure VMs have quite a lot of dependencies - a network interface, a virtual network, a network security group, a public ip address and a disk. Azure CLI and Azure Portal handle the creation of all these resources internally. However, when using the API, the caller is responsible to create all these resources before creating an actual VM.

To handle the creation of all the resources in the right order, a deployment is used. A deployment is a set of resources defined in a JSON document. It can optionally take parameters to customize each deployment. After the deployment is finished, the VM is up and ready to be tested using SSH.

Sadly, the deployments are a bit hard to clean-up. One would expect that deleting a deployment removes all the deployed resources. However, it doesn't work this way and therefore it's needed to clean up all resources "manually". For this reason, our deployment sets a unique tag on all the resources created by the deployment. After this test is finished, the API is queried for all the resources with the tag and then, they're deleted in the right order.


cmd/osbuild-image-tests/main_test.go
```diff
@@ -273,6 +273,42 @@ func testBootUsingAWS(t *testing.T, imagePath string) {
...
+	require.NoError(t, err)
+
+	// if no credentials are given, fall back to qemu
+	if creds == nil {
+		log.Print("no Azure credentials given, falling back to booting using qemu")
+		testBootUsingQemu(t, imagePath)
+		return
+	}
+
+	imageId, err := generateRandomString("")
```
atodorov, line 287: Same comment as previously for your TODO log. Why not use the image UUID instead of a random string? Will make it easier to find on the remote side too.

ondrejbudai: I agree I'm planning to change it, this is just a result of copy-paste.

ondrejbudai:
schutzbot_fedora32_base

cmd/osbuild-image-tests/main_test.go
ondrejbudai: Done

ondrejbudai:
This is now ready to be reviewed. :slightly_smiling_face:

ondrejbudai:
schutzbot_fedora31_base

cmd/osbuild-image-tests/azuretest/azure.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+
+	"github.com/osbuild/osbuild-composer/internal/upload/azure"
+)
+
+// wrapErrorf returns error constructed using fmt.Errorf from format and any
+// other args. If innerError != nil, it's appended at the end of the new
+// error.
```
larskarlitski, line 23: Hehe really looking forward to go 1.13, eh?

teg:
Travis fails with a timed-out Amazon:
```
    --- FAIL: TestImages/test/cases/fedora_31-x86_64-ami-boot.json (1566.46s)

        --- PASS: TestImages/test/cases/fedora_31-x86_64-ami-boot.json/image_info (12.55s)

        --- FAIL: TestImages/test/cases/fedora_31-x86_64-ami-boot.json/boot (1238.37s)

            main_test.go:245: 

                	Error Trace:	main_test.go:245

                	            				main_test.go:320

                	            				main_test.go:360

                	            	cannot register the image: &awserr.baseError{code:"ResourceNotReady", message:"exceeded wait attempts", errs:[]error(nil)}

                	Test:       	TestImages/test/cases/fedora_31-x86_64-ami-boot.json/boot

                	Messages:   	upload to amazon failed, resources could have been leaked
```
That's also failing on master, and is just down to Amazon having a hard time importing images at the moment. This is unrelated to this commit.

PR #515, opened 2020-04-18
ci/travis: Disable the useless arm64 tests

ondrejbudai:
The tests don't do anything useful currently. This commit disables them
to save some resources and possibly speed up the CI a bit (the free plan
allows us to have only 5 concurrent jobs, so if all slots are full,
PRs have to wait).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,8 @@ matrix:
   include:
     - name: fedora_30
       arch: amd64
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Schutzbot failing because of #516

ondrejbudai:
schutzbot

PR #526, opened 2020-04-22
tools/prepare-source: use a well-defined go version

ondrejbudai:
go mod works differently in go 1.12 and go 1.13. When someone uses the
prepare-source tool with go >= 1.13, the ci complains because it uses
go 1.12. This commit changes the script to use a well-defined go version.

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -2,6 +2,15 @@
 
 set -eux
 
-go fmt ./...
-go mod tidy
-go mod vendor
+GO_VERSION=1.12.17
+GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
+
+# this is the official way to get a different version of golang
+# see https://golang.org/doc/install#extra_versions
+go get golang.org/dl/go$GO_VERSION
+$GO_BINARY download
+
+# prepare the sources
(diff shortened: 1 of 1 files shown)


ondrejbudai:
schutzbot_fedora32_base

tools/prepare-source.sh
```diff
@@ -2,6 +2,14 @@
...
 
-go fmt ./...
-go mod tidy
-go mod vendor
+GO_VERSION=1.12.17
+GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
+
+# get golang of version GO_VERSION
+go get golang.org/dl/go$GO_VERSION
+$GO_BINARY download
```
larskarlitski, line 10: Do you mind adding a link to this here: https://golang.org/doc/install#extra_versions, just to clarify this is an officially supported way to get at a different version of go.

Unless I'm the only one that stumbled across that and started researching :)

ondrejbudai: I added the link, thanks for the feedback!

msehnout:
Travis is undergoing maintenance, codecov is not relevant for this PR, the rest of CI checks are green. I'm merging this despite the CI result.

PR #527, opened 2020-04-22
tests: clean up before exiting b/c os.Exit doesn't respect defer

atodorov:
see internal/client/integration_test.go

### DIFF
--- a/internal/client/integration_test.go
+++ b/internal/client/integration_test.go
@@ -19,29 +19,35 @@ var testState *TestState
 
 // Setup the socket to use for running the tests
 // Also makes sure there is a running server to test against
-func TestMain(m *testing.M) {
+func executeTests(m *testing.M) int {
 	var err error
 	testState, err = setUpTestState("/run/weldr/api.socket", 60*time.Second, false)
 	if err != nil {
 		fmt.Printf("ERROR: Test setup failed: %s\n", err)
-		os.Exit(1)
+		panic(err)
 	}
 
 	// Setup the test repo
 	dir, err := test.SetUpTemporaryRepository()
 	if err != nil {
 		fmt.Printf("ERROR: Test repo setup failed: %s\n", err)
-		os.Exit(1)
+		panic(err)
 	}
+
+	// Cleanup after the tests
+	defer func() {
+		err := test.TearDownTemporaryRepository(dir)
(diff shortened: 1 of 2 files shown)


ondrejbudai:
ok to test

major:
@atodorov I manually added you to the Jenkins whitelist so you won't get poked by Schutzbot for `ok to test`. 🤗

bcl:
May as well change integration_test.go too.

atodorov:
I am getting this failure which I don't quite understand:

```
goroutine 19 [running]:
github.com/osbuild/osbuild-composer/internal/client.executeTests.func1(0xc000095110, 0xacc220, 0xc00008b2c0)
	/home/runner/work/osbuild-composer/osbuild-composer/internal/client/unit_test.go:57 +0xa9
created by github.com/osbuild/osbuild-composer/internal/client.executeTests
	/home/runner/work/osbuild-composer/osbuild-composer/internal/client/unit_test.go:54 +0x6f6
FAIL	github.com/osbuild/osbuild-composer/internal/client	0.174s
```

I get it that the test API server which is started in unit_test.go has failed but I don't know why. Doesn't happen for me locally. 

Is it perhaps that I have `executeTests()` function defined in 2 test files ? Doesn't sound like it.

ondrejbudai:
From `server.Serve()` [docs](https://pkg.go.dev/net/http?tab=doc#Server.Serve):
```
ServeTLS always returns a non-nil error. After Shutdown or Close, the returned error is ErrServerClosed. 
```

So I think we should panic only if `err` is `ErrServerClosed`. I think it's basically a race condition issue - on your machine (and mine too) it kills the goroutine faster than it panics. On the CI it has enough time to panic.

If we want to be super-fancy, we can do something like:

```
if err != nil && (err is not ErrServerClosed || tests are not yet finished) {
    panic()
}
```

ondrejbudai:
`if err is **not** ErrServerClosed`

I got that the other way originally, sorry.

atodorov:
@ondrejbudai how does the last commit look like ?  From what I can tell `ErrServerClosed` is a variable in the http package not a data type, duh.

ondrejbudai:
Ouch... It looks good to me, just change the commit message or squash it and I think it's ready.

PR #537, opened 2020-04-23
Fedora31 - fix Size() for vhd

jikortus:
For some reason, the condition to round up the image size for vhd in `imageType.Size()` wrongly used `imageType.filename` instead of `imageType.file` (only for F31, OK for other distros).


ondrejbudai:
ok to test

Dne čt 23. 4. 2020 17:59 uživatel SchutzBot <notifications@github.com>
napsal:

ondrejbudai:
ok to test

Issue #257, opened 2020-02-26
Uploaded AMI images cannot be booted in AWS


ondrejbudai:
When composer builds and uploads an ami image to AWS, it doesn't boot. This is true for both Fedora and RHEL.

### The cause of this issue
We're currently building the ami images (both rhel and fedora) as `raw.xz` archives. These archives are then uploaded. However, AWS doesn't support this format. The current state is the result of c0516420cccd60dce08e083f0c0d55637ce5a7a6. The commit surely moved the ami output closer to the official images (they are indeed raw.xz) but broke the upload functionality. Why the official images are `raw.xz` is still an open question - my opinion is that the only reason is to save disk space and bandwidth on our mirrors, because those images have 6GB but majority of that is just an empty space. Using compression in those cases is indeed very useful.

### Proposed solutions
- change the ami output. AWS should support `raw`, `vmdk`, `vhdx` and `ova`.
  - `ova` - it's just wrapper over vmdk or vhdx, not very useful to us
  - `vmdk` - I couldn't get it to work (`ClientError: Disk validation failed [Unsupported VMDK File Format]`)
  - `vhdx` - **works, imho the best solution**
  - `raw` - works, but it's huge (no compression)
- when uploading to AWS in worker, decompress the image on fly (this would mean that raw is uploaded)
  - this is both slow (decompression) and slow (uploading a huge raw image)

This issue was initially tracked in this [document](https://docs.google.com/document/d/1hau9oss8TmoFvZkYy5GOnu_fmsICbNq8oWr3Q_lrcwY/edit?usp=sharing) but we forgot about it, therefore I decided to track it here.

teg:
I think the highlighted proposed solution makes a lot of sense. I do not know/understand why we are providing `.raw.xz` images, especially as generating them is very slow (as is decompressing them, which you must do before you can do anything with them).
I would be curious to learn what the difference between `vmdk` and `vhdx` is (would `vhdx` be the preferred candidate irrespective of `vmdk` currently not working).

I think we want input from @larskarlitski and @jgiardino on this, as it really fits into their image type definition work.

PR #540, opened 2020-04-24
Change the format of the ami image type to vhdx

ondrejbudai:
Prior this commit the ami image type produced raw.xz images. This was bad for
two reasons:

- The upload was broken because AWS doesn't support tar.xz format
- XZ compression is terribly slow

This commit changes the format to vhdx, which is supported by AWS and also
quite quick. See #257
why vhdx was chosen.

The osbuild requirement is bumped to 12, because that's the first version
that supports vhdx.

Also, I removed the code for handling the `raw.xz` images, so cool!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,6 +70,9 @@ jobs:
       - name: "🗄️ Clone the repository"
         uses: actions/checkout@v2
 
+      - name: "🧱 Pre-install osbuild from updates-testing repo"
+        run: dnf -y --enablerepo=updates-testing install osbuild
+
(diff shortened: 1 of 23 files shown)


ondrejbudai:
- I forgot to bump the submodule
- I enabled updates-testing when installing osbuild on f31 (we do the same thing for f32/f33)

schutzbot seems to have some issues with f32 like he sometimes has :slightly_frowning_face:

ondrejbudai:
@teg After some tinkering, this is now ready to be merged. All the CIs are green and I even tested it manually using the awesome auto-deployer.

PR #545, opened 2020-04-24
🧹 Cleanup after onboarding image tests

major:
There are four commits here:

1. Fix the debug loop variable that shows the list of image tests two times rather than showing them individually 🤦‍♂️
2. Show the test results in a YAML-ish format that is easier to read.
3. Switch from `dl.fedoraproject.org` to `download.fedoraproject.org` for image test cases (to match `osbuild` and) to download packages faster
4. Remove an incomplete comment

The third one is... *large*.


jenkins/test_runner_image.yml
```diff
@@ -5,7 +5,7 @@
     msg: |
       Running the following image test cases:
       {% for test_case in osbuild_composer_image_test_cases %}
-        - {{ osbuild_composer_image_test_cases }}
+        - {{ test_case | splitext | first }}
```
teg, line 8: For future reference, would be nice to put the description "Fix the debug loop variable that shows the list of image tests two times rather than showing them individually man_facepalming" in the commit message rather than (only) in the PR description, as the commit messages stay around in the git history.

major: @teg Good point. I forgot to do that as I stacked them up. :/

PR #546, opened 2020-04-24
Improve coverage for distro pkgs - F31, F32

jikortus:


teg:
ok to test

atodorov:
@major can you add Jiri to Jenkin's whitelist  ?

internal/distro/fedora31/distro_test.go
```diff
@@ -137,3 +140,187 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+			name:       "vhd",
+			inputSize:  10*gigaByte - 1,
+			outputSize: 10 * gigaByte,
+		},
+	}
+
+	distro := fedora31.New()
+	arch, err := distro.GetArch("x86_64")
+	assert.Nil(t, err)
+	if err == nil {
```
atodorov, line 176: that would be `if assert.NoError(t, err)`

internal/distro/fedora31/distro_test.go
```diff
@@ -137,3 +140,187 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+		for _, mapping := range sizeMap {
+			imgType, err := arch.GetImageType(mapping.name)
+			assert.Nil(t, err)
+			if err == nil {
```
atodorov, line 180: same here

internal/distro/fedora31/distro_test.go
```diff
@@ -137,3 +140,187 @@ func TestImageType_BuildPackages(t *testing.T) {
...
+			},
+			excludedPackages: []string{
+				"dracut-config-rescue",
+			},
+			bootable: true,
+		},
+	}
+	distro := fedora31.New()
+	arch, err := distro.GetArch("x86_64")
+	assert.Nil(t, err)
```
atodorov, line 261: NoError is more descriptive

internal/distro/fedora31/distro_test.go
ondrejbudai: Personally, I would prefer using `require.NoError(t, err)` and dropping the `if`.

jikortus: Thanks for the suggestion to use the assertion within the condition statement @atodorov. @ondrejbudai although it wouldn't be a very big difference, using `require` instead of `assert` wouldn't be functionally equivalent, as any failure would mean the whole test fails, as opposed to the current situation where the test can continue with another test scenario within the same test.

I'll fix the asserts and will also create a PR to backport the same changes to the already merged PR for F30.

ondrejbudai: Using 
```
arch, err := distro.GetArch("x86_64")
require.NoError(err)
```

is fine in `TestImageType_Size`, because if `err != nil`, the rest of the test is skipped.

The same in `TestImageType_BasePackages`: if `distro.GetArch` fails, there's no point in continuing the test because the `arch` variable is probably `nil`.

atodorov:
@schutzbot fedora32_base

atodorov:
schutzbot_rhel82_image

atodorov:
schutzbot_fedora32_base

teg:
This PR only affects the unit tests, so the integration test failures are unrelated. I looked through them and I verified that the problem is elsewhere. Merging.

PR #547, opened 2020-04-27
Instruct Jenkins to collect coverage data from integration tests

atodorov:
- first we need to compile the binaries to collect this data
- then when executing them specify where results will be written
- when done report to codecov.io


major:
add to whitelist

atodorov:
Looks like collecting coverage from integration tests (e.g. a stand alone running binary) is rather impossible, see https://www.cyphar.com/blog/post/20170412-golang-integration-coverage due to lack of proper tooling in the language.

I'm closing this PR as it will require too much effort which isn't worth it.

PR #553, opened 2020-04-28
Fix systemd warnings during a package removal/upgrade

ondrejbudai:
See both commits. I tested this using a copr repo and it seems to fix all the warnings.

### DIFF
--- a/golang-github-osbuild-composer.spec
+++ b/golang-github-osbuild-composer.spec
@@ -193,7 +193,8 @@ The worker for osbuild-composer
 %systemd_preun osbuild-worker@.service osbuild-remote-worker@.service
 
 %postun worker
-%systemd_postun_with_restart osbuild-worker@.service osbuild-remote-worker@.service
+# restart all the worker services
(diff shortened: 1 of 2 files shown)


golang-github-osbuild-composer.spec
```diff
@@ -190,7 +190,12 @@ The worker for osbuild-composer
 %systemd_post osbuild-worker@.service osbuild-remote-worker@.service
 
 %preun worker
-%systemd_preun osbuild-worker@.service osbuild-remote-worker@.service
+# systemd_preun uses systemctl disable --now which doesn't work well with template services.
+# The following lines mimicks its behaviour by running two commands:
```
larskarlitski, line 194: Is there an upstream bug for this?

ondrejbudai: There's now: https://github.com/systemd/systemd/issues/15620

larskarlitski: And such a detailed one! Thanks :)

PR #556, opened 2020-04-28
Add a support to upload to Azure

ondrejbudai:
All the bits and pieces were already there, it was just needed to connect them in the worker. Also, I did some small clean-ups.

### DIFF
--- a/internal/upload/azure/azure.go
+++ b/internal/upload/azure/azure.go
@@ -44,7 +44,7 @@ func UploadImage(credentials Credentials, metadata ImageMetadata, fileName strin
 	// Create a default request pipeline using your storage account name and account key.
 	credential, err := azblob.NewSharedKeyCredential(credentials.StorageAccount, credentials.StorageAccessKey)
 	if err != nil {
-		return err
+		return fmt.Errorf("cannot create azure credentials: %v", err)
 	}
 
 	p := azblob.NewPipeline(credential, azblob.PipelineOptions{})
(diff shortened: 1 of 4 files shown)


internal/upload/azure/azure.go
```diff
@@ -34,6 +35,11 @@ type ImageMetadata struct {
 // It can speed up the upload by using goroutines. The number of parallel goroutines is bounded by
 // the `threads` argument.
 func UploadImage(credentials Credentials, metadata ImageMetadata, fileName string, threads int) error {
+	// Azure cannot create an image from a storage blob without .vhd extension
```
teg, line 38: Is the requirement really on the name of the storage blob, or the format? Not a blocker either way, but I think down the line we should make a few changes:
 - make the upload target per image-type in our distro package, so that it is not possible to upload to the wrong one, if that makes sense.
 - if the requirement is about the format, rather than the name, then drop this check again when the format is "correct by construction".

teg:
I don't understand why codecov is complaining here. The coverage is not going down....

internal/upload/azure/azure.go
ondrejbudai: The storage blob must be in the `vhd` format and also its name must end with the `.vhd` extension (otherwise the UI complains).

It makes sense to have the upload target per an image type, we should definitely do it at some point (I think it really only makes sense to upload AMI image type to AWS and VHD to Azure).

atodorov:
> I don't understand why codecov is complaining here. The coverage is not going down....

Statistics I guess. You are adding new lines and the total line count increases, while coverage data remains roughly the same as before and there's a difference in how all of that is calculated. Looking at the link below not even all of the modified files are being reported so my guess is that this isn't very accurate.

internal/upload/azure/azure.go
teg: Thanks! Agreed.

PR #557, opened 2020-04-28
Remove image size from tar assembler options

jkozol:
@gicmo mentioned that  when using the https://github.com/osbuild/osbuild/pull/219 PR against the tar test cases an invalid size argument was passed in the assembler options. 

Image size should not be specified for a tar image. When the tar assembler was being created image size was not passed to it. However, the TarAssemblerOptions contained a Size field which would default to 0. This field is removed so that the tar assembler options no longer specify a size. The assembler tests are also updated to reflect this change.

I believe the tar test cases will need to be updated to remove the image size. Can I do this manually or is there a benefit to regenerating the test cases?


ondrejbudai:
I think it's alright to do the change manually in the test cases.

PR #559, opened 2020-04-28
Enable Travis for Fedora 32

ondrejbudai:
Jenkins still doesn't run all the image tests as Travis does, so let's add Fedora 32 for now, so we know it's working.

Also, I changed the boot type for f32 and rhel 8.2 to azure, I think they were still set to qemu because of " a merge race condition".


teg:
Thanks!

PR #562, opened 2020-04-29
test/cases: remove tar size

jkozol:
When building a tar image the image size should not be specified. All image size fields are removed from the assembler options in the tar test cases. This is a follow up to #557


ondrejbudai:
schutzbot fedora32_base

ondrejbudai:
schutzbot_fedora32_base

PR #567, opened 2020-04-29
Use assert.NoError in F30 distro test

jikortus:
Just a small improvement to keep things consistent (as a follow-up to code review in #546).


teg:
This is a change to the unit tests, so the integration test failure cannot be related. Merging.

PR #571, opened 2020-04-30
🦑 Cache RPMs with squid

major:
Various osbuild-composer tests require downloads from various dnf
mirrors and these can take time on slow network links. Also, it puts
undue strain on mirrors when we run a lot of tests.

Use a small squid proxy on the CI machine to cache these packages, speed
up image tests, reduce mirror strain, and potentially save a little
money in some clouds.


teg:
Sounds good! @dvdhrm is working on a long-term solution to this, but for the time being a proxy sounds like it would help a lot.

PR #573, opened 2020-04-30
test/cases: fix expected values for the image info sub-test

ondrejbudai:
osbuild removed GRUB2_ROOT_FS_UUID from grubenv in 22d131a5. This broke the Jenkins CI because it runs against osbuild master.

This commit fixes all the testcases and bumps the osbuild submodule to version 13 that changed the GRUB2_ROOT_FS_UUID behaviour.


major:
@ondrejbudai Thanks for that!

PR #576, opened 2020-04-30
🐣 Add RHEL 8.3 test cases

major:
Created with:

    ./tools/test-case-generators/generate-test-cases \
      --distro rhel-8.3 --arch x86_64 --store .osbuild/ \
      --output test/cases/


teg:
So, I wonder about this. We are now pointing to the latest nightly in the test case, so it will get out-of-sync as soon as a new nightly is released.

teg:
@dvdhrm this sounds like a job for `fedmir`.

major:
There's an 8.3.0 option, too. It seems there is some compose shuffling happening today.

major:
Nevermind. The redhat-release file for 8.3 is not signed yet.

Issue #401, opened 2020-03-26
Allow specification of an HTTP proxy for use in dnf-json


msehnout:
osbuild-composer downloads a lot of data from the repositories. For example when running our integration tests I often get an HTTP error that originates from dnf because it cannot find a working mirror. I think we could mitigate the issue if we allowed usage of HTTP proxy in dnf:
https://dnf.readthedocs.io/en/latest/conf_ref.html#options-for-both-main-and-repo

My idea is to introduce either a configuration file or a command line switch to osbuild-composer and then pass the proxy option to dnf-json.

The same problem is present in the worker, but @dvdhrm is working on a solution directly in osbuild.

dvdhrm:
Can you elaborate how using a proxy will help here? I don't think I understand.

I am also working on metalink-support for osbuild, yeah, and this should make sure we get better mirror-selection working. No guarantee that the selected mirrors will work, though.

msehnout:
I haven't yet fully investigated the issue, but I think we are downloading the same metadata over and over when running e.g. `cmd/osbuild-dnf-json-tests` which sounds exactly like something http proxy would fix. I know we cache the metadata in osbuild-composer, but I'm not sure about the tests. Anyway they fail pretty often for me now that I run them every day on arm :-)

msehnout:
I should have started with this, but this is an example of the issue:
```
Running /usr/libexec/tests/osbuild-composer/osbuild-weldr-tests
Running tests against osbuild-composer devel server using V1 API

--- FAIL: TestBlueprintDepsolveV0 (1.32s)
    blueprints_test.go:722:
        	Error Trace:	blueprints_test.go:722
        	Error:      	Expected nil, but got: &client.APIResponse{Status:false, Errors:[]client.APIErrorMsg{client.APIErrorMsg{ID:"BlueprintsError", Msg:"test-deps-blueprint-v0: DNF error occured: RepoError: Error occurred when setting up repo: Failed to download metadata for repo 'fedora': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried"}}}
        	Test:       	TestBlueprintDepsolveV0
[...]

PR #578, opened 2020-04-30
Add API v1 support to projects/source route

### LINKED ISSUE #475: The source API only implements v0 even though the server advertises v1
In WELDR API v1 the source response was changed to add an id field. The osbuild-composer API currently doesn't support the id field even though the server advertises V1 support.

Which brings up a good question about API version support. Should a V1 server be expected to support both versions of an API call? Seems to me like it shouldn't need to, but this may cause problems with UI that expects support for both. We never really defined this behavior, since lorax-composer supports both.

bcl:
Fixes #475

### DIFF
--- a/cmd/osbuild-tests/main_test.go
+++ b/cmd/osbuild-tests/main_test.go
@@ -112,7 +112,8 @@ func TestSourcesCommands(t *testing.T) {
 	require.NoErrorf(t, err, "Could not create temporary file: %v", err)
 	defer os.Remove(sources_toml.Name())
 
-	_, err = sources_toml.Write([]byte(`name = "osbuild-test-addon-source"
+	_, err = sources_toml.Write([]byte(`id = "osbuild-test-addon-source"
+name = "Testing sources add command"
 url = "file://REPO-PATH"
 type = "yum-baseurl"
 proxy = "https://proxy-url/"
--- a/internal/client/source.go
+++ b/internal/client/source.go
@@ -45,7 +45,21 @@ func GetSourceInfoV0(socket *http.Client, sourceNames string) (map[string]weldr.
 	if resp != nil || err != nil {
 		return nil, resp, err
 	}
-	var info weldr.SourceInfoV0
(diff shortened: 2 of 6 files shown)


bcl:
ok to test

major:
add to whitelist

major:
ok to test

major:
@bcl remind me to add you to the admin list tomorrow. :)

bcl:
> Very nice! Thanks for this! I noticed some comments about missing proxy support in the store, I don't think that's urgent, but would be good to add issues for missing bits (if you didn't already).

I'll do that. It's going to be another tricky one since it's a change to the store. Created #592 for this specific case, there is also #401 for an overall proxy.

bcl:
NOTE: These tests need to be restarted when jenkins is back in working order.

teg:
Would love to merge this. Seems Schutzbot is still unhappy. Might be worth retrying that now as it has seen some fixes late last week.

larskarlitski:
> Would love to merge this. Seems Schutzbot is still unhappy. Might be worth retrying that now as it has seen some fixes late last week.

No, this seems to be a valid error:
```
--- FAIL: TestSources (0.48s)
    main_test.go:220:
                Error Trace:    main_test.go:220
                                                        main_test.go:82
                Error:          Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc000093d40), Stderr:[]uint8(nil)}
                Test:           TestSources
                Messages:       Command failed: exit status 1 - { 
                                    "sources": {}, 
                                    "errors": [
                                        {
                                            "id": "UnknownSource",
                                            "msg": "osbuild-test-addon-source is not a valid source"
                                        }
                                    ]
                                }   
FAIL
```

bcl:
Yeah, it's failing consistently enough that I don't want to push it -- but I also cannot reproduce it locally so it's kinda stuck until I think of a better way to debug it.

bcl:
Rebased on master, failure is the composer-cli issue which is currently in updates-testing...

PR #580, opened 2020-05-01
test/image: fix qemu args on qemu>=4.2

ondrejbudai:
Running qemu with -accel accel= results in the following error:

qemu-system-x86_64: -accel accel=kvm:hvf:tcg: Don't use ':' with -accel,
use -M accel=... for now instead

Qemu 4.2 deprecated the -accel accel= argument. When the arg is passed in,
qemu exists status code of 1.

This commit changes the qemu command to use the recommended way of specifying
the acceleration options.

See:
https://github.com/qemu/qemu/commit/3d5e90a50bd4ffa199166e01a365f5c5995534ae

### DIFF
--- a/cmd/osbuild-image-tests/context-managers.go
+++ b/cmd/osbuild-image-tests/context-managers.go
@@ -136,7 +136,7 @@ func withBootedQemuImage(image string, ns netNS, f func() error) error {
 			"-smp", strconv.Itoa(runtime.NumCPU()),
 			"-m", "1024",
 			"-snapshot",
-			"-M", "accel=kvm:hvf:tcg",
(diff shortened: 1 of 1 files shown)


ondrejbudai:
I also disabled hvf and tcg, see the second commit for the reasoning.

atodorov:
FYI a year or so ago we had a similar issue for lorax-composer testing, especially on secondary arches and that's why we decided to switch to libvirt, which takes care to supply the appropriate arguments based on arch/specific system.

ondrejbudai:
The `weldr` tests in fedora are failing because of slow `depsolve`. The RHEL 8.2 images are somehow broken, we need to investigate later.

PR #581, opened 2020-05-01
Use sudo/become for tests

major:
Jenkins now uses a non-root user for its agent, so we will need to use
sudo for some commands and `become` within the Ansible playbooks.

### DIFF
--- a/jenkins/test.yml
+++ b/jenkins/test.yml
@@ -1,10 +1,10 @@
 ---
 
 - hosts: localhost
+  become: yes
   vars:
     passed_tests: []
     failed_tests: []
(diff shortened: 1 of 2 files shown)


jenkins/test.yml
```diff
@@ -1,10 +1,10 @@
 ---
 
 - hosts: localhost
+  become: yes
   vars:
     passed_tests: []
     failed_tests: []
-    journald_cursor: "{{ ansible_env.JOURNALD_CURSOR"
```
major, line 7: No idea how this got there, but it doesn't belong there. :)

PR #582, opened 2020-05-01
github: Fix go test coverage report

bcl:
The go test coverage report does not cover other packages unless you
list them with the -coverpkg= argument. This results in an incomplete
coverage report with oddly missing lines.

This commit lists all of the packages so that they will all be included
when running the tests and gathering the results.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,7 +44,7 @@ jobs:
         run: $(go env GOPATH)/bin/golangci-lint run
 
       - name: Run unit tests
-        run: go test -v -race -covermode atomic -coverprofile=coverage.txt ./...
+        run: go test -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
 
       - name: Send coverage to codecov.io
         run: bash <(curl -s https://codecov.io/bash)


.github/workflows/tests.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: $(go env GOPATH)/bin/golangci-lint run
 
       - name: Run unit tests
-        run: go test -v -race -covermode atomic -coverprofile=coverage.txt ./...
+        run: go test -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=$(go list ./... | tr "\n" ",") ./...
```
larskarlitski, line 47: This can be simplified to `-coverpkg=./...`

bcl: I think I tried that variation, but I'll take a look again.

bcl:
> Should testing a package count against the coverage of _another_ package?

Yes. We're trying to measure and track coverage of the whole project. Without doing this the coverage report incorrectly indicates that it didn't run some of the code (in this case it is specifically the v1 branches of the isRequiredVersion check.

teg:
Hmmmm, I'm also not convinced by this. Since coverage is just an indicator rather than making any sort of guarantee, I think it makes sense to aim for full coverage of each package in isolation. As if they were distinct projects/libraries.

To put it differently, the fact that a test covers parts of a package that a different test from a different package also covers doesn't mean that the test is redundant.

I also would find it weird if splitting a package out into a separate project would reduce the coverage of what's left...

Not totally sure, but to me there current behavior seems the most correct.

@atodorov @jikortus @jrusz, thoughts?

bcl:
I guess I'm confused as to why this is controversial. Maybe I didn't explain it well enough.

We have 1 coverage report that supposedly combines the results of all the various unit tests, but it isn't without including `-coverpkg` to combine things. There is no way to get/display coverage on a per-package basis (that I know of).

The result is obvious (to me at least) in the first pass at #578 where I added some checks for v0 vs. v1 api in internal/weldr/api.go -- in the coverage report it was reporting that only one branch was being executed. But that's not correct, the tests run both versions of the query so it was showing less coverage.

This results in much confusion (at least to me) when I look at the report and see all the new code I wrote tests for not being run.

(Note that the current version of #578 includes this patch so it is showing what I think is the correct coverage report)

jikortus:
Hmm, that's is an interesting question. I don't have a strong opinion in either way, but I think I can see Brian's point. Having the coverage as a whole could make more sense, especially if we can assume that the packages will be used together as part of one project in practice, and unlikely in a separate manner. On the other hand if this is not the case, separate testing would seem to be the better direction to follow.

atodorov:
My 2 cents on the topic: 

*Brian:*
> The go test coverage report does not cover other packages unless you
list them with the -coverpkg= argument.

This is often the case with many coverage tools for many languages. Some have arguments to automatically include files based either on package name or directory path or often both.

*Brian:*
> We're trying to measure and track coverage of the whole project.

:+1: here. I don't think I've ever seen coverage reports per module/component as separate entities.  Most services & tools (go included) can count how many times a line has been covered. The native go coverage HTML reports produce what they call a heat map, which is oddly styled with various shades of green - brighter color means more executions. 

I think CodeCov shows numeric annotation but ATM it seems to have a problem matching coverage reports to files in the repo and can't display this info.

*Tom:*
> To put it differently, the fact that a test covers parts of a package that a different test from a different package also covers doesn't mean that the test is redundant.

Nope it doesn't. Like you've said coverage info is just informative, it's not a guarantee about anything. However there are studies made which show parts with more coverage generally suffer from less bugs. You may argue that since these are executed more often if something has changed drastically it is bound to cause a failure much faster than if those lines didn't have any coverage.

The whole topic of how coverage info is used in the testing industry is controversial with lots of debate around it. To top it off there are more than 100 possible coverage metrics where tools usually measure only 2: line & branch coverage b/c these are the easiest ones.


*Tom:*
> I also would find it weird if splitting a package out into a separate project would reduce the coverage of what's left..

That's just probably statistics - include more packages in the report and the total number of lines rises, hence total % coverage drops. I'm not sure what do you mean by "splitting" but I'm also not sure how CodeCov measures the final percentage. I think it's a bit different from what the go cli reporting tools show but haven't checked this in a while.

*Brian:*
> There is no way to get/display coverage on a per-package basis (that I know of).

Correct! CodeCov doesn't support this and we don't really analyze it even if it did. FTR I think go test produces the coverage from each test independently and then combines them so it should be relatively easy to produce a per-package/per-test coverage reports. However there's no point in that. 

The only reason we're looking at coverage metric is to find out possible pieces of code which are not exercised by any kind of automated tests so we make sure to add more testing.  We're not making any other assumptions or decisions based on the coverage metric. In that situation it does make sense to count the coverage metric against all files in the project. That's the reason I wanted to include coverage from integration tests as well (technically not possible though).

Due to the same reasons coverage for anaconda works in a similar way (at least it used to in the past).

If we want to go the per-package way then all internal packages should be separated into independent repositories, with their own tests and coverage. Then osbuild-composer will be a glue between all of these packages + cli tools and whatever tests remain. 

I am :+1: for merging this PR because it will show us a more accurate picture of which parts of the entire project are exercised by automated tests.

atodorov:
> I think it makes sense to aim for full coverage of each package in isolation.

@tom - that makes sense and that's the idea of pure unit testing. However tools and reporting capabilities do not allow for this when everything is in the same repository. It can be done (e.g. execute unit tests for 1 package separately, save the HTML report) but is more work and because the process isn't streamlined (due to tooling) nobody is going to look at this.

teg:
Thanks for all your input! Seems I did not understand the tools well enough :) Let's merge this! Just needs a rebase for CI I think.

PR #583, opened 2020-05-02
osbuild-composer-koji: introduce the koji-specific service

teg:
This API is meant to specifically support building in Koji. This is still work in progress, as testing as well as koji and kerberos support is missing.


teg:
@ondrejbudai @dvdhrm @larskarlitski  I took another stab at this. Dropped the installer part for now so we can focus on landing one and then just duplicating it for the other. I think Koji is the most urgent to finish at the moment.

@Gundersanne might be worth unifying how we do this across the projects. I started out with your implementation and made some changes, so feedback welcome :)

@larskarlitski thanks for the `go generate` pointer, this is indeed very nice (no more `Makefile`).

@dvdhrm any thoughts on doing something like https://github.com/robertlove/jekyll-openapi for our website? Not sure if that's the best one, but something like that would be pretty neat I think.

larskarlitski:
> @ondrejbudai @dvdhrm @larskarlitski I took another stab at this. Dropped the installer part for now so we can focus on landing one and then just duplicating it for the other. I think Koji is the most urgent to finish at the moment.

This is amazing! Thanks for working on this :)

I find it a bit awkward that the generated types and functions are all public.

I have some nits about the server code itself. Let me know when its in a state to be reviewed. Also, I'd drop generating the client code until we need it.

teg:
I verified that the integration test works locally, but it would still need hooking up to CI (I think that can land later).

For now the subpackage is not created for fedora, only for RHEL, where it will not be shipped (at least not yet), so even if this PR is merged as is it will not end up shipping anywhere but for our internal tests.

distribution/osbuild-composer-koji.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=OSBuild Composer Koji
+After=multi-user.target
+Requires=osbuild-composer-koji.socket
+
+[Service]
+Type=simple
+ExecStart=/usr/libexec/osbuild-composer/osbuild-composer-koji
+CacheDirectory=osbuild-composer
+StateDirectory=osbuild-composer
```
larskarlitski, line 10: Does it make sense to share these directories? `fsjobqueue` does not support multiple processes on the same directory.

teg: You are so right.

teg: Fixed.

internal/kojiapi/openapi.yml
```diff
@@ -0,0 +1,174 @@
...
+            text/plain:
+              schema:
+                type: string
+        '415':
+          description: The content type is not supported
+          content:
+            text/plain:
+              schema:
+                type: string
+        '500':
```
msehnout, line 71: Is it necessary to list this return code explicitly? In such case, why not to include it in the other endpoint as well?

teg: I only included the return codes we emit. I think in this case we may want to use a different one than 500, but wasn't entirely sure so left it for now. This is triggered if depsolving fails. Not really an internal server error, just could be a million different things.

teg: On further reflection, I think we should blame the caller for this and return a bad request. I'll update.

teg: Ugh, I'm talking nonsense. We return this if enqueue fails. Maybe @larskarlitski can shed some light on hat is the right thing to return in that case.

teg: I went with a panic, as this would be a programming error.

teg:
Pushed a new version, addressing comments from @larskarlitski and @msehnout. I also added all expected return codes to the spec, and changed the success return on POST from 200 to 201. If CI is happy, I would say this is ready to be merged :)

internal/kojiapi/openapi.yml
msehnout: I'm not sure how chi works, but panicking in an endpoint handler function results in HTTP 500, no? Therefore I think it is "implicit" that 500 can happen. Anyway I was mostly curious, it's not super important.

larskarlitski: It's using `http.Server` as well, so it will return 500 on panics indeed.

I think panicking when `Enqueue` fails should be ok.

internal/kojiapi/server.go
```diff
@@ -0,0 +1,162 @@
...
+func (server *Server) GetComposeId(w http.ResponseWriter, r *http.Request) {
+	id, err := uuid.Parse(r.Context().Value("id").(string))
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Invalid format for parameter id: %s", err), http.StatusBadRequest)
+		return
+	}
+
+	status, err := server.workers.JobStatus(id)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Job %s not found: %s", id.String(), err), http.StatusBadRequest)
```
Gundersanne, line 145: This should probably be `http.StatusNotFound`

Gundersanne: Actually `server.workers.JobStatus` returns 400 when it can't find an ID, that should probably be fixed?

ondrejbudai:
Superseded by #952

PR #575, opened 2020-04-30
🧹 Clean up cruft from old jobs

major:
Delete any leftover `/tmp` files from previous test runs and turn
on the `cleanup_composer_directories` option to clean up the
osbuild-composer directories during the time the services are stopped.

Signed-off-by: Major Hayden <major@redhat.com>

PR #584, opened 2020-05-03
Split the queue from the store

### LINKED ISSUE #342: Split the job queue from the store

larskarlitski:
This introduces a new `jobqueue` package and moves both `weldr` and `rcm` to use that. This means that `rcm` is not dependent on the store anymore.

The new jobqueue saves itself to the `queue` subdirectory, one file for each job. This doesn't scale super well, but is much better than the single file approach in `store`. This, the state directory is now laid out like this (with one image built):
```
/var/lib/osbuild-composer
├── outputs
│   └── 47d47ac9-482a-4f83-864a-85e6bb45f1e9
│       └── 0
│           └── disk.qcow2
├── queue
│   └── c2676c0f-78ad-4343-bd21-c929f4a67e31.json
└── state.json
```

I took care to keep everything backwards compatible. Hope that worked.

This pull request contains a lot of smaller cleanup commits that I needed to facilitate this move.

Fixes #342


larskarlitski:
Updated to fix linting errors and a test that only failed in `make rpm`.

The remaining failure is caused by renaming the local target to `org.osbuild.weldr`. Should I drop the rename completely, or keep the `WeldrTarget` go type with the `org.osbuild.local` name for backwards compatibility?

@major while I'm happy the above was caught, I don't think we should old test state around. Could we delete `/var/lib/osbuild-composer` between runs? (The correct way to catch errors like this is to introduce a test that specifically tests the upgrade path.)

larskarlitski:
> Should I drop the rename completely, or keep the `WeldrTarget` go type with the `org.osbuild.local` name for backwards compatibility?

Answering myself: I've kept it all as is, but continued accepting `org.osbuild.local` when reading state. This can be dropped at some point in the future.

ondrejbudai:
> @major while I'm happy the above was caught, I don't think we should old test state around. Could we delete `/var/lib/osbuild-composer` between runs? (The correct way to catch errors like this is to introduce a test that specifically tests the upgrade path.)

I spoke with Major about it and this should be fixed by #575. I think it needs a rebase though.

cmd/osbuild-worker/main.go
```diff
@@ -79,7 +79,7 @@ func RunJob(job *worker.Job, uploadFunc func(*worker.Job, io.Reader) error) (*co
 
 	for _, t := range job.Targets {
 		switch options := t.Options.(type) {
-		case *target.LocalTargetOptions:
+		case *target.WeldrTargetOptions:
```
teg, line 83: As discussed, we may want to simply drop the target altogether, and make it a high-level concept in the Job. This would make it just an artefact of a job and not weldr specific.

cmd/osbuild-composer/main.go
```diff
@@ -118,11 +119,28 @@ func main() {
...
+		log.Fatalf("cannot create jobqueue: %v", err)
+	}
+
+	outputDir := path.Join(stateDir, "outputs")
+	err = os.Mkdir(outputDir, 0755)
+	if err != nil && !os.IsExist(err) {
+		log.Fatalf("cannot create output directory: %v", err)
+	}
+
+	workers := worker.NewServer(logger, jobs, store.AddImageToImageUpload)
```
teg, line 139: As discussed, we agreed that we can get away with not having the callback, but let weldr query the queue, right?

internal/compose/compose.go
```diff
@@ -21,15 +22,20 @@ func (ste *StateTransitionError) Error() string {
...
-	Targets     []*target.Target       `json:"targets"`
-	JobCreated  time.Time              `json:"job_created"`
-	JobStarted  time.Time              `json:"job_started"`
-	JobFinished time.Time              `json:"job_finished"`
-	Size        uint64                 `json:"size"`
+	Id          int               `json:"id"`
+	ImageType   common.ImageType  `json:"image_type"`
+	Manifest    *osbuild.Manifest `json:"manifest"`
+	Targets     []*target.Target  `json:"targets"`
+	JobCreated  time.Time         `json:"job_created"`
```
teg, line 29: Are these also backwards compat? Doesn't the job also store these?

internal/jsondb/db.go
```diff
@@ -55,6 +56,27 @@ func (db *JSONDatabase) Read(name string, document interface{}) (bool, error) {
 	return true, nil
 }
 
+// Returns a list of all document's names.
```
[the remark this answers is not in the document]
larskarlitski: @dvdhrm would love this.

internal/jobqueue/jobqueue.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type JobStatus int
+
+const (
+	JobPending JobStatus = iota
+	JobRunning
+	JobSucceeded
+	JobFailed
+	JobDependencyFailed
```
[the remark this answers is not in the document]
larskarlitski: To distinguish between a job that failed without a result and a job whose dependencies have failed. This is moot with the idea of not having a "success" state at all. I'll drop it.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -0,0 +1,404 @@
...
+			return nil, err
+		}
+		// Jobs that have already finished are irrelevant here: they
+		// don't need to go into the list of pending jobs, and they
+		// cannot depend on unfinished jobs anymore.
+		if j.Result != nil {
+			continue
+		}
+		// Initialize dependants for this job. Don't need to lock
+		// `q.dependants`, because there's no concurrent access yet.
```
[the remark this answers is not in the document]
larskarlitski: Yeah...

internal/store/store.go
```diff
@@ -523,22 +522,18 @@ func (s *Store) PushCompose(imageType distro.ImageType, bp *blueprint.Blueprint,
...
 			ImageBuilds: []compose.ImageBuild{
 				{
 					QueueStatus: common.IBWaiting,
-					Manifest:    manifestStruct,
+					Manifest:    manifest,
 					ImageType:   imageTypeCommon,
 					Targets:     targets,
 					JobCreated:  time.Now(),
-					Size:        imageType.Size(size),
+					Size:        size,
```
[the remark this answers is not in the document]
larskarlitski: Yeah, this was an intentional change. This is caused by a oddity in `ImageType.Manifest()`: it expects a size, but that size *must* have gone through `ImageType.Size()` before. So I pulled it out into `weldr`'s handler that also calls `Manifest()`. I think we probably want to do something about this in the future.

internal/store/store.go
```diff
@@ -501,15 +500,15 @@ func (s *Store) getImageBuildDirectory(composeID uuid.UUID, imageBuildID int) st
 	return fmt.Sprintf("%s/%d", s.getComposeDirectory(composeID), imageBuildID)
 }
 
-func (s *Store) PushCompose(imageType distro.ImageType, bp *blueprint.Blueprint, repos []rpmmd.RepoConfig, packages, buildPackages []rpmmd.PackageSpec, size uint64, targets []*target.Target) (uuid.UUID, error) {
+func (s *Store) PushCompose(manifest *osbuild.Manifest, imageType string, bp *blueprint.Blueprint, size uint64, targets []*target.Target) (uuid.UUID, error) {
```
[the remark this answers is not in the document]
larskarlitski: Yeah. Done.

cmd/osbuild-worker/main.go
larskarlitski: Agreed. I dropped the rename to `org.osbuild.weldr` and added the new fields to the local target. If we rework this in the future, there's no point to break backwards compat twice.

cmd/osbuild-composer/main.go
larskarlitski: Yeah, but I hope that this can be done in a follow-up?

internal/jobqueue/jobqueue.go
```diff
@@ -0,0 +1,89 @@
...
+// Jobs are pushed to the queue with Enqueue(). Workers call Dequeue() to
+// receive a job and FinishJob() to report one as finished.
+//
+// Each job has a type and arguments corresponding to this type. These are
+// opaque to the job queue, but it mandates that the arguments must be
+// serializable to JSON. Similarly, a job's result has opaque result arguments
+// that are determined by its type.
+//
+// A job can have dependencies. It is not run until all its dependencies have
+// finished successfully. If one of its dependencies fails, it is not run at
```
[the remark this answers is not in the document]
larskarlitski: Yeah, good point. I've changed it to be always run, which allowed me to drop quite some code, because a job now doesn't need track success status anymore. Thanks for the suggestion!

larskarlitski:
Update: incorporate all (but one) of @teg's suggestions and renamed the directory from `queue` to `jobs`.

cmd/osbuild-composer/main.go
teg: Agreed

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -0,0 +1,404 @@
...
+	// through pendingChannel() to ensure concurrent access is restricted
+	// by the mutex.
+	pending      map[string]chan uuid.UUID
+	pendingMutex sync.Mutex
+
+	// Maps job ids to the jobs that depend on it, if any of those
+	// dependants have not yet finished. Only acccess while holding the
+	// associated mutex.
+	dependants map[uuid.UUID][]uuid.UUID
+	depMutex   sync.Mutex
```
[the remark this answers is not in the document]
larskarlitski: Good point. Done.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -0,0 +1,404 @@
...
+		Id:           uuid.New(),
+		Type:         jobType,
+		Dependencies: uniqueUUIDList(dependencies),
+		CreatedAt:    time.Now(),
+	}
+
+	var err error
+	j.Args, err = json.Marshal(args)
+	if err != nil {
+		return uuid.Nil, err
```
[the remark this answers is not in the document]
larskarlitski: Of course. I've added it to all errors where it makes sense. (`readJob` for example already returned a good error - no need to wrap that again.)

internal/compose/compose.go
larskarlitski: Very good point. Had to rework a bit, see below.

larskarlitski:
Update to include @ondrejbudai's comments. Thanks for the review!

> The outstanding item is properly passing job timestamps to `weldr`. This is a regression: only `JobStarted` is reported correctly right now.

I had to rework everything a bit to make this work. Hope it still makes sense. `jobqueue` now returns the job's times in `JobQueue.JobStatus()` and it is piped through the `worker` package up to `weldr`. I think the weldr-bits could use a bit of follow-up thinking.

larskarlitski:
Update: rebase to include @major's changes to the Jenkins setup.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -0,0 +1,340 @@
+// Package fsjobqueue implements a filesystem-backed job queue. It implements
+// the interfaces in package jobqueue.
+//
+// Jobs are stored in the file system, using the `jsondb` package. However,
+// this package does not use the file system as a database, but keeps some
+// state in memory. This means that access to a given directory myst be
```
msehnout, line 6: Just a small typo: 's|myst|must|'

teg: Unless there are other issues, let's fix this in a follow-up :)

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -0,0 +1,340 @@
...
+// (de)serialized on each access.
+type job struct {
+	Id           uuid.UUID       `json:"id"`
+	Type         string          `json:"type"`
+	Args         json.RawMessage `json:"args,omitempty"`
+	Dependencies []uuid.UUID     `json:"dependencies"`
+	Result       json.RawMessage `json:"result,omitempty"`
+
+	Status     jobqueue.JobStatus `json:"status"`
+	QueuedAt   time.Time          `json:"queued-at,omitempty"`
```
teg, line 55: I think we should strive to use underscores in the json format. But really not important, can fix in a follow-up (if at all).

internal/jobqueue/jobqueue.go
```diff
@@ -0,0 +1,105 @@
...
+	"context"
+	"encoding/json"
+	"errors"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// JobQueue is an interface to a simple job queue. It is safe for concurrent use.
+type JobQueue interface {
```
teg, line 25: Should there also be a `cancelled` method? That can obviously be a follow-up, but we discussed it, so just making sure I understood the intentions correctly.

internal/rcm/api.go
```diff
@@ -204,8 +204,19 @@ func (api *API) submit(writer http.ResponseWriter, request *http.Request, _ http
...
+		_, err := writer.Write([]byte(err.Error()))
+		if err != nil {
+			panic("Failed to write response")
+		}
+		return
+	}
+
 	// Push the requested compose to the store
-	composeID, err := api.store.PushCompose(imageType, &blueprint.Blueprint{}, repoConfigs, packages, buildPackages, 0, nil)
+	composeID, err := api.store.PushCompose(manifest, imageType.Name(), &blueprint.Blueprint{}, 0, nil)
```
teg: Shouldn't `size` be passed here?

teg: And also `imageType` instead of `imageType.Name()`?

internal/jobqueue/fsjobqueue/fsjobqueue.go
larskarlitski: Indeed, sorry.

internal/jobqueue/jobqueue.go
larskarlitski: There should be indeed. Also the ability for jobs to time out. I wanted to get this in with the minimal functionality and address those as follow-ups, because this change is already big as it is.

internal/rcm/api.go
larskarlitski: I think you're looking at a commit in the middle of the PR. This is both addressed in the final one.

internal/jobqueue/jobqueue.go
teg: Makes perfect sense to me.

internal/rcm/api.go
teg: Ah, makes sense. Thanks!

PR #587, opened 2020-05-04
test/images: fix running the tests on rhel

ondrejbudai:
RHEL 8 doesn't have qemu-system-x86_64 in PATH. Instead, it is in `/usr/libexec/qemu-kvm`. This commit fixes that by detecting the host distro and choosing the right binary.

~I think I will do a small cleanup before the merge, I just need to test if this helps, therefore marking is a draft.~ Done

### DIFF
--- a/cmd/osbuild-image-tests/context-managers.go
+++ b/cmd/osbuild-image-tests/context-managers.go
@@ -11,10 +11,12 @@ import (
 	"os/exec"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ondrejbudai:
schutzbot_fedora32_base
schutzbot_rhel82_base
schutzbot_rhel82_image

ondrejbudai:
schutzbot_fedora32_base

ondrejbudai:
schutzbot_rhel82_base

ondrejbudai:
schutzbot_rhel82_image

PR #591, opened 2020-05-06
[tests] Improve coverage for distro packages - RHEL-8.2, RHEL-8.3

jikortus:


internal/distro/rhel82/distro_test.go
```diff
@@ -93,3 +96,327 @@ func TestFilenameFromType(t *testing.T) {
...
+		"xz",
+	}
+	buildPackages := map[string][]string{
+		"x86_64":  x8664BuildPackages,
+		"aarch64": aarch64BuildPackages,
+	}
+	d := rhel82.New()
+	for _, archLabel := range d.ListArches() {
+		archStruct, err := d.GetArch(archLabel)
+		if err != nil {
```
atodorov, line 129: You can use assert.NoError or require.NoError everywhere where comparing `err`

internal/distro/rhel82/distro_test.go
```diff
@@ -93,3 +96,327 @@ func TestFilenameFromType(t *testing.T) {
...
+			t.Errorf("d.GetArch(%v) returned err = %v; expected nil", archLabel, err)
+			continue
+		}
+		for _, itLabel := range archStruct.ListImageTypes() {
+			itStruct, err := archStruct.GetImageType(itLabel)
+			if err != nil {
+				t.Errorf("d.GetArch(%v) returned err = %v; expected nil", archLabel, err)
```
atodorov, line 136: that's `assert.NoError` with a message parameter, no ?

internal/distro/rhel82/distro_test.go
```diff
@@ -93,3 +96,327 @@ func TestFilenameFromType(t *testing.T) {
...
+		},
+	}
+	for _, mapping := range imgMap {
+		arch, err := distro.GetArch(mapping.arch)
+		assert.Nil(t, err)
+		if err == nil {
+			for _, imgName := range mapping.imgNames {
+				imgType, err := arch.GetImageType(imgName)
+				assert.Nil(t, err)
+				if err == nil {
```
atodorov, line 182: assert.Nil returns a boolean. Same above & below.

internal/distro/rhel82/distro_test.go
```diff
@@ -93,3 +96,327 @@ func TestFilenameFromType(t *testing.T) {
...
+	}
+
+	for _, a := range arches {
+		actualArch, err := distro.GetArch(a.name)
+		if !a.errorExpected {
+			assert.Equal(t, a.name, actualArch.Name())
+			assert.Nil(t, err)
+		} else {
+			assert.Nil(t, actualArch)
+			assert.NotNil(t, err)
```
atodorov, line 409: See `assert.EqualError` and `assert.Error`

internal/distro/rhel82/distro_test.go
jikortus: I think not in the places with the `continue` statement, but thanks for pointing this out, I fixed the asserts in the remaining occurrences.

atodorov: assert returns a boolean and takes care of logging the issue for you so you can do
```
if assert.NoError(err, "...") {
    continue
}
``

Comment (jikortus) [reply internal/distro/rhel82/distro_test.go]: I see... fixed :)

File: internal/distro/rhel82/distro_test.go [hunk 5]
```diff
@@ -93,3 +96,321 @@ func TestFilenameFromType(t *testing.T) {
...
+		archStruct, err := d.GetArch(archLabel)
+		if assert.NoErrorf(t, err, "d.GetArch(%v) returned err = %v; expected nil", archLabel, err) {
+			continue
+		}
+		for _, itLabel := range archStruct.ListImageTypes() {
+			itStruct, err := archStruct.GetImageType(itLabel)
+			if assert.NoErrorf(t, err, "d.GetArch(%v) returned err = %v; expected nil", archLabel, err) {
+				continue
+			}
+			reflect.DeepEqual(itStruct.BuildPackages(), buildPackages[archLabel])
```
atodorov, line 137: Why do we need `DeepEqual` instead of `assert.Equal` ? What are we trying to validate here that assert isn't capable of ?

atodorov: Note2: as-is `DeepEqual` will return a boolean and we don't really care about the value so in case of failure we'll not know about it.

jikortus: There was no explicit need for `DeepEqual` IMO, it was just a part of a previously existing code I reused (which was, AFAIK, based on a pre-generated boilerplate code). I can change it to `assert.Equal`, however in such a case I should also change other occurrences in the code that has already been merged, for the sake of consistency (in a separate PR). The second note seems to have a bigger impact, thanks for pointing this out.

atodorov: +1 for using `asserts.Equal` if we don't need DeepEqual. After you make the related changes maybe we can drop that package altogether.

jikortus:
I just found out that I could possibly use `assert.ElementsMatch` instead of `assert.Equal`, which wouldn't check for items order in the lists. Would it be a better option? (I also mentioned this in #604).

ondrejbudai:
The jenkins fail is unrelated.

PR #594, opened 2020-05-09
store: make marshalling internal

teg:
Make the types used for marshalling internal to the store package to reduced the interdependencies between our packages. In follow-ups we will simplify the store API further by exposing real types/objects rather than ones meant for marshalling.


internal/store/fixtures.go
```diff
@@ -0,0 +1,124 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/compose"
+	"github.com/osbuild/osbuild-composer/internal/target"
+)
+
+func FixtureBase() *Store {
```
larskarlitski, line 13: It's a bit odd to have the fixtures be public types on the store. Is the idea to make them private at some point?

teg: Not sure about that. I think keeping them public is in keeping with the way the fixtures are meant to be used, but I'm happy to discuss that in a follow-up, unless you think it is a blocker?

internal/weldr/api_test.go
```diff
@@ -432,6 +433,10 @@ func TestBlueprintsDepsolve(t *testing.T) {
 }
 
 func TestCompose(t *testing.T) {
+	arch, err := test_distro.New().GetArch("x86_64")
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
teg: Agreed. Fixed!

internal/store/json.go
```diff
@@ -86,20 +87,28 @@ func newWorkspaceFromV0(workspaceStruct workspaceV0) map[string]blueprint.Bluepr
...
 		c := Compose{
 			Blueprint: composeStruct.Blueprint,
 		}
-		if len(composeStruct.ImageBuilds) == 0 {
-			panic("the was a compose with zero image builds, that is forbidden")
-		}
 		for _, imgBuild := range composeStruct.ImageBuilds {
+			imgTypeString, valid := imgBuild.ImageType.ToCompatString()
+			if !valid {
+				// Invalid type strings in serialization format, this may happen
```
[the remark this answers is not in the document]
teg: Yeah, you are right. Added.

internal/test/helpers.go
```diff
@@ -157,6 +158,14 @@ func Ignore(what string) cmp.Option {
 	return cmp.FilterPath(func(p cmp.Path) bool { return p.String() == what }, cmp.Ignore())
 }
 
+func CompareImageTypes() cmp.Option {
```
msehnout, line 161: What is the idea behind this function? Although the code seems reasonable, it's not obvious why only these 3 properties were selected for comparison as opposed to e.g. MIMEType or ModulePlatformID. Maybe add a comment explaining it?

internal/store/json.go
```diff
@@ -0,0 +1,303 @@
...
+
+func newComposesFromV0(composesStruct composesV0) map[uuid.UUID]compose.Compose {
+	composes := make(map[uuid.UUID]compose.Compose)
+
+	for composeID, composeStruct := range composesStruct {
+		c := compose.Compose{
+			Blueprint: composeStruct.Blueprint,
+		}
+		if len(composeStruct.ImageBuilds) == 0 {
+			panic("the was a compose with zero image builds, that is forbidden")
```
msehnout: `s|the|there|`?

internal/store/json.go
```diff
@@ -0,0 +1,303 @@
...
+		blueprintsCommits: newCommitsFromV0(storeStruct.Commits),
+	}
+
+	// Backwards compatibility: fail all builds that are queued or
+	// running. Jobs status is now handled outside of the store
+	// (and the compose). The fields are kept so that previously
+	// succeeded builds still show up correctly.
+	for composeID, compose := range store.composes {
+		if len(compose.ImageBuilds) == 0 {
+			panic("the was a compose with zero image builds, that is forbidden")
```
msehnout: same as above

internal/store/json.go
```diff
@@ -239,30 +245,46 @@ func newStoreFromV0(storeStruct storeV0, arch distro.Arch) *Store {
...
+	workspaceStruct := make(workspaceV0)
+	for name, blueprint := range workspace {
+		workspaceStruct[name] = blueprint.DeepCopy()
+	}
+	return workspaceStruct
+}
+
+func newComposeV0(compose Compose) composeV0 {
+	return composeV0{
+		Blueprint: compose.Blueprint, // TODO: deep copy
```
msehnout, line 255: Why is this TODO? :) Is there any problem with calling `compose.Blueprint.DeepCopy()`?

internal/test/helpers.go
teg: The idea is that given a triple distro, arch, type the object you get by resolving them should be considered the same, and indeed all the other functions return the same results. I'll add a comment.

internal/store/json.go
teg: Thanks!

internal/store/json.go
teg: Thanks!

internal/store/json.go
teg: This looks like an oversight, I'll fix!

internal/store/json.go
teg: Ah, this was dropped in a later patch. I'll leave the typo in the git history ;)

internal/store/json.go
teg: This was dropped in a later patch, I'll leave the typo in the git history ;)

teg:
Addressed the reviews, added a few more tests (but more would be great!). Ready to be merged from my point of view.

PR #596, opened 2020-05-10
store: simplify TestCompose

larskarlitski:
After the queue rework, TestCompose() was the only user of
UpdateImageBuildInCompose() and Compose.UpdateState(). Set all required
fields in TestCompose() directly instead of going through those.


larskarlitski:
Rebased to include the schutzbot fixes.

ondrejbudai:
Jenkins fail seems to be unrelated.

PR #597, opened 2020-05-11
worker: switch to `--output-diectory=DIR`

dvdhrm:
Since 2 releases `osbuild` accepts an `--output-directory=DIR` argument
which lets us decide where to place generated artifacts. Switch over to
it, rather than digging into the store, to make sure we will not access
the osbuild store when parallel cleanups are ongoing (which are not yet
a thing, though).

### DIFF
--- a/cmd/osbuild-image-tests/constants/constants-travis.go
+++ b/cmd/osbuild-image-tests/constants/constants-travis.go
@@ -4,12 +4,12 @@ package constants
 
 import "os/exec"
 
-func GetOsbuildCommand(store string) *exec.Cmd {
+func GetOsbuildCommand(outputDirectory string) *exec.Cmd {
 	cmd := exec.Command(
 		"python3",
 		"-m", "osbuild",
 		"--libdir", ".",
-		"--store", store,
+		"--output-directory", outputDirectory,
 		"--json",
 		"-",
 	)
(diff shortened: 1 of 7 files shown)


dvdhrm:
The CI failure seems unrelated to me. But maybe someone with more insight can look at it?

```
                	            	- (string) (len=10) "bootloader": (string) (len=4) "grub",

                	            	+ (string) (len=10) "bootloader": (string) (len=7) "unknown",
```

dvdhrm:
> Thanks! osbuild-composer currently depends on [osbuild >= 12](https://github.com/osbuild/osbuild-composer/blob/master/osbuild-composer.spec#L47). Is `--output-directory` already in 12?

Yeah, introduced with `osbuild-12`.

> > The CI failure seems unrelated to me. But maybe someone with more insight can look at it?
> 
> @gicmo has seen these before. I think they're flakes.

Right, then I will ignore it for now. Thanks!

dvdhrm:
I just noticed there are 2 more tests that I could convert as well. We then no longer hard-code `refs/` anywhere. And I can rerun the CI when I repush :)

dvdhrm:
`grep -r generate-test-case .` found nothing, so I am a bit unsure how to verify `generate-test-case` still works. Can someone let me know how I can run and test this tool?

ondrejbudai:
`tools/test-case-generators/generate-test-case` is called by `tools/test-case-generators/generate-test-cases`. I think the following line should work:

```
tools/test-case-generators/generate-test-cases --distro fedora-32 --arch x86_64 --store . --output test/cases
```

dvdhrm:
I used `mkdir store && ... --store store` and it works fine for me.

PR #598, opened 2020-05-11
Switch to using the RHEL CDN

jkozol:
When loading repositories osbuild composer now connects to the cdn instead of the nightly rhel releases. This allows subscribed machines to use osbuild-composer without having to be on the internal vpn. The image tests still use the nightly repos.


repositories/rhel-8.1.json
```diff
@@ -2,24 +2,24 @@
     "aarch64": [
         {
             "id": "baseos",
-            "baseurl": "http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/latest-RHEL-8.1/compose/BaseOS/aarch64/os",
+            "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/aarch64/baseos/os",
```
larskarlitski, line 5: Please include the minor version in these URLs. Otherwise, all of these will resolve to the latest RHEL 8.

`"https://cdn.redhat.com/content/dist/rhel8/8.1/aarch64/baseos/os"`

repositories/rhel-8.3.json
```diff
@@ -2,24 +2,24 @@
     "aarch64": [
         {
             "id": "baseos",
-            "baseurl": "http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/latest-RHEL-8.3/compose/BaseOS/aarch64/os",
+            "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/aarch64/baseos/os",
```
larskarlitski, line 5: 8.3 is not yet released. Let's keep using the internal URLs until we thought about a good solution.

internal/rpmmd/repository.go
```diff
@@ -128,10 +134,10 @@ func (re *RepositoryError) Error() string {
 func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfig, error) {
 	var f *os.File
 	var err error
-	path := "/repositories/" + distro + ".json"
+	repoPath := "/repositories/" + distro + ".json"
 
 	for _, confPath := range confPaths {
-		f, err = os.Open(confPath + path)
+		f, err = os.Open(confPath + repoPath)
```
larskarlitski, line 140: Is this an unrelated change?

internal/rpmmd/repository.go
```diff
@@ -150,6 +156,39 @@ func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfi
 		return nil, err
 	}
 
+	if strings.HasPrefix(distro, "rhel-8") {
```
larskarlitski, line 159: Please don't hard code `rhel-8` here. This module has otherwise no knowledge of the names of distributions. If we ever add another one that needs subscriptions enabled (`rhel-9`, maybe? :wink:), it will be hard to find all the places that we've special-cased names.

Instead, pass this information through `RepoConfig`. I think there's no way around splitting it into two structs—one for reading from `/usr/share/osbuild-composer/repositories`, that contains `"rhsm": true`, and one for passing to `dnf-json`, that contains the `ssl*` fields.

internal/rpmmd/repository.go
```diff
@@ -150,6 +156,39 @@ func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfi
...
+		var sslClientKey string
+		var sslClientCert string
+
+		regexKey := regexp.MustCompile("^[0-9]*-key.pem")
+		regexCert := regexp.MustCompile("^[0-9]*.pem")
+
+		entitlementFiles, err := ioutil.ReadDir("/etc/pki/entitlement/")
+		if err != nil {
+			return nil, &RepositoryError{"LoadRepositories failed: unable to read /etc/pki/entitlement/"}
```
larskarlitski, line 168: Include the error message:
```suggestion
			return nil, &RepositoryError{fmt.Sprintf("unable to read /etc/pki/entitlement/: %v", err)}
```

internal/rpmmd/repository.go
```diff
@@ -150,6 +156,39 @@ func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfi
 		return nil, err
 	}
 
+	if strings.HasPrefix(distro, "rhel-8") {
+		var sslClientKey string
+		var sslClientCert string
+
+		regexKey := regexp.MustCompile("^[0-9]*-key.pem")
+		regexCert := regexp.MustCompile("^[0-9]*.pem")
```
larskarlitski, line 164: Good idea to restrict which files we're reading! `MustCompile` panics when the passed regexp is invalid. It's probably better to move this into a global variable, so that that panic happens when starting the program.

Also ,`[0-9]+` :wink:

internal/rpmmd/repository.go
```diff
@@ -150,6 +156,39 @@ func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfi
...
+			if matchKey {
+				sslClientKey = path.Join("/etc/pki/entitlement/", file.Name())
+				continue
+			}
+
+			matchCert := regexCert.MatchString(file.Name())
+			if matchCert {
+				sslClientCert = path.Join("/etc/pki/entitlement/", file.Name())
+			}
+		}
```
larskarlitski, line 181: This will break if there are multiple entitlements in the directory, because it is not ensuring that the `sslClientCert` and `sslClientKey` are from matching certificates. How about only checking for `regexCert`, and then loading the same file name with `-key` appended?

Also, stop iterating after finding the first one.

internal/rpmmd/repository.go
```diff
@@ -150,6 +156,39 @@ func LoadRepositories(confPaths []string, distro string) (map[string][]RepoConfi
...
+
+		for arch, repoConfigs := range repos {
+			for i := range repoConfigs {
+				repos[arch][i].IgnoreSSL = false
```
larskarlitski, line 185: Why are you overriding this here again?

repositories/rhel-8.1.json
jkozol: This is fixed

internal/rpmmd/repository.go
jkozol: This was because I was using the `path` library but I have now switched to use `path/filepath`

repositories/rhel-8.3.json
jkozol: This is fixed

internal/rpmmd/repository.go
jkozol: I have added the rhsm field to the repo files and use that to determine if a repo needs rhsm certs.

internal/rpmmd/repository.go
jkozol: I changed the error message to be this.

internal/rpmmd/repository.go
jkozol: The regex is global and fixed now.

internal/rpmmd/repository.go
jkozol: This is fixed

internal/rpmmd/repository.go
jkozol: I have now switched IgnoreSSL to be SSLVerify to better match the fields supported by dnf and this section of code has been changed.

jkozol:
This now outputs a manifest where each url can also contain a secret of `org.osbuild.rhsm` which alerts osbuild to add rhsm certs when curling those urls.

This PR still needs to be updated so the tests pass and the sources need to be more specific for when secrets are attached to the urls in the manifest.

jkozol:
@teg This PR is mostly ready. I am unsure where to store the rhsm secrets in repositories.go and the use of the repo id between the 3 different repo structs needs to be clarified more. Also, the test cases need to be regenerated.

teg:
I pushed some patches as suggestions for the open questions. Feel free to rebase/squash/drop/...

Otherwise: this is looking great!

jkozol:
This PR is currently rebased on top of #678 and should not be merged before it.

major:
@jkozol 🤞

PR #602, opened 2020-05-11
codecov: add 5% threshold

larskarlitski:
We're using code coverage as additional information, but don't want to
gate pull requests on it.


larskarlitski:
Update: use `threshold` instead - turns out codecov doesn't post status at all with `informational`.

teg:
CI failures must be unrelated. Merging.

PR #603, opened 2020-05-12
CI: Build RPMs in mock

major:
By building RPMs inside a mock chroot, we build RPMs more like koji and
brew would. This ensures that we build RPMs with the correct
libraries without additional cruft from a customized system.

Also, install RPMs from the mock after they are built instead of
compiling them with Ansible.

Remove the jenkins.osbuild.org configuration now that the Schutzbot PSI Jenkins is stable.


major:
Note that at this moment, the base tests are run first, then the image tests. We can split these in the pipeline since the deployment part is faster now (since we don't compile anything), but I'm trying to keep this change simpler for now.

major:
The only issue found was the intermittent `FAILED`/`FINISHED` bug that intermittently appears on Fedora 32. https://osbuildci.cloud.paas.psi.redhat.com/job/osbuild/job/osbuild-composer/job/PR-603/69/display/redirect

major:
🛑 Hold off on merging this for now. I'd like to get some of this mocking logic into ansible-osbuild instead.

PR #604, opened 2020-05-12
[tests] use assert.Equal instead of reflect.DeepEqual in distro_test

jikortus:
The asserts should do the same job and also at the same time make sure that the two compared lists are equal. In the original state, the comparison was performed, but the result was thrown away - and actually the result was false, as the two lists were different due to non-equal order of items.

This change originates from a suggestion by @atodorov in #591.

EDIT: I just found out that I could possibly use `assert.ElementsMatch` instead of `assert.Equal` and leave the items order in the list as it was before. What would be the better option?

### DIFF
--- a/internal/distro/fedora30/distro_test.go
+++ b/internal/distro/fedora30/distro_test.go
@@ -1,7 +1,6 @@
 package fedora30_test
 
 import (
-	"reflect"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


larskarlitski:
> EDIT: I just found out that I could possibly use `assert.ElementsMatch` instead of `assert.Equal` and leave the items order in the list as it was before. What would be the better option?

Yeah, please do that, because it seems more robust to me.

PR #606, opened 2020-05-12
schutzbot: use -test.timeout argument

larskarlitski:
Use golang's `test.timeout` to set the timeout instead of using
ansible's `async` property. The latter doesn't respect `register` when
running into a timeout, which means there are no logs in that case.

Simplify these playbooks by setting `ignore_errors: true` on the
`command` module, so that it's easier to collect logs.

### DIFF
--- a/schutzbot/test_runner_base.yml
+++ b/schutzbot/test_runner_base.yml
@@ -3,12 +3,10 @@
 - block:
 
     - name: "Run {{ test.name }} test"
-      command: "{{ tests_path }}/{{ test.name }} -test.v"
+      command: "{{ tests_path }}/{{ test.name }} -test.v -test.timeout {{ test.timeout * 60 }}s"
       args:
         chdir: "{{ tests_working_directory }}"
-      register: async_test
-      async: "{{ test.timeout * 60 }}"
-      poll: "{{ polling_interval }}"
+      register: test_cmd
 
     - name: "Mark {{ test.name }} as passed"
       set_fact:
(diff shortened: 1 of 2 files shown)


schutzbot/test_runner_base.yml
```diff
@@ -1,36 +1,31 @@
...
-        content: |
-          Logs from {{ test.name }}
-          ----------------------------------------------------------------------
-          stderr:
-          {{ async_test.stderr }}
-          ----------------------------------------------------------------------
-          stdout:
-          {{ async_test.stdout }}
+- name: "Run {{ test.name }} test"
+  command: "{{ tests_path }}/{{ test.name }} -test.v -test.timeout {{ test.timeout * 60}}s"
```
major, line 4: Need a little bit of whitespace after the `60`, like `{{ test.timeout * 60 }}`

schutzbot/test_runner_image.yml
```diff
@@ -17,48 +17,43 @@
...
-
-    - name: "Mark image tests as failed"
-      set_fact:
-        failed_tests: "{{ failed_tests + ['image_tests'] }}"
+# NOTE(mhayden): The fancy jinja2 here ensures that the last test case
+# in the list does not have a backslash added. That would make the shell
+# upset and nobody likes it when that happens.
+- name: "Run image tests"
+  command: |
+    {{ image_test_executable }} -test.v -test.timeout {{image_test_timeout * 60}}s \
```
major, line 25: Need a little bit of whitespace after the `60`, like `{{ test.timeout * 60 }}`

larskarlitski:
Update to keep using `block` / `rescue`. I was confused in thinking that wouldn't work anymore, but it does and keeps this change much smaller. Thanks @major!

> Need a little bit of whitespace after the `60`, like `{{ test.timeout * 60 }}`

Oops. Fixed.

PR #607, opened 2020-05-12
Add blueprint name validation

### LINKED ISSUE #522: Empty name blueprint should not be allowed to create
Blueprint with empty name can be created.
```toml
name = ""
description = ""
version = "0.0.1"
modules = []
groups = []

[[packages]]
name = "openssh-server"
version = "*"
```
The `state.json` confirmed empty name blueprint have been pushed into store.
```json
{"blueprints":{"":{"name":"","description":"","version":"0.0.1","packages":[{"name":"openssh-server","version":"*"}],"modules":[],"groups":[]}},"workspace":{},"composes":{},"sources":{},"changes":{"":{"c3743814b30e0bb2a5c607d6f85f22e1e126c879":{"commit":"c3743814b30e0bb2a5c607d6f85f22e1e126c879","message":"Recipe , version 0.0.1 saved.","revision":null,"timestamp":"2020-04-21T22:57:37Z"}}},"commits":{"":["c3743814b30e0bb2a5c607d6f85f22e1e126c879"]}}
```

The API "blueprints/new" or "store.push" should have a name validation. Empty name should not be allowed, and the API should response an error message.

bcl:
Closes #522

### DIFF
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -63,6 +63,42 @@ func TestPostEmptyTOMLBlueprintV0(t *testing.T) {
 	require.False(t, resp.Status, "did not return an error")
 }
 
+// POST a TOML blueprint with an empty name
+func TestPostEmptyNameTOMLBlueprintV0(t *testing.T) {
+	// Use a blueprint with an empty Name
+	bp := `
+		name=""
+		version="0.0.1"
+		[package]
+		name="bash"
+		version="*"
+		`
+	resp, err := PostTOMLBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "failed with a client error")
+	require.NotNil(t, resp, "did not return an error")
+	require.False(t, resp.Status, "wrong Status (true)")
+	require.Equal(t, "InvalidChars", resp.Errors[0].ID)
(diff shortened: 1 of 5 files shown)


internal/client/blueprints_test.go
```diff
@@ -63,6 +63,42 @@ func TestPostEmptyTOMLBlueprintV0(t *testing.T) {
...
+		name=""
+		version="0.0.1"
+		[package]
+		name="bash"
+		version="*"
+		`
+	resp, err := PostTOMLBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "failed with a client error")
+	require.False(t, resp.Status, "did not return an error")
+	require.False(t, resp.Status, "wrong Status (true)")
```
atodorov, line 79: What is the point of 2 sequential `require.False` both checking `resp.Status` ? In case `resp.Status` isn't False the test will abort on line 78 and line 79 will never be executed.

atodorov: In some of the other tests below you have only the 2nd require, not the 1st one.

bcl: It makes sure there isn't a race condition with the universe failing inbetween the tests ;)

Or more likely it's a copy-paste error, good catch!

PR #608, opened 2020-05-12
jenkins: use -test.timeout argument

larskarlitski:
This is the same as 6c611b487, but for the jenkins directory
(configuration for Jenkins inside Red Hat's VPN).

Use golang's `test.timeout` to set the timeout instead of using
ansible's `async` property. The latter doesn't respect `register` when
running into a timeout, which means there are no logs in that case.

### DIFF
--- a/jenkins/test_runner_base.yml
+++ b/jenkins/test_runner_base.yml
@@ -3,12 +3,10 @@
 - block:
 
     - name: "Run {{ test.name }} test"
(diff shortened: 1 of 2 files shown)


larskarlitski:
This patch only touches files that are relevant to `schutzbot-psi/pr-merge`. It is expected to be failing, and this patch is working correctly in that it shows the logs for the failing test.

I've verified that the other failures are unrelated and am merging this.

PR #610, opened 2020-05-13
distro/rhel83: use repos from a nightly compose for now

ondrejbudai:
Previously, the path to RHEL 8.3 repos was bad because RHEL 8.3 composes are
not yet under rel-eng directory. This commit changes the path to use
the latest nightly path.


ondrejbudai:
I don't know about this. The nightlies don't have signed packages, so the `org.osbuild.rpm` stage will fail to build the image...

atodorov:
This PR is in conflict with #598, isn't it ?

teg:
Yes, this conflicts, and will be redundant.

PR #613, opened 2020-05-14
Remove most of the image tests from Travis

ondrejbudai:
This PR does 2 things:

1) Schutzbot didn't build ami and openstack images, so I added them.
2) Travis is now used to do only image tests with uploading, because that's not currently supported by Schutzbot.


major:
@ondrejbudai Fixed the refspec in ansible-osbuild.

Issue #611, opened 2020-05-13
osbuild-tests: Intermittent failure in TestEverything on Fedora 32


major:
I'm seeing this failure intermittently on Fedora 32:

```
TASK [Run osbuild-tests test] **************************************************
Wednesday 13 May 2020  14:17:43 +0000 (0:00:00.505)       0:16:04.344 ********* 
fatal: [localhost]: FAILED! => changed=true 
  cmd:
  - /usr/libexec/tests/osbuild-composer/osbuild-tests
  - -test.v
  - -test.timeout
  - 1800s
  delta: '0:18:34.667167'
  end: '2020-05-13 14:36:18.234157'
  msg: non-zero return code
  rc: 1
  start: '2020-05-13 14:17:43.566990'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    === RUN   TestEverything
        TestEverything: main_test.go:104:
                    Error Trace:    main_test.go:104
                                                            main_test.go:61
                    Error:          Not equal:
                                    expected: "FAILED"
                                    actual  : "FINISHED"
  
                                    Diff:
                                    --- Expected
                                    +++ Actual
                                    @@ -1 +1 @@
                                    -FAILED
                                    +FINISHED
                    Test:           TestEverything
                    Messages:       Unexpected compose result: FAILED
    --- FAIL: TestEverything (1114.00s)
    === RUN   TestSources
    --- PASS: TestSources (0.66s)
    FAIL
  stdout_lines: <omitted>
```

It most recently occurred here during work on #603:

https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-603/69/pipeline

atodorov:
@major do we have logs from this failure (I wasn't able to find them)? Could be an issue with the mirrors/depsolve.

major:
Ah darn, this is a duplicate of #566. I'll close this one.

Issue #566, opened 2020-04-29
Fedora 31/32: osbuild-tests failing (got FINISHED instead of FAILED)


major:
This was found when testing RHEL 8.2 with the `osbuild-tests` test in osbuild/osbuild-composer#564:

```
Logs from osbuild-tests
----------------------------------------------------------------------
stderr:

----------------------------------------------------------------------
stdout:
=== RUN   TestEverything
    TestEverything: main_test.go:104: 
        	Error Trace:	main_test.go:104
        	            				main_test.go:61
        	Error:      	Not equal: 
        	            	expected: "FAILED"
        	            	actual  : "FINISHED"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-FAILED
        	            	+FINISHED
        	Test:       	TestEverything
        	Messages:   	Unexpected compose result: FAILED
--- FAIL: TestEverything (68.00s)
=== RUN   TestSources
--- PASS: TestSources (1.23s)
FAIL
```

ondrejbudai:
Can you point me at a build where osbuild-tests failed on RHEL 8.2? I cannot find one.

major:
@ondrejbudai Here's one: https://jenkins.osbuild.org/job/osbuild-composer-fedora32-base/125/

ondrejbudai:
This is Fedora though.

major:
Hmm, I'll keep looking to see if i can find another one. It seems to be intermittent. Sometimes I can tell Jenkins to re-run the test and the issue disappears on the second run.

ondrejbudai:
I think this happens only on Fedora because of bad mirrors, but I'm not sure though. I'm planning to add more logging to the `osbuild-tests` in the case of failure.

atodorov:
This is the same as https://github.com/osbuild/osbuild-composer/issues/611 but we don't have the logs from the compose to see what happened.

@major is this still an issue you are seeing or can this be closed ?

ondrejbudai:
I'm not convinced that this happens on RHEL 8.2. @major can you find a build where this happens? If not, let's change the title to Fedora, this happens often there because of flaky mirrors.

See
https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-621/17/pipeline/14
[...]

PR #616, opened 2020-05-14
tests: swap the places of expected-actual results in asserts

atodorov:
This results in better output. Output before this swap can be
seen at https://github.com/osbuild/osbuild-composer/issues/611.
It looks like we wanted a compose to FAIL and it FINISHED instead
which is making the error log confusing.


major:
Looks like #566 showed up again on this one.

PR #617, opened 2020-05-14
Fixing the logs once again with some refactorings

ondrejbudai:
The fix for the logs is here! This grew larger than I thought...

First 11 commits are just a minor refactoring. Basically, the `worker.JobStatus()` method got renamed to `worker.GetJobStatus()` and it now returns also the `JobResult`. In `weldr`, `getComposeState()` is now called `getComposeStatus()` and returns also `JobResult`. `worker.JobResult()` method was dropped because `GetJobStatus()` can be used instead.

Let's explain more in detail why I merged `JobResult()` and `GetJobStatus()` together:

Both methods actually called `jobqueue.GetJobStatus()`, so performance-wise there's no real difference and it feels to me that it makes the code simpler: You don't have to decide which method to call, you just get all the data about a job in one call. We could split those again when we see some perf issues with retrieving logs on each status check but I don't think we should optimize prematurely. Let's leave some work for the future us.

After this was done, the fix was rather simple - make the logs routes use the updated `getComposeStatus()` method which returns logs in both `compose has a job` and `compose doesn't have a job` situations. In the latter case, this means returning empty log, but hey, we don't have that many users and if someone really wants them, they are still in `/var`.

The second to last commit adds a quick smoke test to see that logs are created.

Finally, the last commit drops all the code managing the `result.json` file because no one is actually using that code now.

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -142,16 +142,24 @@ func (api *API) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 	api.router.ServeHTTP(writer, request)
 }
 
+type composeStatus struct {
+	State    common.ComposeState
+	Queued   time.Time
+	Started  time.Time
+	Finished time.Time
+}
+
 // Returns the state of the image in `compose` and the times the job was
 // queued, started, and finished. Assumes that there's only one image in the
-// compose. Returns CWaiting on error.
-func (api *API) getComposeStatus(compose store.Compose) (state common.ComposeState, queued, started, finished time.Time) {
+// compose.
+func (api *API) getComposeStatus(compose store.Compose) *composeStatus {
 	jobId := compose.ImageBuild.JobID
 
 	// backwards compatibility: composes that were around before splitting
 	// the job queue from the store still contain their valid status and
 	// times. Return those here as a fallback.
 	if jobId == uuid.Nil {
+		var state common.ComposeState
 		switch compose.ImageBuild.QueueStatus {
 		case common.IBWaiting:
 			state = common.CWaiting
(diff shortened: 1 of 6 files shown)


ondrejbudai:
Oh, I didn't know about this convention. I kinda like when methods are starting with a verb but getters could be an exception. I'm fine with reverting, but I'm not sure how to name the new `JobStatus` struct. Any suggestions?

larskarlitski:
> Oh, I didn't know about this convention. I kinda like when methods are starting with a verb but getters could be an exception. I'm fine with reverting, but I'm not sure how to name the new `JobStatus` struct. Any suggestions?

Keep it as `JobStatus`. It's a different namespace and I don't think it is confusing.

For example, the `Stringer` [interface](https://golang.org/pkg/fmt/#Stringer) is defined as:
```go
type Stringer interface {
    String() string
}
``` 
A method called `String` returning a `string`.

ondrejbudai:
Oh, you're right, they are indeed in different namespaces. Force-pushed a new version without the renaming of getters.

internal/weldr/api.go
```diff
@@ -148,7 +148,7 @@ func (api *API) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 // compose. Returns CWaiting on error.
 func (api *API) getComposeState(compose compose.Compose) (state common.ComposeState, queued, started, finished time.Time) {
 	if len(compose.ImageBuilds) == 0 {
-		return
+		panic("no image build in compose")
```
teg, line 159: I think this becomes redundant after I merged PR #594.

ondrejbudai:
Rebased.

> Hopefully that won't be too messy.

It was fairly simple, better than I thought.

PR #618, opened 2020-05-14
CI: Use Jenkins' merged repo for testing

major:
When Jenkins clones a repository for testing, it does the base clone
first and then merges the code from the PR afterwards. This ensures that
the code merges properly and is tested properly, but it also makes a SHA
that only exists inside Jenkins. 😢

Tell ansible-osbuild to use the repository that Jenkins made so that the
SHA is valid.

### DIFF
--- a/schutzbot/run_tests.sh
+++ b/schutzbot/run_tests.sh
@@ -28,9 +28,14 @@ export ANSIBLE_CONFIG=ansible-osbuild/ansible.cfg
 OSBUILD_COMPOSER_VERSION=$(git rev-parse HEAD)
 
 # Deploy osbuild-composer and osbuild using RPMs built in a mock chroot.
+# NOTE(mhayden): Jenkins clones the repository and then merges the code from
+# the pull request into the repo. This creates a new SHA that exists only in
+# Jenkins. We use ${WORKSPACE} below to tell ansible-osbuild to use the clone
+# that Jenkins made for testing osbuild-composer.
 git clone https://github.com/osbuild/ansible-osbuild.git ansible-osbuild
 ansible-playbook \
   -i hosts.ini \
+  -e osbuild_composer_repo_url=${WORKSPACE} \
   -e osbuild_composer_version=${OSBUILD_COMPOSER_VERSION} \
   -e install_source=mock \
   ansible-osbuild/playbook.yml


major:
@ondrejbudai The testing for this one has already passed the git clone step, so we should be all set to merge. I'll let you decide if you want to wait for Jenkins to fully finish. 😜

major:
Hmm #611 showed up again (unrelated).

teg:
This logic is a bit odd, isn't it? The sha being tested is not the one from the PR, and not the one that will be the new master after the merge (as other things may be merged in between). Is this something Jenkins makes us do? Either way: are we sure this is the behavior we want?

major:
@teg The idea here is that any developer could base their PR on whatever version of the codebase they want to. Jenkins takes the code they wrote and rebases it on master, then runs tests. Jenkins is essentially making a copy of what master would look like if the PR was already merged.

teg:
I understand. Before continuing this debate for too long, do we have a say in the matter her, or is this enforced on us by jenkins?

I'm not sure I am convinced that this is the best way to test the code (also not sure it is not the best way...).

I agree that the problem attempting to be solved by this logic is real. However, by not going all the way there I wonder if we add confusion but without actually solving anything. Testing just the PR as it was pushed has the downside that by the time it is merged, the resulting code will be different. It has the upside that if the CI finds an issue, then it is clear that the issue was in the pushed code, and the developer can easily reproduce, rather than the issue being due to some merge problem. So to make debugging manageable, I would have preferred to run CI on the PR as it was pushed.

Ideally, we would also have the possibility of running CI on what will become master after the PR is merged, to avoid any bugs slipping in (debugging this would need some extra output for the developer to reproduce exactly what merge to do locally to get the same code, but that would be ok). My understanding is that Jenkin's approach goes in this direction. However, it is not responsible for doing the actual merging, so if two PRs are open at the same time (seems likely) only one of them will have been tested correctly the way it will land on master. So I'm wondering if this is worth it?

teg:
@dvdhrm @larskarlitski this sounds like the kind of thing you have thoughts about.

major:
@teg This is easy to turn on or off. This is a default setting for Jenkins as it wants to ensure that the code that it is testing will work fine when it is merged into master.

larskarlitski:
Tricky, because GitHub's workflow is a bit broken, as @teg mentioned. I think both solutions have advantages and problems that weigh about equally. Thus, I'm fine either way.

I don't think the "debugging" argument is particularly strong. People need to be aware that this problem exists anyway and check the actually tested commit in the log. Also, they ought to rebase often.

The real solution is to gate merging onto master with the tests, but that's another story :)

PR #619, opened 2020-05-14
CI: Gather log artifacts

major:
Ensure that logs are artifacted after a test run. Also, ensure that
these logs are in unique directories so that they are not overwritten by
Jenkins as it builds the final artifact archive.

For example, if two jobs have a `test.log`, Jenkis will only keep one of
those logs. Each log must be named differently or placed in a unique
directory.

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -13,31 +13,75 @@ pipeline {
                     agent { label "fedora31" }
                     environment { TEST_TYPE = "base" }
                     steps {
-                        sh "schutzbot/run_tests.sh"
+                        run_tests()
+                    }
+                    post {
+                        always {
+                            preserve_logs('fedora31-base')
+                        }
                     }
                 }
                 stage('Fedora 31 image') {
                     agent { label "fedora31" }
                     environment { TEST_TYPE = "image" }
                     steps {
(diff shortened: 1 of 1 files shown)


major:
Artifacts! https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-619/1/artifacts

schutzbot/Jenkinsfile
```diff
@@ -13,31 +13,49 @@ pipeline {
...
         }
     }
 }
+
+// Set up a function to hold the steps needed to run the tests so we don't
+// need to copy/paste the same lines over and over above.
+void run_tests(test_slug) {
+
+    // Run the tests from the repository.
+    sh "schutzbot/run_tests.sh"
```
larskarlitski, line 50: Does this function continue executing if this line fails?

major: @larskarlitski https://www.jenkins.io/doc/book/pipeline/syntax/#post 🤦‍♂️

Good catch. I'll update the PR.

larskarlitski: Heh, I figured it's most important to see the logs when the test fails. Thanks for this!

PR #623, opened 2020-05-15
news: reformat as proper markdown

dvdhrm:
We use 8ch indents right now, which causes everything to be interpreted
as `<pre>` (or whatever your favourite equivalent is). Lets use proper
markdown.

While at it, this fixes a set of minor issues:

  * Surround markdown syntax with `` to suppress interpretation (like in
    the regexp, etc).

  * Replace hyphens before dates with proper em-dash.

  * Fix indentation in some news-entries, which did not align correctly.

Also see the [rendered](https://github.com/dvdhrm/osbuild-composer/blob/wip/formatnews/NEWS.md) version.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,196 +2,196 @@
 
 ## CHANGES WITH 12:
 
-        * In previous versions support for running remote workers was
-         broken. This is now fixed and running remote workers is once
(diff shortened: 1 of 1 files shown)


larskarlitski:
Restarted the travis job: `ndb` error.

PR #627, opened 2020-05-15
worker: Check OSBuildOutput for nil before using it

bcl:
It is possible for it to end up as nil so it needs to be checked
everywhere it is used.

### DIFF
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -101,7 +101,7 @@ func (s *Server) JobStatus(id uuid.UUID) (*JobStatus, error) {
 	if canceled {
 		state = common.CFailed
 	} else if !finished.IsZero() {
-		if result.OSBuildOutput.Success {
+		if result.OSBuildOutput != nil && result.OSBuildOutput.Success {
 			state = common.CFinished
 		} else {
 			state = common.CFailed


bcl:
> Where did you see this happening?

I hit this while running osbuild-tests against freshly built and installed rpms on a Fedora 32 host. The j.Result from readJob (in JobStatus) was a "{}" so it got unmarshaled into an empty struct with no pointer set.

Maybe the actual problem is in readJob? At the same time I'm a bit concerned by all the pointer usage in this code, maybe some of it is unnecessary and can be removed eventually to prevent these kinds of problems.

larskarlitski:
Hm, I cannot reproduce this, sorry. I've read the code carefully again and couldn't find anything obviously wrong, either. Sorry :(

teg:
Hm, we sholud figure out the underlying cause of this I think. @bcl, could you help us reproduce it? Which commit did you see the bug with? Or was it in CI somewhere we can see the logs?

bcl:
I hit it on my local F32 VM when I switched to building rpms (trying to reproduce problems with #578) so I'll see if I can trace it back up the stack. But I still think we have a general problem with pointer usage in the codebase. This kind of error isn't caught by the compiler at all, which is unfortunate.

teg:
I agree that avoiding pointers (or optional things in general) if we can, is nicer. However, having checks in that only check for our own bugs I don't think makes sense. So we should figure out what is actually going on.

bcl:
I tracked it down in more detail:

* I have a number of job that probably crashed, their files in `/var/lib/osbuild-composer/jobs/` have an empty Result object `{}` and are in the `WAITING` state. While this is probably the side-effect of crashes during testing and development I think it's a valid case, users will eventually hit something similar.
* The OSBuildJobResult.OSBuildOutput is a pointer, and has the json tag 'omitempty' so being nil is completely valid.
* JobStatus uses OSBuildJobResult as Result, which makes the presence of a potential nil pointer difficult to know unless you look deeper.
* In JobStatus it uses OSBuildOutput without a nil check and panics.

I'm not sure how else all of this could be structured, all of this code is undocumented, and I'm unfamiliar with the overall design of it. So yeah this patch is a band-aid. It looks like JobStatus assumes that if finished isn't zero then OSBuildOutput is safe to use, but that's not the case, finished is set to a time in the json that is triggering this.

Here's one of the job files:

https://bcl.fedorapeople.org/osbuild-composer/d6bc7a5d-9b5f-4773-ab74-1afeb9bdd66c.json

bcl:
getComposeStatus also suffers from this, it calls JobStatus and returns OSBuildOutput to the callers, who don't check to see if it is nil. I'm seeing this panic in composeLogHandler when it calls  composeStatus.Result.Write(writer)

The compiler doesn't catch the potential problem here at all (which seems like a bug? Or something linting ought to be able to pick up on) so Write ends up actually being called, but the struct it is working with is nil!

larskarlitski:
Sorry, I forgot following up on this. Was reminded by @ondrejbudai fixing it from the other side: https://github.com/osbuild/osbuild-composer/pull/767#pullrequestreview-429535773

We should be robust in the server as well, though. I force-pushed to rebase on master and fix the conflict. Thanks!

teg:
I reviewed the CI failures, and they are unrelated. This patch is also included in #762, which pass CI.

PR #628, opened 2020-05-16
codevoc: fix threshold

larskarlitski:
d49f0fef4 added a 5% threshold to codecov.yml, so that small fixes don't
turn the CI red.

However, it contained a typo and put the setting in the wrong place in
the file. codecov.io ignores unknown keys instead of throwing an error.

Fix this and validate with:

    cat codecov.yml | curl --data-binary @- https://codecov.io/validate

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,9 +1,9 @@
 coverage:
-  project:
-    default:
-      treshold: 5%
   status:
     patch: no
+    project:
+      default:
+        threshold: 5%
 codecov:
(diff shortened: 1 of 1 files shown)


larskarlitski:
I don't understand codecov: how can this PR generate `+2.68%` more coverage?

Schutzbot is running into https://github.com/osbuild/osbuild-composer/issues/566 again. This PR only touches `codecov.yml`, so it's fine to merge.

Issue #632, opened 2020-05-18
To remove the AWS timeout or not to remove the AWS timeout, that is the question


major:
With the changes in #621, we now wait 45 minutes for the AWS import process to finish. There's a 60 minute timeout on the AWS image job as a whole (which leaves extra time for the AMI to be built before uploading).

Some have suggested removing the timeout entirely to improve the customer experience. After all, if a customer wants their image in EC2, they expect it to arrive there, even if it takes a while. A timeout and failure would be fairly frustrating.

Should we remove the timeout for AWS image imports?

major:
If I [understand this code properly](https://github.com/aws/aws-sdk-go/blob/master/aws/request/waiter.go#L171-L192), the `attempt` counter starts at `1` and the code checks to see if the `attempt` number matches the value of `MaxAttempts`:

```
	for attempt := 1; ; attempt++ {
		req, err := w.NewRequest(w.RequestOptions)
		if err != nil {
			waiterLogf(w.Logger, "unable to create request %v", err)
			return err
		}
		req.Handlers.Build.PushBack(MakeAddToUserAgentFreeFormHandler("Waiter"))
		err = req.Send()

		// See if any of the acceptors match the request's response, or error
		for _, a := range w.Acceptors {
			if matched, matchErr := a.match(w.Name, w.Logger, req, err); matched {
				return matchErr
			}
		}

		// The Waiter should only check the resource state MaxAttempts times
		// This is here instead of in the for loop above to prevent delaying
		// unnecessary when the waiter will not retry.
		if attempt == w.MaxAttempts {
			break
		}
```

We should be able to set `MaxAttempts` to `0` and it should keep going forever.

PR #633, opened 2020-05-18
Remove AWS snapshot import timeout

major:
Importing an image into an EBS snapshot can sometimes take a long time.
The customer experience for a timeout is not great.

Allow the import to continue running for as long as needed without
timing out. This allows a customer to monitor the process of their
import until it has a success or failure status.

Fixes osbuild/osbuild-composer#632.


major:
> I wholeheartedly approve :) Thanks for fixing this, and sorry for introducing this in the first place...

Haha, don't thank me until we verify that it works! 🤣

larskarlitski:
It works, but it's running into an unrelated timeout: https://github.com/osbuild/osbuild-composer/blob/master/schutzbot/vars.yml#L17

Merging this despite this failure, because this PR only touches AWS upload.

PR #635, opened 2020-05-18
osbuild: add support for marshalling Fedora IoT manifests

teg:
This has not yet been hooked up from the distro package, so is still unused.

### DIFF
--- a/internal/osbuild/ostree_commit_assembler.go
+++ b/internal/osbuild/ostree_commit_assembler.go
@@ -1,0 +1,22 @@
+package osbuild
+
+// OSTreeCommitAssemblerOptions desrcibe how to assemble a tree into an OSTree commit.
+type OSTreeCommitAssemblerOptions struct {
+	Ref string                          `json:"ref"`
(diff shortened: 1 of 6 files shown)


internal/osbuild/ostree_commit_assembler.go
```diff
@@ -0,0 +1,22 @@
+package osbuild
+
+// OSTreeCommitAssemblerOptions desrcibe how to assemble a tree into an OSTree commit.
+type OSTreeCommitAssemblerOptions struct {
+	Ref string                          `json:"ref"`
+	Tar OSTreeCommitAssemblerTarOptions `json:"tar"`
```
larskarlitski, line 6: The final version that landed in osbuild-15 also has `tmp-is-dir` and `parent` options. Not blocking, of course.

teg: Yeah, we only support the subset we need (in general). Happy to add if they are needed, of course.

teg:
This has been superseded by #656.

PR #639, opened 2020-05-19
test/dnf-json: use the fullpath to dnf-json

ondrejbudai:
osbuild-dnf-json-tests requires dnf-json to be available in the current directory. This confused many people in the past. This commit adds a check if dnf-json is present in the current directory. If not, it prints a helpful message and fails the test immediately.

Another solution is to change the cwd in the test but that would make running the test locally a bit harder, so I chose a more conservative approach. Happy to dicuss though.

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -87,7 +87,9 @@ func main() {
 		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 
-	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"))
+	// osbuild-composer must be run in /usr/libexec/osbuild-composer directory,
+	// therefore use ./dnf-json as the path to dnf-json.
+	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"), "./dnf-json")
 
 	distros, err := distro.NewRegistry(fedora31.New(), fedora32.New(), rhel8.New())
 	if err != nil {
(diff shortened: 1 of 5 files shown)


larskarlitski:
I'm not sure I like this approach, because we might forget adding such a check when we use `rpmmd` again somewhere else.

Could we just always run `dnf-json` from where it is installed, in `/usr/libexec/dnf-json`? We're hard-coding `/usr` paths in other places as well (albeit in `cmd/`, not `internal/`).

ondrejbudai:
Not sure either. But I like that this doesn't break anyone's workflow, yet it gives you a meaningful error when you run it from a wrong directory.

larskarlitski:
Which workflow would we be breaking?

Alternatively, we could pass the `libexec` path to `NewRPMMD`, and default to the current working directory.

ondrejbudai:
> Which workflow would we be breaking?

Someone might want to run the test from the git checkout but that's probably a niche use-case.

> Alternatively, we could pass the `libexec` path to `NewRPMMD`, and default to the current working directory.

I like this! I made `NewRPMMD` take the path to `dnf-json` and used the fullpath to `dnf-json` in the test.

larskarlitski:
> I like this! I made `NewRPMMD` take the path to `dnf-json` and used the fullpath to `dnf-json` in the test.

Awesome, thanks! I'm ok with this as is, but would like to get @teg's opinion too.

cmd/osbuild-composer/main.go
```diff
@@ -90,7 +90,9 @@ func main() {
 		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 
-	rpm := rpmmd.NewRPMMD(path.Join(cacheDirectory, "rpmmd"))
+	// osbuild-composer must be run in /usr/libexec/osbuild-composer directory,
```
teg, line 93: I like this approach a lot. But why not pass the full path here too?

ondrejbudai:
superseded by #746

PR #645, opened 2020-05-19
CI: Run EC2 tests for all instances

major:
Now that EC2 image testing is working well for Fedora 32, enable it for
all of the distros. Also, make it less special and allow it to run with
the other image tests.

Remove the `ami-boot` test from the list of regular image test cases
since it will be thoroughly tested via an import into AWS.


major:
Of course `dnf-json` would start misbehaving now along with EC2 importing so slowly... 🤦‍♂️

PR #647, opened 2020-05-19
distribution: split weldr and job sockets

### LINKED ISSUE #646: composer-cli requires root on RHEL 8 but not on Fedora 31/32
While working on osbuild/ansible-osbuild#58, I found that running `composer-cli` with osbuild-composer on RHEL 8.2 requires `sudo`, but Fedora 31 and 32 allow the normal `fedora` user to use `composer-cli` to talk to osbuild-composer.

I'm not sure if this is by design, or if we have a bug.

larskarlitski:
This fixes permission for the weldr socket: allow users in group `weldr`
to access it, but nobody else. Also add this group to the sysusers file.

Keep the job socket accessible by everyone, as it was before. It will
probably be dropped at some point in favor of remote workers.

Fixes #646

### DIFF
--- a/distribution/osbuild-composer.socket
+++ b/distribution/osbuild-composer.socket
@@ -4,6 +4,8 @@ Description=OSBuild Composer API sockets
 [Socket]
 ListenStream=/run/weldr/api.socket
 ListenStream=/run/osbuild-composer/job.socket
+SocketGroup=weldr
+SocketMode=660
 
 [Install]
 WantedBy=sockets.target
--- a/distribution/osbuild-composer.conf
+++ b/distribution/osbuild-composer.conf
@@ -1,1 +1,2 @@
 u   _osbuild-composer   -   "OSBuild Composer user"
+g   weldr               -


cmd/osbuild-composer/main.go
```diff
@@ -57,6 +59,19 @@ func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
 	}, nil
 }
 
+func firstListener(listeners map[string][]net.Listener, name string) (net.Listener, error) {
```
bcl, line 62: Why is this named firstListener? Seems like getListener would be clearer.

bcl:
It's also not clear to me what's going on with the addition of the job socket. Is that related to the permission change somehow? Or does it belong in a separate commit?

cmd/osbuild-composer/main.go
larskarlitski: It returns the first listener. The `activation` API is a bit cumbersome in that regard. systemd socket units can have multiple listeners, and this API returns all of them by name. This function is meant to only return the first of them,

larskarlitski:
> It's also not clear to me what's going on with the addition of the job socket. Is that related to the permission change somehow? Or does it belong in a separate commit?

I needed to split it, because the weldr socket has different permissions. The `SocketGroup=` and `SocketMode=` options apply to all listeners in a socket unit.

distribution/osbuild-composer.socket
```diff
@@ -1,9 +1,10 @@
 [Unit]
-Description=OSBuild Composer API sockets
+Description=OSBuild Composer socket for the weldr API
 
 [Socket]
 ListenStream=/run/weldr/api.socket
-ListenStream=/run/osbuild-composer/job.socket
+SocketGroup=weldr
+SocketMode=660
```
teg, line 7: Colud we just add these two directives without splitting the socket into two? That would make the rest of the patch redundant (I think). We should still add the user to sysusers though. I think making sure the job socket is not world readable is also the right thing to do. The worker anyway runs as root.

larskarlitski: Convinced. This makes the patch much simpler. I was concerned about exposing the job socket with the `weldr` group as well, but it doesn't really matter.

PR #652, opened 2020-05-20
distro/rhel-8: merge all the RHEL-8 minor releases into one

teg:
Seems there were no objections to this. Details in the commit.


major:
This looks like a decent idea to me. I'm curious to see how the tests go.

PR #653, opened 2020-05-20
GH Actions: Use upstream Fedora containers

major:
Work around the quay.io issues by using the standard Fedora containers.
Also, make most of the dnf operations a little quieter to make it easier
to find problems.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,24 +57,24 @@ jobs:
       matrix:
         fedora_release: ["31"]
     container:
-      image: "quay.io/osbuild/osbuild-fedora${{ matrix.fedora_release }}:latest"
+      image: "docker.io/library/fedora:${{ matrix.fedora_release }}"
     steps:
       - name: "📥 Prepare container"
         run: |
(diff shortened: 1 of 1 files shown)


major:
Essentially the same change as osbuild/osbuild#411.

teg:
What's the difference between the containers? Why did we have our own in quay to begin with? (I seem to remember there was a reason)

teg:
Wasn't the idea to have pre created containers with all dependencies installed so they would be stable between runs?

ondrejbudai:
Hmm, good point... Let's wait for @major

major:
Well, to be honest, now that we're doing mock builds in Jenkins, we probably don't need these jobs any longer. Should we just remove them?

teg:
> Well, to be honest, now that we're doing mock builds in Jenkins, we probably don't need these jobs any longer. Should we just remove them?

Hehe, even better! Let's!

major:
> Hehe, even better! Let's!

https://tenor.com/view/kevin-hart-its-about-to-go-down-calm-before-the-storm-gif-4420127

PR #654, opened 2020-05-20
CI: Skip AWS test if bucket missing

major:
The osbuild CI uses the osbuild-composer CI testing playbooks but
there's no need to run the EC2 import for osbuild's CI. Add a check in
the Ansible playbook to look for the AWS_BUCKET environment variable and
skip the EC2 import if the bucker variable is missing.

### DIFF
--- a/schutzbot/test.yml
+++ b/schutzbot/test.yml
@@ -36,6 +36,8 @@
         loop_var: test_case
       when:
         - test_type == 'image'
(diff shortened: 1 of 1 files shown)


atodorov:
> The osbuild CI uses the osbuild-composer CI testing playbooks 

grrr. Can you change that env to a different name to indicate in which CI environment this is running ? E.g. CI=osbuild vs CI=osbuild-composer ?

PR #656, opened 2020-05-21
Add Fedora IoT support

teg:
This adds support for generating Fedora IoT commit tarballs for Fedora 32. See individual commits for details.

I had a short discussion with @jgiardino about this, and she suggested that we could handle the required ostree-specific options in the same way as we handle image size, for now. That is, in the cockpit-composer UI, the user would can specify the parent commit and the ostree ref in the same place they select the image type (where the image size is currently configured).

Does that make sense to you @jkozol?

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -75,8 +75,8 @@ func main() {
 		panic(err)
 	}
 
-	distro := distros.GetDistro(composeRequest.Distro)
-	if distro == nil {
+	d := distros.GetDistro(composeRequest.Distro)
+	if d == nil {
 		_, _ = fmt.Fprintf(os.Stderr, "The provided distribution '%s' is not supported. Use one of these:\n", composeRequest.Distro)
 		for _, d := range distros.List() {
 			_, _ = fmt.Fprintln(os.Stderr, " *", d)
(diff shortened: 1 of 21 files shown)


jkozol:
> That is, in the cockpit-composer UI, the user would can specify the parent commit and the ostree ref in the same place they select the image type

@teg this makes sense to me. I am a little unsure where the `parent commit` and `ostree ref` are derived from. Is the `parent commit` a commit in an ostree that the `ostree ref` points at and then composer will commit on top of the `parent commit`?

teg:
> I am a little unsure where the `parent commit` and `ostree ref` are derived from. Is the `parent commit` a commit in an ostree that the `ostree ref` points at and then composer will commit on top of the `parent commit`?

Yes, and no. I think @gicmo can explain this the best. But essentially, how it works is this:

We will produce a tarball, containing a ostree repository, containing exactly one commit.

This commit can then be pulled (by others) into an existing repository, and in that case, we need to declare in the commit we make, what the parent commit sha is, so this is all stitched together correctly.

Similarly, the ref is the branch our commit would be pulled onto.

None of this is actually done by composer, but we need to include the information in the produced tarball so the right thing happens when someone does the pull.

jgiardino:
Here is a rough mockup of what Tom describes. You can also see this example in codesandbox: https://codesandbox.io/s/image-parent-commit-joww9?file=/index.js

Some comments and questions I have:

1. What's the best label to use for `ostree ref`? I used "Branch" for now based on Tom's explanation, but I'm not sure if this is the best label.
2. These additional fields should include help text, which is not currently shown in the mockup. Here's an initial draft: 
    - Parent commit - Provide the SHA of the previous commit. This new commit will provide the update for the commit that you specify as the is the parent commit.
        - Does anyone have an example of a SHA that we can include with the help text?
    - Branch - 🤷 
3. We currently display image size for all image types, but does that make sense for this one?

image

teg:
> Here is a rough mockup of what Tom describes.

Thanks Jenn!

> 3. We currently display image size for all image types, but does that make sense for this one?

It does not. It also makes no sense for the current Tar image type.

teg:
I updated the PR to rename the image type internally to "fedora-iot-commit" from "iot-commit". This is ready for review/merge, as far as I'm concerned.

teg:
Ah, and let's continue the UI discussion of course, either here or in a "cockpit-composer" PR/issue.

PR #657, opened 2020-05-21
CI: Add mock builds for RHEL 8.3

major:
Now that mock building seems to be stable, let's start building RPMs for
RHEL 8.3 so we can onboard it for base/image testing soon.


major:
The 8.3 mock build completed just fine, but I stopped the pipeline after that to save PSI resources since we're not deploying to 8.3 quite yet.

PR #642, opened 2020-05-19
tests: split compose sanity tests and add more. Refs #356, #615

atodorov:

Issue #524, opened 2020-04-22
Osbuild-composer-10-1.fc31 fails to build any image


jrusz:
Any image type I try to build fails and no logs can be found anywhere
Steps:

1. Provision a fresh Fedora-31 system
1. `dnf install golang-github-osbuild-composer`
1. Update using this https://bodhi.fedoraproject.org/updates/FEDORA-2020-2720b8a2fb
1. `systemctl start osbuild-composer`
1. Upload a very basic BP `composer-cli blueprints push base.toml` (shown bellow)
1. `composer-cli compose start base ext4-filesystem`

Build fails, no logs are present in `cockpit-compsoer` or in journal or in `/var/lib/osbuild-composer/outputs`. I tried it with several more image types, all seem to fail, some of them just give a bit more info in journal.

Blueprint:
```
name = "base"
description = "A base system with bash v2"
version = "0.0.1"
modules = []
groups = []

[[packages]]
name = "bash"
version = "*"
```
Journal from building ext4-filesystem:
```
Apr 22 10:30:27 kvm-02-guest08.rhts.eng.brq.redhat.com osbuild-worker[17426]: Running job 9bae6fa1-0d27-47b5-a93f-c8f86adbccc4
Apr 22 10:30:27 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: Created slice Virtual Machine and Container Slice.
Apr 22 10:30:27 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: Started Container osbuild-buildroot-ti8oc0ox.
Apr 22 10:30:58 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: osbuild-buildroot-ti8oc0ox.scope: Succeeded.
Apr 22 10:30:58 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: osbuild-buildroot-ti8oc0ox.scope: Consumed 10.873s CPU time.
Apr 22 10:30:58 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: Started Container osbuild-buildroot-mfh2jxa.
Apr 22 10:31:07 kvm-02-guest08.rhts.eng.brq.redhat.com systemd[1]: osbuild-buildroot-mfh2jxa.scope: Succeeded.
[...]

PR #658, opened 2020-05-21
Reproducer for 524

atodorov:
builds on top of #642 and tries to add a quick reproducer for #524. We only need the last commit from here.

### DIFF
--- a/cmd/osbuild-tests/main_test.go
+++ b/cmd/osbuild-tests/main_test.go
@@ -29,7 +29,8 @@ func TestComposeCommands(t *testing.T) {
 
 	bp := blueprint.Blueprint{
 		Name:        "empty",
-		Description: "Test empty blueprint in toml format",
+		Description: "Test blueprint in toml format",
+		Packages: []blueprint.Package{{Name: "bash", Version: "*"}},
 	}
 	pushBlueprint(t, &bp)
 	defer deleteBlueprint(t, &bp)


cmd/osbuild-tests/main_test.go
```diff
@@ -21,24 +21,72 @@ import (
...
 
+func TestComposeCommands(t *testing.T) {
+	// common setup
+	tmpdir := NewTemporaryWorkDir(t, "osbuild-tests-")
+	defer tmpdir.Close(t)
+
+	bp := blueprint.Blueprint{
+		Name:        "empty",
+		Description: "Test empty blueprint in toml format",
+		Packages: []Package{{Name: "bash", Version: "*"}},
```
jrusz, line 32: I think this needs to be `Packages: []blueprint.Package{{Name: "bash", Version: "*"}},`

jrusz:
I can confirm this reproducer works on my F31 vm after the little change I pointed out. 
```
Output:
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
Verifying packages...
Preparing packages...
        file /usr/bin/bashbug conflicts between attempted installs of bash-5.0.11-1.fc31.i686 and bash-5.0.11-1.fc31.x86_64
imported gpg key
imported gpg key
imported gpg key
imported gpg key
fetching sources
creating a fake machine id
  File "/run/osbuild/lib/stages/org.osbuild.rpm", line 121, in <module>
    r = main(args["tree"], args["sources"], args["options"])
  File "/run/osbuild/lib/stages/org.osbuild.rpm", line 103, in main
    ], cwd=f"{sources}/org.osbuild.files", check=True)
  File "/usr/lib64/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['rpm', '--verbose', '--root', '/run/osbuild/tree', '--define', '_pkgverify_level all', '--install', '/tmp/manifest.c7e8uz2i']' returned non-zero exit status 254.


--- FAIL: TestComposeCommands (78.43s)
    main_test.go:139: 
                Error Trace:    main_test.go:139
                                                        main_test.go:50
                Error:          Not equal: 
                                expected: "FINISHED"
                                actual  : "FAILED"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -FINISHED
                                +FAILED
                Test:           TestComposeCommands
                Messages:       Unexpected compose result: FAILED
FAIL
```

teg:
Seems to happen on F32 too, but not RHEL8, right?

Could this be down to a behavioural change in DNF? We should anyway improve on dnf-json to figure out how to avoid this.

atodorov:
@teg ping

teg:
I looked through this again. I believe tests for this was added in #747, if you agree, we could close this I think. Otherwise, the description of the test blueprint is now not quite right.

atodorov:
> I looked through this again. I believe tests for this was added in #747, if you agree, we could close this I think. Otherwise, the description of the test blueprint is now not quite right.

@teg you are right that there was another reproducer/test that was merged together with the bug fix bug I still want to keep the current one too. It is on the integration test level and is a very minimal change anyway.

PR #659, opened 2020-05-22
Remove RPM builds in GitHub Workflows

major:
We are now building RPMs via mock in Jenkins, so we don't need these RPM
build jobs in GitHub Workflows.


major:
I stopped the Jenkins build to save on resources since it's not touched by this change.

Issue #644, opened 2020-05-19
tests: composer-cli compose results fails


atodorov:
```
# composer-cli compose list
56621ea8-de30-41d2-bf78-32aa3cde0669 FINISHED test-blueprint 0.0.2 ami
# composer-cli compose results 56621ea8-de30-41d2-bf78-32aa3cde0669
No Content-Disposition header; cannot get filename
# echo $?
1
```

This causes one of the new sanity tests introduced in #642 to fail.

bcl:
This is the same kind of thing as #643, in lorax-composer results is the metadata + the image file. Not sure if that makes sense for osbuild-composer.

atodorov:
Same comment for reference: https://github.com/osbuild/osbuild-composer/issues/643#issuecomment-631400176

teg:
Same comment from me too ;) I think we should add support.

PR #661, opened 2020-05-22
Add support for /compose/metadata, /compose/results, as well as tests

### LINKED ISSUE #643: tests: composer-cli compose metadata fails
```
# composer-cli compose list
0396b0be-410b-44b6-97dd-e317f2b17c52 FINISHED test-blueprint 0.0.2 AWS

# composer-cli --json compose metadata 0396b0be-410b-44b6-97dd-e317f2b17c52
No Content-Disposition header; cannot get filename
```
exit code 1.

This causes the newly enabled sanity tests in #642 to fail.

bcl:
This fixes #643 and #644

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -112,6 +112,7 @@ func New(rpmmd rpmmd.RPMMD, arch distro.Arch, distro distro.Distro, repos []rpmm
 	api.router.GET("/api/v:version/compose/finished", api.composeFinishedHandler)
 	api.router.GET("/api/v:version/compose/failed", api.composeFailedHandler)
 	api.router.GET("/api/v:version/compose/image/:uuid", api.composeImageHandler)
+	api.router.GET("/api/v:version/compose/metadata/:uuid", api.composeMetadataHandler)
 	api.router.GET("/api/v:version/compose/logs/:uuid", api.composeLogsHandler)
 	api.router.GET("/api/v:version/compose/log/:uuid", api.composeLogHandler)
 	api.router.POST("/api/v:version/compose/uploads/schedule/:uuid", api.uploadsScheduleHandler)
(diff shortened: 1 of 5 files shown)


bcl:
Rebased on master, new vmdk failure seems unrelated.

internal/weldr/api.go
```diff
@@ -2025,6 +2026,67 @@ func (api *API) composeImageHandler(writer http.ResponseWriter, request *http.Re
...
+	if composeStatus.State != common.CFinished && composeStatus.State != common.CFailed {
+		errors := responseError{
+			ID:  "BuildInWrongState",
+			Msg: fmt.Sprintf("Build %s is in wrong state: %s", uuidString, composeStatus.State.ToString()),
+		}
+		statusResponseError(writer, http.StatusBadRequest, errors)
+		return
+	}
+
+	metadata, err := json.Marshal(compose)
```
teg, line 2067: I think we should just expose the manifest here (the `Compose` is really internal state). So I would use `compose.ImageBuild.Manifest` here instead.

internal/worker/server.go
```diff
@@ -97,7 +97,7 @@ func (s *Server) JobStatus(id uuid.UUID) (*JobStatus, error) {
 	if canceled {
 		state = common.CFailed
 	} else if !finished.IsZero() {
-		if result.OSBuildOutput.Success {
+		if result.OSBuildOutput != nil && result.OSBuildOutput.Success {
```
teg, line 100: Hm, as discussed on the other PR, this seems to be masking some underlying problem. I'd have loved to figure out the underlying issue there was. Is this a blocker for the rest of this PR? (My concern would be that even if this avoids us from crashing, we are still in an undefined/inconsistent state that we should probably have caught/handled elsewhere).

internal/weldr/api.go
```diff
@@ -2087,6 +2088,98 @@ func (api *API) composeMetadataHandler(writer http.ResponseWriter, request *http
...
```
teg, line 2128: Same comment as on the other patch.

internal/worker/server.go
bcl: Depending on the state of the system it fails. On a clean install I think it is fine -- but in my opinion if there's a pointer involved there is always going to be the potential for it to be nil and that should be checked, or the pointer should be eliminated.
I can drop this from this PR and we can cover it in the other one, I added it here so I could run the tests in my VM.

internal/weldr/api.go
bcl: Ok, added that, but also a check for nil that returns a 400 with an 'EmptyManifest' error.

internal/worker/server.go
teg: If it doesn't block this, then let's continue the discussion on the other PR

teg:
On closer inspection, the only tests that fail are the RCM tests, which were known to be broken and have been fixed on master. They are also going to be dropped and are independent of this PR. Merging.

PR #669, opened 2020-05-26
worker: dont cache just yet

larskarlitski:
When fdd7536 added `--output-directory` to the invocation of osbuild,
it also removed `--store`.

This was a mistake: osbuild's default store is `.osbuild`, which is not
what we want. Restore the old behavior of passing a temporary directory.


larskarlitski:
Schutzbot reports some image boot failures on RHEL 8.2. I don't think these can be related to this PR. All other tests are green.

PR #670, opened 2020-05-26
Add health check at /status

major:
There are times where it would be good to monitor that osbuild-composer
is up and running. Add a very simple status check that always returns
`200 OK`. This can be expanded later to verify that other parts of
osbuild-composer are working properly.

### DIFF
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -49,6 +49,10 @@ func NewServer(logger *log.Logger, jobs jobqueue.JobQueue, artifactsDir string) 
 	s.router.MethodNotAllowed = http.HandlerFunc(methodNotAllowedHandler)
 	s.router.NotFound = http.HandlerFunc(notFoundHandler)
 
+	// Add a basic status handler for checking if osbuild-composer is alive.
+	s.router.GET("/status", s.statusHandler)
+
+	// Add handlers for managing jobs.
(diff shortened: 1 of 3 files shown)


internal/worker/server.go
```diff
@@ -134,6 +138,15 @@ func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
 	jsonErrorf(writer, http.StatusNotFound, "not found")
 }
 
+func (s *Server) addStatusHandler(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
```
ondrejbudai, line 141: `statusHandler` is probably enough, the method doesn't add anything. :-)

major: 👍

internal/worker/server.go
```diff
@@ -134,6 +138,15 @@ func notFoundHandler(writer http.ResponseWriter, request *http.Request) {
 	jsonErrorf(writer, http.StatusNotFound, "not found")
 }
 
+func (s *Server) StatusHandler(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
```
ondrejbudai, line 141:
```suggestion
func (s *Server) statusHandler(writer http.ResponseWriter, request *http.Request, _ httprouter.Params) {
```

Please, can you change it to a private method for consistency?

PR #674, opened 2020-05-26
Document additional RHEL image types

larskarlitski:
Together with https://github.com/osbuild/osbuild-composer/pull/671, all RHEL image types have a documented purpose.


larskarlitski:
This only changes documentation, so the test failures must be unrelated.

PR #675, opened 2020-05-27
Pre release cleanups

ondrejbudai:
Minor bits changed, see the individual commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,17 +233,17 @@ release:
 	@echo "--------------------------------------------------------------------------------"
 	@echo "## CHANGES WITH $(NEXT_VERSION):"
 	@echo
-	@echo "        * ..."
+	@echo "  * ..."
 	@echo
-	@echo "        * ..."
+	@echo "  * ..."
 	@echo
-	@echo -n "        Contributions from: "
+	@echo -n "Contributions from: "
 # We omit the contributor list if `git log` fails. If you hit this,
 # consider fetching missing tags via `git fetch --tags`, or just copy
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -233,17 +233,17 @@ release:
...
 	@echo
-	@echo -n "        Contributions from: "
+	@echo -n "Contributions from: "
 # We omit the contributor list if `git log` fails. If you hit this,
 # consider fetching missing tags via `git fetch --tags`, or just copy
 # this command and remove the stderr-redirect.
 	@echo `( git log --format='%an, ' v$(VERSION)..HEAD 2>/dev/null | sort -u | tr -d '\n' | sed 's/, $$//' ) || echo`
 	@echo
-	@echo "        - Location, YYYY-MM-DD"
+	@echo "- Location, YYYY-MM-DD"
```
dvdhrm, line 246: I would use `—` rather than `-`, otherwise you end up with a single-item list. Your release-PR has this copied as well.

ondrejbudai: Fixed in both cases. Thanks for pointing out!

golang-github-osbuild-composer.spec
```diff
@@ -40,7 +40,7 @@ BuildRequires:  golang(github.com/stretchr/testify)
 
 Requires: golang-github-osbuild-composer-worker
 Requires: systemd
-Requires: osbuild >= 12
```
teg, line 43: This should probably be 15-{2,3} to pick up the patch that @gicmo added. But we could do that in a follow-up.

PR #677, opened 2020-05-27
Disable automatic local worker startup

major:
Now that we deploy remote workers by default, there's no need for an
automatically-started local worker that listens on a socket. This also
alleviates the need to mask the `osbuild-worker@.service` unit when you
deploy remote workers.


teg:
I would love to only have one type of worker.

My only concern with this is how to make it "just work" out of the box for someone just installing cockpit-composer and interacting with it through the UI. In particular, we would need to provision SSL certificates for this to work, right?

major:
Right. Let's skip this for now, then.

PR #678, opened 2020-05-27
CI: Use RHEL 8 CDN subscribed instance

major:
Replace RHEL 8.2 nightly images with a RHEL 8 CDN subscribed instance so
we can test with the exact content a customer would have.


teg:
CI failed due to failing download for F32, which was untouched by this PR. Merging.

PR #679, opened 2020-05-27
Fix some test failures against lorax-composer

bcl:


bcl:
> Makes sense! Out of curiosity, would it make sense for osbuild to fail in the same way, or is the new way preferable?

I think osbuild is ok, if anything lorax-composer is wrong for including 2 errors for the same issue.

PR #683, opened 2020-05-28
spec: exclude i686 architecture

ondrejbudai:
RHEL 8 doesn't have golang in the i686 buildroot anymore. Fedora doesn't have
i686 images and kernel anymore. Also, osbuild-composer doesn't have support
for building i686.

Let's exclude i686 for these reasons, it's dead.

See rhbz#1752991


ondrejbudai:
quay is currently down, failing the koji test

ondrejbudai:
Koji test failed due to quay being down. Also, it's completely unrelated.

PR #686, opened 2020-05-29
store: don't log if no logger was passed

### LINKED ISSUE #685: composer crashes when store.New() wants to log in non-verbose mode
Logging behaviour got slightly changed in a22cd78e. In non-verbose mode, `nil` Logger is passed to `store.New()`. However, the store tries to use the logger anyway, resulting in crashes when the store is invalid or ignoring some composes.

cc @teg

teg:
`log` is `nil`, unless we are in verbose mode. Skip logging if no logger
was passed.

Long-term I think we should reconsider how we do our logging, and always
log these type of unexpected errors, and hence always have the logger be
non-nil. For now, fix the immediate problem.

The fact that loading the store from disk fails on upgrade is
unexpected, and should be fixed separately.

Fixes #685.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -75,7 +75,7 @@ func New(stateDir *string, arch distro.Arch, log *log.Logger) *Store {
 	if stateDir != nil {
 		db = jsondb.New(*stateDir, 0600)
 		_, err := db.Read(StoreDBName, &storeStruct)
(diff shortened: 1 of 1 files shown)


ondrejbudai:
meme

teg:
All checks passed except for a flaky mirror. Merging.

PR #696, opened 2020-05-31
store: make the upgrade path more solid

teg:
This continues the effort to make the store more resilient to upgrades.

This solves the issue that changing the supported subset of osbulid manifests would invalidate the store, for no reason. We do not need to read the manifests out for any other purpose than passing them on. Represent them now as opaque JSON-serializable objects.

A couple of simple test cases are added, that verify that we can still load stores created with v12 and v13 of osbuild (the two most recent ones).


teg:
Rebased on top of master to get the most recent AWS CI fix.

teg:
CI failed with a flaky mirror and with an AWS timeout, neither related to this PR.

Issue #693, opened 2020-05-29
ppc64le: osbuild-tests failure


major:
From RHEL 8 on ppc64le:

```
TASK [Run osbuild-tests test] **************************************************
Friday 29 May 2020  12:55:02 -0400 (0:00:00.670)       0:10:51.753 ************ 
fatal: [localhost]: FAILED! => changed=true 
  cmd:
  - /usr/libexec/tests/osbuild-composer/osbuild-tests
  - -test.v
  delta: '0:00:22.795913'
  end: '2020-05-29 12:55:25.418691'
  msg: non-zero return code
  rc: 1
  start: '2020-05-29 12:55:02.622778'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    === RUN   TestComposeCommands
    --- FAIL: TestComposeCommands (0.71s)
        main_test.go:257:
                    Error Trace:    main_test.go:257
                                                            main_test.go:40
                    Error:          Received unexpected error:
                                    exit status 1
                    Test:           TestComposeCommands
                    Messages:       Command failed: exit status 1
    === RUN   TestEverything
    --- PASS: TestEverything (21.18s)
    === RUN   TestSourcesCommands
    --- PASS: TestSourcesCommands (0.90s)
    FAIL
  stdout_lines: <omitted>
```

msehnout:
This wasn't finished as part of https://github.com/osbuild/osbuild-composer/issues/353 and the last related PR was https://github.com/osbuild/osbuild-composer/pull/501. I'll revisit this issue now.

msehnout:
https://github.com/osbuild/osbuild-composer/pull/714 should fix this issue. Can I trigger the test to make sure it is fixed? @major

PR #699, opened 2020-06-02
osbuild: change legacy type from bool to string

msehnout:
This work is based on previous PRs, namely:
https://github.com/osbuild/osbuild-composer/pull/501
and
https://github.com/osbuild/osbuild/pull/327

The problem here is that we used to treat legacy as a boolean before we
started introducing support for alternative architectures, but now we
need to specify exact strings for the grub2 stage, for example for
ppc64le the legacy parameter looks like this:

```
"legacy": "powerpc-ieee1275"
```

This patch will allow us to introduce support for ppc64le and fix
associated issues:
https://github.com/osbuild/osbuild-composer/issues/693

### DIFF
--- a/internal/distro/fedora31/distro.go
+++ b/internal/distro/fedora31/distro.go
@@ -44,6 +44,7 @@ type arch struct {
 	name               string
 	bootloaderPackages []string
 	buildPackages      []string
+	legacy             string
 	uefi               bool
 	imageTypes         map[string]imageType
 }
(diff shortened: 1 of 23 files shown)


msehnout:
Schutzbot failed on different stages this time. Last time it was only Fedora 31 image, this time it is Fedora 31 and 32 base. Should we bypass the check, @teg ?

teg:
F31 base tests failure is due to a flaky mirror. That i think we can safely ignore. The other two failures were both due to boot failing in AWS, but it also does work on one, so I think we can safely assume this to be unrelated to this PR as well. I'd have preferred not to circumvent the tests, and hopefully we'll eventually get to a place where we don't have to.

PR #701, opened 2020-06-02
Let AWS figure out the image format during import

major:
Specifying the image format in the `DiskContainers` part of the snapshot
import process sometimes causes AWS to fail the import due to formatting
issues. However, AWS seems to do a decent job at determining the image
format on its own without specifying the format explicitly.


ondrejbudai:
Thanks @major for figuring this out. :champagne: This is really weird.

PR #707, opened 2020-06-03
Add logging for AWS upload tasks

major:
The AWS upload process takes a while to complete and can sometimes have
errors. By adding additional logging, users can see what has already
happened and they will have data available, such as snapshot import task
IDs, that they can use to check on status at AWS.

### DIFF
--- a/internal/upload/awsupload/awsupload.go
+++ b/internal/upload/awsupload/awsupload.go
@@ -1,6 +1,7 @@
 package awsupload
 
 import (
+	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


major:
Darn, I can't get logs to show up in the system journal with `log.Printf` or `fmt.Printf`. 😢

ondrejbudai:
Schutzbot fail is unrelated, this just adds logging.

(if it is related, there's something really fishy in the world of Go and AWS)

Issue #698, opened 2020-06-01
cloud-init installed, but services not started


major:
While working on ansible-osbuild's smoke tests in osbuild/ansible-osbuild#72, I found a problem with the `qcow2` and `openstack` image outputs. They have `cloud-init` installed, but none of the services/targets are enabled.

I used the following blueprint:

```toml
name = "bash"
description = "A base system with bash"
version = "0.0.1"

[[packages]]
name = "bash"
```

Then I started the build with `composer-cli --json compose start bash qcow2`. Fedora 31/32's cloud-init package does not enable cloud-init's services/targets by default and both images failed to prepare the user I asked for in my cloud-init config. RHEL 8 was working fine, but its cloud-init package has [some hacks](https://src.osci.redhat.com/rpms/cloud-init/blob/rhel-8.2.1/f/cloud-init.spec#_134) to ensure services are started after installation.

When I used the following blueprint, all distros worked:

```toml
name = "bash"
description = "A base system with bash"
version = "0.0.1"

[[packages]]
name = "bash"

[customizations.services]
enabled = ["sshd", "cloud-init", "cloud-init-local", "cloud-config", "cloud-final"]
```

As a user, I would expect cloud-init to always be running on bootup whether I asked for it to be started explicitly or not.

msehnout:
For OpenStack image, this sounds reasonable to me ([docs](https://docs.openstack.org/image-guide/openstack-images.html)) but for generic qcow2 I'm unsure about the requirements. Of course, image that you can use sounds like a reasonable requirement but it is still possible to use it without cloud-init for example by setting SSH key in the blueprint.

larskarlitski:
For reference, [Fedora Cloud Base](https://alt.fedoraproject.org/cloud/) has cloud-init installed and enabled.

bcl:
I wonder why qcow2 has it installed? qcow2 should be a generic VM container, the user should decide if they want cloud-init installed and enabled. OTOH openstack should have it enabled by default.

The problem with enabling it by default in plain qcow2 is that it then has to be disabled in order to boot without it setup on your network.

msehnout:
> For reference, [Fedora Cloud Base](https://alt.fedoraproject.org/cloud/) has cloud-init installed and enabled.

The description suggests it is intended for OpenStack, but as @bcl suggested qcow2 can also be used as a regular VM. I think this should be one of the differences between `qcow2` and `openstack` image types.

larskarlitski:
[...]

PR #710, opened 2020-06-03
Run AWS test using osbuild worker code

major:
The `osbuild-image-tests` code does a good job testing image builds for
AWS, but it doesn't use the same code paths that a customer would when
using `composer-cli`.

Add a test that builds a compose and uploads the image using
osbuild-composer and its worker. Also, build an instance in AWS based on
the image we imported and check to see if the smoke test file is
present.


major:
Benefits from this PR:

* We test the AWS image like a customer would end-to-end
* We get console screenshots of the instance as it boots (helps with diagnosing issues)
* Can watch the logs as the compose uploads/imports into AWS

test/image-tests/aws.sh
```diff
@@ -0,0 +1,256 @@
...
+tee $BLUEPRINT_FILE > /dev/null << EOF
+name = "bash"
+description = "A base system with bash"
+version = "0.0.1"
+
+[[packages]]
+name = "bash"
+
+[customizations.services]
+enabled = ["sshd", "cloud-init", "cloud-init-local", "cloud-config", "cloud-final"]
```
teg, line 104: I don't mind putting something into the test blueprint, but these things should be equivalent to an empty blueprint, no?

major: That's a workaround for #698

test/image-tests/aws.sh
```diff
@@ -0,0 +1,256 @@
+#!/bin/bash
+set -euo pipefail
+
+source /etc/os-release
```
teg, line 4: I like this very much! I think we should make it more consistent with the other tests, shipping it as just another test binary, but all that we can discuss in the future, let's merge it as is.

major: Agreed. I tried my best to make it a test that someone could just run from another machine without needing Jenkins.

test/image-tests/aws.sh
```diff
@@ -0,0 +1,256 @@
...
+
+# Fedora 31's composer-cli doesn't support doing image uploads.
+if [[ $ID == fedora ]] && [[ $VERSION_ID == 31 ]]; then
```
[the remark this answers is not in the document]
major: Yeah, we can. We're not running any of those now, so I left it out. It's super easy to add later. 😜

test/image-tests/aws.sh
```diff
@@ -0,0 +1,256 @@
...
+
+[[packages]]
+name = "bash"
+
+[customizations.services]
+enabled = ["sshd", "cloud-init", "cloud-init-local", "cloud-config", "cloud-final"]
+EOF
+
+# Copy the internal Fedora repository into place.
+if curl -fs http://download.devel.redhat.com > /dev/null; then
```
teg, line 108: Should this be specific to only this test?

major: We could move it into a shared `function.sh` at some point when I add the next test for qcow. It could easily be used for other tests.

PR #713, opened 2020-06-04
osbuild: introduce bootloader struct into qemu assembler

msehnout:
osbuild takes a "bootloader" object as an option to the qemu assembler:
https://github.com/osbuild/osbuild/blob/3f14ace5c1ac876555aab6e05aa6f0f8dbbe278b/assemblers/org.osbuild.qemu#L43
we don't use it because for x86_64 with enabled legacy support it
defaults to the right value:
https://github.com/osbuild/osbuild/blob/3f14ace5c1ac876555aab6e05aa6f0f8dbbe278b/assemblers/org.osbuild.qemu#L482
but in order to gain support for ppc64le we need to introduce this.
Example usage can be found in samples directory:
https://github.com/osbuild/osbuild/blob/3f14ace5c1ac876555aab6e05aa6f0f8dbbe278b/samples/f30-ppc64le.json#L819

This change itself does not alter osbuild-composer output.

### DIFF
--- a/internal/osbuild/qemu_assembler.go
+++ b/internal/osbuild/qemu_assembler.go
@@ -9,6 +9,7 @@ import "github.com/google/uuid"
 // containing the indicated partitions. Finally, the image is converted into
 // the target format and stored with the given filename.
 type QEMUAssemblerOptions struct {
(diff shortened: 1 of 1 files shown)


ondrejbudai:
The usual mirror stuff failing.

PR #714, opened 2020-06-04
rhel8 - ppc64le support

msehnout:
Still WIP, I just want to trigger CI, but you can find more info in commit messages if you are interested.


teg:
The exact same commit was merged as #728.

PR #722, opened 2020-06-05
osbuild: update qemu assembler to match osbuild schema better

msehnout:
see commit messages for more info

### DIFF
--- a/internal/distro/fedora31/distro.go
+++ b/internal/distro/fedora31/distro.go
@@ -635,7 +635,6 @@ func (r *imageType) selinuxStageOptions() *osbuild.SELinuxStageOptions {
 func qemuAssembler(format string, filename string, uefi bool, size uint64) *osbuild.Assembler {
 	var options osbuild.QEMUAssemblerOptions
 	if uefi {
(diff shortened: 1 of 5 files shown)


internal/distro/fedora31/distro.go
```diff
@@ -646,7 +645,7 @@ func qemuAssembler(format string, filename string, uefi bool, size uint64) *osbu
 				{
 					Start: 2048,
 					Size:  972800,
-					Type:  &fstype,
```
teg, line 649: Will this pass the tests? It is lower-case and in the same for f32 it is upper-case?

msehnout: Actually unit tests were unhappy when it was the same in f31 and f32. It was either this change or fix in test definition.

msehnout:
Sorry, I had to force-push so that I can rebase #714 on top of this but since we are waiting for CI anyway I hope it does not matter. The patches are still exactly the same.

teg:
This PR does not affect the image tests, as the taste cases are unchanged. The base tests pass on f32 and rhel8, for f31 the error has to do with the move to S3, it appears.

At any rate, this code only affects unit tests, which pass. Merging.

PR #723, opened 2020-06-05
tools: fix generate-test-case script

msehnout:
Commit 8dd45544919961bf6d5d285a7fb6724ec0404e0f introduced a bug because
the output from osbuild is not captured. The problem is that osbuild
itself writes JSON document to STDOUT so two different JSON documents
are written to STDOUT: osbuild output and generated test case. The
generate-test-cases script then fails because it cannot json.loads() the
test case.

This patch simply supresses the output which fixes the issue.

### DIFF
--- a/tools/test-case-generators/generate-test-case
+++ b/tools/test-case-generators/generate-test-case
@@ -41,9 +41,10 @@ def run_osbuild(manifest, store, output):
                     "--store", store,
                     "--output-directory", output,
                     "--json", "-"],
-                   check=True,
-                   encoding="utf-8",
-                   input=json.dumps(manifest))
+                    stdout=subprocess.DEVNULL,
+                    check=True,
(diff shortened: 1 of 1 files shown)


tools/test-case-generators/generate-test-case
```diff
@@ -41,9 +41,10 @@ def run_osbuild(manifest, store, output):
                     "--store", store,
                     "--output-directory", output,
                     "--json", "-"],
-                   check=True,
-                   encoding="utf-8",
-                   input=json.dumps(manifest))
+                    capture_output=True,
```
ondrejbudai, line 44: we don't use the captured output, do we? I would prefer `stdout=subprocess.DEVNULL` then.

msehnout: Sure, I'll change it :)

PR #724, opened 2020-06-05
osbuild: introduce zipl stage

msehnout:
zipl is a z initial program loader used with IBM Z systems. This stage
is required to get support for the s390x architecture.

### DIFF
--- a/internal/osbuild/zipl_stage.go
+++ b/internal/osbuild/zipl_stage.go
@@ -1,0 +1,25 @@
+package osbuild
+
+// The ZiplStageOptions describe how to create zipl stage
+//
+// The only configuration option available is a boot timeout and it is optional
+type ZiplStageOptions struct {
+	Timeout int `json:"timeout,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


internal/osbuild/zipl_stage.go
```diff
@@ -0,0 +1,25 @@
...
+// The ZiplStageOptions describe how to create zipl stage
+//
+// The only configuration option available is a boot timeout and it is optional
+type ZiplStageOptions struct {
+	Timeout int `json:"timeout,omitempty"`
+}
+
+func (ZiplStageOptions) isStageOptions() {}
+
+// NewZiplStageOptions creates a new ZiplStageOptions object with no timeout
```
teg, line 12: I don't think this constructor is necessary, mostly we instantiate the structs directly. But that's a nitpick.

teg:
The test failure on f32 is related to transitioning to storing artifacts in S3, and unrelated to this PR. The tests pass on all other distros, and this functionality is currently only affects unittests. Merging.

PR #729, opened 2020-06-06
tests: use more reliable mirrors for Fedora tests

teg:
Move away from the round-robin fedora mirrors, and use the most reliable
one we have found (kernel.org).


teg:
Test failures are clearly unrelated, as the sha's of the downloaded RPMs are unchanged.

PR #662, opened 2020-05-25
Rename funcs in sanity tests

atodorov:

PR #730, opened 2020-06-07
Rename functions in sanity test

teg:
This is based on #662, but rebased to pick up fixes. Added one trivial patch on top.


teg:
The only failing test is AMI boot test. This is an ongoing problem that this is not stable, but this PR does not change any manifsets, nor the uploading, so it must be unrelated.

PR #734, opened 2020-06-08
Rhel8 s390x support

msehnout:
Ready for review.


internal/distro/rhel8/distro.go
```diff
@@ -544,6 +555,28 @@ func qemuAssembler(format string, filename string, uefi bool, imageOptions distr
...
+				Size:     imageOptions.Size,
+				PTUUID:   "0x14fc63d2",
+				PTType:   "dos",
+				Partitions: []osbuild.QEMUPartition{
+					{
+						Start:    2048,
+						Bootable: true,
+						Filesystem: &osbuild.QEMUFilesystem{
+							Type:       "xfs",
+							UUID:       "76a22bf4-f153-4541-b6c7-0332c0dfaeac",
```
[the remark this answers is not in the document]
msehnout: Thanks! I changed all the UUID to be the same as for the other arches but I'm still in the emergency shell.

internal/distro/rhel8/distro.go
```diff
@@ -906,6 +939,9 @@ func New() distro.Distro {
 		distro: &r,
 		name:   "s390x",
```
gicmo, line 940: s390 seems to be missing `dracut-config-generic` which will lead to a host-only initramfs which does not contain the drivers for virtual disks, making it impossible to boot in qemu:

```
bootloaderPackages: []string{
   "dracut-config-generic",
   "s390utils-base"
}
```
I think  "s390utils-base", which contains `zipl`, is already being pull in, but maybe better be explicit about it.

test/cases/rhel_8-s390x-tar-boot.json
```diff
@@ -2196,6 +2203,10 @@
             }
           }
         },
+        {
+          "name": "org.osbuild.zipl",
```
msehnout, line 2207: @gicmo , @teg : Do we want the zipl stage in `tar` image type or not?

teg: The tar output type is under-specified, so for now I think we should just leave it as is. I would like to revisit this, and get proper specifications.

internal/distro/rhel8/distro.go
teg: This was now added.

teg:
CI failures are unrelated, and down to issues downloading from Fedora mirrors. The unit tests pass, meaning the generated manifests are unchanged, so this is good to go.

PR #738, opened 2020-06-08
Add native qcow2 boot test

major:
Add an end-to-end qcow2 test that follows a customer's steps with
`composer-cli`. The image is booted with libvirt to allow the best
virtualization options to be chosen by libvirt. It also uses libvirt's
default network.

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -92,7 +92,7 @@ pipeline {
                     environment { TEST_TYPE = "base" }
                     steps {
                         unstash 'fedora31'
-                        run_tests()
+                        run_tests('base')
                     }
                     post {
                         always {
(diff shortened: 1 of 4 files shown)


major:
Two benefits here:

* Our qcow2/openstack tests now use `composer-cli` like a customer would.
* We can run these tests on systems without nested virt, like ec2.

teg:
Hm, by dropping the other tests we are no longer doing the image-info tests and this is doing the tests against update repos which are changing under us, right?
I think it is very nice test to have, but could we do it in addition? And maybe not make it blocking? @larskarlitski

major:
@teg Sure, we can set this up however we want. We can run these entirely separately too, if we want.

larskarlitski:
>  I think it is very nice test to have, but could we do it in addition? And maybe not make it blocking? @larskarlitski

Agreed.

major:
Thanks, @teg!

PR #739, opened 2020-06-08
Image info: support for analyzing ostree commits

gicmo:
See the individual commits for details, here an excerpt from the main one:

Support analyzing ostree repositories or directory that contain an ostree commit as created by osbuild. Will return the mode of the repository along with the references and their commits. For the first references, the commit is resolved and checked-out to a temp directory. This directory in turn is then analyzed via the existing append_filesystem function. The latter has gained some small ostree specific tweaks.
NB: for the ostree checkout /usr/etc/ is bind mounted to /etc in order to make append_filesystem happy. The rpm verification step is NOT run, because that is not really compatible with ostree.


teg:
`image-info` tests all pass.

PR #742, opened 2020-06-08
test/image: use tempary osbuild store and improve logging

larskarlitski:


cmd/osbuild-image-tests/main_test.go
```diff
@@ -351,17 +352,20 @@ func testImage(t *testing.T, testcase testcaseStruct, imagePath string) {
 // runTestcase builds the pipeline specified in the testcase and then it
 // tests the result
 func runTestcase(t *testing.T, testcase testcaseStruct) {
-	outputDirectory, err := ioutil.TempDir("/var/tmp", "osbuild-image-tests-*")
-	require.NoErrorf(t, err, "cannot create temporary output directory: %#v", err)
+	tmpdir, err := ioutil.TempDir("/var/tmp", "osbuild-image-tests-*")
```
teg, line 355: This will slow down test runs, could we use a fixed store?

cmd/osbuild-image-tests/main_test.go
```diff
@@ -395,16 +395,16 @@ func getAllCases() ([]string, error) {
 
 // runTests opens, parses and runs all the specified testcases
 func runTests(t *testing.T, cases []string) {
-	for _, path := range cases {
-		t.Run(path, func(t *testing.T) {
-			f, err := os.Open(path)
+	for _, p := range cases {
```
teg, line 398: Are there no users of this in the tests that need to be adjusted?

ondrejbudai: IIRC I don't think so.

cmd/osbuild-image-tests/main_test.go
larskarlitski: Good point. Will fix.

larskarlitski:
Update to also pass fix `constants-travis.go` and to use the same store for all test runs.

PR #746, opened 2020-06-09
Fix dnf-json on RHEL & clean ups

ondrejbudai:
Formerly #639

We had two issues with `dnf-json`:

- `dnf-json-tests` must have bee run with `cwd` set to `/usr/libexec/osbuild-composer`. This confused more people than I would like to admit.
- on RHEL, the "default" Python is in `/usr/libexec/platform-python`. RPMMD used python from `PATH` which might not be installed (and we don't depend on it).

This PR fixes this by:

- RPMMD takes the path to `dnf-json` from outside.
- The callers of RPMMD now specify the full-path to `dnf-json`.
- RPMMD no longer uses `python` from `PATH` to run `dnf-json`. Instead, we now rely on the shebang (that is correctly mangled to `/usr/libexec/platform-python` on RHEL).

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -87,7 +87,9 @@ func main() {
 		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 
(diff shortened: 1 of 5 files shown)


ondrejbudai:
The CI failure looks like misbehaving AWS, otherwise it sounds good to me, @teg?

PR #747, opened 2020-06-09
wip: fix multilib handling

ondrejbudai:
I just want to ask Schutzbot to create me a scratch repo. :-)

PR #749, opened 2020-06-10
test: add a test-case to prevent bad multilib depsolves

ondrejbudai:
Again testing CI, this will be merged with #747.

### DIFF
--- a/cmd/osbuild-tests/main_test.go
+++ b/cmd/osbuild-tests/main_test.go
@@ -9,6 +9,8 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"strings"
+	"testing"
 	"time"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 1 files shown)


cmd/osbuild-tests/main_test.go
```diff
@@ -161,6 +162,43 @@ gpgkey_urls = ["https://url/path/to/gpg-key"]
...
+			}
+
+			if !assert.Equalf(t, 1, jqCount, "jq is specified %d-times in the depsolve, should be there only once", jqCount) {
+				t.Logf("depsolve output:\n%s", output)
+				t.FailNow()
+			}
+		})
+	}
+}
+
```
atodorov, line 201: IMO this kind of test is more suitable for `blueprints_test.go`, for example a slight modification to `TestBlueprintDepsolveV0` will do the job.  `osbuild-tests` should be for sanity testing of the cli commands with minimum assertions.

PR #752, opened 2020-06-10
Convert base test runner to bash

major:
Simplify the base test runner and make it easier to see the output live.


ondrejbudai:
This change only affects base tests and they all passed.

PR #755, opened 2020-06-10
Deploy quickly in CI without Ansible

major:
Speed up the deployment during CI runs and avoid the need for lots of
extra packages to run Ansible. This gets us closer to running bare
images for each distro that better simulates a customer environment.


ondrejbudai:
This was merged as #758 and #761

PR #757, opened 2020-06-11
Support canceling a compose

larskarlitski:
This adds support for the cancel route. The way it is canceling is not the cleanest, but it is very robust.

Sorry @teg, I hadn't seen your earlier commits that added the cancel route to weldr before working on this.


teg:
> Sorry @teg, I hadn't seen your earlier commits that added the cancel route to weldr before working on this.

Good! They were buggy :)

Thanks so much for this!

larskarlitski:
> Good! They were buggy :)

Who says this one isn't? :stuck_out_tongue_closed_eyes:

teg:
> > Good! They were buggy :)
> 
> Who says this one isn't? stuck_out_tongue_closed_eyes

Linter has something to say on that matter ;)

cmd/osbuild-tests/main_test.go
```diff
@@ -81,10 +81,14 @@ func TestComposeCommands(t *testing.T) {
...
 	defer deleteCompose(t, uuid)
 
-	// https://github.com/osbuild/osbuild-composer/pull/180
-	// uuid = startCompose(t, "empty", "tar")
-	// time.Sleep(time.Second)
-	// runComposer(t, "compose", "cancel", uuid.String())
+	// workers ask the composer every 15 seconds if a compose was canceled.
+	// Use 20 seconds here to be sure this is hit.
+	uuid = startCompose(t, "empty", "tar")
+	defer deleteCompose(t, uuid)
```
atodorov, line 87: Why do you have to start another compose while a few lines above we've got one already started ?

larskarlitski: It had a comment saying that it tests reusing the cache. Not sure how it actually does that, but I thought adding one that is immediately canceled wouldn't do that.

ondrejbudai:
Tar is only available on RHEL, I pushed a commit which changes the build image type to qcow2.

larskarlitski:
> Tar is only available on RHEL, I pushed a commit which changes the build image type to qcow2.

Of course! Sorry for the oversight and thanks for fixing this :heart:

cmd/osbuild-tests/main_test.go
atodorov: IDK what the cache is but reordering the actions like currently seems to be better.

cmd/osbuild-tests/main_test.go
```diff
@@ -77,14 +77,18 @@ func TestComposeCommands(t *testing.T) {
...
 	defer os.Remove(uuid.String() + "-disk.qcow2")
 
+	// workers ask the composer every 15 seconds if a compose was canceled.
+	// Use 20 seconds here to be sure this is hit.
+	uuid = startCompose(t, "empty", "qcow2")
+	defer deleteCompose(t, uuid)
+	runComposer(t, "compose", "cancel", uuid.String())
+	time.Sleep(20 * time.Second)
+	status := waitForCompose(t, uuid)
+	assert.Equal(t, "FAILED", status)
```
atodorov, line 87: Isn't there a status CANCELLED to distinguish when the user cancelled the compose or is all FAILED ?

larskarlitski: Alas, there is not.

PR #759, opened 2020-06-11
image-tests: allow disabling of local booting

ondrejbudai:
On some environments (like RHEL gating) there's no virtualization available.
This commit adds -disable-local-boot argument to osbuild-image-tests. When
this argument is present, the local booting is skipped. This doesn't affect
the cloud booting, the test binary still tries to do that. If no credentials
are available, the fall back to local booting will be skipped if
-disable-local-boot is given.

I tested this successfully locally.

### DIFF
--- a/cmd/osbuild-image-tests/main_test.go
+++ b/cmd/osbuild-image-tests/main_test.go
@@ -40,6 +40,8 @@ type testcaseStruct struct {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ondrejbudai:
The failures are interesting but unrelated, image tests are still running local qemu boot.

PR #760, opened 2020-06-11
tests: composer-cli blueprints undo. Closes #615

### LINKED ISSUE #615: tests: uncomment/enable remaining sanity checks in cmd/osbuild-tests/main_test.go
I see commented out sections for some compose and blueprints commands which should be uncommented and implemented properly. The focus here is executing the commands and making sure they don't crash rather than validating the resulting state afterwards. Some asserts/validation is fine but if it needs to go too deep maybe it should be places in the API portion of the test suite.

atodorov:
the remaining test for composer-cli compose cancel is part of #759


ondrejbudai:
Flaky mirror, otherwise good.

PR #761, opened 2020-06-11
Deploy quickly in CI without Ansible

major:
Speed up the deployment during CI runs and avoid the need for lots of
extra packages to run Ansible. This gets us closer to running bare
images for each distro that better simulates a customer environment.


schutzbot/deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Add osbuild team ssh keys.
+cat schutzbot/team_ssh_keys.txt | tee -a ~/.ssh/authorized_keys > /dev/null
+
+# Set up a dnf repository for the RPMs we built via mock.
+sudo cp osbuild-mock.repo /etc/yum.repos.d/osbuild-mock.repo
+sudo dnf repository-packages osbuild-mock list
+
+# Install the Image Builder packages.
+sudo dnf -y install composer-cli osbuild osbuild-ostree \
```
teg, line 18: Shouldn't it be enough to pull in `osbuild-composer-tests`? That can be fixed in a follow-up.

major: I guess so, but I wanted to be explicit.

teg: Reason I suggest it is that it would help us catch missing `Requires` in our spec files.

PR #740, opened 2020-06-08
distro/rhel8: add RHEL for Edge

teg:
This replaces #727, I'm taking it back over again ;)

Rebased on top of master, otherwise the commit is unchanged.

PR #764, opened 2020-06-11
Pending features for 8.3

teg:
This collects #757, #763, and #740 for testing.


teg:
Only #757 is still pending, this PR is not needed.

PR #765, opened 2020-06-11
Add RHEL 8.3 repository override

major:
Override the repository for RHEL 8.3 since they can't be registered yet.


major:
Just fixed the `${VERSION_ID}` bug in #766. This PR is rebased onto that one so we can test out the new repository files.

ondrejbudai:
There was BaseOS repo twice and no Appstream. This failed the builds because we use some packages from AppStream. Pushed a fix on top of this.

PR #767, opened 2020-06-12
worker: ensure that the reported result is always non-nil

ondrejbudai:
When osbuild crashes (e.g. when cp fails because of the machine running
out of disk space), it doesn't produce a machine-readable result. Due to
our suboptimal handling of the result struct (this is my fault), this can
lead to result == nil. However, composer expects that result != nil in all
cases because it uses the Success flag to assess the compose state. If
result == nil, it just crashes terribly.

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -250,6 +250,15 @@ func main() {
 		// signal to WatchJob() that it can stop watching
 		cancel()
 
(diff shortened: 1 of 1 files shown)


ondrejbudai:
I agree, I think this PR should fix all the cases that can happen in a reasonable world. Better safe than sorry, though. :)

ondrejbudai:
(rebased, the cancel implementation caused a conflict)

PR #768, opened 2020-06-13
assorted fixes for RHEL 8.3 Beta

teg:
A bunch of fixes and tests for RHEL 8.3 Beta. See individual patches for details.


teg:
> LGTM, is this ready to be merged? It's currently set as a draft.

Just missed a test-case. I'm about to push a new version ready to be merged.

tools/test-case-generators/generate-test-cases
```diff
@@ -81,25 +100,104 @@ def main(distro, arch, image_types, keep_image_info, store, output):
...
+                                "time.example.com"
+                            ]
+                        },
+                        "locale": {
+                            "languages": [
+                                "en_US"
+                            ],
+                            "keyboard": "dvorak"
+                        },
+#                        "firewall": {
```
ondrejbudai, line 175: why is this commented out?

teg: The firewall stage requires firewalld to be installed in the image (sad, but also fine). Sadly, we cannot currently install the `firewalld` packake on all our image types: https://github.com/osbuild/osbuild-composer/issues/806

ondrejbudai: Ah, interesting, let's add this later then.

test/external-repos/rhel-8.json
```diff
@@ -0,0 +1,66 @@
...
+            "rhsm": true,
+            "check_gpg": true
+        }
+    ]
+/
```
ondrejbudai, line 66: This doesn't look like a valid JSON :stuck_out_tongue_closed_eyes:

ondrejbudai:
Usual suspects failing the tests, otherwise we're green.

PR #771, opened 2020-06-15
image-types: link the Azure requirements

msehnout:
This patch makes it easier to find the right document describing image
requirements for Linux VM images. It does not link the RHEL page
specifically because it is mainly for RHEL 6 and 7 whereas we build RHEL
8. It instead links documentation for generic Linux distribution.


teg:
This PR only touches documentation, CI failures are clearly unrelated.

PR #774, opened 2020-06-15
Print success log line after job is done

major:
We print messages to the log when the build fails, but we don't print
one when the build is successful.

I really want to celebrate our successes more often, so let's print a
success message when osbuild completes successfully. Since success
doesn't feel like success without an emoji, let's add one of those, too. 🎉


major:
Looks like some RPM downloads failed. 😥

PR #776, opened 2020-06-16
tests: upload & boot image in OpenStack. Closes #339

### LINKED ISSUE #339: tests/image: Test openstack images in OpenStack
We are already building this type of output images so need to start looking into actually uploading and booting them in OpenStack.

atodorov:


atodorov:
@teg I hit a bit of a snag here. The gophercloud API client doesn't seem to support block_device_mapping parameters when creating an OpenStack instance and as a result I get an instance which doesn't have any volumes attached to it. I suspect that's the reason it reports it's IP addresses as empty even though I can seen them in the webUI.

Any hints on how to proceed? I guess I can workaround this by manually creating a volume, attaching it to the instance and then trying to boot the instance. However I feel like the API should do this for us already.

major:
@atodorov With OpenStack, you are not required to specify a volume for an instance. They can just boot with ephemeral (local) storage on the hypervisor. Have you tried that option?

atodorov:
> @atodorov With OpenStack, you are not required to specify a volume for an instance. They can just boot with ephemeral (local) storage on the hypervisor. Have you tried that option?

Isn't not specifying anything in the API call the same thing ? 

Another reason for the failure I see could be the missing support for ssh keys. I will give that a try and see how it goes.

atodorov:
@major thanks for the hint. It turned out the VM was booting just fine but the way the code tried to get the IP address was wrong and it was getting an empty string. This is now ready for review - needs some more polishing.

cmd/osbuild-image-tests/openstacktest/openstack.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	defer imageData.Close()
+
+	err = imagedata.Upload(client, image.ID, imageData).ExtractErr()
+	if err != nil {
+		return image, fmt.Errorf("Upload to OpenStack failed: %v", err)
+	}
+
+	// wait for the status to change from Queued to Active
+	err = gophercloud.WaitFor(300, func() (bool, error) {
```
teg, line 47: I'd bump this timeout much higher or remove it completely. Usually 5 min should be fine, but for AWS we had issues where this would suddenly take forever and our tests failing spuriously. Our test framework anyway has a global timeout that would cancel the test-run eventually.

cmd/osbuild-image-tests/main_test.go
```diff
@@ -292,6 +295,50 @@ func testBootUsingAzure(t *testing.T, imagePath string) {
 	require.NoError(t, err)
 }
 
+func testBootUsingOpenStack(t *testing.T, imagePath string) {
+	// TODO: handle missing creds
```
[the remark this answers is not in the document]
atodorov: No, AuthOptionsFromEnv returns an error if any of the required variables are missing.

cmd/osbuild-image-tests/openstacktest/openstack.go
atodorov: This function needs its timeout argument, I'll see what I can do. How much higher is higher enough ? 

FTR: if this test fails because the operation didn't complete in 5 minutes I am perfectly fine with restarting the test job anyway. Same goes for any other environment because they can fail for whatever reason outside our control.

cmd/osbuild-image-tests/main_test.go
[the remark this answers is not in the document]
atodorov: Actually it probably is, but I don't have a good way to test it locally b/c I get 
"JSON Schema validation failed" error when passing the JSON file to the test command.

cmd/osbuild-image-tests/openstacktest/openstack.go
teg: I'm happy to merge as is, but I'd make it 6 hours or so, and then rely on the global timeout instead. If it is genuinely stuck, then restarting isn't a problem, but most of the time it just takes a bit longer than we expected, and if we have enough of these we see a lot of spurious CI failures which would have been avoided by just waiting longer.

teg:
The image test fails on F32, but that was due to a failed RPM download. The other image tests pass.

PR #783, opened 2020-06-17
Pretty print osbuild JSON on failure

### LINKED ISSUE #782: Image test failures dump raw osbuild JSON wall of text
When an image test fails, the osbuild output is dumped as a wall of text instead of pretty printed JSON. This makes it difficult to find out what went wrong.

major:
When osbuild fails, osbuild-composer's image tests print the output as
ugly JSON and that makes it difficult to determine why osbuild failed.

Instead, print the json in pretty format. 🌟

Fixes #782.

### DIFF
--- a/cmd/osbuild-image-tests/main_test.go
+++ b/cmd/osbuild-image-tests/main_test.go
@@ -63,7 +63,11 @@ func runOsbuild(manifest []byte, store, outputDirectory string) error {
 
 	err := cmd.Run()
 	if err != nil {
-		fmt.Println(outBuffer.String())
+		// Pretty print the osbuild error output.
+		buf := new(bytes.Buffer)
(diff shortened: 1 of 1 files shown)


major:
I'm still very new to golang, so if there's an easier/better way to do this, please let me know!

major:
@teg I still got a wall of text. I wonder if I need to parse the osbuild output as json first, and then marshal it?

major:
Aaah, I figured it out. I've gotta do something like this:

```golang
	buf := new(bytes.Buffer)
	json.Indent(buf, []byte(stdout), "", "    ")
	fmt.Println(buf)
```

major:
Ah, it worked! https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-783/7/pipeline/131

PR #786, opened 2020-06-18
distro/{rhel8, fedora32}: no fstab for OSTree

gicmo:
How exactly the final file system layout looks like is determined by the installer (anaconda) and thus can not be known at commit creation time. Thus creating an `/etc/fstab` file is unnecessary and the information in it probably wrong. The file wont be used though because it will be overwritten during the installation process.


teg:
CI failed due to an AWS boot failure on F32. Unrelated to this change.

PR #788, opened 2020-06-18
Use minimal images for mock build

major:
Now that we have CI images built by osbuild-composer, let's use them!


atodorov:
@major is this ready for review ?

PR #792, opened 2020-06-23
Shutzbot: install only osbuild-composer-tests

atodorov:
in order to catch missing dependencies

Note: I saw this as a comment somewhere in the recent pull requests.

Note2: not sure if related to the `execv()` error I'm seeing internally, more precisely why it doesn't happen in Schutzbot.


ondrejbudai:
This needs rebasing, otherwise looking good to me.

PR #793, opened 2020-06-23
CI: print the list of RPMs on the system

atodorov:
this will make it easier to figure out what was there before
installing SUT and help in debugging issues related to dependencies.


schutzbot/ci_details.sh
```diff
@@ -25,4 +25,7 @@ CI MACHINE SPECS
 
 ------------------------------------------------------------------------------
 EOF
-echo -e "\033[0m"
\ No newline at end of file
+echo -e "\033[0m"
+
+rpm -qa | sort
```
ondrejbudai, line 30: Can you add something like this?

```suggestion
echo List of installed packages:
rpm -qa | sort
```

PR #795, opened 2020-06-25
dnf-json: change confusing error message associated with dnf errors

msehnout:
The issue was introduced in 0d3c8329c069dbb8856c6d2fc15ff42a7ff517ea.
The patch correctly changed the base exception class, but it didn't
change the unfortunate use of hardcoded type name. This patch uses
Python's internal `__name__` attribute to get the type (exception) name.

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -107,7 +107,7 @@ with tempfile.TemporaryDirectory() as persistdir:
     try:
         base = create_base(repos, module_platform_id, persistdir, cachedir, arch)
     except dnf.exceptions.Error as e:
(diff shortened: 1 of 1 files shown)


teg:
This looks good to me. What was the confusing error messages though, could you include a before/after example?

msehnout:
> This looks good to me. What was the confusing error messages though, could you include a before/after example?

The confusion part was that we are catching `Error` like this:
```
except dnf.exceptions.Error as e:
```
But we are reporting `RepoError` because we used to catch:
```
except dnf.exceptions.RepoError as e:
```

Issue #781, opened 2020-06-17
Temporary dnf failures need to be retried


major:
During integration tests, a single dnf failure causes the entire depsolve/compose to fail:

```
[2020-06-17T17:27:00.015Z] 📋 Preparing blueprint
[2020-06-17T17:27:01.997Z] 2020-06-17 17:27:00,391: bash: DNF error occured: RepoError: Error occurred when setting up repo: Failed to download metadata for repo '0': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

We ought to retry during these situations.

msehnout:
This could be fixed with a patch like #796 but it looks odd and we are unsure osbuild-composer is the right place to implement the retry mechanism.

major:
Copied from #814:

We occasionally get a failure in CI when the `TestCrossArchDepsolve` test cannot download repo metadata due to temporary network issues or a mirror problem.

```
--- FAIL: TestCrossArchDepsolve (0.00s)
    --- FAIL: TestCrossArchDepsolve/fedora-31 (169.16s)
        --- FAIL: TestCrossArchDepsolve/fedora-31/aarch64 (72.28s)
            --- FAIL: TestCrossArchDepsolve/fedora-31/aarch64/ami (57.40s)
                main_test.go:85: 
                    	Error Trace:	main_test.go:85
                    	Error:      	Received unexpected error:
                    	            	DNF error occured: RepoError: Error occurred when setting up repo: Failed to download metadata for repo '1': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
                    	Test:       	TestCrossArchDepsolve/fedora-31/aarch64/ami
            --- PASS: TestCrossArchDepsolve/fedora-31/aarch64/openstack (7.60s)
            --- PASS: TestCrossArchDepsolve/fedora-31/aarch64/qcow2 (7.28s)
```

It would be nice if we could pre-cache the metadata locally or retry the test.

msehnout:
Tom mentioned this could be fixed by https://github.com/osbuild/osbuild-composer/pull/846 , but I cannot verify that idea.

major:
@msehnout I'm hoping that reduces the chances of dnf having issues.

Choosing the fastest mirror early ensures that the mirrors are actually up. I also removed the hardcoded internal Fedora mirror that was fairly flaky.

My fingers are crossed. 🤞🏻

teg:
[...]

PR #796, opened 2020-06-25
Dnf json retry

msehnout:
Fixes: #781
Requires: #795

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -107,7 +107,7 @@ with tempfile.TemporaryDirectory() as persistdir:
     try:
         base = create_base(repos, module_platform_id, persistdir, cachedir, arch)
     except dnf.exceptions.Error as e:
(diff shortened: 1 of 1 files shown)


teg:
Hm, if dnf has a retry mechanism wouldn't this be a dnf/mirror bug? We are not doing anything out-of-the-ordinary here, right?

If we do go down this path, could we be more specific about which errors we retry, and which we fail?

msehnout:
> Hm, if dnf has a retry mechanism wouldn't this be a dnf/mirror bug? We are not doing anything out-of-the-ordinary here, right?
> 
> If we do go down this path, could we be more specific about which errors we retry, and which we fail?

I was wondering the same (it's not our fault, it should be fixed in dnf in anywhere), but I wanted to at least try it like this. But it is not easy to trigger this specific error.

PR #797, opened 2020-06-25
tests: upload & test in vCenter. Closes #338

### LINKED ISSUE #338: tests/image: Test vmdk images on Vmware
I see that vmdk images can be built so we should start exploring the option of uploading them to vCenter and booting an actual VM from the result.  QE already has access to an appropriate environment to launch vmware VMs.

atodorov:
Very crude, still needs things to do. ATM only upload & delete images work - files are under `/` in the datastore.

I am having issues with creating the VM. There's https://github.com/vmware/govmomi/blob/master/govc/vm/create.go which is my source of inspiration. I was even thinking to mimic the cli by constructing the appropriate list of arguments and calling `create.Run()` from my code.  However I am having troubles executing this command from the terminal and I can't figure out the right set of arguments that it needs. ATM it is choking on the value of `-g` (guest OS type) and the docs are MIA, see
https://github.com/vmware/govmomi/issues/2034

@chris1984 do you have any contacts who could help here ?


atodorov:
@ondrejbudai this requires your work around stream optimized vmdk files and when producing images locally they weren't streamOptimized. Am I missing something ?

ondrejbudai:
They weren't? They should be... The `vmdk` image type creates a streamOptimized `vmdk`...

atodorov:
> They weren't? They should be... The `vmdk` image type creates a streamOptimized `vmdk`...

hmmm, I had a touch luck testing this locally (unrelated issues) but the image I got from `composer-cli compose start ... vmdk` was supposedly not streamOptimized b/c vCenter refused to import it. I had to manually convert it via `qemu-img`. I'll check again though.

atodorov:
@major in the test logs I see: 
`2020/08/10 10:26:58 No vCenter credentials given, falling back to booting using qemu`

can you check out if all of the credentials are fine ?

chris1984:
> @major in the test logs I see:
> `2020/08/10 10:26:58 No vCenter credentials given, falling back to booting using qemu`
> 
> can you check out if all of the credentials are fine ?

The Composer account should have full admin access to everything now.

cmd/osbuild-image-tests/vmwaretest/vmware.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// note: by default this will wait/block until an IP address is returned
+	// note: using exec() instead of running the command b/c .Run() returns an int
+	ipAddress, err := exec.Command(
+		"/usr/bin/govc",
+		"vm.ip",
+		fmt.Sprintf("-u=%s:%s@%s", creds.Username, creds.Password, creds.Host),
+		"-k=true",
+		imageName,
+	).Output()
```
atodorov, line 168: @teg this is the reason we need `/usr/bin/govc` as a dependency. I wasn't able to figure out how to use `cli.Run()` like I do everywhere else and get the output of the commands instead of an exit code.

teg (approved):
This is amazing, thanks for figuring all this out!

Issue #798, opened 2020-06-25
Image build fails with an SELinux denial


jeremycline:
On a Fedora 32 host with builds from the tip of osbuild my build fails with:

```
Stage: org.osbuild.selinux
{
  "file_contexts": "etc/selinux/targeted/contexts/files/file_contexts"
}

Output:
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
setfiles: Could not set context for /run/osbuild/tree/usr/bin/ejabberdctl:  Invalid argument
Traceback (most recent call last):
  File "/run/osbuild/lib/stages/org.osbuild.selinux", line 61, in <module>
    r = main(args["tree"], args["options"])
  File "/run/osbuild/lib/stages/org.osbuild.selinux", line 52, in main
    subprocess.run(["setfiles", "-F", "-r", f"{tree}", f"{file_contexts}", f"{tree}"], check=True)
  File "/usr/lib64/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['setfiles', '-F', '-r', '/run/osbuild/tree', '/run/osbuild/tree/etc/selinux/targeted/contexts/files/file_contexts', '/run/osbuild/tree']' returned non-zero exit status 255.
```

The root cause is:

```
AVC avc:  denied  { mac_admin } for  pid=110119 comm="setfiles" capability=33  scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:osbuild_t:s0 tclass=capability2 permissive=0
```

The blueprint, generated by adding ejabberd-19.09.1-2.fc32:
```
SDL2-2.0.12-1.fc32
abattis-cantarell-fonts-0.201-2.fc32
acl-2.2.53-5.fc32
adobe-source-code-pro-fonts-2.030.1.050-8.fc32
alternatives-1.11-6.fc32
at-spi2-atk-2.34.2-1.fc32
at-spi2-core-2.36.0-1.fc32
atk-2.36.0-1.fc32
audit-libs-3.0-0.19.20191104git1c2f876.fc32
avahi-libs-0.7-23.fc32
basesystem-11-9.fc32
bzip2-libs-1.0.8-2.fc32
checkpolicy-3.0-3.fc32
colord-libs-1.4.4-4.fc32
cpio-2.13-4.fc32
[...]

PR #799, opened 2020-06-26
distro: fix failed image build due to SELinux

msehnout:
Fixes: https://github.com/osbuild/osbuild-composer/issues/798
Depends on: #805


msehnout:
CI failures are coming from `curl` as usual. Otherwise I think it should be fine. Can you please take a look @teg ?

teg:
I think this is ready to go in, though I think it needs a rebase to pick up some CI fix (I think the tests pass but the reporting is wrong).
To make it easier to follow, I suggest reordering the PR something like this (feel free to order them differently of course, just a suggestion if you are going to rebase anyway):
 - preparatory patches (test case generation refactoring, printing debug info on test failure)
 - the functional changes (buildroot labelling + cp labelling)
 - one patch updating all the test cases (ideally would be done as part of the patches introducing the changes making the test updates necessary, but since there are two patches in this PR, that would probably not be worth it). Squash any fixes together into one patch, and explain what was done (probably just re-ran the generation). As aarch64 image tests are not ran in our CI, leave a comment why this is still ok (you ran them manually as part of generating the test cases)
 - bump the spec file

msehnout:
> I think this is ready to go in, though I think it needs a rebase to pick up some CI fix (I think the tests pass but the reporting is wrong).
> To make it easier to follow, I suggest reordering the PR something like this (feel free to order them differently of course, just a suggestion if you are going to rebase anyway):
> 
>     * preparatory patches (test case generation refactoring, printing debug info on test failure)
> 
>     * the functional changes (buildroot labelling + cp labelling)
> 
>     * one patch updating all the test cases (ideally would be done as part of the patches introducing the changes making the test updates necessary, but since there are two patches in this PR, that would probably not be worth it). Squash any fixes together into one patch, and explain what was done (probably just re-ran the generation). As aarch64 image tests are not ran in our CI, leave a comment why this is still ok (you ran them manually as part of generating the test cases)
> 
>     * bump the spec file

I think the bump should go first because without osbuild > 18, the `labels` field won't work. Is it ok like it is now?

msehnout:
The test cases are not working, because they are already outdated. I'm not sure we want to commit to updating them because this patch introduced `updates` repositories:
4b0409df5b8128dbcf3c83f67893720840342e62
If we wanted to keep track od this repo, we would have to update the test-cases all the time. Maybe we could deploy our own mirror, @major ?

major:
@msehnout That's possible, but it may be better to just regenerate the test cases without updates. I forgot that they would disappear in Fedora after they are updated again. 🤦🏻‍♂️

msehnout:
I regenerated the test cases and ran them. It went just fine, but I don't think it is necessary to update them until we agree on the path forward. If that is ok with you @teg , I  would not block this PR on this issue.

PR #804, opened 2020-06-29
test: bump 8.3 repo

ondrejbudai:
The previously used snapshot was deleted. This commit changes the override to
use Internal Snapshot 2.0, which should not be deleted in the near future.


test/internal-repos/rhel-8.json
```diff
@@ -1,13 +1,13 @@
 {
     "x86_64": [
         {
-            "name": "RHEL 8.3 Nightly BaseOS",
-            "baseurl": "http://download.eng.rdu.redhat.com/rhel-8/nightly/RHEL-8/RHEL-8.3.0-20200611.n.0/compose/BaseOS/x86_64/os/",
+            "name": "RHEL 8.3 Internal Snapshot 2.0 BaseOS",
+            "baseurl": "http://download.eng.rdu.redhat.com/rhel-8/rel-eng/RHEL-8/RHEL-8.3.0-InternalSnapshot-2.0/compose/BaseOS/x86_64/os/",
```
atodorov, line 5: Why not use `latest-RHEL-8` symlink instead ?

ondrejbudai: I would prefer the environment to be as fixed as possible, hence the fixed snapshot.

ondrejbudai:
According to Schutzbot, this fixes out 8.3 CI.

major:
Thanks for cleaning this up, @ondrejbudai. It was on my to-do list before PTO. :(

PR #805, opened 2020-06-29
osbuild: Add "labels" field into the SELinux stage

msehnout:


msehnout:
This will be part of #799 .

Issue #754, opened 2020-06-10
Intermittent AWS non-bootable instance failure


major:
The AWS boot test is intermittently failing due to a non-bootable image at AWS. 

![console-screenshot-rhel-8 2-20](https://user-images.githubusercontent.com/89910/84306520-3504d100-ab21-11ea-8d88-0b05c05963e1.jpg)

When this happens, no errors are shown anywhere in osbuild-composer's output. The image imports properly as a snapshot and the AMI gets registered without an issue.

I've proposed #753 to ensure we get the osbuild logs for the compose whether it fails or passes, and this will hopefully help us figure out why this is happening.

major:
So here's an osbuild log from a successful boot:

```
Output:
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
grub2-mkimage: info: the total module size is 0x585c.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/kernel.img.
grub2-mkimage: info: locating the section .text at 0x0.
grub2-mkimage: info: locating the section .rodata at 0x9b6c.
grub2-mkimage: info: locating the section .data at 0xb200.
grub2-mkimage: info: locating the section .module_license at 0xba64.
grub2-mkimage: info: locating the section .bss at 0xba80.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/biosdisk.mod.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/part_msdos.mod.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/fshelp.mod.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/ext2.mod.
grub2-mkimage: info: kernel_img=0x563d1503d770, kernel_size=0xba73.
grub2-mkimage: info: the core size is 0x6917.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/lzma_decompress.img.
grub2-mkimage: info: reading /usr/lib/grub/i386-pc/diskboot.img.
grub2-mkimage: info: writing 0x200 bytes.
grub2-mkimage: info: writing 0x7437 bytes.
mke2fs 1.45.5 (07-Jan-2020)
Discarding device blocks:    4096/1572608               done                            
Creating filesystem with 1572608 4k blocks and 393216 inodes
[...]

PR #807, opened 2020-06-29
16

ondrejbudai:
Release of version 16 & one small spec file patch - I'm looking forward to ditching the old specfile, so much pain...


ondrejbudai:
Fedora 32 integration test is failing because of, this is known to happen randomly, see #754

major:
@ondrejbudai

PR #808, opened 2020-06-30
Switch Fedora mock builds to minimal images

major:
Use minimal images when building RPMs via mock on Fedora (PR for RHEL
will come later).


major:
Failures are in image tests and are unrelated to the mock build.

PR #811, opened 2020-07-02
Fix azure tests leaking disks

ondrejbudai:
Previously, all resources were created with a certain tag. When the cleanup phase came, the Resources - List route\[1\] was used to get all resources with the tag. Then, they were deleted in the right order.

Sadly, the Resources - List API has issues with listing disks. Sometimes, it doesn't return the newly created disks for 15 minutes after they were created. As a result, the disks have been left behind quite often and our bill was higher than necessary.

This PR uses a different method - the Go code now knows all resource names, so it can delete all resources without listing them using the "broken" API route.

\[1\]: https://docs.microsoft.com/en-us/rest/api/resources/resources/list


ondrejbudai:
This is now ready to be reviewed.

ondrejbudai:
This is not connected with failing Schutzbot in any way (Azure code is not run there).

PR #815, opened 2020-07-02
🤔 Check disk space between image tests

major:
Trying to verify if VHD images are failing due to low disk space.


atodorov:
> Trying to verify if VHD images are failing due to low disk space.

regardless of why they fail this is good to have in the logs for future reference.

Issue #816, opened 2020-07-06
useradd step fails on Fedora 31


major:
```
useradd: cannot create directory /home/redhat
Traceback (most recent call last):
File \"/run/osbuild/lib/stages/org.osbuild.users\", line 168, in <module>
    r = main(args[\"tree\"], args[\"options\"])
  File \"/run/osbuild/lib/stages/org.osbuild.users\", line 157, in main
    useradd(tree, name, uid, gid, groups, description, home, shell, password)
  File \"/run/osbuild/lib/stages/org.osbuild.users\", line 100, in useradd
    subprocess.run([\"chroot\", root, \"useradd\", *arguments, name], check=True)
  File \"/usr/lib64/python3.7/subprocess.py\", line 487, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['chroot', '/run/osbuild/tree', 'useradd', 'redhat']' returned non-zero exit status 12.
```

Pipeline: https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/master/148/pipeline

Full osbuild log: https://osbuildci.cloud.paas.psi.redhat.com/job/osbuild/job/osbuild-composer/job/master/148/artifact/fedora31-image/fedora_31-x86_64-vhd-boot.json.log

major:
I can't reproduce this when run manually on a Fedora 31 VM. 🤦🏻‍♂️

major:
Hmm, I shuffled the test order in https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild/detail/PR-467/1/pipeline/111/ and then the openstack image failed instead.

major:
This seems to be happening less often lately, although I can't explain why.

ondrejbudai:
Happens also on `aarch64`:

https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-858/4/pipeline/454

ondrejbudai:
Fedora 31 is EOL

PR #821, opened 2020-07-07
⛅ Run mock builds on AWS and PSI

major:
Increase reliability and performance by running the mock builds at AWS
when PSI is having trouble.


major:
The only CI failure was from #816.

PR #825, opened 2020-07-07
🛃 Ensure the correct packages are installed

### LINKED ISSUE #824: CI is installing osbuild from Fedora repos, not the mock build repo
As @ondrejbudai noted earlier, CI is installing a newer version of osbuild from the Fedora repositories. osbuild 18-2 is in F32's repos, but osbuild-composer still points to version 17.

The rpms from the osbuild-mock repository *must be used every time*.

major:
Some dnf bugs existed in early releases of Fedora 31 + 32 around
repository priorities being ignored (like BZ 1733582. Ensure that we are
running the latest version of dnf before we try to install from the mock
build repo (which has a much higher priority than Fedora's default
repos).

Also, check to see if EPEL is installed already before installing it
again.

Fixes #824. 🥰

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -31,6 +31,10 @@ sudo rm -f /etc/yum.repos.d/fedora*modular*
 # dnf operations.
 echo -e "fastestmirror=1\ninstall_weak_deps=0" | sudo tee -a /etc/dnf/dnf.conf
 
+# Ensure we are using the latest dnf since early revisions of Fedora 31 had
+# some dnf repo priority bugs like BZ 1733582.
+sudo dnf -y upgrade
+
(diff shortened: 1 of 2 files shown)


major:
The F31 integration test failed due to a failed download from the Fedora internal mirror, but that mirror is being removed in #826. Everything else in CI looks ok.

teg:
Why doesn't the repo priorities work? Seems important

major:
From what I can tell, dnf is selecting the newest osbuild version from Fedora's repositories since it's newer than the one in the mock build repo.

If I tell dnf to use ONLY the mock build repo, it complains because it can't get dependencies. If I allow dnf to use all repos on the system, it chooses the newest osbuild from Fedora's repositories.

I'm still trying to figure out how that is happening.

PR #826, opened 2020-07-07
🫀 Use base cloud images for testing

### LINKED ISSUE #822: Use RHEL 8.3 beta content from staging CDN
8.3 Beta is on the staging CDN! http://cdn.stage.redhat.com/content/beta/rhel8/8/x86_64/baseos/os/Packages/r/redhat-release-8.3-0.0.el8.x86_64.rpm

We should use this instead of hardcoding the latest internal snapshot in CI.

### LINKED ISSUE #831: CDN url for RHEL 8.3 should be staging
The CDN url for RHEL 8.3 content is `cdn.redhat.com` but the content is only available on `cdn.stage.redhat.com` right now. This prevents RHEL 8.3 from installing packages in CI.

### LINKED ISSUE #832: Migrate to awscli 2.x
AWS now has a version 2 of it's cli tools and we should be using those for the best support/performance.

major:
By using base cloud images for CI, we better simulate a customer environment and find more bugs from missing dependencies. It also reduces the amount of hacks in the CI system.

Changes include:

* Bump the osbuild submodule to `v18` (fixes RHEL 8.x missing python issue)
  * This is mainly a duck tape fix for #830 but not a long term solution
* Remove internal repositories (they are not reliable and not usable at AWS)
* Updating all packages to pick up dnf/libsolv changes that appeared in F31/32 that breaks repos with priorities (Fixes #822)
* Allow failures in RHEL 8.3 Beta to cause the pipeline to fail
* Change RHEL 8.3 CDN urls to use staging CDN content (fixes #831)
* Install awscli version 2 (fixes #832)


major:
The only CI failure was from the issue in #816.

ondrejbudai:
This needs a rebase.

major:
@ondrejbudai Done!

major:
@ondrejbudai The RHEL breakage was from missing `/usr/bin/python3.6` in RHEL 8.x. See #830 for more discussion about how to handle this going forward.

major:
@ondrejbudai One of the failures is the useradd bug and the other is the occasional broken `image_info` check that only seems to happen with VHDs.

PR #827, opened 2020-07-07
image-info: various cleanups and use loop devices instead of nbd

gicmo:
I am preparing to remove the usage of `nbd` from `image-info` but wanted to give the code a bit of a cleanup before I started doing that. See the individual commit messages for details.

### DIFF
--- a/cmd/osbuild-image-tests/constants/constants-travis.go
+++ b/cmd/osbuild-image-tests/constants/constants-travis.go
@@ -2,7 +2,10 @@
 
 package constants
 
-import "os/exec"
+import (
+	"os"
+	"os/exec"
+)
 
 func GetOsbuildCommand(store, outputDirectory string) *exec.Cmd {
 	cmd := exec.Command(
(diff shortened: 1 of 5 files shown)


ondrejbudai:
Changes looking good to me. However, they uncovered the raciness of nbd in its full potential resulting in 100% failure rate in Travis. Let's put the `nbd` removal also in this PR, so we don't have to merge code which is making our CI really unhappy.

cmd/osbuild-image-tests/constants/constants-travis.go
```diff
@@ -18,13 +21,23 @@ func GetOsbuildCommand(store, outputDirectory string) *exec.Cmd {
...
+}
+
 var TestPaths = struct {
 	ImageInfo               string
 	PrivateKey              string
 	TestCasesDirectory      string
 	UserData                string
 	MetaData                string
 	AzureDeploymentTemplate string
+	PythonPath              string
```
gicmo, line 40: Fixed locally.

gicmo:
If the way to set the `PYTHONPATH` on travis is ok I will move the commit to the very top, so that when the actual change gets introduce it does not break that commit.

cmd/osbuild-image-tests/constants/constants-travis.go
ondrejbudai: LGTM!

ondrejbudai:
The only failure is because of iot, see #798

PR #828, opened 2020-07-08
spec: do not conflict with lorax-composer

teg:
Conflicting makes upgrades fail, obsoletes is sufficient.


ondrejbudai:
Flaky mirrors again. This has nothing to do with the mirror logic, therefore the fails aren't relevant.

PR #829, opened 2020-07-08
17

ondrejbudai:
Release osbuild-composer version 17


ondrejbudai:
Fedora mirrors are behaving weirdly, otherwise it's green.

Issue #834, opened 2020-07-09
CI: run tests on aarch64


teg:
Run the integration tests on aarch64 in the same way as they are run on x86_64.

The orchestration should be the same, but we may need to teach some tests to opt-out of certain cases depending on the architecture (not all image types are available everywhere).

To the extent schutzbot hardcodes any policies about which tests to run, this should be pushed down into the test-package to make this easier.

major:
Once #826 is merged, we should be able to use Fedora's aarch64 cloud image and the Cloud Access image for RHEL 8.2 aarch64 in AWS. The support for the new Graviton instances is in the latest version of the AWS SDK + EC2 plugin for Jenkins.

This will likely require a few tweaks to get package names right and test manifests working.

major:
Oof, we have to do `qemu-img` converts with one subroutine per https://bugs.launchpad.net/qemu/+bug/1805256

```
Command:

qemu-img convert -f qcow2 -O qcow2 ./disk01.qcow2 ./output.qcow2

Hangs indefinitely approximately 30% of the runs.

----

Workaround:

qemu-img convert -m 1 -f qcow2 -O qcow2 ./disk01.qcow2 ./output.qcow2

Run "qemu-img convert" with "a single coroutine" to avoid this issue.
```

teg:
Hm, well, this is sub-optimal. I guess this should go in as a fix to `osbuild`? @gicmo @dvdhrm?

major:
I ended up adding this to osbuild-composer's image-info script temporarily: https://github.com/osbuild/osbuild-composer/pull/882/commits/ac360f821993956c148ddffef889ffc694f5209c

gicmo:
> Hm, well, this is sub-optimal. I guess this should go in as a fix to `osbuild`? @gicmo @dvdhrm?

Yes, totally. I guess someone should file an issue there? :D

major:
@gicmo how about an issue and a PR? https://github.com/osbuild/osbuild/pull/483

atodorov:
We're currently running as many tests as we can on aarch64 and Ondrej is experimenting to see if we can enable a few more. 

That said there are limitations with the existing environment: no VPN, no nested virt which we are aware of and tracking internally. I am closing this issue. In the case something is missing and can be added let's open a more specific issue for it.

PR #837, opened 2020-07-09
🤹🏻‍♂️  Prepare for multi-arch support

major:
As we embark on a CI journey to other architectures, we need to be
explicit about choosing the right architecture for each CI node.

This is a portion of the work required for #834.


major:
@ondrejbudai nice ;)

major:
The deploy step has worked everywhere, so we should be good to go.

PR #842, opened 2020-07-09
⛅ Allow base/integration tests to run in PSI + AWS

major:
Make it easier to handle PSI outages or those times when PSI is overloaded with jobs by allowing base and integration tests to run on PSI OpenStack and AWS.

Also increase the reliability and speed of the integration tests that are booted locally by using a static IP address configuration. Keep checking for the smoke test file repeatedly without messing around with ssh key scanning. This helps when ssh comes up on an instance and then goes back down again.

Fixes #844.
Fixes #847.

PR #852, opened 2020-07-13
🤷🏻‍♂️ Allow tests to run on AWS and PSI

major:
In #842, we verified that our tests run on AWS properly and we should
now let Jenkins decide where to run the tests since they work in both
places.

Remove the hardcoded AWS label and allow tests to run where they please.


major:
@teg Here's that fast follow PR I mentioned in #842.

major:
Two image tests failed due to problems installing the kmod package. Everything else looks okay.

PR #856, opened 2020-07-14
 api/weldr: bring tars back from the future

### LINKED ISSUE #854: weldr/metadata: tar is from the future
```
[2020-07-13T13:34:30.843Z] 💬 Getting compose log and metadata
[2020-07-13T13:34:30.843Z] tar: bf548dfd-0a90-40e6-bbf2-dcdd82fcbb4e.json: time stamp 2020-07-13 13:34:31 is 0.356223173 s in the future
```

https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-849/1/pipeline

I have no words.

ondrejbudai:
By default, go's tar archiver uses USTAR header format. Unfortunately, this
format doesn't support sub-second resolution for ModTime. Go solves this by
*rounding* the time. Sometimes, this creates an archive containing a file
with modtime from the future. When such archive is untarred by GNU tar,
the following message is produced:

```
tar: bf548dfd-0a90-40e6-bbf2-dcdd82fcbb4e.json: time stamp 2020-07-13
13:34:31 is 0.356223173 s in the future
```

We have two options here:

1) Use gnu header format that supports sub-second resolution. Unfortunately,
   it seems that not all tar archivers support this format (e.g. 7-zip).

2) The other option is to truncate the date (instead of rounding).

I went with option 2.

Also, this commit adds a test to check that the header is not from the future.
Without this fix, the test is actually failing, I verified this manually.

Fixes #854

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -2421,9 +2421,10 @@ func (api *API) composeLogsHandler(writer http.ResponseWriter, request *http.Req
 	common.PanicOnError(err)
 
 	header := &tar.Header{
-		Name: "logs/osbuild.log",
-		Mode: 0644,
-		Size: int64(fileContents.Len()),
+		Name:    "logs/osbuild.log",
+		Mode:    0644,
+		Size:    int64(fileContents.Len()),
+		ModTime: time.Now().Truncate(time.Second),
 	}
 
 	err = tw.WriteHeader(header)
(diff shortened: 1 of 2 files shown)


teg:
The CI failure is unrelated, issue #862.

PR #864, opened 2020-07-15
crypt: add automatic darwin support

teg:
Currently one would have to pass a `macos` flag explicitly to build on OSX, but if we replace `macos` with `darwin` this happens automatically.

### DIFF
--- a/internal/crypt/crypt.go
+++ b/internal/crypt/crypt.go
@@ -6,6 +6,9 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


teg:
Thanks Martin! Should now be fixed :)

msehnout:
CI seemed to be stuck so I restarted it. I'll wait for it to complete even though I don't expect any issues.

msehnout:
CI failures are unrelated (failed to download RPMs)

PR #865, opened 2020-07-16
remove unneeded fedora 30 repositories

ondrejbudai:
Fedora 30 support was dropped in e66368b0, this file has had no use since
then. This commit removes it.


msehnout:
CI failures are caused by `curl` as usual.

PR #871, opened 2020-07-17
image-tests: avoid /var/tmp

teg:
`systemd-tmpfiles` will helpfully delete "old"  files in /var/tmp at regular
intervals. The files installed from rpm has the timestamps from when they
were packaged, which causes some to be cleaned up when the timer triggers.

The first timer triggers 15 minutes after boot, so we were sometimes hit
by this when our CI was under load.


NEWS.md
```diff
@@ -32,7 +32,7 @@ Contributions from: Major Hayden, Tom Gundersen
   * Composer now correctly passes UID and GID to org.osbuild.{users,groups}
     stages as ints instead of strings.
 
-  * The subpackages (worker, tests and rcm) now require a matching version of
+  * The subpackages (worker and tests) now require a matching version of
```
ondrejbudai, line 35: I would rather not change the NEWS. It just feels weird to me. I do understand that RCM API was never properly supported but still it's just weird.

teg: I agree, that was unintentional, I'll revert that part

teg: Done.

PR #873, opened 2020-07-17
Do not obsolete lorax-composer

bcl:
Doing this makes it impossible to install lorax-composer when you want
it, which breaks testing in lorax as well as causes problems for users
who specifically want to install it.

Resolves: rhbz#1858051

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -55,7 +55,6 @@ Requires: osbuild-ostree >= 17
 Provides: weldr
 
 %if 0%{?rhel}
-Obsoletes: lorax-composer <= 29
 Conflicts: lorax-composer
 %endif
 


teg:
This partially reverts f87154af750828f4977a9a4c04369a4a366007c7, would it reintroduce the bug that was meant to fix?

atodorov:
> This partially reverts [f87154a](https://github.com/osbuild/osbuild-composer/commit/f87154af750828f4977a9a4c04369a4a366007c7), would it reintroduce the bug that was meant to fix?

How do we deal with the upgrade scenario then ?  You can't have both backends side by side with the new one automatically replacing the old one without obsoletes. Who can tell what's the behavior that we want. 

qa_nack on this PR until we've got answers.

teg:
osbuild-composer is meant to obsolete lorax composer, the way to get around this is by using `excludes` in `dnf`, if necessary. We want to transition users over by default.

PR #874, opened 2020-07-17
🏇🏻 Speed up pipeline with parallelism (again)

major:
Now that we're going about multi-arch testing in a different way, we can
restore some of the old parallelism that we used to have in the pipeline
so it will finish faster.


major:
The big fella has returned, @teg. 🤣

Screenshot from 2020-07-17 11-43-13

PR #882, opened 2020-07-21
Prepare for multi-arch testing in Beaker

major:
This PR:

* Applies a workaround for a qemu-convert bug on aarch64 with lots of CPUs
* Removes nvram as part of the VM deletion in integration tests on aarch64
* Increases sizes for images in image test cases to 4GB on RHEL 8.3
* Moves test cases to use RHEL 8.3.0 Beta 1.0
* Removes Fedora updates repo on aarch64 test cases
* Updates test cases for aarch64, ppc64le, and x86_64


tools/image-info
```diff
@@ -64,8 +65,22 @@ def open_image(ctl, image, fmt):
     with tempfile.TemporaryDirectory() as tmp:
         if fmt != "raw":
             target = os.path.join(tmp, "image.raw")
-            subprocess.run(["qemu-img", "convert", "-O", "raw", image, target],
-                           check=True)
+            # A bug exists in qemu that causes the conversion to raw to fail
```
teg, line 68: Looks reasonable to me. @gicmo, you agree?

internal/distro/rhel8/distro.go
```diff
@@ -847,7 +847,7 @@ func New() distro.Distro {
 		},
 		kernelOptions: "console=ttyS0 console=ttyS0,115200n8 no_timer_check crashkernel=auto net.ifnames=0",
 		bootable:      true,
-		defaultSize:   2 * GigaByte,
+		defaultSize:   4 * GigaByte,
```
teg, line 850: I think this is right if the default images really do need more than 4GB. But this does not only change the test cases, but the actual defaults we provide, so should update the documentation too (each image type has high-level docs). And the commit message should probably reflect that. I assume the actual test cases are regenerated in a separate commit?

major: @teg The test case changes are in [`de1e3c6` (#882)](https://github.com/osbuild/osbuild-composer/pull/882/commits/de1e3c6d18e7d3ce853b640cdc0b71a516efb122).

I'll update the docs/commit in a short while.

major: For what it's worth, we don't mention image sizes in the docs right now (but that might be a good thing for the future.

major:
@teg Let me know how you want to proceed with this. The next step would be to use a keytab to send jobs to Beaker.

internal/distro/rhel8/distro.go
teg: My bad! Could you just update the commit message, so it doesn't look like this is just about changing the tests?

```
distro/rhel8: increase default size to 4GB

RHEL 8.3 requires more than 2GB to install on aarch64/ppc64le.
```
or something to that effect?

tools/image-info
gicmo: For stylistic reasons would have probably gone for a:

```
cmd = ["qemu-img", "convert", "-m", "1", "-O", "raw"]
if platform.machine() == 'aarch64':
   cmd += ["-m", "1"]
subprocess.run(cmd + [image, target], ...
```
but I am totally fine with the logic. :+1:

internal/distro/rhel8/distro.go
major: Sure!

major: @teg Done!

tools/image-info
teg: Let's leave cleaunps for a possible follow-up then.

PR #887, opened 2020-07-23
[WIP] Run osbuild-composer in separate "target" VM in CI

msehnout:
See the commit message

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -474,13 +474,24 @@ void setup_remote_osbuild_composer(template, credentials) {
                 }
             }
             // Install and run osbuild-composer inside the VM
+            // TODO: copy environment variables needed for RHEL setup
+            sh (
+                // The target VM does not contain the git checkout which is available in the test VM
+                // but we still need few files there
+                label: "Prepare directories",
+                script:  "ssh $sshOpts -i ${keyFile} ${username}@${vm.address} 'mkdir /tmp/schutzbot /tmp/test/external-repos/'"
+            )
             sh (
                 label: "Copy repo file to the target VM",
                 script: "scp $sshOpts -i ${keyFile} osbuild-mock.repo ${username}@${vm.address}:/tmp/"
             )
             sh (
                 label: "Copy deployment configuration to the target VM",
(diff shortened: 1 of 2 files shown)


schutzbot/Jenkinsfile
```diff
@@ -417,3 +439,65 @@ void preserve_logs(test_slug) {
...
+                label: "Prepare Weldr API socket for forwarding (allow the Jenkins user to read/write from/to it)",
+                script:  "ssh $sshOpts -i ${keyFile} ${username}@${vm.address} 'sudo chmod go+rw /run/weldr/api.socket'"
+            )
+            sh (
+                label: "Prepare local /run/weldr directory for the forwarded socket",
+                script:  "sudo mkdir /run/weldr"
+            )
+            sh (
+                label: "Forward the Weldr API socket from the target VM to the test VM",
+                script:  "sudo ssh $sshOpts -i ${keyFile} -fN -L /run/weldr/api.socket:/run/weldr/api.socket ${username}@${vm.address}"
```
atodorov, line 499: I remember a bug with this type of forwarding on RHEL but can't find the reference. It was a missing patch that wasn't backported from upstream.

msehnout: Thanks for the note. I'll look into it in case there is an issue, but so far it fails even sooner for RHEL.

major:
The code itself seems to be coming along well, but now this requires creating 2x the amount of VMs and could get us deadlocked if we have a LOT of instances running at once and we hit our quota limit. Also, I worry what happens if we're overloaded in PSI and Jenkins is forced to spawn a worker in AWS instead.

msehnout:
> The code itself seems to be coming along well, but now this requires creating 2x the amount of VMs and could get us deadlocked if we have a LOT of instances running at once and we hit our quota limit. Also, I worry what happens if we're overloaded in PSI and Jenkins is forced to spawn a worker in AWS instead.

That does not sound great. Do you think running the target VMs inside the OpenStack VMs for stages (e.g. using qemu) would be a better solution than creating a separate OpenStack VMs?

major:
@msehnout That could be possible. That would be a little slower in AWS since we don't have nested virtualization.

msehnout:
This does not look like a promising direction. I'll soon submit alternative version using AWS.

PR #897, opened 2020-07-29
store: Fix copy failure when initializing blueprints commit list

bcl:
copy() will not append to a slice, so if the length is 0 nothing will be
copied. Initialize the slice with the length instead of the capacity.

This fixes a problem with blueprint commits vanishing when the server is
restarted.


ondrejbudai:
Can we cover this by a test-case?

bcl:
> Can we cover this by a test-case? Otherwise, the change is looking good to me, thanks!

I'd like to, but I'm not sure if we can. I'll take a closer look today. Maybe just a targeted test of this function instead of needing multiple copies of mocked state.json files.

PR #906, opened 2020-08-03
schutzbot: don't skip image test for fedora iot commit

msehnout:
The SELinux bug has been fixed so there is no need to skip it.


major:
@msehnout Yay! Thanks for turning this back on again.

PR #910, opened 2020-08-05
Add friendly message for logs on running build

major:
Improve the message returned by osbuild-composer when a user asks for
logs of a compose that is still running.


ondrejbudai:
openstack fail is unrelated

PR #925, opened 2020-08-18
Add Install section to remote worker unit file

### LINKED ISSUE #924: Unable to enable remote workers on boot
If I try to enable remote workers at boot time, it fails:

```
# systemctl enable --now osbuild-remote-worker@composer:8700.service
The unit files have no installation config (WantedBy, RequiredBy, Also, Alias
settings in the [Install] section, and DefaultInstance for template units).
This means they are not meant to be enabled using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
   D-Bus, udev, scripted systemctl call, ...).
4) In case of template units, the unit is meant to be enabled with some
   instance name specified.
```

major:
Although the obuild-remote-worker@.service unit can be started, it can't
be enabled at boot time since the `Install` section is missing in the
unit file.

Add a small `[Install]` section with the same `WantedBy` as
osbuild-composr.service.

Fixes #924.

### DIFF
--- a/distribution/osbuild-remote-worker@.service
+++ b/distribution/osbuild-remote-worker@.service
@@ -13,3 +13,6 @@ IOSchedulingClass=idle
 CacheDirectory=osbuild-worker
 # systemd >= 240 sets this, but osbuild-worker runs on earlier versions
 Environment="CACHE_DIRECTORY=/var/cache/osbuild-worker"
+
+[Install]
+WantedBy=multi-user.target


major:
> Looks good! We should have tests for the remote worker as well as the local ones... (pointing this out to no one in particular)

To be fair, we totally missed this with the remote testing we had in CI before now because Ansible will ignore the `systemctl enable` error and just quietly start the service. 🤣

major:
@teg Sure, just let me know when.

teg (approved):
This is good to go as soon as CI is happy again.

major:
@teg Thank ya!

major:
Rebased.

major:
@teg linter timed out (fixed in #945) but this looks green otherwise.

PR #926, opened 2020-08-18
dnf-json: Sort depsolve pkg list

major:
Running `composer-cli blueprints depsolve` gives an unordered list of
packages and this can make it difficult to figure out if your desired
package is in the list.

Sort the list of depsolved packages by the name of the package when
returning it from dnf-json.

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -199,5 +199,5 @@ with tempfile.TemporaryDirectory() as persistdir:
             })
         json.dump({
             "checksums": repo_checksums(base),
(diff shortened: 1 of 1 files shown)


teg:
The order the packages are returned in matters for how they are installed by osbuild, so we should not sort them in `dnf-json`. We could still sort the list of packages in the handler in weldr, at that point the order should not be important.

major:
@teg Oh, I didn't consider that. I'll take a look at the weldr API stuff and figure out where to do that.

PR #947, opened 2020-08-25
osbuild: rpm metadata

teg:
Parse the metadata about the RPM stage as returned from osbuild.

### DIFF
--- a/internal/osbuild/result.go
+++ b/internal/osbuild/result.go
@@ -6,33 +6,74 @@ import (
 	"io"
 )
 
-type assembler struct {
+type rawAssemblerResult struct {
 	Name    string          `json:"name"`
 	Options json.RawMessage `json:"options"`
 	Success bool            `json:"success"`
 	Output  string          `json:"output"`
 }
 
-type stage struct {
-	Name    string          `json:"name"`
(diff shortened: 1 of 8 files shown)


internal/osbuild/result.go
```diff
@@ -6,33 +6,74 @@ import (
...
 	Success bool            `json:"success"`
 	Output  string          `json:"output"`
 }
 
-type stage struct {
-	Name    string          `json:"name"`
-	Options json.RawMessage `json:"options"`
-	Success bool            `json:"success"`
-	Output  string          `json:"output"`
+type StageResult struct {
```
larskarlitski, line 16: Why did you make this public? It doesn't seem to be used anywhere outside this package.

larskarlitski (approved):
Great work, thanks! I have a small question, but don't block on that.

internal/osbuild/result.go
teg: We need to translate this to the koji metadata format, not sure where the translation will live, but i assumed not in this package.

gicmo (approved):
:ok_person: Cool stuff!

ondrejbudai (approved):
Looks good, thanks @teg

PR #948, opened 2020-08-26
rhel8.3: restrict testing after branching

teg:
This branch is where we will base any backports to RHEL8.3. Start off by restricting the testing to only test against the relevant distros.

We also make sure to test against osbuild that has landed in 8.3, rather than the submodule, when we can.


ondrejbudai (commented):
Changes look good to me, I restarted Jenkins (it had lost connection with AWS during the testing, weird).

PR #949, opened 2020-08-27
Skeleton for testing EL8 nightly builds

atodorov:
@major please review. I've added a few TODO comments for items which aren't clear to me ATM.


schutzbot/Jenkinsfile
```diff
@@ -344,6 +344,141 @@ pipeline {
...
+                        unstash 'rhel8json'
+                        run_tests('base')
+                    }
+                    post {
+                        always {
+                            preserve_logs('rhel8-base')
+                        }
+                    }
+                }
+                stage('EL8 Image') {
```
teg, line 420: Note that we cannot run the image tests against a moving target. So we would need to drop this.

I think the solution for this is to move all the boot-tests from image tests to integration tests, which can be run in any environment.

atodorov: > Note that we cannot run the image tests against a moving target. So we would need to drop this.

To be more precise you mean we can't execute the content verification portion of these tests. OTOH the upload & boot to cloud portion can be executed.

For one there's `-disable-local-boot` and we can very easily add another option, e.g. `-disable-content-check`. OTOH I've been running these same tests internally against snapshots and they were passing so I need to check what exactly is happening in that case.

atodorov: FTR these image tests vs. integration tests are getting kind of duplicate & redundant. We should clean them up. I'll open an issue for that.

teg: I agree. We should make the content check and the boot tests completely separate. (And the boot tests should be moved to the integration tests). A bit of rework is required, but I think all the code can easily be reused.

atodorov: QE notes:

- I'm playing around with a way to skip the image info part of these tests but I think it may not be necessary

- ATM we have "EL8 Image" and "EL8.3 Image" stages in the CI pipeline, which differ by the RHN registration command they use, hence the final list of sources. However there is only 1 set of json files for the image tests and inside the manifest section they contain hard-coded URLs for every single RPM. All of this goes directly to osbuild instead of osbuild-composer and I think this is the reason it hasn't failed for me so far in all of my internal executions. 

Essentially the OS version could be a moving target but the content for the image tests is always hardcoded. 

@teg I think the original concern expressed in the first comment is nothing to be worried about ATM and the last 2 commits I've added are not needed. What do you say ?

atodorov:
@major do you mind if we take the first 2 commits from this PR and modify them to call an empty shell script and then merge that so we can see if the pipeline configuration is working or not ?

schutzbot/Jenkinsfile
teg: You are absolutely right. Keeping the image tests as-is is harmless. I'm sorry for causing confusion here.

We should still improve on this in the future by splitting all the boot tests from the image tests so the boot tests can easily be run against the latest nightly (or whatever we point them at), rather than the fixed sources. The way it is now we might think we test more than we do.

atodorov: > We should still improve on this in the future by splitting all the boot tests from the image tests so the boot tests can easily be run against the latest nightly (or whatever we point them at), rather than the fixed sources. The way it is now we might think we test more than we do.

Agreed. I've opened https://github.com/osbuild/osbuild-composer/issues/962 so we don't forget and after we're done with the Jenkins configuration my plan is to start working on improving these tests.

major:
@atodorov That sounds okay to me.

atodorov:
> @atodorov That sounds okay to me.

PR updated, notice the "dummy-" prefix for `run_tests()`.  I've left the TODOs so we can take care of them once we see how all of this works in practice.

PR #950, opened 2020-08-27
⛈  Use AWS more often for testing

major:
Now that we've reduced how much of our PSI quota we are using so the
OpenStack boot tests will work, we need to use AWS for jobs more often.
This should allow test runs to complete a little sooner by freeing up
PSI resources for the jobs that are only able to run there.

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -41,7 +41,7 @@ pipeline {
             parallel {
 
                 stage('F31') {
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):

major:
RHEL 8 integration broke for a weird reason. Everything else passed. Running it one more time...

ondrejbudai:
The image tests are actually broken on `master`, see #953. The other tests are green.

PR #893, opened 2020-07-28
distro/rhel8: Updates for RHEL for Edge

nullr0ute:
We now have greenboot available so update the packages and services
also add exclude sections as subscription-manager is for some reason
getting pulled in which brings dnf and other deps we explicitly don't
want.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

PR #955, opened 2020-09-01
test: Add ostree image installation and upgrade test

henrywang:
`ostree.sh` runs ostree image building, installation, and upgrade test.
`check_ostree.yaml` is a ansible playbook, it runs ostree system test. It does not include `greenboot` related test case in this PR, that'll not be added until PR #893 got merged.

@major, I found that Jenkins would not load jenkinsfile which got updated in a PR. It keeps using jenkinsfile in master because Jenkins does not support jenkinsfile dynamic loading. Could you please tell me a way to trigger my test? Thanks.

My terminal `st` does not support color emoji due to [this issue](https://gitlab.freedesktop.org/xorg/lib/libxft/-/merge_requests/1), I kept old emoji but can't add new one. Sorry about that @major. Feel free to add emoji if you like.


major:
I can't figure out why the script never ran, but I started the job over again to see if it works properly the second time. https://osbuildci.cloud.paas.psi.redhat.com/job/osbuild/job/osbuild-composer/view/change-requests/job/PR-955/2/

schutzbot/Jenkinsfile
```diff
@@ -317,6 +328,18 @@ pipeline {
                         }
                     }
                 }
+                stage('EL8.3 OSTree') {
```
atodorov, line 331: Same question as above, why not a target for EL8 too ?

test/image-tests/ostree.sh
```diff
@@ -0,0 +1,479 @@
...
+done
+
+# Check image installation result
+check_result
+
+##################################################
+##
+## ostree image/commit upgrade
+##
+##################################################
```
atodorov, line 389: At first glance I am not sure what the script below does. I see building 2 images with composer-cli but it is unclear to me why do we have to extract them and then pull/commit with osbuild ? Can you explain what the usage scenario here is ?

schutzbot/Jenkinsfile
```diff
@@ -209,7 +209,18 @@ pipeline {
...
+                    steps {
+                        unstash 'fedora32'
+                        run_tests('ostree')
+                    }
+                    post {
+                        always {
+                            preserve_logs('fedora32-ostree')
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
teg: We only support Fedora IoT on F32.

schutzbot/Jenkinsfile
teg: Same. RHEL for Edge is only supported on 8.3, but EL8 uses the latest GA (8.2).

test/image-tests/ostree.sh
henrywang: The first one is for image installation test, and the second one is for ostree upgrade test. So you have to give it `parent commit` and ostree `ref` args in composer-cli.
To run ostree upgrade test, the ostree repo needs a new commit which made some change, add wget package in this case. So a new image has to be built with ostree `parent commit` and `ref` configured, they're two args of composer-cli.
Then use local pull to `merge` upgrade ostree commit into installation commit and summary the change.
After that the ostree `client` is able to do upgrade.

test/image-tests/ostree.sh
```diff
@@ -0,0 +1,479 @@
...
+[ostree_guest:vars]
+ansible_python_interpreter=/usr/bin/python3
+ansible_user=admin
+ansible_private_key_file=${SSH_KEY}
+ansible_ssh_common_args="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+EOF
+
+# Test IoT/Edge OS
+ansible-playbook -v -i ${TEMPDIR}/inventory -e image_type=${IMAGE_TYPE} -e ostree_commit=${UPGRADE_HASH} $(dirname $0)/check_ostree.yaml || RESULTS=0
+check_result
```
atodorov, line 474: `check_result` will inspect the `RESULTS` variable and I don't see where it is assigned a value if the or condition doesn't branch (e.g. if the part to the right of || operator isn't executed). From what I can tell the last assignment to this variable is on line 449 when waiting for SSH.

test/image-tests/ostree.sh
```diff
@@ -0,0 +1,479 @@
...
+
+# Start VM.
+greenprint "Start VM"
+sudo virsh start ${IMAGE_KEY}
+
+# Check for ssh ready to go.
+greenprint "🛃 Checking for SSH is ready to go"
+for LOOP_COUNTER in `seq 0 30`; do
+    RESULTS="$(wait_for_ssh_up $GUEST_ADDRESS)"
+    if [[ $RESULTS == 1 ]]; then
```
atodorov, line 375: Isn't 0 supposed to mean success and non-zero to mean error? Any reason for flipping them around ?

test/image-tests/ostree.sh
atodorov: OK, I see what the intention here is but that looks like a very complicated use case to me. 
1) Build an image with composer-cli
2) Extract it under /var/www/html
3) Then use anaconda + kickstart to create a VM from the extracted contents

4) Repeat 1-3 only so that we can try an upgrade procedure. 
5) Call the ansible playbook to verify the internals for the running VM


1-3 somewhat duplicate osbuild-image-tests but not entirely. I've opened #962 to start cleaning this up. 

4) looks like a good use-case but in case of problems I have to ask what the root cause of that problem would be ? On line 440 `rpm-ostree upgrade` could possibly fail or maybe fail on 427-428 when pulling the new commits into the ostree repository.  But all of that would indicate a faulty image coming out of osbuild-composer (assuming extraction works fine). Can't this be verified with some sort of a content sanity test? E.g. isn't the "image info" portion of osbuild-image-tests ensuring exactly that ? 

I can see the e2e path here but it really looks like we're testing `osbuild pull-local/update` and `rpm-ostree upgrade` instead of Image Builder.

for 5) shouldn't the script call the ansible playbook somewhere around line 380, e.g. after the initial VM is booted up; That will examine the starting state of the VM before upgrade and looks like if you pass the appropriate values from the outside the playbook will work fine.

atodorov (commented):
A few more comments around checking results value.

test/image-tests/ostree.sh
henrywang: Thanks @atodorov. :+1: 
1. Issue happened between 3 and 4. That means VM installed but can't boot, at least can't ssh to VM.
2. I have no idea about how to do ostree content sanity test. There's no issue or error in build log.
3. I run the same code in another rhel 8.3 VM, everything works well.
4. Adding ansible playbook after ostree installation is a good idea. I'll update that.

test/image-tests/ostree.sh
henrywang: RESULTS will be assigned to `0` or `1` at line 201. function `wait_for_ssh_up` always run before function `check_result`, line 374 and 449.

test/image-tests/ostree.sh
henrywang: No special reason, I just follow [qemu.sh](https://github.com/osbuild/osbuild-composer/blob/master/test/image-tests/qemu.sh) :-)

test/image-tests/ostree.sh
teg: > I can see the e2e path here but it really looks like we're testing osbuild pull-local/update and rpm-ostree upgrade instead of Image Builder.

Just to answer to this: basically the properties we need from the commits is that they can be installed and upgraded (and more things, but those are the basics). There is no way that I know to verify this without going through the motions of the end-to-end test. One could argue that some of this should be tested elsewhere, but as far as I know that is not being done so we are sort of stuck with having to verify that.
I do agree that tests like image-info /should/ catch most things, but that depends on us having gotten the `image-info` tool correct, so I think doing the e2e is the only way to make sure.

teg (approved):
Thanks for all your work on this!

teg:
The F32 Integration test failure must be unrelated.

teg (approved):
Thanks again!

PR #944, opened 2020-08-24
run base tests partially in aws

msehnout:
The idea is to upload OS images already containing latest commit of
osbuild-composer to AWS and boot them for base tests. The Weldr API
socket will be forwarded over SSH just like it was in PR:
https://github.com/osbuild/osbuild-composer/pull/887

PR #956, opened 2020-09-01
Refactor context managers to the /internals directory

msehnout:
This is needed for https://github.com/osbuild/osbuild-composer/pull/944 where the code was simply copy&pasted to show the intention. 

Further changes will be needed in the newly created `/internals/boot` package, but I want to keep this PR minimal. Also feel free to suggest better name than "boot" if you can think of some.


msehnout:
The CI timeouted 3 times, but this time it is the "greenest" of all the runs. The PR is ready for review.

teg (approved):
Nice, thanks! This was easy to review. I think as a matter of style/consistency, the commits should have been squashed into one to make sure each commit is buildable/bisectabel. But that is a very minor point, and should not block this being merged.

PR #959, opened 2020-09-01
distro/rhel/edge: add greenboot support

teg:
This is a copy of #893 and #955. Trying to figure out the changes to schutzbot.


henrywang:
@teg, Thanks for doing this. :+1: 
1. fedora 31 and rhel 8.2 does not support ostree image from [code]( https://github.com/osbuild/osbuild-composer/blob/25c3deda838b4a7757759236229d9a1028dc1d06/internal/distro/fedora31/distro.go#L192). I'll add `exit 0` to bypass ostree test on fedora 31 and rhel 8.2 scenarios. Does that make sense for you?
2. I got `WARNING  KVM acceleration not available, using 'qemu'` from [log](https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-959/1/pipeline/152#step-389-log-328). It seems VM does not support nested?
3. I got another log `error: Cannot run interactive console without a controlling TTY`. That means I have to add `--noautoconsole and --wait=-1` as args of `virt-install`. That'll miss all installation and boot logs. If something wrong, like can't boot, we have to run it again manually to find out root cause.

PR #960, opened 2020-09-02
Run base tests partially in AWS (2nd version)

msehnout:
This PR introduces an alternative version of running base tests for Fedora 32.  The idea is to create a "target" image (that is a pristine image with only osbuild-composer installed) which will run in AWS and the Weldr API socket will be forwarded over SSH to the machine where test binaries (from the -tests RPM subpackage) are running.

Because we already have code for cloud upload and spawning EC2 VMs, this PR contains a lot of refactoring in order to use as much existing code as possible and introduce as little new code as possible.

How it works:
 1. The mock build stage now creates "sources" file alongside the "repo" file
 2. A new stage "Creating target images 🎯" is introduced to run osbuild-composer and create the target image. The cloud upload functionality is used to upload the image directly to EC2.
 3. A new stage "F32 Base with target VM" is introduced to run a newly created binary `cmd/osbuild-base-test-runner` which takes the snapshot/image in EC2 boots it and runs a test against it. At the end of the run, the image is deregistered. This is repeated for every test specified in the `test/base-tests/run.sh` script. At the end of the script a "cleanup" code is invoked which will clean all remaining artifacts from AWS.

### DIFF
--- a/cmd/osbuild-base-test-runner/main.go
+++ b/cmd/osbuild-base-test-runner/main.go
@@ -171,6 +171,8 @@ func runTestOverSSH(address, privateKey, testCommand string) int {
 	// Get details about the machine
 	err := runSSHCommand(address, privateKey, "cat /etc/os-release")
 	panicErr(err)
+	err = runSSHCommand(address, privateKey, "df -h")
+	panicErr(err)
 
 	// Prepare the socket for forwarding because it is not owned by the redhat user
 	err = runSSHCommand(address, privateKey, "sudo chmod go+rw /run/weldr/api.socket")
--- a/cmd/osbuild-base-test-runner/main.go
+++ b/cmd/osbuild-base-test-runner/main.go
@@ -1,0 +1,225 @@
+// +build integration
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/osbuild/osbuild-composer/internal/boot"
+)
+
+type timeoutError struct{}
+
+func (*timeoutError) Error() string { return "Timeout exceeded" }
+
+// GenerateCIArtifactName generates a new identifier for CI artifacts which is based
+// on environment variables specified by Jenkins
+// note: in case of migration to sth else like Github Actions, change it to whatever variables GH Action provides
+func GenerateCIArtifactName(prefix string) (string, error) {
+	distroCode := os.Getenv("DISTRO_CODE")
+	changeId := os.Getenv("CHANGE_ID")
+	buildId := os.Getenv("BUILD_ID")
+	if changeId == "" || buildId == "" || distroCode == "" {
+		return "", fmt.Errorf("The environment variables must specify CHANGE_ID, BUILD_ID, and DISTRO_CODE")
+	}
+
+	return fmt.Sprintf("%s%s-%s-%s", prefix, distroCode, changeId, buildId), nil
+}
+
+// sshDisableHostChecking disables host (=remote machine) key checking
+// because the key of the fresh VM in AWS is not in the list of known_hosts.
+// The function also uses an empty known_hosts file by setting it to /dev/null.
+func sshDisableHostChecking() []string {
+	return []string{
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+	}
+}
+
+func runLocalCommand(name string, args ...string) {
+	cmd := exec.Command(name, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		fmt.Println("Error: ", err)
+	}
+}
+
+func runSSHCommand(address, privateKey, command string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	cmdName := "ssh"
+	cmdArgs := []string{
+		"-i", privateKey,
+	}
+	cmdArgs = append(cmdArgs, sshDisableHostChecking()...)
+	cmdArgs = append(cmdArgs, "redhat@"+address, command)
+
+	var cmd *exec.Cmd = exec.CommandContext(ctx, cmdName, cmdArgs...)
+
+	output, err := cmd.Output()
+
+	if ctx.Err() == context.DeadlineExceeded {
+		return &timeoutError{}
+	}
+
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == 255 {
+				return &timeoutError{}
+			}
+		} else {
+			return fmt.Errorf("ssh command failed from unknown reason: %#v", err)
+		}
+	}
+	outputString := strings.TrimSpace(string(output))
+	fmt.Println(outputString)
+
+	return nil
+}
+
+// testSSHOnce tries to test the running image using ssh once
+// It returns timeoutError if ssh command returns 255, if it runs for more
+// that 10 seconds or if systemd-is-running returns starting.
+// It returns nil if systemd-is-running returns running or degraded.
+// It can also return other errors in other error cases.
+func testSSHOnce(address string, privateKey string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	cmdName := "ssh"
+	cmdArgs := []string{
+		"-i", privateKey,
+	}
+	cmdArgs = append(cmdArgs, sshDisableHostChecking()...)
+	cmdArgs = append(cmdArgs, "redhat@"+address, "systemctl --wait is-system-running")
+
+	var cmd *exec.Cmd = exec.CommandContext(ctx, cmdName, cmdArgs...)
+
+	output, err := cmd.Output()
+
+	if ctx.Err() == context.DeadlineExceeded {
+		return &timeoutError{}
+	}
+
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == 255 {
+				return &timeoutError{}
+			}
+		} else {
+			return fmt.Errorf("ssh command failed from unknown reason: %#v", err)
+		}
+	}
+
+	outputString := strings.TrimSpace(string(output))
+	switch outputString {
+	case "running":
+		fmt.Println("ssh test passed")
+		return nil
+	case "degraded":
+		fmt.Println("ssh test passed, but the system is degraded")
+		return nil
+	case "starting":
+		return &timeoutError{}
+	default:
+		return fmt.Errorf("ssh test failed, system status is: %s", outputString)
+	}
+}
+
+// testSSH tests the running image using ssh.
+// It tries 20 attempts before giving up. If a major error occurs, it might
+// return earlier.
+func testSSH(address string, privateKey string) {
+	const attempts = 20
+	for i := 0; i < attempts; i++ {
+		err := testSSHOnce(address, privateKey)
+		if err == nil {
+			// pass the test
+			return
+		}
+
+		// if any other error than the timeout one happened, fail the test immediately
+		if _, ok := err.(*timeoutError); !ok {
+			panic(err)
+		}
+
+		time.Sleep(10 * time.Second)
+	}
+
+	panic(fmt.Sprintf("ssh test failure, %d attempts were made", attempts))
+}
+
+func main() {
+	fmt.Println("Getting AWS credentials")
+	creds, err := boot.GetAWSCredentialsFromEnv()
+	if err != nil {
+		panic("AWS credentials unavailable")
+	}
+	if creds == nil {
+		panic("Empty AWS credentials")
+	}
+<<<<<<< HEAD
+=======
+
+	// Get environment variables defining the name of CI artifacts
+>>>>>>> 1fc50d4... amend introduce
+	fmt.Println("Getting change and build IDs")
+	changeId := os.Getenv("CHANGE_ID")
+	buildId := os.Getenv("BUILD_ID")
+	if changeId == "" || buildId == "" {
+		panic("The environment variables must specify CHANGE_ID and BUILD_ID")
+	}
+	imageName := fmt.Sprintf("osbuild-composer-base-test-%s-%s", changeId, buildId)
+<<<<<<< HEAD
+	fmt.Println("Getting the EC2 image description")
+=======
+
+	securityGroupName, err := GenerateCIArtifactName("osbuild-image-tests-security-group-")
+	if err != nil {
+		panic(err)
+	}
+
+>>>>>>> 1fc50d4... amend introduce
+	e, err := boot.NewEC2(creds)
+	if err != nil {
+		panic("Failed to obtain credentials for EC2")
+	}
+	imageDesc, err := boot.DescribeEC2Image(e, imageName)
+	if err != nil {
+		panic("Failed to describe EC2 image")
+	}
+	// delete the image after the test is over
+	defer func() {
+		err = boot.DeregisterEC2Image(e, imageDesc)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


dvdhrm:
> See commits for more details.

Can you give a high-level overview of this PR? I don't mind skipping this if there's a couple of commits, but it really helps to get a summary of the intent for big PRs like this.

dvdhrm:
Can you rebase this on master? The `internal/boot` things are already merged, I think.

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+source /etc/os-release
+
+# Register RHEL if we are provided with a registration script.
+if [[ -n "${RHN_REGISTRATION_SCRIPT:-}" ]] && ! sudo subscription-manager status; then
+    sudo chmod +x $RHN_REGISTRATION_SCRIPT
+    sudo $RHN_REGISTRATION_SCRIPT
+fi
+
+# Restart systemd to work around some Fedora issues in cloud images.
+sudo systemctl restart systemd-journald
```
dvdhrm, line 40: Why? Can we drop this?

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Remove Fedora's modular repositories to speed up dnf.
+sudo rm -f /etc/yum.repos.d/fedora*modular*
```
dvdhrm, line 43: Does it? I would prefer to drop it.

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Enable fastestmirror and disable weak dependency installation to speed up
+# dnf operations.
+echo -e "fastestmirror=1\ninstall_weak_deps=0" | sudo tee -a /etc/dnf/dnf.conf
```
dvdhrm, line 47: Is it worth it? I would drop it, but I don't mind much either.

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Ensure we are using the latest dnf since early revisions of Fedora 31 had
+# some dnf repo priority bugs like BZ 1733582.
+# NOTE(mhayden): We can exclude kernel updates here to save time with dracut
+# and module updates. The system will not be rebooted in CI anyway, so a
+# kernel update is not needed.
+if [[ $ID == fedora ]]; then
+    sudo dnf -y upgrade --exclude kernel --exclude kernel-core
+fi
```
dvdhrm, line 56: Can we drop this now? Do we really need to support running on buggy systems?

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Add osbuild team ssh keys.
+cat schutzbot/team_ssh_keys.txt | tee -a ~/.ssh/authorized_keys > /dev/null
```
dvdhrm, line 59:
```suggestion

# Add osbuild team ssh keys.
cat schutzbot/team_ssh_keys.txt >>~/.ssh/authorized_keys
```

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+sudo cp osbuild-mock.repo /etc/yum.repos.d/osbuild-mock.repo
+sudo dnf repository-packages osbuild-mock list
+
+# Ensure osbuild-composer-tests is installed.
+retry sudo dnf -y install osbuild-composer-tests
+
+ls -l /usr/libexec/tests/osbuild-composer
+
+# Change to the working directory.
+cd $WORKING_DIRECTORY
```
dvdhrm, line 71: Belongs into quotes.

test/base-tests/run.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Run
+sudo --preserve-env $TEST_RUNNER /usr/libexec/tests/osbuild-composer/osbuild-weldr-tests
```
dvdhrm: I think you want this in quotes as well, but depends on what this is exactly used for.

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
+set -xeuo pipefail
+
+source /etc/os-release
+
+# Fedora 31's composer-cli doesn't support doing image uploads.
+if [[ $ID == fedora ]] && [[ $VERSION_ID == 31 ]]; then
+    echo "Fedora 31 does not support image uploads with composer-cli."
+    exit 0
```
dvdhrm, line 9: GNU autotools use "77" to skip tests. Is there a way to skip this test rather than make it succeed?

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+if [[ $ID == rhel ]]; then
+    sudo sed -r -i 's#toml.load\(args\[3\]\)#toml.load(open(args[3]))#' \
+        /usr/lib/python3.6/site-packages/composer/cli/compose.py
+    sudo rm -f /usr/lib/python3.6/site-packages/composer/cli/compose.pyc
+fi
+
+# We need jq for parsing composer-cli output.
+if ! hash jq; then
+    greenprint "Installing jq"
+    sudo dnf -qy install jq
```
dvdhrm, line 28: Why `-q` here?

dvdhrm: Also, why single out jq here? We have lots of dependencies, but we just check for jq? Can't we rely on the target image to contain it already?

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+AWS_CONFIG=${TEMPDIR}/aws.toml
+BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
+BLUEPRINT_NAME=targetimage
+COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
+COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
+
+# Get the compose log.
+get_compose_log () {
+    COMPOSE_ID=$1
+    LOG_FILE=${WORKSPACE}/osbuild-${ID}-${VERSION_ID}-aws.log
```
dvdhrm, line 51: I think you want both in quotes.

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    # Download the logs.
+    sudo composer-cli compose log $COMPOSE_ID | tee $LOG_FILE > /dev/null
```
dvdhrm, line 54: Again, variable-expansion belongs into quotes if you want it as a single argument.

dvdhrm: I might be misunderstanding something, but what is the repeated `tee FILE >/dev/null` all about? Why not redirect the original call?

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+}
+
+# Get the compose metadata.
+get_compose_metadata () {
+    COMPOSE_ID=$1
+    METADATA_FILE=${WORKSPACE}/osbuild-${ID}-${VERSION_ID}-aws.json
```
dvdhrm, line 60: Quotes?

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    # Download the metadata.
+    sudo composer-cli compose metadata $COMPOSE_ID > /dev/null
```
dvdhrm, line 63: Quotes

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    # Find the tarball and extract it.
+    TARBALL=$(basename $(find . -maxdepth 1 -type f -name "*-metadata.tar"))
+    tar -xf $TARBALL
+    rm -f $TARBALL
```
dvdhrm, line 68: I think all three lines want quotes. I will skip further comments on that, maybe you have an argument in favor of dropping them, but I think all those belong into quotes to make sure possible spaces are correctly handled.

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    # Move the JSON file into place.
+    cat ${COMPOSE_ID}.json | jq -M '.' | tee $METADATA_FILE > /dev/null
```
dvdhrm, line 71: `tee >/dev/null` again

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+# Prepare the blueprint for the compose.
+greenprint "📋 Preparing blueprint"
+cat $BLUEPRINT_FILE
+sudo composer-cli -j blueprints push $BLUEPRINT_FILE
+greenprint "📋 Depsolving blueprint"
+sudo composer-cli -j blueprints depsolve $BLUEPRINT_NAME
+
+# Get worker unit file so we can watch the journal.
+WORKER_UNIT=$(sudo systemctl list-units | egrep -o "osbuild.*worker.*\.service")
+sudo journalctl -af -n 1 -u ${WORKER_UNIT} &
```
dvdhrm, line 118: Umpf. This will cause random log entries in the output. Preferably the action that waits does this, but I guess that is not how composer-cli works.

test/base-tests/upload-to-aws.sh
```diff
@@ -0,0 +1,151 @@
...
+get_compose_log $COMPOSE_ID
+get_compose_metadata $COMPOSE_ID
+
+# Did the compose finish with success?
+if [[ $COMPOSE_STATUS != FINISHED ]]; then
+    echo "Something went wrong with the compose. 😢"
+    exit 1
+fi
+
+exit 0
```
dvdhrm, line 151: I think that can be dropped.

internal/boot/aws.go
```diff
@@ -151,17 +151,17 @@ func DescribeEC2Image(e *ec2.EC2, imageName string) (*imageDescription, error) {
 	}, nil
```
dvdhrm, line 179: Commit message says "not a functional change", but it kinda is, isn't it? But I get the point, and the intention stays the same, so... ooook :)

dvdhrm (changes requested):
This is **very** nice, I think. It will nicely abstract away the setup for our tests and make them run in a well-defined environment.

I have a couple of comments inline. There is some repeated issues with shell-quoting and `tee`, which I am not sure why you do them, but I tried commenting on all instances.

The last 2 commits could be squashed into the respective commits that introduce the respective files, I think. They do solve some of the things I noted in earlier commits:
    `base-test-runner: introduce cleanup mode`
    `base-tests/run: improve the output of the script`

Thanks!
David

cmd/osbuild-base-test-runner/main.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == 255 {
+				return &timeoutError{}
+			}
+		} else {
+			return fmt.Errorf("ssh command failed from unknown reason: %#v", err)
+		}
+	}
+	outputString := strings.TrimSpace(string(output))
```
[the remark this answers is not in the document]
msehnout: Well, first I get the `output` and `err` values, then I check for errors and if there are none I process the output. Can you think of a better place?

cmd/osbuild-base-test-runner/main.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	if creds == nil {
+		panic("Empty AWS credentials")
+	}
+	fmt.Println("Getting change and build IDs")
+	changeId := os.Getenv("CHANGE_ID")
+	buildId := os.Getenv("BUILD_ID")
+	if changeId == "" || buildId == "" {
+		panic("The environment variables must specify CHANGE_ID and BUILD_ID")
+	}
```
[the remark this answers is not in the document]
msehnout: Not really. Just that Jenkins provide these as environment variables, but I could as well pass them in the bash script running the binary.

cmd/osbuild-base-test-runner/main.go
```diff
@@ -0,0 +1,193 @@
...
+	fmt.Println("Getting the EC2 image description")
+	e, err := boot.NewEC2(creds)
+	if err != nil {
+		panic("Failed to obtain credentials for EC2")
+	}
+	imageDesc, err := boot.DescribeEC2Image(e, imageName)
+	if err != nil {
+		panic("Failed to describe EC2 image")
+	}
+	// delete the image after the test is over
```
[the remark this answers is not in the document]
msehnout: We have the cleanup mode for this and if that fails as well, we need to clean it manually :)

cmd/osbuild-base-test-runner/main.go
```diff
@@ -0,0 +1,193 @@
...
+			err = runSSHCommand(address, privateKey, "cat /etc/os-release")
+			if err != nil {
+				panic(err)
+			}
+			err = runSSHCommand(address, privateKey, "sudo chmod go+rw /run/weldr/api.socket")
+			if err != nil {
+				panic(err)
+			}
+			runLocalCommand("sudo", "mkdir", "/run/weldr")
+			runLocalCommand("sudo", "ssh", "-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-i", privateKey, "-fN", "-L", "/run/weldr/api.socket:/run/weldr/api.socket", fmt.Sprintf("redhat@%s", address))
```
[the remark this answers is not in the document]
msehnout: It is using `sudo` because it creates a socket in `/run/weldr`

test/base-tests/upload-to-aws.sh
msehnout: Let's not care about Fedora 31 any more. We'll soon be interested in f33 instead.

test/base-tests/upload-to-aws.sh
msehnout: I don't know, the definition of the images we use is somewhere in osbuild organization, but I lost track on where it lives. Maybe @major knows from top of his head?

test/base-tests/upload-to-aws.sh
msehnout: I don't know, most of this code is copy pasted because I prefer not to include other bash source files. But I wrote only few lines of this file ¯\_(ツ)_/¯

internal/boot/aws.go
msehnout: :) you are right

cmd/osbuild-base-test-runner/main.go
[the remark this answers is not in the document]
dvdhrm: I don't understand why you trim at all? The caller of this function specifies the bianry to execute, and they know what output to expect. They should be the ones to decide whether to trim or not, right? Trimming in the helper just prevents callers from getting the raw output, in case they ever need so.

I like trimming data at the latest possible stage, if needed at all. So that would mean when you compare the result. Otherwise, we end up trimming at all stages needlessly, or even harmfully dropping content.

cmd/osbuild-base-test-runner/main.go
```diff
@@ -0,0 +1,193 @@
...
+			panic(err)
+		}
+
+		time.Sleep(10 * time.Second)
+	}
+
+	panic(fmt.Sprintf("ssh test failure, %d attempts were made", attempts))
+}
+
+func main() {
```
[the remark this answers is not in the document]
dvdhrm: (FYI, your last commit introduced some grouping to that function, so that seems fine.)

cmd/osbuild-base-test-runner/main.go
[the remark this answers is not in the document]
dvdhrm: Right. We already do that in other parts of the CI, so I guess that is fine for now.

msehnout:
This work will be replaced by something similar, but much better in http://github.com/osbuild/rpmci ! Stay tuned for more AWS coming to your CI!

PR #961, opened 2020-09-03
tests: Add option to fail qemu boot. Fix #888

### LINKED ISSUE #888: Add option for image tests to fail if no cloud credentials are found
In CI we should already have appropriate credentials for AWS, Azure and OpenStack but based on this comment https://github.com/osbuild/osbuild-composer/pull/880#issuecomment-662911028 I see that the image tests for AWS and OpenStack still fall back to qemu. 

I propose a command line option which will cause the tests to fail instead of falling back to qemu. That would be enabled in CI and will help us ensure testing is done in the cloud while also revealing any changes around authentication and credentials.

@ondrejbudai any comments ?

atodorov:
as @teg suggests in
https://github.com/osbuild/osbuild-composer/issues/888#issuecomment-662942314
this is ON by default so we can be alerted for missing cloud
credentials in CI!

If you want to disable it then -fail-local-boot=off

### DIFF
--- a/cmd/osbuild-image-tests/main_test.go
+++ b/cmd/osbuild-image-tests/main_test.go
@@ -46,6 +46,7 @@ type testcaseStruct struct {
 }
 
 var disableLocalBoot = flag.Bool("disable-local-boot", false, "when this flag is given, no images are booted locally using qemu (this does not affect testing in clouds)")
+var failLocalBoot = flag.Bool("fail-local-boot", true, "when this flag is on (default), local boot will fail. Usually indicates missing cloud credentials")
 
 // GenerateCIArtifactName generates a new identifier for CI artifacts which is based
 // on environment variables specified by Jenkins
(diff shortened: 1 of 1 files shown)


atodorov:
> no AWS credentials given, falling back to booting using qemu

then fails with the new option

cmd/osbuild-image-tests/main_test.go
```diff
@@ -179,6 +180,15 @@ func testSSH(t *testing.T, address string, privateKey string, ns *boot.NetNS) {
...
 func testBootUsingQemu(t *testing.T, imagePath string) {
+	if *failLocalBoot {
+		t.Fatal("-fail-local-boot specified. Check missing cloud credentials!")
+	}
+
+	bootWithQemu(t, imagePath)
+}
+
+// will not fail even if -fail-local-boot is specified
+func bootWithQemu(t *testing.T, imagePath string) {
```
ondrejbudai, line 191: I think this function name is kinda confusing. I suggest moving the condition to `testBootUsingVMware` and the other cloud-specific functions. That makes more sense to me.

ondrejbudai (changes requested):
Good idea but I find the change confusing. See my comment inline.

atodorov:
@teg, @ondrejbudai ping?

ondrejbudai (approved):

Issue #951, opened 2020-08-28
url should be part of every source


msehnout:
Unless I'm missing something, the source specification is useless without URL. I suggest to add "url" to this check:
https://github.com/osbuild/osbuild-composer/blob/05fd221bd4763040afd9a17b8c6567dc9ef97409/internal/weldr/api.go#L596

Related: https://github.com/weldr/lorax/issues/1066

teg:
This sounds reasonable. How would this affect lorax compatibility?

bcl:
It shouldn't have any effect. url is assumed to be present. it's useless without it.

PR #968, opened 2020-09-04
weldr: make URL mandatory part of a new source

msehnout:
Sources without URL are useless. Make it mandatory.

Fixes: https://github.com/osbuild/osbuild-composer/issues/951


teg (approved):
This makes sense. Thanks!

PR #971, opened 2020-09-07
Further simplify worker API

larskarlitski:
See individual commits. I'm not yet done, but this is going towards a worker API that supports multiple job types.


larskarlitski:
Update: add a commit which changes the API to use callback URLs instead of predefined routes.

cmd/osbuild-worker/main.go
```diff
@@ -275,14 +276,14 @@ func main() {
 			// flag to indicate all error kinds.
 			result.Success = false
 		} else {
-			log.Printf("  🎉 Job completed successfully: %s", job.Id)
+			log.Printf("  🎉 Job completed successfully: %s", token)
```
teg: We probably don't want to print this in the logs, do we? These tokens are not meaningful to anyone else. This can be updated in a follow-up.

larskarlitski: The token is not sent anymore, but maybe I should print the `location` to make it visible?

Or even send the job id?

cmd/osbuild-worker/main.go
```diff
@@ -120,18 +120,19 @@ func RunJob(token uuid.UUID, manifest distro.Manifest, targets []*target.Target,
 				continue
 			}
 
-			if options.Key == "" {
-				options.Key = token.String()
+			key := options.Key
+			if key == "" {
+				key = uuid.New().String()
```
teg, line 128: We should move this into weldr and make the key mandatory in the worker. In the future we'll do the registering in composer and then we need to know the key. Can be fixed in a follow-up.

teg (approved):
Beautiful!

cmd/osbuild-worker/main.go
larskarlitski: I've added a commit that sends the job id again, for informational purposes only. Please re-review.

larskarlitski:
Update to set `Content-Type` correctly in the client and expose server errors in the worker's log.

larskarlitski:
Update: fix typos, clarify an error message and print errors at all when uploading fails. Also, fix an issue the tests found: the upload URL was wrong.

teg (approved):
Very nice!

teg:
One test failed due to a mirror issue, unrelated to this PR.

Issue #942, opened 2020-08-24
Resources are sometimes left behind in Azure


ondrejbudai:
This morning, @larskarlitski discovered that we had some stray resources laying around in Azure. I'm not sure why - maybe a Jenkins job got interrupted so the testing script cannot perform the clean-up. I think that if a job is canceled manually it's killed immediately, can you confirm that, @major?

msehnout:
I think we could mitigate this issue by having a "clean up" script in the `post {}` section in Jenkinsfile. If we could get rid of the randomness:
https://github.com/osbuild/osbuild-composer/blob/70c32ef7c53dd7ad10f2ea5953ee3aac65a2ea56/cmd/osbuild-image-tests/main_test.go#L267
and instead base the image name on the pipeline ID or something else which is specific to a single CI run, we could easily write a bash script to clean up everything that was left behind.

teg:
We could possibly also use tags for this? But predictable naming might be enough (and anyway a good idea).

major:
@ondrejbudai if i recall correctly, the `post {}` step gets run even if the job is canceled, so @msehnout's suggestion is a good one.

major:
And here's a good link to docs on the `post {}` section: https://www.jenkins.io/doc/book/pipeline/syntax/#post

There's a lot of granular control we have there.

PR #972, opened 2020-09-08
image-tests: don't use random names for CI artifacts

msehnout:
When using random names for artifacts like AWS snapshots, or Azure
images, it becomes hard to clean them up in case of CI failure. See this
issue for more details:
https://github.com/osbuild/osbuild-composer/issues/942

This PR introduces predictable names so that we can easily determine
which artifact belongs to which PR and therefore we can decide to wipe
all resources that are not needed any more.

### DIFF
--- a/internal/boot/helpers.go
+++ b/internal/boot/helpers.go
@@ -7,8 +7,6 @@ import (
 	"os"
 	"syscall"
 	"time"
-
-	"github.com/google/uuid"
 )
 
 // durationMin returns the smaller of two given durations
@@ -51,15 +49,4 @@ func killProcessCleanly(process *os.Process, timeout time.Duration) error {
 	}
 
 	return process.Kill()
-}
-
-// GenerateRandomString generates a new random string with specified prefix.
-// The random part is based on UUID.
-func GenerateRandomString(prefix string) (string, error) {
-	id, err := uuid.NewRandom()
-	if err != nil {
-		return "", err
-	}
-
-	return prefix + id.String(), nil
 }
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -133,6 +133,7 @@ pipeline {
                         AZURE_CREDS = credentials('azure')
                         OPENSTACK_CREDS = credentials("psi-openstack-creds")
                         VCENTER_CREDS = credentials('vmware-vcenter-credentials')
+                        DISTRO_CODE = "fedora31"
                     }
                     steps {
                         unstash 'fedora31'
@@ -182,6 +183,7 @@ pipeline {
                         AZURE_CREDS = credentials('azure')
                         OPENSTACK_CREDS = credentials("psi-openstack-creds")
                         VCENTER_CREDS = credentials('vmware-vcenter-credentials')
+                        DISTRO_CODE = "fedora32"
                     }
                     steps {
                         unstash 'fedora32'
@@ -246,6 +248,7 @@ pipeline {
                         OPENSTACK_CREDS = credentials("psi-openstack-creds")
                         RHN_REGISTRATION_SCRIPT = credentials('rhn-register-script-production')
                         VCENTER_CREDS = credentials('vmware-vcenter-credentials')
+                        DISTRO_CODE = "rhel8"
                     }
                     steps {
                         unstash 'rhel8cdn'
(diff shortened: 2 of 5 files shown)


internal/boot/helpers.go
```diff
@@ -63,3 +64,16 @@ func GenerateRandomString(prefix string) (string, error) {
 
 	return prefix + id.String(), nil
 }
+
+// GenerateCIArtifactName generates a new identifier for CI artifacts which is based
+// on environment variables specified by Jenkins
+// note: in case of migration to sth else like Github Actions, change it to whatever variables GH Action provides
+func GenerateCIArtifactName(prefix string) (string, error) {
+	changeId := os.Getenv("CHANGE_ID")
+	buildId := os.Getenv("BUILD_ID")
```
larskarlitski, line 73: I don't like reading environment variables from a library, because it makes it hard to figure out how a program is meant to be run. Is there a way to pass those through from `main_test.go`?

Don't block on it, especially since there seems to be a lot of such use here, and it's only test code.

schutzbot/run_image_tests.sh
```diff
@@ -62,7 +62,7 @@ run_test_case () {
 
     # Run the test and add the test name to the list of passed or failed
     # tests depending on the result.
-    if sudo $TEST_CMD 2>&1 | tee ${WORKSPACE}/${TEST_NAME}.log; then
+    if sudo --preserve-env $TEST_CMD 2>&1 | tee ${WORKSPACE}/${TEST_NAME}.log; then
```
larskarlitski, line 65: Is this necessary because of the additional environment variables that are read above? Why wasn't it necessary for the other vars?

larskarlitski (changes requested):
I really like this, thanks!

Please drop `GenerateRandomStrings` now. It doesn't seem to be used anymore.

internal/boot/helpers.go
msehnout: We could move the whole function out of the library into `main_test.go`, the only function which requires separate solutions is `WithBootedImageInEC2` because it also creates a security group, but we can pass the change/build id to it.

schutzbot/run_image_tests.sh
larskarlitski: Ah I see. Thanks!

msehnout:
@larskarlitski the PR is again ready for review. RHEL8.3 stages fail due to corrupted mock repo and I don't know what is wrong with codecov, but I can't see any errors in the UI.

larskarlitski (approved):
> @larskarlitski the PR is again ready for review. RHEL8.3 stages fail due to corrupted mock repo and I don't know what is wrong with codecov, but I can't see any errors in the UI.

Nice! Thanks.

The corrupt repo error seems to be gone from the latest log, but vhd-boot is failing. Is that unrelated?

I've restarted codecov. Seems to have failed running.

msehnout:
> The corrupt repo error seems to be gone from the latest log, but vhd-boot is failing. Is that unrelated?

This seems related, I'll check it.

schutzbot/Jenkinsfile
```diff
@@ -133,6 +133,7 @@ pipeline {
                         AZURE_CREDS = credentials('azure')
                         OPENSTACK_CREDS = credentials("psi-openstack-creds")
                         VCENTER_CREDS = credentials('vmware-vcenter-credentials')
+                        DISTRO_CODE = "fedora31"
```
ondrejbudai, line 136: I think you can reuse `ID` and `VERSION_ID`. See https://github.com/osbuild/osbuild-composer/blob/85423d403c031f9931568bb2525bcb2d4dd7eb94/test/image-tests/aws.sh#L7

ondrejbudai (commented):
Looks very nice, thanks! See my note inline.

schutzbot/Jenkinsfile
msehnout: I was afraid that it would clash for RHEL 8 in case we test against beta and nightly channels. This seemed to be a simple and safe workaround, but if you think there is nothing to worry about I can change it for sure.

msehnout:
> > The corrupt repo error seems to be gone from the latest log, but vhd-boot is failing. Is that unrelated?
> 
> This seems related, I'll check it.

I haven't found anything that could cause this yet. So I at least tried restarting the job. It passed this time, but failed with VMWare this time :(

schutzbot/Jenkinsfile
ondrejbudai: Right, let's merge this and worry about this later if we want to.

ondrejbudai (approved):
Very nice, thanks!

ondrejbudai:
For the record, VMWare was broken yesterday, this was an unrelated failure.

PR #974, opened 2020-09-08
tools/prepare-source: also call go generate ./...

larskarlitski:
To ensure that all generated code is up to date.


teg (approved):
Thanks!

larskarlitski:
A single test on Schutzbot ran into the timeout — everything else is green. This PR only touches a `tools/prepare-source`.

PR #977, opened 2020-09-10
🦛 Bump osbuild submodule to v21

major:


ondrejbudai (approved):
`groovy.lang.MissingPropertyException: No such property: none for class: groovy.lang.Binding` seems to be unrelated.

PR #975, opened 2020-09-09
azuretest: Refactor clean up code to separate functions

msehnout:
In order to run the clean up in a separate binary it is necessary to
refactor the clean up code away from the context manager.

PR #978, opened 2020-09-10
Introduce cloud cleaner

msehnout:
WIP, depends on:
#972 
#975

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -119,6 +119,7 @@ go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-weldr-te
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-image-tests %{goipath}/cmd/osbuild-image-tests
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-composer-cloud-tests %{goipath}/cmd/osbuild-composer-cloud-tests
 go test -c -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/osbuild-auth-tests %{goipath}/cmd/osbuild-auth-tests
+go build -tags=integration -ldflags="${TEST_LDFLAGS}" -o _bin/cloud-cleaner %{goipath}/cmd/cloud-cleaner
 
 %endif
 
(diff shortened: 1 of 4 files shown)


ondrejbudai:
This should be no longer blocked, can you rebase the PR, @msehnout?

osbuild-composer.spec
```diff
@@ -148,6 +149,7 @@ install -m 0755 -vp _bin/osbuild-tests                          %{buildroot}%{_l
 install -m 0755 -vp _bin/osbuild-weldr-tests                    %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 install -m 0755 -vp _bin/osbuild-dnf-json-tests                 %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 install -m 0755 -vp _bin/osbuild-image-tests                    %{buildroot}%{_libexecdir}/tests/osbuild-composer/
+install -m 0755 -vp _bin/cloud-cleaner                      %{buildroot}%{_libexecdir}/tests/osbuild-composer/
```
ondrejbudai, line 152: There's a whitespace issue.

ondrejbudai (changes requested):
I looked into the Azure Portal and we indeed need this thing (I need to delete 2 forgotten VMs manually).

I've found one whitespace issue and it needs one more rebase. Otherwise, I think this is a great start. Thanks, @msehnout!

osbuild-composer.spec
msehnout: Fixed :)

ondrejbudai: Hmm, I gave this another look and I think we should put the cloud-cleaner into `%{buildroot}%{_libexecdir}/osbuild-composer/`. I think in the future we want to be able to just iterate through the test libexec directory and run them all so we cannot put the cloud-cleaner there. Sorry!

msehnout: That is correct, but it also seems that the cloud cleaner cannot access any environment variables. Otherwise I cannot explain why the credentials structure is `nil`. Though I'm not sure why the environment variables are not available. I'm trying to read more about the "post" section, but it seems they should be available even in there.

msehnout:
The CI failure seems unrelated. If I understand correctly, it fails to upload to S3. Can you check this again @ondrejbudai ?

ondrejbudai (commented):
I like the cloud cleaner!

Mockbuild logs are also nice but I think we should merge `mockbuild-logs.sh` into `mockbuild.sh` because our Jenkinsfile grows bigger and bigger and it would be nice if we could rely on Jenkins specialties as low as possible. What do you think?

msehnout:
> I like the cloud cleaner!
> 
> Mockbuild logs are also nice but I think we should merge `mockbuild-logs.sh` into `mockbuild.sh` because our Jenkinsfile grows bigger and bigger and it would be nice if we could rely on Jenkins specialties as low as possible. What do you think?

On one hand yes, on the other hand it seems like good way to do it. Do you think other CI systems won't have such feature (like `post {}` section)? Anyway as this is not strictly related to this PR, I'd prefer to discuss it separately and fix in a following PR.

ondrejbudai:
I guess a nice solution is to merge only the first commit in this PR and open a new PR for the second one. Is it ok?

msehnout:
Again, only RHEL8.3 fails because of missing repository. This has already been fixed in #1027. Can I ask for one more review, @ondrejbudai ?

ondrejbudai (approved):
Thanks!

PR #982, opened 2020-09-14
distro: add fedora 33 support

jkozol:
Fedora 33 images can now be built. These are based off of what is installed in fedora 32.

According to image-info there are a couple of package differences between the qcow2 osbuild-composer generates and the official image found [here](https://www.happyassassin.net/nightlies.html). The official image is built from [this kickstarter](https://pagure.io/fedora-kickstarts/blob/master/f/fedora-cloud-base.ks).

The following packages are contained in our image but not the official image:
```
polkit-0.117-2.fc33.x86_64
polkit-libs-0.117-2.fc33.x86_64
polkit-pkla-compat-0.1-18.fc33.x86_64
geolite2-city-20191217-3.fc33.noarch
geolite2-country-20191217-3.fc33.noarch
linux-firmware-20200817-111.fc33.noarch
zram-generator-0.2.0-4.fc33.x86_64
zram-generator-defaults-0.2.0-4.fc33.noarch
```

I removed polkit from the list of packages and I have marked `geolite2-city`, `geolite2-country`, and `zram-generator-defaults` as excluded packages. This resolves the package differences except for `linux-firmware`. I am unsure how to remove the linux-firmware package. Any advice would be much appreciated.

I am opening this as a draft to get feedback before generating fedora 33 x86_64 test cases and updating schutzbot.


jkozol:
There are also differences in the `boot-environment` and `bootloader` options. 

For a qcow2 image we set `kernelOptions: "ro biosdevname=0 net.ifnames=0"`. This is not set for the official f33 image. 

The [official kickstarter](https://pagure.io/fedora-kickstarts/blob/master/f/fedora-cloud-base.ks#_35) also sets `bootloader --timeout=1 --append="no_timer_check net.ifnames=0 console=tty1 console=ttyS0,115200n8"`. Whereas the osbuild image generates with the options set to `ro rootflags=subvol=root rhgb quiet`

I am unsure which options we should set and where.

larskarlitski:
Very nice, thanks! For reference, here's the official [release schedule for 33](https://fedorapeople.org/groups/schedule/f-33/f-33-key-tasks.html). Also, the [list of high-level changes](https://fedoraproject.org/wiki/Releases/33/ChangeSet) might help guide some of the decisions we'll have to make. 

> I removed polkit from the list of packages and I have marked `geolite2-city`, `geolite2-country`, and `zram-generator-defaults` as excluded packages. This resolves the package differences except for `linux-firmware`. I am unsure how to remove the linux-firmware package. Any advice would be much appreciated.

We're not removing linux-firmware from fedora-32, are we? I don't think it is possible to remove cleanly — the kickstart you linked uses `rpm -e` directly, circumventing dnf.

> I am unsure which options we should set and where.

It probably makes sense to consult @gicmo on boot issues, but in general: the output image-info between the official images and the one osbuild-composer is producing should be minimal. Whatever the official one does, we do too :smile:

gicmo:
NB: grub changed the way kernel command line arguments are handled. Before F33, they were stored in `grubenv` and and then pulled in via a environment variable. On F33 they are directly in the BLS snippet, generated at RPM install time (relevant [change](https://src.fedoraproject.org/rpms/grub2/c/4a742183a39)). To generate the kernel command line the locations `/etc/kernel/cmdline`, ` /usr/lib/kernel/cmdline` and finally `/proc/cmdline` are read. This is what sd-boot is doing and we also do on s390x. Therefore the quickest fix is to have the `org.osbuild.kernel-cmdline` stage before the `org.osbuild.rpm` stage, like
```
 {
  "name": "org.osbuild.kernel-cmdline",
  "options": {
   "root_fs_uuid": "76a22bf4-f153-4541-b6c7-0332c0dfaeac",
   "kernel_opts": "net.ifnames=0 crashkernel=auto ..."
   }
 },
```
Ideally we should also change the `org.osbuild.grub2` stage to not optionally not write the kernel args to the grubenv, but that can be done later.

jkozol:
@gicmo Thanks for figuring this out. I added the stage and now the images are very close to being equivalent. There are still a few minor differences and I am not sure which should be fixed and which are fine as is. When I run image-info on the osbuild fedora 33 qcow and the official image, I get the following differences:

**osbuild** 
```
boot-environment": {
  "kernelopts": "root=UUID=76a22bf4-f153-4541-b6c7-0332c0dfaeac"
}
```
**official**
```
boot-environment": {
      "saved_entry": "dcac8d584c554b458d718c3025928093-5.8.6-301.fc33.x86_64"
}
```

Also, image-info shows a few extra fields for the **official image** that are not set in the generated image:

```
"rpm-verify": {
  "changed": {
    "/etc/machine-id": ".M.......",
    ...
  },
  "missing": [
    "/run/cloud-init",
    "/run/console-login-helper-messages",
    "/run/console-login-helper-messages/issue.d",
    "/run/console-login-helper-messages/motd.d",
    "/run/motd",
    "/run/motd.d",
    "/run/motd.d",
    "/run/setrans",
    "/var/run/console",
    "/var/run/faillock",
    "/var/run/sepermit"
  ]
},
"timezone": "UTC"
```

gicmo:
> I get the following differences:
> 
> **osbuild**
> 
> ```
> boot-environment": {
>   "kernelopts": "root=UUID=76a22bf4-f153-4541-b6c7-0332c0dfaeac"
> }
> ```
> 
> **official**
> 
> ```
> boot-environment": {
>       "saved_entry": "dcac8d584c554b458d718c3025928093-5.8.6-301.fc33.x86_64"
> }
> ```

We don't boot our images, so we dont have a saved entry. I think that is all fine.

jkozol:
@larskarlitski @teg do you have any thoughts on whether we should set a default timezone to UTC?

teg:
This might land me in bike shed territory, but i would say we do what the official image does, if it would be simple for us to do. I'm pretty sure this is a noop though and not setting a timezone has the same effect.

jkozol:
> This might land me in bike shed territory, but i would say we do what the official image does, if it would be simple for us to do. I'm pretty sure this is a noop though and not setting a timezone has the same effect.

So the timezone seems to default to UTC when we don't set one but there is no `/etc/localtime` in our image whereas there is one in the official image. I am happy to add the timezone. I think it is kind of strange to have set a default ourselves but if the official image is doing it then I guess we can too. @larskarlitski any additional thoughts on this?

Also, since the qcow2 is basically equivalent now I will add some tests and move on to the other image types.

larskarlitski:
> So the timezone seems to default to UTC when we don't set one but there is no `/etc/localtime` in our image whereas there is one in the official image. I am happy to add the timezone. I think it is kind of strange to set a default ourselves but if the official image is doing it then I guess we can too. @larskarlitski any additional thoughts on this?

Agreed. Let's do what the official image does.

> Also, since the qcow2 is basically equivalent now I will add some tests and move on to the other image types.

Nice work!

jkozol:
The timezone is now set and I have test cases for x86_64 qcow2

jkozol:
I added support for vmdk images. I will add individual commits for each image type and then squash the commits at the end.

jkozol:
I have added support for all images. I can squash/restructure the commits once looks good. I still need to generate tests but have been having issues with the package versions in the manifests not matching the package versions available on mirrors.kernel.org.

jkozol:
Commits are squashed, tests are added, and this is ready for review.

internal/distro/fedora33/distro_test.go
```diff
@@ -0,0 +1,344 @@
...
+		errorExpected bool
+	}{
+		{
+			name: "x86_64",
+		},
+		{
+			name: "aarch64",
+		},
+		{
+			name:          "foo-arch",
```
gicmo, line 319: Some extra whitespace here.

larskarlitski (approved):
Amazing work, and great commit message. Thanks a lot!

@teg I think this could go on like this, but we might want to swap for repo snapshots at some point, right?

gicmo (approved):
LGTM, paid attention mostly to OSTree/IoT changes.

internal/distro/fedora33/distro_test.go
jkozol: go fmt seems to want this so it aligns with the `errorExpected: true` field.

gicmo: oh i see, looked weird to my eyes. :+1:

ondrejbudai (approved):
I love this! Thanks so much!

PR #986, opened 2020-09-16
Expose the Koji API on the standard https port

teg:
See individual commits for details.


Gundersanne (dismissed):

larskarlitski (changes requested):
I agree that it should use standard ports, but unfortunately the tests need to be adjusted (predicted by @gicmo):
```[2020-09-16T17:56:30.176Z] Error: cannot listen on the TCP port: listen tcp4 :443: bind: address already in use
```
Port 443 of one of the containers (kojid?) is forwarded to localhost.

internal/upload/koji/run-koji-container.sh
```diff
@@ -81,7 +81,7 @@ koji_start() {
   ${CONTAINER_RUNTIME} run -d --name org.osbuild.koji.koji --network org.osbuild.koji \
     -v "${SHARE_DIR}:/share:z" \
     -p 80:80 \
-    -p 443:443 \
+    -p 443:4343 \
```
ondrejbudai, line 84: This is backwards, see https://docs.docker.com/config/containers/container-networking/

```suggestion
    -p 4343:443 \
```

ondrejbudai (changes requested):

internal/upload/koji/run-koji-container.sh
gicmo, line 84: I think it is a bit weird to have koji exposed on standard `80` but then the corresponding standard https port (`443`) be composer. Maybe also move koji to `8080` or something?

gicmo:
I like the idea! It will need a lot of changes in the koji-plugin test because they all assume that koji runs on the default http/https port.

internal/upload/koji/run-koji-container.sh
teg: Changed that too.

teg: Fixed!

ondrejbudai (approved):
Looks good to me. Let's wait for the CI though.

teg:
All koji tests passed, the others are unrelated.

PR #988, opened 2020-09-17
kojiapi: make repo gpgkey optional

### LINKED ISSUE #985: koji api: gpgkey is not required even though the openapi spec says it is
The subject says it all.

ondrejbudai:
The kojiapi shouldn't require it as we need to build images from
unsigned packages.

Fixes #985

### DIFF
--- a/internal/kojiapi/server.go
+++ b/internal/kojiapi/server.go
@@ -112,7 +112,9 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 		repositories := make([]rpmmd.RepoConfig, len(ir.Repositories))
 		for j, repo := range ir.Repositories {
 			repositories[j].BaseURL = repo.Baseurl
-			repositories[j].GPGKey = repo.Gpgkey
+			if repo.Gpgkey != nil {
+				repositories[j].GPGKey = *repo.Gpgkey
+			}
 		}
 		bp := &blueprint.Blueprint{}
 		err = bp.Initialize()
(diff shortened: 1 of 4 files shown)


teg (approved):
Thanks!

teg:
This is looking good, but needs a rebase.

internal/kojiapi/server.go
```diff
@@ -106,7 +106,7 @@ func (server *Server) PostCompose(w http.ResponseWriter, r *http.Request) {
 		repositories := make([]rpmmd.RepoConfig, len(ir.Repositories))
 		for j, repo := range ir.Repositories {
 			repositories[j].BaseURL = repo.Baseurl
-			repositories[j].GPGKey = repo.Gpgkey
+			repositories[j].GPGKey = *repo.Gpgkey
```
Gundersanne, line 109: Hm this might be nil now, no? Not sure what happens in go when you deref nil.

ondrejbudai: You are right, thanks! Fixed.

ondrejbudai:
Rebased and modified a test - now we build images both with a gpg key and without one.

teg (approved):
Nice! But requires rebase.

ondrejbudai:
Rebased

larskarlitski (approved):
Nice! Thanks :)

ondrejbudai:
ostree failure is fixed in #1027. F32 tests failed due to flakey mirrors.

PR #992, opened 2020-09-17
kojiapi: move to echo

teg:
Unify with the worker API and move from the chi to the echo libraries.


internal/kojiapi/api/api.go
```diff
@@ -0,0 +1,3 @@
+//go:generate go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen -package=api -generate types,server -o api.gen.go openapi.yml
```
Gundersanne, line 1: Hm `client` was remote from -generate here, do we not use it?

internal/kojiapi/server.go
```diff
@@ -24,66 +23,70 @@ import (
...
-		return
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid Koji server: %s", request.Koji.Server)
 	}
-	creds, exists := server.kojiServers[kojiServer.Hostname()]
+	creds, exists := h.server.kojiServers[kojiServer.Hostname()]
 	if !exists {
-		http.Error(w, fmt.Sprintf("Koji server has not been configured: %s", kojiServer.Hostname()), http.StatusBadRequest)
-		return
-
+		echo.NewHTTPError(http.StatusBadRequest, "Koji server has not been configured: %s", kojiServer.Hostname())
```
Gundersanne, line 89: missing return statement?

internal/kojiapi/api/api.go
larskarlitski: If we don't use it, I recommend not generating it. The client code generated by `deepmap` is all over the place. I've opted doing it manually in the `worker` package as well.

teg: The code that used the client api was dropped (it was an unused test). Generating it still was a left-over.

internal/kojiapi/server.go
teg: Fixed.

teg:
Test failures are due to #997, and unrelated to this PR.

internal/kojiapi/server.go
```diff
@@ -23,66 +23,70 @@ import (
...
-		return
+		return err
 	}
 
-	d := server.distros.GetDistro(request.Distribution)
+	d := h.server.distros.GetDistro(request.Distribution)
 	if d == nil {
-		http.Error(w, fmt.Sprintf("Unsupported distribution: %s", request.Distribution), http.StatusBadRequest)
-		return
+		return echo.NewHTTPError(http.StatusBadRequest, "Unsupported distribution: %s", request.Distribution)
```
larskarlitski, line 80: Unfortunately, this function does not take format strings (cf [echo.NewHTTPError](https://pkg.go.dev/github.com/labstack/echo/v4#NewHTTPError). I realized this too late myself when working on the worker API and fixed it in https://github.com/osbuild/osbuild-composer/commit/ccebfb014c4bb53870d1c876ab4868c936d5178a.

larskarlitski (changes requested):
Thanks! The failure is unrelated indeed (sad that we're having to say this more often again...)

Super sorry: I overlooked something in the first round. Comment inline.

larskarlitski (approved):
Force-pushed a fix for my latest nitpick myself. Was an easy one.

@teg, please have a look at these changes before merging. Otherwise it's all good. Thanks again!

PR #995, opened 2020-09-18
📜 mockbuild: Reduce log output

major:
Running mock with verbose enabled made sense a while back when we were
still unsure if things would work, but it's generating a ton of logs now
that eventually clogs Jenkins' disk.

Disable the verbose flag to reduce mock's log output.


ondrejbudai (approved):

major:
The mock build went through fine so I stopped the rest of the job so we could free up resources in Jenkins sooner.

Issue #997, opened 2020-09-19
EL8.3 ostree test fails with a wrong checksum


ondrejbudai:
Example here: https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/master/244/pipeline/162

```
Running job 76d62b4c-5d05-44dc-8cbf-74cec1dd6bab
sha256sum: WARNING: 1 computed checksum did NOT match
Job failed: running osbuild failed: exit status 1
Waiting for a new job...
```

teg:
These tests are run against nightly, which changes under our feet. We should update them to run against snapshots from https://github.com/osbuild/manifestdb/tree/main/rpmrepos/by-name
Thoughts on this @henrywang and @dvdhrm?

henrywang:
@teg, the [rhel-8-beta.json](https://github.com/osbuild/osbuild-composer/blob/master/test/image-tests/rhel-8-beta.json) we are using right now is the same as what https://github.com/osbuild/manifestdb/tree/main/rpmrepos/by-name referd to. I have no idea the failed reason.

teg:
Did some digging, and this turned out to be quite nasty.

tl;dr: our rpm snapshots are broken, but a fix has been identified and is in the works.

If you look in the Jenkins artifacts for the ostree log in question you'll see this line:
```
RuntimeError: org.osbuild.files: checksum mismatch: sha256:741da77a8af0cd127ee6a3d65cd6e1e0a1e38b58eedb9ddb89aa67f39143b63b {'url': 'https://rpmrepo.osbuild.org/psi/el8/el8-20200901-x86_64-baseos-8.3.0.n-2bfaa8f1e249f5d937e42b21d9eb63f1efe26d60ccaf20518ad0ae7f89e6fe53/Packages/shim-x64-15-15.el8.x86_64.rpm'}
```
And, indeed, the yum metadata indicates a different checksum for the given NEVRA than the RPM we provide.

Turns out, NEVRA's are required to match the checksum of the payload of an RPM (not including signature headers), but not actual file checksums (which is what is recorded in the yum metadata). Two snapshots were taken, containing the same NEVRA, but where the checksum of the RPM differed. The reason was that one version was signed and the other was not.

This turned out to be a problem, as we deduplicate our snapshots based on NEVRA, which breaks down when we mix signed and unsigned RPMs.

[...]

PR #998, opened 2020-09-20
worker API: introduce job types and architecture

larskarlitski:
This is an effort to future-proof the worker API. It removes the assumption that all jobs are osbuild jobs and requires a worker to pass both the jobs it can handle and the architecture it runs on.

### DIFF
--- a/cmd/osbuild-composer-cloud/main.go
+++ b/cmd/osbuild-composer-cloud/main.go
@@ -106,17 +106,31 @@ func main() {
 		log.Fatalf("cannot create queue directory: %v", err)
 	}
 
-	jobs, err := fsjobqueue.New(queueDir, []string{"osbuild"})
+	distros, err := distro.NewRegistry(fedora31.New(), fedora32.New(), rhel8.New())
+	if err != nil {
+		log.Fatalf("Error loading distros: %v", err)
+	}
+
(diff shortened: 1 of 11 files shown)


teg (dismissed):
Perfect!

larskarlitski:
Update to resolve the conflict and set a correct list of accepted job types.

teg (dismissed):
Thanks! (CI failures are, again, unrelated and caused by the ostree tests).

larskarlitski:
> Thanks! (CI failures are, again, unrelated and caused by the ostree tests).

There was a failure in `aws.sh` as well, which might have been a flake. I restarted the tests to make sure.

larskarlitski:
> There was a failure in `aws.sh` as well, which might have been a flake. I restarted the tests to make sure.

This was an actual bug (yay testing and thanks for tracking it down @ondrejbudai). I've copied the logic to assemble job types from `osbuild-composer` to `osbuild-composer-cloud`. I didn't bother sharing the code because we're about to merge the two.

larskarlitski:
That last fix did it indeed. The only remaining failure is due to https://github.com/osbuild/osbuild-composer/issues/997.

teg (approved):
Nice!

PR #999, opened 2020-09-21
Add domain allowlist to remote worker and Koji APIs

ondrejbudai:
This PR adds:

- list of allowed domains for both Koji and remote worker APIs. After this change, the server accepts only certificates issued to pre-defined domains.
- tests for this.
- By default, koji and remote worker APIs now use the host's trusted CAs to validate client certificates.
  
  This is configurable in `osbuild-composer.toml` by `koji.ca` and `worker.ca` fields. These accept a path to a certificate that should be used to the validation instead of the default ones. If there's a need for using multiple certificates, it's possible to just append them to one file. (`cat cert1.pem cert2.pem > /etc/osbuild-composer/ca-crt.pem`)

What's still missing:

- wildcards for allowed domains (nice to have)
- SAN certificates for tests. Currently, only certificates with common names are used. This just requires a bit of tinkering with `openssl`.

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -72,11 +72,13 @@ func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
 
 func main() {
 	var config struct {
-		KojiServers map[string]struct {
-			Kerberos *struct {
-				Principal string `toml:"principal"`
-				KeyTab    string `toml:"keytab"`
-			} `toml:"kerberos,omitempty"`
+		Koji *struct {
+			Servers map[string]struct {
+				Kerberos *struct {
+					Principal string `toml:"principal"`
+					KeyTab    string `toml:"keytab"`
+				} `toml:"kerberos,omitempty"`
+			} `toml:"servers"`
 		} `toml:"koji"`
 		Worker *struct {
 			AllowedDomains []string `toml:"allowed_domains"`
@@ -190,8 +192,11 @@ func main() {
 
 	// Optionally run Koji API
 	if kojiListeners, exists := listeners["osbuild-composer-koji.socket"]; exists {
+		if config.Koji == nil {
+			log.Fatal("koji not configured in the config file")
+		}
 		kojiServers := make(map[string]koji.GSSAPICredentials)
-		for server, creds := range config.KojiServers {
+		for server, creds := range config.Koji.Servers {
 			if creds.Kerberos == nil {
 				// For now we only support Kerberos authentication.
 				continue
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -79,6 +79,7 @@ func main() {
 					KeyTab    string `toml:"keytab"`
 				} `toml:"kerberos,omitempty"`
 			} `toml:"servers"`
+			AllowedDomains []string `toml:"allowed_domains"`
 		} `toml:"koji"`
 		Worker *struct {
 			AllowedDomains []string `toml:"allowed_domains"`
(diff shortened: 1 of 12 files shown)


cmd/osbuild-composer/main.go
```diff
@@ -55,6 +59,18 @@ func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
 		Certificates: []tls.Certificate{cert},
 		ClientAuth:   tls.RequireAndVerifyClientCert,
 		ClientCAs:    roots,
+		VerifyPeerCertificate: func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+			for _, chain := range verifiedChains {
+				for _, domainGlob := range c.DomainAllowlist {
+					commonName := chain[0].Subject.CommonName
```
teg, line 65: Should we match on SAN insetad? Isn't CN deprecated? @larskarlitski @gicmo ?

cmd/osbuild-composer/main.go
```diff
@@ -55,6 +59,18 @@ func createTLSConfig(c *connectionConfig) (*tls.Config, error) {
...
+					if glob.MustCompile(domainGlob).Match(commonName) {
```
teg, line 66: Should we pre-compile this when loading the config? My main concern is that this would panic at runtime if the config file is wrong, would be better to error out early.

cmd/osbuild-composer/main.go
```diff
@@ -66,6 +82,9 @@ func main() {
 				KeyTab    string `toml:"keytab"`
 			} `toml:"kerberos,omitempty"`
 		} `toml:"koji"`
+		RemoteWorker *struct {
+			DomainAllowlist []string `toml:"domain_allowlist"`
+		} `toml:"remote_worker,omitempty"`
```
teg, line 90: Name it `worker` for simplicity. The non-remote case is basically legacy and should be going away (if we need config for that we could always call that `local_worker`).

cmd/osbuild-composer/main.go
```diff
@@ -66,6 +82,9 @@ func main() {
 				KeyTab    string `toml:"keytab"`
 			} `toml:"kerberos,omitempty"`
 		} `toml:"koji"`
+		RemoteWorker *struct {
+			DomainAllowlist []string `toml:"domain_allowlist"`
```
teg, line 89: Just call this `domains`?

teg (changes requested):
Looking good, minor nits. Main point is CN v SAN.

cmd/osbuild-composer/main.go
larskarlitski: It has been deprecated for > 10 years, but still works. @gicmo was seeing some warnings in python client libraries. Probably makes sense to switch or support both. I wouldn't block on it.

gicmo: This is the mentioned python library issue: https://github.com/urllib3/urllib3/issues/497, for background information.

cmd/osbuild-composer/main.go
ondrejbudai: `allowedDomains`? I like more context...

cmd/osbuild-composer/main.go
ondrejbudai: Done!

cmd/osbuild-composer/main.go
ondrejbudai: I used https://golang.org/pkg/crypto/x509/#Certificate.VerifyHostname in the end. Thanks, @gicmo, for the suggestion!

cmd/osbuild-composer/main.go
ondrejbudai: This code was dropped.

test/image-tests/osbuild-composer.toml
```diff
@@ -1,3 +1,6 @@
+[koji]
+allowed_domains = [ "localhost", "*.osbuild.org" ]
```
teg: Wildcard?

teg (approved):
Great stuff! Left some comments, but nothing worth rebasing for.

teg:
#997 is still hitting us, but otherwise this is good.

PR #1003, opened 2020-09-22
tests: Make run_image_tests.sh fail on ls errors. Fix #1002

### LINKED ISSUE #1002: ls: cannot access 'rhel_8-x86_64*.json': No such file or directory doesn't fail  run_image_tests.sh
From this execution:
https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-961/2/pipeline/160

```
[2020-09-21T11:29:54.415Z] + schutzbot/run_image_tests.sh

[2020-09-21T11:29:54.415Z] ls: cannot access 'rhel_8-x86_64*.json': No such file or directory

[2020-09-21T11:29:54.415Z] ------------------------------------------------------------------------------

[2020-09-21T11:29:54.415Z] 😃 Passed tests: 

[2020-09-21T11:29:54.415Z] ☹ Failed tests: 

[2020-09-21T11:29:54.415Z] ------------------------------------------------------------------------------

[2020-09-21T11:29:54.415Z] 🎉 All tests passed.
```

Which reports PASS and IMO should clearly be a FAIL. 

Not only is this wrong but it hides the actual results from #961 which is supposed to fail b/c previous test jobs (osbuild-image-tests) were reporting that no AWS Credentials were found.

atodorov:
man set says:

... The shell does not exit if the command that fails is part
of  the  command list  immediately  following a while or until
keyword ...

It doesn't say anything about `for` but it looks like it's the
same.

### DIFF
--- a/schutzbot/run_base_tests.sh
+++ b/schutzbot/run_base_tests.sh
@@ -53,8 +53,8 @@ echo "😃 Passed tests:" "${PASSED_TESTS[@]}"
 echo "☹ Failed tests:" "${FAILED_TESTS[@]}"
 test_divider
 
-# Exit with a failure if any tests failed.
-if [ ${#FAILED_TESTS[@]} -eq 0 ]; then
+# Exit with a failure if tests were executed and any of them failed.
+if [ ${#PASSED_TESTS[@]} -gt 0 ] && [ ${#FAILED_TESTS[@]} -eq 0 ]; then
     echo "🎉 All tests passed."
     exit 0
 else
--- a/schutzbot/run_image_tests.sh
+++ b/schutzbot/run_image_tests.sh
@@ -91,8 +91,8 @@ echo "😃 Passed tests: " "${PASSED_TESTS[@]}"
 echo "☹ Failed tests: " "${FAILED_TESTS[@]}"
 test_divider
 
-# Exit with a failure if any tests failed.
-if [ ${#FAILED_TESTS[@]} -eq 0 ]; then
+# Exit with a failure if tests were executed and any of them failed.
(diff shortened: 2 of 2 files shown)


ondrejbudai:
This doesn't make the tests fail. The other issue is in the command substitution. See https://stackoverflow.com/questions/58859554/how-to-use-set-e-inside-a-bash-command-substitution

ondrejbudai (changes requested):
This doesn't make the tests fail. The other issue is in the command substitution. See https://stackoverflow.com/questions/58859554/how-to-use-set-e-inside-a-bash-command-substitution

atodorov:
> This doesn't make the tests fail.

Right. When I tested it locally with a simplified version of this script I got a non-zero exit code but not in Jenkins. Not sure why. Here's my test script for reference:

```
#/bin/bash
set -euo pipefail

get_test_cases () {
    ls /tmp/non-existing
}

CASES=$(get_test_cases)
for TEST_CASE in $CASES; do
    echo "Running $TEST_CASE"
done
```

> The other issue is in the command substitution. See https://stackoverflow.com/questions/58859554/how-to-use-set-e-inside-a-bash-command-substitution

I see lots of info here, not sure what are you pointing me at exactly.

teg (approved):
Looks good.

atodorov:
> Looks good.

I can't merge with failing checks. #1005 fixes the failures exposed here. The ostree test failure is unrelated.

ondrejbudai:
Superseded by #1005

PR #1004, opened 2020-09-23
api/worker, koji: fix race condition when using multiple listeners

ondrejbudai:
When remote worker socket was enabled, this was happening:

```
e := echo.New()

go func() {
    e.Listener = listener1
    e.Start("")
}()

e.Listener = listener2
e.Start("")
```

Yeah, this is a race condition. None of the echo's Start methods cannot safely
handle multiple listeners.

This commit fixes this issue by using Echo only as a router for standard
`http.Server` which handles multiple listeners in a non-racy way.

### DIFF
--- a/internal/kojiapi/server.go
+++ b/internal/kojiapi/server.go
@@ -24,7 +24,7 @@ import (
 
 // Server represents the state of the koji Server
 type Server struct {
-	echo        *echo.Echo
+	server      *http.Server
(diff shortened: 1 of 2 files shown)


larskarlitski (approved):
Thanks! And sorry for introducing this race in the first place.

ondrejbudai:
#997, otherwise good.

PR #1005, opened 2020-09-23
tests/image: fix the selector

ondrejbudai:
**Currently, the image tests are not running!**

They're only reporting `ls: cannot access 'rhel_8-x86_64*.json': No such file or directory` and ending with success immediately. #1002 and #1003 tried (so far unsuccessfully) to fail the test if `ls` fails. This PR instead fixes the cause[1]. Nevertheless, we should still implement a better error handling if something like this ever happens again.

[1]Using a wildcard in quotes doesn't trigger shell expansion, therefore this PR moves the wildcard out of the quotes

### DIFF
--- a/schutzbot/run_base_tests.sh
+++ b/schutzbot/run_base_tests.sh
@@ -54,8 +54,8 @@ echo "😃 Passed tests:" "${PASSED_TESTS[@]}"
 echo "☹ Failed tests:" "${FAILED_TESTS[@]}"
 test_divider
 
-# Exit with a failure if any tests failed.
-if [ ${#FAILED_TESTS[@]} -eq 0 ]; then
+# Exit with a failure if tests were executed and any of them failed.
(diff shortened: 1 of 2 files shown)


teg (approved):

teg:
Turns out there are other bugs (probably shellcheck fallout) revealed by this. I guess this should all be fixed in the same PR.

atodorov:
@teg can you hold off on merging this so can have a good reproducer for #1002

teg (dismissed):
Thanks for this!

atodorov:
@ondrejbudai can you cherry-pick my commit from #1003 here so we can merge with minimum failures ?

ondrejbudai:
#1003 cherry-picked. The image tests are now failing because we don't have AWS credentials in Jenkins. (see #961 for an explanation of why the tests without credentials are failing).

ondrejbudai:
Superseded by #1012

PR #1009, opened 2020-09-25
Start osbuild-composer-koji.socket only if installed

atodorov:
Tom says this shouldn't be started unconditionally because most
of the tests don't need it. In fact it isn't even available in
RHEL trees.


ondrejbudai (approved):
Thanks!

PR #1010, opened 2020-09-27
osbuild-composer: cleanup main.go

larskarlitski:
`main.go` was quite unwieldy and hard to see what's going on. This series of patches try to clean it up a bit, mainly by splitting running the actual service into its own file.

This has the additional benefit that it is a step towards running osbuild-composer in an environment different than what systemd sets up (:wave: @major).

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -1,0 +1,268 @@
+package main
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"path"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/jobqueue/fsjobqueue"
+	"github.com/osbuild/osbuild-composer/internal/kojiapi"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"github.com/osbuild/osbuild-composer/internal/store"
+	"github.com/osbuild/osbuild-composer/internal/upload/koji"
+	"github.com/osbuild/osbuild-composer/internal/weldr"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+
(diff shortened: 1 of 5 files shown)


major:
> This has the additional benefit that it is a step towards running osbuild-composer in an environment different than what systemd sets up ( @major).

@larskarlitski  🥰 😍 🤗

cmd/osbuild-composer/composer.go
```diff
@@ -0,0 +1,268 @@
...
+
+	c.kojiListener = tls.NewListener(l, tlsConfig)
+
+	return nil
+}
+
+func (c *Composer) InitRemoteWorkers(cert, key string, l net.Listener) error {
+	tlsConfig, err := createTLSConfig(&connectionConfig{
+		CACertFile:     c.config.Worker.CA,
+		ServerKeyFile:  cert,
```
ondrejbudai, line 160: As you probably already know, these lines are in the wrong order. :upside_down_face:

larskarlitski: Indeed I had seen this yesterday, but forgot to actually push the update. Thanks!

ondrejbudai (approved):
I like this! Thanks!

cmd/osbuild-composer/composer.go
```diff
@@ -0,0 +1,268 @@
...
+	rpm rpmmd.RPMMD
+
+	workers *worker.Server
+	weldr   *weldr.API
+	koji    *kojiapi.Server
+
+	weldrListener, localWorkerListener, workerListener, kojiListener net.Listener
+}
+
+func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *log.Logger) (*Composer, error) {
```
teg, line 45: The canonical name would just be `New()`, wouldn't it?

teg (approved):
Very nice! Comment not a blocker.

teg:
CI failures are unrelated.

cmd/osbuild-composer/composer.go
larskarlitski: It would, but this is not a package. Having a blank `New(...)` in `main.go` seemed confusing to me.

PR #1014, opened 2020-10-01
schutzbot: enable F33 testing

ondrejbudai:
## Update as of 2020-11-11
This is now done, see today's comment.

## Done so far
- :heavy_check_mark: Fedora 33 beta images are available in both PSI and AWS
- :heavy_check_mark: Jenkins could not connect to F33 instances. The cause is not entirely clear but it's probably connected to changes in [crypto](https://fedoraproject.org/wiki/Changes/StrongCryptoSettings2). Major was able to find a workaround. (set crypto policy to LEGACY)
- :heavy_check_mark: `kolo/xmlrpc` was rebased in F33 and has a slightly different API. This PR ships a somewhat ugly workaround, see individual commits for more info.
- :heavy_check_mark: `s3cmd` used in `mockbuild.sh` does not work on F33 because of changes in Python 3.9. I've sent a [PR upstream](https://github.com/s3tools/s3cmd/pull/1137) and a [bug downstream](https://bugzilla.redhat.com/show_bug.cgi?id=1884607). This PR applies the same patch to `s3cmd` as is in the upstream PR.
- :heavy_check_mark: F33 cloud base image doesn't have `openssl` installed so I have to install it manually so we can generate TLS keys for remote worker/cloud/koji API.

## Needs work
- :no_entry_sign: auth tests are broken because they rely on common names in TLS certificates that are deprecated in Go 1.15. This is just an issue in the test, actual implementation should be good.
- :no_entry_sign: image tests need updated cases because the base repo is still moving. We should use a snapshot.
- :no_entry_sign: koji test is broken. Not sure why, looks like an issue with importing schema into the database.
- :no_entry_sign: `test/image-tests/ostree.sh: line 167: IMAGE_TYPE: unbound variable` - no idea, I haven't had time to investigate yet.


dvdhrm:
FYI: F33 snapshots are now available. First one is:

    https://rpmrepo.osbuild.org/v1/anon/f33/f33-x86_64-devel-20201006/repodata/repomd.xml

I will make another one the coming weekend. You can enumerate the snapshots here:

    https://rpmrepo.osbuild.org/control/snapshots

If you need another snapshot earlier than next Monday, lemme know!

ondrejbudai:
After 42 days, this is finally finished and green! This is just the tip of the iceberg, many stuff needed to properly support Fedora 33 testing was merged earlier:

- #1035
- #1057
- #1068
- #1071 

This PR does basically 3 things:

1) Adds support for F33 to Ostree tests
2) Fixes the test case generator, updates the repos used by it and regenerates the test cases
3) Enables F33 testing on Schutzbot

But there's one more thing!

Qemu tests downloaded the image to `/tmp` that doesn't have enough space on F33 (possibly connected with the ZRAM change?\[1\]. I modified the test to use `/var/lib/osbuild-composer-tests/` for this purpose. This is btw the same dir as image-tests use for temporary data.

\[1\]: https://fedoraproject.org/wiki/Changes/SwapOnZRAM

schutzbot/Jenkinsfile
```diff
@@ -165,6 +165,69 @@ pipeline {
...
+                        VCENTER_CREDS = credentials('vmware-vcenter-credentials')
+                        DISTRO_CODE = "fedora33"
+                    }
+                    steps {
+                        unstash 'fedora33'
+                        run_tests('image')
+                    }
+                    post {
+                        always {
+                            preserve_logs('fedora33-image')
```
msehnout, line 198:
```suggestion
                            preserve_logs('fedora33-image')
                            sh (
                                label: "Run cloud cleaner just in case something failed",
                                script: "schutzbot/run_cloud_cleaner.sh"
                            )
```

test/cases/qemu.sh
```diff
@@ -205,10 +205,18 @@ sudo kill ${WORKER_JOURNAL_PID}
...
 greenprint "📥 Downloading the image"
-sudo composer-cli compose image "${COMPOSE_ID}" > /dev/null
-IMAGE_FILENAME=$(basename "$(find . -maxdepth 1 -type f -name "*.${IMAGE_EXTENSION}")")
-LIBVIRT_IMAGE_PATH=/var/lib/libvirt/images/${IMAGE_KEY}.${IMAGE_EXTENSION}
-sudo mv "$IMAGE_FILENAME" "$LIBVIRT_IMAGE_PATH"
+
+# Current $PWD is inside /tmp, there may not be enough space for an image.
+# Let's use a bigger temporary directory for this operation.
+BIG_TEMP_DIR=/var/lib/osbuild-composer-tests
+sudo rm -rf "${BIG_TEMP_DIR}" || true
```
msehnout, line 212: Why the `|| true` part? Shouldn't it be fine with non-existing or empty directory because of `-f`?

msehnout (commented):
I haven't checked the test cases properly, they are rather long, but otherwise this looks very good!

test/cases/qemu.sh
ondrejbudai: Ah, right, didn't know about this `-f`'s behavior.

schutzbot/Jenkinsfile
ondrejbudai: Indeed, thanks! I added those lines before the cloud cleaner was introduced.

msehnout (approved):
My previous comments are not blockers and CI is all green. Thank you @ondrejbudai !

PR #1015, opened 2020-10-02
tests: Conditionally enable osbuild-composer-cloud.socket

atodorov:


ondrejbudai (changes requested):
This change makes sense, shellcheck doesn't like it though.

atodorov:
The proposed changes are already in master. Closing.

PR #1016, opened 2020-10-03
Test cleanup

teg:
Make integration tests more uniform: ship all tests in the -tests RPM, rather than run some from git.

Without changing the tests themselves the aim is to make them easier to run locally / outside schutzbot. This PR only addresses the most low hanging fruit, but hopefully it moves things in the right direction.

PR #1017, opened 2020-10-04
Merge cloud API into main binary

larskarlitski:
As discussed, we'd like to expose one endpoint on `:443` with both the composer API (née "cloud", at `/api/composer/v1`) and the koji api (at `/api/composer-koji/v1`).

This is a draft, because it doesn't contain an integration test yet, because I don't want too many conflicts with #1016.


distribution/osbuild-composer-api.socket
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=OSBuild Composer Koji API socket
```
ondrejbudai, line 2:
```suggestion
Description=OSBuild Composer API socket
```

ondrejbudai (commented):
Looks alright to me so far!

distribution/osbuild-composer-api.socket
larskarlitski: Thanks!

larskarlitski:
> This is a draft, because it doesn't contain an integration test yet, because I don't want too many conflicts with #1016.

#1016 has landed and I've included an integration test that depends on it, based on the one @Gundersanne had put into `aws.sh`.

This is ready for review now.

larskarlitski:
Update: fix `osbuild-auth-tests` and change `api.sh` to make a RHEL image, because `fedora-32` cannot be built on older systems because of https://bugzilla.redhat.com/show_bug.cgi?id=1777831.

test/cmd/api.sh
```diff
@@ -0,0 +1,158 @@
...
+cat > "$REQUEST_FILE" << EOF
+{
+  "distribution": "rhel-8",
+  "image_requests": [
+    {
+      "architecture": "x86_64",
+      "image_type": "ami",
+      "repositories": [
+        {
+          "baseurl": "http://download.eng.rdu.redhat.com/rel-eng/rhel-8/RHEL-8/RHEL-8.3.0-Beta-1.0/compose/BaseOS/x86_64/os",
```
atodorov, line 72: Shouldn't these URLs be pointed to the rel-end repos? See #1027

test/cmd/api.sh
```diff
@@ -0,0 +1,158 @@
...
+}
+trap cleanup EXIT
+
+
+#
+# Install the aws client from the upstream release, because it doesn't seem to
+# be available as a RHEL package.
+#
+
+if ! hash aws; then
```
ondrejbudai, line 42: What about awscli from EPEL?

cmd/osbuild-composer/composer.go
```diff
@@ -197,9 +201,25 @@ func (c *Composer) Start() error {
 		}()
 	}
 
-	if c.kojiListener != nil {
+	if c.apiListener != nil {
 		go func() {
-			err := c.koji.Serve(c.kojiListener)
+			const apiRoute = "/api/composer/v1"
```
Gundersanne, line 206: ideally the version would come from the openapi.yml spec, but if there isn't an elegant way to do it, it's no big deal

Gundersanne (approved):
very elegant

test/cmd/api.sh
Gundersanne, line 72: Just saw tests are failing: for `fedora-*` tests, you'll need to change this request to fedora ones, unless we wanna test cross-distro as well.

Gundersanne (dismissed):

larskarlitski:
Rebased to resolve conflicts and to switch `api.sh` to build `fedora-31`. This could be extended to other (or more) distros in the future, but it's not really what this script is testing. Instead, it's testing that the API works end-to-end.

test/cmd/api.sh
larskarlitski: They do not. Thanks for pointing it out. I've changed it to build fedora-31 (32 cannot be built on other systems currently, due to an selinux bug).

cmd/osbuild-composer/composer.go
larskarlitski: I had them in there at first (in `api.go`), but did not want to depend on those packages in the main binary. I think it should not, because the top-level entry points should really all be specified next to each other to ensure they're not conflicting, no?

test/cmd/api.sh
larskarlitski: Good point! I've taken this verbatim from `aws.sh`. Do you mind if we do this in a follow-up?

Filed https://github.com/osbuild/osbuild-composer/issues/1037 to not forget.

test/cmd/api.sh
larskarlitski: Yeah, these were wrong. See https://github.com/osbuild/osbuild-composer/pull/1017#issuecomment-708622274

cmd/osbuild-composer/composer.go
Gundersanne: that is a good point, agreed

larskarlitski:
> Rebased to resolve conflicts and to switch `api.sh` to build `fedora-31`. This could be extended to other (or more) distros in the future, but it's not really what this script is testing. Instead, it's testing that the API works end-to-end.

And of course that didn't work, because we cannot build Fedora on RHEL (yay runners, @gicmo :wink:)

Added a commit which installs `/usr/share/tests/osbuild-composer/repositories.json`, meant to be used by all tests. `api.sh` uses it to build an image on the OS it's running on.

teg (approved):
This is great! Thanks!

ondrejbudai (approved):
image

schutzbot/deploy.sh
```diff
@@ -71,15 +71,8 @@ retry sudo dnf -y install osbuild-composer-tests
...
-    greenprint "Overriding default osbuild-composer rhel-8 sources"
-    sudo cp rhel-8.json /etc/osbuild-composer/repositories/
-fi
-
-if [[ -f "rhel-8-beta.json" ]]; then
-    greenprint "Overriding default osbuild-composer rhel-8-beta sources"
-    sudo cp rhel-8-beta.json /etc/osbuild-composer/repositories/
-fi
+greenprint "Overriding default osbuild-composer sources"
+sudo cp schutzbot/*.json /etc/osbuild-composer/repositories/
```
teg, line 75: Should be `schutzbot/repositories/*.json`.

teg (changes requested):

ondrejbudai:
`tools/deploy/test/files/etc/osbuild-composer/repositories`

^ is this file good for anything?

schutzbot/deploy.sh
larskarlitski: Indeed :facepalm: thanks

larskarlitski:
> ^ is this file good for anything?

It is! It's a link in `tools/deploy/test` which makes sure that a local test run with `tools/test-qemu` includes those files as well.

teg (approved):

ondrejbudai (approved):

teg (dismissed):

larskarlitski:
All tests are green, but Jenkins failed after the pipline ran successfully with:
```
[2020-10-15T23:19:52.512Z] groovy.lang.MissingPropertyException: No such property: none for class: groovy.lang.Binding
```

ondrejbudai (approved):

ondrejbudai:
#1032

Issue #1013, opened 2020-09-30
Add RHEL 8.4 support


jkozol:
osbuild-composer needs to support producing rhel 8.4 images. This work will be based off of the current support for rhel 8.3 images and off of the official 8.4 nightlies found here: http://download-ipv4.eng.brq.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8/compose/BaseOS/x86_64/iso/

jkozol:
There are numerous differences between the official rhel 84 qcow2 and our image based off of the rhel 83 image. They are as follows:

An additional "rescue" bootmenu entry.
```
    {
      "grub_arg": "--unrestricted",
      "grub_class": "kernel",
      "grub_users": "$grub_users",
      "id": "rhel-20200929154808-0-rescue-724317cf905047a2b52844ebed98adf9",
      "initrd": "/initramfs-0-rescue-724317cf905047a2b52844ebed98adf9.img",
      "linux": "/vmlinuz-0-rescue-724317cf905047a2b52844ebed98adf9",
      "options": "$kernelopts",
      "title": "Red Hat Enterprise Linux (0-rescue-724317cf905047a2b52844ebed98adf9) 8.4 (Ootpa)",
      "version": "0-rescue-724317cf905047a2b52844ebed98adf9"
    },
```

The firewall is enabled for cockpit.
```
  "firewall-enabled": ["ssh", "dhcpv6-client", "cockpit"],
```

There is an additional efi partition.
```
    [
      "UUID=7B77-95E7",
      "/boot/efi",
      "vfat",
      "defaults,uid=0,gid=0,umask=077,shortname=winnt",
      "0",
      "2"
    ],
```

There is a default hostname:
```
  "hostname": "localhost.localdomain",
```

The official image's default timezone is New_York:
```
  "timezone": "New_York"
```

Our osrelease is still for rhel 8.3. This is the official one:
```
  "os-release": {
    "ANSI_COLOR": "0;31",
    "BUG_REPORT_URL": "https://bugzilla.redhat.com/",
    "CPE_NAME": "cpe:/o:redhat:enterprise_linux:8.4:beta",
    "HOME_URL": "https://www.redhat.com/",
    "ID": "rhel",
    "ID_LIKE": "fedora",
    "NAME": "Red Hat Enterprise Linux",
    "PLATFORM_ID": "platform:el8",
[...]

PR #1018, opened 2020-10-05
distro: add rhel 84 support

jkozol:
Nightly repos are are added for testing.

The distro name for 84 is rhel-84 unlike prior rhel releases which fall under the umbrella name rhel-8. This fix will be generalized.

This runs on rhel 84 and can build a qcow2. But, there are many differences with the official image.

These differences are documented in #1013 

I am opening this as a draft PR to make it easier to collaborate on this with @chloenayon 

This PR is dependent on https://github.com/osbuild/osbuild/pull/518


cmd/osbuild-composer/composer.go
```diff
@@ -107,6 +108,9 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener, localWorkerListe
 
 	// TODO: refactor to be more generic
 	name := hostDistro.Name()
+	if name == "rhel-84" {
+		name = "rhel-8"
+	}
```
jkozol, line 113: Will we eventually split up the rhel-8 distro into it's minor releases? If not should I make this more generic now?

jkozol:
I can now generate a qcow2 image and boot it using a manifest built by this PR. But, for some reason, I am unable to get the image to build on my rhel 8.4 machine.

internal/distro/rhel84/distro.go
```diff
@@ -0,0 +1,1107 @@
...
+		packages: []string{
+			"@core",
+			"chrony",
+			"dnf",
+			"dosfstools",
+			"efi-filesystem",
+			"efivar",
+			"efivar-libs",
+			"grub2-efi-x64",
+			"shim-x64",
```
jkozol, line 824: @gicmo I added these files to the qcow2 image type to follow what is supported in the official image. All x86_64 images will have grub2-efi-x64 and shim-x64 in their build packages. Do you have thoughts on which image types might need these packages installed? The official kickstarts do not explicitly install these packages.

jkozol:
I am now able to successfully build rhel 8.4 images on rhel 8.4 so I am opening this up for review.

atodorov:
> I am now able to successfully build rhel 8.4 images on rhel 8.4 so I am opening this up for review.

I don't see any changes to Jenkinsfile and I don't think schutzbot is running any jobs against the 8.4 definitions. If I'm wrong can you explain why ?

jkozol:
> I don't see any changes to Jenkinsfile and I don't think schutzbot is running any jobs against the 8.4 definitions. If I'm wrong can you explain why ?

There is currently no test runner for 8.4. @major do you have any thoughts on what it will take to add 8.4 testing to schutzbot?

cmd/osbuild-composer/composer.go
teg: I think that would make sense. There are no immediate plans to split into the minor releases, that would anyway require more work as some customers would be on the generic RHEL8, but others would opt into a specific minor version.

schutzbot/deploy.sh
```diff
@@ -51,3 +51,14 @@ fi
 greenprint "Installing the Image Builder packages"
 # Note: installing only -tests to catch missing dependencies
 retry sudo dnf -y install osbuild-composer-tests
+
+# Provision the software under test.
+/usr/libexec/osbuild-composer-test/provision.sh
```
teg, line 56: How come this line is added (back?)?

schutzbot/deploy.sh
```diff
@@ -51,3 +51,14 @@ fi
...
+
+# RHEL 8.4 still requires the nightly repos to be used
```
teg, line 58: I think this chunk should be done in `provision.sh`, and `deploy.sh` should be kept generic. The thinking being that `deploy.sh` is specific to running these tests in schutzbot, but they should behave the same if you just execute them from the test RPM on a non-schutzbot system too.

teg (changes requested):
Loving this!
One more thing I think we need: we must test that the hybrid images boot both using BIOS and UEFI. I know @gicmo had this set up, at some point, we just need to add the conditionals to the `qemu.sh` integration test.

schutzbot/deploy.sh
jkozol: So, each test should individually call `provision.sh` if they need it and `deploy.sh` wouldn't be needed for the tests to run unless they are in a schutzbot system? This makes sense to me.

ondrejbudai:
This needs a rebase after #1063 lands.

larskarlitski:

PR #1020, opened 2020-10-06
tools: add deploy-qemu and gen-user-data

larskarlitski:
tools/gen-user-data generates a cloud-init user-data file from a
configuration directory. It is mostly useful to embed files in the
user-data.

tools/deploy-qemu uses above tool to make a user-data file and spins up
a virtual machine with it. This is useful to locally run, test, and
debug osbuild-composer.

A simple user-data directory for running tests locally is included in
tests/deploy-user-data. It expects a repository with osbuild-composer
rpms to be served on the host's port 8000.

### DIFF
--- a/tools/deploy/test/user-data.yml
+++ b/tools/deploy/test/user-data.yml
@@ -1,0 +1,16 @@
+#cloud-config
+yum_repos:
+  osbuild:
+    name: osbuild
+    baseurl: "http://10.0.2.2:8000"
+    enabled: true
+    gpgcheck: false
+user: admin
+password: foobar
+ssh_pwauth: True
+chpasswd:
+  expire: False
+sudo: 'ALL=(ALL) NOPASSWD:ALL'
+runcmd:
+  - /run/provision-scripts/deploy.sh
+  - /run/provision-scripts/provision.sh
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,35 @@
+# Hacking on osbuild-composer
+
+*osbuild-composer* cannot be run from the source tree, but has to be installed
+onto a system. We recommend doing this by building rpms, with:
+
+    make rpm
+
+This will build rpms from the latest git HEAD (remember to commit changes), for
+the current operating system, with a version that contains the commit hash. The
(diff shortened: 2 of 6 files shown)


ondrejbudai:
My first thought is that we should have something like `HACKING.md` that would contain instructions on how to use these scripts. I like that the scripts are documented inside them but it would be nice to have a central location for this kind of documentation so external contributors can find it easily.

larskarlitski:
> My first thought is that we should have something like `HACKING.md` that would contain instructions on how to use these scripts. I like that the scripts are documented inside them but it would be nice to have a central location for this kind of documentation so external contributors can find it easily.

I very much agree. I've added one that explains this workflow. Hope it makes sense — feedback welcome.

tools/deploy/test/user-data.yml
```diff
@@ -0,0 +1,16 @@
+#cloud-config
+yum_repos:
+  osbuild:
+    name: osbuild
+    baseurl: "http://10.0.2.2:8000"
```
msehnout, line 5: Where does this IP address come from? Is it always the ip of the host machine available from the guest?

msehnout:
> > My first thought is that we should have something like `HACKING.md` that would contain instructions on how to use these scripts. I like that the scripts are documented inside them but it would be nice to have a central location for this kind of documentation so external contributors can find it easily.
> 
> I very much agree. I've added one that explains this workflow. Hope it makes sense — feedback welcome.

It's nice :+1:

tools/deploy/test/user-data.yml
larskarlitski: Yeah, documented [here](https://wiki.qemu.org/Documentation/Networking#User_Networking_.28SLIRP.29). Just learned about this myself for this PR. I'll add a comment with a link in the file.

ondrejbudai (approved):
Very cool, thanks!

tools/deploy/test/user-data.yml
larskarlitski: Oops, Ondřej already merged it (thanks!)

I'll post a separate PR.

PR #1021, opened 2020-10-07
test: depend on rpms rather than installing them explicitly

teg:
The -test subpackage should depend on all that it needs, and we should
never need to install anything explicitly from the test scripts.


atodorov (approved):
LGTM but why do we keep having 2 .spec files ?

ondrejbudai (approved):
That's a lot of stuff. :-O Nevertheless, I agree that it makes sense, thanks!

ondrejbudai:
> LGTM but why do we keep having 2 .spec files ?

We still need to use the old one on Fedora 31. It will soon be gone.

PR #1022, opened 2020-10-07
test: enable epel on RHEL

ondrejbudai:
I broke master by merging #1021 - ansible and koji are not in the official
RHEL repositories so we need to enable epel on RHEL to be able to get them.


teg (approved):
Thanks!

PR #1024, opened 2020-10-09
spec: obsolete lorax-composer in Fedora 34

ondrejbudai:
Lorax-composer won't be shipped into Fedora 34. To upgrade users from it to osbuild-composer we need to obsolete it in our spec file.

Fixes RHBZ#1886405

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -67,6 +67,12 @@ Obsoletes: lorax-composer <= 29
 Conflicts: lorax-composer
 %endif
 
+# Remove when we stop releasing into Fedora 35
+%if 0%{?fedora} >= 34
+# lorax 34.3 is the first one without the composer subpackage
+Obsoletes: lorax-composer < 34.3
+%endif
+
 # remove in F34
 Obsoletes: golang-github-osbuild-composer < %{version}-%{release}
 Provides:  golang-github-osbuild-composer = %{version}-%{release}


osbuild-composer.spec
```diff
@@ -67,6 +67,12 @@ Obsoletes: lorax-composer <= 29
 Conflicts: lorax-composer
 %endif
 
+# Remove when we stop releasing into Fedora 35
+%if 0%{?fedora} >= 34
+# lorax 34.3 is the first one without the composer subpackage
+Obsoletes: lorax-composer < 34.3
```
hroncok, line 73: This way, you can easily reintroduce it if needed if this ends up in the Fedora 34 GA repo.

```suggestion
Obsoletes: lorax-composer < 34.3-100
```

If you are 100 % positive you won't even need to reintroduce lorax-composer, you can ignore this suggestion.

hroncok (commented):
Do you want osbuild-composer installed when users do: `dnf install lorax-composer`? If so, consider adding `Provides: lorax-composer 1:%{version}-%{release}`.

osbuild-composer.spec
bcl: lorax-composer is gone, the code has been removed from master and rawhide and will not be returning.

bcl (approved):
ACK

ondrejbudai:
Thanks, @hroncok!

The CI is red because of flakey F32 mirrors.

PR #1027, opened 2020-10-13
test/ostree: use releng repo instead of nightly for rhel8.3

ondrejbudai:
The previously used repo doesn't exist anymore making this test fail with:

Error validating install location: Could not find an installable distribution
at 'http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.3/compose/BaseOS/x86_64/os/':
The URL could not be accessed, maybe you mistyped?

Let's switch to a releng repo that shouldn't be deleted.


larskarlitski (approved):
Oh dear. Yeah. Thanks!

atodorov (approved):

msehnout (approved):
Looks good to me and Schutzbot reports failure even though everything is green. I'm merging this. Thanks!

PR #1028, opened 2020-10-13
tests: Fail if python36 is installed. Closes #794

### LINKED ISSUE #794: tests: Use minimal images for testing CI w/o python36
https://github.com/osbuild/osbuild-composer/pull/788 says 
"Now that we have CI images built by osbuild-composer, let's use them!" but from what I can see it deals only with mock. I don't see how the images themselves are being built. 

OTOH I've been seeing the following error on 8.3:
```
execv(/run/osbuild/lib/runners/org.osbuild.rhel83) failed: No such file or directory
```

which is fixed by https://github.com/osbuild/osbuild/pull/452 but our CI systems come with python36 preinstalled onto them which was hiding the error (https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-793/1/pipeline/131 from PR #793). 

cc @major

atodorov:


schutzbot/ci_details.sh
```diff
@@ -37,6 +37,13 @@ echo "List of installed packages:"
 rpm -qa | sort
 echo "------------------------------------------------------------------------------"
 
+if [ $(rpm -qa | grep -c python36) == 0 ]; then
```
larskarlitski, line 40: I think you can simplify this to:
```suggestion
if rpm --quiet -q python36; then
```

That also gets rid of the shellcheck warning.

teg:
I agree with the intention, but I'm not sure if we should hard-code packages we don't want. Ideally we should move schutzbot towards being agnostic to the package/distro under test, and this goes in the other direction by hard-coding a composer/rhel specific test in schutzbot itself.

We could have a dedicated test that does environment validation, but I'm not sure if that's right, at least let's not have it in schutzbot or `ci_details.sh`.

I think the underlying issue is that it is not straight-forward to know what images we are testing on top of, so that is what I think we should improve (so we just get the base image + the deps declared by the package under test, and then whatever is pulled in is pulled in).

@msehnout and @dvdhrm are working on decoupling the tests from the package under test (having the tests run remotely), which should make these sort of dependency issues go away.

atodorov:
> I agree with the intention, but I'm not sure if we should hard-code packages we don't want. Ideally we should move schutzbot towards being agnostic to the package/distro under test, and this goes in the other direction by hard-coding a composer/rhel specific test in schutzbot itself.
> 
> We could have a dedicated test that does environment validation, but I'm not sure if that's right, at least let's not have it in schutzbot or `ci_details.sh`.
> 
> I think the underlying issue is that it is not straight-forward to know what images we are testing on top of, so that is what I think we should improve (so we just get the base image + the deps declared by the package under test, and then whatever is pulled in is pulled in).
> 
> @msehnout and @dvdhrm are working on decoupling the tests from the package under test (having the tests run remotely), which should make these sort of dependency issues go away.


I agree in principle with everything in this comment but in practice all of the groundwork for what you propose is missing ATM and this one is a clear case/easy fix. I'm happy to move it elsewhere but let's not keep this PR open for long.

teg:
> > I agree with the intention, but I'm not sure if we should hard-code packages we don't want. Ideally we should move schutzbot towards being agnostic to the package/distro under test, and this goes in the other direction by hard-coding a composer/rhel specific test in schutzbot itself.
> > We could have a dedicated test that does environment validation, but I'm not sure if that's right, at least let's not have it in schutzbot or `ci_details.sh`.
> > I think the underlying issue is that it is not straight-forward to know what images we are testing on top of, so that is what I think we should improve (so we just get the base image + the deps declared by the package under test, and then whatever is pulled in is pulled in).
> > @msehnout and @dvdhrm are working on decoupling the tests from the package under test (having the tests run remotely), which should make these sort of dependency issues go away.
> 
> I agree in principle with everything in this comment but in practice all of the groundwork for what you propose is missing ATM and this one is a clear case/easy fix. I'm happy to move it elsewhere but let's not keep this PR open for long.

Yeah, you are right, let's merge this now, and rework it in the future when the groundwork is in place.

teg (approved):

PR #1029, opened 2020-10-13
Fix osbuild-composer in ELN

ondrejbudai:
See individual commits.

I would introduce a test building our package in ELN using `mock` but `mock` with ELN target seems to be currently broken. I already asked for help in fedora devel channel.

There's at least a manual ELN scratch build: https://koji.fedoraproject.org/koji/taskinfo?taskID=53369289

Fixes rhbz#1884383


larskarlitski (approved):
Thanks for figuring this out and the clear explanation. Sounds reasonable to me.

ondrejbudai:
Seeing #1032, the tests are actually green.

Issue #1031, opened 2020-10-13
Many tests fail because of faulty mirrors


larskarlitski:
osbuild-composer's tests are downloading a lot of packages from a myriad of mirrors. Some of those are flaky, or not in sync, or rate-limit us.

Thist most often manifests itself like this in osbuild's output:
```
RuntimeError: org.osbuild.files: curl: error downloading {'url': 'https://mirrors.rit.edu/fedora/fedora/linux/releases/32/Everything/x86_64/os/Packages/g/geolite2-city-20191217-2.fc32.noarch.rpm'}: error code 28
```

atodorov:
> osbuild-composer's tests are downloading a lot of packages from a myriad of mirrors. Some of those are flaky, or not in sync, or rate-limit us.

@larskarlitski I am curious about 2 items:
1) If a download fails doesn't dnf go to the next one, then next one, etc ? Is this causing real issues and failures to download packages or just annoying error messages in the logs ? 

2) Isn't there a Fedora mirror in AWS ? There should be. Quick search found the following resources, all of them a bit unclear:
https://fedoraproject.org/wiki/EC2_Mirrors
https://fedoraproject.org/wiki/Archive:Infrastructure/Mirroring/Amazon - contains a list of regional mirror URLs
https://fedora-infra-docs.readthedocs.io/en/latest/sysadmin-guide/sops/mirrormanager-S3-EC2-netblocks.html

but it looks like there should be mirroring infrastructure inside AWS and dnf is supposed to be using it.

larskarlitski:
> @larskarlitski I am curious about 2 items:
> 
>     1. If a download fails doesn't dnf go to the next one, then next one, etc ? Is this causing real issues and failures to download packages or just annoying error messages in the logs ?

It does, but osbuild's file source doesn't implement the same logic. Mostly because we didn't know that dnf tries fetching files from mirrors that are *not* the one it downloaded the metadata from. This could be added, of course, but is a bit tricky.

>     2. Isn't there a Fedora mirror in AWS ? There should be. Quick search found the following resources, all of them a bit unclear:
>        https://fedoraproject.org/wiki/EC2_Mirrors
>        https://fedoraproject.org/wiki/Archive:Infrastructure/Mirroring/Amazon - contains a list of regional mirror URLs
>        https://fedora-infra-docs.readthedocs.io/en/latest/sysadmin-guide/sops/mirrormanager-S3-EC2-netblocks.html
> 
> 
[...]

PR #1033, opened 2020-10-14
make: put all binaries into bin directory

ondrejbudai:
Currently, we have osbuild-image-tests binary committed in the master branch. IMHO the root cause is that we don't have it in .gitignore. Actually, I think that it is pretty hard to keep .gitignore in sync with the build target.

This commit solves the situation by putting all the built binaries into bin directory and adding this directory into .gitignore. This way, it's much harder to actually commit a new Go binary into the repository.

This commit doesn't remove the binary as #1017 already does that.

We don't want to merge this before #1017

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,4 @@
 __pycache__
 
-/osbuild-composer
-/osbuild-worker
-/osbuild-weldr-tests
-/osbuild-pipeline
-/osbuild-upload-azure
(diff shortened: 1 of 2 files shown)


larskarlitski (approved):
Yeah, much better. Thanks!

ondrejbudai:
#1031 and #1027 showed, otherwise, we are green!

Issue #1032, opened 2020-10-13
jenkins is red even though all tests are green


ondrejbudai:
An example here: https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1029/1/pipeline

PR #1034, opened 2020-10-14
ci: fix artifact name generation in cloud cleaner

msehnout:
The issue comes from the fact that the PR introducing it was very old
and meanwhile the variable used for image name creation has changed.
This patch makes sure both functions are the same.


larskarlitski (approved):
:+1: Thanks!

ondrejbudai:
#1032 is showing, otherwise, we are good.

PR #1036, opened 2020-10-14
schutzbot/run_cloud_cleaner: pass BRANCH_NAME instead of CHANGE_ID

msehnout:
`CHANGE_ID` bug strikes for the 3rd time :-/. Hope this is the last one. Recursive grep does not show any other usage:
```
$ rg CHANGE_ID
$
```


msehnout:
#1032 strikes again

ondrejbudai (approved):

PR #1038, opened 2020-10-14
test: install koji-compose.py

larskarlitski:
And use it from its installed location in test/cmd/koji.sh.


ondrejbudai (approved):
Makes sense, thanks!

atodorov (approved):

ondrejbudai:
openstack failed for some unknown reason but that is irrelevant.

PR #976, opened 2020-09-09
schutzbot: Replace EL8 nightly pipeline with conditional steps

atodorov:
- looks like having 2 pipelines defined in the same Jenkinsfile
  is not working as we expect
- remove the existing EL8 nightly pipeline definition
- use detect_build_cause() funtion and set a global env.BUILD_CAUSE
  variable for use in conditionals
- add a cron job trigger - this will work together with the
  GitHub pull request trigger
- use conditional blocks for all steps we want to be executed
  outside of cron jobs
- only EL8.3 jobs will be executed unconditionally, both in cron
  and for PRs. The preparation stage for cron jobs makes sure to
  use the same name for osbuild-mock.repo so that the jobs can
  unstash it later!

PR #1040, opened 2020-10-16
schutzbot: remove rhel nightly pipeline

### LINKED ISSUE #1032: jenkins is red even though all tests are green

ondrejbudai:
Jenkins doesn't support two pipelines in one file. We had two and it caused
an exception in Jenkins marking all builds as failed even though all stages
passed successfully.

As this pipeline isn't used for anything right now (see #976), it's safe to
delete it.

Fixes #1032


larskarlitski (approved):
Makes sense to me with the context of the linked bug. Thanks!

PR #1041, opened 2020-10-19
spec: remove executable bit from certificates

ondrejbudai:
Trust me, it's not very useful.


larskarlitski (approved):
I execute my certificates all the time!

Sorry, that was me, eh? :roll_eyes:

PR #1042, opened 2020-10-19
schutzbot: move test scripts into the test package

teg:
`run_{base,image}_tests.sh` are osbuild-composer-specific, and should
live in the `-test` subpackage, rather than as part of the testing
infratructure.

This is not a functional change.


teg:
I added a lot of reviewers, don't need everyone to review this, just wanted to bring it to your attention as you are/have been involved with schutzbot in various ways.

larskarlitski (commented):
*nod* Thank you!

schutzbot/Jenkinsfile
```diff
@@ -436,14 +436,14 @@ void run_tests(test_type) {
     if (test_type == 'base') {
         sh (
             label: "Base tests",
-            script: "schutzbot/run_base_tests.sh"
+            script: "/usr/libexec/tests/osbuild-composer/base_tests.sh"
```
msehnout, line 366: This breaks the idea of iterating over `/usr/libexec/tests/osbuild-composer/`, no? Shouldn't we move this somewhere else?

schutzbot/deploy.sh
```diff
@@ -41,16 +41,6 @@ sudo rm -f /etc/yum.repos.d/fedora*modular*
...
 echo -e "fastestmirror=1\ninstall_weak_deps=0" | sudo tee -a /etc/dnf/dnf.conf
 
-# Ensure we are using the latest dnf since early revisions of Fedora 31 had
-# some dnf repo priority bugs like BZ 1733582.
-# NOTE(mhayden): We can exclude kernel updates here to save time with dracut
-# and module updates. The system will not be rebooted in CI anyway, so a
-# kernel update is not needed.
-if [[ $ID == fedora ]]; then
-    greenprint "Upgrading system to fix dnf issues"
-    sudo dnf -y upgrade --exclude kernel --exclude kernel-core
```
msehnout, line 51: 🎉

msehnout (dismissed):
Very nice cleanup 👍 , but I'm not sure about the location for the scripts. Can you please explain more how this plays with the idea of iterating over the test files?

schutzbot/Jenkinsfile
```diff
@@ -443,7 +370,8 @@ void run_tests(test_type) {
     if (test_type == 'image') {
         sh (
             label: "Image tests",
-            script: "/usr/libexec/tests/osbuild-composer/run_image_tests.sh"
+            script: "/usr/libexec/tests/osbuild-composer/
+            image_tests.sh"
```
larskarlitski, line 374: Was this accidental?

schutzbot/deploy.sh
larskarlitski: Indeed! So good.

larskarlitski (changes requested):
Very nice cleanups indeed. Martin raised a good point and I think an unrelated change sneaked in. Otherwise good to me.

ondrejbudai (dismissed):
I like the cleanups! Some of my concerns were already covered by others but I have other kinda big for me:

This PR removes all testing for F31 images. If we still want to keep support for building them, we need some kind of testing. So, either we drop support for building F31 images, leave the tests for now (I can imagine a subset but that's possibly an extra work), or introduce some kind of cross-distro image tests (this makes the most sense because the only way how to build Fedora 31 images right now is to install osbuild-composer on F32/33 or RHEL and use Composer API or Composer API for Koji to perform a cross-distro build).

schutzbot/Jenkinsfile
teg: Very good point! My idea was to move the test binaries that are not called directly by schutzbot somewhere else, so that schutzbot really just iterates over the scripts in `/usr/libexec/tests/osbuild-composer/*`. Now we have both the scripts in there and the binaries they call, which isn't quite right. Does that make sense? I'll add a patch to do that.

schutzbot/Jenkinsfile
teg: Yikes. Indeed. Thanks!

teg:
> This PR removes all testing for F31 images. If we still want to keep support for building them, we need some kind of testing. 

Very good point! Is there any reason to still support them? We never supported/tested cross-distro builds, so I would just rip it all out (I should have done that in the same patch).

schutzbot/Jenkinsfile
larskarlitski: Hm I don't understand. Where would that "somewhere else" be?

ondrejbudai:
Rip it all out then.

schutzbot/Jenkinsfile
teg: `/usr/libexec/osbuild-copmoser-test/`. I'd move all the helper scripts there (some are currently in `/usr/libexec/osbuild-composer/`).

larskarlitski: Sounds good to me. Thanks for clarifying.

schutzbot/Jenkinsfile
teg: Fixed.

ondrejbudai (commented):
Schutzbot is failing because of the lines 203 and 204:

```
install -m 0755 -vd                    %{buildroot}%{_datadir}/tests/osbuild-composer/cases
install -m 0644 -vp test/data/cases/*  %{buildroot}%{_datadir}/tests/osbuild-composer/cases/
```

`cases` were renamed to `manifests`

larskarlitski (dismissed):
Amazing! Love seeing this come along so nicely. Thanks!

larskarlitski (approved):
Thanks again! :heart: 

It's all green except for `fedora_32-x86_64-openstack-boot.json`, which is to be expected with the recent outages of the OpenStack instance we're using.

Ondřej's comment has been addressed (rip out fedora-31). I think this is good to go in.

larskarlitski:
(Martin's comments have been addressed as well.)

ondrejbudai (approved):
Amazing, thanks! The changes look alright to me.

I especially like that `test/ca/ca-key.pem` was deleted. I learned about the presence of this file one week ago but I failed to post a small PR to remove it. Thanks a lot for easing my mind!

There are still some leftovers from Fedora 31 but they can be handled in a follow-up:

`schutzbot/repositories/fedora-31.json`
=> We don't need this repo override anymore.

`test/cases/aws.sh`
=> There's still a condition for disabling this test for F31.



*For curious people, I used the following command to inspect our codebase for F31 left-overs:*

`rg -g "!/vendor/*" -g "!/osbuild/*" -g "!*.svg" -g "!/test/data/manifests/*" -g "!/internal/store/test/*" 31`

PR #1043, opened 2020-10-20
docs/HACKING: add a note about how to grow the image

ondrejbudai:
Yesterday, I spent more hours than I would like to admit debugging tests that
was failing because the Fedora cloud base image has only 4 GB disk space.

This commit adds a note for the future me.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -21,6 +21,9 @@ To start a ephemeral virtual machine using this repository, run:
 
 `IMAGE` has to be a path to an cloud-init-enabled image matching the host
 operating system, because that's what the packages where built for above.
+Note that the Fedora/RHEL cloud images might be too small for some tests
+to pass. Run `qemu-img resize IMAGE 10G` to grow them, cloud-init's growpart
+module will grow the root partition automatically during boot. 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -21,6 +21,8 @@ To start a ephemeral virtual machine using this repository, run:
 
 `IMAGE` has to be a path to an cloud-init-enabled image matching the host
 operating system, because that's what the packages where built for above.
+Note that the Fedora/RHEL cloud images might be too small for some tests
+to pass. Run `qemu-img resize IMAGE 10G` to grow them.
```
larskarlitski, line 25: Good point. Could you also add a note that the filesystems are grown automatically, thanks to cloud-init's [growpart module](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#growpart), which is enabled by default on the root fs.

ondrejbudai: Like this?

larskarlitski (approved):
> Like this?

Perfect. Thanks a lot for this!

And sorry you had to debug for so long :disappointed:

PR #1044, opened 2020-10-20
Add tools/deploy-openstack

larskarlitski:
Similar to deploy-qemu, but deploys into an OpenStack cluster. It uses
the same logic for user-data.

No downstream testing is necessary for this commit, as it doesn't change *osbuild-composer* itself.


teg (approved):
This is great!

PR #1046, opened 2020-10-21
tests: remove stray F31 occurrences

ondrejbudai:
osbuild-composer doesn't support F31 anymore, this commit rips out the last pieces.

This is a follow-up to #1042.


larskarlitski (approved):
Thanks!

PR #1051, opened 2020-10-23
spec: generate packageinfo files for composer and worker

ondrejbudai:
For Koji builds, it would be nice to know the version of osbuild-composer and worker that was used to build the image.

This commit explores the approach of putting the information about the package version in two json files:

```
/usr/share/osbuild-composer/packageinfo.json
/usr/share/osbuild-worker/packageinfo.json
```

For now, I chose to have separate NVRA fields in the files. From my test run:

```
$ cat /usr/share/osbuild-composer/packageinfo.json 
{
  "name": "osbuild-composer",
  "version": "22",
  "release": "1.20201023git820a330.fc33",
  "arch": "x86_64"
}
$ cat /usr/share/osbuild-worker/packageinfo.json 
{
  "name": "osbuild-composer-worker",
  "version": "22",
  "release": "1.20201023git820a330.fc33",
  "arch": "x86_64"
}
```

The other approach is to query rpm database for this information but this changes `osbuild-composer` dependant on `rpm`, that is something that we probably don't want.

I'm open to all suggestions!

We will need something similar in `osbuild` (composer needs to know its version), so I'm asking for a review and suggestions also from @gicmo and @dvdhrm because it would nice to do this the same way in both packages.

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -101,6 +101,24 @@ sed -i "s$// +build kolo_xmlrpc_oldapi$// +build !kolo_xmlrpc_oldapi$" internal/
 sed -i "s$// +build !kolo_xmlrpc_oldapi$// +build kolo_xmlrpc_oldapi$" internal/upload/koji/xmlrpc-response.go
 %endif
 
+# create the packageinfo files for composer and worker
+tee osbuild-composer-packageinfo.json << EOF
+{
+  "name": "%{name}",
+  "version": "%{version}",
+  "release": "%{release}",
+  "arch": "%{_arch}"
+}
+EOF
+tee osbuild-worker-packageinfo.json << EOF
+{
+  "name": "%{name}-worker",
+  "version": "%{version}",
+  "release": "%{release}",
+  "arch": "%{_arch}"
+}
+EOF
+
 %build
 %if 0%{?rhel}
 GO_BUILD_PATH=$PWD/_build
@@ -147,8 +165,13 @@ install -m 0755 -vp _bin/osbuild-composer                       %{buildroot}%{_l
 install -m 0755 -vp _bin/osbuild-worker                         %{buildroot}%{_libexecdir}/osbuild-composer/
 install -m 0755 -vp dnf-json                                    %{buildroot}%{_libexecdir}/osbuild-composer/
 
+install -m 0755 -vd                                             %{buildroot}%{_datadir}/osbuild-composer
+install -m 0644 -vp osbuild-composer-packageinfo.json           %{buildroot}%{_datadir}/osbuild-composer/packageinfo.json
 install -m 0755 -vd                                             %{buildroot}%{_datadir}/osbuild-composer/repositories
 install -m 0644 -vp repositories/*                              %{buildroot}%{_datadir}/osbuild-composer/repositories/
+
+install -m 0755 -vd                                             %{buildroot}%{_datadir}/osbuild-worker
+install -m 0644 -vp osbuild-worker-packageinfo.json             %{buildroot}%{_datadir}/osbuild-worker/packageinfo.json
 
 install -m 0755 -vd                                             %{buildroot}%{_unitdir}
 install -m 0644 -vp distribution/osbuild-composer.service       %{buildroot}%{_unitdir}/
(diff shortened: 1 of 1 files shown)


larskarlitski:
I like this a lot.

> For Koji builds, [...]

In addition, I'd love to (out of scope for this PR) return this information from the `/status` http routes of all APIs. This will bring the concept of being able to know what you're running on a system (via `rpm -q`) to knowing what's running on a service you're talking to. After all, being able to trace from running software back to the source is one of the main benefits of Open Source software.

> The other approach is to query rpm database for this information but this changes `osbuild-composer` dependant on `rpm`, that is something that we probably don't want.

I tend to agree, but I'm having a hard time to find a reason. osbuild-composer already depends on rpm when it is installed, and we don't support deploying any other way. Maybe it could run on minimal systems (containers?) without an rpm binary and database?

ondrejbudai:
> I tend to agree, but I'm having a hard time to find a reason. osbuild-composer already depends on rpm when it is installed, and we don't support deploying any other way. Maybe it could run on minimal systems (containers?) without an rpm binary and database?

Well, if we for some reason want to ship composer in a different way than in a RPM package, we can still maintain backwards compatibility using the proposed "file method".

osbuild-composer.spec
```diff
@@ -101,6 +101,24 @@ sed -i "s$// +build kolo_xmlrpc_oldapi$// +build !kolo_xmlrpc_oldapi$" internal/
...
+EOF
+tee osbuild-worker-packageinfo.json << EOF
+{
+  "name": "%{name}-worker",
+  "version": "%{version}",
+  "release": "%{release}",
+  "arch": "%{_arch}"
+}
+EOF
+
```
dvdhrm, line 121: Are you explicitly skipping `epoch`?

dvdhrm (commented):
I don't mind making software introspect its own version. This is crucial for debugging and tracing errors. I think checksums would be beneficial, but that is not how systems work right now, so I guess we can skip this.

However, I wonder how to pick the information to store there. I think we have 2 possibilities:

1) Pick the version we as *upstream* control. This means `name`, `version`, `architecture`.
2) Pick the version `rpm` controls, this means calling out to RPM to fill this in.

I think you picked a middle-ground, which might get us into problems. In particular, with modular repositories, you want NSVCA to identify the stream+configuration of the build. Maybe `%{release}` includes this, but I thought it didn't. I don't even know whether RPM at this stage knows the NSVCA, or whether this is only added a step later when determining stream adherence.

Anyway, I wonder whether we want a `"rpm": "..."` field, rather than hardcoding them manually. So we can keep all the individual fields we control, but rather than adding things like `%{release}`, we just store the entire RPM version, as controlled by RPM? I didn't find any macros that neatly format this information, though. But maybe you guys know better ;)

teg:
I like this!

I think we should not depend at runtime on `rpm` and having this as a static file is nice. Ideally this would be something we would do for all our packages, so should not rely on the fact that we happen to have `rpm` around in the case of `osbuild-composer`.

I agree with @dvdhrm that there are essentially two approaches (and what they are). The way I see it is that we have the source-code on one hand and the build on the other. As a consumer of the service, I would be mostly interested in the source-code as it would allow me to figure out how things work (in case our docs are less than perfect). As we aspire to have a 1-to-1 mapping between git commits and SRPMs, I think all I would care about is the git SHA.

I can see the use for specifying the build too (and I think this is what koji would want). I would follow the scheme that they use, which is NEVRA, as structured data with `epoch` being optional (meaning `0`). Which is what you currently did.

How about:

```
$ cat /usr/share/osbuild-composer/packageinfo.json 
{
  "rpm": {
    "name": "osbuild-composer",
    "version": "22",
    "release": "1.20201023git820a330.fc33",
    "arch": "x86_64"
  },
  "commit": "0cb45a1cc2f7a15b914ef3374f350e2bf175a3c6"
}
$ cat /usr/share/osbuild-worker/packageinfo.json 
{
  "rpm": {
    "name": "osbuild-composer-worker",
    "version": "22",
    "release": "1.20201023git820a330.fc33",
    "arch": "x86_64"
  },
  "commit": "0cb45a1cc2f7a15b914ef3374f350e2bf175a3c6"
}
```

gicmo:
I like Tom's approach. One day composer might also be running on non-RPM based distros (Arch?), so something independent of RPM would be nice. Maybe `"rpm": { ... }` could be `"package": { "type": "rpm", ... }` because NVRA could probably be shared across different package managers.

larskarlitski:

PR #1052, opened 2020-10-23
Rework x509 certificate generation in our testing setup

ondrejbudai:
I decided to rework the current x509 setup we have in our tests. The main reason is that we need SAN certificates for Go 1.15 (Fedora 33, RHEL 8.4).

To achieve this, I created 3 helper scripts that handle X.509 certificates with SAN:

```
tools/x509/generate-certificate
tools/x509/generate-csr
tools/x509/sign-csr
```

I think it's pretty obvious what they do, feel free to inspect them, I tried to document them.

The scripts are now used everywhere where we need X.509 certificates. This also means one potentially controversial thing: We're no longer shipping pre-generated certificates in the test subpackage but instead, we're generating them on fly. 

I have several reasons to do this:

1) The pre-generated certificates are not documented. If someone wants to inspect them, he must know the right openssl incantation. This way, you are able to see what's inside the certificates in plain text.

2) The pre-generated certificates are going to expire at one point and someone will be surprised.

3) Shipping private keys in RPMs is iffy. I know, it's just for testing but still...

4) Auth tests are generating their own certificates. To achieve consistency, we have two options:
    a) Ship also all certificates for auth tests. That's extra 8 ones or something like that.
    b) Generate all certificates on fly. This commit does that.


teg (approved):
I like this 👍 
I think there was benefit to static files when it was harder to provision the tests, but now that the tests take care of all their own setup, I see no need. @larskarlitski @gicmo?

ondrejbudai:
I force-pushed a fix to `api.sh` failure. I put a comment in `tools/provision.sh` explaining why this was needed.

tl;dr: our Schutzbot setup currently violates the "run every test case on a pristine machine" principle.

tools/x509/openssl.py
```diff
@@ -0,0 +1,124 @@
...
+req_extensions = v3_req
+prompt = no
+
+[req_distinguished_name]
+CN = {common_name}
+
+[v3_req]
+keyUsage = critical,keyEncipherment, dataEncipherment, digitalSignature
+extendedKeyUsage = critical,serverAuth,clientAuth,emailProtection
+basicConstraints = critical,CA:FALSE
```
larskarlitski, line 31: This generates certificates with a wide range of usages. This is fine for tests, but we ought to never use them in production.

In my understanding, `CA` needs to be `true` for the ca certificate. It seems to work regardless — do you know why?

larskarlitski (changes requested):
Thanks for working on this! It's really nice to see this cleaned up. Generating certificates is indeed a preferable approach, for the reasons you listed.

I am however not a fan of the openssl wrapper scripts. They have similar-but-not-equal API, which might confuse people that already know `openssl`. Also, it means most certificates have the same options (as for example noted in a comment below).

I think I'd prefer using `openssl` directly, passing a config file that is specific to the test we generate certificates for. That would be easier to review than to read code to find out which options (mostly SANs) were passed.

However, that would be quite the rework :) This is a great improvement over what we have currently. So go ahead if you don't have the time or feel differently.

tools/provision.sh
```diff
@@ -12,14 +12,58 @@ sudo mkdir -p /etc/osbuild-composer/repositories
...
 
 sudo systemctl start osbuild-remote-worker.socket
 sudo systemctl start osbuild-composer.socket
 sudo systemctl start osbuild-composer-api.socket
 
+# The keys were regenerated but osbuild-composer might be already running.
+# Let's try to restart it. In ideal world, this shouldn't be needed as every
+# test case is supposed to run on a pristine machine. However, this is
+# currently not true on Schutzbot
+sudo systemctl try-restart osbuild-composer
```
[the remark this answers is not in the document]
ondrejbudai: This is the issue with the way we currently run the integration tests on Schutzbot.

In `Jenkinsfile` we call all integration test cases one after each other. However, each of these integration test scripts calls `provision.sh` which clearly breaks the pristine system assumption. I'm not sure how to solve this situation in a clear way, therefore I introduced this workaround.

tools/x509/openssl.py
ondrejbudai: I have no idea, the config is taken from `koji-osbuild`. I'm open to all suggestions.

ondrejbudai:
> I am however not a fan of the openssl wrapper scripts. They have similar-but-not-equal API, which might confuse people that already know `openssl`. Also, it means most certificates have the same options (as for example noted in a comment below).

Honestly, I think most people (including me) don't know `openssl` well. And that's exactly why I tried to hide it. If we have a wrapper reviewed by someone who does, then everybody on the team can use it to easily generate **safe** certificates. Restricting users to be able to use just a subset of `openssl` options sounds like a good way to prevent people from getting burned by delicate internals of the `openssl` hydra.

> 
> I think I'd prefer using `openssl` directly, passing a config file that is specific to the test we generate certificates for. That would be easier to review than to read code to find out which options (mostly SANs) were passed.

In `auth-tests`, we generate a lot of similar certificates. Honestly, I don't like the idea of shipping multiple very similar config files for `openssl`. Possibly, we can have an `openssl` configuration generator script, that sounds like a nice middleground between "have a wrapper for openssl" and "have a lot of similar configs in the repo".

tools/provision.sh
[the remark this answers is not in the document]
larskarlitski: Ah, I see. Hm. Maybe @teg has an idea?

tools/x509/openssl.py
larskarlitski: It needs to be like this when using the same config for all certificates. On a production system, you'd *only* allow `serverAuth` for the server cert, only `clientAuth` for a client cert, etc. (Again, it is fine for testing as is.)

I'm not sure why `CA:FALSE` works. The man page `X509V3_CONFIG(5)` says:

> A CA certificate must include the basicConstraints value with the CA field set to TRUE.

larskarlitski:
> Honestly, I think most people (including me) don't know `openssl` well. And that's exactly why I tried to hide it. If we have a wrapper reviewed by someone who does, then everybody on the team can use it to easily generate **safe** certificates. Restricting users to be able to use just a subset of `openssl` options sounds like a good way to prevent people from getting burned by delicate internals of the `openssl` hydra.

Very fair point. Except one thing: I wouldn't say these certificates are "safe" — they are ok for testing, but go against common practice of giving the least permissions possible.

> In `auth-tests`, we generate a lot of similar certificates. Honestly, I don't like the idea of shipping multiple very similar config files for `openssl`. Possibly, we can have an `openssl` configuration generator script, that sounds like a nice middleground between "have a wrapper for openssl" and "have a lot of similar configs in the repo".

Ah, I didn't know auth-tests generated that many certificates. I like the idea of a config generator.

ondrejbudai:
Superseeded by #1057

PR #1053, opened 2020-10-23
koji: add logs API route

chloenayon:
opening draft PR for koji logs API route and new tests

### DIFF
--- a/internal/kojiapi/api/openapi.yml
+++ b/internal/kojiapi/api/openapi.yml
@@ -55,6 +55,38 @@ paths:
             text/plain:
               schema:
                 type: string
+  '/compose/{id}/logs':
+    get:
+      summary: Get logs for a compose.
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+            example: 123e4567-e89b-12d3-a456-426655440000
+          required: true
+          description: ID of compose status to get
+      description: 'Get the status of a running or finished compose. This includes whether or not it succeeded, and also meta information about the result.'
+      responses:
+        '200':
+          description: The logs for the given compose, in no particular format (though valid JSON).
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ComposeLogs'
+        '400':
+          description: Invalid compose id
+          content:
+            text/plain:
+              schema:
+                type: string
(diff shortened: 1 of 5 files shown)


ondrejbudai:
This needs a rebase after #1063 lands.

teg (changes requested):
This is very nice, and a feature that has been requested by the brew developers :)

One request though: A compose can consist of several builds, even though that is not yet hooked up, so we should return an array of logs, one for each image build. For now, that just means changing the type from object to array of objects, but the rest of the code would be virtually unchanged.

gicmo:
Very cool! Eager to see this land so I can add this to `koji-osbuild`. One little thing `koji: add tests` is not very descriptive as a commit message.

larskarlitski:
Amended @chloenayon's patch slightly to return an array in `"image_logs"` and add an integration test to `Koji-compose.py`. I think this is ready for reviewing now.

Thanks @chloenayon!

internal/kojiapi/api/openapi.yml
```diff
@@ -115,6 +147,12 @@ components:
         koji_task_id:
           type: integer
           example: 203143
+    ComposeLogs:
+      required:
+        - image_logs
+      properties:
+        image_logs:
+          type: array
```
teg, line 155: If an array is returned we should document what each part of the array is. I thought you intended to just return one opaque blob? `image_logs` makes it sound like this is one blob per image build, but we also will have koji-init and koji-finalize logs, where would they go?

teg (changes requested):

internal/kojiapi/api/openapi.yml
larskarlitski: The array elements are undocumented because each of those is an opaque blob – one for each image request that was sent, in the same order. For now, they contain the contents of the job result, but we don't specify that.

The top level has `image_logs` so that we can put other things next to it (for example output from init/finalize jobs) if we want to.

teg: As promised, you explained why I was wrong. This makes perfect sense to me then.

teg (approved):

gicmo (approved):
:deciduous_tree:

PR #1056, opened 2020-11-02
test/koji: move the koji test to schutzbot

ondrejbudai:
The Koji test in Github actions was always a bit quick and dirty solution. I think it's a much nicer solution to run it on Schutzbot.

Therefore, this commit moves the `koji_test.go` to a new `osbuild-koji-tests` executable. This new test isn't run in the base test suite as one would anticipate but inside the `koji.sh` test. This is needed because `osbuild-koji-tests` requires a running koji instance.


osbuild-composer.spec
```diff
@@ -175,6 +176,7 @@ install -m 0755 -vp _bin/osbuild-weldr-tests                    %{buildroot}%{_l
 install -m 0755 -vp _bin/osbuild-dnf-json-tests                 %{buildroot}%{_libexecdir}/osbuild-composer-test/
 install -m 0755 -vp _bin/osbuild-image-tests                    %{buildroot}%{_libexecdir}/osbuild-composer-test/
 install -m 0755 -vp _bin/osbuild-auth-tests                     %{buildroot}%{_libexecdir}/osbuild-composer-test/
+install -m 0755 -vp _bin/osbuild-koji-tests                     %{buildroot}%{_libexecdir}/osbuild-composer-test/
```
msehnout, line 179: Should we introduce a new directory for test executables which are not base tests? I'm working on running all the files in `/usr/libexec/osbuild-composer-tests` as a "test" in `rpmci`. This is the code I use ATM (far from nice, but believe me, I want to make it nicer):
```
 sudo find /usr/libexec/osbuild-composer-tests/ -type f -exec sudo {} \\;"
```
where

msehnout (approved):
Looks good to me 👍

ondrejbudai:
ostree test for RHEL 8.3 failed. This is clearly an unrelated issue as this PR doesn't touch the ostree test code.

PR #1057, opened 2020-11-03
Rework x509 certificate generation in our testing setup: Second edition

ondrejbudai:
Another iteration of reworking our X.509 setup. This time, it uses the same stuff as @larskarlitski uses for generating certificates for composer.osbuild.org (aka the osbuild CA).

I can think of some cleanups in a follow-up but I wanted to make this PR as simple as possible.


test/cases/koji.sh
```diff
@@ -46,10 +46,10 @@ sudo cp \
     "${OSBUILD_COMPOSER_TEST_DATA}"/kerberos/krb5-local.conf \
     /etc/krb5.conf.d/local
 
-greenprint "Adding generated CA cert for Koji"
+greenprint "Adding the testsuite's CA cert to the system trust store"
 sudo cp \
-    /tmp/osbuild-composer-koji-test/ca-crt.pem \
-    /etc/pki/ca-trust/source/anchors/koji-ca-crt.pem
+    /etc/osbuild-composer/ca-crt.pem \
+    /etc/pki/ca-trust/source/anchors/osbuild-composer-tests-ca-crt.pem
```
teg, line 52: I remember doing this out of laziness, is there a way to drop in the right config to point to /etc/osbuild-composer/ca-crt.pem directly instead?

cmd/osbuild-auth-tests/certificates.go
```diff
@@ -28,61 +34,153 @@ func (ckp certificateKeyPair) key() string {
...
+
+func (c ca) certificate() string {
+	return path.Join(c.BaseDir, "ca.cert.pem")
+}
+
+func (c ca) key() string {
+	return path.Join(c.BaseDir, "private", "ca.key.pem")
+}
+
+func newCA(subj string) (*ca, error) {
```
teg, line 78: Could we just generate all these things in the provision script and not anything from golang?

teg (approved):
Couple of comments, but overall I like this. In particular, I'm convinced by the argument that this makes the certs self-documenting. I don't mind shipping test secrets (we still have ssh keys), I'd merge patches to move away from that, but I don't think it is the end of the world either.

Happy to have my comments addressed in follow-ups.

test/cases/koji.sh
ondrejbudai: `koji` can be indeed run with a custom CA but I didn't want to make more changes in this PR. Will do in a follow-up!

cmd/osbuild-auth-tests/certificates.go
ondrejbudai: We definitely can, I'm just not sure how to do this elegantly in `bash` without repeating many lines of `openssl` calls. Any ideas?

test/cases/koji.sh
teg: Thanks!

cmd/osbuild-auth-tests/certificates.go
teg: Could have a little helper for this? Either a function or a separate script. Not sure what's best to be honest.

test/data/x509/openssl.cnf
```diff
@@ -0,0 +1,85 @@
...
+distinguished_name = osbuild_distinguished_name
+
+
+#
+# Only prompt for CN
+#
+
+[osbuild_distinguished_name]
+CN = Common Name
+emailAddress = E-Mail Address
```
larskarlitski, line 85: These are for interactive processing (`openssl req` prompts for those). I don't think they're needed here, but also don't hurt – I think because you're providing both fields on the command line.

cmd/osbuild-auth-tests/certificates.go
larskarlitski: I think I'd prefer having this in the provision script as well, but wouldn't block on it. Not a regression.

larskarlitski (approved):
Beautiful. Sorry for the long delay for me to review this.

Thanks a lot!

ondrejbudai:
This PR is finally green! Thanks for the reviews, let's merge this!

PR #1058, opened 2020-11-03
Prepare worker package and binary for more job types

larskarlitski:
Since we only had one job type `osbuild` so far, a lot of code has assumptions about it. This PR aims to clean this up and allow for adding additional job types in the future.

This got quite large, but I hope I have explained the individual commits well enough.

No functionality is changed by these patches.


internal/cloudapi/server.go
```diff
@@ -225,3 +232,23 @@ func (server *Server) ComposeStatus(w http.ResponseWriter, r *http.Request, id s
 		panic("Failed to write response")
 	}
 }
+
+func composeStatusFromJobStatus(js *worker.JobStatus) string {
+	if js.Canceled {
+		return StatusFailure
```
teg, line 242: Should we introduce a `cancelled` state? Seems useful. Then again, I guess we do not yet support cancelling composes from this API.

teg: This would be a future improvement.

internal/kojiapi/server.go
```diff
@@ -223,34 +229,24 @@ func splitExtension(filename string) string {
...
+	if js.Canceled {
+		return StatusFailure
+	}
+
+	if js.Started.IsZero() {
+		return StatusPending
+	}
+
+	if js.Finished.IsZero() {
+		return StatusRunning
```
teg, line 247: This is a change of behavior. Is that intended? Did the koji openapi spec not need adjusting like the cloud one?

internal/kojiapi/server.go
```diff
@@ -223,34 +229,24 @@ func splitExtension(filename string) string {
...
-}
 
-func composeStateToImageStatus(state common.ComposeState) string {
-	switch state {
-	case common.CFailed:
-		return "failure"
-	case common.CFinished:
-		return "success"
-	case common.CRunning:
-		return "building"
```
teg, line 248: This is now mapped to `running`. Intentional?

internal/kojiapi/server.go
```diff
@@ -223,34 +229,24 @@ func splitExtension(filename string) string {
...
 }
 
-func composeStateToStatus(state common.ComposeState) string {
-	switch state {
-	case common.CFailed:
-		return "failure"
-	case common.CFinished:
-		return "success"
-	case common.CRunning:
-		return "pending"
```
teg, line 233: This is now mapped to `running`. Intentional?

cmd/osbuild-worker/main.go
```diff
@@ -261,6 +263,14 @@ func RunJob(job worker.Job, store string, kojiServers map[string]koji.GSSAPICred
 				}
 			}()
 
+			if result.Success == false {
```
teg: Why are all the others `!result.Success` and this `result.Success == false`? ;P

internal/cloudapi/server.go
larskarlitski: Agreed, let's not muddy this PR even more. I very much like the idea of such a state, though.

cmd/osbuild-worker/main.go
larskarlitski: The linter complained, too :)

internal/kojiapi/server.go
larskarlitski: Indeed! Thanks for the catch.

teg (approved):
Love this! So good! Much simpler to wrap ones head around now :) We are close to multi-build composes and all the glory that will bring us!

cmd/osbuild-worker/main.go
```diff
@@ -459,38 +450,21 @@ func main() {
...
-			// successful. There's no different way how to inform composer that
-			// e.g. an upload fail. Therefore, this line reuses the osbuild success
-			// flag to indicate all error kinds.
-			result.Success = false
-		} else {
-			log.Printf("  🎉 Job completed successfully: %v", job.Id())
+		cancelWatcher()
+		if err != nil {
+			log.Printf("Job %s failed: %v", job.Id(), err)
+			continue
```
ondrejbudai, line 186: I'm afraid this change will lead to a job stuck forever in the `WAITING` state. We must somehow send information to `composer` that a job failed even in cases where it wasn't an osbuild error.

ondrejbudai (changes requested):
Awesome changes! However, I think we can now have jobs stuck forever if something other than `osbuild` fails in the worker. See the inline comment.

cmd/osbuild-worker/main.go
larskarlitski: Good catch! This set of patches sends more errors in a structured way (i.e., the ones from targets), but doesn't set `osbuild.Result.Success` anymore on other errors. This was the wrong thing to do, but at least the job was marked as failed in composer.

The remaining possible errors are related to the worker:
1. not being able to create temporary directories
2. OSBuild returning invalid JSON
3. not being able to talk to composer.

We should not try to communicate (3) to composer, because the error itself was not being able to reach it.

The only thing composer can do for errors (1) and (2) (and possibly others in that category) is to reschedule the job on a different worker. That's something composer will need to do very soon anyway, when a job times out or a watchdog notices a worker is gone. For example, because we're using spot instances on AWS. If we don't add it, misbehaving workers can swallow all jobs, which will then never be completed.

I hesitate to add this logic with this PR, because it's already quite big. But I also hesitate to add another way of reporting errors of the (1) and (2) type, given that we want this other solution soon anyway.

Would you be ok merging this without, with this (rather small?) regression – that some jobs which used to be marked as failed, with no error message, are now marked as pending. In either case, someone with access to the worker has to check its logs.

larskarlitski:
I think CI is failing because of certificates expiring, which #1057 fixes. I'll rebase once that lands.

cmd/osbuild-worker/main.go
ondrejbudai: > Good catch! This set of patches sends more errors in a structured way (i.e., the ones from targets), but doesn't set `osbuild.Result.Success` anymore on other errors. This was the wrong thing to do, but at least the job was marked as failed in composer.
> 
> The remaining possible errors are related to the worker:
> 
>     1. not being able to create temporary directories
> 
>     2. OSBuild returning invalid JSON
> 
>     3. not being able to talk to composer.
> 
> 
> We should not try to communicate (3) to composer, because the error itself was not being able to reach it.
> 
> The only thing composer can do for errors (1) and (2) (and possibly others in that category) is to reschedule the job on a different worker.

This is the part I was missing. These errors are indeed caused by a misbehaving worker. I think that the worker should return some kind of "misbehaved" status at some point so we don't rely solely on timeouts.

I'm OK to merge with this small regression, I'm glad we cleared things up. Thanks!

ondrejbudai (commented):
My point was resolved, so I'm OK with merging. :) Therefore, this just needs a rebase once #1057 is in.

ondrejbudai (approved):

larskarlitski:
Rebased now that #1057 is in. Thanks @ondrejbudai!

teg (dismissed):
This PR did not change, so still ✅

ondrejbudai (approved):

ondrejbudai (approved):

ondrejbudai (changes requested):
The test failure is an actual bug in the PR. From `journalctl`:

```
Nov 06 11:31:31 rhel8cloudbase-50 osbuild-worker[28230]: 2020/11/06 11:31:31 Job 53cc5072-4fcd-45eb-8a7f-7ce406529a24 failed: error uploading artifact: Put "http://localhost/api/worker/v1/jobs/c9221e86-5b39-4d51-bb6e-9f2a9329f918/artifacts/": read /var/tmp/osbuild-worker-760549219: is a directory
```

The same issue is in the F32 job.

larskarlitski:
> The test failure is an actual bug in the PR. From `journalctl`:

Indeed, thanks! I was trying to be clever and detect whether a job supports uploading of artifacts based on which API is enabled. We enable all APIs in the tests, so even jobs scheduled through the koji API tried uploading artifacts. That failed, because jobs scheduled through that API do no carry the filename of the image to be uploaded. And it shouldn't: only the weldr API uploads back to composer.

Fix this by dropping the commit that tried to be clever. Enable artifacts for all jobs server-side, but only upload if the filename is actually set.

teg (approved):
I would have preferred an explicit boolean in the args that indicate that an image should be uploaded back, it was not clear to me that ImageName had this meaning.

But that is for a follow-up, if Schutzbot is happy this is good to go!

ondrejbudai:
`api.sh` test was also broken by this PR. I pushed a fix directly to this PR (see the last commit) so we can move on with this.

larskarlitski:
> `api.sh` test was also broken by this PR. I pushed a fix directly to this PR (see the last commit) so we can move on with this.

Thanks! I've rebased on `master` (with your quay.io fix) and put your commit (with attribution) in the right place.

ondrejbudai:
I messed up the `koji-compose.py` change a bit, so I had to force-pushed a fix.

ondrejbudai (approved):
Still awesome!

PR #1065, opened 2020-11-06
jobqueue: pass results between dependent jobs

teg:
This should be the last few changes we need in preparation for splitting our jobs apart so we can have multi-{build,arch} composes for koji.

This still needs some tests, and to actually introduce the new job types. 

This PR depends on #1058, so please only look at the last few commits.


teg:
Replaced by #1069.

PR #1067, opened 2020-11-09
test: don't install epel in test cases

ondrejbudai:
We don't install any packages in test cases anymore, therefore we don't need to install EPEL there.

A slightly different explanation:
osbuild-composer-tests depends on packages from EPEL on RHEL. Therefore, you cannot run the test cases without EPEL installed. Therefore, there's no point in installing EPEL there.


larskarlitski (approved):
Love it when stuff gets simpler ❤️

Thanks!

PR #1068, opened 2020-11-09
Switch the tests to use RPMRepo

ondrejbudai:
~`ostree`~\[1\] and `koji` tests haven't use RPMrepo yet. This PR changes that.

The only remaining place where we use the official repositories is the image tests. I will switch them later as that will require a bit of work and regenerating all the test cases.

This PR also does 2 related changes:
1) Bumps the test repositories to the latest snapshot.
2) Enables GPG check for Fedora snapshots.

\[1\]: `ostree.sh` cannot use RPMRepo, see [this comment](https://github.com/osbuild/osbuild-composer/pull/1068#issuecomment-724128216)


larskarlitski (dismissed):
Nice work!

@gicmo do the `koi-compose.py` changes look ok to you?

gicmo (approved):
Cool!

ondrejbudai:
I messed up the `koji-compose.py` change a bit, so I had to force-push a fix.

larskarlitski (approved):
> I messed up the koji-compose.py change a bit, so I had to force-push a fix.

Ah, `arch`, of course.

ondrejbudai:
Sigh, `$ARCH` wasn't defined in `koji.sh`.

Also, I had to revert the changes in `ostree.sh`. The $BOOT_LOCATION isn't actually an URL to a RPM repository but an URL of an installable tree. More information can be found here\[1\]\[2\]. As RPMRepo doesn't currently mirror the installable tree part of a repository and `.treeinfo` file, we cannot use it here.

Closing remarks: I've just discovered instalable trees so I have very little understanding of them. I just for sure know that we cannot currently use RPMRepo for `ostree.sh`.

\[1\]: https://release-engineering.github.io/productmd/treeinfo-1.0.html
\[2\]: https://docs.pulpproject.org/en/2.9/plugins/pulp_rpm/tech-reference/rpm.html#composes

larskarlitski (approved):
Bummer about using rpmrepo for ostree tests. Thanks for fixing it up!

PR #1070, opened 2020-11-10
distribution: remove osbuild-composer-koji

larskarlitski:
Remove both the package osbuild-composer-koji, and the only file it
shipped: osbuild-composer-koji.socket.

It's been deprecated since 835b556, but the backwards-compatible
solution in that commit never worked, because osbuild-composer only
checks for "osbuild-composer-api.socket" when starting up.

Since this has been meant to be deprecated for a while, just remove it
outright.

Add an "Obsoletes:" for the package, so that it gets uninstalled on
existing systems.

Box of shame

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -77,6 +77,9 @@ Obsoletes: lorax-composer < 34.3
 Obsoletes: golang-github-osbuild-composer < %{version}-%{release}
 Provides:  golang-github-osbuild-composer = %{version}-%{release}
 
+# remove when F34 is EOL
+Obsoletes: osbuild-composer-koji
+
 %description
 %{common_description}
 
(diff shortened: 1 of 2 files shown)


osbuild-composer.spec
```diff
@@ -77,6 +77,9 @@ Obsoletes: lorax-composer < 34.3
 Obsoletes: golang-github-osbuild-composer < %{version}-%{release}
 Provides:  golang-github-osbuild-composer = %{version}-%{release}
 
+# remove in F35
```
ondrejbudai, line 80: As we use the same spec file for every distro, I suggest `remove when F34 is EOL`. I think this is more obvious to future readers.

teg (dismissed):
Thanks! I was fighting with this last night, didn't get as far as figuring out what was going on ;)

ondrejbudai (commented):
Nice, just a little nit-pick.

osbuild-composer.spec
larskarlitski: Makes sense. Thanks!

larskarlitski:
Update: address @ondrejbudai's comment and remove `osubild-composer-test`'s dependency on `osbuild-composer-koji`.

ondrejbudai (approved):
Looks good now, thanks!

PR #1075, opened 2020-11-11
schutzbot: don't copy dnf repo file between stages

larskarlitski:
Now that the repository URLs are predictable, don't use Jenkins' stash
feature to pass the repo file between stages.

Instead, simply create the repo file where it is needed, in deploy.sh.

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -51,10 +51,6 @@ pipeline {
                         retry(3) {
                             sh "schutzbot/mockbuild.sh"
                         }
-                        stash (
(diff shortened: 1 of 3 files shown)


ondrejbudai (approved):
Awesome, thanks!

teg (approved):
Nice! I love how the mock stage becomes more independent.

larskarlitski:
Forgot a `sudo` and added a commit that exists early when the repository already exists.

teg (approved):

ondrejbudai (approved):

PR #1078, opened 2020-11-12
fsjobqueue: accept jobs of any type

ondrejbudai:
Soon, we want to begin tagging the jobs with the name of its submitter. The simplest way to add a tag to a job is to put it into its type string. However, as we don't know (and don't want to know) the submitters' names when osbuild-composer is initialized, we need to be able to push arbitrary job types into the jobqueue.

This commit therefore lifts the restriction that a jobqueue accepts only a predefined set of job types. Now, jobqueue clients can push jobs of arbitrary names.

### DIFF
--- a/internal/jobqueue/fsjobqueue/fsjobqueue.go
+++ b/internal/jobqueue/fsjobqueue/fsjobqueue.go
@@ -59,6 +59,10 @@ type job struct {
 	Canceled bool `json:"canceled,omitempty"`
 }
 
+// The size of channels used in fsJobQueue for queueing jobs.
+// Note that each job type has its own queue.
+const channelSize = 100
+
 // Create a new fsJobQueue object for `dir`. This object must have exclusive
 // access to `dir`. If `dir` contains jobs created from previous runs, they are
 // loaded and rescheduled to run if necessary.
@@ -70,7 +74,7 @@ func New(dir string, acceptedJobTypes []string) (*fsJobQueue, error) {
 	}
 
 	for _, jt := range acceptedJobTypes {
-		q.pending[jt] = make(chan uuid.UUID, 100)
+		q.pending[jt] = make(chan uuid.UUID, channelSize)
 	}
 
 	// Look for jobs that are still pending and build the dependant map.
--- a/internal/jobqueue/fsjobqueue/fsjobqueue.go
+++ b/internal/jobqueue/fsjobqueue/fsjobqueue.go
@@ -66,15 +66,11 @@ const channelSize = 100
 // Create a new fsJobQueue object for `dir`. This object must have exclusive
 // access to `dir`. If `dir` contains jobs created from previous runs, they are
 // loaded and rescheduled to run if necessary.
-func New(dir string, acceptedJobTypes []string) (*fsJobQueue, error) {
+func New(dir string) (*fsJobQueue, error) {
 	q := &fsJobQueue{
 		db:         jsondb.New(dir, 0600),
 		pending:    make(map[string]chan uuid.UUID),
 		dependants: make(map[uuid.UUID][]uuid.UUID),
-	}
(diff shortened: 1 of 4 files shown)


internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -59,6 +59,10 @@ type job struct {
 	Canceled bool `json:"canceled,omitempty"`
 }
 
+// The size of channels used in fsJobQueue for queueing jobs.
+// Note that each job type has its own queue.
+const channelSize = 100
```
teg, line 64: What happens when we reach this limit. Maybe that's a question for @larskarlitski ;)

teg (dismissed):
Very nice! We currently have a `testjobqueue` pacakge as well. I discussed this with @larskarlitski and we agreed that it is not needed, we should just allocate a temp dir and use `fsjobqueue` on that in our tests instead. That would avoid us having to adapt that too and get some more testing coverage while at it.

For a follow-up.

ondrejbudai:
> Very nice! We currently have a `testjobqueue` pacakge as well. I discussed this with @larskarlitski and we agreed that it is not needed, we should just allocate a temp dir and use `fsjobqueue` on that in our tests instead. That would avoid us having to adapt that too and get some more testing coverage while at it.
> 
> For a follow-up.

`testjobqueue` doesn't have restricted job types so no change is needed there.

larskarlitski:
> `testjobqueue` doesn't have restricted job types so no change is needed there.

It does not indeed, because it is not meant to be thread-safe, like `fsjobqueue`.

internal/jobqueue/fsjobqueue/fsjobqueue.go
larskarlitski: Bad things. (Enqueuing will hang forever.)

This has always been the case. Ondřej just moved it to a constant.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -321,7 +325,8 @@ func (q *fsJobQueue) maybeEnqueue(j *job, updateDependants bool) error {
 	if depsFinished {
 		c, exists := q.pending[j.Type]
 		if !exists {
-			return fmt.Errorf("this queue doesn't accept job type '%s'", j.Type)
+			c = make(chan uuid.UUID, channelSize)
+			q.pending[j.Type] = c
```
larskarlitski, line 329: This needs the lock (`q.mu`) held. We do hold it when calling `maybeEnqueue` from `Enqueue` and `FinishJob`, but not in `New`. That's ok, because nothing can run in parallel when the object is just being constructed. However, I think it will make it easier for people reading this code to understand it if we clearly state that the lock must be held during the duration of `maybeEnqueue`.

 (Are Golang's locks recursive?)

larskarlitski (changes requested):
A. Maz. Ing.

Thanks so much for this! One tiny nit.

internal/jobqueue/fsjobqueue/fsjobqueue.go
ondrejbudai: I added a comment to address this. Thanks!

larskarlitski: Perfect!

larskarlitski (approved):
Thanks! Great work on this.

internal/jobqueue/fsjobqueue/fsjobqueue.go
teg: Yeah, this is nothing to address in this PR, but we probably need to. Or at least bump it by a factor of ten or something?

teg (approved):

PR #1082, opened 2020-11-12
spec: fix the osbuild dependency

ondrejbudai:
osbuild-composer doesn't actually require osbuild. osbuild-composer-worker does. Let's remove the dependency from osbuild-composer and depend on the right version of osbuild in the worker.


teg (approved):
Perfect!

PR #1083, opened 2020-11-12
workerapi: serialize koji errors as stringss

teg:
Serializing an interface does not work, let us simply use the string
representation and treat the empty string as no error. This is
compatible with the current API in the success case, and fixes the
error case, which is currently broken.

Also extend the test matrix for the kojiapi to ensure that all the
different kinds of errors can be serialized correctly and leads to
the correct status being returned.


larskarlitski (approved):
Thanks for fixing this. Looks reasonable to me.

ondrejbudai (approved):
Awesome, thanks for a quick fix!

PR #1085, opened 2020-11-14
schutzbot/mockbuild: test against osbuild in target release by default

teg:
To make sure we test against what we will release with, osbuild from the
target distro.

The only exception is in RHEL8.3, where we are not rebasing any longer,
but expect users who use new osbulid-composer from upstream to use it
with new osbuild from upstream too. In this case, use the RPM from
osbuild upstream, for now pinned to the same version as was in the
submodule.

This introduces a new configuration file: Schutzfile, which is meant
to contain the full test-matrix. For now it only points to the upstream
osbuild commit to test a given distro against.

The submodule is now unused and is therefore removed.


larskarlitski (approved):
I really like were this is going. Thanks!

PR #1087, opened 2020-11-16
Allow osbuild-composer to run without Weldr API

ondrejbudai:
I think we should be able to ditch the local worker altogether. This is the first step. See the individual comments for details.

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -156,6 +156,15 @@ func (c *Composer) InitRemoteWorkers(cert, key string, l net.Listener) error {
 //
 // Running without the weldr API is currently not supported.
 func (c *Composer) Start() error {
+	// sanity checks
+	if c.localWorkerListener == nil && c.workerListener == nil {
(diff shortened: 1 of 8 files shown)


cmd/osbuild-composer/composer.go
```diff
@@ -200,7 +199,17 @@ func (c *Composer) Start() error {
...
+		go func() {
+			err := c.weldr.Serve(c.weldrListener)
+			if err != nil {
+				panic(err)
+			}
+		}()
+	}
+
+	// wait indefinitely
+	select {}
```
larskarlitski, line 221: Note: we probably want to stop when any of the APIs `panic`, which is easily solvable by passing a channel to each of the above goroutines. Out of scope for this PR.

larskarlitski (approved):
Nice! I really like this. Thanks!

cmd/osbuild-composer/composer.go
ondrejbudai: I think that's the default Go behaviour: https://play.golang.org/p/pL4FVhEZ3-c

teg (approved):
Great stuff! It is a bit fiddly to get both the RPM and systemd behavior backwards compatible, but this seems correct. Thanks!

PR #1088, opened 2020-11-16
mockbuild: use mock to build source rpm

larskarlitski:
Running `make srpm` on the host can have different results from building
the SRPM in mock.

Stop using the Makefile altogether for this, which is meant for
developer convenience only. The interface is clear: a tarball generated
from `git archive` and defining the `commit` variable. Less stuff to
install on the host, too.

Also includes a commit simplifying the way we `make rpm`.


larskarlitski:
Update to make `spellcheck` happy.

teg (approved):
nice!

PR #1090, opened 2020-11-19
worker/server: move it to the style of koji server

ondrejbudai:
The previous code was smelling a bit (e.g. Server.server field) so I decided
to rewrite it in the style of the much nicer koji server.

Not a functional change.


teg (approved):
Nice!

PR #1093, opened 2020-11-19
worker/osbuild: print raw osbuild output if it cannot be parsed

ondrejbudai:
osbuild output should be always JSON-parsable. However, if a user has a weird installation of osbuild, it can return malformed JSON. In this case, it's very hard to debug what happened because the worker doesn't provide any useful information. This commit adds the non-parsable osbuild output to the error so there's at least some clue in the logs what happened.

Let me reiterate: In 99% of these cases, this shouldn't be needed. It will give us some hints in these 1% cases though.

You may want to ask if using `json.Decoder` isn't better because it doesn't buffer the input. However, it does buffer. Firstly, `json.Decoder.Decode()` buffers the whole JSON value (it contains kinda cool state machine to do it) and in the second run, it parses the buffer. Therefore, the only thing that this commit actually does is that it moves the buffer out of `json.Decoder`.

### DIFF
--- a/cmd/osbuild-worker/osbuild.go
+++ b/cmd/osbuild-worker/osbuild.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


larskarlitski (changes requested):
I like this and I agree with your reasoning that buffering is fine. Manifests need to fit into memory anyway, because osbuild loads them, too.

However, I think you accidentally added an extraneous file `register.sh`.

ondrejbudai:
indeed, fixed, thanks!

larskarlitski (approved):
Thank you!

teg (approved):

PR #1095, opened 2020-11-22
Add ability to run reverse dependency tests, and run one for koji-osbuild

larskarlitski:
Now that koji-osbuild pushes rpms for (most) commits in the same format as osbuild and osbuild-composer, it is trivial to add reverse dependency tests for it. 🎈

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -164,6 +164,17 @@ pipeline {
                         }
                     }
                 }
+                stage('F32: koji-osbuild') {
+                    agent { label "f32cloudbase && x86_64 && aws" }
+                    steps {
+                        run_project_tests('koji-osbuild', 'integration.sh')
+                    }
+                    post {
+                        always {
+                            preserve_logs('fedora32-revdep-koji-osbuild')
+                        }
+                    }
+                }
                 stage('F33 Base') {
                     agent { label "f33cloudbase && x86_64 && aws" }
                     environment { TEST_TYPE = "base" }
@@ -288,6 +299,17 @@ pipeline {
                     post {
                         always {
                             preserve_logs('rhel8-ostree')
+                        }
+                    }
+                }
+                stage('EL8: koji-osbuild') {
+                    agent { label "rhel8cloudbase && psi && x86_64" }
+                    steps {
+                        run_project_tests('koji-osbuild', 'integration.sh')
+                    }
+                    post {
+                        always {
+                            preserve_logs('rhel8-revdep-koji-osbuild')
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


schutzbot/Jenkinsfile
```diff
@@ -466,7 +510,14 @@ void run_tests(test_type, boot_type) {
...
         )
     }
+}
 
+// Similar to run_tests, but with a more general signature: allows setting the
+// project whose -tests package to install, and the name of the test to execute
+void run_project_tests(project, test) {
+    sh "schutzbot/ci_details.sh"
+    sh "schutzbot/deploy.sh ${project}"
+    sh "/usr/libexec/tests/koji-osbuild/${test}"
```
gicmo, line 520: Use $project instead of koji-osbuild?

gicmo (dismissed):
:heart: This is super cool

larskarlitski:
Force-pushed to remove fedora-33 and rhel-8.4 tests, because upstream doesn't yet build packages for those. Thanks @gicmo!

larskarlitski:
Update to include @gicmo's fix from https://github.com/osbuild/koji-osbuild/pull/50, which stops pulling containers from docker.io.

This is ready to be reviewed now.

gicmo (dismissed):
😍

schutzbot/Jenkinsfile
```diff
@@ -164,6 +164,17 @@ pipeline {
                         }
                     }
                 }
+                stage('F32: koji-osbuild') {
+                    agent { label "f32cloudbase && x86_64" }
```
ondrejbudai, line 168: Does it use stuff from RH internal repos or nested virt? If not, I suggest adding `&& aws` so we don't use more resource from PSI OpenStack than we really need. If it does, `&& psi` is needed.

schutzbot/Jenkinsfile
```diff
@@ -291,6 +302,17 @@ pipeline {
                         }
                     }
                 }
+                stage('EL8: koji-osbuild') {
+                    agent { label "rhel8cloudbase && psi && x86_64" }
```
ondrejbudai, line 306: The same as in the previous comment.

ondrejbudai (changes requested):
Very nice, I have just concerns about image labelling in Jenkinsfile.

schutzbot/Jenkinsfile
gicmo: The Fedora tests should not need anything from the internal repos IIRC.

schutzbot/Jenkinsfile
gicmo: The RHEL tests do need access, and I think `psi` is already present here, no?

schutzbot/Jenkinsfile
larskarlitski: Ah, I just removed the `psi` here (copied from the stage above), but didn't think to add `aws` in. Pushed a fix. Thanks!

schutzbot/Jenkinsfile
larskarlitski: Yep, leaving as is with `psi` in there then.

gicmo (approved):
:balloon:

ondrejbudai (approved):
Awesome!

schutzbot/Jenkinsfile
ondrejbudai: Thanks!

Issue #983, opened 2020-09-14
Make composer easier to run in a container


major:
Running osbuild-composer in a container (especially on OpenShift) has some benefits:

- OpenShift can verify that composer started properly and re-verify that it is working properly on an interval
- We can ensure that the OS + dependencies + composer are identical each time and we can test them together
- Once there's a shared database, we can easily scale composer up/down based on our needs

However, composer's startup is heavily tied to systemd's socket activation. That is not available in containers on OpenShift and we need a way to start the daemon and have it listen on a port.

teg:
Absolutely. We should figure out exactly what this would take and how to test it. Happy to discuss that. Not sure we'll have the resources to schedule this work in the immediate future, but definitely high on the list.

dvdhrm:
This is "easy" to do with `ubi-init`, *IF* you run it on modern podman, that is >=OpenShift-4.

teg:
@major did we conclude that OpenShift-4 is not an option? so we have to run this in a `ubi` container?

major:
From what I understand, you still need to run privileged containers (which isn't possible). I also hunted around and can't find anyone using ubi-init on OpenShift 4 on clouddot right now.

ondrejbudai:
Fixed in  #1097

ondrejbudai:
Ah, it isn't merged yet, reopening this.

dvdhrm:
There is now a base container in the repository, filed as `./containers/osbuild-composer/`. While there might be adjustments necessary, this does allow running composer in a container with its full feature set.

Closing this now. For more specific requests, we can start opening new issues.

PR #1097, opened 2020-11-23
containers: add containerized osbuild-composer

dvdhrm:
Add a Dockerfile that creates a container based on Fedora with
osbuild-composer deployed. Create a suitable entrypoint that runs
osbuild-composer in the container and creates the required sockets
without systemd.

To test this, build the container via:

    docker build ./containers/osbuild-composer

Then create your certificates in /etc/osbuild-composer/. Then run
composer with something like:

    docker run --rm -v /etc/osbuild-composer:/etc/osbuild-composer <id>

(Where <id> is the container ID returned by `docker build`.)

Note that this container is not yet hooked up to anything. Furthermore,
it might need adjustments to be ready for deployment. This is mostly a
proof-of-concept that shows how to deploy Composer in a container.

### DIFF
--- a/containers/osbuild-composer/Dockerfile
+++ b/containers/osbuild-composer/Dockerfile
@@ -1,0 +1,71 @@
+#
+# osbuild-composer - Containerized OSBuild Composer
+#
+# This container provides a minimal fedora image with the osbuild-composer
+# application installed and configured as default entrypoint.
+#
+# Build Arguments:
+#
+#   * OSB_FROM
+#       This specifies the host image to use. It must be an RPM-based
+#       distribution image with all osbuild-composer requirements
+#       pre-installed.
+#
+#       Example: "docker.io/library/fedora:latest"
+#
+#   * OSB_RPMREPO
+#       Base URL of an RPM repository from which to install osbuild-composer
+#       from.
+#
+#       Example: "https://dl01.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/"
+#
+
+# Image arguments must be imported before `FROM`.
+ARG             OSB_FROM="docker.io/library/fedora:latest"
+
+# Prepare our host environment.
+FROM            "${OSB_FROM}" AS base
+
+# Import build parameters.
+ARG             OSB_RPMREPO="https://dl01.fedoraproject.org/pub/fedora/linux/releases/\$releasever/Everything/\$basearch/os/"
+
+# Create our state directory and use it as anchor.
+WORKDIR         "/var/lib/osb"
+
+# Create and switch into our src directory, which we use as temporary storage
+# for all sources during the install.
+WORKDIR         "./src"
+
+# Install all global dependencies.
+RUN \
+                dnf \
+                        -y \
+                        "--repofrompath=ephemeral0,${OSB_RPMREPO}" \
+                        "--setopt=ephemeral0.gpgcheck=0" \
+                        "--setopt=ephemeral0.priority=10" \
+                        install "osbuild-composer" \
+                && dnf clean all
+
+# Copy all our local sources, so we can access them from within the container
+# build. They will be cleaned in a later step.
+COPY            "." "."
+
+# Prepare the runtime configuration and state.
+RUN             mkdir -p "../bin"
+RUN             mkdir -p "/etc/osbuild-composer/"
+RUN             mkdir -p "/run/osbuild-composer/"
+RUN             mkdir -p "/run/weldr/"
+RUN             mkdir -p "/var/cache/osbuild-composer/"
+RUN             mkdir -p "/var/cache/osbuild-worker/"
+RUN             mkdir -p "/var/lib/osbuild-composer/"
+
+# Install all required sources into the persistent directory.
+RUN             cp "entrypoint.py" "../bin/"
+
+# Leave and delete our temporary source directory.
+WORKDIR         ".."
+RUN             rm -rf "./src"
+
+# Prepare the runtime entrypoint and empty working directory.
+WORKDIR         "./workdir"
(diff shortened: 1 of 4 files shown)


containers/osbuild-composer/entrypoint.py
```diff
@@ -0,0 +1,136 @@
...
+    assert(sock.fileno() == index)
+    index += 1
+
+    # osbuild-composer.socket:1
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    sock.bind("/run/osbuild-composer/job.socket")
+    sock.listen(256)
+    sockets.append(sock)
+    fds.append(sock.fileno())
+    names.append("osbuild-composer.socket")
```
larskarlitski, line 61: I don't think exposing the weldr API makes sense in container deployments. We could make it configurable, defaulting to off. I wouldn't mind removing it altogether, either.

containers/osbuild-composer/entrypoint.py
```diff
@@ -0,0 +1,136 @@
...
+            cmd,
+            stdin=subprocess.DEVNULL,
+            stderr=subprocess.STDOUT,
+            pass_fds=fds,
+            preexec_fn=preexec_setenv,
+        )
+        return proc.wait()
+    except:
+        proc.terminate()
+        raise
```
larskarlitski, line 124: I'm not sure this will be a common use case, but I'm always getting a 23-line stack trace when stopping the container with `Ctrl-C`.

We could not re-raise the exception on `KeyboardInterrupt` and `SIGTERM`. Can be in a follow-up though.

larskarlitski (commented):
It works! ❤️ 

Really really cool. I'm a bit hesitant to add this without testing it in CI. If it's in the repository, people might start using it... I wouldn't block on it though if other feel differently.

ondrejbudai (changes requested):
https://github.com/osbuild/osbuild-composer/commit/dc1b84fcfeb946a1a0553c7af9df0c448cacc21a split the local worker socket and the weldr socket so this PR needs some adjustments. Sorry, David!

This leads to an interesting question: The image built by the new `Dockerfile` is tight to a downstream version of osbuild-composer. However, it's in the upstream repository. This will surely lead to some funny situations. We even have one right now - see the first paragraph. Not sure how to solve it though...

msehnout:
> [dc1b84f](https://github.com/osbuild/osbuild-composer/commit/dc1b84fcfeb946a1a0553c7af9df0c448cacc21a) split the local worker socket and the weldr socket so this PR needs some adjustments. Sorry, David!
> 
> This leads to an interesting question: The image built by the new `Dockerfile` is tight to a downstream version of osbuild-composer. However, it's in the upstream repository. This will surely lead to some funny situations. We even have one right now - see the first paragraph. Not sure how to solve it though...

Good point! Although I understand the reasoning behind this approach, I'm also afraid this is not an expected behavior from a Dockerfile. By default I expect the Dockerfile in a repository checkout to use the source code from the checkout.

What about using our RPM repo in S3? Wouldn't it be at least slightly better? You could push a PR and then use it.

containers/osbuild-composer/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+#
+# Build Arguments:
+#
+#   * OSB_HOST="docker.io/library/fedora:latest"
+#       This specifies the host image to use. It must be an RPM-based
+#       distribution image with `dnf` ready for use.
+#
+
+# Image arguments must be imported before `FROM`.
+ARG             OSB_HOST="docker.io/library/fedora:latest"
```
msehnout, line 17: Wouldn't it be better to use `33` instead of `latest`?

dvdhrm:
I updated the container with the provided suggestions. A few more notes:

This is mostly a proof-of-concept for now, but allows for integration into our build-infrastructure. Before we merge this, I want to add an automatic build for this container which adds the RPM-repository we create for every commit and then pulls in composer via those RPMs. Hence, even though this looks like it pulls in composer from downstream, this is not how it will be built in our pipeline. However, keeping the dockerfile like this allows building it locally as well.

teg:
After discussing this with @dvdhrm I think this makes sense and should be merged so we can work on top of it. However, I do think @msehnout and @ondrejbudai raise some good questions.

The way this Dockerfile is currently written is independent of a given checkout/version of osbuild-composer. To the extent that it might as well live (with the entrypoint) in a different repository alltogether. In practice though, @dvdhrm explained to me that the idea would be to override the `FROM` argument with a fixed base, and to insert our repositories with a given osbuild-composer version so we also get a fixed version of our software.

This is neat in the sense that the Dockerfile is independent and can be used as-is, giving you the most recent fedora with the most recently released version of composer. However, in practice, I think we should treat this more like our spec file: only support the version of composer from the commit of the Dockerfile, even if it needs to be passed in manually. I would also, for the avoidance of doubt, require the FROM to be passed in explicitly without a default.

Lastly, I am not a huge fan of the entrypoint logic. I think we should merge it as-is, but I wonder if it would not be nicer to just move this to composer itself, that would definitely be something for a follow-up though. Let's first start producing containers, then bikeshed later ;)

ondrejbudai (approved):
I usually don't agree with "merge first, fix later" approach but in this case, I think it makes sense. It's a new way to distribute our project and it doesn't change the way we build RPMs - it rather complements it. I just wonder if we should add a note in the `Dockerfile` that it's still a tech-preview and it doesn't have any backward compatibility promises yet...

ondrejbudai:
Please, close #983 when this is merged.

dvdhrm:
I updated the PR and it now builds the containers in CI. It does not push them, yet, but this would just be a matter of adding `podman push <...>` after the `podman build` invocation.

teg (approved):
Nice! Thanks for this! The koji failure in CI is unrelated.

PR #1098, opened 2020-11-23
cloudapi: Share an ec2 snapshot/ami with an account

Gundersanne:
Share an ec2 snapshot and ami with other aws accounts.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -160,7 +160,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			}
 
 			/* TODO: communicate back the AMI */
-			_, err = a.Register(t.ImageName, options.Bucket, key)
+			_, err = a.Register(t.ImageName, options.Bucket, key, options.ShareWithAccounts)
 			if err != nil {
 				r = append(r, err)
 				continue
--- a/internal/cloudapi/openapi.yml
+++ b/internal/cloudapi/openapi.yml
@@ -203,6 +203,11 @@ components:
         snapshot_name:
           type: string
           example: 'my-snapshot'
+        share_with_accounts:
+          type: array
+          example: ['123456789012']
+          items:
+            type: string
     Customizations:
       type: object
       properties:
--- a/cmd/osbuild-upload-aws/main.go
+++ b/cmd/osbuild-upload-aws/main.go
@@ -16,6 +16,7 @@ func main() {
 	var keyName string
 	var filename string
 	var imageName string
+	var shareWith string
 	flag.StringVar(&accessKeyID, "access-key-id", "", "access key ID")
 	flag.StringVar(&secretAccessKey, "secret-access-key", "", "secret access key")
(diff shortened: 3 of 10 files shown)


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -159,8 +159,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				continue
 			}
 
-			/* TODO: communicate back the AMI */
```
teg, line 162: But, the TODO isn't done ;)

internal/cloudapi/openapi.yml
```diff
@@ -203,6 +204,11 @@ components:
         snapshot_name:
           type: string
           example: 'my-snapshot'
+        share_with_accounts:
+          type: array
+          example: ['123456789012']
+          items:
+            type: string
```
teg, line 211: Should this be numbers rather than strings?

teg (changes requested):
Love this! A tiny nit (please put back the TODO), and a question.

cmd/osbuild-worker/jobimpl-osbuild.go
Gundersanne: Is it still needed though? Like now the thing shares?

internal/cloudapi/openapi.yml
Gundersanne: Hm for a number we'd have to convert it back to a string to call the aws api. And who knows we might wanna accept account aliases or something at some point?

internal/cloudapi/openapi.yml
```diff
@@ -192,6 +192,7 @@ components:
       required:
         - access_key_id
         - secret_access_key
+        - share_with_accounts
```
Gundersanne, line 195: Hm one thing to consider is if this should be a required property, wdyt @teg ?

cmd/osbuild-worker/jobimpl-osbuild.go
teg: Yeah, we still need to tell the caller which AMI was shared with them.

internal/cloudapi/openapi.yml
teg: Fair enough, if this is the API there is no reason for us to artificially restrict it.

internal/cloudapi/openapi.yml
teg: I think it should be optional as it is now. From image-builder it would always be used, but we could have other services consume this API that does not use the same model (some CI service, for instance?).

teg (approved):

internal/cloudapi/openapi.yml
Gundersanne: Ah, in that case i'll repush tomorrow without it being a required property. Right now it will return bad request if you omit it.

Gundersanne:
@teg I removed the required bit, can you ack again?

internal/cloudapi/openapi.yml
teg: Thanks!

teg (approved):
Perfect!

PR #1099, opened 2020-11-23
tools/deploy-qemu: add macOS support 

larskarlitski:
Also, add a `QEMU_EXTRA_ARGS` environment variable for people to set arguments from their `bashrc` (or similar) that better suit their workflow.

### DIFF
--- a/tools/gen-user-data
+++ b/tools/gen-user-data
@@ -20,17 +20,15 @@ The configuration directory may contain:
                    to this directore (`files/etc/hosts` → `/etc/hosts`). Its
                    permissions are copied over, but the owner will always be
                    root:root. Empty directories are ignored.
-
-The `python3-pyyaml` package is required to run this tool.
 """
 
 
 import argparse
 import base64
+import json
 import os
 import stat
 import sys
-import yaml
 
 
 def octal_mode_string(mode):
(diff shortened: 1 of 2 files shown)


ondrejbudai (approved):
I tested this on Fedora and it still works like a charm!

tools/gen-user-data
```diff
@@ -48,27 +46,26 @@ def main():
...
                 "content": content,
                 "permissions": octal_mode_string(os.lstat(path).st_mode)
             })
 
-    sys.stdout.write("#cloud-config\n")
-    yaml.dump(userdata, sys.stdout, Dumper=yaml.SafeDumper)
+    with open(f"{args.configdir}/user-data.yml") as f:
+        sys.stdout.write(f.read())
+    sys.stdout.write("write_files: ")
+    json.dump(write_files, sys.stdout)
```
msehnout, line 67: Do I read this correctly that an JSON oneliner is a valid array in YAML?

larskarlitski: Yes, as of YAML 1.2, it is a superset of JSON. The [spec](https://yaml.org/spec/1.2/spec.html) states that the "primary objective of this revision is to bring YAML into compliance with JSON as an official subset".

PR #1102, opened 2020-11-24
Add an owner to a job and allow workers to filter jobs by owners

ondrejbudai:
This PR needs some tests but the basic gist is ready to be reviewed. It consists of several parts:

1) `arch` is merged in the job type.
2) Job queue no longer works with the concept of a job type, it instead uses a queue name. Queue name can be constructed from a job type and vice versa.
3) Jobs now have a owner. Queue names are now constructed in a backward compatible way from a job type and an owner. The inverse function now also has the support for job owners.
4) The worker can now accept jobs only from a pre-selected set of owners (configurable).
5) Jobs created by Koji and Cloud APIs have now owners based on the SAN field of the certificate that was used to create the job.
6) Jobs created by Weldr API have now _weldr owner. Unconfigured workers fall back to jobs owned by _weldr.

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -151,11 +151,11 @@ func main() {
 	}
 
 	jobImpls := map[string]JobImplementation{
-		"osbuild": &OSBuildJobImpl{
+		"osbuild:" + common.CurrentArch(): &OSBuildJobImpl{
 			Store:       store,
 			KojiServers: kojiServers,
 		},
-		"osbuild-koji": &OSBuildKojiJobImpl{
+		"osbuild-koji:" + common.CurrentArch(): &OSBuildKojiJobImpl{
 			Store:       store,
 			KojiServers: kojiServers,
 		},
@@ -174,7 +174,7 @@ func main() {
 
 	for {
 		fmt.Println("Waiting for a new job...")
-		job, err := client.RequestJob(acceptedJobTypes, common.CurrentArch())
+		job, err := client.RequestJob(acceptedJobTypes)
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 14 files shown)


internal/kojiapi/server.go
```diff
@@ -60,6 +60,17 @@ type apiHandlers struct {
...
+	tls := ctx.Request().TLS
+	if tls == nil {
+		panic("kojiapi must be always run on TLS listener")
+
+	}
+	if tls.PeerCertificates == nil || len(tls.PeerCertificates) > 0 {
+		panic("kojiapi must be always run with required peer certificates")
+	}
+	// take the job owner name from the first SAN name
+	jobOwner := (*tls.PeerCertificates[0]).DNSNames[0]
```
teg, line 72: Are we guaranteed that this is set? We no longer support CN? (I would not mind deprecating CN, but should make sure our our issued certs are up-to-date).

cmd/osbuild-worker/main.go
```diff
@@ -114,6 +114,10 @@ func main() {
 		log.Fatalf("Could not load config file '%s': %v", configFile, err)
 	}
 
+	if len(config.AcceptedJobOwners) == 0 {
+		config.AcceptedJobOwners = []string{"_weldr"}
```
teg, line 118: Why not just leave it without an owner? Won't unconfigured workers pick it up then?

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +84,7 @@ func main() {
 				KeyTab    string `toml:"keytab"`
 			} `toml:"kerberos,omitempty"`
 		} `toml:"koji"`
+		AcceptedJobOwners []string `toml:"accepted_job_owners,omitempty"`
```
teg, line 87: Not sure about the naming of this. But have no good suggestions...

teg:
This is great! Had some minor nits, and CI is not happy. But apart from that: happy to merge it!

PR #1104, opened 2020-11-25
distro/rhel84: don't install uefi packages on all arches

### LINKED ISSUE #1103: rhel-84 builds fail on non-x86 arches
```
GenericError: Failed to create the compose request: {"message":"Failed to depsolve base base packages for qcow2/s390x/rhel-84: DNF error occured: MarkingErrors: Error occurred when marking packages for installation: Problems in request:\nmissing packages: efi-filesystem, efivar, efivar-libs, grub2-efi-x64, shim-x64"}
```

image

larskarlitski:
These packages are architecture-specific. Including them in the generic
package list broke image builds for everything but x86_64.

The corresponding architecture-dependant packages were already included
in `arch.bootloaderPackages` (some via dependencies).

Fixes #1103


teg (approved):
Which arches was this built on? Is it correct that the test-cases should be unchanged?

larskarlitski:
> Which arches was this built on? Is it correct that the test-cases should be unchanged?

I am testing this right now on s390x, but wanted to start CI in parallel. Test cases should be unchanged, because the same packages get pulled in in the end.

gicmo:
For the record, `efivars` is required by `dbxtool`, which is required by `shim`. The latter also requires `efi-filesystem` (which is also directly required by `grub2-efi`).

larskarlitski:
I have just verified building an image on a s390x machine manually. It works as expected.

gicmo (approved):
:green_heart: Sorry for not catching that in the original PR :/

ondrejbudai (approved):

PR #1062, opened 2020-11-05
Enable gosec in golangci-lint

ondrejbudai:
The title says it all...

PR #1106, opened 2020-11-25
Golangci lint

msehnout:
Rebased #1062


internal/jsondb/db_test.go
```diff
@@ -49,7 +49,7 @@ func TestDegenerate(t *testing.T) {
 		db := jsondb.New(dir, 0755)
 
 		// write-only file
-		err = ioutil.WriteFile(path.Join(dir, "one.json"), []byte("{"), 0644)
+		err = ioutil.WriteFile(path.Join(dir, "one.json"), []byte("{"), 600)
```
teg, line 52: Dropping the leading `0` changes this from octal to decimal.

teg (changes requested):
Would love to merge this :) Found a minor issue.

internal/jsondb/db_test.go
msehnout: I was wondering the same. Was it intentional @ondrejbudai ?

ondrejbudai: I think that gosec didn't like `0644` and preferred `0600`. `600` is definitely a mistake.

PR #1111, opened 2020-11-25
cloudapi: add support for mirrorlist and metalink repos

chloenayon:
Previously, baseurl was required in openapi.yaml. In order to add support for metalink 
and mirrorlist repos as well, make all optional, since openapi does not support mutually 
exclusive parameters. Instead, enforce this logic in server.go, and if no repo has been 
specified, return a 400 bad request error.

### DIFF
--- a/internal/cloudapi/openapi.gen.go
+++ b/internal/cloudapi/openapi.gen.go
@@ -82,8 +82,10 @@ type ImageStatus struct {
 
 // Repository defines model for Repository.
 type Repository struct {
-	Baseurl string `json:"baseurl"`
-	Rhsm    bool   `json:"rhsm"`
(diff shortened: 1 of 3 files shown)


teg (approved):
Very nice! Thanks @chloenayon 🎉

ondrejbudai (approved):
> openapi does not support mutually exclusive parameters.

This is not entirely true, there's `oneOf`, see \[1\]. However, there's a big catch: It isn't supported by the code generator we use\[2\] and therefore, we cannot use it.

Big thanks, @chloenayon, great patch! :slightly_smiling_face: 

\[1\]: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
\[2\]: https://github.com/deepmap/oapi-codegen#whats-missing-or-incomplete

Issue #1113, opened 2020-11-26
Make cross-arch depsolving test smarter and add RHEL


msehnout:
https://github.com/osbuild/osbuild-composer/blob/69e78834214d808203ca3f2775aa731e2efae4f7/cmd/osbuild-dnf-json-tests/main_test.go#L55

Our current implementation of the depsolving test does not make sure it works for RHEL. It should be possible to pass the distro as an argument to the binary so that only RHEL machines run RHEL depsolving.

PR #1114, opened 2020-11-26
Run RHEL depsolving in dnf-json-tests

msehnout:
Fixes: https://github.com/osbuild/osbuild-composer/issues/1113

See the commit message as an up-to-date source of truth.


test/cases/depsolving.sh
```diff
@@ -0,0 +1,67 @@
...
+
+    test_divider
+    echo
+}
+
+# Use repository overrides because RHEL subscription does not allow usage of repositories
+# for multiple architectures (it provides content only for the host's architecture).
+sudo mkdir -p /etc/osbuild-composer/repositories
+sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-8.json /etc/osbuild-composer/repositories/
+sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-84.json /etc/osbuild-composer/repositories/
```
teg, line 43: I think this should be copied to `rhel-8.json` in the same way we do for other tests. Even when we support 8.3 and 8.4 as separate distros in composer, we only have the one set of repositories defined for RHEL8.

If you re-run the tests on each distro we support you'll also catch any libdnf compat issues, so that might be worth it, even though this test already would catch any issues with the packagelists in our image definitions.

teg (changes requested):

ondrejbudai:
Hmm, the depsolving tests for Fedora ran for 2 hours, and then the global pipeline timeout killed them. Do we know why they run for such a long time? This doesn't sound right to me.

ondrejbudai:
Also, a rebase would be nice to fix the `koji-osbuild` rev-dep test.

msehnout:
I think it is safe to say now that this is not going to happen :-)

PR #1115, opened 2020-11-26
distro/rhel84: no hybrid boot on aarch64

gicmo:
Hybrid boot, i.e. MBR+UEFI, is only supported on X86_64, but not on aarch64, which only supports UEFI. Therefore, disable legacy, i.e. MBR, boot for aarch64 and remove the corresponding BIOS boot partition.

Since hybrid boot is always used on X86_64, the "uefi" parameter is removed completely, together with the "legacy only" option on X86_64. Thus we are left with a simple one to one mapping of architecture to qemu assembler stage options in `qemuAssembler`.

### DIFF
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -49,7 +49,7 @@ type imageType struct {
 	bootable         bool
 	rpmOstree        bool
 	defaultSize      uint64
-	assembler        func(uefi bool, options distro.ImageOptions, arch distro.Arch) *osbuild.Assembler
+	assembler        func(options distro.ImageOptions, arch distro.Arch) *osbuild.Assembler
 }
 
 func (a *architecture) Distro() distro.Distro {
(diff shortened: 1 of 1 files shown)


larskarlitski (approved):
Yep! Thank you so much for this. So much simpler now.

teg (approved):

teg (approved):
Thanks for the quick rebase! And thanks for the fix and testing.

gicmo:
For the record, I manually checked `aarch64` and `s390x` (built an image and booted it) .

gicmo:
Manually also checked `ppc64le` now too (built an image and booted it).

teg:
The F32 and F33 openstack image tests failed with:
```
Cannot create instance: The service is currently unable to handle the request due to a temporary overloading or maintenance. This is a temporary condition. Try again later.
```
This is unrelated to this PR, which only touches RHEL.

teg:
There was an issue with fedora iot on F33, but again that is clearly unrelated to this PR. Merge to let CI re-run on `main`.

PR #1116, opened 2020-11-26
schutzbot: add required env variable to 8.4 integration test

larskarlitski:
22c9f6a introduced a new environment variable to api.sh,
AWS_API_TEST_SHARE_ACCOUNT, but only set it in the RHEL 8 tests.

Seems like the corresponding branch was not rebased to a version which
already had those 8.4 tests:

    https://github.com/osbuild/osbuild-composer/pull/1098


teg (approved):
Thanks!

gicmo (approved):
:+1: Defined variables are better than not defined variables, at least when you need them.

PR #1117, opened 2020-11-30
api/koji: return pending status until all jobs are finished

ondrejbudai:
Previously, the compose status returned failure as soon as possible. koji-osbuild considers the job as done when its status == failure and proceeds with uploading the logs to koji and marking the job as failed. However, not all osbuild-composer jobs might be done at this point so the logs might be incomplete making the debugging hard.

This commit changes the behaviour: Now, the compose status is pending until ALL jobs belonging to it are finished.

### DIFF
--- a/internal/kojiapi/server.go
+++ b/internal/kojiapi/server.go
@@ -224,6 +224,10 @@ func composeStatusFromJobStatus(js *worker.JobStatus, initResult *worker.KojiIni
 		return "failure"
 	}
 
+	if js.Finished.IsZero() {
+		return "pending"
+	}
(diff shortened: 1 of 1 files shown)


teg (dismissed):
Makes sense!

gicmo (dismissed):
Makes sense indeed!

ondrejbudai:
Rebased to add #1122.

teg (approved):

ondrejbudai:
F33 aarch64 flakey mirror, otherwise green. I will fix this later.

PR #1118, opened 2020-11-30
{koji,worker}/server: log errors returned from handlers

ondrejbudai:
Previously, we had no clue what errors were caught by the default echo's error handler. Thus, in the case of an error, we were basically blind. Let's log all errors so we can investigate them later.


teg (approved):

ondrejbudai:
Rebased to add #1122.

teg (approved):

PR #1120, opened 2020-11-30
distro/rhel8*: change the default size of qcow2 to 10 GiB

ondrejbudai:
The default size of RHEL 8 qcow2 images is 10 GiB, let's align our default.

Related: rhbz#1846087


teg:
Thanks for picking this up so quickly! I would only change the 8.4 images (no need to touch 8.3 unless there is a bug, I would say). And we need to update the test cases.

ondrejbudai:
Thanks for your quick review! I force-pushed a new version that changes only rhel-84 and I regenerated the test-cases.

gicmo (approved):
:moon_cake:

teg (dismissed):

ondrejbudai:
Rebased to add #1122.

gicmo (approved):
:ok_man:

PR #1121, opened 2020-11-30
drop the osbuild submodule

ondrejbudai:
64432c70 promised to remove it but this didn't actually happen. We don't use the submodule anymore so let's indeed drop it this time.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "osbuild"]
-	path = osbuild
-	url = https://github.com/osbuild/osbuild.git


teg (dismissed):
🎉 
(sorry for not delivering on my promise)

msehnout:
I love it, but is the CI error unrelated? It should be, but I'd like someone else to also take a look :)

ondrejbudai:
Rebased to add #1122.

teg (approved):
Thanks!

ondrejbudai:
EL8.4 Integration test hit the timeout, all other test cases are green, they would surely catch if there was something wrong with this PR.

PR #1123, opened 2020-12-01
tests: move the epel dependencies to provisioning

ondrejbudai:
koji and ansible are not in RHEL repositories. Depending on them breaks RHEL gating (see OSCI-1541): It tries to build a custom image with -tests package in it but in the build environment there's no EPEL.

This commit makes the RPM independent from EPEL. However, we still need koji and ansible, so the provision script now enables EPEL and installs the packages from there. This is not nice but we have to live with that until OSCI-1541 is solved.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -14,7 +14,10 @@
     },
     "dependants": {
       "koji-osbuild": {
-        "commit": "4fdc457745e1147475ea3ac1e3b073e592d7b174"
+        "commit": "4fdc457745e1147475ea3ac1e3b073e592d7b174",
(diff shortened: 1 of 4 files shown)


teg (approved):
Least bad compromise: 👍

ondrejbudai:
I needed to come up with a method for installing EPEL before `koji-osbuild-tests` in the rev dep test. See the first commit, it's a tiny bit ugly but it gets us there.

teg (approved):

ondrejbudai:
OpenStack upload failed, otherwise we're green!

PR #1126, opened 2020-12-03
tests: use a newer rpmrepo snapshot

ondrejbudai:
Also, this uses the new aarch64 snapshot for F33 that should reduce the amount
of CI fails.


msehnout:
This is included in #1114 . I'd personally prefer to merge that one.

ondrejbudai:
Ah, OK, I missed that, closing this one.

PR #1127, opened 2020-12-04
distro/rhel84: remove rng-tools

jkozol:
rng-tools is no longer included as a package in the RHEL 8.4 image. This package is both removed from being an included package and also specifically declared as excluded. The test manifests are updated.

I am unsure why rng-tools needs to be explicitly declared as excluded. But, when I just removed it from the included packages list, it was still getting added to the manifest. I tried running `dnf repoquery --whatdepends` to see what might be pulling it in but that didn't reveal anything. If anyone has thoughts on this I'd appreciate the info.

### DIFF
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -666,7 +666,7 @@ func New() distro.Distro {
 			"fwupd", "usbguard",
 			"bash-completion", "tmux",
 			"ima-evm-utils",
-			"audit", "rng-tools",
+			"audit",
 			"podman", "container-selinux", "skopeo", "criu",
 			"slirp4netns", "fuse-overlayfs",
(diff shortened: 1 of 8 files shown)


ondrejbudai (changes requested):
Thanks! The image tests don't like the changes though. Weird... Ping me if you need help investigating. :)

jkozol:
The manifests seemed to have dropped too many packages. I regenerated them and they seem to be fixed. I wasn't able to replicate the manifests generating incorrectly so I'm not sure what went wrong.

ondrejbudai (approved):
Looks good to me, thanks!

ondrejbudai:
openstack image test failed because of PSI Openstack outage, this is a known issue.

PR #1128, opened 2020-12-04
distro/rhel84: disable rhsmcertd.service

jkozol:
The rhsmcertd service is now disabled for all image types for the RHEL 8.4 guest image. The test manifests are updated.

I am opening this as a draft because alongside disabling the rhsmcertd service we need to set `enabled=0` in both `/etc/yum/pluginconf.d/product-id.conf` and `/etc/yum/pluginconf.d/subscription-manager.conf`. In the official kickstarter this is done with `sed`. @gicmo do you think this should be done in any particular osbuild stage?

### DIFF
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -287,7 +287,7 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 		p.AddStage(osbuild.NewUsersStage(options))
 	}
 
-	if services := c.GetServices(); services != nil || t.enabledServices != nil {
+	if services := c.GetServices(); services != nil || t.enabledServices != nil || t.disabledServices != nil {
 		p.AddStage(osbuild.NewSystemdStage(t.systemdStageOptions(t.enabledServices, t.disabledServices, services, t.defaultTarget)))
 	}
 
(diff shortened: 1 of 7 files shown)


teg:
I often argued against generic configuration stages in `osbuild`, but maybe it is time to revisit that stance *ducks*. The important point I always wanted to make was that we should not allow arbitrary modifications to the filesystem outside the control of RPM. However, the argument could be made that modifications to `/etc` are perfectly fine, and had `/etc` been structured as some well-defined key-value store of configuration overrides, I think this would be a no-brainer.

As it is, `/etc` is a mess of different more-or-less defined formats, and I think we only really have two choices:
 - require a custom osbuild stage to be added for every type of config we support editing
 - add the support for editing/replacing/adding configuration files in a generic way. I think we should consider this option (even though I have hesitated for the longest time, to avoid creating a mess).

I would be open to something like:
```
{
  "name": "org.osbuild.etc",
  "options": {
    "sed": [
      {
        "path": "foobar.conf",
        "script": "s/foo/bar/g"
      }
    ],
    "file": [
      {
        "path": "baz.conf",
        "contents": "\n\n\tTEST\n"
      }
    ]
  }
}
```

This applies the sed script to `/etc/foobar.conf` and creates the file `/etc/baz.conf`. Note that all filenames are relative to `/etc`.

I'll stress again that this is not something I'm hugely happy about, but I'm thinking that
 1. this is how `/etc` is meant to be used: you are meant as an admin to edit the config files in there by hand
 2. if we don't come up with some generic solution we'll be spending a lot of time and effort adding a million different stages to everything that comes up, with no clear delinietion of when we are "done". For now this is all about hard-coded configuration, but surely at some point users will want to make arbitrary changes too, and we'll need some non-turing-complete way of making that happen.

jkozol:
@teg Thank you for this thoughtful write up.

My only question is what sort of limits would we want to have? I guess the basic operations would be to:

- Add a new file
- Remove a file
- Add text to a file
- Replace/Remove text in a file

jkozol:
Thanks to conversations on rhel-devel facilitated by @thozza it has been determined that the rhsmcertd.service should be kept enabled for the rhel 8.4 kvm guest image.

teg:
Also, after several discussions, in particular with @gicmo , i would like to take back my suggestion of a generic etc stage. let's make domain specific ones for the time being still.

PR #1129, opened 2020-12-04
distro/rhel84: update kernel options

jkozol:
The kernel options are updated to remove the read only option "ro" from all image types that had it set. Also, the qcow2's kernel options are updated to only set console=ttyS0 once. It was declared twice which is redundant so now it is set for both tty0 and ttyS0.

### DIFF
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -814,7 +814,7 @@ func New() distro.Distro {
 			"timedatex",
 		},
 		defaultTarget: "multi-user.target",
-		kernelOptions: "ro console=ttyS0,115200n8 console=tty0 net.ifnames=0 rd.blacklist=nouveau nvme_core.io_timeout=4294967295 crashkernel=auto",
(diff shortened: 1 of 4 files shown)


ondrejbudai (commented):
The `qcow2` image changes look alright - the kernel options are now the same as in the kickstart. I cannot decide though if removing `ro` from the other image types is the right thing to do.

jkozol:
> The qcow2 image changes look alright - the kernel options are now the same as in the kickstart. I cannot decide though if removing ro from the other image types is the right thing to do.

I also feel unsure about how to decide what to include in the other image types. In general I've been trying to apply the changes to the qcow2 to the other image types. But, I am happy to only drop the `ro` option from the qcow2 and amazon image since that is what is set in the kickstarters.

ondrejbudai:
We have `ro` for other image types than `qcow2` also in `rhel8` definitions so I think I would prefer doing the change only for `qcow2`. We know for sure that it shouldn't have `ro`. We don't know it for the other types so it's probably safer to leave them as they are.

ondrejbudai (approved):
Thanks a lot for this! :) The Openstack tests failed but this is expected as PSI is still not fully operational.

PR #1132, opened 2020-12-08
spec: build & install osbuild-composer(7) man-page

thozza:
Build the `osbuild-composer(7)` man-page as part of the RPM build and include it in the osbuild-composer RPM. Previously the man-page was not shipped at all in any of the produced RPMs. I found out about the problem, because `README.md` refers to the man-page.

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -31,6 +31,8 @@ Source0:        %{gosource}
 BuildRequires:  %{?go_compiler:compiler(go-compiler)}%{!?go_compiler:golang}
 BuildRequires:  systemd
 BuildRequires:  krb5-devel
+BuildRequires:  python3-docutils
+BuildRequires:  make
 %if 0%{?fedora}
 BuildRequires:  systemd-rpm-macros
 BuildRequires:  git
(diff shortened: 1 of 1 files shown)


osbuild-composer.spec
```diff
@@ -117,6 +118,7 @@ export GOFLAGS=-mod=vendor
 %gobuild -o _bin/osbuild-composer %{goipath}/cmd/osbuild-composer
 %gobuild -o _bin/osbuild-worker %{goipath}/cmd/osbuild-worker
 
+make man
```
msehnout, line 121: Since this is the first time we use make in our spec file, I'd also like to have an explicit "make" buildrequires for it. It is currently included in all buildroots by default, but it will bite us in few months:
https://fedoraproject.org/wiki/Releases/34/ChangeSet#Remove_make_from_BuildRoot

msehnout (changes requested):
Thanks for catching this! What an over oversight on our side ... I'd prefer to have `make` mentioned explicitly in the spec file, otherwise the PR looks good to me.

(CI failures are "regular" OpenStack issues that are not related to this PR)

osbuild-composer.spec
thozza: I added make explicitly as BuildRequires. Thanks for pointing out the Fedora change.

dvdhrm (approved):

msehnout (approved):
LGTM, CI failures are unrelated (OpenStack issues)

Issue #1135, opened 2020-12-15
Tests that use podman fail on F32


ondrejbudai:
From `/usr/libexec/tests/osbuild-composer/koji.sh`:

```
Error: default OCI runtime "crun" not found: invalid argument
```

It happened for example here: https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1131/5/pipeline/179

atodorov:
@ondrejbudai few notes here:

1) I've found https://github.com/containers/podman/issues/8227 and it looks like the problem is an older version of podman. That issue says 2.1.1 fixes the problem

2) From what I can see we're using v3.0.1 and what's more containerbuild (which uses podman) is executed on Fedora 33 now.

3) From what I can see we haven't had this issue for a while

I think we can close this issue as not being relevant, what do you think ?

ondrejbudai:
Yeah, this should be fixed now.

PR #1136, opened 2020-12-15
composer: seed the random number generator

ondrejbudai:
I thought rand in Go is auto-seeded but I was wrong, see \[1\].
This commit adds seed initialization.

\[1\]: https://golang.org/pkg/math/rand/#Seed

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -3,9 +3,11 @@
 package cloudapi
 
 import (
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"math/rand"
+	"math"
+	"math/big"
 	"net/http"
 
 	"github.com/go-chi/chi"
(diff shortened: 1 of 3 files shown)


bcl:
Where is math/rand being used? I see crypto/rand being used in several places but not math/rand -- they are different :) crypto/rand gets its numbers from getrandom or /dev/urandom and is much better than math/rand which, even with it seeded, has a higher chance of collisions between systems (unlikely with the uses of Nano but still possible if a number of systems all start at the same time).

ondrejbudai:
I added `math/rand` here: https://github.com/osbuild/osbuild-composer/pull/1131 I guess I can seed the `Manifest()` function with `crypto/rand`...

ondrejbudai:
I force-pushed a version with `crypto/rand`. The global `rand.Seed()` call that I had before was ugly, I think this is nicer.

teg (approved):
Nice! I agree with Brian, nice to use proper random seed, even if it would not really make a difference for this usecase.

bcl:
Ha, I was looking at the wrong branch when searching for math/rand use :) Looks good to me.

ondrejbudai:
https://github.com/osbuild/koji-osbuild/pull/54 and https://github.com/osbuild/osbuild-composer/issues/1135 appeared, otherwise we're green. :green_heart:

PR #1140, opened 2020-12-17
containers: add ubi8 container

teg:
Build ubi8 containers in CI. This requires some preliminary changes, see individual commits for details.

### DIFF
--- a/containers/osbuild-composer/Dockerfile
+++ b/containers/osbuild-composer/Dockerfile
@@ -43,7 +43,7 @@ RUN \
                         "--repofrompath=ephemeral0,${OSB_RPMREPO}" \
                         "--setopt=ephemeral0.gpgcheck=0" \
                         "--setopt=ephemeral0.priority=10" \
-                        install "osbuild-composer" \
+                        install "osbuild-composer-core" \
                 && dnf clean all
 
 # Copy all our local sources, so we can access them from within the container
--- a/schutzbot/containerbuild.sh
+++ b/schutzbot/containerbuild.sh
@@ -13,18 +13,22 @@ COMMIT=$(git rev-parse HEAD)
 IMG_TAGS=(
         "quay.io/osbuild/osbuild-composer:f32-${COMMIT}"
         "quay.io/osbuild/osbuild-composer:f33-${COMMIT}"
+        "quay.io/osbuild/osbuild-composer:ubi8-${COMMIT}"
 )
 IMG_PATHS=(
+        "./containers/osbuild-composer/"
         "./containers/osbuild-composer/"
         "./containers/osbuild-composer/"
 )
 IMG_FROMS=(
         "docker.io/library/fedora:32"
         "docker.io/library/fedora:33"
+        "registry.access.redhat.com/ubi8"
 )
 IMG_RPMREPOS=(
         "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/fedora-32/${ARCH}/${COMMIT}"
         "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/fedora-33/${ARCH}/${COMMIT}"
+        "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/rhel-8.3/${ARCH}/${COMMIT}"
 )
 IMG_COUNT=${#IMG_TAGS[*]}
 
--- a/containers/osbuild-composer/entrypoint.py
+++ b/containers/osbuild-composer/entrypoint.py
@@ -32,20 +32,6 @@ class Cli(contextlib.AbstractContextManager):
             prog="container/osbuild-composer",
         )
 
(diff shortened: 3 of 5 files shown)


ondrejbudai (dismissed):
Awesome! :)

containers/osbuild-composer/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN \
                         "--repofrompath=ephemeral0,${OSB_RPMREPO}" \
                         "--setopt=ephemeral0.gpgcheck=0" \
                         "--setopt=ephemeral0.priority=10" \
-                        install "osbuild-composer" \
+                        install "osbuild-composer-core" \
```
dvdhrm, line 46: The container-entrypoint supports `--builtin-worker`. This will not work with the worker not deployed in the container.

schutzbot/containerbuild.sh
```diff
@@ -13,18 +13,22 @@ COMMIT=$(git rev-parse HEAD)
...
 )
 IMG_FROMS=(
         "docker.io/library/fedora:32"
         "docker.io/library/fedora:33"
+        "registry.access.redhat.com/ubi8"
 )
 IMG_RPMREPOS=(
         "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/fedora-32/${ARCH}/${COMMIT}"
         "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/fedora-33/${ARCH}/${COMMIT}"
+        "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild-composer/rhel-8.3/${ARCH}/${COMMIT}"
```
dvdhrm, line 31: You use the `ubi8` container and tie it to `rhel-8.3`. Once `ubi8` updates to rhel-8.4, this will likely break.

dvdhrm (changes requested):

containers/osbuild-composer/Dockerfile
teg: Fair point, let's drop the support for the built-in worker?

schutzbot/containerbuild.sh
teg: Hmm, I don't see an alternative as there are no minor releases of the UBI. That said, minor releases should be backwards compatible, and we don't depend on much (anything?) from the container that could break, I think, so I don't see how this would likely break. I posted an RFC PR last night that would make all this moot though, thoughts on that?

containers/osbuild-composer/Dockerfile
dvdhrm: This really depends on what this container is used for. If it shall just provide a running composer instance, then yeah, it really should go. But if it should also support building images locally via weldr, then it should stay.

There is no real description of the purpose of this container, yet. So this really all depends on what this is meant to do.

schutzbot/containerbuild.sh
dvdhrm: Right. I thought I remembered ubi-images to having minor-release tags as well. But the information on any of this is quite sparse. If any `8.x` RPMs are meant to work on `ubi8` containers, then this should be fine.

containers/osbuild-composer/Dockerfile
larskarlitski: > This really depends on what this container is used for. If it shall just provide a running composer instance, then yeah, it really should go. But if it should also support building images locally via weldr, then it should stay.

I don't think running workers together with composer in the same container is worthwhile. The whole point of having a local worker (i.e., the reason why `osbuild-composer` depends on `osbuild-worker`) is to provide backwards compatibility for people using Image Builder on a local RHEL machine via the weldr API. While that's a fine use case, it's covered with the usual way of installing on RHEL: via the `osbuild-composer` package. It does not scale well to the service we're imagining, where we really want fleets of workers on different machines. I'd say there's no use-case to support it and would only make the container more complex than it needs to be.

> There is no real description of the purpose of this container, yet. So this really all depends on what this is meant to do.

Yeah, good point. Probably makes sense to add a README to `containers/osbuild-composer` stating what it is used for, or adding it to the header comment in `Dockerfile`.

larskarlitski:
I like this a lot. Thanks!

containers/osbuild-composer/Dockerfile
teg: I dropped `--builtin-worker` from `entrypoint.py`.

dvdhrm (approved):

containers/osbuild-composer/Dockerfile
dvdhrm: Some unused code is now left, which implements `builtin_worker`, but I guess it does no harm.

teg: I started cleaning it up, but realised I was more likely to break something, so figured I'd leave that for a future clean-up ;)

teg:
openstack failure is unrelated.

PR #1141, opened 2020-12-18
containers: prepare for deployment

teg:
This changes the container logic a bit to bring it more in line with image-builder and closer to what AppSRE expects to deploy. The main changes is that we no longer build from RPM and that we only build UBI containers.


dvdhrm:
>  - Build directly from git. Now that the RPM we include in our
>   container does not have any dependencies and only contains a
>   couple of executables, the indirection via RPM has less value.
>   Eventually the value will be reduced even further as we merge
>   the entrypoint into the main binary and move dnf-json into the
>   worker, leaving us with only a go binary. The only potential
>   benefit might be that the build environment of RPMs is more
>   clearly defined, but there is no real reason to believe that
>   our mockbuild is any better than using the UBI golang build
>   container.

Given that with this our own deployments would not be based on rpms, would that also deprecate our reasoning that tests should strictly run in rpm-based deployments?

Furthermore, `osbuild-composer` will still have dependencies, won't it? Otherwise, there would be no reason to put it into a UBI container. You could just place your go binary in the container and that would be it.

Regarding the container itself: What purpose does it serve? Is it meant solely for AppSRE purposes? Is it meant as general-purpose containerized composer? What is the target environment? What is it supposed to run? How is it meant to be configured? What services exactly is it meant to provide?

Lastly, the `build_deploy.sh` looks quite misplaced in the root-directory of an upstream project repository. Can't we put this together with the other configuration and code of the actual deployment?

teg:
> > * Build directly from git. Now that the RPM we include in our
> >   container does not have any dependencies and only contains a
> >   couple of executables, the indirection via RPM has less value.
> >   Eventually the value will be reduced even further as we merge
> >   the entrypoint into the main binary and move dnf-json into the
> >   worker, leaving us with only a go binary. The only potential
> >   benefit might be that the build environment of RPMs is more
> >   clearly defined, but there is no real reason to believe that
> >   our mockbuild is any better than using the UBI golang build
> >   container.
> 
> Given that with this our own deployments would not be based on rpms, would that also deprecate our reasoning that tests should strictly run in rpm-based deployments?

Might make sense to discuss. I don't think it necessarily follows. The rpm-based tests can still be run externally against a container.

> Furthermore, `osbuild-composer` will still have dependencies, won't it? Otherwise, there would be no reason to put it into a UBI container. You could just place your go binary in the container and that would be it.

We might be able to drop the container altogether, but for now we at least have the couple of python scripts still. I'm also not sure if there are any implicit dependencies on things like a basic filesystem layout, /etc, and stuff like that.

> Regarding the container itself: What purpose does it serve? Is it meant solely for AppSRE purposes? Is it meant as general-purpose containerized composer? What is the target environment? What is it supposed to run? How is it meant to be configured? What services exactly is it meant to provide?

For now this is only targeting what we need for running as part of insights. Happy to consider generalizing it once we have that working, of course. For this we need to provide the composer API and the worker API. I'm still investigating exactly how this will be invoked, but I figured a good starting-point would be to test this just with podman, and then eventually hook it up with Insight's CI.

> Lastly, the `build_deploy.sh` looks quite misplaced in the root-directory of an upstream project repository. Can't we put this together with the other configuration and code of the actual deployment?

Yeah, we may want to reconsider this, but for now I wanted to do what works for image builder so we have a starting point to iterate from. I'm not at all married to this being the permanent solution.

Gundersanne:
> > > * Build directly from git. Now that the RPM we include in our
> > >   container does not have any dependencies and only contains a
> > >   couple of executables, the indirection via RPM has less value.
> > >   Eventually the value will be reduced even further as we merge
> > >   the entrypoint into the main binary and move dnf-json into the
> > >   worker, leaving us with only a go binary. The only potential
> > >   benefit might be that the build environment of RPMs is more
> > >   clearly defined, but there is no real reason to believe that
> > >   our mockbuild is any better than using the UBI golang build
> > >   container.
> > 
> > 
> > Given that with this our own deployments would not be based on rpms, would that also deprecate our reasoning that tests should strictly run in rpm-based deployments?
> 
> Might make sense to discuss. I don't think it necessarily follows. The rpm-based tests can still be run externally against a container.
 
Personally I like the indirection of rpms. Just having the single way to build/install the project for distributing makes sense imo, and having the container essentially as a wrapper around those rpms. We don't do it in image-builder, but it's something i'd like to do.

> > Furthermore, `osbuild-composer` will still have dependencies, won't it? Otherwise, there would be no reason to put it into a UBI container. You could just place your go binary in the container and that would be it.
> 
> We might be able to drop the container altogether, but for now we at least have the couple of python scripts still. I'm also not sure if there are any implicit dependencies on things like a basic filesystem layout, /etc, and stuff like that.

Within app-sre containers have to be based on the ubi8 one, there's a ubi8-minimal container however.

> > Regarding the container itself: What purpose does it serve? Is it meant solely for AppSRE purposes? Is it meant as general-purpose containerized composer? What is the target environment? What is it supposed to run? How is it meant to be configured? What services exactly is it meant to provide?
> 
> For now this is only targeting what we need for running as part of insights. Happy to consider generalizing it once we have that working, of course. For this we need to provide the composer API and the worker API. I'm still investigating exactly how this will be invoked, but I figured a good starting-point would be to test this just with podman, and then eventually hook it up with Insight's CI.

If we want a general container we should start releasing it when we do releases. And document the registry. But for those I wouldn't upload every main build.

> > Lastly, the `build_deploy.sh` looks quite misplaced in the root-directory of an upstream project repository. Can't we put this together with the other configuration and code of the actual deployment?
> 
> Yeah, we may want to reconsider this, but for now I wanted to do what works for image builder so we have a starting point to iterate from. I'm not at all married to this being the permanent solution.

`build_deploy.sh` can be put anywhere, in app-intf just set `build_deploy_script_path` under the config>project>jobs in the build.yml, relative to the project root. So we could put it in tools quite easily.

Gundersanne:
So after getting access to e2e-deploy, it's not really involved with our stage deployment. So we'll need this PR to build images and push them to quay. And we'll need a build_deploy script (the latter can also be in tools).

schutzbot/containerbuild.sh
```diff
@@ -2,64 +2,20 @@
...
-                build \
-                "--build-arg=OSB_FROM=${i_from}" \
-                "--build-arg=OSB_RPMREPO=${i_rpmrepo}" \
-                "--tag=${i_tag}" \
-                "${i_path}"
-        echo
-done
+podman \
+	build \
+	"--file=distribution/Dockerfile-ubi"
```
Gundersanne, line 19: there's a \ missing here

distribution/Dockerfile-ubi
```diff
@@ -0,0 +1,18 @@
...
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+RUN microdnf install python3
+RUN mkdir -p "/usr/libexec/osbuild-composer"
+RUN mkdir -p "/etc/osbuild-composer/"
+RUN mkdir -p "/run/osbuild-composer/"
+RUN mkdir -p "/var/cache/osbuild-composer/"
+RUN mkdir -p "/var/lib/osbuild-composer/"
+COPY --from=builder /opt/app-root/src/go/bin/osbuild-composer /usr/libexec/osbuild-composer/
+COPY ./containers/osbuild-composer/entrypoint.py /opt/entrypoint.py
+COPY ./dnf-json /usr/libexec/osbuild-composer/
```
Gundersanne, line 15: Won't we need the repositories in `/usr/share` as well here?

Gundersanne:
For the build_deploy script do something similar here: https://github.com/osbuild/image-builder/blob/main/build_deploy.sh, but then rename it to tools/clouddot_build_deploy.sh or something

dvdhrm:
Can you rebase this on `main`? I thought the ubi8+rpm changes were merged already.

tools/clouddot_build_deploy.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# AppSRE runs this script to build the container and push it to Quay.
+set -exv
+
+IMAGE_NAME="quay.io/cloudservices/image-builder"
```
Gundersanne, line 5: This needs to be `quay.io/cloudservices/osbuild-composer`

Gundersanne (changes requested):

teg:
Thanks for the fix @Gundersanne! Feel free to approve/merge when green.

Gundersanne (approved):
lgtm

PR #1142, opened 2020-12-21
Introduce support for CentOS Stream

### LINKED ISSUE #431: Add CentOS support
We should definitely support CentOS at some point. Here are some blockers we discussed with @teg:

- [ ] `internal/distro` needs refactoring, it contains a high amount of duplicate code. After a refactoring, it should be fairly easy to add support for a new distro.
- [ ] we want to have Fedora fully covered by proper tests, let's wait for them until we start working on a new distro.

ondrejbudai:
This is my attempt at introducing CentOS Stream support into osbuild-composer. I have 3 discussion topics:

1) I'm reusing `rhel84` to define CentOS 8. I think that this makes sense because whatever change we make in RHEL, it should go into CentOS. Defining RHEL 8.4 and CentOS 8 in two files would most certainly lead to these two distributions slowly drifting apart (remember the times when we had two spec files, they drifted apart quite a bit). The only differences in CentOS are the lack of `insights-client`, a different EFI vendor, and a different osbuild runner (technically we can use the rhel one but it just feels weird?).

2) How to name CentOS Stream 8? It turns out that both CentOS 8 and Stream 8 have the same `${ID}_${VERSION_ID}`: `centos-8`. Therefore I decided to name the new distribution `centos-8`. However, this doesn't mean that we have CentOS 8 support, this PR adds only support for Stream 8 (because the repofiles for centos are missing). Not sure what should be the convention for this... I kinda think that we should just live in a bit messy world for a year with the name `centos-8` and then, it will not matter because CentOS 8 will be EOL.

3) What about "CentOS Stream for Edge"? It obviously doesn't exist... Should we support this image type for CentOS? How should it be named?

btw, if we want to merge this including the test, we need to make an RPMRepo snapshot first.

Closes #431


larskarlitski:
This is so cool :tada: 

I haven't had time to look at the code yet, but might be able to shed some first light on your discussion points:

> 1. I'm reusing `rhel84` to define CentOS 8. I think that this makes sense because whatever change we make in RHEL, it should go into CentOS. Defining RHEL 8.4 and CentOS 8 in two files would most certainly lead to these two distributions slowly drifting apart (remember the times when we had two spec files, they drifted apart quite a bit). The only differences in CentOS are the lack of `insights-client`, a different EFI vendor, and a different osbuild runner (technically we can use the rhel one but it just feels weird?).

Your reasoning sounds good to me. We can always split it later, should the differences become to much. I don't suspect so, because as you say, the whole idea of CentOS Stream is to be very close to RHEL. But yeah, let's have a separate osbuild runner for it.

> 2. How to name CentOS Stream 8? It turns out that both CentOS 8 and Stream 8 have the same `${ID}_${VERSION_ID}`: `centos-8`. Therefore I decided to name the new distribution `centos-8`. However, this doesn't mean that we have CentOS 8 support, this PR adds only support for Stream 8 (because the repofiles for centos are missing). Not sure what should be the convention for this... I kinda think that we should just live in a bit messy world for a year with the name `centos-8` and then, it will not matter because CentOS 8 will be EOL.

Yeah. Let's only support stream and call it `centos-8`.

> 3. What about "CentOS Stream for Edge"? It obviously doesn't exist... Should we support this image type for CentOS? How should it be named?

I love that this is so easy to add support for now :smile: It might be a thing, but we probably want to discuss with the Edge/IoT people. I suggest not blocking on it and leaving it out for now.

teg:
> This is so cool 🎉

Very cool!

> > 1. I'm reusing `rhel84` to define CentOS 8. I think that this makes sense because whatever change we make in RHEL, it should go into CentOS. Defining RHEL 8.4 and CentOS 8 in two files would most certainly lead to these two distributions slowly drifting apart (remember the times when we had two spec files, they drifted apart quite a bit). The only differences in CentOS are the lack of `insights-client`, a different EFI vendor, and a different osbuild runner (technically we can use the rhel one but it just feels weird?).
> 
> Your reasoning sounds good to me. We can always split it later, should the differences become to much. I don't suspect so, because as you say, the whole idea of CentOS Stream is to be very close to RHEL. But yeah, let's have a separate osbuild runner for it.

As CentOS Stream turns into RHEL, shouldn't the idea be that we at regular intervals copy and "freeze" the CentOS Stream definition to become the new RHEL minor definition?

> > 1. How to name CentOS Stream 8? It turns out that both CentOS 8 and Stream 8 have the same `${ID}_${VERSION_ID}`: `centos-8`. Therefore I decided to name the new distribution `centos-8`. However, this doesn't mean that we have CentOS 8 support, this PR adds only support for Stream 8 (because the repofiles for centos are missing). Not sure what should be the convention for this... I kinda think that we should just live in a bit messy world for a year with the name `centos-8` and then, it will not matter because CentOS 8 will be EOL.
> 
> Yeah. Let's only support stream and call it `centos-8`.

I see pitchforks in our future 🔮

teg:
Thanks for reviving this @ondrejbudai! As soon as schutzi is happy, I think this can go in. Or is there anything else we are missing?

ondrejbudai:
2 things:

"CentOS for Edge". The ostree test is currently disabled but the image type is still available under the name "rhel-edge-commit". If we want to keep the image type even for CentOS, we need a name for it and we need to make few adjustments (for example change the ref from `rhel`).

Do we want to add support for the regular CentOS? We would just need to add repositories for it (currently, I just assume that every centos == centos stream). I think that testing it is impossible.

teg:
I would skip CentOS for Edge for now, as it would otherwise be something we would invent ourselves without an owner for it. We should sync up with @nullr0ute on the future of that at some point, but I would say that's not urgent at all.
I don't think it makes sense to try to add regular CentOS support at this point. At least it should be a completely separate effort to this.

ondrejbudai:
OK, this is now ready to go in!

## My decisions

1) The Weldr API will build classic CentOS on classic CentOS and Stream on Stream. My previous version would just build Stream everywhere which I found confusing. Therefore, I decided to distinguish between those distros, and from there, it's trivial to put both classic CentOS and Stream repos in the code.
   Just a note I like to put everywhere: Due to the weird lifecycle of the classic CentOS, testing this is weird, so we don't do it. I would like to apologize to all future readers wondering why composer is broken on classic CentOS. Sorry, I didn't mean to break it!

2) I disabled `s390x` for CentOS as `s390x` isn't supported by it. Also, I disabled the rhel-edge-commit image type because we need to define it and name it first.

## Future work

I would like to enable `aarch64` tests for CentOS Stream. This way, we can sort of test the upcoming RHEL 8.y version on at least one "alternative" architecture. (Testing unreleased RHEL 8 on aarch64 is impossible because we need to access the internal repos but we don't have any internal aarch64 machines).

BTW this way, we would use CentOS Stream to catch bugs in the upcoming RHEL version which is one of the points of CentOS Stream. Awesome, isn't it?

teg:
Love where this is heading, we definitely want CentOS Stream support.
I wonder about the life-cycles though:
 - Isn't the idea that CentOS Stream is a rolling release that turns into RHEL? Your suggestion of tracking the most recent RHEL minor release seems to have it the other way around, or did I read you wrong? Also, if it is indeed rolling, we should add it to RPMRepo and test against that.
 - Not sure about the attempt at regular CentOS support. If we can't test it, then I'd leave that to be a downstream problem, this way we may need to deal with it after the fact which just sounds hairy. Especially as it will be EOL soon anyway. This is not a hill I'm willing to die on though :)

Dropping s390x + edge support makes perfect sense. Very much support aarch64 testing support eventually too!

ondrejbudai:
Let's divide our support for CentOS Stream into two parts: repositories and distro definition (aka `distro.go`).

**Repositories**
In the RPM, we're shipping the official CS repositories. They are the same for the whole 8 lifecycle and indeed rolling (or continuously delivered as some people prefer).
For testing, I'm overriding them with a specific Stream compose. These composes seem to live on the mirror for at least a half year so I'm fine with them for now. Of course, we might want to switch to the RPMRepo but that can be handled in the future.


**Distro definition**
Now, we are doing this: `rhel84.New(), rhel84.NewCentos()`. The first call provides the `rhel-84` distro definition, the second one the  `centos-8` one. My thought is that the `centos-x` definition should be always provided by the newest `rhel-xy` definition we have. Example: When we decide to branch `rhel85` from `rhel84`, we should also move the `centos-8` definition like this: `rhel84.New(), rhel85.New(), rhel85.NewCentos()`.

I do agree that this looks like we're doing things the other way around but I'm not sure how to implement it differently... Honestly, this is just an implementation detail in composer in order to not duplicate the code between RHEL and CentOS Stream.

## The classic CentOS issue

I thought of it some more and I think we're mostly fine. CentOS should a bug for bug compatible RHEL rebuild, right? Therefore, if we cannot build CentOS using centos repos + rhel `distro.go`, there's somewhere bug in CentOS, which means it's breaking the "bug for bug compatible" promise and it's not our problem. The only parts in which we can introduce a bug are the ones that are different between CentOS and RHEL in our code:

- repositories
- disabling of certain arches and image types
- additional differences in `distro.go` (currently there are only two: removal of `insights-client` and different uefi vendor)

Disabling of certain arches and image types - that's covered by unit tests. Additional differences are mostly covered by CentOS Stream. The only thing which is questionable is the repositories. Let's be honest though: We're using repo overrides for all distros so we can break the repositories in all distros and not catch it.

Note that I'm thinking mostly in terms of the Weldr API. Building classic CentOS using the Cloud API is... weird... Because CentOS 8 is in fact also rolling (there are no minor versions) but it tracks RHEL which is not considered a rolling distro in our distro definitions in osbuild-composer.

teg:
Ah, I missed that you are able to pin tests to a compose. I'm good with that then.

I agree that the way the distro definition is shared is an implementation detail, and that this makes the most sense for now. We may revesit this in the future.

Classic CentOS support: Sounds reasonable to me.

teg (approved):
Go CentOS Stream! 🎉

stefwalter:
Great stuff @ondrejbudai \o/

PR #1146, opened 2021-01-05
Docker-compose and Dockerfile for worker container

achilleas-k:
Adding *osbuild-composer-worker* Dockerfile (`Dockerfile-worker`) and a `docker-compose.yml` file that makes it straightforward to set up a multi-container environment.  This is a lot faster than building RPMs and installing them in VMs for when you need to do quick compile-debug cycles.

## SSL cert generation script

The biggest change to existing tools is the SSL certificate generation script: I've moved the cert generation parts from `tools/provisioning.sh` into a new file (`tools/gen-certs.sh`).  The original workflow that uses this script should be unaffected, but having the cert generation separate makes it easy to quickly create certificates and keys in any directory using the `openssl.cnf` file from any location.

## HACKING.md

I added a section to HACKING.md that explains how to get the Docker setup up and running.  The docker-compose.yml file points to the Dockerfiles directly, so running `docker-compose up` the first time will build the images as well.

## File location

I added the new docker files to the `distribution` directory because that's where the `Dockerfile-ubi` file is kept.  I was thinking it might be more appropriate to put these in the `containers` directory instead.  Any thoughts on this?

You might also notice an `.env` file that defines the project name.  For those unfamiliar with docker-compose, this affects the names of the containers, networks, and volumes that are created.  In our case, naming the project `osbuild` makes the containers have the names `osbuild_composer_1` and `osbuild_worker_1` and the network gets named `osbuild_net`.

## Other

Added `python-dnf` to the *osbuild-composer* Dockerfile (`Dockerfile-ubi`), since it's required by the `dnf-json` script.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -52,19 +52,36 @@ containers. Building and starting containers is generally faster than building
 RPMs and installing them in a VM, so this method is more convenient for
 developing and testing changes quickly.
 
+### Configuration
+
 Each service (*composer* and *worker*) requires a configuration file and a set
-of certificates. Use the [`tools/gen-certs.sh`](./tools/gen-certs.sh) script to
+of certificates. The storage location for these is shared between the
+containers for simplicity. By default it's configured to be at
+`./containers/config`, but this can be changed in the
+[`./distribution/.env`](./distribution/.env) file by modifying the value of the
+`$CONTAIN£R_CONFIG_DIR` variable (both absolute and relative paths work).
+
+Use the [`tools/gen-certs.sh`](./tools/gen-certs.sh) script to
 generate the certificates (using the test OpenSSL config file):
 
     ./tools/gen-certs.sh ./test/data/x509/openssl.cnf ./containers/config  ./containers/config/ca
 
-The services also require a config file (each) which they expect to be in the
+Note that the two arguments `./containers/config` and `./containers/config/ca`
+should be the same location as the `$CONTAINER_CONFIG_DIR` described above
+
+The services also require a config file each which they expect to be in the
 same directory. The following test files can be copied into it:
 
     cp ./test/data/composer/osbuild-composer.toml ./test/data/composer/osbuild-worker.toml ./containers/config/
 
-The `containers/config` directory will be mounted inside both containers (see
+The `$CONTAINER_CONFIG_DIR` (default `containers/config`) directory will be mounted inside both containers (see
 the [`docker-composer.yml`](./distribution/docker-compose.yml) file).
+
+### Build and run
+
+To build the containers, change into the `distribution/` directory and run:
+
+    docker-compose build
 
 To start the containers, change into the `distribution/` directory and run:
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -50,7 +50,18 @@ unit tests before going through the full workflow described above.
 *osbuild-composer* and *osbuild-composer-worker* can be run using Docker
 containers. Building and starting containers is generally faster than building
 RPMs and installing them in a VM, so this method is more convenient for
-developing and testing changes quickly.
+developing and testing changes quickly. However, using this method has several
+downsides:
+- It doesn't build the RPMs so the `.spec` file isn't tested.
+- The environment is quite different from production (e.g., installation paths,
+  privileges and permissions).
+- The setup is not complete for all required services, so some functionality
+  isn't available for testing this way (e.g., Koji Hub and all its dependent
+  services).
+
+The containers are a good way to quickly test small changes, but before
+submitting a Pull Request, it's recommended to run through all the tests using
+the [Virtual Machine](#virtual-machine) setup described above.
 
 ### Configuration
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,5 +1,7 @@
 # Hacking on osbuild-composer
 
+## Virtual Machine
+
 *osbuild-composer* cannot be run from the source tree, but has to be installed
 onto a system. We recommend doing this by building rpms, with:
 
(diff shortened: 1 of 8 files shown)


achilleas-k:
Added the gen-certs.sh to the spec file for the test package.  I think this is ready now.

HACKING.md
```diff
@@ -42,3 +44,37 @@ For a quick compile and debug cycle, we recommend iterating code using thorough
...
+*osbuild-composer* and *osbuild-composer-worker* can be run using Docker
+containers. Building and starting containers is generally faster than building
+RPMs and installing them in a VM, so this method is more convenient for
+developing and testing changes quickly.
+
+Each service (*composer* and *worker*) requires a configuration file and a set
+of certificates. Use the [`tools/gen-certs.sh`](./tools/gen-certs.sh) script to
+generate the certificates (using the test OpenSSL config file):
+
+    ./tools/gen-certs.sh ./test/data/x509/openssl.cnf ./containers/config  ./containers/config/ca
```
teg, line 59: In a follow-up this might be nice to add as a make target or something like that, just to make it even easier to get going. Or maybe I'm overthinking that :)

HACKING.md
```diff
@@ -42,3 +44,37 @@ For a quick compile and debug cycle, we recommend iterating code using thorough
 unit tests before going through the full workflow described above.
 
 [1]: https://wiki.qemu.org/Documentation/Networking#User_Networking_.28SLIRP.29
+
+## Containers
+
+*osbuild-composer* and *osbuild-composer-worker* can be run using Docker
+containers. Building and starting containers is generally faster than building
+RPMs and installing them in a VM, so this method is more convenient for
+developing and testing changes quickly.
```
teg, line 53: Maybe also mention the possible downside (not testing the .spec file / integration into RHEL/Fedora), otherwise one might reasonably ask why we test in VMs at all.

teg (approved):
This is really great! Can't wait to use it. Something for @msehnout to keep in mind for the developer guide too. I had some minor nits, but nothing that was actually wrong, so feel free to merge as-is (after checking CI) and address / discuss these in follow-ups.

There is a comment in the PR description about it being DRAFT, which I think can be removed now?

distribution/Dockerfile-worker
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:latest AS builder
+COPY . .
+ENV GOFLAGS=-mod=vendor
+RUN go install ./cmd/osbuild-worker
+
+FROM fedora
+RUN dnf install -y python3 python3-pip git bubblewrap systemd policycoreutils osbuild
```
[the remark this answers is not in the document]
achilleas-k: Right, should've have gone to the spec file from the start.
`systemd qemu-img osbuild osbuild-ostree` should be enough.  Will test and update.

HACKING.md
```diff
@@ -42,3 +44,37 @@ For a quick compile and debug cycle, we recommend iterating code using thorough
...
+Each service (*composer* and *worker*) requires a configuration file and a set
+of certificates. Use the [`tools/gen-certs.sh`](./tools/gen-certs.sh) script to
+generate the certificates (using the test OpenSSL config file):
+
+    ./tools/gen-certs.sh ./test/data/x509/openssl.cnf ./containers/config  ./containers/config/ca
+
+The services also require a config file (each) which they expect to be in the
+same directory. The following test files can be copied into it:
+
+    cp ./test/data/composer/osbuild-composer.toml ./test/data/composer/osbuild-worker.toml ./containers/config/
```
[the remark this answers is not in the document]
achilleas-k: They could but then we'd be generating certs inside a directory checked into git and that feels a bit messy.  The way the guide is written now, the `config` directory can be placed anywhere and the `docker-composer.yml` volumes updated for anyone who wants to work outside the source tree.

Now that I think about it, I'd prefer to make that path configurable (through `distribution/.env`) and add an instruction at the start that it can be changed to be anywhere (and keep it in `containers/config` by default).

HACKING.md
achilleas-k: Good idea.  I'll add some caveats.

tools/gen-certs.sh
```diff
@@ -0,0 +1,97 @@
...
+    # Copy the private key to the location expected by the tests
+    cp ca.cert.pem "$CERTDIR"/ca-crt.pem
+
+    # Generate a composer certificate.
+    openssl req -config "$OPENSSL_CONFIG" \
+        -keyout "$CERTDIR"/composer-key.pem \
+        -new -nodes \
+        -out /tmp/composer-csr.pem \
+        -subj "/CN=localhost/emailAddress=osbuild@example.com" \
+        -addext "subjectAltName=DNS:localhost"
```
achilleas-k, line 48: Found another issue: For this to work, the internal hostname of the container needs to be added to the alt names of the certs.

distribution/Dockerfile-worker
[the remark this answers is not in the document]
teg: systemd i think is redundant too, might be wrong though...

HACKING.md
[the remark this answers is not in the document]
teg: ah, got it, makes sense

distribution/Dockerfile-worker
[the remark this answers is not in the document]
achilleas-k: You're right.  It's listed in the spec file, but it's a dependency of osbuild so it's redundant.

HACKING.md
achilleas-k: Oh yeah, that would be really useful for some other things that I have in mind as well.  I'll come back to this later for sure.

HACKING.md
achilleas-k: Let me know what you think of the new text.

teg (approved):
Love it!

distribution/Dockerfile-worker
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:latest AS builder
+COPY . .
+ENV GOFLAGS=-mod=vendor
+RUN go install ./cmd/osbuild-worker
+
+FROM fedora
```
msehnout, line 6: Why fedora and not ubi8-minimal?

distribution/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '2.4'
+services:
+  composer:
+    image: local/osbuild-composer
+    build:
+      context: ..
+      dockerfile: ./distribution/Dockerfile-ubi
```
msehnout, line 7: Would it be possible to change the name from Dockerfile-ubi to something like Dockerfile-composer without breaking anything? It would make the naming more consistent.

msehnout (commented):
Looks very good!

distribution/Dockerfile-worker
msehnout: Answer from chat: ubi8-minimal is missing bubblewrap

distribution/docker-compose.yml
achilleas-k: I see it mentioned a couple of times in the repo (`schutzbot/containerbuild.sh` and `tools/clouddot_build_deploy.sh`) and I don't know if it's also used anywhere outside.  I'd be in favour of renaming for consistency, but didn't want to make any unrelated changes.

I'd also be in favour of either:
- Naming them plain `Dockerfile` in a separate directory each (i.e., `distribution/composer/Dockerfile` and `distribution/worker/Dockerfile`).
- Naming them `x.Dockerfile` (i.e., `composer.Dockerfile` and `worker.Dockerfile`).

The former is conventional (in my experience) and it can be used to separate each with their scripts or accompanying files, though we don't seem to have any that are specific to each now.
The latter is also conventional when there are multiple Dockerfiles and helps identify the "type" for editors and browsers (and GitHub).

teg: I think we can rename this now that we are actually not deploying from this file. Just naming it `Dockerfile` seems appealing to me, do you lose anything compared to the second suggestion, or will all the tools still identify it correctly?

teg: For a follow-up though!

teg:
The CI failure was an issue with booting in OpenStack with F32. Clearly unrelated to this PR.

PR #1148, opened 2021-01-06
kojiapi,cloudapi: exclude packages when depsolving

jkozol:
When rpmmd's Depsolve function is called we need to pass in the image type's excluded packages. These excluded packages are retrieved when we get the packages we include from each image type.

This should fix the issue we've been seeing with brew builds including excluded packages such as firewalld.

Thanks @ondrejbudai for the help figuring this out!!


ondrejbudai (approved):
💚

PR #1151, opened 2021-01-08
image-types: Updated RHEL-8 KVM and EC2 images information

thozza:
Updated RHEL-8 KVM and EC2 images information.

Updates in KVM image type:
- Updated the intended use of this image type, based on discussion with Virt PM.
- Added note about the most common way to consume official Red Hat content using this image type.

Updates in EC2 image type:
- Added note about the most common way to consume official Red Hat content using this image type.

### DIFF
--- a/image-types/rhel8/kvm-guest-image.md
+++ b/image-types/rhel8/kvm-guest-image.md
@@ -1,7 +1,10 @@
 # KVM Guest Image
 
-This is an image meant to be used as a generic base image for various
-virtualization environments.
+This is an image meant to be used as a generic base image for KVM
+based hypervisors. It is mainly used with [Red Hat Virtualization
+(RHV)][rhv] or [Red Hat OpenStack Platform (RHOSP)][rhosp]. Using this
+type of image with other types of hypervisors or with cloud providers
+is not supported.
 
 To be usable in common environments, it is provided in the *qcow2* format and
 has *cloud-init* installed and enabled.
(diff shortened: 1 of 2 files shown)


image-types/rhel8/kvm-guest-image.md
```diff
@@ -1,7 +1,9 @@
 # KVM Guest Image
 
-This is an image meant to be used as a generic base image for various
-virtualization environments.
+This is an image meant to be used as a generic base image for KVM
+based hypervisors. It is mainly used with RHV or RHOSP. Using this
```
larskarlitski, line 4: I'd prefer spelling `RHV` and `RHOSP` out once, and linking to https://access.redhat.com/products/red-hat-virtualization and https://access.redhat.com/products/red-hat-openstack-platform (if these are the best product links), respectively.

larskarlitski (changes requested):
Nice, thanks for updating those!

Nitpickiest nitpick: we usually use imperative style for commit messages, describing what the commit does, not what the author did. For example "Update something..." vs "Updated something...". No need to change it in this PR, but please change it to include the spelled-out forms of the product abbreviations.

image-types/rhel8/kvm-guest-image.md
msehnout: I also think we should at least once write the full name before we start using abbreviations. There is a lot of TLAs in IT world and way too many conflicting TLAs in RH portfolio. So I'd also mention what RHSM and RHUI stands for because every time I read it, I think about user interface :-).

(RHEL, CDN, and KVM might be fine without)

thozza: Sure, makes sense. I'll write full names and add links to products. Thanks.

thozza: I kept KVM and CDN as abbreviations, but other than these, I included also full names, together with links where appropriate. I also changed commit messages to use imperative.

larskarlitski (approved):
Perfect. Thank you!

teg (approved):

PR #1155, opened 2021-01-11
Koji: Retrieve the Manifests for a compose job

### LINKED ISSUE #996: kojiapi: ablity to get the manifests for a compose job
If a compose fails, It probably would be handy to get the complete manifests for all the image builds and attach that as a log to the (failed) koji task. Could be combined with the full log API or a separate call.

achilleas-k:
Added an API endpoint that returns the manifest for a given osbuild-koji job.

The Koji API function assumes that the second dependency of the incoming job ID (which should match a Finalize Job) is the first (and perhaps only) Build Job. It retrieves the arguments of the Build Job to extract the Manifest.

The tests cover basic (non-failing) cases for now. More notable is that in all test cases the manifest is empty, which isn't very interesting.

Closes #996.

### DIFF
--- a/internal/kojiapi/api/openapi.yml
+++ b/internal/kojiapi/api/openapi.yml
@@ -87,6 +87,36 @@ paths:
             text/plain:
               schema:
                 type: string
+  '/compose/{id}/manifests':
+    get:
+      summary: Get the manifests for a compose.
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+            example: 123e4567-e89b-12d3-a456-426655440000
+          required: true
+          description: ID of compose status to get
+      description: 'Get the manifests of a running or finished compose. Returns one manifest for each image in the request. Each manifest conforms to the format defined at https://www.osbuild.org/man/osbuild-manifest.5'
+      responses:
+        '200':
+          description: The manifest for the given compose.
+          content:
+            application/json:
+        '400':
+          description: Invalid compose id
+          content:
+            text/plain:
+              schema:
+                type: string
+        '404':
+          description: Unknown compose id
+          content:
+            text/plain:
+              schema:
+                type: string
   /compose:
     post:
       summary: Create compose
--- a/internal/kojiapi/server.go
+++ b/internal/kojiapi/server.go
@@ -384,6 +384,37 @@ func (h *apiHandlers) GetComposeIdLogs(ctx echo.Context, idstr string) error {
 	return ctx.JSON(http.StatusOK, response)
 }
 
+// GetComposeIdManifests returns the Manifests for a given Compose (one for each image).
+func (h *apiHandlers) GetComposeIdManifests(ctx echo.Context, idstr string) error {
+	id, err := uuid.Parse(idstr)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter id: %s", err))
+	}
+
+	var finalizeResult worker.KojiFinalizeJobResult
+	_, deps, err := h.server.workers.JobStatus(id, &finalizeResult)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, fmt.Sprintf("Job %s not found: %s", idstr, err))
+	}
+
+	if len(deps) <= 1 {
+		// ID matched a job ID that's not a Finalize job. Treat it as "not found".
+		return echo.NewHTTPError(http.StatusNotFound, fmt.Sprintf("Job %s not found: job does not exist", idstr))
+	}
+
+	manifests := make([]distro.Manifest, len(deps)-1)
+	for i, id := range deps[1:] {
+		var buildJob worker.OSBuildKojiJob
+		if _, err := h.server.workers.JobArgs(id, &buildJob); err != nil {
+			// This is a programming error.
+			panic(err)
+		}
+		manifests[i] = buildJob.Manifest
+	}
+
+	return ctx.JSON(http.StatusOK, manifests)
+}
+
 // A simple echo.Binder(), which only accepts application/json, but is more
 // strict than echo's DefaultBinder. It does not handle binding query
 // parameters either.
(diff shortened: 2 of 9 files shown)


gicmo (commented):
Very cool! Looks very good. One thing, since the manifest is different for each architecture, I think we want the manifest for all the builds; vermuch as with status and logs. :D

achilleas-k:
> Very cool! Looks very good. One thing, since the manifest is different for each architecture, I think we want the manifest for all the builds; vermuch as with status and logs. :D

Thanks.  Should have realised that the manifests are different after being submitted (even though the same request is sent) since different archs have different hashes and maybe even packages after depsolve.

internal/kojiapi/api/openapi.yml
```diff
@@ -160,6 +192,10 @@ components:
         koji_import_logs: {}
         image_logs:
           type: array
+    Manifest:
```
ondrejbudai, line 195: Will this definition still be valid after the DAG changes in osbuild, @gicmo? Maybe, we should just leave `Manifest` as a generic object here and add a note that it's a valid osbuild manifest with a link to https://www.osbuild.org/man/osbuild-manifest.5?

internal/kojiapi/server.go
```diff
@@ -384,6 +384,43 @@ func (h *apiHandlers) GetComposeIdLogs(ctx echo.Context, idstr string) error {
...
+	}
+
+	var finalizeResult worker.KojiFinalizeJobResult
+	_, deps, err := h.server.workers.JobStatus(id, &finalizeResult)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, fmt.Sprintf("Job %s not found: %s", idstr, err))
+	}
+
+	if len(deps) > 1 {
+		// Get the first build job ID (second dependency) to get its manifest
```
ondrejbudai, line 401: Hmm, different architectures have different manifests so we probably need to return all of them.

internal/kojiapi/server.go
```diff
@@ -155,7 +155,7 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 		Release: request.Release,
 	})
 	if err != nil {
-		// This is a programming errror.
+		// This is a programming error.
```
ondrejbudai, line 158: :heart:

internal/worker/server_test.go
```diff
@@ -142,6 +142,43 @@ func TestUpdate(t *testing.T) {
...
+	jobId, err := server.EnqueueOSBuild(arch.Name(), &job)
+	require.NoError(t, err)
+
+	_, _, _, args, _, err := server.RequestJob(context.Background(), arch.Name(), []string{"osbuild"})
+	require.NoError(t, err)
+	require.NotNil(t, args)
+
+	var jobArgs worker.OSBuildJob
+	rawArgs, err := server.JobArgs(jobId, &jobArgs)
+	if err != nil {
```
ondrejbudai, line 175: I suggest using `require.NoError()` (there are in total 4 cases of `if err != nil`).

ondrejbudai (changes requested):
Yeah, I have the same concern as Christian. Otherwise, this looks awesome! :heart_eyes: I have just minor nits inline.

internal/kojiapi/api/openapi.yml
larskarlitski: Yeah, I'd leave it opaque.

larskarlitski (commented):
Really nice work! Nothing to add beyond the other comments.

As for testing, it might make sense to exercise this route once in `test/api/koji.sh`.

internal/kojiapi/api/openapi.yml
larskarlitski: Yeah, I'd leave it opaque.

achilleas-k:
Addressed all your comments.  It's now also rebased on #1158 since that changes the `fsjobqueue` testing which this PR adds to.

internal/kojiapi/api/openapi.yml
```diff
@@ -87,6 +87,36 @@ paths:
             text/plain:
               schema:
                 type: string
+  '/compose/{id}/manifest':
```
larskarlitski, line 90: This might be nitpicky, but should this not be `manifests` now? At least it should be documented as such, I think.

Which order are the manifests returned in?

achilleas-k: You're right.  I updated the description to mention it returns multiples, but the path and summary should be pluralised too.

The order of the manifests _should_ be the same order as when submitted.  The order should be stable for the `fsjobqueue` since dependencies are defined in an array in memory and on disk.  This will likely become an issue with the `dbjobqueue` ([see relevant comment](https://github.com/osbuild/osbuild-composer/discussions/1149#discussioncomment-275388)).

Logs use the same logic (and will have the same issue).

larskarlitski: Yeah I was wondering about how we handle logs. Thanks!

internal/kojiapi/api/openapi.yml
gicmo: This has already been addressed, but just for the record: might indeed change.

internal/worker/server_test.go
```diff
@@ -172,6 +172,39 @@ func TestUpdate(t *testing.T) {
...
+	imageType, err := arch.GetImageType("qcow2")
+	require.NoError(t, err)
+	manifest, err := imageType.Manifest(nil, distro.ImageOptions{Size: imageType.Size(0)}, nil, nil, nil, 0)
+	require.NoError(t, err)
+
+	tempdir, err := ioutil.TempDir("", "worker-tests-")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempdir)
+	server := newTestServer(t, tempdir)
+	server.Handler()
```
ondrejbudai, line 188: This is a no-op. Maybe we want to actually use `test.TestRoute` to test the also that the API works as expected?

achilleas-k: Thanks for catching that.  Probably missed that it's not necessary when copying from other tests.

There's no API endpoint for getting args from the **worker**.  It just grabs it directly from the `fsJobQueue`.  The `newTestServer()` is used to initialise the job queue to test the worker methods that wrap the new `JobQueue` methods.

teg (approved):
Nice work!

internal/worker/server_test.go
ondrejbudai: Right, sorry!

ondrejbudai (approved):
Awesome!

PR #1156, opened 2021-01-11
 🖼️ test/image: Fix test cases directory path in doc and code

thozza:
The directory with image-tests test cases has been renamed from `cases`
to `manifests`. This has not been previously reflected in the test/README.md
and osbuild-image-tests code. osbuild-image-tests hardcodes the test
cases directory path and uses it in case no test case are passed
to it on the command line. Since the image_tests.sh CI test case looks
for image-tests test cases in the correct directory and passes the
relevant ones to osbuild-image-tests, the CI didn't detect this issue.

Running osbuild-images-tests without any argument and let it run all
test cases from the default test cases directory as part of CI probably
does not make sense. Due to this reason, I'm not adding any new test.

### DIFF
--- a/cmd/osbuild-image-tests/constants/constants.go
+++ b/cmd/osbuild-image-tests/constants/constants.go
@@ -31,7 +31,7 @@ var TestPaths = struct {
 }{
 	ImageInfo:               "/usr/libexec/osbuild-composer-test/image-info",
 	PrivateKey:              "/usr/share/tests/osbuild-composer/keyring/id_rsa",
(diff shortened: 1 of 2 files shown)


ondrejbudai (approved):
Thanks and sorry for the confusion!

thozza:
> Thanks and sorry for the confusion!

No problem. At least there is no doubt that I look at it and tried to use it :smile:

ondrejbudai:
PSI OpenStack is misbehaving, otherwise we're green.

PR #1158, opened 2021-01-11
jobqueue: remove testjobqueue

larskarlitski:
testjobqueue did not implement the JobQueue interface correctly (noted
in its package comment), making it impossible to write tests for
JobQueue itself.

Replace its use everywhere with fsjobqueue operating on a temporary
directory.


internal/client/unit_test.go
```diff
@@ -37,7 +37,11 @@ func executeTests(m *testing.M) int {
 	}
 
 	// Create a mock API server listening on the temporary socket
-	fixture := rpmmd_mock.BaseFixture()
+	err = os.Mkdir(tmpdir+"/jobs", 0755)
```
ondrejbudai, line 40: `path.Join` would be a bit cleaner but that's just the tiniest nitpick ever.

ondrejbudai (dismissed):
Looks good! :)

larskarlitski:
Update to remove `t.TempDir()`.

internal/client/unit_test.go
larskarlitski: Fixed, because I had to make a change anyway.

achilleas-k: Since we're picking nits, `filepath.Join` would be more appropriate, but I think we're going to be fine with `path.Join` as well :)

achilleas-k (approved):
LGTM

larskarlitski:
As far as I can see, only openstack-boot tests failed due to PSI issues. @ondrejbudai could you please confirm?

ondrejbudai:
Yeah, this change doesn't interfere with OpenStack in any way. :/

ondrejbudai (approved):
Awesome, thanks! :)

PR #1159, opened 2021-01-12
distro/{rhel8,rhel84}: set systemd default targets

jkozol:
An image only had a systemd stage added if its blueprint contained or if its image type contained enabled services. The systemd
stage is now also added if the image type contains disabled services or a default target.

The RHEL 8.4 qcow2 image type now specifies the multi-user default target.

In order to test this the image-info tool now includes the default target in its output. Image test manifests are updated to include this change.

### DIFF
--- a/internal/distro/rhel8/distro.go
+++ b/internal/distro/rhel8/distro.go
@@ -288,7 +288,7 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 		p.AddStage(osbuild.NewUsersStage(options))
 	}
 
-	if services := c.GetServices(); services != nil || t.enabledServices != nil {
+	if services := c.GetServices(); services != nil || t.enabledServices != nil || t.disabledServices != nil || t.defaultTarget != "" {
 		p.AddStage(osbuild.NewSystemdStage(t.systemdStageOptions(t.enabledServices, t.disabledServices, services, t.defaultTarget)))
 	}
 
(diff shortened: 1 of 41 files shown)


teg (dismissed):
Nice! And this was unintentionally left out of 8.3, I guess, as the config is there, just not applied?

jkozol:
> Nice! And this was unintentionally left out of 8.3, I guess, as the config is there, just not applied?

@teg Yes. This was fairly invisible since if there were enabled services we would apply the other parameters.

jkozol:
@teg I don't believe the test failure is related to this PR. Could I get another review and merge this?

ondrejbudai (approved):
Looks good! The CI failures are definitely unrelated, PSI OpenStack had several outages during the last days.

thozza:
Hey. I just noticed that the default target for images like `rhel_8-s390x-qcow2-boot.json` is `graphical.target`. Is that actually correct? I'm not expert on mainframes, but it seems weird to me that someone would use the qcow2 image on s390x with graphical UI...

jkozol:
@thozza I agree that that is a bit weird. But, it is expected from what is currently implemented. This PR specifically fixed the default target for RHEL 8.4 qcow2 images.

thozza:
@jkozol sure, I'm not saying that this PR did a wrong change, just pointing out that while 8.4 image's default-target has been fixed, it should be probably changed also for 8.3. Although one test case for `rhel_8-s390x-tar-boot.json` has been missed by this PR, I'm going to submit another PR with updated repos for 8.3 and 8.4, so I'll regenerate it anyway...

PR #1162, opened 2021-01-14
distro/rhel84: fix s390x kernel options

jkozol:
Our s390x images now use the kernel commandline options set in the image type's declaration.

This should fix the missing kernel cmdline args pointed out by Wei Shi.


ondrejbudai (approved):

ondrejbudai:
Openstack failure, otherwise we're green.

PR #1163, opened 2021-01-14
schutzbot: prolong the timeout to 12 hours

ondrejbudai:
With our limited number of machines in OpenStack, we cannot run unlimited jobs
in parallel. When there's a lot of activity in the repository, there's a high
probability that a job must wait for a free VM for quite a long time. In these cases,
tests commonly fail due to the timeout.

I decided to prolong the timeout to 12 hours. This should lower the amount of
these kinds of jobs.


achilleas-k (approved):
Very much in favour of this.  It's not nice seeing red `x`s on PRs just because of timeouts.

PR #1165, opened 2021-01-18
Jobqueue: get all job parameters

### LINKED ISSUE #1154: KojiAPI: Server panic if job ID matches InitJob
This is a pretty minor and unrealistic edge case.

The Koji API handlers require the ID of a Finalize Job because that's the job ID that's returned to the caller.  If the ID of an Init Job is submitted, the `JobStatus()` call to the worker succeeds (it's a valid Job ID) but the dependencies are empty.  The server assumes non-zero dependencies and tries to index the `deps` (see below).

https://github.com/osbuild/osbuild-composer/blob/d9f09c66f2ea4fd8ea15ca8a267e94fe988c0310/internal/kojiapi/server.go#L305
https://github.com/osbuild/osbuild-composer/blob/d9f09c66f2ea4fd8ea15ca8a267e94fe988c0310/internal/kojiapi/server.go#L358

Realistically, this can only happen if a caller manages to _guess_ the UUID of an Init Job, but perhaps we should detect this and return a 404 anyway.

achilleas-k:
This pull request includes:

*Not sure if it's adequate, but it covers some error conditions as well as successful calls.*

## New JobQueue method

This PR replaces the recently introduced `JobArgs()` method of the `JobQueue` interface (see #1155) with a more general `Job()` method that returns all the arguments that were supplied to the `Enqueue()` method: `jobType`, `args`, and `dependencies`.

## Job type verification

Having access to the `jobType` makes it possible to validate the type when reading jobs from the queue, which is relevant for the Koji API.

The Koji API now includes convenience methods for reading each job type from the queue, resulting in errors if the ID doesn't match the expected type.  A downside to these functions is that it becomes harder to distinguish between an ID that doesn't exist in the queue and a type mismatch, though the error message does differ.  I'm not sure if the error message for a type mismatch is appropriate to be returned to an API client (e.g., `Job <ID> not found: expected "koji-finalize", found "koji-init" job instead`).
- Is this information relevant to the client?
- Does a 404 Not Found make sense for this error?

The type verification is also now also used when fetching job status and logs.

## Docs

Added a news item in `docs/news/unreleased` describing the introduction and purpose of the new API endpoint.  Since we didn't agree on exact naming, formatting, or linking, let me know if you think I should make any changes.

I linked to the relevant PRs (this one and #1155 where it was introduced).  I figure that might make it easier to add it to release notes in case there are any gaps in the description.

## Issues

Closes #1154

### DIFF
--- a/docs/news/unreleased/koji-manifests.md
+++ b/docs/news/unreleased/koji-manifests.md
@@ -1,0 +1,9 @@
+# Koji API: New endpoint for getting the manifests of a compose job
+
+A new endpoint is available in the Koji API: `GET /compose/{ID}/manifests`.
+Returns the manifests for a running or finished compose. Returns one manifest
+for each image in the request, in the order they were defined.
+
+Relevant PRs:
+https://github.com/osbuild/osbuild-composer/pull/1155
+https://github.com/osbuild/osbuild-composer/pull/1165
--- a/internal/jobqueue/fsjobqueue/fsjobqueue.go
+++ b/internal/jobqueue/fsjobqueue/fsjobqueue.go
@@ -287,13 +287,16 @@ func (q *fsJobQueue) JobStatus(id uuid.UUID) (result json.RawMessage, queued, st
 	return
 }
 
-func (q *fsJobQueue) JobArgs(id uuid.UUID) (args json.RawMessage, err error) {
+func (q *fsJobQueue) Job(id uuid.UUID) (jobType string, args json.RawMessage, dependencies []uuid.UUID, err error) {
 	j, err := q.readJob(id)
 	if err != nil {
 		return
 	}
 
+	jobType = j.Type
 	args = j.Args
+	dependencies = j.Dependencies
+
 	return
 }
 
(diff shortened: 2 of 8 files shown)


larskarlitski (approved):
This is amazing, thanks!

I think it makes a lot of sense to strictly validate incoming ids.

> * Is this information relevant to the client?
> * Does a 404 Not Found make sense for this error?

I think so, yes, for both. Going forward, I think it could make sense to include machine-readable error ids for these cases. But there's no framework to do so just yet.

> Added a news item in docs/news/unreleased describing the introduction and purpose of the new API endpoint. Since we didn't agree on exact naming, formatting, or linking, let me know if you think I should make any changes.

It's perfect to me. Thanks!

ondrejbudai (approved):
Awesome! :heart_eyes: I agree with all your points.

ondrejbudai:
PSI OpenStack has another bad day, otherwise, we're green.

elianerpereira:
@achilleas-k may I suggest a nitpicking for the paragraph below, just to remove the double "also":
"The type verification is now also used when fetching job status and logs." What do you think? Thank you. :-)

achilleas-k:
Edited.  Thanks!

PR #1166, opened 2021-01-18
cloudapi: expose upload status

chloenayon:
Expose a more detailed job status result - specifically, include upload status alongside
image status. Expand openapi.yml accordingly and add an UploadStatus field to the 
OSBuildJobResult struct. At the moment, only represent the "success" and "failure" states 
of UploadStatus - to differentiate between "pending" and "running" would involve significant 
design decisions and should be addressed in a separate commit.

This pull request includes:

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -145,10 +145,12 @@ do
     --cert /etc/osbuild-composer/client-crt.pem \
     https://localhost/api/composer/v1/compose/"$COMPOSE_ID")
 
-  COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.status')
+  COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.status')
+  UPLOAD_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.upload_status')
 
   if [[ "$COMPOSE_STATUS" != "pending" && "$COMPOSE_STATUS" != "running" ]]; then
     test "$COMPOSE_STATUS" = "success"
+    test "$UPLOAD_STATUS" = "success"
     break
   fi
 
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -146,11 +146,13 @@ do
     https://localhost/api/composer/v1/compose/"$COMPOSE_ID")
 
   COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.status')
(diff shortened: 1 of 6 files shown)


Gundersanne:
Looks good! 

The test failures are due to the status response having changed, needs a tweak here https://github.com/osbuild/osbuild-composer/blob/01c21b5a47dc6d96e7a81b204902d904790e5e91/test/cases/api.sh#L148 at least.

test/cases/api.sh
```diff
@@ -145,7 +145,7 @@ do
     --cert /etc/osbuild-composer/client-crt.pem \
     https://localhost/api/composer/v1/compose/"$COMPOSE_ID")
 
-  COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.status')
+  COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.status')
 
   if [[ "$COMPOSE_STATUS" != "pending" && "$COMPOSE_STATUS" != "running" ]]; then
     test "$COMPOSE_STATUS" = "success"
```
Gundersanne, line 151: We could tes tthe upload status by adding a test here by extracting `upload_status` from the COMPOSE_STATUS here and see if it's `success` too.

Gundersanne (commented):
The idea looks good, I'd add a test though. It should be easy enough to expand our current api test to include checking the upload_status.

Gundersanne (approved):
The only failed test now is `fedora_33-x86_64-fedora_iot_commit-boot.json`, unrelated to the PR. Also it succeeded in pipeline #7 which also ran on commit `aa355f117a38c8d22222fb53bff37834601eb285`.

I'd say we merge this.

larskarlitski:
It looks to me like the failure is known and tracked in https://github.com/osbuild/osbuild-composer/issues/1196. Feel free to merge,

thozza:
Hey.

I noticed that after this PR, there is an unused component in the cloudapi scheme:
https://github.com/osbuild/osbuild-composer/blob/d686abfffbb4d805aa55ef3a27673ad7561e3849/internal/cloudapi/openapi.yml#L119-L124
This has been previously not really used, but I thought that eventually we will return more information related to the upload status, than just the type of the upload. E.g for AWS it would be the `ami_id`, for GCP this would be `project_id` and `image_name`. I expect that we will have to return more granular information, which should be presented to the user, so that they can identify the image and use it with the cloud provider. Isn't this change a potential step away from that direction?

PR #1172, opened 2021-01-20
Apply tags to registered AMI

### LINKED ISSUE #1171: Tags missing on AMI after import
We are applying tags to the snapshot in AWS, but not to the AMI itself after import. AWS does have an AMI name field, but that doesn't appear in the `Name` field within most of the AWS UI.

major:
Adding the tag called `Name` to the AMI ensures that the name appears in
the *Name* column inside AWS' web console.

Fixes #1171.

This pull request includes:

### DIFF
--- a/internal/upload/awsupload/awsupload.go
+++ b/internal/upload/awsupload/awsupload.go
@@ -243,6 +243,23 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	log.Printf("[AWS] 🎉 AMI registered: %s", *registerOutput.ImageId)
 
+	// Tag the image with the image name.
+	req, _ = a.ec2.CreateTagsRequest(
+		&ec2.CreateTagsInput{
+			Resources: []*string{registerOutput.ImageId},
+			Tags: []*ec2.Tag{
+				{
+					Key:   aws.String("Name"),
+					Value: aws.String(name),
+				},
+			},
+		},
+	)
+	err = req.Send()
+	if err != nil {
+		return nil, err
+	}
+
 	if len(shareWith) > 0 {
 		log.Println("[AWS] 💿 Sharing ec2 AMI")
 		var launchPerms []*ec2.LaunchPermission
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -263,6 +263,12 @@ for LOOP_COUNTER in {0..10}; do
     sleep 5
 done
 
+# Ensure the image was properly tagged.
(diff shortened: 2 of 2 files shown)


internal/upload/awsupload/awsupload.go
```diff
@@ -243,6 +243,23 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	log.Printf("[AWS] 🎉 AMI registered: %s", *registerOutput.ImageId)
 
+	// Tag the image with the image name.
+	req, _ := a.ec2.CreateTagsRequest(
```
ondrejbudai, line 247:
```suggestion
	req, _ = a.ec2.CreateTagsRequest(
```

Go doesn't like that you use `:=` without introducing a new variable.

Also, a quick test in `aws.sh` might be very cool. Otherwise, the change looks good! :-)

larskarlitski: Is this ignoring the error return?

major: Ah, thank y'all for that. My golang is still improving.

test/cases/aws.sh
```diff
@@ -263,6 +263,12 @@ for LOOP_COUNTER in {0..10}; do
     sleep 5
 done
 
+# Ensure the image was properly tagged.
+IMAGE_TAG=$($AWS_CMD ec2 describe-images --image-ids "${AMI_IMAGE_ID}" jq -r '.Images[0].Tags[] | select(.Key=="Name") | .Value')
```
ondrejbudai, line 267:
```suggestion
IMAGE_TAG=$($AWS_CMD ec2 describe-images --image-ids "${AMI_IMAGE_ID}" | jq -r '.Images[0].Tags[] | select(.Key=="Name") | .Value')
```

You lost a pipe.

image

major: Oopsies. It should be fixed now. 👨🏻‍🔧

internal/upload/awsupload/awsupload.go
```diff
@@ -243,6 +243,23 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	log.Printf("[AWS] 🎉 AMI registered: %s", *registerOutput.ImageId)
 
+	// Tag the image with the image name.
```
teg, line 246: What's the relation to the `Name` specified when importing the image? Is the difference just what is used in the UI?

major: Right, the `Name` tag key is a special one and it's used in the UI in AWS when you display an AMI.

ondrejbudai (approved):
Looks good!

PR #1173, opened 2021-01-22
distro/rhel84: include timedatex in qcow2 image

jkozol:
timedatex was an exlcuded package due to an selinux-policy issue. This
issue is resolved and timedatex is no longer excluded.

This pull request includes:

I do not believe this has adequate documentation about the change but I am unsure where I would document our package list.

### DIFF
--- a/docs/news/unreleased/timedatex.md
+++ b/docs/news/unreleased/timedatex.md
@@ -1,0 +1,7 @@
+# RHEL 8.4: Include timedatex in qcow2 images
+
+Timedatex was an excluded package due to an selinux-policy issue that has been
+fixed. Therefore, timedatex should be in the qcow2 image we build. Our list of 
+excluded packages for RHEL 8.4 was not being included in our nightly builds so 
+we did not realize that timedatex was still being excluded. The issue with the 
+excluded packages is now fixed and timedatex is now removed from this list.
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -953,10 +953,6 @@ func New() distro.Distro {
 			"fedora-release",
 			"fedora-repos",
 			"rng-tools",
-
-			// TODO setfiles failes because of usr/sbin/timedatex. Exlude until
-			// https://errata.devel.redhat.com/advisory/47339 lands
-			"timedatex",
 		},
 		defaultTarget:           "multi-user.target",
 		kernelOptions:           "console=tty0 console=ttyS0,115200n8 no_timer_check net.ifnames=0 crashkernel=auto",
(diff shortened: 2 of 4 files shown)


marusak:
Thank you @jkozol for looking into this!

I was honestly surprised when I saw this fix. Was not expecting it just being excluded by osbuild. When I was opening [1] (which this PR fixes) I thought that either RCM decided to drop/replace it or that maintainers are migrating to some other tooling and there is a sync issue. I opened the issue against `timedatex` even though I wasn't thinking it is really issue in this component but rather somewhere else. But I was expecting that maintainers would be aware if there was something happening and they would be able to just tell me, what's the problem. But seems they had no idea it was shadowed by osbuild.

That brings me to my question - should osbuild, a building tool, be able to decide what goes into official images and what not? I thought this decision is made by RCM and other entities and osbuild just puts those things into image. If the image cannot be composed due to a bug, then image should not be produced?
Also skipping it somewhere in depths of github is tricky, as it may get unnoticed for a long time until some customer would complain, which would be a worse case. If it has to be skipped in this way, then maybe having some automatization that would check if it still applies? (as does cockpit with naughties)


[1] https://bugzilla.redhat.com/show_bug.cgi?id=1918340

marusak:
> Also skipping it somewhere in depths of github is tricky, as it may get unnoticed for a long time until some customer would complain, which would be a worse case. If it has to be skipped in this way, then maybe having some automatization that would check if it still applies? (as does cockpit with naughties)

That linked errata is for rhel-8.2, so this possibly was not even needed for 8.4 in the first place, it was just copied from `internal/distro/rhel8/distro.go` (where it is still now) when 8.4 was being introduced in #1089?

teg:
@jkozol could you redo the verification of the package list of the last images produced by image factory with our current image, to make sure there is nothing else we accidentally missed? Otherwise: looks good to me, and I don't think any explicit documentation is needed for this, but you could add a news entry (possibly explaining what happened).

The context is that we had a bug that did not apply the exclusions at all, this was detected by QE, we fixed that and applied this old list which had not been properly verified. @marusak is right that this particular package should not have been excluded at all and is a left-over. I agree that having some verification of this sort of workaround would have been great, and we should look into that before introducing something like it again. I hope we will not need to do that though.

@marusak, the ownership of the image definitions are moving from EXD to devel, in the same way the compose itself is for RHEL9. We may still want to work on making this more discoverable, but long-term our goal is for customers running this on-prem to get exactly the same images as if they were built as part of the compose, and EXD is just another consumer.

teg (changes requested):
Great stuff! Please put in a release note, and we are good to go!

teg (approved):

Issue #1177, opened 2021-01-26
image-tests: enable testing of image customizations done via `ImageOptions` (e.g. via Cloud API)


thozza:
Image test cases can be currently generated using the `tools/test-case-generators/generate-test-cases` tool only for images customized via Blueprint customizations. However there are some types of image customizations done e.g. using Cloud API, which can't be done via Blueprint customizations, but only via [`ImageOptions`](https://github.com/osbuild/osbuild-composer/blob/2618e11bfecc580e589874a17137241f3eb73d78/internal/distro/distro.go#L86-L89) structure. Example of such customization is [`subscription` of RHEL images using `subscription-manager`](https://github.com/osbuild/osbuild-composer/blob/2618e11bfecc580e589874a17137241f3eb73d78/internal/cloudapi/server.go#L142-L150). `tools/test-case-generators/generate-test-cases` and `cmd/osbuild-pipeline` tools should be enhanced to allow image customizations possible to achieve only through [`ImageOptions`](https://github.com/osbuild/osbuild-composer/blob/2618e11bfecc580e589874a17137241f3eb73d78/internal/distro/distro.go#L86-L89) structure.

Along with the support for testing image customizations done via [`imageOptions`](https://github.com/osbuild/osbuild-composer/blob/2618e11bfecc580e589874a17137241f3eb73d78/internal/distro/distro.go#L86-L89), test cases for each [possible customization allowed via `imageOptions`](https://github.com/osbuild/osbuild-composer/blob/2618e11bfecc580e589874a17137241f3eb73d78/internal/distro/distro.go#L86-L89) should be added as well.

PR #1178, opened 2021-01-26
Image test cases tooling fix & data update

thozza:
This pull request includes:
- Fix of a bug in `generate-test-cases` tool, which caused the blueprint customization `overrides` from `format-request-map.json` to not be properly applied and used in the resulting test case manifests.
- Removed the `subscription` customization from the `tools/test-case-generators/format-request-map.json` as it did not affect generated image test cases. Images currently can not be subscribed via Blueprint customizations, only via Cloud API, which uses `imageOptions` to modify the resulting image. Our current tooling for image test cases relies only on Blueprint customizations for image customizations and does not allow `imageOptions` to be modified. Created issue #1177 so that we don't forget to add the support to our tooling and actual test cases for RHEL image subscription in the future.
- Updated `tools/test-case-generators/repos.json` to use the latest RHEL-8.3 and RHEL-8.4 RPMrepo snapshots.
- Regenerated all image test cases affected by changes done to `tools/test-case-generators/format-request-map.json` and `tools/test-case-generators/repos.json`, namely all `rhel-8` and `rhel-84` images.


thozza:
There seems to be an issue with the changes to image test cases for `rhel-84`. I'll have to investigate it...

thozza:
I'm currently blocked on #1189 to get merged, so that I can regenerate the `rhel-84` `qcow2` image test cases...

ondrejbudai:
#1189 is now merged, so if you can rebase this PR, we should be good to go. :)

thozza:
Thanks! Regeneration of image test cases is in progress... I'll post an update here, once the PR is updated with the latest image test cases...

teg (approved):
Thanks for all the work on this!

PR #1179, opened 2021-01-26
[RFC] tools: Add script to generate all test cases in VMs

thozza:
Add the `generate-all-test-cases` tool, which generates all defined image
test cases based on provided "distro x arch x image-type" combination
matrix and CLI arguments. Test cases are generated by running the
`generate-test-cases` tool inside an architecture-specific VM.

The tool requires a qcow2 image to be provided for each requested
architecture. These images are used to create ephemeral VMs used
for test case generation.

There appears to be an issue with entropy on all VMs for other
architecture than the host's one. It makes the generation of the first
test case usually fail due to the `go get` command failing with a "TLS
handshake timeout" error. There is a workaround to retry the generation
of a test case a maximum of 3 times in this case. In the usual case,
the test case generation passes on the second try. All subsequent test
case generations should complete successfully on the first try.

Add the `distro-arch-imagetype-map.json` file as the default source of
the "distro x arch x image-type" combination matrix for generating test
cases. Its content is based on the current content of the
`test/data/manifests` directory.

Update `test/README.md` to mention the `generate-all-test-cases` tool.
Add a note, that `generate-test-cases` must be run on a host with the
same architecture, as the intended test case's.

**Needs discussion:**
- The content of `distro-arch-imagetype-map.json` and its idea in general.
- Whether the tool works also on some else's computer 😇

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -69,6 +69,20 @@ boot-test the image.
 
 To (re)generate these test cases use the tool
 `tools/test-case-generators/generate-test-cases`.
+Note that the `generate-test-cases` tool must be run on a host with
+the same architecture, as the one intended for the generated test
+cases. In other words, you need to generate e.g test cases for `aarch64`
+images on an `aarch64` host.
+
+Alternatively to (re)generate test cases for all architectures, or just
+the ones different from your host's architecture, you can use the tool
+`tools/test-case-generators/generate-all-test-cases`. It creates
+an ephemeral virtual machine for each necessary architecture using the
+`qemu-system-<arch>` command and generates test cases using the
+`generate-test-cases` tool inside the virtual machine. It is important
+to note that test case generation in virtual machines may take several
+hours. The `generate-all-test-cases` currently does not work with RHEL
+images because of missing "9p" filesystem support.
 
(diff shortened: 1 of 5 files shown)


teg (approved):
Love this! Happy to discuss the points you mentioned.

thozza:
Thanks for the feedback @teg. I added a note to `test/README.md` that the tool unfortunately does not work with RHEL images. The tool uses "9p virtio" to share directories from the Host to the Guest but unfortunately the "9p" filesysten is not supported on RHEL. The way that sources are shared with the Guest and the results are retrieved could be definitely reworked to support also RHEL images, but I would not spend more time on this at this moment. The PoC works fine with Fedora 33 images.

teg (approved):
Thanks for doing this Tomas!

ondrejbudai (approved):
Awesome, thanks! :-) 

I like the idea of `distro-arch-imagetype-map.json`, we can always refine it later if needed.

ondrejbudai:
(CI was unhappy but this was because of some Jenkins issues. As this change only code for development, we're safe to merge this.)

PR #1186, opened 2021-01-29
Bump go version to 1.14

ondrejbudai:
The oldest distros we support are Fedora 32 and RHEL 8.3. As both have Go 1.14, we're safe to upgrade.

This pull request includes:

     The tests that we already have are enough.
     Not sure, is this change that important? It feels to me that the change in README is enough.


internal/tools.go
```diff
@@ -0,0 +1,15 @@
+// +build tools
+
+// This file is here to just explicitly tell `go mod vendor` that we depend
+// on oapi-codegen. Without this file, `go generate ./...` in Go >= 1.14 gets
+// confused because oapi-codegen is not being vendored.
```
larskarlitski, line 5: Oh dear.

larskarlitski (approved):
Finally. Thanks! ❤️

teg (approved):
🥳

PR #1187, opened 2021-01-29
image-info: Add workaround for listing services by status

thozza:
This pull request includes:

  - Fixes an issue, affected test cases were updated 
  - No documentation other than comments in the code is included. This is not user-visible change
 
**Summary**
`image-info` tools parses output of `systemctl list-unit-files` run on a
different tree (with `--root` option), to determine the list of enabled
and disabled services on the inspected image. However since Fedora 33
(and presumably since systemd v246), the output of `systemctl
list-unit-files` changed. Some units previously reported as "enabled" or
"disabled" are now reported as "alias", which means, that they are just
a symlink to a different unit.

There is no systemd command, that would take an "alias" unit and would
report its state as "enabled" or "disabled" and could run on a different
tree (with "--root" option).

To make the list of reported services in the given state consistent on
systems with older and new (v246+) systemd version, check all "alias"
units and append them to the list of services with a specific status,
if their target is also listed in in the list.

Example of the `systemctl list-unit-files` output change:
```
~]# rpm -q systemd
systemd-246.6-3.fc33.x86_64
~]# systemctl list-unit-files ctrl-alt-del.target
UNIT FILE           STATE VENDOR PRESET
ctrl-alt-del.target alias -
```
```
~]# rpm -q systemd
systemd-245.8-2.fc32.x86_64
~]# systemctl list-unit-files ctrl-alt-del.target
UNIT FILE           STATE   VENDOR PRESET
ctrl-alt-del.target enabled disabled
```
This change makes it possible to produce consistent output for an
inspected image, regardless if the `image-info` tool is run on Fedora
32, Fedora 33 or RHEL-8.

Also regenerate all Fedora 33 test cases, since this commit changes the
content of produced list of enabled / disabled services since Fedora 33.
The list is now consistent with what would be produced by `image-info`
for an image on older Fedora (e.g. 32) or RHEL-8.


ondrejbudai (approved):
:heart: Thanks!

ondrejbudai:
PSI OpenStack failure, clearly unrelated.

PR #1188, opened 2021-01-29
Add sysconfig stage

jkozol:
This pull request includes:

The org.osbuild.sysconfig stage is now supported. Config updates can be made to the kernel and network files. Currently, only the default values are used for all image types in rhel84. The image-info script is updated to allow testing the sysconfig info.

The default values are:
```
  kernel:
    UPDATEDEFAULT=yes
    DEFAULTKERNEL=kernel
    
  network:
    NETWORKING=yes
    NOZEROCONF=yes
```

### DIFF
--- a/test/data/manifests/rhel_84-x86_64-vhd-boot.json
+++ b/test/data/manifests/rhel_84-x86_64-vhd-boot.json
@@ -3499,6 +3499,19 @@
           "options": {
             "file_contexts": "etc/selinux/targeted/contexts/files/file_contexts"
           }
+        },
+        {
+          "name": "org.osbuild.sysconfig",
+          "options": {
+            "kernel": {
+              "update_default": true,
+              "default_kernel": "kernel"
+            },
+            "network": {
+              "networking": true,
+              "no_zero_conf": true
+            }
+          }
         }
       ],
       "assembler": {
@@ -10248,6 +10261,16 @@
       "unbound-anchor.timer",
       "waagent.service"
     ],
+    "sysconfig": {
+      "kernel": {
+        "DEFAULTKERNEL": "kernel",
+        "UPDATEDEFAULT": "yes"
+      },
+      "network": {
+        "NETWORKING": "yes",
+        "NOZEROCONF": "yes"
+      }
+    },
     "timezone": "New_York"
   }
 }
--- a/tools/image-info
+++ b/tools/image-info
@@ -358,6 +358,30 @@ def read_fstab(tree):
             result = sorted([line.split() for line in f if line and not line.startswith("#")])
     return result
 
+# Create a nested dictionary for all supported sysconfigs 
+def read_sysconfig(tree):
+    result = {}
+    sysconfig_paths = {
+        "kernel": f"{tree}/etc/sysconfig/kernel",
+        "network": f"{tree}/etc/sysconfig/network"
+    }
+    # iterate through supported configs
+    # based on https://github.com/osbuild/osbuild/blob/main/osbuild/util/osrelease.py#L17
+    for name, path in sysconfig_paths.items():
+        with contextlib.suppress(FileNotFoundError):
+            with open(path) as f:
+                # if file exists start with empty array of values
+                result[name] = {}
+                for line in f:
+                    line = line.strip()
+                    if not line:
+                        continue
+                    if line[0] == "#":
+                        continue
+                    key, value = line.split("=", 1)
+                    result[name][key] = value.strip('"')
+    return result
+ 
 
 def append_filesystem(report, tree, *, is_ostree=False):
     if os.path.exists(f"{tree}/etc/os-release"):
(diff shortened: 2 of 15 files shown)


tools/image-info
```diff
@@ -301,6 +301,20 @@ def read_fstab(tree):
...
+def read_sysconfig(tree):
+    result = {
+        "kernel": [],
+        "network": []
+    }
+    with contextlib.suppress(FileNotFoundError):
+        with open(f"{tree}/etc/sysconfig/kernel") as f:
+            result["kernel"] = f.read().splitlines()
+        with open(f"{tree}/etc/sysconfig/network") as f:
+            result["network"] = f.read().splitlines()
```
gicmo, line 313: If you put both open calls in just one `with` block then you wan't attempt to read the second file in case the first one don't exists.
Also maybe parse the actual key, value pairs. You should be able to re-use the code we use in osbuild to parse [`os-release`](https://github.com/osbuild/osbuild/blob/main/osbuild/util/osrelease.py#L17).
I think I would also not pre-fill the result dict so you don't get a lot of basically empty entries when the files don't exist.

tools/image-info
```diff
@@ -331,6 +345,10 @@ def append_filesystem(report, tree, *, is_ostree=False):
         if fstab:
             report["fstab"] = fstab
 
+        sysconfig = read_sysconfig(tree)
+        if sysconfig:
```
gicmo, line 349: Since your `result` currently always has two keys inside, the condition is always `true`.

gicmo (changes requested):
Overall looks good to me! I wonder if in the case of OSTree commits they should really go into the commit itself.

gicmo:
Schutzi is unhappy because the osbuild requirements was not updated so the sysconfig stage is actually not there, hence you get missing schema errors.

gicmo:
One other little other thing, in your commit you write:
```
Currently, only the default values are used for all image types in rhel84.
```
I stumbled a bit over `default`, since if they were the default we would not have to set them, right?

tools/image-info
jkozol: sysconfig's default will now be an empty dict if no files are found. This condition will then be false.

tools/image-info
jkozol: Updated this to use the parsing done in `os-release`

jkozol:
> I stumbled a bit over default, since if they were the default we would not have to set them, right?

@gicom I see what you're saying here. I changed the wording to just indicate that we are setting values for all rhel 8.4 image types.

teg (approved):
Great work @jkozol! Make sure to sync with @thozza on the merge conflicts, but as soon as that is taken care of and schutzbot is happy I think this is ready to be merged!

tools/image-info
gicmo: Not a blocker, but good practice (and sometimes required for copyright) is to mention where the `os-release` code is coming from (in the commit message or as a comment in the code)

test/data/manifests/rhel_84-x86_64-vhd-boot.json
```diff
@@ -9761,6 +9774,16 @@
...
+      },
+      "network": {
+        "NETWORKING": "yes",
+        "NOZEROCONF": "yes"
+      }
+    },
     "timezone": "New_York"
   }
-}
+}
```
gicmo, line 9789: Lets hope David doesn't see this :P

gicmo:
> @gicom

That's not my name! *sing* 😄

test/data/manifests/rhel_84-x86_64-vhd-boot.json
jkozol: hahaha shhhh. It's the test generator's fault...

ondrejbudai:
Blocked until osbuild 24 is in both Fedora 32 and Fedora 33:

https://bodhi.fedoraproject.org/updates/FEDORA-2021-1f28f28ea7
https://bodhi.fedoraproject.org/updates/FEDORA-2021-1f070f8684

gicmo:
> Blocked until osbuild 24 is in both Fedora 32 and Fedora 33:
> 
> https://bodhi.fedoraproject.org/updates/FEDORA-2021-1f28f28ea7
> https://bodhi.fedoraproject.org/updates/FEDORA-2021-1f070f8684

On its way now for F33, but needs karma or time for F32.

jkozol:
Also blocked until osbuild 24 is in rhel 8.4 nightly.

jkozol:
Just an update. The latest version of osbuild is now in the RHEL 8.4 nightly. Just waiting on f32 to make it into stable. It is currently on its magical journey from testing to stable. After this all the tests should pass.

gicmo:
Not blocked anymore as osbuild 24 should have made it to stable everywhere, restarting Schutzi.

teg (approved):

gicmo (approved):
🙆

PR #1189, opened 2021-01-29
distro/rhel84: remove rng-tools from qcow2

jkozol:
`rng-tools` was added back into the qcow2 packages by mistake. It should be an excluded package and `rngd.service` should not be enabled. After #1178 `rng-tools` won't need to be an excluded package since it won't be included by `@core` but it is easiest to explicitly exclude it for now.

This pull request includes:


thozza (approved):
Changes look good. Thanks for a quick fix!

ondrejbudai (approved):

PR #1191, opened 2021-02-02
image-tests: fixup for rhel-8 aarch64 edge test case after PR #1178

thozza:
@teg  was too quick to merge PR #1178 and as a result, one image test case
was not properly regenerated in that PR. The CI passed for PR #1178,
because RHEL-8.3 images in CI are not tested on aarch64.

Add properly regenerated image test case for rhel-8 aarch64 edge commit.


teg (approved):
Thanks and sorry!

PR #1192, opened 2021-02-02
cloudapi: expose AWS target information

chloenayon:
Extend OSBuildJobResult to contain a TargetResults field, and append 
to it AWS target information like amiID and region. Add TargetResults to 
the UploadStatus struct so that it is exposed on a cloudapi status call.

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -115,6 +115,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	}
 
 	var r []error
+	var targetResults []*target.TargetResult
 
 	for _, t := range args.Targets {
 		switch options := t.Options.(type) {
@@ -189,6 +190,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			if !osbuildOutput.Success {
 				continue
 			}
+
 			a, err := awsupload.New(options.Region, options.AccessKeyID, options.SecretAccessKey)
 			if err != nil {
 				r = append(r, err)
@@ -206,12 +208,21 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				continue
 			}
 
-			/* TODO: communicate back the AMI */
-			_, err = a.Register(t.ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
+			ami, err := a.Register(t.ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
 			if err != nil {
 				r = append(r, err)
 				continue
 			}
+
+			if ami == nil {
+				r = append(r, fmt.Errorf("No ami returned"))
+				continue
+			}
+
+			targetResults = append(targetResults, target.NewAWSTargetResult(&target.AWSTargetResultOptions{
+				Ami:    *ami,
+				Region: options.Region,
+			}))
 		case *target.AzureTargetOptions:
 			if !osbuildOutput.Success {
 				continue
@@ -406,6 +417,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		Success:       osbuildOutput.Success && len(targetErrors) == 0,
 		OSBuildOutput: osbuildOutput,
 		TargetErrors:  targetErrors,
+		TargetResults: targetResults,
 		UploadStatus:  uploadstatus,
 	})
 	if err != nil {
--- a/internal/cloudapi/openapi.yml
+++ b/internal/cloudapi/openapi.yml
@@ -110,16 +110,23 @@ components:
       required:
         - status
         - type
+        - options
       properties:
         status:
           type: string
           enum: ['success', 'failure', 'pending', 'running']
         type:
           $ref: '#/components/schemas/UploadTypes'
+        options:
+          oneOf:
+            - $ref: '#/components/schemas/AWSUploadStatus'
     AWSUploadStatus:
       type: object
+      required:
+        - ami
+        - region
       properties:
-        ami_id:
+        ami:
           type: string
           example: 'ami-0c830793775595d4b'
     GCPUploadStatus:
--- a/internal/worker/json.go
+++ b/internal/worker/json.go
@@ -21,10 +21,11 @@ type OSBuildJob struct {
 }
 
 type OSBuildJobResult struct {
-	Success       bool            `json:"success"`
-	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
-	TargetErrors  []string        `json:"target_errors,omitempty"`
-	UploadStatus  string          `json:"upload_status"`
+	Success       bool                   `json:"success"`
+	OSBuildOutput *osbuild.Result        `json:"osbuild_output,omitempty"`
+	TargetResults []*target.TargetResult `json:"target_results,omitempty"`
+	TargetErrors  []string               `json:"target_errors,omitempty"`
+	UploadStatus  string                 `json:"upload_status"`
 }
 
 type KojiInitJob struct {
--- a/internal/cloudapi/openapi.gen.go
+++ b/internal/cloudapi/openapi.gen.go
@@ -42,6 +42,11 @@ type AWSUploadRequestOptionsS3 struct {
 	SecretAccessKey string `json:"secret_access_key"`
 }
 
+// AWSUploadStatus defines model for AWSUploadStatus.
+type AWSUploadStatus struct {
+	Ami string `json:"ami"`
+}
+
 // ComposeRequest defines model for ComposeRequest.
 type ComposeRequest struct {
 	Customizations *Customizations `json:"customizations,omitempty"`
@@ -137,8 +142,9 @@ type UploadRequest struct {
 
 // UploadStatus defines model for UploadStatus.
 type UploadStatus struct {
-	Status string      `json:"status"`
-	Type   UploadTypes `json:"type"`
+	Options interface{} `json:"options"`
+	Status  string      `json:"status"`
+	Type    UploadTypes `json:"type"`
 }
 
 // UploadTypes defines model for UploadTypes.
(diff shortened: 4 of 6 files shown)


Gundersanne (changes requested):
The []string approach will be difficult to use on the other end. I think we need a different approach where we do something similar here as we do in `internal/stage`. 

So we'd have different sets of options (TargetAWSOptions as a first), and then cast depending on the type into the relevant options.

```
type TargetResult struct {
        Type string `json:"type"`
        Options TargetResultOptions `json:"options"`
}

type TargetResultOptions interface {
	isTargetResultOptions()
}

type TargetResultAWSOptions struct {
        Ami string `json:"ami"`
        Region string `json:"region"`
}

func (TargetResultAWSOptions) isTargetResultOptions() {}
```

ObuildJobResult would change to:
```
type OSBuildJobResult struct {
	Success       bool            `json:"success"`
	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
	TargetResults []TargetResult  `json:"target_results,omitempty"`
	TargetErrors  []string        `json:"target_errors,omitempty"`
	UploadStatus  string          `json:"upload_status"`
}
```

Gundersanne:
So where this needs changes:

- [ ] Declare structs in internal/target; similar to Target/TargetOptions there, make TargetResult/TargetResultOptions
- [ ] Change OSBuildJobResult to have `	TargetResults []*target.TargetResult  `json:"target_results,omitempty"`` field
- [ ] In jobImpl-osbuild make the list of targetresults, aws would look something like:
```
opts := target.TargetResultAWSOptions{
	Ami: *ami, //might need to check if Register ever returns nil for the ami pointer
	Region: options.Region,
}

targetResults = append(targetResults, target.NewAWSTargetResult(&opts))
```

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -160,11 +162,21 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			}
 
 			/* TODO: communicate back the AMI */
```
teg, line 164: Can drop the `TODO` now :)

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -160,11 +162,21 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
-			_, err = a.Register(t.ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
+			ami, err := a.Register(t.ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
 			if err != nil {
 				r = append(r, err)
 				continue
 			}
+
+			var amiResult string = ""
+			if ami != nil {
```
teg, line 172: This should probably be treated as an error, don't you think? Probably the underlying library already does, but just in case, it might makes sense to error out here. Also feel free to leave as is, up to you!

teg (dismissed):
Looking great!

cmd/osbuild-worker/jobimpl-osbuild.go
chloenayon: You're right, thanks for pointing it out!

thozza (commented):
> ```
> type OSBuildJobResult struct {
> 	Success       bool            `json:"success"`
> 	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
> 	TargetResults []TargetResult  `json:"target_results,omitempty"`
> 	TargetErrors  []string        `json:"target_errors,omitempty"`
> 	UploadStatus  string          `json:"upload_status"`
> }
> ```
@Gundersanne I wonder whether it would not make more sense to have the `UploadStatus` and `TargetErrors` merged into the `TargetResults`. This way it would be easier to connect the specific requested target with any related errors, upload status and any target-specific results. With the two distinct lists, it may not be that easy to connect specific `TargetError` to a `TargetResult`.

My idea would be something like:
```
type TargetResult struct {
        Type string `json:"type"`
        Error string `json:"error"`
        UploadStatus `json:"upload_status"`
        Options TargetResultOptions `json:"options"`
}

type TargetResultOptions interface {
	isTargetResultOptions()
}

type TargetResultAWSOptions struct {
        Ami string `json:"ami"`
        Region string `json:"region"`
}

func (TargetResultAWSOptions) isTargetResultOptions() {}
```

```
type OSBuildJobResult struct {
	Success       bool            `json:"success"`
	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
	TargetResults []TargetResult  `json:"target_results,omitempty"`
}
```
My main reasoning is that if we want to keep the option to have multiple targets for a compose request in the future, we should ideally keep all the information related to specific target in the same data structure. This will make it easier to make sense out of the results once we will get more than one target results.

What do you think?

thozza:
> ```
> type OSBuildJobResult struct {
> 	Success       bool            `json:"success"`
> 	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
> 	TargetResults []TargetResult  `json:"target_results,omitempty"`
> 	TargetErrors  []string        `json:"target_errors,omitempty"`
> 	UploadStatus  string          `json:"upload_status"`
> }
> ```
@Gundersanne I wonder whether it would not make more sense to have the `UploadStatus` and `TargetErrors` merged into the `TargetResults`. This way it would be easier to connect the specific requested target with any related errors, upload status and any target-specific results. With the two distinct lists, it may not be that easy to connect specific `TargetError` to a `TargetResult`.

My idea would be something like:
```
type TargetResult struct {
        Type string `json:"type"`
        Error string `json:"error"`
        UploadStatus `json:"upload_status"`
        Options TargetResultOptions `json:"options"`
}

type TargetResultOptions interface {
	isTargetResultOptions()
}

type TargetResultAWSOptions struct {
        Ami string `json:"ami"`
        Region string `json:"region"`
}

func (TargetResultAWSOptions) isTargetResultOptions() {}
```

```
type OSBuildJobResult struct {
	Success       bool            `json:"success"`
	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
	TargetResults []TargetResult  `json:"target_results,omitempty"`
}
```
My main reasoning is that if we want to keep the option to have multiple targets for a compose request in the future, we should ideally keep all the information related to specific target in the same data structure. This will make it easier to make sense out of the results once we will get more than one target results.

What do you think?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -113,6 +113,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	}
 
 	var r []error
+	var targetResults []*target.TargetResult
```
thozza, line 116: My suggestion would be to merge the `r` with `targetResults`, so that any errors relevant to the target are part of its result. This will in my opinion make it easier to connect errors and target-specific results on the caller side, once we would allow more than one target to be specified.

internal/cloudapi/openapi.yml
```diff
@@ -110,18 +110,45 @@ components:
       required:
         - status
         - type
```
thozza, line 112: `type` from `UploadStatus` should be probably removed in favor of the `TargetResult`'s `name` property which technically holds the same information and also identifies the type of the upload target.

internal/cloudapi/server.go
```diff
@@ -254,12 +254,22 @@ func (server *Server) ComposeStatus(w http.ResponseWriter, r *http.Request, id s
...
+	}
+
 	response := ComposeStatus{
 		ImageStatus: ImageStatus{
 			Status: composeStatusFromJobStatus(status, &result),
 			UploadStatus: &UploadStatus{
-				Status: result.UploadStatus,
-				Type:   "aws",
+				Status:        result.UploadStatus,
+				Type:          "aws",
```
thozza, line 271: Keeping `"aws"` value hard-coded here is not beneficial in my opinion, especially once we introduce additional targets. The type of the target is now part of the `TargetResults` items as their `Name` property. Therefor this information is redundant. This is related to my other comment I left in the `internal/cloudapi/openapi.yml` file.

test/cases/api.sh
```diff
@@ -194,11 +194,14 @@ do
   COMPOSE_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.status')
   UPLOAD_STATUS=$(echo "$OUTPUT" | jq -r '.image_status.upload_status.status')
   UPLOAD_TYPE=$(echo "$OUTPUT" | jq -r '.image_status.upload_status.type')
-
+
```
thozza, line 197: unnecessary space added

internal/worker/json.go
```diff
@@ -21,10 +21,22 @@ type OSBuildJob struct {
...
+	UploadStatus  string                 `json:"upload_status"`
+}
+
+type TargetResult struct {
+	Type string      `json:"type"` // aws, gcp, azure, ....
+	Data interface{} `json:"data"` //cast this to a specific result depending on the type
+}
+
+/* if targetResult.Type == "aws" { targetResult.Data.(TargetAWSResult).Ami } */
+type TargetAWSResult struct {
```
thozza, line 37: I'm a bit confused, but it seems that the `TargetResult` and `TargetAWSResult` defined here are actually not used anywhere. Instead, for internal communication with worker, the structure from `target` package is used and for cloudapi the one generated by `openapi`.

thozza (commented):
Again, nice work! I can't wait for this to land, so that it can be used also got GCP target. I added a few comments below....

internal/worker/json.go
Gundersanne: Indeed, I think these can just be removed. I think it's just noise from my initial faffing about https://github.com/Gundersanne/osbuild-composer/commit/2e0475b061f996de9995814952b6cf1b649e59bd

Gundersanne (commented):
> > ```
> > type OSBuildJobResult struct {
> > 	Success       bool            `json:"success"`
> > 	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
> > 	TargetResults []TargetResult  `json:"target_results,omitempty"`
> > 	TargetErrors  []string        `json:"target_errors,omitempty"`
> > 	UploadStatus  string          `json:"upload_status"`
> > }
> > ```
> 
> @Gundersanne I wonder whether it would not make more sense to have the `UploadStatus` and `TargetErrors` merged into the `TargetResults`. This way it would be easier to connect the specific requested target with any related errors, upload status and any target-specific results. With the two distinct lists, it may not be that easy to connect specific `TargetError` to a `TargetResult`.
> 
> My idea would be something like:
> 
> ```
> type TargetResult struct {
>         Type string `json:"type"`
>         Error string `json:"error"`
>         UploadStatus `json:"upload_status"`
>         Options TargetResultOptions `json:"options"`
> }
> 
> type TargetResultOptions interface {
> 	isTargetResultOptions()
> }
> 
> type TargetResultAWSOptions struct {
>         Ami string `json:"ami"`
>         Region string `json:"region"`
> }
> 
> func (TargetResultAWSOptions) isTargetResultOptions() {}
> ```
> 
> ```
> type OSBuildJobResult struct {
> 	Success       bool            `json:"success"`
> 	OSBuildOutput *osbuild.Result `json:"osbuild_output,omitempty"`
> 	TargetResults []TargetResult  `json:"target_results,omitempty"`
> }
> ```
> 
> My main reasoning is that if we want to keep the option to have multiple targets for a compose request in the future, we should ideally keep all the information related to specific target in the same data structure. This will make it easier to make sense out of the results once we will get more than one target results.
> 
> What do you think?


I agree! I was thinking about eliminating UploadStatus or integrating it into TargetResults somehow, but your proposed solution makes a lot more sense. It seems like TargetErrors isn't really used anywhere except in `internal/worker/server.go` so this shouldn't be too big of a change.

internal/cloudapi/openapi.yml
```diff
@@ -110,18 +110,45 @@ components:
...
         status:
           type: string
           enum: ['success', 'failure', 'pending', 'running']
         type:
           $ref: '#/components/schemas/UploadTypes'
+        target_results:
+          type: array
+          items:
+            $ref: '#/components/schemas/TargetResult'
     AWSUploadStatus:
```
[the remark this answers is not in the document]
Gundersanne: I would actually keep `AWSUploadStatus` in favour of the `TargetResult`. `TargetResult` as an object shouldn't be exposed via the cloudapi.

Imo a status should look like:
```
{
    status: "success",
    upload_status: {
        type: "aws",
        status: "success",
        target_results: [{
            ami: "ami-0219e0129",
            region: "us-east-1"
        }]
    }
}
```

internal/cloudapi/openapi.yml
```diff
@@ -110,18 +110,45 @@ components:
...
       type: object
       properties:
         ami_id:
           type: string
           example: 'ami-0c830793775595d4b'
+    TargetResult:
```
Gundersanne, line 130: I woudn't expose this. I think the `TargetResults` from the Job should be converted into an `UploadStatus`.

internal/cloudapi/openapi.yml
```diff
@@ -110,18 +110,45 @@ components:
...
+        - name
+        - options
+      properties:
+          name:
+            type: string
+            example: 'aws'
+          options:
+            oneOf:
+              - $ref: '#/components/schemas/AWSTargetResultOptions'
+    AWSTargetResultOptions:
```
Gundersanne, line 142: ditto, use the `UploadStatus` I'd say

internal/cloudapi/server.go
Gundersanne: The type needs to be extracted from the TargetResult into the UploadStatus exactly. But I think we need to make a decision first on if we're going to support multiple upload_requests per compose or not.

Either we support multiple upload_requests and then we need to have an array in UploadStatus. Or we don't and then this can be a single object.

internal/cloudapi/openapi.yml
Gundersanne: So in image-builder and the cloudapi we use `type` to specify the target in `upload_request`. It's indeed a bit confusing and was just decided earlier on. If we switch to using `name` here we need to change it in all 3 projects for `upload_requests`.

chloenayon:
Thanks @thozza @Gundersanne for your good points! Currently making changes

internal/cloudapi/server.go
thozza: My thought process with `UploadStatus` and `TargetResults` was that these are technically the same thing. The target is basically an upload + additional processing needed to be able to use the image for VM instances. So the overall upload status does not makes so much sense, because with multiple targets you would be interested of the result of each one of them (basically the upload status). The same goes for errors and additional target-specific results (AMI ID, etc...).

Based on the discussion I had so far about this part of composer, I think that we still want to keep the ability to have multiple targets, even if not really allowed to be requested via cloudapi for now.

Since `ComposeRequest` -> `ImageRequest`'s `upload_requests` property is an array of `UploadRequest` I would propose to make the `ImageStatus` structure in the Compose status response symmetric. I would propose to keep only the `UploadStatus` as you mentioned and not expose the `TargetResults`. Also because the naming seems to be more of an internal thing, than anything we use in the cloudapi. In addition I would make the `upload_status` property of `ComposeStatus` an array. It should contain an item for each `UploadRequest`.

chloenayon:
Made fixes addressing above comments

test/cases/api.sh
```diff
@@ -198,7 +198,10 @@ do
   if [[ "$COMPOSE_STATUS" != "pending" && "$COMPOSE_STATUS" != "running" ]]; then
     test "$COMPOSE_STATUS" = "success"
     test "$UPLOAD_STATUS" = "success"
-    test "$UPLOAD_TYPE" = "aws"
+    if [[ "$UPLOAD_TYPE" == "aws" ]]; then
+      AMI_ID=$(echo "$OUTPUT" | jq -r '.image_status.upload_status.target_results[0].options.ami' | cut -c1-4)
```
teg, line 202: I think the `target_results[0]` needs to be dropped.

chloenayon: you're absolutely right, thank you!

thozza:
Hi @chloenayon. You may want to rebase on the latest `main` HEAD, because of the GCP support merged last Thu. The change touched quite a few places that are relevant also to your PR. In case you would have any questions related to the GCP support or would like some help with integrating the GCP-specific upload results, feel free to reach out to me. I'll be happy to help :wink:

chloenayon:
rebased on main, resolved some merge conflicts re: new GCP support

teg (approved):
This is coming along nicely :)

Gundersanne:
I think we can consider schutzbot as having passed:

https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1192/18/pipeline/304/  had el8 integration cancelled in libvirt.sh

https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1192/17/pipeline/304/ based on the same commit worked.

@teg want to rerun ci anyway?

teg (approved):
let's merge it! green enough

PR #1193, opened 2021-02-02
Add support for `org.osbuild.rhsm` osbuild stage + disable RHSM DNF plugins on RHEL qcow2 images

thozza:
This pull request includes:
  - Unit tests have been added.
  - Image test cases for affected images have been regenerated.
- [X] adequate documentation informing people about the change
  - Added new entry in `docs/news/unreleased/`

Add support to configure `org.osbuild.rhsm` osbuild stage. This stage
allows the configuration of Red Hat Subscription Manager (RHSM) related
components. Currently it is possible to configure only the enablement
status of RHSM DNF plugins.

Modify RHEL 8.3 and 8.4 KVM guest images definition to produce osbuild
manifest with `org.osbuild.rhsm` stage to disable both RHSM DNF plugins
(`product-id` and `subscription-manager`).

Add `/docs/news/unreleased/osbuild-rhsm-stage.md` with information about
the added support for `org.osbuild.rhsm` osbuild stage. Also note that RHEL
8.3 and 8.4 qcow2 image definitions are updated to disable RHSM DNF
plugins by default.

Enhance `tools/image-info` tool to add RHSM-specific section to its
output in case RHSM DNF plugins configuration exists in the tree.

Regenerate all RHEL image test cases affected by the patch set.

### DIFF
--- a/docs/news/unreleased/osbuild-rhsm-stage.md
+++ b/docs/news/unreleased/osbuild-rhsm-stage.md
@@ -1,0 +1,6 @@
+# Add support for `org.osbuild.rhsm` osbuild stage
+
+Add support for `org.osbuild.rhsm` osbuild stage. This stage is available in
+osbuild since version 24. The stage currently allows only configuring the
+enablement status of two RHSM DNF plugins, specifically of `product-id` and
+`subscription-manager` DNF plugins.
--- a/docs/news/unreleased/osbuild-rhsm-stage.md
+++ b/docs/news/unreleased/osbuild-rhsm-stage.md
@@ -4,3 +4,10 @@ Add support for `org.osbuild.rhsm` osbuild stage. This stage is available in
 osbuild since version 24. The stage currently allows only configuring the
 enablement status of two RHSM DNF plugins, specifically of `product-id` and
 `subscription-manager` DNF plugins.
+
+# RHEL 8.3 & 8.4: Disable all RHSM DNF plugins on qcow2 image
+
+Disable both available RHSM DNF plugins (`product-id` and
+`subscription-manager`) on rhel-8 and rhel-84 qcow2 images. The reason for
+disabling these DNF plugins is to make the produced images consistent in this
+regard, with what had been previously produced by the imagefactory.
--- a/internal/distro/rhel8/distro.go
+++ b/internal/distro/rhel8/distro.go
@@ -324,6 +324,20 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 			WaitForNetwork: true,
 		},
 		))
+	} else {
+		// RHSM DNF plugins should be by default disabled on RHEL Guest KVM images
+		if t.Name() == "qcow2" {
+			p.AddStage(osbuild.NewRHSMStage(&osbuild.RHSMStageOptions{
+				DnfPlugins: &osbuild.RHSMStageOptionsDnfPlugins{
+					ProductID: &osbuild.RHSMStageOptionsDnfPlugin{
+						Enabled: false,
+					},
+					SubscriptionManager: &osbuild.RHSMStageOptionsDnfPlugin{
+						Enabled: false,
+					},
+				},
+			}))
+		}
 	}
 
 	p.Assembler = t.assembler(t.arch.uefi, options, t.arch)
@@ -1002,6 +1016,7 @@ func New() distro.Distro {
 			"glibc",
 			"policycoreutils",
 			"python36",
+			"python3-iniparse", // dependency of org.osbuild.rhsm stage
 			"qemu-img",
 			"selinux-policy-targeted",
 			"systemd",
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -363,6 +363,20 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 			WaitForNetwork: true,
 		},
 		))
+	} else {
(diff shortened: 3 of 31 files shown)


teg (commented):
Nice to see this all come together! I wonder about the best way to verify that the images actually contain the right settings. Maybe simply extending `image-info` to expose this information and then the manifest tests would verify it?

docs/news/unreleased/osbuild-rhsm-stage.md
```diff
@@ -4,3 +4,10 @@ Add support for `org.osbuild.rhsm` osbuild stage. This stage is available in
 osbuild since version 24. The stage currently allows only configuring the
 enablement status of two RHSM DNF plugins, specifically of `product-id` and
 `subscription-manager` DNF plugins.
+
+# RHEL 8.3 & 8.4: Disable all RHSM DNF plugins on qcow2 image
+
+Disable both available RHSM DNF plugins (`product-id` and
+`subscription-manager`) on rhel-8 and rhel-84 qcow2` images. The reason for
```
ondrejbudai, line 11: There's an extra `.

ondrejbudai (commented):
This is looking very good. I agree with Tom, extending `image-info` seems very reasonable so we're able to test this change.

thozza:
@teg Yeah, I have been thinking about this as well... The `org.osbuild.rhsm` stage and its functionality is tested as part od osbuild's CI (https://github.com/osbuild/osbuild/pull/555/commits/ec2d3e396e82fe18b154c02877606658a7417f7e). Then the fact that the stage is added to the relevant rhel-8 rhel-84 manifest is tested as part of this PR.
Example snippet from `test/data/manifests/rhel_84-x86_64-qcow2-boot.json`
```
@@ -3486,6 +3492,19 @@
           "options": {
             "file_contexts": "etc/selinux/targeted/contexts/files/file_contexts"
           }
+        },
+        {
+          "name": "org.osbuild.rhsm",
+          "options": {
+            "dnf-plugins": {
+              "product-id": {
+                "enabled": false
+              },
+              "subscription-manager": {
+                "enabled": false
+              }
+            }
+          }
         }
       ],
       "assembler": {
```
And the fact that the RHSM DNF plugin configurations have been modified is captured as part of `rpm-verify` output:
```
@@ -10180,6 +10217,8 @@
       "changed": {
         "/boot/efi/EFI/redhat/grubx64.efi": ".......T.",
         "/etc/chrony.conf": "S.5....T.",
+        "/etc/dnf/plugins/product-id.conf": "..5....T.",
+        "/etc/dnf/plugins/subscription-manager.conf": "..5....T.",
         "/etc/machine-id": ".M.......",
         "/proc": ".M.......",
         "/run/cockpit": ".M.......",
```

Nevertheless I agree that all of this information is not really good enough for verifying what is the actual content of the plugin configurations and how were they modified.

So `image-info` could be definitely extended to list additional information specific to the RHSM stage. The question is how... I'm not really keen on having a generic section listing enablement status of all DNF plugins. The reason is that each plugin can handle its configuration in a different way and although plugin configurations are usually in INI format, it is not really mandatory. For this reason it would be hard to implement something really generic when we don't have any constrains on the plugin configuration format. I would suggest two options:

1. Have a RHSM specific section in the `image-info` output and for now, include only the enablement status of the RHSM DNF plugins. List the information only if the respective plugin configuration file exists.
2. Have a generic DNF plugins section in the `image-info` output, but check specifically only a defined list of plugin configurations. For now, the list would contain only RHSM DNF plugins. The information listed could be either the enablement status of the plugin, or all the listed options in the configuration, I don't have a strong preference... And also list the information only if the respective plugin configuration file exists.

What do you think?

docs/news/unreleased/osbuild-rhsm-stage.md
thozza: Nice catch... It took me a few seconds to spot it even now :smile:

teg:
Ah, you gave this more thought than me @thozza! I think the argument could be made that this is already good enough (as the contents of the modified files is presumably tested in osbuild). However, if we were to improve the image-info tests (still nice, but I wouldn't say a requirement, so up to you). I would try to mirror the osbuild stage in functionality. I.e., collect exactly the info it is able to set.

ondrejbudai:
Blocked by https://bodhi.fedoraproject.org/updates/FEDORA-2021-1f28f28ea7.

thozza:
So osbuild v24 should be already in F32 and F33 stable, as well as in el8.4 nightly. I'll wait for the PR #1188 to get merged and rebase my PR after that...

gicmo:
#1188 is in, osbuild 24 in stable, should not be blocked anymore.

thozza:
I enhanced the `image-info` tool to include RHSM-specific section in its output in case the RHSM DNF plugins configurations are in the image tree. I'm regenerating image test cases now and will update the PR once this is done. This should be (hopefully) the last change needed to get this merged...

thozza:
The PR has been rebased and all RHEL image test cases have been regenerated. :partying_face:  Once the Schutzbot CI passes, this PR should be good to go (unless there is any additional feedback on the code changes).

teg (approved):
Amazing! Thanks for your hard work on this :)

ondrejbudai (approved):
Awesome! :)

gicmo (approved):
👏👏

thozza:
All tests in Schutzbot passed, except for "F33 aarch64 Base" test. The failure was caused by an inconsistent RPM mirror, which returned 404 while fetching RPM from it. Since this change set did not touch any F33 images or their definitions, I propose to ignore this failure and merge the PR.

PR #1195, opened 2021-02-03
Azure: test on V2 Hyper-V generation

atodorov:
Related: rhbz#1896264

Update - a few more questions raised here:

1) Should we drop testing on Gen1 hypervisors or not ? This is being tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1896264

2) How the heck v2 image (hypervisor) works with RHEL 8.3 that is bios-only? (we're not sure if hypervisor can auto-detect which boot mode to use)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -151,6 +151,11 @@ Integration:
           - aws/rhel-8-x86_64
           - aws/rhel-8.4-x86_64
         INTERNAL_NETWORK: ["true"]
+      - SCRIPT:
+         - azure_hyperv_gen2.sh
+        RUNNER:
+          - aws/rhel-8.4-x86_64
+        INTERNAL_NETWORK: ["true"]
       - <<: *INTEGRATION_TESTS
         RUNNER:
           - aws/rhel-8-x86_64
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -5,6 +5,7 @@ source /etc/os-release
 DISTRO_CODE="${DISTRO_CODE:-${ID}_${VERSION_ID//./}}"
 BRANCH_NAME="${BRANCH_NAME:-${CI_COMMIT_BRANCH}}"
 BUILD_ID="${BUILD_ID:-${CI_BUILD_ID}}"
+HYPER_V_GEN="${HYPER_V_GEN:-V1}"
 
 # Colorful output.
 function greenprint {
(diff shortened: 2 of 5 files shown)


atodorov:
@ondrejbudai this will need backport downstream.

ondrejbudai:
I'm not sure if we want to test only the v2 gen. Shouldn't we test both?

atodorov:
> I'm not sure if we want to test only the v2 gen. Shouldn't we test both?

Do we default to UEFI or not ? If we do then V2 IMO is fine. 

OTOH what's the recommendation from Microsoft? I'd imagine they want customers always running on their latest version of the hypervisor.

atodorov:
> OTOH what's the recommendation from Microsoft? 

From https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/plan/should-i-create-a-generation-1-or-2-virtual-machine-in-hyper-v

```
We recommend that you create a generation 2 virtual machine to take advantage of features like Secure Boot unless one of the following statements is true:

- The VHD you want to boot from is not UEFI-compatible.                           <- we have UEFI boot
- Generation 2 doesn't support the operating system you want to run on the virtual machine. <- RHEL is supported on Gen2
- Generation 2 doesn't support the boot method you want to use.  <- not sure what this is but I think it ties back to UEFI
```

From what I can see we can safely default to Gen2 hypervisors only!

atodorov:
Digging through our latest test suite I see the following:

1) `test/data/azure/deployment-template.json` uses hyperVGeneration V1 (ignoring the change in this PR) which is used by osbuild-image-tests, used by image_tests.sh, in turn executed by the osbuild repository (recently disabled for the osbuild-composer repository).

2) test/data/azure/main.tf is used in integration testing and will also deploy & boot a VM. I believe this is V2 by default but I can't find any information if it is possible to specify that explicitly, see https://discuss.hashicorp.com/t/how-do-i-explicitly-specify-gen1-or-gen2-for-azurerm-linux-virtual-machine/24272. If we can figure that out it will be very straight forward to cover the testing requirements for rhbz#1896264.

ondrejbudai:
I see that the Azure v2 test failed in the first run: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1339981146

Let's wait for the second attempt to finish.

atodorov:
> I see that the Azure v2 test failed in the first run: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1339981146
> 
> Let's wait for the second attempt to finish.

I've restarted this job couple of times today but it still fails in the same way. I'm not convinced the failure is related to the current changes though b/c the 8.4 job reported pass.

ondrejbudai:
The test is actually running on an RHEL 8.3 instance so any images are built using the 8.3 definitions that are not UEFI-compatible. Could this be the issue?

atodorov:
> The test is actually running on an RHEL 8.3 instance so any images are built using the 8.3 definitions that are not UEFI-compatible. Could this be the issue?

well it could be. I will disable it and we can re-enable once we move forward.

atodorov:
Test job reporting PASS - https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1344976801

ondrejbudai (approved):
Looks good, waiting for CI to finish all jobs.

atodorov:
All green, merging.

PR #1198, opened 2021-02-03
cloudapi: Packages endpoint

Gundersanne:
- [ ] Probably should also add the ability to specify packages for a compose to the cloudapi in this PR

### DIFF
--- a/internal/cloudapi/openapi.yml
+++ b/internal/cloudapi/openapi.yml
@@ -60,6 +60,24 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ComposeResult'
+  /packages:
+    post:
+      summary: Get packages
+      description: Get a list of packages for a given distribution and repositories
+      operationId: packages
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PackagesRequest'
+      responses:
+        '200':
+          description: List of packages
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PackagesResponse'
 
 components:
   schemas:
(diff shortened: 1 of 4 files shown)


internal/cloudapi/openapi.yml
```diff
@@ -60,6 +60,24 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ComposeResult'
+  /packages:
+    post:
```
teg, line 64: Should this be a `GET`?

internal/cloudapi/openapi.yml
```diff
@@ -60,6 +60,24 @@ paths:
...
+      summary: Get packages
+      description: Get a list of packages for a given distribution and repositories
```
teg, line 66: Instead of getting the full list, should we make this a search, and limit the number of replies (possibly as an argument)? Returning 50000 packages (on fedora) might be a bit sluggish.

teg (commented):
Nice start! Would be fine to make the package requests separate or part of the same. Up to you.

internal/cloudapi/openapi.yml
Gundersanne: Doesn't allow a body and the repository descriptions are quite hard to pass as path parameters, but if you'd prefer that it would be more idiomatic I guess.

Gundersanne:
Moving this over to image-builder, decided that as much of the ui-driving should happen in image-builder, not in composer.

Furthermore the packages endpoint shouldn't return all the available packages, rather it should let you search for a package. Because the list of repositories  are stored within image-builder, it's probably not too bad to make a GET request out of this now, which is a lot more idiomatic as well.

PR #1200, opened 2021-02-03
image-info: do not include inputhash in the report

ondrejbudai:
The algorithm from calculating changed in osbuild 24, thus breaking the
testsuite. As the the inputhash is merely a implementation detail, there's
no need to test it in the image tests. Nevertheless, the inputhash is also
tested in the osbuild's testsuite.

This pull request includes:

This is a minor change of the test suite, both items are not needed imho.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -499,9 +499,6 @@ def append_ostree_repo(report, repo):
     resolved = {r: ostree("rev-parse", r).stdout.strip() for r in refs}
     commit = resolved[refs[0]]
 
-    refs = {r: {"inputhash": ostree("show", "--print-metadata-key=rpmostree.inputhash", resolved[r]).stdout.strip("'\n")} for r in refs}
-    report["ostree"]["refs"] = refs
-
     with tempfile.TemporaryDirectory(dir="/var/tmp") as tmpdir:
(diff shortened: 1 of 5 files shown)


tools/image-info
```diff
@@ -499,9 +499,6 @@ def append_ostree_repo(report, repo):
     resolved = {r: ostree("rev-parse", r).stdout.strip() for r in refs}
     commit = resolved[refs[0]]
 
-    refs = {r: {"inputhash": ostree("show", "--print-metadata-key=rpmostree.inputhash", resolved[r]).stdout.strip("'\n")} for r in refs}
-    report["ostree"]["refs"] = refs
-
```
gicmo, line 504: Sad to see this go, be neat to (in the future) have a `--complete` flag or so for `image-info` to get info that are not meant for tests, or something like this.

gicmo (approved):
🙆

tools/image-info
ondrejbudai: Yeah, this would be nice I think. :)

ondrejbudai:
I had to force-push a fix. Editing JSON by hand is hard and I forgot to remove a set of parentheses.

gicmo (approved):
Even better!

PR #1208, opened 2021-02-09
cloudapi: add support for additional packages

teg:
Optionally allow a pacakge set to be included in the compose request.

The specified packages are added to the base packages before
depsolving. As the base packages differ between the image types
the package customizations may have different results on the different
images part of the compose request.

This pull request includes:

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -77,6 +77,20 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var bp = blueprint.Blueprint{}
+	err = bp.Initialize()
+	if err != nil {
+		http.Error(w, "Unable to initialize blueprint", http.StatusInternalServerError)
+		return
+	}
+	if request.Customizations != nil && request.Customizations.Packages != nil {
+		for _, p := range *request.Customizations.Packages {
+			bp.Packages = append(bp.Packages, blueprint.Package{
+				Name: p,
+			})
+		}
+	}
+
 	type imageRequest struct {
 		manifest distro.Manifest
 		arch     string
@@ -116,13 +130,6 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, "Must specify baseurl, mirrorlist, or metalink", http.StatusBadRequest)
 				return
 			}
-		}
-
-		var bp = blueprint.Blueprint{}
-		err = bp.Initialize()
-		if err != nil {
-			http.Error(w, "Unable to initialize blueprint", http.StatusInternalServerError)
-			return
 		}
 
 		packageSpecs, excludePackageSpecs := imageType.Packages(bp)
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -90,6 +90,11 @@ cat > "$REQUEST_FILE" << EOF
       "architecture": "$ARCH",
       "image_type": "ami",
       "repositories": $(jq ".\"$ARCH\"" /usr/share/tests/osbuild-composer/repositories/"$DISTRO".json),
+      "customizations": {
+        "packages": [
+          "postgres"
+        ]
+      },
       "upload_requests": [
         {
           "type": "aws",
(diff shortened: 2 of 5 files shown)


larskarlitski (commented):
Lovely :heart:! I'd say this is ok with a test in api.sh for now, but would be cool if we could check that the package ended up on the image as well at some point.

internal/cloudapi/server.go
```diff
@@ -77,6 +77,20 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
...
+	var bp = blueprint.Blueprint{}
+	err = bp.Initialize()
+	if err != nil {
+		http.Error(w, "Unable to initialize blueprint", http.StatusInternalServerError)
+		return
+	}
+	if request.Customizations != nil && request.Customizations.Packages != nil {
+		for _, p := range *request.Customizations.Packages {
+			bp.Packages = append(bp.Packages, blueprint.Package{
+				Name: p,
```
jkozol, line 89: @Gundersanne so, to make sure I understand this properly, when we create our image we will pass a list of package names as a customization. The list of names would look something like `['0ad', 'curl', 'openssh']` correct? In the future will there be a desire to support multiple package versions?

jkozol (approved):

test/cases/api.sh
```diff
@@ -90,6 +90,11 @@ cat > "$REQUEST_FILE" << EOF
       "architecture": "$ARCH",
       "image_type": "ami",
       "repositories": $(jq ".\"$ARCH\"" /usr/share/tests/osbuild-composer/repositories/"$DISTRO".json),
+      "customizations": {
+        "packages": [
+          "postgres",
```
Gundersanne, line 95: Hm I guess there's no way to verify this if it actually gets installed since we don't boot the image. Maybe something to do in future, we would check the subscriptions that way as well as those are currently untested in the same way.

Gundersanne (approved):
Nice, lgtm :)

test/cases/api.sh
teg: Yeah, was thinking the same. We definitely should boot the image, and check all these things. I guess for a follow-up :)

Gundersanne: It seems using  [`aws ec2-instance-connect send-ssh-public-key`](https://docs.aws.amazon.com/cli/latest/reference/ec2-instance-connect/send-ssh-public-key.html) we could inject a public ssh key after creating the instance. Doesn't need to be a part of this PR imo but worth doing in future I think.

internal/cloudapi/server.go
teg: Correct!
Of desires of the future I know not much. However, not supporting package versions was intentional from my side. The thinking was that we mostly don't have different versions of the same package in our repos (so selecting anything but the most recent one will just fail), and if we do, mixing and matching versions surely have not been tested well. So let's start off with only supporting the most recent versions, and if we want more granularity in the future we should discuss exactly what is wanted / what that means.

teg:
Added docs, just need schutzbot again :)

Gundersanne (dismissed):
still looks good :)

teg:
The failing `Checks` test is due to the new github action PR trigger. This is being reworked in a separate PR, so I'm fine with merging with that being red (assuming schutzbot approves).

ondrejbudai (changes requested):
Schutzbot sadly didn't approve this PR:

```
curl --silent --show-error --cacert /etc/osbuild-composer/ca-crt.pem --key /etc/osbuild-composer/client-key.pem --cert /etc/osbuild-composer/client-crt.pem --header 'Content-Type: application/json' --request POST --data @/tmp/tmp.T63Ecb0GvI/request.json https://localhost/api/composer/v1/compose
+ OUTPUT='Could not parse JSON body'
++ echo 'Could not parse JSON body'
++ jq -r .id
parse error: Invalid numeric literal at line 1, column 6
```

test/cases/api.sh
Gundersanne, line 95: Ah I think this final `,` might make the json of the request invalid :)

teg: Oops, yeah this is invalid.

teg:
Thanks for the fix @Gundersanne!

ondrejbudai (approved):
Awesome!

PR #1209, opened 2021-02-09
⚡ Trigger schutzbot using github actions

major:


teg (approved):

teg:
@major @ondrejbudai I got a bit of a talking to for approving a PR without a commit message 😱 And rightly so! We should try to make sure to explain our commits, even just for infrastructure type things, so that others can better understand our thinking in case something goes pear-shaped and they want to understand. This made sense to me as I followed the chat, but obviously not everyone does that all the time, so should keep that in mind.

PR #1212, opened 2021-02-10
Send webhooks without credentials

major:
We can now send webhook data to an SQS queue at AWS without signing the
request with credentials. This allows us to trigger Schutzbot from
forks and from branches on the main repository.


major:
@teg Looks like schutzbot was triggered from a fork!

teg (approved):
Nice! 🥳

teg:
No need to wait for schutzbot, we already verified that this triggers it correctyl, and without this PR it doesn't work at all.

PR #1214, opened 2021-02-11
weldr: Remove underscores from FreezeHandler error

bcl:
This pull request includes:

I didn't include testing or documentation because it's a string meant to be read by the user when trying to freeze a blueprint that doesn't exist.


ondrejbudai (approved):
looks_good_to_me

I don't think we guarantee the stability of the error messages so this is imho OK. Maybe we should mention this fact in the commit message?

ondrejbudai (approved):
Thanks! :)

PR #1218, opened 2021-02-11
distro/rhel84: remove dbxtool

jkozol:
dbxtool is not available on s390x or ppc64le. So, it cannot be included for now without breaking image builds for these architectures.

It seems that dbxtool will be installed without the explicit declaration since the regenerating the test manifests for x86_64 did not remove it. I am unsure why it was missing in the image-info for the nightly I was comparing the package list for.


teg (approved):

ondrejbudai (approved):
Thanks! :)

PR #1224, opened 2021-02-12
test: Fix the compose request in the api test case

Gundersanne:
Customizations should be alongside image_requests, not inside of it.

This pull request includes:


ondrejbudai (approved):
Nice catch!

Gundersanne:
Definitely wait for the tests, as `postgres` doesn't exist as a package, `postgresql` does, I hope it's called the same in rhel, otherwise we need a different package :)

teg (approved):
👍

PR #1225, opened 2021-02-13
Investigate how often the openstack test fails

ondrejbudai:
do not merge this ever


msehnout:
50 successful attempts? I find it hard to believe :D

ondrejbudai:
They added more hardware to the PSI this weekend, maybe this issue might be gone without any changes on our side...

teg:
Thanks for looking into this. How about we make the openstack not actually fail CI, but instead just count how often it fails/succeeds in our regular runs, and if we see the stats improve we can reenable it?

ondrejbudai:
See #1254

PR #1229, opened 2021-02-15
Schutzbot: Start building on aarch64 for RHEL/CentOS

atodorov:
This pull request includes:

### DIFF
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -138,6 +138,20 @@ pipeline {
                         }
                     }
                 }
+                stage('EL8 aarch64') {
+                    agent { label "rhel8cloudbase && aarch64 && aws" }
+                    environment {
+                        AWS_CREDS = credentials('aws-credentials-osbuildci')
+                        AWS_IMAGE_TEST_CREDS = credentials('aws-credentials-osbuild-image-test')
+                        RHN_REGISTRATION_SCRIPT = credentials('rhn-register-script-production')
+                    }
+                    steps {
+                        sh "schutzbot/ci_details.sh"
+                        retry(3) {
+                            sh "schutzbot/mockbuild.sh"
+                        }
+                    }
+                }
                 stage('EL8.4') {
                     agent { label "rhel84cloudbase && x86_64 && psi" }
                     environment {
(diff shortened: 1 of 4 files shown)


atodorov:
@major IDK if the slave systems will be provisioned automatically or if you need to make some adjustments to our AWS account. Let's see how this goes.

teg (commented):
Not sure we have the required workers set up (at least not in PSI), but if we can, this would be good to have!

ondrejbudai:
AFAIK, PSI doesn't have support for aarch64. It's possible though to test CS on AWS, I just need to add an image definition to the Jenkins settings. I will ping you when it's done.

atodorov:
> AFAIK, PSI doesn't have support for aarch64. 

I think we can use an AWS host for RHEL + aarch64 and limit the types of testing we do there / targets we build against!


> It's possible though to test CS on AWS, I just need to add an image definition to the Jenkins settings. I will ping you when it's done.

I can probably do that myself, let me know where to look.

ondrejbudai:
> I think we can use an AWS host for RHEL + aarch64 and limit the types of testing we do there / targets we build against!

However, we currently cannot access the internal repos (and thus RHEL 8.4) from AWS.

atodorov:
> 
> However, we currently cannot access the internal repos (and thus RHEL 8.4) from AWS.

Yes but we can only build & test on released RHEL 8 like we already do. This is still better than not testing aarch64 at all.

ondrejbudai:
Ah, yeah, that's definitely possible. I think many people are interested in enabling testing for more distros - I will try to prepare a guide. :)

ondrejbudai:
I added these AMI instance types to Jenkins config:

```
rhel8cloudbase && aarch64 && aws
cs8cloudbase && aarch64 && aws
```

As I said, I don't think that can currently test RHEL 8.4 on aarch64 but at least stable RHEL and Centos Stream should be possible.

ondrejbudai:
Hmm, the tests seem to be broken right now. Alex, can you please take a look?

schutzbot/Jenkinsfile
```diff
@@ -675,36 +675,6 @@ pipeline {
                         }
                     }
                 }
-                stage('EL8 aarch64 Image') {
```
teg: Should this be part of this patch?

atodorov: Nope, squashed into the previous commit.

atodorov:
> Hmm, the tests seem to be broken right now. Alex, can you please take a look?

There's a failure trying to download a nightly repo with dnf. The reason this happens is because a file check 
`-f rhel8nightly.repo` succeeds, which happens because a previous step manages to somehow unstash files with the same name: `Restore files previously stashed` which are not supposed to exist IMO. Still investigating.

atodorov:
Merged as #1287

PR #1234, opened 2021-02-17
Enable nfs convert

jkozol:
nfs-convert service is not currently being enabled by the redhat-release package but it should be enabled. For now, osbuild-composer will explicitly enable it for RHEL 8.4 qcow2 images.

This pull request includes:


test/data/manifests/rhel_84-x86_64-qcow2-customize.json
```diff
@@ -12,6 +12,11 @@
...
+      },
+      {
+        "name": "rt",
```
ondrejbudai, line 17: This looks... unrelated. I think it's because of @achilleas-k's changes but I have no idea why this wasn't added before.

achilleas-k: Took me a while to figure out what was happening, thought a package was being added.

This is fine, we just added a new repo source for the rt kernel, but I neglected to generate all the existing test cases, so they weren't added.

ondrejbudai: OK, thanks! :)

docs/news/unreleased/nfs-convert.md
```diff
@@ -0,0 +1,5 @@
+# RHEL 8.4: enable nfs-convert.service
```
ondrejbudai, line 1: I wonder if we should mention that this change also affects CentOS Stream... Not sure...

jkozol: Ya I will add that in. I think it makes sense to mention.

test/data/manifests/rhel_84-x86_64-qcow2-customize.json
jkozol: What is the recommended way to regenerate the cases now? There are quite a few arches and distros now so it doesn't seem practical to regenerate the cases locally.

achilleas-k: There's a new script, `tools/test-case-generators/generate-all-test-cases`, that can generate all the test cases for all the arches inside VMs, but like you said, it's not very practical and takes hours.

jkozol: oh gosh haha. Let's see how this goes...

jkozol:
Closing this as there is no need for us to explicitly enable nfs-convert. We can just default to the packages behaviour on install.

PR #1240, opened 2021-02-19
tools: move to new utility-container builds

dvdhrm:
Use the new tags of the `osbuild/containers` repository. The
infrastructure was simplified a lot and the new tags are much easier to
handle without any conflicts when building other images.

Note that this change uses the `latest` tag of all images. As an
alternative, we can also switch to the `latest-<date>` tags, which would
use a pinned immutable image tagged by the given date. However, these
tags are burried somewhere deep down in the ./tools/ directory and are
easy to forget upgrading. So for now use `latest` until we have better
synchronization infrastructure in place.

However, if these images ever break and we want to get back stable CI
behavior, we can always switch back to older builds by picking those
immutable tags instead of `latest`.

### DIFF
--- a/tools/run-koji-container.sh
+++ b/tools/run-koji-container.sh
@@ -57,13 +57,13 @@ koji_start() {
     -e POSTGRES_USER=koji \
     -e POSTGRES_PASSWORD=kojipass \
     -e POSTGRES_DB=koji \
-    quay.io/osbuild/postgres:v1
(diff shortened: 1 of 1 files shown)


teg (approved):

ondrejbudai (approved):
cool!

ondrejbudai:
Both failures are unrelated to the koji test that uses these containers.

PR #1241, opened 2021-02-19
rhel84/grub2: set saved_entry

teg:
Explicitly set the kernel to boot into.

This hooks up the recent changes @gicmo made in osbuild to fix a rhel8.4 regression. No tests yet.

This pull request includes:

### DIFF
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -155,11 +155,21 @@ func (c *Customizations) GetGroups() []GroupCustomization {
 }
 
 func (c *Customizations) GetKernel() *KernelCustomization {
-	if c == nil {
-		return nil
+	var name string
+	var append string
+	if c != nil && c.Kernel != nil {
+		name = c.Kernel.Name
+		append = c.Kernel.Append
 	}
 
-	return c.Kernel
+	if name == "" {
+		name = "kernel"
+	}
+
+	return &KernelCustomization{
+		Name:   name,
+		Append: append,
+	}
 }
 
 func (c *Customizations) GetFirewall() *FirewallCustomization {
--- a/internal/distro/fedora32/distro.go
+++ b/internal/distro/fedora32/distro.go
@@ -426,7 +426,7 @@ func (t *imageType) fsTabStageOptions(uefi bool) *osbuild.FSTabStageOptions {
 func (t *imageType) grub2StageOptions(kernelOptions string, kernel *blueprint.KernelCustomization, uefi bool) *osbuild.GRUB2StageOptions {
 	id := uuid.MustParse("76a22bf4-f153-4541-b6c7-0332c0dfaeac")
 
-	if kernel != nil {
+	if kernel != nil && kernel.Append != "" {
 		kernelOptions += " " + kernel.Append
 	}
 
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -295,7 +295,7 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 
 	if t.bootable {
 		if t.arch.Name() != "s390x" {
-			p.AddStage(osbuild.NewGRUB2Stage(t.grub2StageOptions(pt, t.kernelOptions, c.GetKernel(), t.arch.uefi, t.arch.legacy)))
+			p.AddStage(osbuild.NewGRUB2Stage(t.grub2StageOptions(pt, t.kernelOptions, c.GetKernel(), packageSpecs, t.arch.uefi, t.arch.legacy)))
 		}
 	}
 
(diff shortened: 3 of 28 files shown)


larskarlitski (approved):
nice work 👍

internal/distro/rhel84/distro.go
```diff
@@ -547,11 +543,22 @@ func (t *imageType) grub2StageOptions(pt *disk.PartitionTable, kernelOptions str
 		legacy = t.arch.legacy
 	}
 
+	var kernelEntry string
+	if kernel != nil {
+		kernelOptions += " " + kernel.Append
+		for _, pkg := range packages {
+			if pkg.Name == kernel.Name {
+				kernelEntry = "ffffffffffffffffffffffffffffffff-" + pkg.Version + "-" + pkg.Release + "." + pkg.Arch + ".conf"
```
gicmo, line 551: I think it is without the `.conf` at the end:
```
at /boot/grub2/grubenv 
# GRUB Environment Block
kernelopts=root=UUID=0194fdc2-fa2f-4cc0-81d3-ff12045b73c8 console=tty0 console=ttyS0,115200n8 no_timer_check net.ifnames=0 crashkernel=auto
saved_entry=ffffffffffffffffffffffffffffffff-4.18.0-275.el8.x86_64
```

teg: Thanks!

ondrejbudai:
I think we need to bump the osbuild dependency in both the spec file and README.

teg:
> I think we need to bump the osbuild dependency in both the spec file and README. Also what about Fedora 32? Doesn't it need a pinned dependency?

Good points! Fixed.

internal/blueprint/customizations.go
```diff
@@ -155,11 +155,21 @@ func (c *Customizations) GetGroups() []GroupCustomization {
...
+	var name string
+	var append string
+	if c != nil && c.Kernel != nil {
+		name = c.Kernel.Name
+		append = c.Kernel.Append
 	}
 
-	return c.Kernel
+	if name == "" {
+		name = "kernel"
```
gicmo, line 166: Just a thing to remember, if we were to use that from the sysconfig stage in the future, the default there for fedora is `kernel-core`:

```
# DEFAULTKERNEL specifies the default kernel package type
DEFAULTKERNEL=kernel-core
```

internal/distro/fedora32/distro.go
```diff
@@ -426,7 +426,7 @@ func (t *imageType) fsTabStageOptions(uefi bool) *osbuild.FSTabStageOptions {
 func (t *imageType) grub2StageOptions(kernelOptions string, kernel *blueprint.KernelCustomization, uefi bool) *osbuild.GRUB2StageOptions {
 	id := uuid.MustParse("76a22bf4-f153-4541-b6c7-0332c0dfaeac")
 
-	if kernel != nil {
+	if kernel != nil && kernel.Append != "" {
```
gicmo, line 429: `kernel` can't e `nil` here anymore now, can it? But I guess it does not hurt to check anyway.

gicmo (approved):
:rescue_worker_helmet: Looks good to me!

internal/blueprint/customizations.go
teg: Ah, thank for pointing that out. I wonder if it matters / what difference it makes. It seems like `kernel` and `kernel-core` go together. Is this just another random difference between RHEL and Fedora?

internal/distro/fedora32/distro.go
teg: Yeah, can't be nil, so probably want to refactor the whole thing to not be a pointer at all. But kept it minimal for now.

internal/blueprint/customizations.go
gicmo: Good question, need to investigate this.

gicmo: I think it is a by-product of the auto-detection mechanism in anaconda: https://github.com/rhinstaller/anaconda/blob/6bbc4f12b1a3e61790ae3d742459714a82eb7c0b/pyanaconda/modules/storage/bootloader/utils.py#L149

gicmo: On Fedora it is auto-generated via anaconda, which detects the package by looking which package provides the actual kernel image (`kernel-core`, which is also true on RHEL). But on RHEL the `/etc/sysconfig/kernel` file for the guest images is explicitly written in the [kickstart](http://git.app.eng.bos.redhat.com/git/spin-kickstarts.git/tree/rhel8/rhel-8.4-kvm-x86_64.ks#n216) file and there set to `DEFAULTKERNEL=kernel`.

gicmo: Ok, I think this is only used by the [`new-kernel-pkg` script](https://github.com/rhboot/grubby/blob/99d10a31cee89eb45ac0743f484018d36500ce66/new-kernel-pkg#L846) and there both it checks for `$package` and `$package-core`, so both `kernel` and `kernel-core` will work. `new-kernel-pkg` is in `grubby-deprecated` and I think not used anymore on systems using BLS. Right @martinezjavier ?

martinezjavier: > . Right @martinezjavier ?

@gicmo yes, that's correct. The `DEFAULTKERNEL` option is `/etc/sysconfig/kernel` is not used anymore in neither Fedora nor RHEL8.

teg:
Only OpenStack boot test failed. This is a known problem and unrelated. This is ready to go in once the release has been tagged.

ondrejbudai (approved):

PR #1242, opened 2021-02-20
28

msehnout:
Release osbuild-composer version 28

This pull request includes:


NEWS.md
```diff
@@ -1,5 +1,37 @@
...
+  * OSBuild dependency now requires version 24 because of its new sysconfig and
+    RHSM stages.
+
+  * RHEL qcow2 images now come with disabled product-id and subscription-manager
+    plugins in dnf to make them consistent with images produced by imagefactory.
+
+  * OSTree `parent` can now be replaced with `url` pointing to a repository
+    that contains the parent commit.
+
+Contributions from: Achilleas Koutsou, Aleksandar Todorov, Alexander Todorov,
```
ondrejbudai, line 28: I guess we can mention Alex only once. :)

NEWS.md
```diff
@@ -1,5 +1,37 @@
...
+
+  * Cloudapi contains new /openapi.json and /version endpoints. The already
+    existing `POST /compose` endpoint now supports specification of additional
+    packages.
+
+  * Blueprints now support definition of alternative kernels, such as kernel-rt,
+    using the `customizations.kernel.name` key.
+
+  * Attempt to use the customization.kernel key in a blueprint for the OSTree
+    image type now results in error, because the customization was ignored.
```
achilleas-k, line 17:
```suggestion
  * Attempt to use the `customization.kernel.append` key in a blueprint for the OSTree
    image type now results in error, because the customization was ignored.
```

I realise this was unclear in my original description of the change, but it's actually the `append` key that's ignored.

achilleas-k (dismissed):
Looks good.  One comment for clarification.

NEWS.md
msehnout: hmm but which one? :-D

NEWS.md
msehnout: Thanks, I'll fix it.

NEWS.md
ondrejbudai: he uses Alexander on social networks.

teg (approved):
Thanks for this Martin! CI is flaky today, and is failing for CentOS Stream, which we don't release into yet, so that's ok. One test is still running, the new ostree test on F33, between this test passing on F32 and the almost identical old ostree test passing on F33, I'll declare that good enough.

PR #1231, opened 2021-02-16
internal/distro/rhel8: un-exclude subman from edge

runcom:
We aim at shrinking our deps eventually but we need subman for the time
being. This patch basically un-exclude subman which was introduced by
https://github.com/osbuild/osbuild-composer/pull/893

Signed-off-by: Antonio Murdaca <runcom@linux.com>

PR #1244, opened 2021-02-22
Support osbuild format v2

achilleas-k:
This pull request includes:

----

_Opening as draft for review and CI.  Unit tests are covered but needs image type test cases and news item._

----

## OSBuild Manifest Format Version 2!

This PR adds support for the new Manifest schema for OSBuild, introduced as a tech preview in [v25](https://github.com/osbuild/osbuild/releases/tag/v25).  The new packages, `osbuild2` contains the new types (Stages, Sources, etc.) which mostly consists of copies of the types from the old `osbuild` package with small adaptations, and adds new types introduced in OSBuild.  The old package is renamed to `osbuild1`.

## New Image Types

A new implementation of the ImageType interface is added in the `rhel84` distro which generates Manifests based on the new schema.  It adds two new image types (`rhel-edge-container` image type for x86_64 and aarch64).  The target creates a container that, when run, serves an OSTree commit.

### Image Type Name
I've been using `rhel-edge-container` while writing it and I see that name is also in @jkozol's draft PR in composer.  I'm wondering if that's accurate for the container that's serving the commit.  I was thinking about renaming it to `rhel-edge-commit-container` but that's a bit long, though I'm not sure it matters much one way or another.

### Decoding types with interface fields
Types that contain interfaces with multiple implementations implement their own `UnmarshalJSON()` method.  They use a JSON decoder with the `DisallowUnknownFields` option to catch errors during the deserialization while trying to determine which implementation matches the raw data.

### Log output
Log output from OSBuild has a very different format when using the new schema.  The `osbuild1.Result` object now supports unmarshalling the new format and adapting it to the old format.  Refer to the corresponding commit message for more information.

### Stage packages
The ImageType interface defines two sets of packages: Build packages and image packages.  The former is used to set up the build environment (during the Build stage) and the latter is the list of packages that are meant to be installed in the final output image.

With the new Manifest, the idea of an explicit Build stage is gone.  Instead we can define multiple Pipelines, each with its own set of stages.  Each of these Pipelines might require its own set of packages.  For the new image type, for example, we would need three package sets: One for the first pipeline (which in the old format would have been the build stage), one for the commit that will be delivered, and one for the container that serves the commit (which should only be `httpd` + dependencies).  Since we can't define that last package set explicitly, the image type reuses the build packages for the container.

In the (very near) future, we should generalise the ImageType interface to allow defining a separate package set for each Pipeline in the Manifest.


jkozol:
> I've been using rhel-edge-container while writing it and I see that name is also in @jkozol's draft PR in composer. I'm wondering if that's accurate for the container that's serving the commit. I was thinking about renaming it to rhel-edge-commit-container but that's a bit long, though I'm not sure it matters much one way or another.

I would prefer for it to remain `rhel-edge-container` since that is what we have specified in the UI.

gicmo:
> > I've been using rhel-edge-container while writing it and I see that name is also in @jkozol's draft PR in composer. I'm wondering if that's accurate for the container that's serving the commit. I was thinking about renaming it to rhel-edge-commit-container but that's a bit long, though I'm not sure it matters much one way or another.
> 
> I would prefer for it to remain `rhel-edge-container` since that is what we have specified in the UI.

Not a big deal I think, but it is a good point that an "edge container" is a bit undefined, now that @achilleas-k mentions it.

internal/worker/json.go
```diff
@@ -18,6 +18,7 @@ type OSBuildJob struct {
 	Targets         []*target.Target `json:"targets,omitempty"`
 	ImageName       string           `json:"image_name,omitempty"`
 	StreamOptimized bool             `json:"stream_optimized,omitempty"`
+	Exports         []string         `json:"export_stages,omitempty"`
```
teg, line 21: We have to be careful when extending this API, keeping in mind that composer and workers will not be updated in lockstep. Adding a field is not a problem, as long as we can handle the case when composer passes it to a worker that does not expect it, and that workers can deal with not being passed it.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -91,9 +91,13 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	streamOptimizedPath := ""
 
+	// NOTE: Currently OSBuild supports multiple exports, but this isn't used
+	// by any of the image types and it can't be specified during the request.
+	// Use the first (and presumably only) export for the imagePath.
+	exportPath := args.Exports[0]
```
teg, line 97: Keep in mind we need to handle the case of old composer with new workers. Maybe handle this the same as in the koji case if `len(args.Exports)==0`?

Should we could also refuse jobs with `len(args.Exports)>1`? If someone is requesting two artifacts and we return only one, something is not right...

internal/osbuild1/result.go
```diff
@@ -123,3 +126,52 @@ func (cr *Result) Write(writer io.Writer) error {
 
 	return nil
 }
+
+func (cr *Result) UnmarshalJSON(data []byte) error {
```
teg, line 130: This is obviously completely bonkers, but in a very impressive way. I think it hits the right balance between getting something useful and something we can use now. Let's go with this, and refactor once we have settled on the final format and have the luxury of more time.

teg (changes requested):
This is amazing! An impressive amount of work in very challenging circumstances. Love that this sets up the basics for the future, while making the right tradeoffs on temporary solutions. A couple of requests inline about the worker API.

internal/distro/rhel84/distro_v2.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (t *imageTypeS2) ostreePullStageInputs(options distro.ImageOptions) *osbuild.OSTreePullStageInputs {
+	pullStageInput := new(osbuild.OSTreePullStageInput)
+	pullStageInput.Type = "org.osbuild.ostree"
+	pullStageInput.Origin = "org.osbuild.pipeline"
+
+	inputRefs := make(map[string]osbuild.OSTreePullStageReference)
+	ref := options.OSTree.Ref
+	if ref == "" {
+		// TODO: Merge from Tom's PR (#1235)
```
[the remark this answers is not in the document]
achilleas-k: What good is a `todo` if I don't check for it before PRing?

Thanks for pointing it out.  Fixing.

internal/osbuild2/curl_source.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type CurlSource struct {
+	Items map[string]CurlSourceItem `json:"items"`
+}
+
+func (CurlSource) isSource() {}
+
+// CurlSourceItem can be either a URL string or a URL paired with a secrets
+// provider
```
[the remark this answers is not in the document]
achilleas-k: Especially when one is a superset of the other, it does make life simpler.  I got a bit carried away with following the spec.  I'll keep it simpler next time :)

internal/osbuild1/result.go
achilleas-k: This definitely needs a bit of discussion.  We could define the output schema for osbuild and rely on it, or we could keep it flexible not bother processing it on the receiving end in composer.  I think @gicmo told me he'd prefer it wasn't parsed and was assumed completely opaque, but I might be misremembering.

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: > Keep in mind we need to handle the case of old composer with new workers. Maybe handle this the same as in the koji case if `len(args.Exports)==0`?

Good point.  I'll cover the compatibility case by falling back to `"assembler"` when `args.Exports` isn't set.

> Should we could also refuse jobs with `len(args.Exports)>1`? If someone is requesting two artifacts and we return only one, something is not right...

There's no mechanism currently to request multiple exports (I think?).  I suppose this could happen if a newer (future) composer is paired with an older (this version) worker.  If that's a possibility, then we would have to fail the request, yes.

teg: That's something we probably want to be prepared for.

achilleas-k: Covering both cases now.

achilleas-k:
New commits add the `--export` flag to the `osbuild` call in the `generate-test-cases` script and generate a new test case for `osbuild-container`.  For x86_64 I also regenerated the existing test cases to update them with some of the recent changes (kernel metapackage, new sources, new packages).

Marking PR as ready.

internal/osbuild1/result.go
teg: I'm fine with keeping this as-is for now, what do you think @gicmo? I agree that keeping it opaque might be nicer and all we really need (we just dump it as logs anyway), but we do anyway parse the result in some cases (koji), so I'm not sure we gain anything by not parsing it?

test/data/manifests/fedora_32-x86_64-vmdk-boot.json
```diff
@@ -1,6 +1,6 @@
 {
   "boot": {
-    "type": "vmware"
```
teg, line 3: Is this right?

teg (changes requested):
This is great! Let's work with @henrywang to adapt the `ostree-ng.sh` test-case, and then this is good to go I think!
Forgot: had one nit, see inline comment.

test/data/manifests/fedora_32-x86_64-vmdk-boot.json
achilleas-k: I don't know but in the `format-request-map` for the test case generation it's set to `qemu`.  Also for fedora-33 it's already set to `qemu`.  Not sure if they're meant to be different, or if f33 is the wrong one.

teg: Ah, I see. The commit that moved from `qemu` to `vmware` did not change the generator: https://github.com/osbuild/osbuild-composer/commit/9cce43d384bb499c7667ba471cde08dd7f1b6a14

So I guess the correct fix is to change the generator here (could also file an issue and do that in a follow-up).

We may also want to drop the boot testing from the image-tests for those image types we have proper integration tests for. The image tests really mostly verify that osbuild itself doesn't change behavior (might not even belong in composer, but that's another topic). What do you think @atodorov, should we drop the boot tests from the image tests in favor of the integration tests (where we have them)?

achilleas-k: > So I guess the correct fix is to change the generator here (could also file an issue and do that in a follow-up).

Changing and regenerating.

ondrejbudai: I think that the reason is that we're using vcenter in the internal network but the integration tests for Fedora run in aws.

henrywang:
@achilleas-k, I found worker unit file changed from `osbuild-local-worker.service` to `osbuild-local-worker.socket`. Will we change the unit file name forever? If yes, I need update [test](https://github.com/osbuild/osbuild-composer/blob/ce23f97cb9c07b6065543e8e35495d8479b86bf0/test/cases/ostree-ng.sh#L148).
And will this change apply to both `Fedora` and `RHEL 8.3`? Thanks.

henrywang:
@achilleas-k, I can imported edge container image without any error. But I can't run it.
```
[cloud-user@rhel-8-4 rhel-edge]$ podman image import 15461c29-8c15-49dc-80ad-1ec90ca751e8-rhel84-container.tar rhel-edge-container:latest
Getting image source signatures
Copying blob 681618047ca9 done
Copying config bb8c7f9749 done
Writing manifest to image destination
Storing signatures
bb8c7f97494a4c3d8ffc3ecf5f3efee8611fa4c798301f51a733bb39bfae838f
[cloud-user@rhel-8-4 rhel-edge]$ podman images
REPOSITORY                               TAG     IMAGE ID      CREATED        SIZE
docker.io/library/rhel-edge-container    latest  bb8c7f97494a  7 minutes ago  832 MB
registry.access.redhat.com/ubi8-minimal  latest  7331d26c1fdf  2 months ago   105 MB
[cloud-user@rhel-8-4 rhel-edge]$ podman run -d --network host docker.io/library/rhel-edge-container:latest
Error: no command or entrypoint provided, and no CMD or ENTRYPOINT from image
[cloud-user@rhel-8-4 rhel-edge]$ podman run -d --network host docker.io/library/rhel-edge-container:latest httpd -D
Error: container_linux.go:370: starting container process caused: exec: "httpd": executable file not found in $PATH: OCI not found
cloud-user@rhel-8-4 rhel-edge]$ podman run -d --network host docker.io/library/rhel-edge-container:latest /usr/sbin/httpd -D FOREGROUND
Error: container_linux.go:370: starting container process caused: exec: "/usr/sbin/httpd": stat /usr/sbin/httpd: no such file or directory: OCI not found
```
From [code](https://github.com/osbuild/osbuild-composer/pull/1244/files#diff-9fa981e4b66dacb0d4825d435b2eb67e5978b6de086737d080fd362cff8bbe99R229), httpd should be installed and it's run by CMD. But looks it does not work in this case.
Could you please take a look if you have time? Thanks.

gicmo:
> @achilleas-k, I can import edge container image without any error. But I can't run it.

`podman import` is meant as a counter part of `podman export` and it is not specified which format is used.

Try one of the following:
- `podman pull oci-archive:15461c29-8c15-49dc-80ad-1ec90ca751e8-rhel84-container.tar`
- `cat 15461c29-8c15-49dc-80ad-1ec90ca751e8-rhel84-container.tar | podman load`

achilleas-k:
@henrywang I get the same behaviour when I `import` the file.  @gicmo's suggestion should work.

As for the service/socket files: I don't think anything has changed there recently.  The worker service depends on worker socket, so the socket starts automatically when you start the service.  I think the test doesn't require any changes there.

henrywang:
> @henrywang I get the same behaviour when I `import` the file. @gicmo's suggestion should work.
> 
> As for the service/socket files: I don't think anything has changed there recently. The worker service depends on worker socket, so the socket starts automatically when you start the service. I think the test doesn't require any changes there.

Thanks for clearing this.
1. Test code in [this line](https://github.com/osbuild/osbuild-composer/blob/ce23f97cb9c07b6065543e8e35495d8479b86bf0/test/cases/ostree-ng.sh#L148) failed because worker unit file changed from `osbuild-local-worker.service` to `osbuild-local-worker.socket`. After I update code to `osbuild-local-worker.socket`, it works.
2. @gicmo's comment works for me. But I still get error. Anything I missed? Thanks.
```
[cloud-user@rhel-8-4 rhel-edge]$ podman pull oci-archive:15461c29-8c15-49dc-80ad-1ec90ca751e8-rhel84-container.tar
Getting image source signatures
Copying blob 3e6464084597 done
Copying config 67607b0bec done
Writing manifest to image destination
Storing signatures
67607b0bece0b407c860dec7d20018c4c8ac396a805d7f3c276d23207f6c6e34

[cloud-user@rhel-8-4 rhel-edge]$ podman images
REPOSITORY                               TAG     IMAGE ID      CREATED       SIZE
<none>                                   <none>  67607b0bece0  6 hours ago   1.34 GB
registry.access.redhat.com/ubi8-minimal  latest  7331d26c1fdf  2 months ago  105 MB

[cloud-user@rhel-8-4 rhel-edge]$ podman run -d --name edge --network host 67607b0bece0
Error: container_linux.go:370: starting container process caused: exec: "httpd": executable file not found in $PATH: OCI not found

[cloud-user@rhel-8-4 rhel-edge]$ podman ps -a
CONTAINER ID  IMAGE         COMMAND               CREATED         STATUS   PORTS   NAMES
d2d3757bd2b7  67607b0bece0  httpd -D FOREGROU...  34 seconds ago  Created          edge
```

teg:
rather than unmarshaling the pipeline, could we just always name the last one "assembler"?

achilleas-k:
> rather than unmarshaling the pipeline, could we just always name the last one "assembler"?

 I suppose that would make things simpler.

achilleas-k:
> 2\. @gicmo's comment works for me. But I still get error. Anything I missed? Thanks.

Found the issue there.  https isn't added to the container.  At some point while finalising the changes, I must have dropped it from the package list for the container.  I'm fixing it now.

internal/distro/rhel84/distro.go
```diff
@@ -1202,6 +1228,113 @@ func newDistro(isCentos bool) distro.Distro {
...
+			"greenboot", "greenboot-grub2", "greenboot-rpm-ostree-grub2", "greenboot-reboot", "greenboot-status",
+			// x86 specific
+			"grub2", "grub2-efi-x64", "efibootmgr", "shim-x64", "microcode_ctl",
+			"iwl1000-firmware", "iwl100-firmware", "iwl105-firmware", "iwl135-firmware",
+			"iwl2000-firmware", "iwl2030-firmware", "iwl3160-firmware", "iwl5000-firmware",
+			"iwl5150-firmware", "iwl6000-firmware", "iwl6050-firmware", "iwl7260-firmware",
+		},
+		excludedPackages: []string{
+			"rng-tools",
+			"subscription-manager",
```
henrywang, line 1275: Looks like `subscription-manager` from [this PR](https://github.com/osbuild/osbuild-composer/pull/1231) already.

achilleas-k: Yes, I missed that.  Thanks.

achilleas-k:
New commits contain some rebasing with older commits so here's a list of changes:
- Added `httpd` to the `BuildPackages()` method of the new `imageTypeS2`.
- Renamed the last pipeline for the new manifests to "assembler" for compatibility with the old format.
- Changed the `packages`, `excludedPackages`, and `enabledServices` fields of the new image types to reuse the ones defined for the `rhel-edge-commit` image types since the payloads should be the same.

teg (dismissed):
Thanks! I particularly like that you reuse the package sets/services now.

gicmo:
>     * Renamed the last pipeline for the new manifests to "assembler" for compatibility with the old format.

Heh, that seems like a hack. Not strongly against it, but is that meant to stay like that? Or just for now?

achilleas-k:
> > ```
> > * Renamed the last pipeline for the new manifests to "assembler" for compatibility with the old format.
> > ```
> 
> Heh, that seems like a hack. Not strongly against it, but is that meant to stay like that? Or just for now?

For now.  Internally now it doesn't matter since the image types declare their exports themselves and that works.  This is for the image tests that read the pre-generated manifests from the test cases.  I initially made it so that the test unmarshals the manifest figuring out if it's v1 or v2 and adding the appropriate value to the export option.  Making them all "assembler" simplifies the situation, but in the future, we read the manifest and get the export that way, or we can add it to the test cases as an additional option.

teg:
> > > ```
> > > * Renamed the last pipeline for the new manifests to "assembler" for compatibility with the old format.
> > > ```
> > 
> > 
> > Heh, that seems like a hack. Not strongly against it, but is that meant to stay like that? Or just for now?
> 
> For now. Internally now it doesn't matter since the image types declare their exports themselves and that works. This is for the image tests that read the pre-generated manifests from the test cases. I initially made it so that the test unmarshals the manifest figuring out if it's v1 or v2 and adding the appropriate value to the export option. Making them all "assembler" simplifies the situation, but in the future, we should read the manifest and get the export that way, or we can add it to the test cases as an additional option.

I like adding an additional option to the test cases (in the future), I think we already have that in a couple of places to avoid having to parse the manifest.

henrywang:
@achilleas-k, Thanks for fixing issue! :+1: 
When I run container, I got the following error. Could you please take a look if you have time? Thanks!
```
[cloud-user@rhel-8-4 rhel-edge]$ sudo podman ps -a
CONTAINER ID  IMAGE         COMMAND               CREATED        STATUS                    PORTS   NAMES
204fcb0b1263  7baa87c7264e  httpd -D FOREGROU...  6 seconds ago  Exited (1) 6 seconds ago          rhel-edge

[cloud-user@rhel-8-4 rhel-edge]$ sudo podman logs rhel-edge
httpd: Syntax error on line 59 of /etc/httpd/conf/httpd.conf: Could not open config directory /etc/httpd/conf.modules.d: Permission denied
```

achilleas-k:
> @achilleas-k, Thanks for fixing issue! 
> When I run container, I got the following error. Could you please take a look if you have time? Thanks!
> 
> ```
> [cloud-user@rhel-8-4 rhel-edge]$ sudo podman ps -a
> CONTAINER ID  IMAGE         COMMAND               CREATED        STATUS                    PORTS   NAMES
> 204fcb0b1263  7baa87c7264e  httpd -D FOREGROU...  6 seconds ago  Exited (1) 6 seconds ago          rhel-edge
> 
> [cloud-user@rhel-8-4 rhel-edge]$ sudo podman logs rhel-edge
> httpd: Syntax error on line 59 of /etc/httpd/conf/httpd.conf: Could not open config directory /etc/httpd/conf.modules.d: Permission denied
> ```

Turns out this is an SELinux issue.  OSBuild is creating the container archive with SELinux labels that end up on the host when doing `podman pull`.
@gicmo opened a RHBZ for it just now: https://bugzilla.redhat.com/show_bug.cgi?id=1933774

gicmo:
> Turns out this is an SELinux issue. OSBuild is creating the container archive with SELinux labels that end up on the host when doing `podman pull`.
> @gicmo opened a RHBZ for it just now: https://bugzilla.redhat.com/show_bug.cgi?id=1933774

As a work around for now, to get this PR moving forward, we could so something like `sudo restorecon -R /var/lib/containers/storage/overlay/` after the `podman pull/load` command to label it properly.

achilleas-k:
Relevant fix in osbuild https://github.com/osbuild/osbuild/pull/598

henrywang:
@gicmo, awesome! Your work around works for me.
@achilleas-k, I run test locally, everything works well. Thanks.

Integration test code (Installation only) is ready, but there's an issue about upgrade.
Integration test [ostree-ng.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree-ng.sh) workflow is:
1. build edge image
2. install edge image on VM
3. build edge upgrade image
4. upgrade(`rpm-ostree upgrade`) on VM

If the edge image is a container, the `upgrade` part will not work because `pull-local` and `summary` will not work in this case.

My plan is dropping upgrade test in [ostree-ng.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree-ng.sh) because upgrade test is covered already in [ostree.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree.sh).

Do you have any thoughts and ideas? Add @teg here for discussion. Thanks.

gicmo:
> When I check installed VM, one issue I found about `journal`. No this issue found before. I have no idea about this issue. Could you please take a look if you have time? Thanks.

Yeah, this seems to be a permission problem. Maybe something in the journal policy changed so that being in the `wheels` group is not sufficient anymore?

achilleas-k:
> @gicmo, awesome! Your work around works for me.
> @achilleas-k, I run test locally, everything(image build, ostree install, and ostree upgrade) works well. Thanks.
> I'm working on Integration test code right now.
> 
> When I check installed VM, one issue I found about `journal`. No this issue found before. I have no idea about this issue. Could you please take a look if you have time? Thanks.
> 
> ```
> [admin@vm ~]$ journalctl -u ostree-remount
> Hint: You are currently not seeing messages from other users and the system.
>       Users in the 'systemd-journal' group can see all messages. Pass -q to
>       turn off this notice.
> No journal files were opened due to insufficient permissions.
> 
> [admin@vm ~]$ journalctl -q -u ostree-remount
> No journal files were opened due to insufficient permissions.
> 
> [admin@vm ~]$ sudo journalctl -u ostree-remount
> -- Logs begin at Tue 2021-03-02 06:56:00 UTC, end at Tue 2021-03-02 06:58:06 UTC. --
> Mar 02 06:56:04 localhost.localdomain systemd[1]: Starting OSTree Remount OS/ Bind Mounts...
> Mar 02 06:56:04 localhost.localdomain ostree-remount[711]: Remounted rw: /sysroot
> Mar 02 06:56:04 localhost.localdomain ostree-remount[711]: Remounted rw: /var
> Mar 02 06:56:04 localhost.localdomain systemd[1]: Started OSTree Remount OS/ Bind Mounts.
> ```
> 
> ```
> [admin@vm ~]$ rpm-ostree status
> State: idle
> Warning: failed to query journal: couldn't find current boot in journal
> Deployments:
> * ostree://rhel-edge-container:rhel/8/x86_64/edge
>                    Version: 8.4 (2021-03-02T06:52:35Z)
>                     Commit: e25308679ef305496c5cbe5f05ecbe1da16e182fc6363394dc6ceb9ff62c39fb
> 
>   ostree://rhel-edge-container:rhel/8/x86_64/edge
>                    Version: 8.4 (2021-03-02T06:39:42Z)
>                     Commit: 772049c102aa0e85c2f62c06bccf51d7bd4d4cd344db07c24862aeb848df5e46
> [admin@vm ~]$ sudo rpm-ostree status
> State: idle
> Deployments:
> * ostree://rhel-edge-container:rhel/8/x86_64/edge
>                    Version: 8.4 (2021-03-02T06:52:35Z)
>                     Commit: e25308679ef305496c5cbe5f05ecbe1da16e182fc6363394dc6ceb9ff62c39fb
> 
>   ostree://rhel-edge-container:rhel/8/x86_64/edge
>                    Version: 8.4 (2021-03-02T06:39:42Z)
>                     Commit: 772049c102aa0e85c2f62c06bccf51d7bd4d4cd344db07c24862aeb848df5e46
> ```

This was a minor bug cause by a change in osbuild.  I've reproduced the issue locally and confirmed that https://github.com/osbuild/osbuild/pull/600 fixes it.

teg:
To get this PR to green, we could add a temporary work-around by pinning osbuild `main` in the `Schutzfile`. That way we could get the PR tested and we could take that to the exception meeting on (a) Thursday, to get it approved. Then, once `osbuild` lands downstream we can remove the work-around and land the PR. We could also add the installer stuff to the same PR, so we end up with only one RPM to verify / take to the exception meeting.

achilleas-k:
So checklist now for this PR is:
- [ ] Pin to `osbuild/main` for tests
- [ ] Include RHEL for Edge commit installer image type

henrywang:
@achilleas-k, @teg. I have [my test update](https://github.com/osbuild/osbuild-composer/pull/1244/commits/1535cf5dd0ee6b6e935ac706bae61c622b7b4db1) pushed already. [This test](https://github.com/osbuild/osbuild-composer/blob/f091af55d89ac9e77aa34b94e0180aacead3f32e/test/data/ansible/check_ostree.yaml#L206) will fail due to osbuild/osbuild#600. But I can work around this issue with `sudo` added. Do you have any other thoughts? Thanks.

achilleas-k:
> @achilleas-k, @teg. I have [my test update](https://github.com/osbuild/osbuild-composer/pull/1244/commits/1535cf5dd0ee6b6e935ac706bae61c622b7b4db1) pushed already. [This test](https://github.com/osbuild/osbuild-composer/blob/f091af55d89ac9e77aa34b94e0180aacead3f32e/test/data/ansible/check_ostree.yaml#L206) will fail due to [osbuild/osbuild#600](https://github.com/osbuild/osbuild/pull/600). But I can work around this issue with `sudo` added. Do you have any other thoughts? Thanks.

New commit runs tests using current `osbuild/main` which includes https://github.com/osbuild/osbuild/pull/600.

henrywang:
@achilleas-k @gicmo. Found another issue about greenboot. All greenboot* services are not enabled by default. You can find the test result from https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1244/16/pipeline/312/#step-864-log-140
Could you please have a look if you have time? Thanks.

gicmo:
> @achilleas-k @gicmo. From the EL8.4 New OSTree result, everything(install, upgrade) works pretty well. But I found one more issue about greenboot.
> All `greenboot*` services are not enabled by default. You can find the test result from https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1244/16/pipeline/312/#step-864-log-140
> Could you please have a look if you have time? Thanks.

We are missing the enabled services that the "old" [definition](https://github.com/osbuild/osbuild-composer/blob/f091af55d89ac9e77aa34b94e0180aacead3f32e/internal/distro/rhel84/distro.go#L830) has.

I actually think the right way is to, maybe after this PR is good to go, build the ostree-commit itself with the very same pipeline. This means porting the tar artefact to the v2 manifest. Otherwise we might get different content based on the "wrapping" which I think could be a major headache.

achilleas-k:
Updated PR with rebase on main and (more importantly) new stages that were added and changes that were made between osbuild v25 and v26.

achilleas-k:
Updated with new depsolving workflow inside `ImageType.Manifest()` as discussed with @gicmo and @teg.

**Here's a description of the new process:**
The new ImageType now includes a `packageSets [][]string` field to replace the `packages []string` field of existing types.  Similarly, `excludedPackages []string` has been replaced by `exludedPackageSets [][]string`, though this might not be necessary.
For the ImageType to run depsolve, it needs access to the `rpmmd.Depsolve()` method and information defined in the weldr API package.  The new image type adds a `SetSolver()` method (and the type is now public) so that external packages can add a solver function and the blueprint for the compose.  The weldr API uses this method when the name of the image type is `"rhel-edge-container"`.
We could simplify this by just testing the type assertion for `ImageTypeS2`.

The existing `Packages()` and `BuildPackages()` method now return empty arrays to avoid unnecessary depsolving from the existing process.

The mapping of the package sets to purpose is implicit and depends on the image type knowing the order in which they're defined.  This is probably fine, but it would perhaps be more readable as a map, where each set has a descriptive name (e.g., `"commit-packages"`).  On a similar note, we could have package sets that define a included and excluded packages that go together (recommended by @gicmo), assuming we continue using package exclusion as we do now.

I'm not entirely certain this is the best way but _It Works_.  However, it's a pretty big change and still needs a little bit of compatibility tweaks for testing to work.  Certain tests and other routes (e.g., `osbuild-pipeline`) fail because the depsolving of the packages for the image type doesn't work the same way anymore.

To quote a modern philosopher:
> ... that hackery code we currently have will hopefully vanish soon. I am sorry though for this mess.

achilleas-k:
Newest commits now include the `rhel-edge-installer` image definition.  It's now rebased on main and uses the PackageSets() from #1279.

henrywang:
@ondrejbudai @teg Do you know why our Jenkins CI tests is not listed here? Thanks.

teg:
unit tests failed, they need to pass before jenkins is triggered

achilleas-k:
Getting to the unit tests now.  I'll need to adjust some things with the generated manifests.

achilleas-k:
Manifests updated.  Unit tests should be passing now @henrywang.

henrywang:
@achilleas-k @gicmo I got the following error when I boot from ISO and enter into `Dracut Emergency Shell`
Could you please have a look if you have time? Thanks.
```
[  OK  ] Started Open-iSCSI.
         Starting dracut initqueue hook...
[  143.738176] dracut-initqueue[1025]: Warning: dracut-initqueue timeout - starting timeout scripts
[  144.313823] dracut-initqueue[1025]: Warning: dracut-initqueue timeout - starting timeout scripts
[  144.870715] dracut-initqueue[1025]: Warning: dracut-initqueue timeout - starting timeout scripts
[  145.440527] dracut-initqueue[1025]: Warning: dracut-initqueue timeout - starting timeout scripts
[  145.995295] dracut-initqueue[1025]: Warning: dracut-initqueue timeout - starting timeout scripts
[  209.634282] localhost dracut-initqueue[1025]: Warning: Could not boot.
[  209.654526] localhost systemd[1]: Starting Setup Virtual Console...
[  209.663000] localhost systemd-vconsole-setup[6881]: KD_FONT_OP_GET failed while trying to get the font metadata: Invalid argument
[  209.663291] localhost systemd-vconsole-setup[6881]: Fonts will not be copied to remaining consoles
[  209.663926] localhost systemd[1]: systemd-vconsole-setup.service: Succeeded.
[  209.664286] localhost systemd[1]: Started Setup Virtual Console.
[  209.667936] localhost systemd[1]: Starting Dracut Emergency Shell...
[  209.688070] localhost systemd[1]: Received SIGRTMIN+21 from PID 973 (plymouthd).
[  209.698742] localhost systemd[1]: Received SIGRTMIN+21 from PID 973 (plymouthd).
dracut:/#
```

gicmo:
Github doesn't want me to comment in-line so: Looking at [osinfo-db](https://gitlab.com/libosinfo/osinfo-db/-/blob/master/data/os/redhat.com/rhel-8.4.xml.in#L20) I think we have to stick to `RHEL-8-4-(([0-9]|[1-9][0-9])-)?BaseOS-x86_64`, i.e. `RHEL-8-4-0-BaseOS-x86_64`, for the volume id of the iso.

achilleas-k:
Installer image updated to pull commit from URL during build.  The ostree source functionality requires a fix that just landed in osbuild today (https://github.com/osbuild/osbuild/pull/610), so we're going to have to update the commit for Schutzbot again.

The Volume ID for the ISO was changed to match @gicmo's comment above `RHEL-8-4-0-BaseOS-{arch}`.

Regarding test cases: We agreed that building an installer without a commit URL should fail, since it doesn't make sense to build the installer ISO with nothing in it.  This prevents us from creating a test case manifest for the image type.  If we do add a URL to the test manifest, it needs to be a valid (and live) URL for the image to be built successfully.  Do we skip having an image test for it and rely on integration test scripts alone?  If so, do we still need to update the OSBuild dependency in the Schutzfile?

internal/distro/rhel84/distro_v2.go
```diff
@@ -448,3 +572,173 @@ func (t *imageTypeS2) systemdStageOptions(enabledServices, disabledServices []st
...
+	}
+}
+
+func (t *imageTypeS2) bootISOMonoStageOptions(kernelVer string) *osbuild.BootISOMonoStageOptions {
+	return &osbuild.BootISOMonoStageOptions{
+		Product: osbuild.Product{
+			Name:    "Red Hat Enterprise Linux",
+			Version: "8.4",
+		},
+		ISOLabel: "RHEL-8-4-X86_64",
```
teg, line 681: Should this be arch-independent?

henrywang:
@achilleas-k @gicmo I got the following error. Does `rhel-edge-installer` not support rt-kernel and user? Thanks.
```
failed to create osbuild manifest: boot ISO image type "rhel-edge-installer" does not support blueprint customizations
```

internal/distro/rhel84/distro_v2.go
gicmo: This is actually used to build the kernel command line `cmdline = f"inst.stage2=hd:LABEL={isolabel}"` which means it needs to match the `VolID` of the below, i.e. be `fmt.Sprintf("RHEL-8-4-0-BaseOS-%s", t.Arch().Name()),`.

gicmo:
> @achilleas-k @gicmo I got the following error. Does `rhel-edge-installer` not support rt-kernel and user? Thanks.
> 
> ```
> failed to create osbuild manifest: boot ISO image type "rhel-edge-installer" does not support blueprint customizations
> ```

In the new version the installer just embeds the commit that is specified via the blueprint, but the commit self can only be customised (i.e. kernel) during the build of the commit itself. So I do think this is indeed correct.

achilleas-k:
> > @achilleas-k @gicmo I got the following error. Does `rhel-edge-installer` not support rt-kernel and user? Thanks.
> > 
> > ```
> > failed to create osbuild manifest: boot ISO image type "rhel-edge-installer" does not support blueprint customizations
> > ```
> 
> In the new version the installer just embeds the commit that is specified via the blueprint, but the commit self can only be customised (i.e. kernel) during the build of the commit itself. So I do think this is indeed correct.


Yes, just to be clear, the following is how we agreed it should work:
- rhel-edge-container: OCI container with rhel-edge-commit inside that starts a web server when run
- rhel-edge-installer: boot ISO image that embeds the rhel-edge-commit.  The commit is embedded in the ISO by being pulled from the running container during the compose of the boot ISO.

- Blueprints and customizations go into the rhel-edge-container.
- Blueprint options are ignored when building a rhel-edge-installer.
- Requesting a rhel-edge-installer without specifying a URL should return an error.

I decided to make it an error if there are customisations to make it clear to the user that they have no effect.  The installer will set up the system based on the commit served at the URL specified in the request.

henrywang:
Thanks @achilleas-k @gicmo! That makes more sense to me. I think we need Doc team to update 8.4 doc with all these tips added.

achilleas-k:
> Thanks @achilleas-k @gicmo! That makes more sense to me. I think we need Doc team to update 8.4 doc with all these tips added.

Definitely.  I'm sure it can be quite confusing (it was for me :) )

internal/distro/rhel84/distro_v2.go
achilleas-k: Fixed to match the VolID.  Thanks!

henrywang:
@achilleas-k @gicmo A `worker` error found when build installer image. That caused compose keeps at RUNNING status. Could you please have a look if you have time? Thanks.
```
Mar 13 08:46:25 rhel-8-4 osbuild-worker[1761134]: Waiting for a new job...
Mar 13 08:47:11 rhel-8-4 osbuild-worker[1761134]: Running 'osbuild' job 5f55ae49-296b-4ad4-b23f-63c24020a444
Mar 13 08:47:18 rhel-8-4 osbuild-worker[1761134]: ostree init --mode=archive --repo=/var/cache/osbuild-worker/osbuild-store/sources/org.osbuild.ostree/repo
Mar 13 08:47:18 rhel-8-4 osbuild-worker[1761134]: ostree config set repo.locking true --repo=/var/cache/osbuild-worker/osbuild-store/sources/org.osbuild.ostree/repo
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: error: Importing a6be77f528449fd8b330171ac84f849114fe1a08a8467535c9f028dcfdc0b477.commit: linkat: No such file or directory
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: ostree init --mode=archive --repo=/var/cache/osbuild-worker/osbuild-store/tmp/store-server-wcpja5ag/files-outputcq4cwqju/repo
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: exporting a6be77f528449fd8b330171ac84f849114fe1a08a8467535c9f028dcfdc0b477
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: ostree pull-local /var/cache/osbuild-worker/osbuild-store/sources/org.osbuild.ostree/repo a6be77f528449fd8b330171ac84f849114fe1a08a8467535c9f028dcfdc0b477 --repo=/var/cache/osbuild-worker/osbuild-store/tmp/store-server-wcpja5ag/files-outputcq4cwqju/repo
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: Traceback (most recent call last):
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/osbuild/inputs/org.osbuild.ostree", line 129, in <module>
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:     r = main()
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/osbuild/inputs/org.osbuild.ostree", line 123, in main
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:     export(refs, source, output)
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/osbuild/inputs/org.osbuild.ostree", line 86, in export
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:     repo=repo_out)
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/osbuild/inputs/org.osbuild.ostree", line 71, in ostree
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:     check=True)
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib64/python3.6/subprocess.py", line 438, in run
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]:     output=stdout, stderr=stderr)
Mar 13 08:52:29 rhel-8-4 osbuild-worker[1761134]: subprocess.CalledProcessError: Command '['ostree', 'pull-local', '/var/cache/osbuild-worker/osbuild-store/sources/org.osbuild.ostree/repo', 'a6be77f528449fd8b330171ac84f849114fe1a08a8467535c9f028dcfdc0b477', '--repo=/var/cache/osbuild-worker/osbuild-store/tmp/store-server-wcpja5ag/files-outputcq4cwqju/repo']' returned non-zero exit status 1.
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]: Traceback (most recent call last):
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/bin/osbuild", line 11, in <module>
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     load_entry_point('osbuild==26', 'console_scripts', 'osbuild')()
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/main_cli.py", line 141, in osbuild_cli
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     output_directory=output_directory
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/pipeline.py", line 285, in build
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     res = pl.run(store, monitor, libdir, output_directory)
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/pipeline.py", line 245, in run
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     results, _, obj = self.build_stages(store, monitor, libdir)
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/pipeline.py", line 213, in build_stages
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     libdir)
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/pipeline.py", line 96, in run
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     path, data = ip.run(storeapi)
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:   File "/usr/lib/python3.6/site-packages/osbuild/inputs.py", line 94, in run
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]:     raise RuntimeError(f"{name}: error: {r.stderr}") from None
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]: RuntimeError: org.osbuild.ostree: error: None
Mar 13 08:52:32 rhel-8-4 osbuild-worker[1761134]: 2021/03/13 08:52:32 Job 5f55ae49-296b-4ad4-b23f-63c24020a444 failed: error decoding osbuild output: unexpected end of JSON input
```

Schutzfile
```diff
@@ -21,6 +21,13 @@
       }
     }
   },
+  "rhel-8.4": {
+    "dependencies": {
+      "osbuild": {
+        "commit": "846b3274e35d7d088373be5601442374f628c3d0"
```
teg, line 27: This needs to be updated to `2dcc1d9ceea9df74f0176b503b1c59cb5ae122ad`.

henrywang:
@achilleas-k Could you please update Schutzfile with new osbuild master commit which has osbuild PR#611 included? I need CI to build new osbuild to run test. Thanks!

achilleas-k:
> @achilleas-k Could you please update Schutzfile with new osbuild master commit which has osbuild PR#611 included? I need CI to build new osbuild to run test. Thanks!

Yes, of course.  Thanks for the reminder.

henrywang:
@achilleas-k @gicmo @teg everything works now. :100:  `--location` works with `installer iso`. Thanks @gicmo! :+1: 

Installer iso is able to be installed and can be updated after re-configure ostree repo. But there's still two issues I found in my test.

1. Only %post part of injected kickstart file can be run, instructions like `user`, `ostreesetup` does not work. Looks they can't be overwritten by kickstart file inside iso.
2. Creating `home` directory in osbuild `users` stage does not work with ostree image. We should make `home` as `/var/home/` for ostree image. But it's not issue caused by this PR.

I'm working on automation right now. Thanks all for helping me on testing! :1st_place_medal:

gicmo:
> @achilleas-k @gicmo @teg everything works now. 💯 `--location` works with `installer iso`. Thanks @gicmo! 👍
> 
> Installer iso is able to be installed and can be updated after re-configure ostree repo. But there's still two issues I found in my test.

Yay! ;) Very cool


> 1. Only %post part of injected kickstart file can be run, instructions like `user`, `ostreesetup` does not work. Looks they can't be overwritten by kickstart file inside iso.

That `user` bit is on purposes, since the user configuration should be done when you create the commit. We disable most of the anaconda modules. `ostreesetup` is interesting since we use that to actually tell anaconda where the ostree repo is. It *should* be overwritable, I think.

> 2. Creating `home` directory in osbuild `users` stage does not work with ostree image. We should make `home` as `/var/home/` for ostree image. But it's not issue caused by this PR.

I thought the `ostree.preptree` stage should migrate existing user homes, maybe it does not. I will have a look tomorrow (on PTO today) if nobody else is faster.

> I'm working on automation right now. Thanks all for helping me on testing! 🥇

Thanks for dong the tests! Very much appreciated.

achilleas-k:
> Thanks all for helping me on testing! 1st_place_medal

Thanks for the amazing work on the testing!

henrywang:
@achilleas-k @gicmo @teg All green for all `ostree*` tests. YEAH! 😍🤞

achilleas-k:
That's amazing!  Thank you so much.

Now I need to figure out how I broke the Azure credentials in CI.

achilleas-k:
> Now I need to figure out how I broke the Azure credentials in CI.

Looks like it may had been an outdated run we were looking at.  Now the only failure seems to be good old openstack boot.

internal/distro/rhel84/distro_v2.go
```diff
@@ -0,0 +1,745 @@
...
+			"root":    rootUser,
+			"install": installUser,
+		},
+	}
+
+	p.AddStage(osbuild.NewUsersStage(usersStageOptions))
+	p.AddStage(osbuild.NewAnacondaStage(t.anacondaStageOptions()))
+	p.AddStage(osbuild.NewLoraxScriptStage(t.loraxScriptStageOptions()))
+	p.AddStage(osbuild.NewDracutStage(t.dracutStageOptions(kernelVer)))
+	p.AddStage(osbuild.NewKickstartStage(t.kickstartStageOptions(options.OSTree.URL)))
```
gicmo, line 415: Since the commit is embedded into the installer, this should not be `options.OSTree.URL` but `file:///ostree/repo`

internal/distro/rhel84/distro_v2.go
```diff
@@ -532,6 +534,27 @@ func (t *imageTypeS2) userStageOptions(users []blueprint.UserCustomization) (*os
 	return &options, nil
 }
 
+func (t *imageTypeS2) usersFirstBootOptions(usersStageOptions *osbuild.UsersStageOptions) *osbuild.FirstBootStageOptions {
+	cmds := make([]string, 0, 2*len(usersStageOptions.Users))
```
gicmo, line 538: `3*`? `3*len()+1`? :)

achilleas-k: Good catch.  It's not biggie since it's just capacity, but might as well fix it.

teg (approved):
Thanks to everyone who helped on this massive effort! @achilleas-k @henrywang and @gicmo  in particular!

PR #1245, opened 2021-02-22
Cloud API: Add support for Google Cloud Platform as a target

thozza:
- Add internal Google Cloud Platform API supporting uploading images to GCP and sharing them with specified accounts.
- Add CLI tool `osbuild-upload-gcp` using the internal GCP API.
- Extend Cloud API to support GCP as an upload target.
- Extend worker job implementation to be able to handle GCP as an upload target.
- Rework `api.sh` to allow integration testing for multiple cloud providers.
- Extend `api.sh` to also test GCP.

This pull request includes:

  - Integration test `api.sh` has been extended to test also GCP integration.
  - Updated `test/README.md` to note integration testing of Cloud API
  - Added NEWS entry for the new functionality.


thozza:
I still need to verify that the changes don't affect `cockpit-composer`, but feel free to review in the meantime. Thanks!

internal/cloudapi/openapi.yml
```diff
@@ -243,6 +253,36 @@ components:
...
+        - bucket
+      properties:
+        region:
+          type: string
+          example: 'eu'
+          description: 'A valid GCP location. See https://cloud.google.com/storage/docs/locations'
+        bucket:
+          type: string
+          example: 'my-example-bucket'
+# don't expose the os type for now
```
teg, line 268: Any particular reason why/why not?

internal/cloudapi/server.go
```diff
@@ -204,9 +204,48 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
...
+			object := fmt.Sprintf("composer-api-%s", uuid.New().String())
+			t := target.NewGCPTarget(&target.GCPTargetOptions{
+				Filename:          imageType.Filename(),
+				Region:            region,
+				Os:                "", // not exposed in cloudapi for now
+				Bucket:            gcpUploadOptions.Bucket,
+				Object:            object,
+				ShareWithAccounts: share,
+			})
+			// TODO: randomize the image name by appending some suffix?
```
teg, line 238: I would say: either let the caller decide the name, and require them to choose a unique one, or don't and use the random string as below.

tools/provision.sh
```diff
@@ -12,8 +12,11 @@ if [[ $ID == rhel || $ID == centos ]]; then
 fi
 
 sudo mkdir -p /etc/osbuild-composer
-sudo cp -a /usr/share/tests/osbuild-composer/composer/*.toml \
+sudo mkdir -p /etc/osbuild-worker
+sudo cp -a /usr/share/tests/osbuild-composer/composer/osbuild-composer.toml \
     /etc/osbuild-composer/
+sudo cp -a /usr/share/tests/osbuild-composer/composer/osbuild-worker.toml \
```
teg, line 18: Maybe we should have a separate `worker` subdirectory which is copied to `/etc/osbuild-worker`?

test/cases/api.sh
```diff
@@ -198,73 +325,104 @@ do
...
+
+  # Verify that the image was shared
+  SHARE_OK=1
+  $GCP_CMD compute images get-iam-policy "$GCP_IMAGE_NAME" > "$WORKDIR/image-iam-policy.json"
+  SHARED_ACCOUNT=$(jq -r '.bindings[0].members[0]' "$WORKDIR/image-iam-policy.json")
+  SHARED_ROLE=$(jq -r '.bindings[0].role' "$WORKDIR/image-iam-policy.json")
+  if [ "$SHARED_ACCOUNT" != "$GCP_API_TEST_SHARE_ACCOUNT" ] || [ "$SHARED_ROLE" != "roles/compute.imageUser" ]; then
+    SHARE_OK=0
+  fi
+
```
teg, line 414: This should probably also instantiate / boot the image like for AWS, right? For a follow-up?

internal/cloudapi/server.go
```diff
@@ -298,7 +298,7 @@ func (server *Server) ComposeStatus(w http.ResponseWriter, r *http.Request, id s
 			Status: composeStatusFromJobStatus(status, &result),
 			UploadStatus: &UploadStatus{
 				Status: result.UploadStatus,
-				Type:   "aws",
```
teg, line 262: @Gundersanne this won't be a problem for image-builder, I assume? This value is currently ignored?

test/README.md
```diff
@@ -206,6 +217,65 @@ from the local directory without installing it:
 * `make build` - will build everything under `cmd/`
 * `./osbuild-composer-cli-tests` - will execute the freshly built integration test suite
 
+### Cloud API integration testing
```
teg, line 220: Love this documentation! Another way to test this would be to use `docker-compose up --build` (introduced by @achilleas-k) rather than building/installing on your local machine. To make that really smooth, we may want to introduce a third container in which to actually run the test scripts, but not sure if that's necessary.
No need to change this or look into the docker stuff now though, just wanted to mention it in case you didn't know about it.

teg (approved):
This is really great! Thanks so much @thozza!

teg:
Seems we need some creds in Jenkins though. Do we have some we can use, or are we blocked on this?

internal/cloudapi/server.go
Gundersanne: This is fine yea, it's currently ignored as you say.

internal/cloudapi/openapi.yml
thozza: Honestly, this is one thing that I wanted to discuss as part of the PR. The "os type" is used when importing an image into Compute Node from previously uploaded image from Storage. There are basically three options in the related behavior of the import:

1. You don't specify the "os type" and you also don't specify that the imported image should be treated as "data disk" - In this scenario the GCP import job will try to inspect the image and if it finds out that the OS is supported, it will modify it on the fly by installing some GCP repositories and packages. If the OS is not supported (e.g. Fedora), then the import fails.
2. You specify the "os type" - GCP will inspect the image, but will disregard the OS type and treat it as if it was the specified OS type. Modification and installation of repos and packages is handled the same way as in case (1.)
3. You specify that the image is "data disk" - GCP treats this image as if it was not bootable and does not try to modify it. This does not have any other effect on the usage of the image, other than that GCP won't be able to inject some information into it (like create user, inject SSH keys, ...) and won't be able to use RHUI infrastructure inside GCP.

Based on the discussion we had, we agreed that we don't want GCP to modify our images for now and that it is OK to expect that the image had user (with set SSH keys) created as part of the image build. So as a result, the code now imports images build through cloudapi always as "data disks".

My motivation to not expose the "os" in the cloudapi was to keep it minimal for now, so that we don't need to worry about backward compatibility and API stability if we decide to remove it or not use it. But other than that, it could be exposed if API stability is not really important for now...

internal/cloudapi/server.go
thozza: The problem from my point of view is that the image name must be unique within the GCP Project that we will use on our side for importing and sharing the image. Since the user won't see what image names we already have within the project, they could easily choose an image name that already exists. The unfortunate thing is that the workflow will fail quite late while importing the already uploaded image. So there is currently no real feedback when creating the compose request if the chosen image name already exists.

Nevertheless the internal GCP API could be extended with a function which would check if an image with the given name already exists within the GCP Project. But there could be race conditions (especially when more users use the API simultaneously) and also in order to expose this in cloudapi, the check would probably need to be done by a worker, which has GCP credentials.

So since I don't see any ideal solution here, I'll incline to whatever you think will be OK for the user...

teg: We could leave this up to the caller (`image-builder` and make them responsible for avoiding clashes), but they'd basically be left with the same problem of having to pick a random name, they couldn't actually expose it to the user for the reasons you point out. I'd be ok with keeping it exactly as you have it here, then either `image-builder` would provide a name (appropriately randomized) or we'd create a random one internally.

internal/cloudapi/openapi.yml
teg: Makes sense! Let's discuss how to extend this after the first version is done.

tools/provision.sh
thozza: Yeah, that probably makes sense. Both composer and worker configurations are currently stored in `test/data/composer` directory in git and installed to this directory.

teg: Right, that seems like a mistake. Should have been split all the way.

test/cases/api.sh
thozza: Yes, this should be definitely added eventually. @Gundersanne also said that this should be good enough to start with. My expectation is that this may not be trivial, because GCP currently can not inject SSH keys into the instance which uses our imported image. This may require introducing the GCP image type.

test/README.md
thozza: Thanks for the note. I know about the containers work, but didn't use it much because it didn't work with `podman-compose` last time I tried it. And unfortunately I don't have docker installed, just podman. Nevertheless I definitely see how running the whole setup in separate containers could be beneficial for testing and debugging.

For now I mostly copied the wording from the composer-cli testing section which talked about running the test case locally. Personally I found running the cloudapi integration test directly on the GCP instance very useful, but it would probably run just OK in containers on the instance.

I may get back to the containers work and try to make it work with `podman-compose`.

thozza:
> Seems we need some creds in Jenkins though. Do we have some we can use, or are we blocked on this?

I already added credentials to the Jenkins instance. I used the `cockpituous` project for now (CC @larskarlitski to expect some bills :smile: ). The error is actually a bug in the code, I'll fix it. The thing is that I currently made Jenkins to inject the credentials only for RHEL integration tests, that's why I used condition in the `provision.sh`.

test/cases/api.sh
teg: Makes sense.

internal/cloudapi/server.go
thozza: @Gundersanne do you have any preference here? I basically did it the same way as AWS part does it. If the image name (snapshot name for AWS) was provided by the caller, just use it, otherwise use `composer-api-<UUID>`...

Gundersanne: `fmt.Sprintf("composer-api-%s", uuid.New().String())` is fine for now I think. It's useful to have the default be some unique thing which is useful during testing.

ondrejbudai (commented):
Looks great! Found two nits and one bigger thing for a discussion.

internal/cloudapi/server.go
thozza: OK, thanks for all the comments. I'll just remove the "TODO" comment.

cmd/osbuild-upload-gcp/main.go
```diff
@@ -0,0 +1,92 @@
...
+	var credentials []byte
+	if credentialsPath != "" {
+		var err error
+		credentials, err = ioutil.ReadFile(credentialsPath)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error while reading credentials: %s\n", err)
+			return
+		}
+	} else {
+		credentials = nil
```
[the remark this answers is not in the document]
thozza: True, thanks!

internal/upload/gcp/gcp.go
```diff
@@ -0,0 +1,417 @@
...
+	fmt.Printf("[GCP] 🤔 Image import build status: %+v\n", imageBuild.Status)
+
+	getBuldReq := &cloudbuildpb.GetBuildRequest{
+		ProjectId: imageBuild.ProjectId,
+		Id:        imageBuild.Id,
+	}
+
+	// Wait for the build to finish
+	fmt.Println("[GCP] 🥱 Waiting for the image import to finish")
+	for {
```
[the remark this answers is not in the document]
teg: In general I think we should allow things to be cancelled by the caller. Not a fan of timeouts, except as a very last resort at the highest level (usually I'd say the user getting bored and clicking "cancel" is timeout good enough).
Might want to do the cancel logic in a follow-up though, seems like more invasive surgery would be needed.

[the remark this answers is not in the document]
ondrejbudai: Agreed!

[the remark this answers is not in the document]
thozza: Yeah, we probably should. I'm open to suggestions. The GCP build job has a timeout set to 7200 sec (2 hours) [on line 190](https://github.com/osbuild/osbuild-composer/pull/1245/files/cc4759b408a16325f4aeef7b9cc3efb4ef1714e2#diff-fa36b8cedd24680d43cde61fbdcd8681c6ccf4ebc452e394c34e692d575af6d3R190), so it will eventually fail. I could either expose this timeout or pass a context as you suggested. The magic value of 7200 sec is from GCP API example...

I didn't do it for the first iteration, because none of the internal cloud provider APIs allow to set timeout while they are still blocking.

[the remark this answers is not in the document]
teg: I'd keep it like this for now. Then go through in a follow-up and make sure all of  this (for all the providers) is cancelable.

tools/provision.sh
thozza: This is fixed now.

cmd/osbuild-upload-gcp/main.go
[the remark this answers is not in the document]
thozza: This is fixed now.

internal/upload/gcp/gcp.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	return nil
+}
+
+// Import imports a previously uploaded image by submitting a Cloud Build API
+// job. The job builds an image into Compute Node from an image uploaded to the
+// storage.
+//
+// The source image file is deleted from the storage bucket after a successfull
```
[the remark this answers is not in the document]
thozza: This is fixed now.

test/cases/api.sh
```diff
@@ -198,73 +335,106 @@ do
...
 
-SHARED_ID=$(jq -r '.LaunchPermissions[0].UserId' "$WORKDIR/ami-attributes.json")
-if [ "$AWS_API_TEST_SHARE_ACCOUNT" != "$SHARED_ID" ]; then
+  SHARED_ID=$(jq -r '.LaunchPermissions[0].UserId' "$WORKDIR/ami-attributes.json")
+  if [ "$AWS_API_TEST_SHARE_ACCOUNT" != "$SHARED_ID" ]; then
     SHARE_OK=0
-fi
-
-# Create key-pair
-$AWS_CMD ec2 create-key-pair --key-name "key-for-$AMI_IMAGE_ID" --query 'KeyMaterial' --output text > keypair.pem
```
ondrejbudai, line 239: I think you removed this step causing the CI to fail.

internal/cloudapi/openapi.yml
```diff
@@ -243,6 +253,36 @@ components:
           example: ['123456789012']
           items:
             type: string
+    GCPUploadRequestOptions:
+      type: object
+      required:
+        - bucket
```
ondrejbudai, line 259: Just something that I've encountered right now: Can you document what happens if you don't pass the individual optional values?

test/cases/api.sh
thozza: Yeah, not sure how that happened. Thanks Ondrej!

schutzbot/Jenkinsfile
```diff
@@ -11,6 +11,8 @@ pipeline {
         AWS_REGION = "us-east-2"
         AWS_BUCKET = "imagebuilder-jenkins-testing-use2"
         BUILD_CAUSE = detect_build_cause()
+        GCP_BUCKET = "osbuild-composer-testing"
```
ondrejbudai, line 14: I wonder if we should move these two variables in `test/cases/api.sh`. IMHO, it would be more convenient to have the tests as much self-contained as possible and don't put additional login in Jenkins.

thozza: Personally I don't think that we should. The reason I see is that the value is specific to the testing environment used for running the test case. This bucket is not auto-created by the test, therefore the test case script would be not usable out of the box if you would not have access to the same project that we use for CI. Also Bucket name must be unique globally (among all GCP projects), so user can not create it in their own project.

internal/cloudapi/openapi.yml
thozza: I added description to each value with explanation of what will happen.

test/cases/api.sh
thozza: This is fixed now...

msehnout:
Should we also extend cloud-cleaner with support for GCP? It is supposed to clean up any artifacts left behind in case of test failure so it the new test cases upload anything to GCP, we probably should.

thozza:
> Should we also extend cloud-cleaner with support for GCP? It is supposed to clean up any artifacts left behind in case of test failure so it the new test cases upload anything to GCP, we probably should.

That's a fair point. There are currently code paths when the uploaded image can stay in the storage if something failed. I would rather refactor the internal API code a bit to ensure that the proper cleanup is done in case of an error when using the API. The service should not leave behind any leftover files in case of fail in production, where `cloud-cleaner` would not be run. There may be a failure with GCP when the upload succeeds, but image import fails because an image with chosen name already exists.

I'm happy to do it as part of this PR or create an issue and follow up on this...

teg:
Good point @msehnout, I think landing this first and doing a follow-up is ok though.

msehnout:
> > Should we also extend cloud-cleaner with support for GCP? It is supposed to clean up any artifacts left behind in case of test failure so it the new test cases upload anything to GCP, we probably should.
> 
> That's a fair point. There are currently code paths when the uploaded image can stay in the storage if something failed. I would rather refactor the internal API code a bit to ensure that the proper cleanup is done in case of an error when using the API. The service should not leave behind any leftover files in case of fail in production, where `cloud-cleaner` would not be run. There may be a failure with GCP when the upload succeeds, but image import fails because an image with chosen name already exists.
> 

The cloud cleaner is targeted at cases where Jenkins kills the job for some reason so any cleanup code path is useless. That's why it runs in the
```
post {
                        always {
```
section.

(Ideally it would also run as part of the cron job, but noone has written the code for querying images and removing anything older than whatever timeout we decide to use)

> I'm happy to do it as part of this PR or create an issue and follow up on this...

As Tom said, follow-up is fine :-)

internal/cloudapi/openapi.yml
```diff
@@ -243,6 +253,60 @@ components:
...
+            Each string must contain a specifier of the account type. Valid formats are:
+              - 'user:{emailid}': An email address that represents a specific
+                Google account. For example, 'alice@example.com'.
+              - 'serviceAccount:{emailid}': An email address that represents a
+                service account. For example, 'my-other-app@appspot.gserviceaccount.com'.
+              - 'group:{emailid}': An email address that represents a Google group.
+                For example, 'admins@example.com'.
+              - 'domain:{domain}': The G Suite domain (primary) that represents all
+                the users of that domain. For example, 'google.com' or 'example.com'.
+
```
Gundersanne, line 305: nitpick: why the newline here?

I'm not opposed to introducing newlines to make it easier to read but per `schemas` entry then

thozza: I didn't want the case "if not specified" to blend with the list. So just a personal preference. Happy to remove it if you think that it should not be there :wink:

thozza: I removed the extra newline...

tools/provision.sh
```diff
@@ -12,9 +12,23 @@ if [[ $ID == rhel || $ID == centos ]]; then
...
+sudo mkdir -p /etc/osbuild-worker
+sudo cp -a /usr/share/tests/osbuild-composer/worker/osbuild-worker.toml \
+    /etc/osbuild-worker/
+
+# if GCP credentials are defined in the ENV, add them to the worker's configuration
+GOOGLE_APPLICATION_CREDENTIALS="${GOOGLE_APPLICATION_CREDENTIALS:-}"
+if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
+    # The credentials file must be copied to a different location. Jenkins places
+    # it into /tmp and as a restult, the worker would not see it due to using PrivateTmp=true.
+    GCP_CREDS_WORKER_PATH="/etc/osbuild-worker/$(basename "$GOOGLE_APPLICATION_CREDENTIALS")"
```
ondrejbudai, line 27: We might even use a static filename.

thozza: We can. I probably don't have a strong preference.

ondrejbudai: Me neither, I'm fine with this. :)

thozza: I changed it :smile:

teg (approved):
Amazing work! Thanks so much for this!

thozza:
> I still need to verify that the changes don't affect `cockpit-composer`, but feel free to review in the meantime. Thanks!

Since no new image type has been added for GCP as well as no Weldr API support, there was no effect on `cockpit-composer` or `composer-cli` from what I tested. In case there are no outstanding issues that would be blocking this PR and CI passes, this change should be OK to merge. Thanks to all for a great feedback and discussion!

ondrejbudai (approved):
Thanks!

Amazing!

PR #1250, opened 2021-02-24
schutzbot: Work around mock RHBZ#1931452 on F34

thozza:
Mockbuild using systemd-nspawn currently fails on Fedora 34. The
workaround is to use "simple" isolation method - the traditional
chroot() call.

Reported as: https://bugzilla.redhat.com/show_bug.cgi?id=1931452

This pull request includes:

  - This is a workaround for distribution issue on the CI system. No new tests were added.
  - Not a user-visible change, therefore no documentation was added.

### DIFF
--- a/schutzbot/mockbuild.sh
+++ b/schutzbot/mockbuild.sh
@@ -78,12 +78,19 @@ if [[ $VERSION_ID == 8.4 ]]; then
     echo '"""' | sudo tee -a /etc/mock/templates/rhel-8.tpl
 fi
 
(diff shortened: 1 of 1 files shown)


ondrejbudai:
This might be an issue in `dbus-broker-26-2`. I updated the CI images so we have the newest `dbus-broker-27-2` that should fix this issue. Let's see if that helps, otherwise, we might indeed need this.

ondrejbudai (approved):
Sigh, this issue still exists. Let's merge this so it doesn't ruin CI runs.

PR #1254, opened 2021-02-25
test/openstack: increase timeout for the instance to become ACTIVE

ondrejbudai:
The openstack boot test often ruins our days with:

```
Waiting for instance 63ac19be-2e19-44e2-8bef-9770d68a190c to become Active failed: A timeout occurred
```

I decided to investigate. It turns out the first boot of an image can take
up to 18 minutes. The subsequent ones are usually much faster (but don't rely
on this fact, I saw 15 minutes there).

This commit bumps the timeout to 30 minutes. This should be plenty of time
for the instance to spin up and get into the ACTIVE state.

Honestly, I'm not very happy with the solution but it should help with the
failing Schutzbot. As a follow-up, I will reach to the PSI OpenStack team
and ask them if we could somehow speed up the process (maybe by using another
flavor, ci.m1.medium.ephemeral just might be slow for some reason, I don't
know).

Anyway, this should help us in the short term because I strongly believe that
a slow test is still better than a failing one.

This pull request includes:

just test adjustements.

### DIFF
--- a/internal/boot/openstacktest/openstack.go
+++ b/internal/boot/openstacktest/openstack.go
@@ -8,12 +8,12 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
-	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
-	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
 
-const WaitTimeout = 600 // in seconds
+const WaitTimeout = 30 * 60 // 30 minutes in seconds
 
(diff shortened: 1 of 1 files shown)


internal/boot/openstacktest/openstack.go
```diff
@@ -8,12 +8,12 @@ import (
...
 	"github.com/gophercloud/gophercloud/openstack"
-	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
-	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
 
-const WaitTimeout = 600 // in seconds
+const WaitTimeout = 20 * 60 // 20 minutes in seconds
```
teg, line 16: Did you intend to make this 30?

ondrejbudai: good catch, thanks!

ondrejbudai:
The test suite passed on the first attempt. I fixed the wrong timeout (20 minutes instead of 30) and force-pushed again. Let's see if we will have green Schutzi again.

ondrejbudai:
second pass and also :green_heart:

teg (approved):
All the green is disconcerting, but ok.

PR #1255, opened 2021-02-26
Add support for uploading an Azure image in cloud API

ondrejbudai:
Testing the CI. Will write more later.

This pull request includes:


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -282,6 +300,112 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			// imageName := t.ImageName
 			// projectID := gcp.GetProjectID()
 
+		case *target.AzureImageTargetOptions:
+			ctx := context.Background()
+
+			var authorizer autorest.Authorizer
+			if impl.AzureCreds == nil {
+				authorizer, err = auth.NewAuthorizerFromEnvironment()
```
teg, line 308: The idea is that this would use an Azure Managed Identity if no environment variables are set, right?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -282,6 +300,112 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+					ImageName:     blobName,
+				},
+				path.Join(outputDirectory, options.Filename),
+				azure.DefaultUploadThreads,
+			)
+			if err != nil {
+				return fmt.Errorf("uploading the image failed: %v", err)
+			}
+
+			err = azure.RegisterImage(
```
teg, line 395: For the future: we probably want to have some idea of how long this takes so if it is worth splitting out like what we want to do for AWS. Some logging around the call would do it, I guess.

cmd/osbuild-worker/main.go
```diff
@@ -154,12 +157,21 @@ func main() {
...
+		}
+	}
+
 	jobImpls := map[string]JobImplementation{
 		"osbuild": &OSBuildJobImpl{
 			Store:        store,
 			Output:       output,
 			KojiServers:  kojiServers,
 			GCPCredsPath: config.GCP.Credentials,
+			AzureCreds:   azureCredentials,
```
teg, line 174: Any reason this is different from GCP? The Azure config is a pointer to a struct, while GCP is not, and the Azure config is parsed in `main()` whilst GCP is passed on as just the path to the creds.

test/cases/api.sh
```diff
@@ -57,13 +61,21 @@ function checkEnvGCP() {
...
+}
+
 case $CLOUD_PROVIDER in
   "$CLOUD_PROVIDER_AWS")
     checkEnvAWS
     ;;
   "$CLOUD_PROVIDER_GCP")
     checkEnvGCP
     ;;
+  "$CLOUD_PROVIDER_GCP")
```
thozza, line 76: Copy&Paste error, should be `"$CLOUD_PROVIDER_AZURE"`

cmd/osbuild-worker/jobimpl-osbuild.go
ondrejbudai: I've decided to remove this feature for now. With `auth.NewAuthorizerFromEnvironment()`, we cannot specify the tenant ID per job. This should surely be possible but needs more research and it's actually not needed for the MVP.

cmd/osbuild-worker/jobimpl-osbuild.go
ondrejbudai: Logging added, fyi it takes about 1 minute to register the image.

cmd/osbuild-worker/main.go
ondrejbudai: I added a comment to this: Basically, the fileformat for Azure credentials is created by us and I want to make sure to fail as soon as possible if it's malformed. We probably want to do something similar for GCP too. If we have wrong credentials, we would like to fail as soon as possible and not wait for the first job.

test/cases/api.sh
ondrejbudai: good catch, thanks!

ondrejbudai:
Pushed a new version. I cleaned up the whole PR, added commit messages, cleaned up the `internal/azure/upload` directory and changed my mind on some decisions (see discussion).

The target result for Azure is not implemented here to prevent conflicts with #1262

teg (dismissed):

teg:
Very nice!

ondrejbudai:
Had to force-push to fix a mockbuild failure.

cmd/osbuild-worker/main.go
thozza: Yeah, sounds reasonable to do also for GCP, although it would be just `[]byte`, not our own structure... For GCP, not providing the credentials can still work if the worker is not running on GCP instance with proper roles associated with it. So to exclude the possibility to fail late, we would need to try to authenticate when the worker is starting up (which we discussed with @ondrejbudai). But then again it would probably make sense to be able to configure supported features in worker's configuration, because now every worker would have to be able to authenticate with GCP.

internal/cloudapi/server.go
```diff
@@ -205,7 +205,7 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 			}
 
 			targets = append(targets, t)
-		} else if uploadRequest.Type == "gcp" {
+		} else if uploadRequest.Type == UploadTypes_gcp {
```
thozza, line 208: :heart:

internal/target/target.go
```diff
@@ -71,6 +71,8 @@ func UnmarshalTargetOptions(targetName string, rawOptions json.RawMessage) (Targ
 		options = new(AWSTargetOptions)
 	case "org.osbuild.gcp":
 		options = new(GCPTargetOptions)
+	case "org.osbuild.azure.image":
```
thozza, line 74: Wouldn't it make sense to have this case right under `org.osbuild.azure`? Not really a big issue...

schutzbot/Jenkinsfile
```diff
@@ -13,6 +13,10 @@ pipeline {
         BUILD_CAUSE = detect_build_cause()
         GCP_BUCKET = "osbuild-composer-testing"
         GCP_REGION = "us-east4"
+        AZURE_TENANT_ID = "1710d22c-ccf0-4421-8ba7-0135cfaecb90"
+        AZURE_SUBSCRIPTION_ID = "8d026bb1-2a65-454d-a88f-c896db94c4f8"
```
thozza, line 17: Are any of these IDs secret (not any more anyway)? Just wondering whether someone could use them to upload to our account...

teg (approved):
Still great!

internal/target/target.go
ondrejbudai: Good point, will move it in some kind of a follow-up.

schutzbot/Jenkinsfile
ondrejbudai: Without a client ID and secret, these are just random UUIDs. https://stackoverflow.com/questions/45661109/are-azure-subscription-id-aad-tenant-id-and-aad-app-client-id-considered-secre

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -290,9 +297,122 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+				r = append(r, errors.New("osbuild job has org.osbuild.azure.image target but this worker doesn't have azure credentials"))
+				continue
+			}
+
+			c, err := azure.NewClient(*impl.AzureCreds, options.TenantID)
+			if err != nil {
+				r = append(r, err)
+				continue
+			}
+			log.Print("[Azure] 🔑 Logged in Azure")
```
thozza, line 316: Not an issue, just a comment probably for further discussion. This kind of logging is done in GCP's and AWS's case in the `internal/upload/...` implementation, not in worker's job implementation. In case of GCP, I followed what was already implemented for AWS, but I can see that Azure never had any logging in the internal upload implementation. We should probably decide what we want and should make it consistent for all types of upload targets.

internal/upload/azure/azurestorage.go
```diff
@@ -0,0 +1,203 @@
...
+	// Check properties, specifically MD5 sum of the blob
+	props, err := blobURL.GetProperties(ctx, azblob.BlobAccessConditions{})
+	if err != nil {
+		return fmt.Errorf("getting the properties of the new blob failed: %v", err)
+	}
+	var blobChecksum []byte = props.ContentMD5()
+	var fileChecksum []byte = imageFileHash.Sum(nil)
+
+	if !bytes.Equal(blobChecksum, fileChecksum) {
+		return errors.New("error during image upload. the image seems to be corrupted")
```
thozza, line 171: I implemented a similar check for GCP, but I'm thinking that maybe the uploaded file should be deleted from the storage if the checksum check fails. Not an issue for this set of changes, but I would like to get your point of view anyway.

internal/upload/azure/azure.go
```diff
@@ -1,158 +1,149 @@
...
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("sending the create storage account request failed: %v", err)
 	}
-	// Move the cursor back to the start of the imageFile
-	if _, err := imageFile.Seek(0, 0); err != nil {
-		return fmt.Errorf("cannot seek the image: %v", err)
+
+	err = result.WaitForCompletionRef(context.Background(), c.Client)
```
thozza, line 82: Shouldn't the method's `ctx` argument be used here as well?

thozza (commented):
Very nice work!

test/cases/api.sh
```diff
@@ -102,6 +114,20 @@ function cleanupGCP() {
   fi
 }
 
+function cleanupAzure() {
+  # since this function can be called at any time, ensure that we don't expand unbound variables
+  AZURE_CMD="${AZURE_CMD:-}"
+  if [ -n "$AZURE_CMD" ]; then
+    set +e
+    $AZURE_CMD image delete --resource-group sharing-research --name "$AZURE_IMAGE_NAME"
```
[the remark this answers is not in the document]
ondrejbudai: Fixed, I'm now skipping the clean-up when AZURE_IMAGE_NAME is empty.

test/cases/api.sh
```diff
@@ -162,13 +191,32 @@ EOM
   $GCP_CMD --version
 }
 
+function installClientAzure() {
+  if ! hash az; then
+    sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
+    echo -e "[azure-cli]
```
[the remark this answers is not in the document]
ondrejbudai: I've just copy-pasted this from the official docs... I added a link to the docs to the script and left `-e` to prevent confusion as this is just a copy. :)

test/cases/api.sh
```diff
@@ -296,13 +344,45 @@ function createReqFileGCP() {
...
+  ]
+}
+EOF
+}
+
 case $CLOUD_PROVIDER in
   "$CLOUD_PROVIDER_AWS")
     createReqFileAWS
-    ;;
+  ;;
```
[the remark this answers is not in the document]
ondrejbudai: Fixed! Thanks!

test/cases/api.sh
```diff
@@ -469,12 +549,30 @@ function verifyInGCP() {
   ssh -oStrictHostKeyChecking=no -i "$GCP_SSH_KEY" "$SSH_USER"@"$HOST" rpm -q postgresql
 }
 
+# Verify image in Azure
+function verifyInAzure() {
+  set +x
+  $AZURE_CMD login --service-principal --username "${AZURE_CLIENT_ID}" --password "${AZURE_CLIENT_SECRET}" --tenant "${AZURE_TENANT_ID}"
```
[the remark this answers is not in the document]
ondrejbudai: `set +x` is used to prevent printing of credentials.

tools/provision.sh
```diff
@@ -23,12 +23,24 @@ sudo cp -a /usr/share/tests/osbuild-composer/worker/osbuild-worker.toml \
...
     sudo cp "$GOOGLE_APPLICATION_CREDENTIALS" "$GCP_CREDS_WORKER_PATH"
     echo -e "\n[gcp]\ncredentials = \"$GCP_CREDS_WORKER_PATH\"\n" | sudo tee -a /etc/osbuild-worker/osbuild-worker.toml
 fi
 
+# if Azure credentials are defined in the env, create the credentials file
+AZURE_CLIENT_ID="${AZURE_CLIENT_ID:-}"
+AZURE_CLIENT_SECRET="${AZURE_CLIENT_SECRET:-}"
+if [[ -n "$AZURE_CLIENT_ID" && -n "$AZURE_CLIENT_SECRET" ]]; then
+    set +x
+    sudo tee /etc/osbuild-worker/azure-credentials.toml > /dev/null << EOF
```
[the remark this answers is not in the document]
ondrejbudai: `set +x` is used to prevent printing of credentials.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -290,9 +297,122 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				continue
 			}
 			// TODO: report back the information below, which is necessary to find and use the image
-			// imageName := t.ImageName
+			// imageName := t.BlobName
```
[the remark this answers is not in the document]
ondrejbudai: Fixed by rebase. :-D

cmd/osbuild-worker/jobimpl-osbuild.go
ondrejbudai: Yeah, my thinking is that `internal/upload/...` should serve mainly as a library and thus not log anything... But yeah, we should have a discussion. :)

internal/upload/azure/azurestorage.go
ondrejbudai: Good point, we should open an issue to track it.

internal/upload/azure/azure.go
ondrejbudai: Fixed in the whole `azure.go` file.

ondrejbudai:
pushed a new version. I addressed all comments and implemented a simple TargetResult that returns just the image name. We might want to extend it later, this is just a start. :)

test/cases/api.sh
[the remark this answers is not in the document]
thozza: Yeah, I meant confused `set +x` with `set +e`. Sorry for the noise.

teg (approved):
🥳

ondrejbudai:
Tom's changes looks good to me (as an author I cannot approve).

teg (approved):
Between us, we can approve of each other, I think ;)

PR #1258, opened 2021-03-02
Revert "internal/distro/rhel8: un-exclude subman from edge"

runcom:
We needed this patch to demonstrate RHC capabilities but we don't want to ship such a huge dependencies tree with subman/dnf - for 8.4 we'd revert this and revisit the dep tree.

@nullr0ute @gicmo ptal


teg:
The 8.4 beta freeze is passed, so any changes we merge to master will no longer land in 8.4. If you still need this, feel free to reach out to me directly and I can walk you through the process.

teg:
Based on side-channel discussion: I don't think this is needed any more? Should we close it?

PR #1261, opened 2021-03-03
spec: Improve description of the osbuild-composer package

msehnout:
The previous version referred to lorax-composer as the definition of
what osbuild-composer does. This worked fine while osbuild-composer was
considered an alternative for it. Now that osbuild-composer is the
default one, it should describe what it does without references to
lorax. Furthemore, composer is now able to build OSTree commits as well
as VM images, to the previous description was slightly incomplete.

This commit introduces description which is up-to-date and does not
refer to lorax any more.

This pull request includes:


andreasn (dismissed):
Looks good to me.

thozza:
The change looks good, but personally I'm not sure if calling osbuild-composer a "web service" is correct. It still provides the Weldr API which works over a UNIX socket and it has no frontend of its own. Maybe just call it a "service"?

msehnout:
> The change looks good, but personally I'm not sure if calling osbuild-composer a "web service" is correct. It still provides the Weldr API which works over a UNIX socket and it has no frontend of its own. Maybe just call it a "service"?

Good point, I've considered this and I think the terminology here is either misleading or overloaded. If I used just "service" I'd personally understand it as a systemd service, not necessarily having REST API. But this is just my bias as a long term systemd user, others might disagree. When I call it "web service" one could expect it to be available over TCP port instead of UNIX socket as you said.

I concluded that there is no short, yet precise way of describing it I could think of.

ondrejbudai:
REST service maybe?

thozza:
> Good point, I've considered this and I think the terminology here is either misleading or overloaded. If I used just "service" I'd personally understand it as a systemd service, not necessarily having REST API. But this is just my bias as a long term systemd user, others might disagree. When I call it "web service" one could expect it to be available over TCP port instead of UNIX socket as you said.

In my opinion calling it a service has nothing to do with the fact, that it is started by systemd. I'm not sure what "systemd service" is, other than just the notion of a process controlled or supervised by systemd. Which would be basically any service running on the system. But I understand that you wanted to be more specific in the description... :wink:

msehnout:
Maybe just service is better and my view is biased :-).

This is how I think about it: What do we want to stress in the description:
 * The fact it provides API over HTTP?
 * That we try to make it RESTful?
 * It is running over UDS, but now it can run also over TCP if used with composer api?

Maybe we should not stress any of those technical details and instead focus on the fact there is no client in the package itself.

thozza:
> Maybe we should not stress any of those technical details and instead focus on the fact there is no client in the package itself (=> use only service)
...so "backend service"? :smiley:

achilleas-k:
> Maybe we should not stress any of those technical details and instead focus on the fact there is no client in the package itself (=> use only service)

I agree with this.  No qualifiers leaves it general enough that all instances are covered.

larskarlitski:
Oh nice! The old description is moot now indeed.

Not strong opinion, but I agree that just "service" is enough. If we want to provide more detail, maybe add "providing a HTTP API". I would not call it "REST", because that term means various things to people.

ondrejbudai:
I realized something yesterday: The new description doesn't mention anything about the composer's feature to upload the image directly to multiple cloud providers. As this is becoming the biggest part of the project, we should surely mention it.

msehnout:
> I realized something yesterday: The new description doesn't mention anything about the composer's feature to upload the image directly to multiple cloud providers. As this is becoming the biggest part of the project, we should surely mention it.

ok, I've force-pushed a 2nd proposal. It does not mention any cloud providers specifically because the list of supported platforms have been growing recently.

ondrejbudai (approved):
:heart:

achilleas-k (approved):

teg (approved):

thozza (approved):
Looks great!

PR #1271, opened 2021-03-08
Don't check branch name when deciding if build was started by cron

atodorov:
because we already have cron jobs on multiple branches!

This pull request includes:


atodorov:
The effect if this patch is not applied is that `telegramSend` doesn't send messages for jobs on the 8.4 branch (which are now passing) because the build cause is returned as empty string.

teg (approved):
Makes sense to me!

ondrejbudai (approved):

PR #1279, opened 2021-03-09
Distro: replace Packages() / BuildPackages() with PackageSets()

teg:
This is an alternative to calling depsolve from within the distros. @achilleas-k first suggested this approach, but we decided to go with calling depsolve directly to make the change simpler. I really don't see how to hook that up with the manifest tests we currently have, and I think skipping those would be a shame.

Thoughts?

### DIFF
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -80,12 +80,12 @@ func TestCrossArchDepsolve(t *testing.T) {
 							imgType, err := arch.GetImageType(imgTypeStr)
 							require.NoError(t, err)
 
-							buildPackages := imgType.BuildPackages()
-							_, _, err = rpm.Depsolve(buildPackages, []string{}, repos[archStr], distroStruct.ModulePlatformID(), archStr)
+							packages := imgType.PackageSets(blueprint.Blueprint{})
+
+							_, _, err = rpm.Depsolve(packages["build-packages"], repos[archStr], distroStruct.ModulePlatformID(), archStr)
 							assert.NoError(t, err)
 
-							basePackagesInclude, basePackagesExclude := imgType.Packages(blueprint.Blueprint{})
-							_, _, err = rpm.Depsolve(basePackagesInclude, basePackagesExclude, repos[archStr], distroStruct.ModulePlatformID(), archStr)
+							_, _, err = rpm.Depsolve(packages["packages"], repos[archStr], distroStruct.ModulePlatformID(), archStr)
 							assert.NoError(t, err)
 						})
 					}
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -72,17 +72,16 @@ type ImageType interface {
 	// is 0 the default value for the format will be returned.
 	Size(size uint64) uint64
 
-	// Returns the default packages to include and exclude when making the image
-	// type.
-	Packages(bp blueprint.Blueprint) ([]string, []string)
-
-	// Returns the build packages for the output type.
-	BuildPackages() []string
+	// Returns the sets of packages to include and exclude when building the image.
+	// Indexed by a string label. How each set is labeled and used depends on the
+	// image type.
+	PackageSets(bp blueprint.Blueprint) map[string]rpmmd.PackageSet
 
 	// Returns an osbuild manifest, containing the sources and pipeline necessary
 	// to build an image, given output format with all packages and customizations
-	// specified in the given blueprint.
-	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecs, buildPackageSpecs []rpmmd.PackageSpec, seed int64) (Manifest, error)
+	// specified in the given blueprint. The packageSpecSets must be labelled in
+	// the same way as the originating PackageSets.
+	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, seed int64) (Manifest, error)
 }
 
 // The ImageOptions specify options for a specific image build
--- a/internal/distro/fedora32/distro.go
+++ b/internal/distro/fedora32/distro.go
@@ -187,20 +187,32 @@ func (t *imageType) BuildPackages() []string {
 	return packages
 }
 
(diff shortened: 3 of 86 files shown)


internal/distro/fedora32/distro.go
```diff
@@ -190,9 +190,14 @@ func (t *imageType) BuildPackages() []string {
 func (t *imageType) Manifest(c *blueprint.Customizations,
 	options distro.ImageOptions,
 	repos []rpmmd.RepoConfig,
-	packageSpecs,
-	buildPackageSpecs []rpmmd.PackageSpec,
+	packageSpecSets [][]rpmmd.PackageSpec,
 	seed int64) (distro.Manifest, error) {
+	var packageSpecs []rpmmd.PackageSpec
+	var buildPackageSpecs []rpmmd.PackageSpec
+	if len(packageSpecSets) == 2 {
```
larskarlitski, line 210: This seems brittle. Is is so that you can pass `nil` from tests?

internal/distro/distro.go
```diff
@@ -82,7 +82,7 @@ type ImageType interface {
 	// Returns an osbuild manifest, containing the sources and pipeline necessary
 	// to build an image, given output format with all packages and customizations
 	// specified in the given blueprint.
-	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecs, buildPackageSpecs []rpmmd.PackageSpec, seed int64) (Manifest, error)
+	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets [][]rpmmd.PackageSpec, seed int64) (Manifest, error)
```
larskarlitski, line 90: I wonder if a `map[string][]rpmmd.PackageSpec` is preferable here, with string like "build" and "base". Not sure if strings as indices are less error-prone, but it might make it easier to debug?

larskarlitski (commented):
I *think* I like this as an alternative, because it's a tiny bit less hacky than calling `Depsolve()` from within the distros. If it solves the CI problem, I'm all for it.

internal/distro/distro.go
achilleas-k: I agree.  My current version of the open PR (not pushed yet) uses a map and it's much cleaner.  The names/keys don't need to have any real meaning (e.g., they don't need to match pipeline or stage names), but it's easier to work with and more readable.

teg: good point, will change

teg: Done.

internal/distro/fedora32/distro.go
teg: It did.

achilleas-k (dismissed):
I like this a lot.  Much cleaner than depsolving in `Manifest()` and definitely the better way forward.

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -80,12 +80,12 @@ func TestCrossArchDepsolve(t *testing.T) {
...
-							buildPackages := imgType.BuildPackages()
-							_, _, err = rpm.Depsolve(buildPackages, []string{}, repos[archStr], distroStruct.ModulePlatformID(), archStr)
+							packages := imgType.PackageSets(blueprint.Blueprint{})
+
+							_, _, err = rpm.Depsolve(packages["build-packages"].Include, packages["build-packages"].Exclude, repos[archStr], distroStruct.ModulePlatformID(), archStr)
 							assert.NoError(t, err)
 
-							basePackagesInclude, basePackagesExclude := imgType.Packages(blueprint.Blueprint{})
-							_, _, err = rpm.Depsolve(basePackagesInclude, basePackagesExclude, repos[archStr], distroStruct.ModulePlatformID(), archStr)
+							_, _, err = rpm.Depsolve(packages["packages"].Include, packages["build-packages"].Exclude, repos[archStr], distroStruct.ModulePlatformID(), archStr)
```
ondrejbudai, line 88: `packages["build-packages"].Exclude` is wrong on this line.

Maybe let's move `PackageSet` into `rpmmd` and make `rpmmd.Depsolve` taking `PackageSet` as an argument? This change would prevent this kind of mistakes (it happened in past, see #1148)
Maybe `rpm.Depsolve` should take `distro.PackageSet` as an argument?

internal/weldr/api.go
```diff
@@ -2738,32 +2750,13 @@ func (api *API) allRepositories() []rpmmd.RepoConfig {
...
-	specs := bp.GetPackages()
-	excludeSpecs := []string{}
-	if imageType != nil {
-		// When the output type is known, include the base packages in the depsolve
-		// transaction.
-		specs, excludeSpecs = imageType.Packages(*bp)
-	}
-
-	packages, _, err := api.rpmmd.Depsolve(specs, excludeSpecs, repos, api.distro.ModulePlatformID(), api.arch.Name())
+func (api *API) depsolveBlueprint(bp *blueprint.Blueprint) ([]rpmmd.PackageSpec, error) {
```
ondrejbudai, line 2753: Nice cleanup!

ondrejbudai (changes requested):
> I really don't see how to hook that up with the manifest tests we currently have, and I think skipping those would be a shame.

I'm confused, everything seems to work alright. Did I miss anything?

cmd/osbuild-dnf-json-tests/main_test.go
teg: Good catch, and good suggestion! Done.

teg:
> > I really don't see how to hook that up with the manifest tests we currently have, and I think skipping those would be a shame.
> 
> I'm confused, everything seems to work alright. Did I miss anything?

The current tests are ok, but in #1244 the plan was to call `Depsolve()` from within `Manifest()` to be able to deal with more than the two standard package sets. If we do that, we would not be able to test the new image types in the same way as the current image tests (as they depend on pinning the rpmmd depsolve results).

ondrejbudai:
Do you have any plans for dropping the now obsoleted `Packages()` and `BuildPackages()` from the actual distro implementations?

teg:
> Do you have any plans for dropping the now obsoleted `Packages()` and `BuildPackages()` methods from the actual distro implementations?

Ah, yeah, they should be made private at least (and tests adjusted). I'd do that in a follow-up though.

ondrejbudai (approved):
LGTM, very nice changes, thanks!

Issue #1276, opened 2021-03-08
osbuild-worker jobimpl-osbuild contains duplicate code and as a result osbuild-worker does not send compose status update to osbuild-composer


msehnout:
https://github.com/osbuild/osbuild-composer/blob/167fdc44bc82744b759a059b197d4415523e83da/cmd/osbuild-worker/jobimpl-osbuild.go#L101

Almost the same piece of code lives in the Run function twice. This is the first one:
```go
	if osbuildOutput.Success && args.ImageName != "" {
		var f *os.File
		imagePath := path.Join(outputDirectory, args.ImageName)
		if args.StreamOptimized {
			f, err = vmware.OpenAsStreamOptimizedVmdk(imagePath)
			if err != nil {
				return err
			}
			streamOptimizedPath = f.Name()
		} else {
			f, err = os.Open(imagePath)
			if err != nil {
				return err
			}
		}
		err = job.UploadArtifact(args.ImageName, f)
		if err != nil {
			return err
		}
	}
```
and this is the second one:
```go
			if !osbuildOutput.Success {
				continue
			}
			var f *os.File
			imagePath := path.Join(outputDirectory, options.Filename)
			if options.StreamOptimized {
				f, err = vmware.OpenAsStreamOptimizedVmdk(imagePath)
				if err != nil {
					r = append(r, err)
					continue
				}
			} else {
				f, err = os.Open(imagePath)
				if err != nil {
					r = append(r, err)
					continue
				}
			}
```

Now the problem is that the first one contains the `return` statement which makes the function end without calling `job.Update` at the end:
https://github.com/osbuild/osbuild-composer/blob/167fdc44bc82744b759a059b197d4415523e83da/cmd/osbuild-worker/jobimpl-osbuild.go#L545

Now this causes osbuild-worker to correctly log the fact that a compose failed:
```
Mar 08 09:39:00 vm osbuild-worker[26301]: Running 'osbuild' job 24b774cf-4ae0-49b2-8f35-cdcd17c83763
Mar 08 09:43:52 vm osbuild-worker[26301]: 2021/03/08 09:43:52 Job 24b774cf-4ae0-49b2-8f35-cdcd17c83763 failed: open /var/cache/osbuild-worker/output/24b774cf-4ae0-49b2-8f35-cdcd17c83763-803191209/commit.tar: no such file or directory
Mar 08 09:43:52 vm osbuild-worker[26301]: Waiting for a new job...
```
[...]

Issue #1270, opened 2021-03-08
Return clear error message when out of disk space


yih-redhat:
If the image builder machine does not have enough disk space, the compose command will fail, and in the log it only reports "non-zero exit status 233", suggest to return clear error message to tell user the fail reason is out of disk space.

Env:
RHEL8.4
osbuild-composer-28-1.el8.x86_64

Reproduce steps:
1. Run command: composer-cli compose start ostree rhel-edge-commit
2. The command will fail soon, check log:

Output:
[/usr/lib/tmpfiles.d/httpd.conf:1] Unknown group 'apache'.
[/usr/lib/tmpfiles.d/httpd.conf:2] Unknown user 'apache'.
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
[/usr/lib/tmpfiles.d/radvd.conf:1] Unknown user 'radvd'.
[/usr/lib/tmpfiles.d/rpcbind.conf:2] Unknown user 'rpc'.
[/usr/lib/tmpfiles.d/setroubleshoot.conf:1] Unknown user 'setroubleshoot'.
Failed to create file /sys/fs/selinux/checkreqprot: Read-only file system
Preparing packages...
  installing package selinux-policy-targeted-3.14.3-65.el8.noarch needs 62MB on the / filesystem
  installing package libsecret-0.18.6-1.el8.x86_64 needs 63MB on the / filesystem
  installing package pinentry-1.1.0-2.el8.x86_64 needs 63MB on the / filesystem
  installing package grub2-tools-extra-1:2.02-97.el8.x86_64 needs 70MB on the / filesystem
  installing package rpm-plugin-systemd-inhibit-4.14.3-13.el8.x86_64 needs 70MB on the / filesystem
  installing package tpm2-tss-2.3.2-3.el8.x86_64 needs 71MB on the / filesystem
  installing package ima-evm-utils-1.3.2-12.el8.x86_64 needs 72MB on the / filesystem
  installing package libusbx-1.0.23-4.el8.x86_64 needs 72MB on the / filesystem
  installing package gnupg2-smime-2.2.20-2.el8.x86_64 needs 73MB on the / filesystem
  installing package gnupg2-2.2.20-2.el8.x86_64 needs 83MB on the / filesystem
  installing package gpgme-1.13.1-7.el8.x86_64 needs 84MB on the / filesystem
[...]

PR #1281, opened 2021-03-10
osbuild-worker: Report job status back to worker API in all cases

msehnout:
The previous implementation exited before reporting back to the worker
API in few branches. This left the compose status in RUNNING state even
though the worker did not work of the job any more. Refactoring the
API call into the `deref` part makes sure it gets called every time.

Fixes: #1276

This pull request includes:

  - This functionality is already covered by tests. (well not the cases of failure, but we don't have any test cases like that)
- [X] adequate documentation informing people about the change
  - Ideally this change does not affect anyone.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -63,36 +63,27 @@ func osbuildStagesToRPMs(stages []osbuild.StageResult) []koji.RPM {
 	return rpms
 }
 
+func appendTargetError(res *worker.OSBuildJobResult, err error) {
+	errStr := err.Error()
+	fmt.Printf("target errored: %s", errStr)
+	res.TargetErrors = append(res.TargetErrors, errStr)
+}
+
 func (impl *OSBuildJobImpl) Run(job worker.Job) error {
-	// Initialize variables needed for reporting back to osbuild-composer
-	var outputDirectory string
-	var r []error
-	var targetResults []*target.TargetResult
-	var osbuildOutput *osbuild.Result = &osbuild.Result{
+	// Initialize variable needed for reporting back to osbuild-composer.
+	var osbuildJobResult *worker.OSBuildJobResult = &worker.OSBuildJobResult{
 		Success: false,
+		OSBuildOutput: &osbuild.Result{
+			Success: false,
+		},
+		UploadStatus: "failure",
 	}
 
+	var outputDirectory string
+
+	// In all cases it is necessary to report result back to osbuild-composer worker API.
 	defer func() {
-		var targetErrors []string
-		for _, err := range r {
-			errStr := err.Error()
-			fmt.Printf("target errored: %s", errStr)
-			targetErrors = append(targetErrors, errStr)
-		}
-
-		var uploadstatus string = "failure"
-		if len(targetErrors) == 0 {
-			uploadstatus = "success"
-		}
-
-		// In all cases it is necessary to report result back to osbuild-composer worker API
-		err := job.Update(&worker.OSBuildJobResult{
-			Success:       osbuildOutput.Success && len(targetErrors) == 0,
-			OSBuildOutput: osbuildOutput,
-			TargetErrors:  targetErrors,
-			TargetResults: targetResults,
-			UploadStatus:  uploadstatus,
-		})
+		err := job.Update(osbuildJobResult)
 		if err != nil {
 			log.Printf("Error reporting job result: %v", err)
 		}
@@ -133,13 +124,13 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	}
 
 	// Run osbuild and handle two kinds of errors
-	osbuildOutput, err = RunOSBuild(args.Manifest, impl.Store, outputDirectory, exports, os.Stderr)
+	osbuildJobResult.OSBuildOutput, err = RunOSBuild(args.Manifest, impl.Store, outputDirectory, exports, os.Stderr)
 	// First handle the case when "running" osbuild failed
 	if err != nil {
 		return err
 	}
 	// Second handle the case when the build failed, but osbuild finished successfully
-	if !osbuildOutput.Success {
+	if !osbuildJobResult.OSBuildOutput.Success {
 		return nil
 	}
 
@@ -149,7 +140,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	// by any of the image types and it can't be specified during the request.
 	// Use the first (and presumably only) export for the imagePath.
 	exportPath := exports[0]
-	if osbuildOutput.Success && args.ImageName != "" {
+	if osbuildJobResult.OSBuildOutput.Success && args.ImageName != "" {
 		var f *os.File
 		imagePath := path.Join(outputDirectory, exportPath, args.ImageName)
 		if args.StreamOptimized {
@@ -170,7 +161,11 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		}
 	}
 
-	if len(args.Targets) == 1 {
+	if len(args.Targets) == 0 {
+		// There is no upload target, mark this job a success.
+		osbuildJobResult.Success = true
+		osbuildJobResult.UploadStatus = "success"
+	} else if len(args.Targets) == 1 {
 		switch options := args.Targets[0].Options.(type) {
 		case *target.VMWareTargetOptions:
 			credentials := vmware.Credentials{
@@ -184,7 +179,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 			tempDirectory, err := ioutil.TempDir(impl.Output, job.Id().String()+"-vmware-*")
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
@@ -200,20 +195,22 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			imagePath := path.Join(tempDirectory, imageName)
 			err = os.Symlink(streamOptimizedPath, imagePath)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
 			err = vmware.UploadImage(credentials, imagePath)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
 		case *target.AWSTargetOptions:
 			a, err := awsupload.New(options.Region, options.AccessKeyID, options.SecretAccessKey)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
@@ -224,29 +221,32 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 			_, err = a.Upload(path.Join(outputDirectory, options.Filename), options.Bucket, key)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
 			ami, err := a.Register(args.Targets[0].ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
 			if ami == nil {
-				r = append(r, fmt.Errorf("No ami returned"))
+				appendTargetError(osbuildJobResult, fmt.Errorf("No ami returned"))
 				return nil
 			}
 
-			targetResults = append(targetResults, target.NewAWSTargetResult(&target.AWSTargetResultOptions{
+			osbuildJobResult.TargetResults = append(osbuildJobResult.TargetResults, target.NewAWSTargetResult(&target.AWSTargetResultOptions{
 				Ami:    *ami,
 				Region: options.Region,
 			}))
+
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
 		case *target.AzureTargetOptions:
 			azureStorageClient, err := azure.NewStorageClient(options.StorageAccount, options.StorageAccessKey)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return err
 			}
 
@@ -264,13 +264,16 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			)
 
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
+
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
 		case *target.GCPTargetOptions:
 			g, err := gcp.New(impl.GCPCreds)
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
@@ -278,7 +281,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			_, err = g.StorageObjectUpload(path.Join(outputDirectory, options.Filename),
 				options.Bucket, options.Object, map[string]string{gcp.MetadataKeyImageName: args.Targets[0].ImageName})
 			if err != nil {
-				r = append(r, err)
+				appendTargetError(osbuildJobResult, err)
 				return nil
 			}
 
@@ -305,7 +308,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 			// check error from ComputeImageImport()
 			if importErr != nil {
-				r = append(r, importErr)
+				appendTargetError(osbuildJobResult, importErr)
 				return nil
 			}
 			log.Printf("[GCP] 💿 Image URL: %s", g.ComputeImageURL(args.Targets[0].ImageName))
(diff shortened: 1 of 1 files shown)


msehnout:
Second attempt on fixing #1276

ondrejbudai:
If I get this PR correctly, then if osbuild succeeds but the `vmdk` conversion fails, the worker will still report success. Do I get it right?

msehnout:
> If I get this PR correctly, then if osbuild succeeds but the `vmdk` conversion fails, the worker will still report success. Do I get it right?

It should not. I've modified the code to append errors to `r` even in cases where it wasn't reported previously.

jmikovic (dismissed):
looks good :+1:

thozza (dismissed):
All looking good. Very nice work!

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -89,8 +122,9 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	start_time := time.Now()
 
-	osbuildOutput, err := RunOSBuild(args.Manifest, impl.Store, outputDirectory, os.Stderr)
+	osbuildOutput, err = RunOSBuild(args.Manifest, impl.Store, outputDirectory, os.Stderr)
 	if err != nil {
+		r = append(r, err)
```
ondrejbudai, line 127: I'm a little worried about appending the error from osbuild to the list of target errors. Target errors have never been strongly defined but the idea is that they should be somewhat connected with upload targets.

Btw once upon a time, there was an idea to distinguish between a misbehaving worker and a failed job.

When building a particular manifest fails or an upload fails, this is a failed job that results in a failed compose.

When some other stuff fails (like creating a temporary directory on the worker or running osbuild for different reasons than a bad manifest), the worker should be treated like a misbehaving and the job should be rescheduled. We never really finished this part.

So yeah, I think this is an area where definitely is a lot of room for improvements and discussions. Overall, I think that your workaround makes the situation slightly better, I'm just worried that we're abusing `targetErrors` for something they were never intended for.

msehnout: Yes, this is exactly what I was confused about. There is no obvious place to report different kinds of errors. Let's remove this particular line and keep the rest. We can follow up later with bigger refactoring and extending the Result structure.

teg:
I really like where this is going! I have some nitpicks (related to @ondrejbudai's), but maybe taking a step back and simplifying the whole `Run()` method might make it all easier:
 - drop support for `LocalTarget`, this has not been used in a long time, and we don't really need to stay compatible across many releases (just as long as we don't get problems with having to deploy in lock-step), at least not yet.
 - drop support for `KojiTarget`, this has been replaced by the `osbuild-koji` job type.
 - drop support for more than one `Target` in the targets array. We currently only ever have 0 or 1 targets, let's just error out if we are passed anything else? This does not change the actual API spec.
 - drop the loop over the targets array now that we know we have at most one.
 - similarly always return 0 or 1 TargetResult OR TargetError (again without changing the spec).
 - add a new field: `InternalError` for any errors that are not returned from the remote side (`TargetErrors`).

I really like the new defer logic, though I would go all the way and simply `return` on error rather than `continue` as we do now. Then we don't have to keep checking if `osbuild` failed, and the overall `Success` field can be set to `true` at the end of the function, without having to deduce it when putting together the struct.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -70,17 +70,50 @@ func osbuildStagesToRPMs(stages []osbuild.StageResult) []koji.RPM {
...
+		var targetErrors []string
+		for _, err := range r {
+			errStr := err.Error()
+			fmt.Printf("target errored: %s", errStr)
+			targetErrors = append(targetErrors, errStr)
+		}
+
+		var uploadstatus string = "failure"
+		if len(targetErrors) == 0 {
+			uploadstatus = "success"
```
teg, line 91: Is this right even if we did not get to the uploads yet? I forgot what statuses we support, but maybe the default should be "pending" or something like that?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -70,17 +70,50 @@ func osbuildStagesToRPMs(stages []osbuild.StageResult) []koji.RPM {
...
+		}
+
+		// In all cases it is necessary to report result back to osbuild-composer worker API
+		err := job.Update(&worker.OSBuildJobResult{
+			Success:       osbuildOutput.Success && len(targetErrors) == 0,
```
teg, line 96: This seems fragile, could we instead set a variable `success := true` at the end of the `Run()` function and rely on that? That would require us to exit early on errors.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -104,24 +137,24 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		if args.StreamOptimized {
 			f, err = vmware.OpenAsStreamOptimizedVmdk(imagePath)
 			if err != nil {
+				r = append(r, err)
```
teg, line 140: These errors are not really `TargetErrors`, I suggest introducing a new `InternalError` and setting that to the first internal error we get (and exiting when that happens).

teg (dismissed):
Left a few nitpicks, but as I said, maybe the simplification I suggested in my first comment would make this easier/redundant.

achilleas-k:
I wonder if it would be simpler and more readable if, instead of several variables, a whole instance of `worker.OSBuildJobResult` was initialised at the beginning and that was used directly throughout the function.

For example
```go
result := &worker.OSBuildJobResult{
     // empty values with Success = false
}

defer func() {
	if err := job.Update(result); err != nil {
		log.Printf("Error reporting job result: %v", err)
	}

	if err := os.RemoveAll(outputDirectory); err != nil {
		log.Printf("Error removing temporary output directory (%s): %v", outputDirectory, err)
	}
}()
```

This way errors can be collected on the result object manipulated directly and the `UploadStatus` can be set to `"success"` after all uploads have succeeded, etc.

msehnout:
Thanks for suggestions @teg and @achilleas-k . I included most of them in this PR, but I'm not sure about introducing the `InternalError` field. After all it is already reported in the osbuild-worker process, so isn't it reporting the same twice? I mean we sure need better error reporting from osbuild to worker to composer (see #1270 for example), but that will require bigger effort across the whole project.

msehnout:
I'm not sure what's wrong with `golangci-lint run` :-/. It reports error on line that does not exist...

ondrejbudai:
Ah, `Checks` are running on the merge commit instead of PR head, see https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

This is then probably caused by a merge conflict. Rebasing your PR on top of `main` should fix it.

achilleas-k:
Looks like the rebases brought in the offending line now (L260).

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -422,151 +412,29 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
-		Success:       osbuildOutput.Success && len(targetErrors) == 0,
-		OSBuildOutput: osbuildOutput,
-		TargetErrors:  targetErrors,
-		TargetResults: targetResults,
-		UploadStatus:  uploadstatus,
-	})
-	if err != nil {
-		return fmt.Errorf("Error reporting job result: %v", err)
+	} else if len(args.Targets) > 1 {
+		panic("The job request contains multiple upload targets. This is not supported!")
```
achilleas-k, line 437: If this occurs (if it's possible to occur), will the image be accessible after the panic/return?
If not, probably best to panic/return before running the build.

msehnout: It shouldn't be, but probably is :-D. So this makes sense! Thanks :-)

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -119,6 +118,9 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	if err != nil {
 		return err
 	}
+	if osbuildOutput.Success == false {
+		panic("Osbuild failed, yet did not return an error. This is an implementation bug.")
```
ondrejbudai, line 113: I'm not sure about this. From `RunOSBuild()`:


```golang
err = cmd.Wait()

// try to decode the output even though the job could have failed
var result osbuild.Result
decodeErr := json.Unmarshal(stdoutBuffer.Bytes(), &result)
if decodeErr != nil {
	return nil, fmt.Errorf("error decoding osbuild output: %v\nthe raw output:\n%s", decodeErr, stdoutBuffer.String())
}

if err != nil {
	// ignore ExitError if output could be decoded correctly
	if _, isExitError := err.(*exec.ExitError); !isExitError {
		return nil, fmt.Errorf("running osbuild failed: %v", err)
	}
}

return &result, nil
```

If osbuild returns non-zero but its output can be parsed (i.e. it failed to build the given manifest), then we should report this failure back to composer.

msehnout: Yes, I got confused with the code. After reading the commit message again, I get it.

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: hmm, in the subsequent commits I default to "failed". I think the code for each upload target should set "pending" whenever it starts the upload.

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: Hopefully fixed in following commits.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -119,41 +147,16 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
-				r = append(r, err)
-				continue
-			}
+	if len(args.Targets) == 0 {
+		// There is no upload target, mark this job a success.
+		osbuildJobResult.Success = true
+		osbuildJobResult.UploadStatus = "success"
+	} else if len(args.Targets) == 1 {
+		// Although the API takes a list of targets, we only support one upload target
+		// per job, so we only need to handle this case and panic in case of multiple
```
thozza, line 156: The comment talks about "panic", but there is no panic... And probably IMHO should not be, because providing multiple targets is not a programmer's error, but caller's error(?)

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Since the worker does not send any update to composer before the upload is finished or failed, setting `uploadstatus` to "pending" seems unnecessary to me. That is unless you also call `job.Update()`. Personally I think that defaulting to "failed" is reasonable.

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: You are right, I forgot to remove this comment. See the "if" statement checking for a single upload target. It returns a failure to osbuild-composer. I think that makes more sense, WDYT?

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: See my comment bellow. I think this is a good idea, but I'd implement it as part of a new PR because this might again turn out to be a bigger issue than just sending a single message.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -69,23 +63,55 @@ func osbuildStagesToRPMs(stages []osbuild.StageResult) []koji.RPM {
...
 	err = job.Args(&args)
 	if err != nil {
 		return err
 	}
+	// The specification allows multiple upload targets because it is an array, but we don't support it.
+	// Return an error to osbuild-composer.
+	if len(args.Targets) > 1 {
+		log.Printf("The job specification contains more than one upload target. This is not supported any more. " +
+			"This might indicate a deployment of incompatible osbuild-worker and osbuild-composer versions.")
+		return nil
```
thozza, line 113: I wonder whether it would not make more sense to create the above comment as an `error` and add it to target errors?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -133,41 +161,13 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
-			err = job.UploadArtifact(options.Filename, f)
-			if err != nil {
-				r = append(r, err)
-				continue
-			}
+	if len(args.Targets) == 0 {
+		// There is no upload target, mark this job a success.
+		osbuildJobResult.Success = true
+		osbuildJobResult.UploadStatus = "success"
+	} else if len(args.Targets) == 1 {
```
thozza, line 168: I don't have a problem with being explicit here and have the `else if` branch, but it should not be needed because of the `if len(args.Targets) > 1 {` earlier.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -436,152 +434,28 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
-			_, err = k.CGImport(build, buildRoots, output, options.UploadDirectory, options.Token)
-			if err != nil {
-				r = append(r, err)
-				continue
-			}
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
 		default:
-			r = append(r, fmt.Errorf("invalid target type"))
+			err = fmt.Errorf("invalid target type")
```
thozza, line 454: You could optionally also log the target type here to have more information for debugging possible issues in the future. Not needed, since it was never done before.

thozza (commented):
The code looks very good. I added a few comments, nothing major.

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: I'm not sure really. I think this would make sense as the `InternalError`, but I don't want to add it as part of this PR.

cmd/osbuild-worker/jobimpl-osbuild.go
msehnout: I know, but it feels easier to read for me, because it makes the intention clear. But I'm also ok removing it ....

thozza: That's fine by me... :wink:

msehnout:
@thozza I've fixed most of the issues. I used new commits instead of amending because none of the issues was introduced in this PR, they were only part of the refactoring process.

thozza (approved):
Looking good!

msehnout:
There are two failed tests:
1. F32 OSTree:
```
[2021-03-19T10:17:52.149Z] TASK [assert] ******************************************************************

[2021-03-19T10:17:52.149Z] fatal: [192.168.100.50]: FAILED! => {

[2021-03-19T10:17:52.149Z]     "assertion": "'Boot Status is GREEN - Health Check SUCCESS' in result_greenboot_log.stdout",

[2021-03-19T10:17:52.149Z]     "changed": false,

[2021-03-19T10:17:52.149Z]     "evaluated_to": false,

[2021-03-19T10:17:52.149Z]     "msg": "Fallback log not found"

[2021-03-19T10:17:52.149Z] }

```
2. CS8 Image:
```
[2021-03-19T09:33:14.576Z] 🏃🏻 Running test: centos_8-x86_64-ami-boot.json

[2021-03-19T09:33:14.576Z] ------------------------------------------------------------------------------

[2021-03-19T09:33:14.576Z] === RUN   TestImages

[2021-03-19T09:33:14.576Z] === RUN   TestImages/centos_8-x86_64-ami-boot.json

[2021-03-19T09:39:07.128Z] === RUN   TestImages/centos_8-x86_64-ami-boot.json/image_info

[2021-03-19T09:39:07.129Z] qemu-img: Could not open '/var/lib/osbuild-composer-tests/osbuild-image-tests-087454950/assembler/image.raw': Could not open '/var/lib/osbuild-composer-tests/osbuild-image-tests-087454950/assembler/image.raw': No such file or directory

[2021-03-19T09:39:07.129Z] --- Output from ['qemu-img', 'info', '--output=json', '/var/lib/osbuild-composer-tests/osbuild-image-tests-087454950/assembler/image.raw']:

[2021-03-19T09:39:07.129Z] 

[2021-03-19T09:39:07.129Z] --- End of the output

[2021-03-19T09:39:07.129Z] Traceback (most recent call last):

[2021-03-19T09:39:07.129Z]   File "/usr/libexec/osbuild-composer-test/image-info", line 668, in <module>

[2021-03-19T09:39:07.129Z]     main()

[2021-03-19T09:39:07.129Z]   File "/usr/libexec/osbuild-composer-test/image-info", line 662, in main

[2021-03-19T09:39:07.129Z]     report = analyse_image(target)

[2021-03-19T09:39:07.129Z]   File "/usr/libexec/osbuild-composer-test/image-info", line 529, in analyse_image

[2021-03-19T09:39:07.129Z]     imgfmt = read_image_format(image)

[2021-03-19T09:39:07.129Z]   File "/usr/libexec/osbuild-composer-test/image-info", line 154, in read_image_format

[2021-03-19T09:39:07.129Z]     qemu = subprocess_check_output(["qemu-img", "info", "--output=json", device], json.loads)

[2021-03-19T09:39:07.129Z]   File "/usr/libexec/osbuild-composer-test/image-info", line 143, in subprocess_check_output

[2021-03-19T09:39:07.129Z]     output = subprocess.check_output(argv, encoding="utf-8")

[2021-03-19T09:39:07.129Z]   File "/usr/lib64/python3.6/subprocess.py", line 356, in check_output

[2021-03-19T09:39:07.129Z]     **kwargs).stdout

[2021-03-19T09:39:07.129Z]   File "/usr/lib64/python3.6/subprocess.py", line 438, in run

[2021-03-19T09:39:07.129Z]     output=stdout, stderr=stderr)

[2021-03-19T09:39:07.129Z] subprocess.CalledProcessError: Command '['qemu-img', 'info', '--output=json', '/var/lib/osbuild-composer-tests/osbuild-image-tests-087454950/assembler/image.raw']' returned non-zero exit status 1.
```
I believe both are unrelated to this PR.

PR #1282, opened 2021-03-10
distro/rhel9: add RHEL9 support

jmikovic:
This pull request includes:

requires: https://github.com/osbuild/osbuild/pull/612


cmd/osbuild-composer/composer.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/distro/fedora33"
 	"github.com/osbuild/osbuild-composer/internal/distro/rhel8"
 	"github.com/osbuild/osbuild-composer/internal/distro/rhel84"
+	"github.com/osbuild/osbuild-composer/internal/distro/rhel9"
```
teg, line 29: let's call this `rhel90`, to avoid repeating the mistake of `rhel8` (which really is `rhel83`)

repositories/rhel-9.json
```diff
@@ -0,0 +1,66 @@
+{
+    "aarch64": [
+        {
+            "name": "baseos",
+            "baseurl": "http://download-node-02.eng.bos.redhat.com/rhel-9/composes/RHEL-9/RHEL-9.0.0-20210228.7/compose/BaseOS/aarch64/os",
```
teg, line 5: This should probably be the CDN address, and then we override it in tests, as this URL will expire.

test/data/repositories/rhel-9.json
```diff
@@ -0,0 +1,66 @@
+{
+    "aarch64": [
+        {
+            "name": "baseos",
+            "baseurl": "http://download-node-02.eng.bos.redhat.com/rhel-9/composes/RHEL-9/RHEL-9.0.0-20210228.7/compose/BaseOS/aarch64/os",
```
teg, line 5: This is fine for now I think, but we should ask @dvdhrm to add this to his snapshots so we can use a RHEL9 snapshot here as we do for RHEL8.

tools/test-case-generators/repos.json
```diff
@@ -176,5 +176,55 @@
         "check_gpg": true
       }
     ]
+  },
+  "rhel-9": {
+    "x86_64": [
+      {
+        "name": "baseos",
+        "baseurl": "http://download-node-02.eng.bos.redhat.com/rhel-9/composes/RHEL-9/RHEL-9.0.0-20210228.7/compose/BaseOS/x86_64/os",
```
teg, line 184: Same as other comment, let's move this to rpmrepos before we land this.

internal/distro/rhel9/distro.go
```diff
@@ -0,0 +1,870 @@
...
+	"github.com/osbuild/osbuild-composer/internal/osbuild"
+
+	"github.com/google/uuid"
+
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/crypt"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+)
+
+const name = "rhel-9"
```
teg, line 22: Let's go with `rhel-90`.

teg (commented):
Nice start! If you can build an image but not boot it, I suggest grabbing the most recent nightly qcow2 and run image-info on that as well as ours to compare.

cmd/osbuild-composer/composer.go
ondrejbudai: Or maybe go even deeper and call it `rhel90beta`?

internal/distro/rhel9/distro.go
ondrejbudai: Same as before.

jmikovic:
Changed the name to `rhel90` and made changes to distro implementation so now the image is bootable but there are still some issues as `image-info` is failing because it's unable to mount root partition as `read-only`, test case generation also fails because of it

atodorov:
@jmikovic would you copy the EL8.4 sections in Jenkinsfile and make similar ones so we can consume the EL9 contents via Schutzbot ?

jmikovic:
yes, I want to do that but first I need to figure out how to run test for different distro than the one test is running on, since we want to test el9 build on el8, right @teg ?

teg:
> yes, I want to do that but first I need to figure out how to run test for different distro than the one test is running on, since we want to test el9 build on el8, right @teg ?

Sorry, missed this comment. Yeah, I think the first aim should be to hook this up with the koji API, and test it on the `EL8`, as this is firstly about building images for brew. Secondly we should enable this for the weldr API, which would require actual EL9.0 images to test on, but that can be a follow-up, in my opinion.

teg:
Seems like the compose we are testing against is gone.

@dvdhrm did you write up any documentation yet on how to use the new rpmrepo APIs? I see the snapshots are there for RHEL9, but how do I query for the most recent snapshot locations?

dvdhrm:
> Seems like the compose we are testing against is gone.
> 
> @dvdhrm did you write up any documentation yet on how to use the new rpmrepo APIs? I see the snapshots are there for RHEL9, but how do I query for the most recent snapshot locations?

You can just enumerate them all and then filter via `jq`, like this:

    curl -s "https://rpmrepo.osbuild.org/v2/enumerate" | jq 'map(select(startswith("el9-")))'

Alternatively, you can just query for all snapshots of a single repository by appending the snapshot without the date suffix:

    curl -s "https://rpmepo.osbuild.org/v2/enumerate/el9-x86_64-baseos-n9.0" | jq .

The current snapshots all have different timestamps. I have scheduled snapshots for tomorrow which will all have the `-20210326` suffix.

The current snapshots for EL9 are:

    [
      "el9-aarch64-appstream-n9.0-202103231341",
      "el9-aarch64-baseos-n9.0-202103231341",
      "el9-ppc64le-appstream-n9.0-202103231341",
      "el9-ppc64le-baseos-n9.0-202103231345",
      "el9-s390x-appstream-n9.0-202103231349",
      "el9-s390x-baseos-n9.0-202103231356",
      "el9-x86_64-appstream-n9.0-202103231357",
      "el9-x86_64-baseos-n9.0-202103231359",
      "el9-x86_64-rt-n9.0-202103231401"
    ]

teg:
Thanks @dvdhrm ! I guess we should go work the v2 repos if that's the future. What do you think @jmikovic ?

repositories/rhel-9.json
jmikovic: is there already CDN address for el9?

teg (approved):
This is great! Thanks so much @jmikovic! 🥳

PR #1283, opened 2021-03-10
GCP refactoring: logging, storage cleanup, worker credentials

thozza:
This PR addresses couple of outstanding things, which popped up during the initial GCP PR and later during review of Azure support for Cloud API.

Main changes:
- Moved all logging out of the GCP internal library to the callers (worker and the CLI upload tool), similar as it is done for Azure.
- Made all GCP methods a single purpose and not to do any extra tasks (e.g. cleanup). Rather than cleaning up under the hood, an error is returned and the caller can use provided cleanup methods to delete any unwanted resources. This is true for importing image. If the image upload to Storage fails, there is no need to clean up the object since it has been not created.
- Modified the worker and CLI upload tool code to do all necessary clean up in case of errors 
- Modified worker GCP-related code to try to share the imported image only if the list of accounts is not empty.
- Minor simplification of Google Compute Node API calls in the GCP library based on project's code examples.
- Added building of `osbuild-upload-gcp` to Makefile's build target. This was previously missed.
- Refactored handling of GCP credentials file on the worker. Instead of passing the path to the file to job, the file is read on the start up and its content is passed to the job. This mimics what is done for Azure and helps to detect issues with the credentials file early. Note that there is no easy way to actually test the validity of the content of the file and credentials other than making some actual API calls. I decided to not to do any "random" calls for now.

This pull request includes:

 - no new tests have been added
 - changes should be transparent to the end user, thus no documentation has been added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ build:
 	go build -o bin/osbuild-pipeline ./cmd/osbuild-pipeline/
 	go build -o bin/osbuild-upload-azure ./cmd/osbuild-upload-azure/
 	go build -o bin/osbuild-upload-aws ./cmd/osbuild-upload-aws/
+	go build -o bin/osbuild-upload-gcp ./cmd/osbuild-upload-gcp/
 	go test -c -tags=integration -o bin/osbuild-composer-cli-tests ./cmd/osbuild-composer-cli-tests/main_test.go
 	go test -c -tags=integration -o bin/osbuild-weldr-tests ./internal/client/
 	go test -c -tags=integration -o bin/osbuild-dnf-json-tests ./cmd/osbuild-dnf-json-tests/main_test.go
--- a/cmd/osbuild-upload-gcp/main.go
+++ b/cmd/osbuild-upload-gcp/main.go
@@ -4,7 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"os"
+	"log"
 
 	"github.com/osbuild/osbuild-composer/internal/upload/gcp"
 )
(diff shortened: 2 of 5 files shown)


teg (commented):
This makes a lot of sense to me. Something also worth considering is to pass `ctx` to our various long-running library calls that can support it so that we can have better cancellation support. Though osbuild itself might not be ready for that yet, so utility might be limited still.

thozza:
> This makes a lot of sense to me. Something also worth considering is to pass `ctx` to our various long-running library calls that can support it so that we can have better cancellation support. Though osbuild itself might not be ready for that yet, so utility might be limited still.

Yes, good point. This is still on my TODO...

thozza:
CI "almost" passed - only "F33 aarch64 Base" test was aborted. This change set didn't touch any Fedora-related. I'm happy to reschedule to make it completely green...

ondrejbudai (approved):
Awesome work!

As a caller of the `StorageObjectUpload` method, I would expect that it either uploads a file correctly or it fails and doesn't leave any traces in the storage. This can be implemented in a future PR though.

thozza:
> Awesome work!
> 
> As a caller of the `StorageObjectUpload` method, I would expect that it either uploads a file correctly or it fails and doesn't leave any traces in the storage. This can be implemented in a future PR though.

Thanks for your opinion Ondrej. I reworked the code a bit more yesterday and it should be already behaving this way. I just didn't update the PR text. Now the MD5 checksum is provided when uploading the file and the server side rejects the object on Close() if there is a mismatch, rather than checking it on our side after the file has been uploaded.

ondrejbudai:
`// Uploaded data is rejected if its MD5 hash does not match the set value.`

Ah, I missed this comment, thanks for the clarification!

thozza:
The only failed test in CI is "F32 New OSTree" which passed for the same code (commit) in previous [run number 4](https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1283/4/pipeline/296).

I would call this as overall pass and merge. Any objections @ondrejbudai @teg ?

ondrejbudai:
Merged, thanks!

PR #1286, opened 2021-03-10
[PoC]worker/depsolve: add a depsolve job type

teg:
Allow depsolving to be done in a worker through the job queue rather
than synchronously in composer.

The benefit this might unlock include:
 - no more blocking calls in the cloud/koji APIs
 - only workers accessing repositoires
   - no VPN access from composer
   - composer not needing to be subscribed to CDN, etc
 - no dnf cache managment in composer

Potential problems:
 - the version of composer (so the distro definitions) that
   triggered a depsolve, may not be the same that uses the
   result to generate a manfiset

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+)
+
+type DepsolveJobImpl struct {
+	RPMMD rpmmd.RPMMD
+}
+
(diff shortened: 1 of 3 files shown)


ondrejbudai:
Nice! Btw this will require some fine handling of `dnf-json` - we need to move it from `osbuild-composer` to `osbuild-composer-worker.

teg:
I thought some more about the problem I highlighted above, and I'm really not sure how to solve it. Imagine this situation:
 - osbuild-composer-100 generates a `PackageSets` map to be depsolved for a given image-request
 - osbuild-worker does as requested and returns the result in the queue to composer
 - by the time osbuild-composer gets around to calling `Manifest()` on the result, it has been updated to osbuild-composer-104, and the `PackageSets` no longer match the `Manifest()` definition, and things are sad.

A solution would be for the worker to not only do the depsolving, but also the manifest generation, which gets us down the rabbit hole of where manifest generators should live (in the distro would work, on each worker node would be.... confusing?).

ondrejbudai:
One solution is to add a composer version to the job. When it finishes, check if the version is still same. Otherwise, reschedule. Might be troublesome when running multiple composers in k8s/OpenShift.

But I guess the final solution is to move the distro distribution in the distro itself...

PR #1288, opened 2021-03-11
Refactor distro registry to simplify the process of adding a new distro

ondrejbudai:
I made several changes to simplify the process of adding a new distribution, see individual commits for a deeper explanation.

This pull request includes:

- [x] adequate testing for the new functionality or fixed issue (I added some tests)
- [x] adequate documentation informing people about the change (not a functional change)

### DIFF
--- a/internal/distroregistry/distroregistry.go
+++ b/internal/distroregistry/distroregistry.go
@@ -1,0 +1,60 @@
+package distroregistry
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+
+	"github.com/osbuild/osbuild-composer/internal/distro"
+)
+
+type Registry struct {
+	distros map[string]distro.Distro
+}
+
+func New(distros ...distro.Distro) (*Registry, error) {
+	reg := &Registry{
+		distros: make(map[string]distro.Distro),
+	}
+	for _, d := range distros {
+		name := d.Name()
+		if _, exists := reg.distros[name]; exists {
+			return nil, fmt.Errorf("New: passed two distros with the same name: %s", d.Name())
+		}
+		reg.distros[name] = d
+	}
+	return reg, nil
+}
+
+func (r *Registry) GetDistro(name string) distro.Distro {
+	d, ok := r.distros[name]
+	if !ok {
+		return nil
+	}
+
+	return d
+}
+
+// List returns the names of all distros in a Registry, sorted alphabetically.
+func (r *Registry) List() []string {
+	list := []string{}
+	for _, d := range r.distros {
+		list = append(list, d.Name())
+	}
+	sort.Strings(list)
+	return list
+}
+
+func (r *Registry) FromHost() (distro.Distro, bool, bool, error) {
+	name, beta, isStream, err := distro.GetHostDistroName()
+	if err != nil {
+		return nil, false, false, err
+	}
+
+	d := r.GetDistro(name)
+	if d == nil {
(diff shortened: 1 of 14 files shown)


internal/distroregistry/distroregistry.go
```diff
@@ -26,6 +40,23 @@ func New(distros ...distro.Distro) (*Registry, error) {
...
+// supportedDistros variable.
+func NewDefault() *Registry {
+	var distros []distro.Distro
+	for _, distroInitializer := range supportedDistros {
+		distros = append(distros, distroInitializer())
+	}
+
+	registry, err := New(distros...)
+	if err != nil {
+		panic(fmt.Sprintf("two supported distros have the same name, this is a programming error: %v", err))
```
msehnout, line 54: Not super important: I think this error message would be better in the `New()` function, or maybe it's not necessary to have it in the first place? The line `return nil, fmt.Errorf("New: passed two distros with the same name: %s", d.Name())` says mostly the same.

internal/distroregistry/distroregistry_test.go
```diff
@@ -19,3 +21,16 @@ func TestRegistry_List(t *testing.T) {
...
+func TestRegistry_GetDistro(t *testing.T) {
+	distros := NewDefault()
+
+	t.Run("distro exists", func(t *testing.T) {
+		expectedDistro := rhel8.New()
+		require.Equal(t, expectedDistro.Name(), distros.GetDistro(expectedDistro.Name()).Name())
+	})
+
+	t.Run("distro doesn't exist", func(t *testing.T) {
+		require.Nil(t, distros.GetDistro("toucan-os"))
```
msehnout, line 34: What's the reason we don't support Toucan OS?

msehnout (approved):
Looks fine to me! Should I merge it right away?

teg (approved):
Love it! Suggestion for follow-up: assert that all manifest test cases have image-info/manifest/blueprint sections, and that each distro/arch/image-type has a manifest test case. Now it is possible to forget to add/update these, let's make that a tiny bit harder ;)
Similarly: let's drop support for only generating the manifests without also generating the image tests? We have the new `generate-all-test-cases` script now to help us...

internal/distroregistry/distroregistry_test.go
ondrejbudai: Because toucans can fly so they would make Tux jealous.

internal/distroregistry/distroregistry.go
ondrejbudai: Yeah, good point, I will try to remember to change it in a follow-up. :)

ondrejbudai:
> Love it! Suggestion for follow-up: assert that all manifest test cases have image-info/manifest/blueprint sections, and that each distro/arch/image-type has a manifest test case. Now it is possible to forget to add/update these, let's make that a tiny bit harder ;)
> Similarly: let's drop support for only generating the manifests without also generating the image tests? We have the new `generate-all-test-cases` script now to help us...

Yeah, I have similar thoughts, thanks!

internal/distroregistry/distroregistry_test.go
msehnout: Not true! https://youtu.be/9dfWzp7rYR4

ondrejbudai: LOL, I love everything about the video. :D:D:D

PR #1291, opened 2021-03-12
Extend cloud-cleaner and clean up after GCP integration tests

thozza:
This change set extends `cloud-cleaner` to clean up GCP resources. The internal GCP library is extended to allow deleting additional types of resources, such as images and instances.

Changes include:
- Move internal GCP library to from `internal/upload` to `internal/cloud`, because it now supports more than just uploading files. The idea is that functionality for other cloud-providers could be consolidated under `internal/cloud` in the future.
- Split the GCP library into multiple files, roughly based on the API that is being used (e.g. Storage, Compute).
- Changes to internal GCP library:
  - Method `StorageObjectUpload()` is extended to allow setting custom metadata on uploaded objects. Worker's job implementation as well as the GCP CLI upload tool have been modified to set the intended image name as a custom metadata on the uploaded object. This makes it possible to find the Storage object using the image name.
  - Add methods for deleting Compute Node image and instance.
- Changes to `cloud-cleaner`:
   - Move the Azure cleanup to separate function.
   - Add function for cleaning up GCP resources.
   - Run all available cleanup functions as goroutines.
- Extend Jenkinsfile to run `cloud-cleaner` after `rhel8`, `rhel84` and `cs8` integrations tests, which test GCP.
- Modify `api.sh` to use predictable GCP resource names, to make cleanup by `cloud-cleaner` possible. Predictable resource names are based on `TEST_ID`, which is in Jenkins environment constructed in the same way as done by `GenerateCIArtifactName()` from `internal/test/helpers.go`.

This pull request includes:

  - No new tests have been added, but reduced the CI to test cloud-cleaner for GCP before merging.
  - Add News entry about image name being added as metadata to uploaded GCP Storage object as part of worker job.


thozza:
`cloud-cleaner` successfully verified for GCP -> https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fosbuild-composer/detail/PR-1291/4/pipeline/54

teg (approved):

teg (changes requested):
Code looks great, but CI is failing on Fedora, so need to be able to handle the situation where the env vars are missing (or just enable it everywhere).

thozza:
> Code looks great, but CI is failing on Fedora, so need to be able to handle the situation where the env vars are missing (or just enable it everywhere).

I defined `DISTRO_CODE` for all integration testing stages, including Fedora, so it should pass now...

msehnout (approved):
Very nice! Thanks for improving cloud cleaner :-). 

One thing to consider though: Me and few other people in the team are not big fans of environment variables. I know I used them myself in the cloud cleaner code so I'm the one to blame in the first place :facepalm: , but do you think it would make sense to rewrite cloud-cleaner in a way that it takes JSON object, specifying the work it should do, as input and does not read any environment variables? The problem is that now we already need to set multiple variables and the list is only going to get bigger as we add support for AWS, VMWare, and maybe more. The JSON object would be composed in the `run_cloud_cleaner.sh` script and then fed into cloud-cleaner binary.

thozza:
> specifying the work it should do, as input and does not read any environment variables?

Thanks for the review!

My understanding of `cloud-cleaner`'s purpose is that it should try to do as much clean up as possible when run. For that it needs at least two things:

1. Credentials to authenticate with the Cloud provider's API.
2. A way to determine cloud resource names to try to clean up.

For the 1. point, I think that the test cases already need all the necessary environment variables to be defined in order to run. So these variables should be already available in the environment when `cloud-cleaner` is run. I'm not a Jenkins expert, but providing the variables in `run_cloud_cleaner.sh` to `cloud-cleaner` seems completely unnecessary to me. But I didn't test it without it, so I don't know if there is any reason to do it this way. AFAIK `cloud-cleaner` does not run in PrivateTmp, so it should be able to inherit variables set in the environment. If there are credentials to authenticate with "Cloud provider X" missing, then `cloud-cleaner` should just log the fact and skip the clean up for it.

As for point 2., I would argue that we should probably have predictable `TEST_ID` be set for each stage, which can be used by the test for resource names, as well as by `cloud-cleaner` to determine resource names to clean up. Composing this in any way in the `run_cloud_cleaner.sh` seems wrong to me. It should be ideally set in the environment globally or very early for all scripts. If there are any other environment variables needed to clean up resources (e.g. Bucket name, Zone/Region, ...), these would have to be set already for the test script. If not available, `cloud-cleaner` should probably just log the fact and skip the clean up. I admit that the GCP cleanup code could be modified to not fail too early if environment variable needed to clean up just one type of resource is not set (e.g. `GCP_BUCKET`).

Personally I don't have any issue with using environment variables. Especially when test scrips rely on them. Constructing JSON and passing it to `cloud-cleaner` seems unnecessary to me. In this case I see it as just another way of doing the same thing. But I'm not against it in any way...

msehnout:
> Personally I don't have any issue with using environment variables. Especially when test scrips rely on them. Constructing JSON and passing it to `cloud-cleaner` seems unnecessary to me. In this case I see it as just another way of doing the same thing. But I'm not against it in any way...

As I said, it's only a suggestion. Just to give you the perspective why doing the same thing in another way might be considered better: we have a lot of testing binaries and we also have a lot of configuration options encoded in the Jenkinsfile, so no one really knows what variables each of them require. But it works, which is fine. Until it doesn't, or until you want to run the test elsewhere. Then you end up reading the source code looking for references to environment variables all around the place (like we did when we attempted to migrate to Github workflows). So even though constructing a JSON from environment variables and running the tool with the JSON as input sounds ridiculous without context, it might be worth it in the long run.

thozza:
> > Personally I don't have any issue with using environment variables. Especially when test scrips rely on them. Constructing JSON and passing it to `cloud-cleaner` seems unnecessary to me. In this case I see it as just another way of doing the same thing. But I'm not against it in any way...
> 
> As I said, it's only a suggestion. Just to give you the perspective why doing the same thing in another way might be considered better: we have a lot of testing binaries and we also have a lot of configuration options encoded in the Jenkinsfile, so no one really knows what variables each of them require. But it works, which is fine. Until it doesn't, or until you want to run the test elsewhere. Then you end up reading the source code looking for references to environment variables all around the place (like we did when we attempted to migrate to Github workflows). So even though constructing a JSON from environment variables and running the tool with the JSON as input sounds ridiculous without context, it might be worth it in the long run.

Thanks or the wider context. I'm inclining to not to rework `cloud-cleaner` to accept JSON as part of this change set. It may make sense to rework it as part of adding support for AWS, since it will bring yet another set of variables.

teg (approved):

PR #1293, opened 2021-03-13
image-tests: checkpoint build root

teg:
Most build roots for a given distro/arch are identical, so let's checkpoint them to avoid having to regenerate them
all the time.

This risks increasing the disk space usage in the case the build roots do not match, but in practice that should be rare.


ondrejbudai (commented):
Looks good!

thozza (approved):
This makes sense from my point of view.

PR #1299, opened 2021-03-15
Main compose packagesets

bcl:
This pull request includes:

### DIFF
--- a/cmd/osbuild-store-dump/main.go
+++ b/cmd/osbuild-store-dump/main.go
@@ -170,11 +170,11 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	manifest, packages = getManifest(bp2, t2, a, d, rpmmd, repos)
+	manifest, packages = getManifest(bp1, t2, a, d, rpmmd, repos)
 	err = s.PushCompose(id2,
 		manifest,
 		t2,
-		&bp2,
+		&bp1,
 		0,
 		[]*target.Target{
 			awsTarget,
--- a/cmd/osbuild-store-dump/main.go
+++ b/cmd/osbuild-store-dump/main.go
@@ -16,7 +16,7 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/target"
 )
 
-func getManifest(bp blueprint.Blueprint, t distro.ImageType, a distro.Arch, d distro.Distro, rpm_md rpmmd.RPMMD, repos []rpmmd.RepoConfig) distro.Manifest {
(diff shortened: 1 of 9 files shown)


ondrejbudai:
This PR modified the signature of `Store.PushCompose()` but did not modify the calls, therefore the tests are failing.

I think we also need to add `PackageSets` to `internal/store/json.go` so the package sets are persistent.

internal/weldr/api.go
```diff
@@ -2370,6 +2365,18 @@ func (api *API) composeInfoHandler(writer http.ResponseWriter, request *http.Req
 		reply.Uploads = targetsToUploadResponses(compose.ImageBuild.Targets, composeStatus.State)
 	}
 
+	// Add package dependencies from the compose
+	// This information may not be included for the compose
+	if _, ok := compose.PackageSets["packages"]; ok {
+		dependencies := compose.PackageSets["packages"]
```
teg, line 2371: Could we just store the `packages` PackageSet in the store? That way we don't make the names of the sets API, and we would also avoid storing the build-root packages etc which we currently have no use for / way to export.

cmd/osbuild-store-dump/main.go
```diff
@@ -170,11 +170,11 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	manifest, packageSpecSets = getManifest(bp2, t2, a, d, rpmmd, repos)
+	manifest, packageSpecSets = getManifest(bp1, t2, a, d, rpmmd, repos)
```
teg, line 173: Could you explain the context a bit in the commit message? :)

internal/store/json.go
```diff
@@ -29,8 +30,9 @@ type workspaceV0 map[string]blueprint.Blueprint
 // It contains all the information necessary to generate the inputs for the job, as
 // well as the job's state.
 type composeV0 struct {
-	Blueprint   *blueprint.Blueprint `json:"blueprint"`
-	ImageBuilds []imageBuildV0       `json:"image_builds"`
+	Blueprint   *blueprint.Blueprint           `json:"blueprint"`
+	ImageBuilds []imageBuildV0                 `json:"image_builds"`
+	PackageSets map[string][]rpmmd.PackageSpec `json:"package_sets"`
```
teg, line 35: As suggested elsewhere: should we make this only be the `packages` set, so we don't make the names of the sets API, nor the build-root++

teg (changes requested):
I really like this! Just one suggestion for changes inline.

teg (approved):
Nice! Thanks!

ondrejbudai (approved):

bcl:
@teg Can this be merged? I don't see why Schutzbot says failed, everything looks green on the details page.

teg:
Base tests timed out on aarch64 EL8. This is caused by `qemu-img` hanging, not related to this PR.

PR #1303, opened 2021-03-16
GCP: Report the right error when image import fails

thozza:
Fix a bug in the worker job implementation and GCP CLI upload tool,
which causes the code to report wrong error instance in case the image
import failed for some reason.


ondrejbudai (approved):
Good catch!

PR #1309, opened 2021-03-18
Fix for RHEL for Edge Ref in new image types

achilleas-k:
This pull request includes:

- [x] adequate testing for the new functionality or fixed issue
- [x] adequate documentation informing people about the change

The new image type introduced in #1244 ignores the ostree ref supplied by the user and always uses the default `rhel/8/<arch>/edge` instead.

This PR fixes the issue by using the user-supplied ref when building the image.  It also updates the integration test script to use a non-default ref by prefixing it with `test/`.

For all distros/image types, the default is only read by the weldr API when a ref isn't specified.

PR #1310, opened 2021-03-18
RHEL 8.4.0: Fix for RHEL for Edge Ref in new image types

achilleas-k:
_Duplicate of #1309 for RHEL 8.4.0 branch._

This pull request includes:

The new image type introduced in #1244 ignores the ostree ref supplied by the user and always uses the default rhel/8/<arch>/edge instead.

This PR fixes the issue by using the user-supplied ref when building the image. It also updates the integration test script to use a non-default ref by prefixing it with test/.

For all distros/image types, the default is only read by the weldr API when a ref isn't specified.


ondrejbudai (approved):
Thanks!

PR #1312, opened 2021-03-19
ostree-ng.sh: simplify .ks

teg:
Bring the kickstart closer to the embedded kicktstart.

The embedded kickstart performs a default installation, but requires you to select the partitioning. This does the same, but applies autopart.

This pull request includes:


test/cases/ostree-ng.sh
```diff
@@ -340,45 +340,10 @@ sudo qemu-img create -f qcow2 "${LIBVIRT_IMAGE_PATH}" 20G
...
 zerombr
 clearpart --all --initlabel --disklabel=msdos
 autopart --nohome --noswap --type=plain
-ostreesetup --nogpg --osname=rhel-edge --remote=rhel-edge --url=file:///ostree/repo --ref=${OSTREE_REF}
-poweroff
-
-%post --log=/var/log/anaconda/post-install.log --erroronfail
-
-# no sudo password for user admin
-echo -e 'admin\tALL=(ALL)\tNOPASSWD: ALL' >> /etc/sudoers
```
henrywang, line 362: We need this one for avoid sudo password when run command via ssh.

henrywang:
@teg, test is stuck at anaconda installation. That means we need more "instructions" to make it automatically.

atodorov:
> @teg, test is stuck at anaconda installation. That means we need more "instructions" to make it automatically.

At which screen ? That's the command which is missing.

From the old tar kickstart test in lorax composer we use the following kickstart:

```
cmdline
lang en_US.UTF-8
timezone America/New_York
keyboard us
rootpw --lock
sshkey --username root "$PUB_KEY"
bootloader --location=mbr
zerombr
clearpart --initlabel --all
autopart
# reboot is used together with --no-reboot qemu-kvm parameter, which makes the qemu-kvm
# process exit after the installation is complete and anaconda reboots the system
# (using 'poweroff' ks command just halted the machine without powering it off)
reboot

liveimg --url file:///var/lib/lorax/composer/results/$UUID/root.tar.xz
```

test/cases/ostree-ng.sh
teg: Thanks! I tried fixing this by adding a password instead.

teg:
@henrywang @atodorov seems like I should not have dropped the `poweroff` command. Pushed with that readded.

osbuild-composer.spec
```diff
@@ -334,6 +334,7 @@ Requires:   libvirt-daemon-driver-secret
 Requires:   libvirt-daemon-driver-storage
 Requires:   libvirt-daemon-driver-storage-disk
 Requires:   libvirt-daemon-kvm
+Requires:   lorax
```
henrywang, line 337: @teg Looks `mkksiso` not included in RHEL release. https://src.osci.redhat.com/rpms/lorax/blob/rhel-8.4.0/f/lorax.spec#_208. It's only available for Fedora.

teg: Yeah, this is a pain. I got this working locally by spinning up a container, but that might be overkill.

ondrejbudai: Is it in EPEL? If so, we can just add it here: https://github.com/osbuild/osbuild-composer/blob/c2dc9d65fb71add09d5afc4f6e66991c46a7a1c0/tools/provision.sh#L11

teg: @bcl is there a convenient way to get `mkksiso` on RHEL8? Is it in EPEL?

PR #1319, opened 2021-03-22
Cloud API: fix `image_status.status` value for running compose

thozza:
Previously, the Cloud API endpoint `/v1/compose/{id}` return value's `image_status.status` for a running worker job was "running", which didn't comply with the Cloud API specification. Equivalents allowed by the API specification are "building", "uploading" or "registering".

As a result, the Image Builder API also does not comply in this regard to its specification, because it currently just copies the status value string returned by osbuild-composer.

Define the `image_status.status` as a reusable type in the Cloud API specification. This forces openapi to generate an explicit type for it, which can be then explicitly used in the code, instead of plain strings. The name `ImageStatusValue` is probably not ideal, but IMHO something like `ImageStatusStatus` would be even worse...

Return "building", instead of "running"  for running compose.

Modify api integration test to check for all valid `image_status.status` values for a compose.

Add News entry explaining this change.

This change will affect: 
* image-builder-frontend (PR https://github.com/osbuild/image-builder-frontend/pull/121)
  * [it expects "running" as the `image_status.status` value](https://github.com/osbuild/image-builder-frontend/blob/c6c616a9dc20166c465d05526dff041d20f42b47/src/PresentationalComponents/ImagesTable/ImageBuildStatus.js#L29-L38)
* image-builder (PR https://github.com/osbuild/image-builder/pull/163)
  * [api integration test will expect (once PR is merged) "running" as the `image_status.status` value](https://github.com/osbuild/image-builder/pull/156/files#diff-8c9fc1f11806f990402d945e7a6a3bdcf8be429120a4b8c2ec18e9ff5251f02fR344)
  * [code in `internal/cloudapi/client_test.go` expects "running"](https://github.com/osbuild/image-builder/blob/34f544c03b220d2114aa14c3f3e9f0ff6895869c/internal/cloudapi/client_test.go#L54-L83)
  * [code in `internal/server/server_test.go` expects "running"](https://github.com/osbuild/image-builder/blob/34f544c03b220d2114aa14c3f3e9f0ff6895869c/internal/server/server_test.go#L189-L231)

There may be more, but this is what I found so far...

This pull request includes:

  - no new testing was added, affected tests were modified.
  - News entry was added.

### DIFF
--- a/docs/news/unreleased/cloudapi-image_status-status-value.md
+++ b/docs/news/unreleased/cloudapi-image_status-status-value.md
@@ -1,0 +1,10 @@
+# Cloud API: fix `image_status.status` value for running compose
+
+Previously, the Cloud API endpoint `/v1/compose/{id}` return value's
+`image_status.status` for a running worker job was "running", which didn't
+comply with the Cloud API specification. Equivalents allowed by the API
+specification are "building", "uploading" or "registering".
+
+Return "building" as the `image_status.status` value for a running compose,
+instead of "running". Returning the remaining "uploading" and "registering"
+values is not yet implemented.
--- a/internal/cloudapi/openapi.gen.go
+++ b/internal/cloudapi/openapi.gen.go
@@ -150,9 +150,22 @@ type ImageRequest struct {
 
 // ImageStatus defines model for ImageStatus.
 type ImageStatus struct {
-	Status       string        `json:"status"`
-	UploadStatus *UploadStatus `json:"upload_status,omitempty"`
+	Status       ImageStatusValue `json:"status"`
+	UploadStatus *UploadStatus    `json:"upload_status,omitempty"`
 }
+
+// ImageStatusValue defines model for ImageStatusValue.
+type ImageStatusValue string
+
+// List of ImageStatusValue
+const (
+	ImageStatusValue_building    ImageStatusValue = "building"
+	ImageStatusValue_failure     ImageStatusValue = "failure"
+	ImageStatusValue_pending     ImageStatusValue = "pending"
+	ImageStatusValue_registering ImageStatusValue = "registering"
+	ImageStatusValue_success     ImageStatusValue = "success"
+	ImageStatusValue_uploading   ImageStatusValue = "uploading"
+)
 
 // Repository defines model for Repository.
 type Repository struct {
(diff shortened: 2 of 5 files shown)


thozza:
@teg, @Gundersanne, @ondrejbudai please have a look at this... I think we should have a discussion before merging this, but I wanted to draft a fix, becuase it was simple.... I think that the consequences of the change on all the other components is what is more complicated.

chloenayon:
Hey, a question - why not change the API specification to expect running instead of building?

thozza:
> Hey, a question - why not change the API specification to expect running instead of building?

I would say that in the context of having also "uploading" and "registering" statuses, "running" is too broad. It basically covers all three situations (including "building").

I would say that the question is what makes the most sense for API and the frontend? IMO we should either have only "running" or the three variants "building" / "uploading" / "registering" but not a combination.

teg:
This makes sense to me. To avoid having to coordinate the upgrade, should we add support for all values to image-builder first?
The reason we distinguished building from uploading and registering was that the latter two might be long-running and are out of our control, whereas building should be more predictable and something we could optimize. Actually hooking up the reporting of these values is still on the TODO though.

thozza:
> This makes sense to me. To avoid having to coordinate the upgrade, should we add support for all values to image-builder first?
> The reason we distinguished building from uploading and registering was that the latter two might be long-running and are out of our control, whereas building should be more predictable and something we could optimize. Actually hooking up the reporting of these values is still on the TODO though.

Sure, I'll look at this once I'm finished with documenting how to use images shared by the image-builder using the information returned from the API. I'll mark this PR as a draft for now...

thozza:
The Image Builder front-end change has been merged.

The Image Builder change is under review and good for merging. It is also backward compatible, so it will work fine with the current as well as new version of osbuild-composer.

This change is ready for review and for merging. The rationale is that this change won't affect Image Builder until we deploy it in staging / production, which will most probably happen after the Image Builder part has been already merged as well.

teg (approved):
Thanks! I'd probably merge this after image builder has been deployed, just to avoid any issues in case they get deployed in the wrong order.

Gundersanne (approved):
lgtm!

PR #1322, opened 2021-03-23
readme: require systemd 239

### LINKED ISSUE #1216: README says that composer requires systemd >= 244 but that's not true
Composer runs just fine on RHEL 8.3 with `systemd` 239.

Can any of our systemd experts (@teg, @dvdhrm) clarify which systemd version we indeed require?

ondrejbudai:
Systemd 239 is used in RHEL 8 and we test osbuild-composer against it
therefore it's safe to say that 239 is the minimal version we require.

Fixes #1216

[skip ci]

This pull request includes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ provided infrastructure and services.
 The requirements for this project are:
 
  * `osbuild >= 26`
- * `systemd >= 244`
+ * `systemd >= 239`
 
 At build-time, the following software is required:
 


teg:
I have no strong opinion. But note that systemd in RHEL has a million back ported patches, so it might very well be that things don't work with systemd 239.

ondrejbudai:
Hmm, so maybe let's approach this section differently? Something like that osbuild-composer is supported and tested on stable Fedora versions, RHEL 8.3+ and CentOS Stream 8?

teg:
Good idea, because that's what we actually do. Love it!

msehnout:
> I have no strong opinion. But note that systemd in RHEL has a million back ported patches, so it might very well be that things don't work with systemd 239.

The same will eventually apply to rhel-8.4 version of osbuild. So I agree that dropping this section altogether makes sense.

PR #1323, opened 2021-03-23
[POC] [RFC] Enable testing on s390x

jrusz:
This will enable testing on s390x in beaker on latest nightly
composes.

This pull request includes:


test/cases/image_tests.sh
```diff
@@ -5,6 +5,13 @@ set -euo pipefail
 source /etc/os-release
 ARCH=$(uname -m)
 
+# Set WORKSPACE to current dir if unset
+WORKSPACE=${WORKSPACE:=$(pwd)}
+
+# Set variables for tests to pass on s390x
+BUILD_ID=${BUILD_ID:='s390x'}
+DISTRO_CODE=${BUILD_ID:='EL8'}
+
```
jrusz, line 14: Since the testing is not running on regular Jenkins slave, these variables are unset and so I have to set them to something, I don't know what is appropriate though.

test/cases/libvirt.sh
```diff
@@ -1,19 +1,27 @@
...
+ARCH=$(uname -m)
 
 # Provision the software under test.
 /usr/libexec/osbuild-composer-test/provision.sh
 
 # Test the images
-/usr/libexec/osbuild-composer-test/libvirt_test.sh qcow2
-/usr/libexec/osbuild-composer-test/libvirt_test.sh openstack
-/usr/libexec/osbuild-composer-test/libvirt_test.sh vhd
+if [[ "${ARCH}" == "s390x" ]]; then
```
jrusz, line 12: There is only this image time on s390x.

test/cases/libvirt.sh
```diff
@@ -1,19 +1,27 @@
...
+fi
 
 # RHEL 8.4 and Centos Stream 8 images also supports uefi, check that
 if [[ "${ID}-${VERSION_ID}" == "rhel-8.4" || "${ID}-${VERSION_ID}" == "centos-8" ]]; then
   echo "🐄 Booting qcow2 image in UEFI mode on RHEL/Centos Stream"
   /usr/libexec/osbuild-composer-test/libvirt_test.sh qcow2 uefi
 fi
+
+echo "Tests passed."
+exit 0
```
jrusz, line 27: I needed some exit code returned otherwise it would not end when ran over ssh.

jrusz:
I added some comments to the changes I made, please add more comments if you have questions or suggestions for improvement. 

This is intended to run on a slave with kerberos and access to internal network. The way it works is that a s390x machine is reserved from a certain range of machines in beaker, then osbuild-composer repository is cloned there, rpms built and installed and then the selected tests are ran. After that is finished the machine is returned. 

Currently it can run any of the 3 test runs that can be seen in the case statement. Also it's using latest nightly right now and I just created one stage in the Jenkinsfile for testing and only the base test will succeed as the other tests need the changes in this PR to succeed. This certainly needs some more work and polishing and right now I have a few questions. 

1. Do you think it's ok to run the tests like this over ssh?
2. How often do we want to run this and on what composes?
3. Any other thoughts about this?

test/cases/libvirt.sh
atodorov: This comment sounds important. It should be in the source code.

test/cases/image_tests.sh
atodorov: See Jenkinsfile for DISTRO_CODE.  BUILD_ID I believe is the ID of the Jenkins build job. You can replace that with the beaker job id if that's available.

tools/run-s390x-tests.sh
```diff
@@ -0,0 +1,111 @@
...
+      </task>
+    </recipe>
+  </recipeSet>
+</job>
+EOF
+
+# Submit the job
+JOB_ID=$(bkr job-submit beaker.xml | awk -F'['\''|'\'']' '{print $2}')
+
+# Wait for the machine to get provisioned
```
atodorov, line 65: Can you walk me through how the provisioning/test+reporting of the results is intended to happen ? IMO we can eliminate the intermediary PSI machine (or at least keep it around for much shorter) depending on what's intended to happen.

jrusz: I guess you're right as this can easily be run from your local machine with the same results... Provisioning is as simple as submitting the above beaker.xml, which simply reserves a s390x machine that meets the requirements on line 47. That's a range of machines with the same hostname but different numbers, they are all the same and fit our needs. It periodically checks whether the machine is ready or gives up after 2 hours. Also it's reserving it only for 2 hours which should be more then enough time to run the tests. Then everything just happens over ssh, cloning the repository, building the rpms, installing them and then running the selected tests and returns the machine after all is done. It's expected to either just pass or exit with non zero. I guess I'm missing something like `set -e` but for that I would need better checking than what's on line 68 because that will fail the whole script otherwise.

tools/run-s390x-tests.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/bash
+
+# This script prepares testing environment on a remote s390x
+# machine in beaker and runs specified tests on it over ssh.
+
+# Run on lastest nightly
+DISTRO=$(curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.4/COMPOSE_ID)
```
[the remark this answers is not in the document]
jrusz: I see, I'll take a look at that and see how can I use this here.

schutzbot/Jenkinsfile
```diff
@@ -738,6 +738,28 @@ pipeline {
                         }
                     }
                 }
+                stage('EL8.4 s390x Base') {
+                    when {
+                        beforeAgent true
+                        expression {
+                            return env.BUILD_CAUSE != 'cron';
+                        }
```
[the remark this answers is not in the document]
jrusz: I didn't really think about this, I just copied one of the stages here and used it. I wanted to discuss this here how often and where we want to run this.

test/cases/libvirt.sh
```diff
@@ -1,19 +1,27 @@
...
+    /usr/libexec/osbuild-composer-test/libvirt_test.sh qcow2
+    /usr/libexec/osbuild-composer-test/libvirt_test.sh openstack
+    /usr/libexec/osbuild-composer-test/libvirt_test.sh vhd
+fi
 
 # RHEL 8.4 and Centos Stream 8 images also supports uefi, check that
 if [[ "${ID}-${VERSION_ID}" == "rhel-8.4" || "${ID}-${VERSION_ID}" == "centos-8" ]]; then
   echo "🐄 Booting qcow2 image in UEFI mode on RHEL/Centos Stream"
   /usr/libexec/osbuild-composer-test/libvirt_test.sh qcow2 uefi
 fi
```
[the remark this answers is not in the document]
jrusz: It's passing like this so I believe it does.

jrusz:
Since we're going away from using Jenkins this PR no longer makes sense. Testing on s390x will be revisited later. Closing for now.

PR #1330, opened 2021-03-25
osbuild1: make the conversion from v2 result stable

ondrejbudai:
A result from manifest v2 contains logs from pipelines. The individual
pipelines are stored as an object, thus they have no order. Well, at least
in Go, because it doesn't guarantee one when parsing maps, see:

https://github.com/golang/go/issues/27179

Unfortunately, this makes the Result.fromV2 method return unpredictable
results because the pipeline results are processed in basically a random
order.

This caused the TestUnmarshalV2Failure test (result_test.go:124) to randomly
fail because it expects the ordering to be stable. I decided to fix this by
ordering the pipeline results by their name. When this fix is applied, the
output from Result.fromV2 is well-defined.

This pull request includes:

### DIFF
--- a/internal/osbuild1/result.go
+++ b/internal/osbuild1/result.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sort"
 
 	"github.com/osbuild/osbuild-composer/internal/osbuild2"
 )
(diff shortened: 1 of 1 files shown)


msehnout:
Isn't it possible to parse dictionary as a list of tuples instead? Or use an implementation of ordered map, but that would have to be from a 3rd party library ...

ondrejbudai:
It certainly is, although one might argue that it's against the standard: https://www.json.org/json-en.html

> An object is an unordered set of name/value pairs.

teg (approved):

msehnout:
> It certainly is, although one might argue that it's against the standard: https://www.json.org/json-en.html
> 
> > An object is an unordered set of name/value pairs.

Sure, that makes sense. Maybe I should have asked a different question. Is the order of logs important for us? Sorry for my ignorance, but I'm not sure what logs exactly are in this structure.

ondrejbudai:
From my perspective, it's better to define the `Result.fromV2` function as a pure function - it's then easier to test. Of course, we can say that its output is in an undefined order... Not sure what's better, I think I prefer a defined order slightly more.

achilleas-k (approved):
I agree with this.  Sorting might not be ideal—the assembler pipeline is always last for example and now it will almost surely show up first—but it's all we can do right now for order stability.

achilleas-k:
In the long term we can talk about other ways to make this stable:
- change all pipeline names in our image types to have numbered prefixes,
- change the output format coming from osbuild to use arrays of pipelines instead of map/object,
- change the osbuild output completely??

msehnout:
> In the long term we can talk about other ways to make this stable:
> 
>     * change all pipeline names in our image types to have numbered prefixes,
> 
>     * change the output format coming from osbuild to use arrays of pipelines instead of map/object,
> 
>     * change the osbuild output completely??

If we want to use JSON, which I guess we want, and we want the output to be ordered, then using list, as an ordered type, sounds like a logical choice to me.

PR #1335, opened 2021-03-30
GCP: Allow the caller of the internal library to pass the ctx

### LINKED ISSUE #1304: GCP: Allow the caller of the internal library to pass the `ctx`
GCP internal library methods currently create their own context, instead of allowing the caller to pass it to the method. This should be fixed to allow higher-level logic to potentially cancel the operation by cancelling the passed context.

thozza:
Modify all relevant methods in the internal GCP library to accept context from the caller.

Modify all places which call the internal GCP library methods to pass the context.

Made it possible to clean up resources after canceled image import:
- Add method to fetch Cloudbuild job log.
- Add method to parse Cloudbuild job log for created resources. Parsing is specific to the Image import Cloudbuild job and its logs format. Add unit tests for the parsing function.
- Add method to clean up all resources (instances, disks, storage objects) after a Cloudbuild job.
- Modify the worker osbuild job implementation and also the GCP upload CLI tool to use the new cleanup method `CloudbuildBuildCleanup()`.
- Keep the `StorageImageImportCleanup()` method, because it is still used by the cloud-cleaner tool. There is no way for the cloud-cleaner to figure out the Cloudbuild job ID to be able to call `CloudbuildBuildCleanup()` instead.
- Add methods to delete Compute instance and disk.
- Add method to get Compute instance information. This is useful for checking if the instance has been already deleted, or whether it still exists.

Rename all occurrences of "Compute Node" to "Compute Engine". This is an error, there is no such thing as "Compute Node" in GCP.

Fixes #1304 

This pull request includes:

  - unit test was added for code which can be tested this way.
  - no new documentation was added. Changes should be transparent to the user.

### DIFF
--- a/internal/cloud/gcp/cloudbuild.go
+++ b/internal/cloud/gcp/cloudbuild.go
@@ -1,0 +1,226 @@
+package gcp
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+	"time"
+
+	cloudbuild "cloud.google.com/go/cloudbuild/apiv1"
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	cloudbuildpb "google.golang.org/genproto/googleapis/devtools/cloudbuild/v1"
+)
+
+// Structure with resources created by the Build job
+// Intended only for internal use
+type cloudbuildBuildResources struct {
+	zone             string
+	computeInstances []string
+	computeDisks     []string
+	storageCacheDir  struct {
+		bucket string
+		dir    string
+	}
+}
+
+// CloudbuildBuildLog fetches the log for the provided Build ID and returns it as a string
+//
+// Uses:
+//	- Storage API
+//	- Cloud Build API
+func (g *GCP) CloudbuildBuildLog(ctx context.Context, buildID string) (string, error) {
+	cloudbuildClient, err := cloudbuild.NewClient(ctx, option.WithCredentials(g.creds))
+	if err != nil {
+		return "", fmt.Errorf("failed to get Cloud Build client: %v", err)
+	}
+	defer cloudbuildClient.Close()
+
+	storageClient, err := storage.NewClient(ctx, option.WithCredentials(g.creds))
+	if err != nil {
+		return "", fmt.Errorf("failed to get Storage client: %v", err)
+	}
+	defer storageClient.Close()
+
+	getBuldReq := &cloudbuildpb.GetBuildRequest{
+		ProjectId: g.creds.ProjectID,
+		Id:        buildID,
+	}
+
+	imageBuild, err := cloudbuildClient.GetBuild(ctx, getBuldReq)
+	if err != nil {
+		return "", fmt.Errorf("failed to get the build info: %v", err)
+	}
+
+	// Determine the log file's Bucket and Object name
+	// Logs_bucket example: "gs://550072179371.cloudbuild-logs.googleusercontent.com"
+	// Logs file names will be of the format `${logs_bucket}/log-${build_id}.txt`
+	logBucket := imageBuild.LogsBucket
+	logBucket = strings.TrimPrefix(logBucket, "gs://")
+	// logBucket may contain directory in its name if set to a custom value
+	var logObjectDir string
+	if strings.Contains(logBucket, "/") {
+		ss := strings.SplitN(logBucket, "/", 2)
+		logBucket = ss[0]
+		logObjectDir = fmt.Sprintf("%s/", ss[1])
+	}
+	logObject := fmt.Sprintf("%slog-%s.txt", logObjectDir, buildID)
+
+	// Read the log
+	logBuilder := new(strings.Builder)
+	rd, err := storageClient.Bucket(logBucket).Object(logObject).NewReader(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to create a new Reader for object '%s/%s': %v", logBucket, logObject, err)
+	}
+	_, err = io.Copy(logBuilder, rd)
+	if err != nil {
+		return "", fmt.Errorf("reading data from object '%s/%s' failed: %v", logBucket, logObject, err)
+	}
+
+	return logBuilder.String(), nil
+}
+
+// CloudbuildBuildCleanup parses the logs for the specified Build job and tries to clean up all resources
+// which were created as part of the job. It returns list of strings with all resources that were deleted
+// as a result of calling this method.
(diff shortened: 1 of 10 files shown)


internal/cloud/gcp/cloudbuild.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get the build info: %v", err)
+	}
+
+	// Determine the log file's Bucket and Object name
+	// Logs_bucket example: "gs://550072179371.cloudbuild-logs.googleusercontent.com"
+	// Logs file names will be of the format `${logs_bucket}/log-${build_id}.txt`
+	logBucket := imageBuild.LogsBucket
+	logBucket = strings.TrimPrefix(logBucket, "gs://")
+	// logBucket may can contain directory in its name if set to a custom value
```
msehnout, line 63: typo? either "may" or "can"

internal/cloud/gcp/cloudbuild.go
```diff
@@ -0,0 +1,224 @@
...
+		err = g.ComputeInstanceDelete(ctx, resources.zone, instance)
+		if err == nil {
+			deletedResources = append(deletedResources, fmt.Sprintf("instance: %s (%s)", instance, resources.zone))
+		}
+	}
+
+	// Deleting instances in reality takes some time. Deleting a disk while it is still used by the instance, will fail.
+	// Iterate over the list of instances and wait until they are all deleted.
+	for _, instance := range resources.computeInstances {
+		for {
```
msehnout, line 124: Did you consider a timeout for this loop?

internal/cloud/gcp/cloudbuild.go
```diff
@@ -0,0 +1,224 @@
...
+			// Getting the instance information failed, it is ideleted.
+			if err != nil {
+				break
+			}
+			time.Sleep(1 * time.Second)
+		}
+	}
+
+	// Delete all Compute Engine Disks
+	// Deleting Compute instances takes time, therefore it may happen that deleting the disk will fail, because
```
msehnout, line 135: This shouldn't happen if the above loop succeeded, right?

internal/cloud/gcp/cloudbuild.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+
+	return deletedResources, nil
+}
+
+// cloudbuildResourcesFromBuildLog parses the provided Cloud Build log for any
+// resources that were created by the job as part of its work. The list of extracted
+// resources is returned as cloudbuildBuildResources struct instance
+func cloudbuildResourcesFromBuildLog(buildLog string) (*cloudbuildBuildResources, error) {
```
msehnout, line 169: Wow, there are no functions for this in the official GCP API?

msehnout (approved):
Looks very nice, I only have few questions/comments. :+1:

internal/cloud/gcp/compute.go
```diff
@@ -105,6 +108,24 @@ func (g *GCP) ComputeImageImport(bucket, object, imageName, os, region string) (
 
 	// Wait for the build to finish
 	for {
+		select {
+		case <-time.After(30 * time.Second):
+			// Just check the build status below
+		case <-ctx.Done():
+			// cancel the build
+			cancelBuildReq := &cloudbuildpb.CancelBuildRequest{
```
ondrejbudai, line 116: If I understand this correctly, in case of `ctx.Done()`, we still return `err == nil` which means that the upload will be marked as a success. I might be missing something though.

internal/cloud/gcp/cloudbuild.go
thozza: Good catch!

internal/cloud/gcp/cloudbuild.go
thozza: Not really. Instances are deleted in the step before and if the code does not wait for the GCP service to finish deleting them, then the subsequent deletion of disks will fail if the instance is still not deleted. While an instance can have "Deletion Protection" set and in this case we could end up in an infinite loop, it is not set on the instances created by the image import job.

I'm open to suggestion how to make the code more robust. However I don't think that timeout makes sense here. Mainly because timing out will potentially cause the subsequent disk deletions to fail. Maybe at least check whether the instance does not have deletion protection set?

internal/cloud/gcp/cloudbuild.go
thozza: Yes. This is mostly FYI, but I should probably remove it from here, because the same information is already included in the comment above the previous loop.

internal/cloud/gcp/cloudbuild.go
thozza: No, more info in https://issuetracker.google.com/issues/182680680

internal/cloud/gcp/compute.go
thozza: Great question!

In case it is canceled, an error will be returned. On line no. 139, there is a check of the actual build job status and if it is not `cloudbuildpb.Build_SUCCESS`, an error is returned in all cases. In case the job is canceled, its status will be `cloudbuildpb.Build_CANCELLED`

ondrejbudai: Ah, I see, thanks!

internal/cloud/gcp/cloudbuild.go
thozza: Fixed

internal/cloud/gcp/cloudbuild.go
thozza: I added check if the instance is protected against deletion so prevent a potential infinite loop. I don't think that having a timeout here makes sense, but I'm open to more feedback and also to changing my opinion.

internal/cloud/gcp/cloudbuild.go
thozza: fixed

internal/cloud/gcp/cloudbuild.go
thozza: Just to have a complete information, the GCP has a [shutdown period](https://cloud.google.com/compute/docs/instances/deleting-instance#delete_timeout), which is in most cases 90 seconds. If the instance does not shut down gracefully within that time, it will shut it down forcefully. Since all instances are deleted first and only after that the code waits for the shutdown to complete, the combined maximum time that it can take should be also around 90 seconds. The reason is that all of the instances will be effectively shutting down in parallel.

msehnout: Thanks for the detailed explanation!

PR #1349, opened 2021-04-09
distro: remove excluded package if explicitly specified in the bp

msehnout:
When a users wants to install a package that itself is excluded or its
dependency is excluded, it fails the build. There is no known workaround
for this shorcoming of our current design.

Therefore, remove a package from the list of excluded if it is
explicitly mentioned in a blueprint. This will not solve the issue with
dependencies, but it will create a possibility of a workaround.

This pull request includes:

### DIFF
--- a/internal/distro/rhel84/distro.go
+++ b/internal/distro/rhel84/distro.go
@@ -217,7 +217,18 @@ func (t *imageType) Packages(bp blueprint.Blueprint) ([]string, []string) {
 		packages = removePackage(packages, "insights-client")
 	}
 
-	return packages, t.excludedPackages
+	// copy the list of excluded packages from the image type
+	// and subtract any packages found in the blueprint (this
+	// will not handle the issue with dependencies present in
+	// the list of excluded packages, but it will create a
+	// possibility of a workaround at least)
+	excludedPackages := append([]string(nil), t.excludedPackages...)
+	for _, pkg := range bp.GetPackages() {
+		// removePackage is fine if the package doesn't exist
+		excludedPackages = removePackage(excludedPackages, pkg)
+	}
+
+	return packages, excludedPackages
 }
 
 func (t *imageType) BuildPackages() []string {
--- a/test/cases/regression.sh
+++ b/test/cases/regression.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# Provision the software under tet.
+/usr/libexec/osbuild-composer-test/provision.sh
+
+BLUEPRINT_FILE=/tmp/blueprint.toml
+COMPOSE_START=/tmp/compose-start.json
+COMPOSE_INFO=/tmp/compose-info.json
+
+# Write a basic blueprint for our image.
+tee "$BLUEPRINT_FILE" > /dev/null << EOF
+name = "redhat-lsb-core"
+description = "A base system with redhat-lsb-core"
+version = "0.0.1"
+
+[[packages]]
+name = "redhat-lsb-core"
+
+[[packages]]
+# The nss package is excluded in the RHEL8.4 image type, but it is required by the
+# redhat-lsb-core package. This test verifies it can be added again when explicitly
+# mentioned in the blueprint.
+name = "nss"
+EOF
+
+sudo composer-cli blueprints push "$BLUEPRINT_FILE"
+sudo composer-cli blueprints depsolve redhat-lsb-core
+sudo composer-cli --json compose start redhat-lsb-core qcow2 | tee "${COMPOSE_START}"
+COMPOSE_ID=$(jq -r '.build_id' "$COMPOSE_START")
+
+# Wait for the compose to finish.
+echo "⏱ Waiting for compose to finish: ${COMPOSE_ID}"
+while true; do
+    sudo composer-cli --json compose info "${COMPOSE_ID}" | tee "$COMPOSE_INFO" > /dev/null
+    COMPOSE_STATUS=$(jq -r '.queue_status' "$COMPOSE_INFO")
+
+    # Is the compose finished?
+    if [[ $COMPOSE_STATUS != RUNNING ]] && [[ $COMPOSE_STATUS != WAITING ]]; then
+        break
+    fi
+
+    # Wait 30 seconds and try again.
+    sleep 30
+done
+
+jq . "${COMPOSE_INFO}"
+
+# Did the compose finish with success?
+if [[ $COMPOSE_STATUS != FINISHED ]]; then
+    echo "Something went wrong with the compose. 😢"
+    exit 1
+fi
--- a/internal/distro/fedora32/distro.go
+++ b/internal/distro/fedora32/distro.go
@@ -51,6 +51,19 @@ type imageType struct {
 	assembler        func(uefi bool, options distro.ImageOptions, arch distro.Arch) *osbuild.Assembler
 }
 
+func removePackage(packages []string, packageToRemove string) []string {
+	for i, pkg := range packages {
+		if pkg == packageToRemove {
+			// override the package with the last one from the list
+			packages[i] = packages[len(packages)-1]
+
+			// drop the last package from the slice
+			return packages[:len(packages)-1]
+		}
+	}
+	return packages
+}
+
 func (a *architecture) Distro() distro.Distro {
 	return a.distro
 }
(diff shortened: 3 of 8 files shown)


teg:
I would love to see this fixed. As you say, this still leaves the issue with a dependency of an included package being excluded. I think there are two other potential solutions:

1) We stop including the `core` group, and instead list all the packages. We also disable weak dependencies if necessary. This would allow us to stop excluding any packages. The downside is that we would not automatically reflect changes in the `core` group. Note that this is already how `rhel-for-edge` is done.
2) We could append several package sets to the dnf transaction, first the base one (with excludes), then the user provided ones. I think @gicmo looked into that and said it would work, but I may misremember.

msehnout:
> I would love to see this fixed. As you say, this still leaves the issue with a dependency of an included package being excluded. I think there are two other potential solutions:
> 
>     1. We stop including the `core` group, and instead list all the packages. We also disable weak dependencies if necessary. This would allow us to stop excluding any packages. The downside is that we would not automatically reflect changes in the `core` group. Note that this is already how `rhel-for-edge` is done.
> 
>     2. We could append several package sets to the dnf transaction, first the base one (with excludes), then the user provided ones. I think @gicmo looked into that and said it would work, but I may misremember.

The second one sounds very nice, I'll look into it. Nevertheless would it make sense to you to have both the solution I've currently implemented and one of those you propose? Unless we have 0 excluded packages, this would still make sense, wouldn't it?

thozza:
The change IMO makes sense at least as a temporary workaround. If the package specified as "wanted" by the user is part of our Image Type's `PackageSet()` exclude list, this change allows the user to have this package installed on the image. Installation of the user-requested package can however still fail if its dependency is excluded. Nevertheless this change allows the user to workaround this case as well by specifying this dependency also as explicitly "wanted", thus removing it from the list of packages to exclude.

internal/distro/rhel84/distro.go
```diff
@@ -218,6 +218,19 @@ func (t *imageType) Packages(bp blueprint.Blueprint) ([]string, []string) {
...
+	// copy the list of excluded packages from the image type
+	// and subtract any packages found in the blueprint (this
+	// will not handle the issue with dependencies present in
+	// the list of excluded packages, but it will create a
+	// possibility of a workaround at least)
+	excludedPackages := append([]string(nil), t.excludedPackages...)
+	for _, pkg := range bp.GetPackages() {
+		// removePackage is fine if the package doesn't exist
+		excludedPackages = removePackage(excludedPackages, pkg)
+	}
```
achilleas-k, line 230: Had an immediate reaction to this nested iteration here.  I guess blueprints don't tend to have hundreds of packages listed in them, and we don't exclude more than a handful, so in the end it's fine.

achilleas-k:
> 2\. We could append several package sets to the dnf transaction, first the base one (with excludes), then the user provided ones. I think @gicmo looked into that and said it would work, but I may misremember.

I like this idea.  It's straightforward and simple enough.

internal/distro/rhel84/distro.go
msehnout: Exactly and Go is said to be fast, right? :-D

msehnout:
Just for the record: The error message we return is pretty helpful. If we had this workaround it should be simple enough to exclude the conflicting packages by hand.
```
[2021-04-12T15:46:02.568Z] + sudo composer-cli --json compose start redhat-lsb-core qcow2
[2021-04-12T15:46:02.568Z] + tee /tmp/compose-start.json
[2021-04-12T15:46:03.137Z] {
[2021-04-12T15:46:03.137Z]     "status": false,
[2021-04-12T15:46:03.137Z]     "errors": [
[2021-04-12T15:46:03.137Z]         {
[2021-04-12T15:46:03.137Z]             "id": "DepsolveError",
[2021-04-12T15:46:03.137Z]             "msg": "DNF error occured: DepsolveError: There was a problem depsolving ['@core', 'authselect-compat', 'chrony', 'cloud-init', 'cloud-utils-growpart', 'cockpit-system', 'cockpit-ws', 'dhcp-client', 'dnf', 'dnf-utils', 'dosfstools', 'dracut-norescue', 'insights-client', 'NetworkManager', 'net-tools', 'nfs-utils', 'oddjob', 'oddjob-mkhomedir', 'psmisc', 'python3-jsonschema', 'qemu-guest-agent', 'redhat-release', 'redhat-release-eula', 'rsync', 'subscription-manager-cockpit', 'tar', 'tcpdump', 'yum', 'redhat-lsb-core', 'kernel', 'dracut-config-generic', 'grub2-pc', 'grub2-efi-x64', 'shim-x64']: \n Problem: conflicting requests\n  - package redhat-lsb-core-4.1-47.el8.i686 requires redhat-lsb-submod-security(x86-32) = 4.1-47.el8, but none of the providers can be installed\n  - package redhat-lsb-core-4.1-47.el8.x86_64 requires redhat-lsb-submod-security(x86-64) = 4.1-47.el8, but none of the providers can be installed\n  - package redhat-lsb-submod-security-4.1-47.el8.i686 requires nss(x86-32), but none of the providers can be installed\n  - package redhat-lsb-submod-security-4.1-47.el8.x86_64 requires nss(x86-64), but none of the providers can be installed\n  - package nss-3.53.1-17.el8.i686 is filtered out by exclude filtering\n  - package nss-3.53.1-17.el8.x86_64 is filtered out by exclude filtering"
[2021-04-12T15:46:03.137Z]         }
[2021-04-12T15:46:03.137Z]     ]
[2021-04-12T15:46:03.137Z] }
```

internal/distro/rhel84/distro.go
ondrejbudai: Yeah, Go is so fast that it can do `O(n^2)` in `O(n)` :-D /s

ondrejbudai (changes requested):
This is a reasonable workaround covering some of the cases. I don't see a reason to not include it.

I would suggest changing the test though. PSI resources are scarce and I overall prefer to run tests on Fedora/CentOS Stream just because they're more available to contributors. I see two options:

1) Running the new test on Fedora/CS on AWS.
2) Given that the test itself takes only 5 minutes, I think we can even include it in the base tests. Base tests are usually the fastest tests so I'm not worried too much about making them bigger. We can always gather more data about the runtime of our test suite and split it.

msehnout:
> This is a reasonable workaround covering some of the cases. I don't see a reason to not include it.
> 
> I would suggest changing the test though. PSI resources are scarce and I overall prefer to run tests on Fedora/CentOS Stream just because they're more available to contributors. I see two options:
> 
>     1. Running the new test on Fedora/CS on AWS.

Sounds good :+1: 

> 
>     2. Given that the test itself takes only 5 minutes, I think we can even include it in the base tests. Base tests are usually the fastest tests so I'm not worried too much about making them bigger. We can always gather more data about the runtime of our test suite and split it.

I'm ok with running the test on the same machine but in such case I'd not call them "base" tests anymore to prevent confusion.

achilleas-k:
>     2\. We could append several package sets to the dnf transaction, first the base one (with excludes), then the user provided ones. I think @gicmo looked into that and said it would work, but I may misremember.

Sorry if I'm repeating stuff but I can't remember why (or if) we dismissed this idea.  What would happen if we added the blueprint packages as a separate package set, depsolved it, and then merged that package set into the base packages during manifest creation?

msehnout:
> > ```
> > 2\. We could append several package sets to the dnf transaction, first the base one (with excludes), then the user provided ones. I think @gicmo looked into that and said it would work, but I may misremember.
> > ```
> 
> Sorry if I'm repeating stuff we already covered but I can't remember why (or if) we dismissed this idea. What would happen if we added the blueprint packages as a separate package set, depsolved it, and then merged that package set into the base packages during manifest creation?

According to my research of the dnf API, communication with the dnf team, and discussion with Christian there is no way to do this. We could do something like this, but that doesn't look nice IMO: https://github.com/weldr/lorax/blob/46ab9e55f553667ef8538a1002272fe365272574/src/pylorax/ltmpl.py#L581

That said, I'm all in for suggestions!

msehnout:
@ondrejbudai : I've made the changes you suggested. I also had to stop using the `run_tests` function in the Jenkinsfile because running it twice resulted in failure. The `ci_details.sh` script failed on:
```bash
if rpm --quiet -q python36; then
    echo -e "\n FAIL: python36 is installed, see #794 ..."
    exit 1
```
because after running base tests, it was already installed.

test/cases/regression.sh
```diff
@@ -0,0 +1,50 @@
...
+tee "$BLUEPRINT_FILE" > /dev/null << EOF
+name = "redhat-lsb-core"
+description = "A base system with redhat-lsb-core"
+version = "0.0.1"
+
+[[packages]]
+name = "redhat-lsb-core"
+
+[[packages]]
+name = "nss"
```
ondrejbudai, line 21: I would like a comment here that explains that `nss` is excluded from RHEL 8.4 and this test proves that it can re-added into a package set using a blueprint.

msehnout: Fixed :+1:

msehnout:
CI failure should be unrelated, any additional concerns @ondrejbudai ?

thozza (approved):
Changes look good to me and this approach has been discussed as appropriate fix / workaround for now. The CI failure is caused by the ostree test requiring repository, which is not available any more.

PR #1350, opened 2021-04-10
distro/rhel84: build qcow2 images with compat=0.10

ondrejbudai:
*"backport" from the rhel-8.4.0 branch*

By default, `qemu-img convert` creates qcow2 images usable in qemu 1.1 and
newer. RHEL 8 guest images are meant to be bootable on RHEL 6 though.
Unfortunately, RHEL 6 has qemu 0.12, therefore these images cannot be used
there.

To fix this, we need to use the new qcow2_compat option in qemu assembler
to override the default compat version and make qcow2 images that can be used
in qemu 0.10 and newer.

For this, we need osbuild 28 that is currently only available in RHEL 8.4,
therefore it's being pinned for now in Fedora, CentOS Stream and RHEL 8.3

This pull request includes:

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -1,5 +1,17 @@
 {
+  "fedora-32": {
+    "dependencies": {
+      "osbuild": {
+        "commit": "eb74ddf2ef8c203b5d354dfb97adee65babdc147"
+      }
+    }
+  },
   "fedora-33": {
+    "dependencies": {
+      "osbuild": {
+        "commit": "eb74ddf2ef8c203b5d354dfb97adee65babdc147"
+      }
+    },
     "dependants": {
       "koji-osbuild": {
         "commit": "4fdc457745e1147475ea3ac1e3b073e592d7b174"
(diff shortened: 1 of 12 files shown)


gicmo (approved):
Thanks!

teg (dismissed):
Thanks!

ondrejbudai:
Ah, I need to pin osbuild also in RHEL 8.4 because it only has `27.2`. We might be fine with `Requires: osbuild >= 27.2` here but I think that depending on `osbuild 28` is just cleaner.

teg:
We shouldn't be testing 8.4 on `main` anymore though, should we?

ondrejbudai:
Let's leave it as follow-up?

ondrejbudai:
I would also argue that main should still be able to build 8.4 so we need some coverage.

teg:
Yeah, leave it as a follow-up. But I think the test should be: `main` should be able to build all the `rhel` verions on the `koji` API against the most recent `osbuild` major release on a RHEL8.3 host. Needs some work to make this elegant in shutzbot though.

teg (approved):

Issue #1318, opened 2021-03-22
tests: provision.sh overrides RHEL repositories twice - this leads to problems


atodorov:
In provision.sh we have:

```
     54 # RHEL nightly repos need to be overriden in rhel-8.json and rhel-8-beta.json
     55 case "${ID}-${VERSION_ID}" in
     56     "rhel-8.4")
     57         # Override old rhel-8.json and rhel-8-beta.json because RHEL 8.4 test needs nightly repos
     58         sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-84.json /etc/osbuild-composer/repositories/rhel-8.json
     59         # If multiple tests are run and call provision.sh the symlink will need to be overriden with -f
     60         sudo ln -sf /etc/osbuild-composer/repositories/rhel-8.json /etc/osbuild-composer/repositories/rhel-8-beta.json;;
     61     *) ;;
     62 esac
     63 
     64 # overrides for RHEL nightly builds testing
     65 if [ -f "rhel-8.json" ]; then
     66     sudo mv rhel-8.json /etc/osbuild-composer/repositories/
     67 fi
     68 
     69 if [ -f "rhel-8-beta.json" ]; then
     70     sudo mv rhel-8-beta.json /etc/osbuild-composer/repositories/
     71 fi
```

Lines 55-65 I believe take the repository definitions from `test/data/repositories/rhel-8*` (points to latest internal trees) while lines 65-70 rely on the information generated by `schutzbot/prepare-rhel-internal.sh` which pins the URLs to the current compose which is being tested (usually the latest but can be overridden). 

This leads to a discrepancy such as in #1317 where all changes for the missing RT repository (for kernel-rt RPM) have been added and PRs are passing, everything has been backported and passing but the nightly test job still fails. 

FTR the changes are in 28.1 where as the nightly uses 28.2 and still fails.

A secondary issue is that we haven't tested the relevant changes (e.g. #1243, #1244) against the nightly trees because we didn't have a nightly tree containing the RPM. This made it impossible to discover the failure earlier and necessitates more changes after we thought we were done with the rebase. This applies to all PRs in general, not only to those 2. 

IOW the existing CI configuration tells us that everything works fine when built from source but doesn't tell us if this is going to continue working fine once the RPMs land into the nightly trees.

atodorov:
Also see the `rhel-85.json` file added by https://github.com/osbuild/osbuild-composer/pull/1391 and 1351 above.

PR #1274, opened 2021-03-08
8.4 cleanup after transitioning to 8.5 

atodorov:
This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change

PR #1351, opened 2021-04-12
Edge image types for RHEL 8.5

achilleas-k:
This pull request includes:

## Summary

Adding new distro RHEL 8.5 with support for the three Edge image types: `edge-commit`, `edge-container`, and `edge-installer`.

## New pipeline definition

The definition of the image types differs from older implementations:
- Reusable functions define alternatives to the `pipelines()` function for different image types.  These are bound to the image type instance on creation.  This pattern avoids the frequent conditional branching we ended up having in older pipeline generation functions.  This is especially visible in the definition of the `edge-installer` pipelines that share almost no common components with the other types (only the build pipeline is the same).  Similarly, the other (general) image types will be quite different from the edge image types when they're added, so they can be defined in their own functions while reusing common pipelines.
- Pipelines, options, and inputs are in different files in the package.  A minor change, but I think this makes navigating the code easier.

We could have each image type have its own struct implementation of the `ImageType` interface with its own `pipelines()` method, but then we would need to rewrite all the other common methods as well (e.g., `Arch()`, `Name()`, `PackageSets()`).  I prefer the way I did it now, but I'm open to hearing different opinions on this.

I'm quite happy with the way the pipelines are defined now.  It makes it very easy to look at a `pipelines()` function and see the list of pipelines that it contains.  Same goes for looking at every individual pipeline and seeing the stages.  

A big part of the redesign came out of discussions with @thozza (thanks Tomáš) about having a more declarative implementation of the manifests in osbuild-compser.

## Image names

As discussed in other channels, the new image types are named `edge-*`, without the `rhel-` prefix that they have in 8.4.

## Fixes to 8.4

There's a small fix for the `rhel-edge-container` image type in 8.4: The commit's parent is no longer added to the manifest's sources.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -133,7 +133,7 @@ OSTree:
           # See COMPOSER-919
           # - openstack/fedora-34-x86_64
           - openstack/rhel-8-x86_64
-          - openstack/rhel-8.4-x86_64
+          - openstack/rhel-8.5-x86_64
           # do we want centos as well ?
           # do we want secondary architectures here ??
 
@@ -146,7 +146,7 @@ New OSTree:
   parallel:
     matrix:
       - RUNNER:
-          - openstack/rhel-8.4-x86_64
+          - openstack/rhel-8.5-x86_64
 
 .INTEGRATION_TESTS: &INTEGRATION_TESTS
   SCRIPT:
--- a/internal/distro/rhel85/distro.go
+++ b/internal/distro/rhel85/distro.go
@@ -298,14 +298,229 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 		ostreeRef:        ostreeRef,
 	}
 
+	// Shared Package sets
+	edgeCommitCommonPkgSet := rpmmd.PackageSet{
+		Include: []string{
+			"redhat-release",
+			"glibc", "glibc-minimal-langpack", "nss-altfiles",
+			"dracut-config-generic", "dracut-network",
+			"basesystem", "bash", "platform-python",
+			"shadow-utils", "chrony", "setup", "shadow-utils",
+			"sudo", "systemd", "coreutils", "util-linux",
+			"curl", "vim-minimal",
+			"rpm", "rpm-ostree", "polkit",
+			"lvm2", "cryptsetup", "pinentry",
+			"e2fsprogs", "dosfstools",
+			"keyutils", "gnupg2",
+			"attr", "xz", "gzip",
+			"firewalld", "iptables",
+			"NetworkManager", "NetworkManager-wifi", "NetworkManager-wwan",
+			"wpa_supplicant",
+			"dnsmasq", "traceroute",
+			"hostname", "iproute", "iputils",
+			"openssh-clients", "procps-ng", "rootfiles",
+			"openssh-server", "passwd",
+			"policycoreutils", "policycoreutils-python-utils",
+			"selinux-policy-targeted", "setools-console",
+			"less", "tar", "rsync",
+			"fwupd", "usbguard",
+			"bash-completion", "tmux",
+			"ima-evm-utils",
+			"audit",
+			"podman", "container-selinux", "skopeo", "criu",
+			"slirp4netns", "fuse-overlayfs",
+			"clevis", "clevis-dracut", "clevis-luks",
+			"greenboot", "greenboot-grub2", "greenboot-rpm-ostree-grub2", "greenboot-reboot", "greenboot-status",
+		},
+		Exclude: []string{"rng-tools"},
+	}
+	edgeBuildPkgSet := rpmmd.PackageSet{
+		Include: []string{
+			"dnf", "dosfstools", "e2fsprogs", "efibootmgr", "genisoimage",
+			"grub2-efi-ia32-cdboot", "grub2-efi-x64", "grub2-efi-x64-cdboot",
+			"grub2-pc", "grub2-pc-modules", "grub2-tools", "grub2-tools-efi",
+			"grub2-tools-extra", "grub2-tools-minimal", "isomd5sum",
+			"lorax-templates-generic", "lorax-templates-rhel",
+			"policycoreutils", "python36", "python3-iniparse", "qemu-img",
+			"rpm-ostree", "selinux-policy-targeted", "shim-ia32", "shim-x64",
+			"squashfs-tools", "syslinux", "syslinux-nonlinux", "systemd",
+			"tar", "xfsprogs", "xorriso", "xz",
+		},
+		Exclude: nil,
+	}
+	edgeInstallerPkgSet := rpmmd.PackageSet{
+		Include: []string{
+			"aajohan-comfortaa-fonts", "abattis-cantarell-fonts",
+			"alsa-firmware", "alsa-tools-firmware", "anaconda",
+			"anaconda-dracut", "anaconda-install-env-deps", "anaconda-widgets",
+			"audit", "bind-utils", "biosdevname", "bitmap-fangsongti-fonts",
+			"bzip2", "cryptsetup", "curl", "dbus-x11", "dejavu-sans-fonts",
+			"dejavu-sans-mono-fonts", "device-mapper-persistent-data",
+			"dmidecode", "dnf", "dracut-config-generic", "dracut-network",
+			"dump", "efibootmgr", "ethtool", "ftp", "gdb-gdbserver", "gdisk",
+			"gfs2-utils", "glibc-all-langpacks",
+			"google-noto-sans-cjk-ttc-fonts", "grub2-efi-ia32-cdboot",
+			"grub2-efi-x64-cdboot", "grub2-tools", "grub2-tools-efi",
+			"grub2-tools-extra", "grub2-tools-minimal", "grubby",
+			"gsettings-desktop-schemas", "hdparm", "hexedit", "hostname",
+			"initscripts", "ipmitool", "iwl1000-firmware", "iwl100-firmware",
+			"iwl105-firmware", "iwl135-firmware", "iwl2000-firmware",
+			"iwl2030-firmware", "iwl3160-firmware", "iwl3945-firmware",
+			"iwl4965-firmware", "iwl5000-firmware", "iwl5150-firmware",
+			"iwl6000-firmware", "iwl6000g2a-firmware", "iwl6000g2b-firmware",
+			"iwl6050-firmware", "iwl7260-firmware", "jomolhari-fonts",
+			"kacst-farsi-fonts", "kacst-qurn-fonts", "kbd", "kbd-misc",
+			"kdump-anaconda-addon", "kernel", "khmeros-base-fonts", "less",
+			"libblockdev-lvm-dbus", "libertas-sd8686-firmware",
+			"libertas-sd8787-firmware", "libertas-usb8388-firmware",
+			"libertas-usb8388-olpc-firmware", "libibverbs",
+			"libreport-plugin-bugzilla", "libreport-plugin-reportuploader",
+			"libreport-rhel-anaconda-bugzilla", "librsvg2", "linux-firmware",
+			"lklug-fonts", "lohit-assamese-fonts", "lohit-bengali-fonts",
+			"lohit-devanagari-fonts", "lohit-gujarati-fonts",
+			"lohit-gurmukhi-fonts", "lohit-kannada-fonts", "lohit-odia-fonts",
+			"lohit-tamil-fonts", "lohit-telugu-fonts", "lsof", "madan-fonts",
+			"memtest86+", "metacity", "mtr", "mt-st", "net-tools", "nfs-utils",
+			"nmap-ncat", "nm-connection-editor", "nss-tools",
+			"openssh-clients", "openssh-server", "oscap-anaconda-addon",
+			"ostree", "pciutils", "perl-interpreter", "pigz", "plymouth",
+			"prefixdevname", "python3-pyatspi", "rdma-core",
+			"redhat-release-eula", "rng-tools", "rpcbind", "rpm-ostree",
+			"rsync", "rsyslog", "selinux-policy-targeted", "sg3_utils",
+			"shim-ia32", "shim-x64", "sil-abyssinica-fonts",
+			"sil-padauk-fonts", "sil-scheherazade-fonts", "smartmontools",
+			"smc-meera-fonts", "spice-vdagent", "strace", "syslinux",
+			"systemd", "system-storage-manager", "tar",
+			"thai-scalable-waree-fonts", "tigervnc-server-minimal",
+			"tigervnc-server-module", "udisks2", "udisks2-iscsi", "usbutils",
+			"vim-minimal", "volume_key", "wget", "xfsdump", "xfsprogs",
+			"xorg-x11-drivers", "xorg-x11-fonts-misc", "xorg-x11-server-utils",
+			"xorg-x11-server-Xorg", "xorg-x11-xauth", "xz",
+		},
+		Exclude: nil,
+	}
+	edgeCommitX86PkgSet := rpmmd.PackageSet{
+		Include: append(edgeCommitCommonPkgSet.Include,
+			// x86 specific
+			"grub2", "grub2-efi-x64", "efibootmgr", "shim-x64",
+			"microcode_ctl", "iwl1000-firmware", "iwl100-firmware",
+			"iwl105-firmware", "iwl135-firmware", "iwl2000-firmware",
+			"iwl2030-firmware", "iwl3160-firmware", "iwl5000-firmware",
+			"iwl5150-firmware", "iwl6000-firmware", "iwl6050-firmware",
+			"iwl7260-firmware"),
+		Exclude: edgeCommitCommonPkgSet.Exclude,
+	}
+	edgeCommitAarch64PkgSet := rpmmd.PackageSet{
+		Include: append(edgeCommitCommonPkgSet.Include,
+			// aarch64 specific
+			"grub2-efi-aa64", "efibootmgr", "shim-aa64",
+			"iwl7260-firmware"),
+		Exclude: edgeCommitCommonPkgSet.Exclude,
+	}
+
+	// Shared Services
+	edgeServices := []string{
+		"NetworkManager.service", "firewalld.service", "sshd.service",
+	}
+
+	// Image Definitions
+	edgeCommitImgTypeX86_64 := imageType{
+		name:     "edge-commit",
+		filename: "commit.tar",
+		mimeType: "application/x-tar",
+		packageSets: map[string]rpmmd.PackageSet{
+			"build":    edgeBuildPkgSet,
+			"packages": edgeCommitX86PkgSet,
+		},
+		enabledServices: edgeServices,
+		rpmOstree:       true,
+		pipelines:       edgeCommitPipelines,
+		exports:         []string{"commit-archive"},
+	}
+	edgeOCIImgTypeX86_64 := imageType{
+		name:     "edge-container",
+		filename: "container.tar",
+		mimeType: "application/x-tar",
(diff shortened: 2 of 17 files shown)


jkozol:
Great stuff! The naming changes will require an update in cockpit-composer before it can release into 8.5. I captured this [here](https://github.com/osbuild/cockpit-composer/issues/1283) and will soon create a PR.

ondrejbudai:
Looks good so far! :)

achilleas-k:
Update on the status here:
- Testing 8.5 needs an osbuild nightly build.
- osbuild nightly build depends on https://github.com/osbuild/osbuild/pull/635 being merged so we can pin the post-merge commit here.
- osbuild PR is failing on openstack boot tests.

ondrejbudai:
osbuild/osbuild#635 merged :)

internal/distroregistry/distroregistry.go
```diff
@@ -21,6 +22,7 @@ var supportedDistros = []func() distro.Distro{
 	rhel8.New,
 	rhel84.New,
 	rhel84.NewCentos,
```
teg, line 24: Centos Should be moved to 8.5 now too, right?

internal/distro/rhel85/distro.go
```diff
@@ -163,7 +163,20 @@ func (t *imageType) Size(size uint64) uint64 {
...
-	return nil
+	sets := make(map[string]rpmmd.PackageSet)
+	for name, pkgSet := range t.packageSets {
+		if name == "packages" {
+			// combine image packages with blueprint
+			pkgSet.Include = append(pkgSet.Include, bp.GetPackages()...)
+			timezone, _ := bp.Customizations.GetTimezoneSettings()
+			if timezone != nil {
+				pkgSet.Include = append(pkgSet.Include, "chrony")
+			}
```
teg, line 168: Don't we also include the kernel here, or do I misremember?

internal/distro/rhel85/distro.go
```diff
@@ -1036,7 +1052,7 @@ func New() distro.Distro {
 	edgeInstallerImgTypeX86_64 := imageType{
 		name:     "edge-installer",
 		filename: "installer.iso",
-		mimeType: "application/x-tar",
+		mimeType: "application/x-iso9660-image",
```
teg, line 442: Bug fix of previous commit?

internal/distro/rhel85/distro.go
```diff
@@ -1089,7 +1105,7 @@ func New() distro.Distro {
 	edgeInstallerImgTypeAarch64 := imageType{
 		name:     "edge-installer",
 		filename: "installer.iso",
-		mimeType: "application/x-tar",
+		mimeType: "application/x-iso9660-image",
```
teg, line 498: Ditto?

internal/distro/rhel84/distro_v2.go
```diff
@@ -127,7 +127,7 @@ func (t *imageTypeS2) Manifest(c *blueprint.Customizations,
 	}
 
 	var commits []ostreeCommit
-	if options.OSTree.Parent != "" && options.OSTree.URL != "" {
+	if t.bootISO && options.OSTree.Parent != "" && options.OSTree.URL != "" {
```
teg, line 130: Even though changing old distro definitions should be a no-no, and also should have no effect as we don't plan to do any more 8.4 releases, I think this is ok, as it only avoids a download but otherwise should have no effect. Want to point out that in general we should avoid it though.

teg (commented):
Looks great! I just had a few questions, no need to rebase unless there are actual issues though.

achilleas-k:
Thanks @ondrejbudai.

Pinned osbuild to the post-merge commit and changed the condition in mockbuild.sh to match your comments on the corresponding change in osbuild.

internal/distroregistry/distroregistry.go
achilleas-k: Yup, that's the plan.  I think we should do it in a separate PR though.  I wanted to keep this one short to see what people think of the way I wrote the image type definitions and pipelines.

internal/distro/rhel85/distro.go
achilleas-k: The kernel comes with the blueprint packages: https://github.com/osbuild/osbuild-composer/blob/cf73edd2d2248bbb5e561a8beeab694857be2132/internal/blueprint/blueprint.go#L107
including the fallback when unspecified:
https://github.com/osbuild/osbuild-composer/blob/cf73edd2d2248bbb5e561a8beeab694857be2132/internal/blueprint/customizations.go#L165-L167

teg: Ah, of course. I should know that ;)

internal/distroregistry/distroregistry.go
teg: Makes sense. I guess we also need to implement all the other image types for 8.5 first too.

internal/distro/rhel85/distro.go
achilleas-k: Yes.  Rebased to keep the history clean.

internal/distro/rhel85/distro.go
achilleas-k: Ditto

thozza (commented):
I like how you structured the code and split it into separate files. I think that this version is good for 8.5, however I think that we should take this approach further.

In order to make addition of new non-edge image types simpler, I would propose to somehow transform `ostreeTreePipeline()` function into a generic function, which can be used by any image type. My reasoning is that it creates stages for all potential Blueprint customizations and Image options. Therefore this will be the most commonly used function by all image types.

Please update `tools/test-case-generators/distro-arch-imagetype-map.json` with new distribution and proper image types.

Otherwise this looks very good to me. Thanks for all your effort!

internal/distro/rhel85/distro.go
```diff
@@ -0,0 +1,517 @@
...
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+)
+
+const name = "rhel-85"
+const osVersion = "8.5"
+const modulePlatformID = "platform:el8"
+const ostreeRef = "rhel/8/%s/edge"
+
+type distribution struct {
+	arches map[string]architecture
```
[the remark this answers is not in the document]
achilleas-k: Good point.  Changed and rebased.

achilleas-k:
> In order to make addition of new non-edge image types simpler, I would propose to somehow transform `ostreeTreePipeline()` function into a generic function, which can be used by any image type. My reasoning is that it creates stages for all potential Blueprint customizations and Image options. Therefore this will be the most commonly used function by all image types.

I've already started with this kind of renaming and reuse for a followup PR.  Most of the `ostreeTreePipeline()` function is going to become a set of common image stages (I'm going with the name `coreStages` for now), but some parts, the preptree stage in particular, are ostree specific.

I can include the changes in this PR, but the purpose of the separation wont become clear until I include the tar as well.  That said, I could add the tar-installer here as well, but it'll change the scope of the PR.

thozza:
> > In order to make addition of new non-edge image types simpler, I would propose to somehow transform `ostreeTreePipeline()` function into a generic function, which can be used by any image type. My reasoning is that it creates stages for all potential Blueprint customizations and Image options. Therefore this will be the most commonly used function by all image types.
> 
> I've already started with this kind of renaming and reuse for a followup PR. Most of the `ostreeTreePipeline()` function is going to become a set of common image stages (I'm going with the name `coreStages` for now), but some parts, the preptree stage in particular, are ostree specific.

Great, thus sounds exactly like what I was thinking about.

> 
> I can include the changes in this PR, but the purpose of the separation wont become clear until I include the tar as well. That said, I could add the tar-installer here as well, but it'll change the scope of the PR.

Follow-up PR is fine by me. I just think that we should do it rather sooner than later, mainly because we should start producing official 8.5 KVM guest images as well as eventually the AWS EC2 images for 8.5.

Thanks!

achilleas-k:
Removed the explicit enabling of greenboot services (rebased).

See [rhbz#1935177](https://bugzilla.redhat.com/show_bug.cgi?id=1935177)

achilleas-k:
> Removed the explicit enabling of greenboot and redboot services (rebased).
> 
> See [rhbz#1935177](https://bugzilla.redhat.com/show_bug.cgi?id=1935177)

ostree tests are failing because of this change.  The services should now be enabled by the package rhel-release, but it hasn't been updated to include the fix yet.

internal/distro/rhel85/distro.go
```diff
@@ -0,0 +1,168 @@
...
+	"errors"
+	"fmt"
+	"sort"
+
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/distro"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+)
+
+const name = "rhel-85"
```
ondrejbudai, line 16: I think we decided to change the name to `rhel-8.5`, right? Should we do it straight away?

test/data/repositories/rhel-85.json
```diff
@@ -0,0 +1,19 @@
+{
+  "x86_64": [
+    {
+      "name": "baseos",
+      "baseurl": "http://download.eng.bos.redhat.com/rhel-8/development/RHEL-8/latest-RHEL-8.5.0/compose/BaseOS/x86_64/os/",
```
ondrejbudai, line 5: Why don't we use rpmrepo?

ondrejbudai (commented):
Looks great, two comments inline.

test/data/repositories/rhel-85.json
thozza: Just a note, that we actually never used it for unreleased RHEL versions. We do the same for 8.4...

internal/distro/rhel85/distro.go
thozza: Just a note, that this will affect Weldr and other parts, which rely on the name to be without a dot. Something to consider, but I agree that it would make sense to start using the dot.

achilleas-k: I started the work with the new naming scheme on the branch and realised that the distro name detection would need to change, or we would need a condition special for 8.5.  Since the concept of aliases came up in that discussion (which is a good idea), I think we should do it in a separate PR where we add support for distro aliases and rename or alias existing distros.

test/data/repositories/rhel-85.json
achilleas-k: Like @thozza said.  I went with what we used for 8.4.

ondrejbudai: Ah, I guess it makes sense. Let's change it if it makes us some troubles.

internal/distro/rhel85/distro.go
ondrejbudai: Makes sense!

ondrejbudai (approved):
Awesome!

test/cases/ostree.sh
```diff
@@ -34,6 +34,11 @@ case "${ID}-${VERSION_ID}" in
         OSTREE_REF="rhel/8/${ARCH}/edge"
         OS_VARIANT="rhel8-unknown"
         BOOT_LOCATION="http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/os/";;
+    "rhel-8.5")
+        IMAGE_TYPE=edge-commit
+        OSTREE_REF="rhel/8/${ARCH}/edge"
+        OS_VARIANT="rhel8-unknown"
+        BOOT_LOCATION="http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/os/";;
```
ondrejbudai, line 41: This needs to `8.5`, schutzbot is failing on it. :( Can you please fix it?

ondrejbudai (changes requested):

test/cases/ostree.sh
achilleas-k: Thank you.  Fixed.

yih-redhat:
@achilleas-k  Do you know when will composer-cli use the new names 'edge-commit, edge-container, edge-installer'?

achilleas-k:
> @achilleas-k  Do you know when will composer-cli use the new names 'edge-commit, edge-container, edge-installer'?
composer-cli reads the image names from the API so there's no need to build in support.  If you run this branch on 8.5 and call `composer-cli compose types` it will list the supported image types from composer.

test/cases/ostree.sh
```diff
@@ -34,6 +34,11 @@ case "${ID}-${VERSION_ID}" in
         OSTREE_REF="rhel/8/${ARCH}/edge"
         OS_VARIANT="rhel8-unknown"
         BOOT_LOCATION="http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/os/";;
+    "rhel-8.5")
+        IMAGE_TYPE=edge-commit
+        OSTREE_REF="rhel/8/${ARCH}/edge"
+        OS_VARIANT="rhel8-unknown"
+        BOOT_LOCATION="http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8.5/compose/BaseOS/x86_64/os/";;
```
atodorov, line 41: `BOOT_LOCATION="$COPOSE_URL/compose/BaseOS/x86_64/os/"`

otherwise we're not properly testing against specific composes, e.g. rel-eng snapshots. See #1388

atodorov (changes requested):

henrywang:
@achilleas-k according to [errata](https://errata.devel.redhat.com/advisory/68370#c14) and [redhat-release-8.5-0.3.el8.x86_64.rpm](http://download-node-02.eng.bos.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.5.0/compose/BaseOS/x86_64/os/Packages/redhat-release-8.5-0.3.el8.x86_64.rpm), we can re-run test and land this PR if [`bz#1935177`](https://bugzilla.redhat.com/show_bug.cgi?id=1935177) is the only blocker.

test/cases/ostree.sh
```diff
@@ -34,6 +34,11 @@ case "${ID}-${VERSION_ID}" in
         OSTREE_REF="rhel/8/${ARCH}/edge"
         OS_VARIANT="rhel8-unknown"
         BOOT_LOCATION="http://download.devel.redhat.com/nightly/rhel-8/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/os/";;
```
atodorov, line 36: FYI https://github.com/osbuild/osbuild-composer/pull/1388#issuecomment-833330668

achilleas-k:
Currently working on splitting out the foundational parts of this PR.  I will open a separate PR that includes the 8.5 distro support without any of the image types.  This is to unblock other image types that have been waiting for this one to merge.

achilleas-k:
Rebased on main.

test/cases/ostree.sh
```diff
@@ -33,6 +33,12 @@ case "${ID}-${VERSION_ID}" in
         OS_VARIANT="rhel8-unknown"
         COMPOSE_URL="${COMPOSE_URL:-http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.4}"
         BOOT_LOCATION="$COMPOSE_URL/compose/BaseOS/x86_64/os/";;
+    "rhel-8.5")
+        IMAGE_TYPE=edge-commit
+        OSTREE_REF="rhel/8/${ARCH}/edge"
+        OS_VARIANT="rhel8-unknown"
+        COMPOSE_URL="${COMPOSE_URL:-http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.5}"
```
atodorov, line 40: This is in conflict/at odds with https://github.com/osbuild/osbuild-composer/issues/1397 and also partially with https://github.com/osbuild/osbuild-composer/pull/1274. Need to be careful not to end up with locations pointing both to 8.4 and 8.5 nightlies.

tools/provision.sh
```diff
@@ -51,12 +51,17 @@ sudo mkdir -p /etc/osbuild-composer/repositories
...
     "rhel-8.4")
         # Override old rhel-8.json and rhel-8-beta.json because RHEL 8.4 test needs nightly repos
         sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-84.json /etc/osbuild-composer/repositories/rhel-8.json
-        # If multiple tests are run and call provision.sh the symlink will need to be overriden with -f
+        # If multiple tests are run and call provision.sh the symlink will need to be overridden with -f
+        sudo ln -sf /etc/osbuild-composer/repositories/rhel-8.json /etc/osbuild-composer/repositories/rhel-8-beta.json;;
+    "rhel-8.5")
+        # Override old rhel-8.json and rhel-8-beta.json because RHEL 8.5 test needs nightly repos
+        sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-85.json /etc/osbuild-composer/repositories/rhel-8.json
+        # If multiple tests are run and call provision.sh the symlink will need to be overridden with -f
```
atodorov, line 64: Related to https://github.com/osbuild/osbuild-composer/issues/1318

test/data/repositories/rhel-85.json
```diff
@@ -0,0 +1,19 @@
...
+    }
+  ]
+}
```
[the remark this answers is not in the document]
achilleas-k: Thanks.  Fixed :)

test/cases/ostree.sh
atodorov: Please rebase again and drop line 40.

achilleas-k: Done

henrywang:
`error: Sanity-checking final rpmdb: Didn't find package 'greenboot-failing-unit-1.0-1.el8.noarch'` error in RHEL 8.5 `ostree-ng.sh` test is caused by [rpm-ostree issue](https://bugzilla.redhat.com/show_bug.cgi?id=1959874)

achilleas-k:
Rebased on `main` and cleaned up commits relating to the old Jenkins CI.  Need to define a RHEL 8.5 runner in https://github.com/osbuild/gitlab-ci-terraform first.

achilleas-k:
Added 8.5 to GitLab CI, but https://github.com/osbuild/gitlab-ci-terraform/pull/9 should be merged first

.gitlab-ci.yml
```diff
@@ -109,6 +109,7 @@ OSTree:
           # - openstack/fedora-34-x86_64
           - openstack/rhel-8-x86_64
           - openstack/rhel-8.4-x86_64
+          - openstack/rhel-8.5-x86_64
```
atodorov, line 112: @achilleas-k I propose to replace the 8.4 runner with 8.5 instead, not add another one. Also once the runner PR lands I prefer to actually change all of the 8.4 runners (we can do that separately).

achilleas-k: Changed ostree tests to run in 8.5 and the rest in 8.4 until we support all image types in 8.5.

schutzbot/mockbuild.sh.orig
```diff
@@ -0,0 +1,124 @@
...
+greenprint "📦 SHA: ${COMMIT}"
+greenprint "📤 RPMS will be uploaded to: ${REPO_URL}"
+
+<<<<<<< HEAD
+# rhel 8.4 will run off of the internal repos and does not have a redhat subscription
+if [[ $VERSION_ID == 8.4 || $VERSION_ID == 8.5 ]]; then
+=======
+# rhel 8.4 and 8.5 will run off of the internal repos and does not have a redhat subscription
+if [[ $VERSION_ID == 8.4 ]]; then
+>>>>>>> fe2b26a3 (schutzbot/mockbuild.sh: RHEL 8.5 support)
```
atodorov, line 77: Unresolved conflict.

achilleas-k: accidentally added a merge conflict `.orig` file during rebase

atodorov (approved):

teg (approved):

PR #1353, opened 2021-04-13
Schutzbot on GitLab CI PR

ondrejbudai:
> TL;DR: Running our own Jenkins instance is cumbersome, this PR explores an option of using GitLab CI instead.

This PR enables a parallel testing pipeline that is completely independent from our Jenkins pipeline. The goal of this effort is to explore the possibility of dumping Jenkins and replacing it with GitLab CI. Note that we will still host our code on GitHub, GitLab serves only as an external CI service (like Travis or Circle CI can be used).

After this PR is merged, all PRs will have another check (Schutzbot on GitLab). Schutzbot on GitLab currently runs only a subset of tests (base & ostree on RHEL 8 & CS8 & F32&33 on x86_64 & aarch64 on AWS & PSI OpenStack). The goal is to test this setup for approximately a week before deciding on which path to use. The first option is to continue using Jenkins, the second option is to dump it and use GitLab instead. During this period, Jenkins is still authoritative for deciding whether a PR should be merged. Maintainers are encouraged to make GitLab happy too but this is not required in this phase.

## How does it work

The following paragraphs explain how a new PR is processed using the new setup.

### Triggering GitLab

This PR adds a new workflow that's responsible for triggering the GitLab CI. In contrast with the existing GitHub workflow, it uses the new-ish [`pull_request_target` event][pull_request_target]. This event runs in the context of the base of the pull request, rather than in the merge commit as the `pull_request` event does. This means that modifying the workflow in a PR doesn't change the workflow until the PR is merged. This event also has the advantage that it has access to the secrets.

The new workflow does a very important thing in the first step: It checks if the event was created by someone who has *admin* rights. I was inspired by the [anaconda] repository where they do the same thing. If the check passes, the PR is pushed to a [GitLab repository][gitlab-ci-repo].

> This was previously done by Jenkins.

### GitLab self-hosted worker with a custom Terraform executor

After the PR is pushed into our [GitLab CI repository][gitlab-ci-repo], GitLab CI will start running all jobs as they are defined in `gitlab-ci.yml` (this PR adds this file).

The interesting part is that all RPM and test jobs are marked with a `terraform` tag. This causes GitLab to run them only on runners marked with such tag and not on any of the "public" workers. We run a self-hosted runner with the `terraform` tag on an EC2 instance.

Self-hosted workers can be configured to use several types of executors. There's a [docker executor][docker-exec] that runs jobs in a docker container, there's a [kubernetes executor][k8s-exec] that runs them in a Kubernetes cluster and many others. Our runner uses a super-cool GitLab feature called the [custom executor][custom-executor].

Our custom executor creates a new Terraform deployment for each job and runs the test scripts using ssh. It's hosted in the [osbuild/gitlab-ci-terraform-executor repository][terraform-exec]. Writing a custom executor is simple: It's just needed to write `prepare`, `exec` and `cleanup` scripts. Their job is to create the deployment, execute the scripts and tear down the deployment when everything is done. From my experience, these not too many lines of Bash work quite nicely.

The terraform templates themselves are also stored in a GitHub repository ([osbuild/gitlab-ci-terraform]). The terraform executor uses a pinned version of the template as specified in `schutzbot/terraform` file in the `osbuild-composer` repository.

> #### Security note
> 
> PRs on GitLab are never run on self-hosted runners. Therefore, if anyone opens a PR for the CI repository, the code won't be run on our workers.
> 
> Just in case, I also disabled pull requests in the CI repository. Both these features aren't sadly available on GitHub. GitHub also doesn't have support for custom executors so we need to use a proxy git repo on GitLab.

### Integration with GitHub

The first job in the GitLab pipeline creates a pending GitHub status using [this API][github-status-api] on the respective PR. Whenever a job fails, it sends a "failure status". The last job in the pipeline runs only on success and it marks the job as done. The status also contains a handy link to the GitLab pipeline.

## Gotchas

There are some gotchas in this setup.

### Canceling a job in GitLab is weird

Yeah, canceling a job doesn't run `after_script`, therefore we cannot send an update that the pipeline failed - the job will be shown as pending forever in GH UI. Also, a canceled job cannot run something like the cloud cleaner.

The forever pending status is annoying but I think we can live with it, the cloud cleaner stuff is potentially a problem - we must be careful when canceling a job. One should be aware of what artifacts might the job leave.

This issue is [known and tracked][canceling-job].

### When our self-hosted runner is restarted, we lose all running jobs

However, this shouldn't be a huge issue, EC2 instances are not restarted often. When this happens, we just need to restart all affected jobs.

We might need to take care of all Terraform resources that the runner had allocated: This is simple with the instances - just call `terraform destroy` and they are gone. We might have an issue with the cloud cleaner though.

### Storage

I'm not sure from the docs how much storage we have for logs on GitLab. We need to be careful here.

## Benefits

Finally. I think we have many more benefits than there are gotchas. :)

### Public logs

Yeah, they are just visible on gitlab.com publicly, isn't that nice?

### Ability to restart jobs

Jenkins can restart only the whole pipeline, GitLab has the power to restart individual stages.

### Terraform config

OMG, these terraform configs are so simple. Much better than configuring instance types in Jenkins. We can also use Azure if we want nested virtualization. With Terraform, this should be dead simple.

### No Jenkins

No 10s of plugins, no out of space issues, simply no Jenkins.

### Rev-dep tests

We might be able to hook up very simple reverse dependency tests with this setup. Stay tuned.

### Nightly tests

The same as above, this should be much simpler than with Jenkins.

### Beaker executor

Yeah, we can definitely create one and run our tests for alternative architectures on Beaker. Alternatively, we might be able to use QEMU emulation on an EC2 machine. Both approaches have their downsides.

[pull_request_target]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
[anaconda]: https://github.com/rhinstaller/anaconda/blob/master/.github/workflows/tests-owners.yml
[gitlab-ci-repo]: https://gitlab.com/osbuild/ci/osbuild-composer
[docker-exec]: https://docs.gitlab.com/runner/executors/docker.html
[k8s-exec]: https://docs.gitlab.com/runner/executors/kubernetes.html
[custom-executor]: https://docs.gitlab.com/runner/executors/custom.html
[terraform-exec]: https://github.com/osbuild/gitlab-ci-terraform-executor
[osbuild/gitlab-ci-terraform]: https://github.com/osbuild/gitlab-ci-terraform
[github-status-api]: https://docs.github.com/en/rest/reference/repos#statuses
[canceling-job]: https://gitlab.com/gitlab-org/gitlab/-/issues/15603


teg (approved):
Nice! Can't wait to switch over!

PR #1355, opened 2021-04-14
mockbuild: use s3cmd put instead of sync to speed up the RPM upload

ondrejbudai:
s3cmd sync actually downloads metadata for all objects in a s3 bucket.
We have built a lot of RPMs, thus this takes 5 minutes on AWS and 25 minutes
on my laptop (!!!).

Let's use recursive put instead. This doesn't delete any files on the remote
side. As we upload RPMs only once, this also shouldn't fail on "the
object already exists". Using this method, we should be able to upload the
RPMs in seconds.

This pull request includes:


larskarlitski (approved):
Yep! 🐎

PR #1357, opened 2021-04-14
test: Install and run ostree commit on UEFI VM

henrywang:
Update `ostree-ng.sh` to install and run ostree commit on UEFI VM.  `ostree.sh` keeps on BIOS VM. Both of BIOS and UEFI are covered.
`check_ostree.yaml` ansible playbook has to be updated to support both BISO and UEFI.

### DIFF
--- a/test/cases/ostree-ng.sh
+++ b/test/cases/ostree-ng.sh
@@ -168,11 +168,7 @@ wait_for_ssh_up () {
 clean_up () {
     greenprint "🧼 Cleaning up"
     sudo virsh destroy "${IMAGE_KEY}"
-    if [[ $ARCH == aarch64 ]]; then
-        sudo virsh undefine "${IMAGE_KEY}" --nvram
-    else
-        sudo virsh undefine "${IMAGE_KEY}"
-    fi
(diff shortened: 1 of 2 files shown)


henrywang:
This PR only changed file `ostree-ng.sh` and `check_ostree.yaml`. The failed tests `CS8 Integration` and `EL8.4 Integration` are not related with this PR.

teg (approved):
I agree UEFI is the most important to test, and we have now some coverage of both BIOS and UEFI, which is great.

I wonder, should we also test our installer on BIOS though? Now only the commits are tested on BIOS with the standard installer, but not the one we produce? That would anyway be a separate PR.

henrywang:
> I agree UEFI is the most important to test, and we have now some coverage of both BIOS and UEFI, which is great.
> 
> I wonder, should we also test our installer on BIOS though? Now only the commits are tested on BIOS with the standard installer, but not the one we produce? That would anyway be a separate PR.

Thanks @teg! Great idea. Let's install both BIOS and UEFI VMs on the same `ostree-ng.sh`

PR #1358, opened 2021-04-15
Enable testing of images in Azure using terraform

jrusz:
This pull request includes:

### DIFF
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -1,0 +1,237 @@
+#!/bin/bash
+set -euo pipefail
+
+source /etc/os-release
+
+# Colorful output.
+function greenprint {
+    echo -e "\033[1;32m${1}\033[0m"
+}
+
+#TODO: Remove this once there is rhel9 support for Azure image type
+if [[ $DISTRO_CODE == rhel_90 ]]; then
+    greenprint "Skipped"
+    exit 0
+fi
+
+# Provision the software under tet.
+/usr/libexec/osbuild-composer-test/provision.sh
+
+# Terraform needs azure-cli to talk to Azure.
+if ! hash az; then
+    # this installation method is taken from the official docs:
+    # https://docs.microsoft.com/cs-cz/cli/azure/install-azure-cli-linux?pivots=dnf
+    sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
+    echo -e "[azure-cli]
+name=Azure CLI
+baseurl=https://packages.microsoft.com/yumrepos/azure-cli
+enabled=1
+gpgcheck=1
+gpgkey=https://packages.microsoft.com/keys/microsoft.asc" | sudo tee /etc/yum.repos.d/azure-cli.repo
+
+  greenprint "Installing azure-cli"
+  sudo dnf install -y azure-cli
+  az version
+fi
+
+# We need terraform to provision the vm in azure and then destroy it
+if [ "$ID" == "rhel" ] || [ "$ID" == "centos" ]
+then
+    release="RHEL"
+elif [ "$ID" == "fedora" ]
+then
+    release="fedora"
+else
+    echo "Test is not running on neither Fedora, RHEL or CentOS, terminating!"
+    exit 1
+fi
+sudo dnf config-manager --add-repo https://rpm.releases.hashicorp.com/$release/hashicorp.repo
+sudo dnf install -y terraform
+
+ARCH=$(uname -m)
+TEST_ID="$DISTRO_CODE-$ARCH-$BRANCH_NAME-$BUILD_ID"
+IMAGE_KEY=image-${TEST_ID}
+
+# Jenkins sets WORKSPACE to the job workspace, but if this script runs
+# outside of Jenkins, we can set up a temporary directory instead.
+if [[ ${WORKSPACE:-empty} == empty ]]; then
+    WORKSPACE=$(mktemp -d)
+fi
+
+# Set up temporary files.
+TEMPDIR=$(mktemp -d)
+AZURE_CONFIG=${TEMPDIR}/azure.toml
+BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
+COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
+COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
+
+# Check for the smoke test file on the Azure instance that we start.
+smoke_test_check () {
+    SMOKE_TEST=$(sudo ssh -i key.rsa redhat@"${1}" -o StrictHostKeyChecking=no 'cat /etc/smoke-test.txt')
+    if [[ $SMOKE_TEST == smoke-test ]]; then
+        echo 1
+    else
+        echo 0
+    fi
+}
+
+# Get the compose log.
+get_compose_log () {
+    COMPOSE_ID=$1
+    LOG_FILE=${WORKSPACE}/osbuild-${ID}-${VERSION_ID}-azure.log
+
+    # Download the logs.
+    sudo composer-cli compose log "$COMPOSE_ID" | tee "$LOG_FILE" > /dev/null
+}
+
+# Get the compose metadata.
+get_compose_metadata () {
+    COMPOSE_ID=$1
+    METADATA_FILE=${WORKSPACE}/osbuild-${ID}-${VERSION_ID}-azure.json
+
+    # Download the metadata.
+    sudo composer-cli compose metadata "$COMPOSE_ID" > /dev/null
+
+    # Find the tarball and extract it.
+    TARBALL=$(basename "$(find . -maxdepth 1 -type f -name "*-metadata.tar")")
+    tar -xf "$TARBALL"
+    rm -f "$TARBALL"
+
+    # Move the JSON file into place.
+    cat "${COMPOSE_ID}".json | jq -M '.' | tee "$METADATA_FILE" > /dev/null
+}
+
+# Export Azure credentials
+exec 4<"$AZURE_CREDS"
+readarray -t -u 4 vars
+for line in "${vars[@]}"; do export "${line?}"; done
+exec 4<&-
+
+# Write an Azure TOML file
+tee "$AZURE_CONFIG" > /dev/null << EOF
+provider = "azure"
+
+[settings]
+storageAccount = "${AZURE_STORAGE_ACCOUNT}"
+storageAccessKey = "${AZURE_STORAGE_ACCESS_KEY}"
+container = "${AZURE_CONTAINER_NAME}"
+EOF
+
+# Write a basic blueprint for our image.
+tee "$BLUEPRINT_FILE" > /dev/null << EOF
+name = "bash"
+description = "A base system with bash"
+version = "0.0.1"
+
+[[packages]]
+name = "bash"
+
+[[packages]]
+name = "cloud-init"
+
+[customizations.services]
+enabled = ["sshd", "cloud-init", "cloud-init-local", "cloud-config", "cloud-final"]
+EOF
+
+# Prepare the blueprint for the compose.
+greenprint "📋 Preparing blueprint"
+sudo composer-cli blueprints push "$BLUEPRINT_FILE"
+sudo composer-cli blueprints depsolve bash
+
+# Get worker unit file so we can watch the journal.
+WORKER_UNIT=$(sudo systemctl list-units | grep -o -E "osbuild.*worker.*\.service")
+sudo journalctl -af -n 1 -u "${WORKER_UNIT}" &
+WORKER_JOURNAL_PID=$!
+
+# Start the compose and upload to Azure.
+greenprint "🚀 Starting compose"
+sudo composer-cli --json compose start bash vhd "$IMAGE_KEY" "$AZURE_CONFIG" | tee "$COMPOSE_START"
+COMPOSE_ID=$(jq -r '.build_id' "$COMPOSE_START")
+
+# Wait for the compose to finish.
+greenprint "⏱ Waiting for compose to finish: ${COMPOSE_ID}"
+while true; do
+    sudo composer-cli --json compose info "${COMPOSE_ID}" | tee "$COMPOSE_INFO" > /dev/null
+    COMPOSE_STATUS=$(jq -r '.queue_status' "$COMPOSE_INFO")
+
+    # Is the compose finished?
+    if [[ $COMPOSE_STATUS != RUNNING ]] && [[ $COMPOSE_STATUS != WAITING ]]; then
+        break
+    fi
+
+    # Wait 30 seconds and try again.
+    sleep 30
+done
+
+# Capture the compose logs from osbuild.
+greenprint "💬 Getting compose log and metadata"
+get_compose_log "$COMPOSE_ID"
+get_compose_metadata "$COMPOSE_ID"
+
+# Did the compose finish with success?
+if [[ $COMPOSE_STATUS != FINISHED ]]; then
+    echo "Something went wrong with the compose. 😢"
+    exit 1
+fi
+
+# Stop watching the worker journal.
+sudo pkill -P ${WORKER_JOURNAL_PID}
+
+# Set up necessary variables for terraform
+export TF_VAR_RESOURCE_GROUP="$AZURE_RESOURCE_GROUP"
+export TF_VAR_STORAGE_ACCOUNT="$AZURE_STORAGE_ACCOUNT"
+export TF_VAR_CONTAINER_NAME="$AZURE_CONTAINER_NAME"
+export TF_VAR_BLOB_NAME="$IMAGE_KEY".vhd
+export TF_VAR_TEST_ID="$TEST_ID"
+export BLOB_URL="https://$AZURE_STORAGE_ACCOUNT.blob.core.windows.net/$AZURE_CONTAINER_NAME/$IMAGE_KEY.vhd"
+export ARM_CLIENT_ID="$AZURE_CLIENT_ID" > /dev/null
+export ARM_CLIENT_SECRET="$AZURE_CLIENT_SECRET" > /dev/null
+export ARM_SUBSCRIPTION_ID="$AZURE_SUBSCRIPTION_ID" > /dev/null
+export ARM_TENANT_ID="$AZURE_TENANT_ID" > /dev/null
+
+# Copy terraform main file and cloud-init to current working directory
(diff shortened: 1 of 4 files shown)


jrusz:
Hi @atodorov, I believe it's finally ready for review, could you please take a look at this and share your thought about this?

I took `aws.sh` as a template and made it work with azure with some changes and this is basically how it works:

1. Osbuild-composer is deployed and all the dependencies are installed first (like `azure-cli` and `terraform`)
1. Then a basic blueprint and azure config is created which is needed for `composer-cli`
1. Compose is started which uploads the created image as a page blob to the specified azure storage account and container
1. Terraform is initialized then which downloads it's own dependencies and prepares to run
1. The previously uploaded page blob is imported into terraform which makes terraform aware of it's presence and can work with it
1. `terraform plan` reads the `main.tf` file and runs whatever is specified there
1. The `main.tf` file is a modified example from the docs with a few additions, I believe it contains the minimum that is needed to spin up a linux vm there
2. After the vm is up it checks for the `smoke-test.txt` file which is created by `cloud-init`
3. `terraform destroy` then deletes everything including the page blob uploaded by `composer-cli`

ondrejbudai:
Please remove the second commit. I'm marking this as a draft so we don't merge it.

jrusz:
> Please remove the second commit. I'm marking this as a draft so we don't merge it.

I wanted to do that once I get a review on the test and apply any additional changes, but you're right, it could have been merged like this by accident. Thanks for pointing it out :)

test/cases/azure.sh
```diff
@@ -0,0 +1,240 @@
...
+if [ "$ID" == "rhel" ] 
+then
+    release="RHEL"
+elif [ "$ID" == "fedora" ]
+then
+    release="fedora"
+else
+    echo "Test is not running on neither Fedora nor RHEL, terminating!"
+    exit 1
+fi
```
atodorov, line 55: How about CentOS ? We offer CentOS Stream 8 as an option in the Image Builder service and that can be paired with an Azure upload target. We should be testing it as well.

test/cases/azure.sh
```diff
@@ -0,0 +1,240 @@
...
+# Set up temporary files.
+TEMPDIR=$(mktemp -d)
+AZURE_CONFIG=${TEMPDIR}/azure.toml
+BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
+COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
+COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
+
+# Check for the smoke test file on the Azure instance that we start.
+smoke_test_check () {
+    SMOKE_TEST=$(sudo ssh -i key.rsa redhat@"${1}" -o StrictHostKeyChecking=no 'cat /etc/smoke-test.txt')
```
atodorov, line 78: Is `redhat` always the default username for ssh ? IIRC that was coming from cloud-init and in some older versions it was different between RHEL/Fedora (possibly CentOS). In the old lorax-composer test suite we're using `azure-user` instead but IIRC we weren't using cloud-init back then.

atodorov: Also in main.tf , line 163 you have `azureuser`. What's the relation with that ?

atodorov (changes requested):
Overall looks good, added a few comments. 

However you've got to drop the second commit which shortens Jenkinsfile and make sure to add azure.sh to all of our testing targets (OS/arch variants), including CentOS if possible so we can see how it works.

test/cases/azure.sh
jrusz: Hm, I found only `fedora` and `RHEL` here but it says `CentOS` is supported, although I'm not sure what to put in `$release`... I'll get a CentOS machine and try it.

ondrejbudai: `RHEL`. CentOS Linux is just a rebuild. With CentOS Stream it gets a bit more complicated but the compatibility should be still there (if not, it might be a bug).

test/cases/azure.sh
jrusz: `redhat` username comes from the cloud-init. It was `azureuser` by default but after I included the cloud-init file it changed to `redhat`... About the `azureuser` specified in `main.tf` right now I'm not even sure if that user exists there as I was unable to log in using that user... I can do some more investigation or just change the `azureuser` in `main.tf` to `redhat` and see if that works so that it's consistent.

test/cases/azure.sh
```diff
@@ -0,0 +1,240 @@
...
+
+source /etc/os-release
+
+# Colorful output.
+function greenprint {
+    echo -e "\033[1;32m${1}\033[0m"
+}
+
+#TODO: Remove this once there is rhel9 support for Azure image type
+if [[ $DISTRO_CODE == rhel_90 ]]; then
```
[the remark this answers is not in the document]
jmikovic: looking at the distro implementation, there is still only qcow2 image type for 9.0

test/cases/azure.sh
atodorov: @ondrejbudai can you comment on that? I think we're fine as-is without further changes.

ondrejbudai: Hmm, we are initializating the user twice - once in the terraform file and once by injecting a cloud-init. I think we should choose one. Also, the test seems to be failing on Fedora.

test/cases/azure.sh
```diff
@@ -0,0 +1,232 @@
...
+    greenprint "Skipped"
+    exit 0
+fi
+
+# Provision the software under tet.
+/usr/libexec/osbuild-composer-test/provision.sh
+
+# We need azure-cli to talk to Azure.
```
ondrejbudai, line 20: Do we? I don't see it being used anywhere.

test/cases/azure.sh
```diff
@@ -0,0 +1,232 @@
...
+then
+    release="RHEL"
+elif [ "$ID" == "fedora" ]
+then
+    release="fedora"
+else
+    echo "Test is not running on neither Fedora, RHEL or CentOS, terminating!"
+    exit 1
+fi
+sudo dnf install -y dnf-plugins-core
```
ondrejbudai, line 48: I suggest making `dnf-plugins-core` a depedency of `osbuild-composer-tests` in the specfile.

test/cases/azure.sh
```diff
@@ -0,0 +1,232 @@
...
+get_compose_metadata "$COMPOSE_ID"
+
+# Did the compose finish with success?
+if [[ $COMPOSE_STATUS != FINISHED ]]; then
+    echo "Something went wrong with the compose. 😢"
+    exit 1
+fi
+
+# Stop watching the worker journal.
+sudo kill ${WORKER_JOURNAL_PID}
```
ondrejbudai, line 175: This doesn't work as we expected, see https://github.com/osbuild/osbuild-composer/commit/e7b3063f7ec06fdb547515c24fd956ed1af74e0b

test/data/azure/main.tf
```diff
@@ -0,0 +1,171 @@
...
+    sensitive = true
+}
+
+# Create virtual machine
+resource "azurerm_linux_virtual_machine" "testterraformvm" {
+    name                  = "testVM"
+    location              = data.azurerm_resource_group.testResourceGroup.location
+    resource_group_name   = data.azurerm_resource_group.testResourceGroup.name
+    network_interface_ids = [azurerm_network_interface.testterraformnic.id]
+    size                  = "Standard_DS1_v2"
```
ondrejbudai, line 148: Can we try using something cheaper? B1S might work for us.

test/data/azure/main.tf
```diff
@@ -0,0 +1,171 @@
...
+  algorithm = "RSA"
+  rsa_bits = 4096
+}
+output "tls_private_key" {
+    value = tls_private_key.test_ssh.private_key_pem
+    sensitive = true
+}
+
+# Create virtual machine
+resource "azurerm_linux_virtual_machine" "testterraformvm" {
```
ondrejbudai, line 143: Can we also try `priority = "Spot"` to get a spot instance and save some more money?

test/data/azure/main.tf
atodorov: lorax-composer used Standard_B2s, while `test/data/azure/deployment-template.json` uses Standard_B1s

test/cases/azure.sh
jrusz: Terraform needs it. `Error building AzureRM Client: Please ensure you have installed Azure CLI version 2.0.79 or newer. Error parsing json result from the Azure CLI: Error launching Azure CLI: exec: "az": executable file not found in $PATH.`

atodorov: Does it need to be installed explicitly or only the repos configured so that terraform can pull it as a dependency? In any case a comment saying TF needs this is a good idea.

test/cases/azure.sh
jrusz: Uhm, terraform requires to specify either `admin_ssh_key` or `admin_password` and I guess we want to use cloud-init for the smoke test. I could try creating another cloud-init file but I see very little reason to do so... Also the failure on Fedora is resolved, it was due to missing cloud-init package in the blueprint.

test/cases/azure.sh
jrusz: It needs to be installed, it won't look for it unless it tries to process the `main.tf` file and then it just gives error if it can't find it. I updated the comment.

test/cases/azure.sh
jrusz: Done.

test/data/azure/main.tf
jrusz: Done, I hope there won't be any problems with it, but we'll see in time.

test/data/azure/main.tf
jrusz: There are no `B-series` machines available as `Spot` instances. The cheapest ones are `D-series` and since it's up a minute or two tops, I guess we'll be fine.

jrusz:
I'm making a few more tests before it's ready for another round of review. I realized that the current `cloud-cleaner` won't clean up the resources if they're left there so I need to address that and also I had to change the name of the resources so that they're unique for each system it's running on since it's running in parallel and there were conflicts.

jrusz:
Hi @atodorov and @ondrejbudai , I believe it's ready for another round of review. I addressed your comments and also made sure `cloud-cleaner` cleans up the created resources if the script fails and `terraform destroy` is not ran. There are some failures from Schutzbot but I don't think they're related, one seems to be lost connection to slave and I'm not sure what happened with the other.

test/data/azure/main.tf
atodorov: Looking at the Azure calculator if we take the B1S as the base price then B2S is 4x more expensive while DS1-v2 is 7x more expensive. 

Azure says that is supports "per minute" billing unlike AWS which bills you per "started hour" which is nice.  OTOH my experience with spot instances in AWS is that while most of the time they are indeed very cheap during rush periods (e.g. major holidays) they become more expensive even than their regular price.

I don't have metrics about the hours we consume but it looks to me that a `Standard_B2s` instance is a good compromise.

@ondrejbudai what do you think ?

test/data/azure/main.tf
atodorov: FTR does `priority = "Spot"` work for B series instances or does it cause a failure ? If these are connected and cause failures when one of the values doesn't match the other we at least need a comment around then, or visual separation with new lines + comments etc.

atodorov:
The test reports PASS but I see something which doesn't look well:

```ERROR:

[2021-04-30T18:48:51.367Z]                                                                                          -> github.com/Azure/azure-pipeline-go/pipeline.NewError, /builddir/build/BUILD/osbuild-composer-d50ef703d50e225d68c88a44eef81527feb9f943/_build/src/github.com/osbuild/osbuild-composer/vendor/github.com/Azure/azure-pipeline-go/pipeline/error.go:157

[2021-04-30T18:48:51.367Z]                                                                                          HTTP request failed

[2021-04-30T18:48:51.367Z]                                                                                          

[2021-04-30T18:48:51.367Z]                                                                                          Put "https://travisciuploads.blob.core.windows.net/travisciuploads/image-rhel_84-x86_64-PR-1358-44.vhd?comp=page&timeout=61": read tcp 10.0.147.97:36458->20.150.42.4:443: read: connection reset by peer
```

This is at the end of the upload to Azure. Not sure if a real error but maybe we need an error check after commands in the script.

Then I see: 

```
 🛃 Checking for smoke test file

[2021-04-30T18:54:56.536Z] ssh: connect to host 20.86.107.65 port 22: Connection refused

[2021-04-30T18:55:01.802Z] ssh: connect to host 20.86.107.65 port 22: Connection refused

[2021-04-30T18:55:07.069Z] Warning: Permanently added '20.86.107.65' (ECDSA) to the list of known hosts.

[2021-04-30T18:55:08.443Z] Smoke test passed! 🥳
```

this looks like a retry sequence and the wording can be updated to make it more clear that we're retrying to connect so we don't have to wonder if this is an error or not.

test/data/azure/main.tf
jrusz: I don't have the exact error it gives but during my manual testing it just gave an error with something like "Selected instance can not be found/provisioned".. The failure is shown during the time terraform tries to provision the resource, terraform doesn't know this before and believes this combination is fine and proceeds. If you try to manually provision a B-series virtual machine in Azure portal it will tell you "This size does not support Azure Spot"...

test/data/azure/main.tf
jrusz: I looked at this again and found out there's `A-series` and that seems to be actually cheaper when used as a `Spot` instance. Also the eviction rate seems a lot lower and they suggest this series to be used for testing, so that might be another option to consider. I don't know what's the difference if we run a Spot instance for 2 minutes or regular instance for 2 minutes though...
azure_prices

atodorov: @yuxisun1217 do you have any experience insights here? We're looking for the cheapest possible instance to use for testing and preferably use spot instances to go even cheaper. Do you have any practical experience on this subject ?

ondrejbudai: we agreed on our weekly meeting that we will use a B1S non-spot instance for now and we can iterate from there.

jrusz: `B1s` seems to be working fine, all stages passed.

jrusz:
> The test reports PASS but I see something which doesn't look well:
> 
> ```
> 
> [2021-04-30T18:48:51.367Z]                                                                                          -> github.com/Azure/azure-pipeline-go/pipeline.NewError, /builddir/build/BUILD/osbuild-composer-d50ef703d50e225d68c88a44eef81527feb9f943/_build/src/github.com/osbuild/osbuild-composer/vendor/github.com/Azure/azure-pipeline-go/pipeline/error.go:157
> 
> [2021-04-30T18:48:51.367Z]                                                                                          HTTP request failed
> 
> [2021-04-30T18:48:51.367Z]                                                                                          
> 
> [2021-04-30T18:48:51.367Z]                                                                                          Put "https://travisciuploads.blob.core.windows.net/travisciuploads/image-rhel_84-x86_64-PR-1358-44.vhd?comp=page&timeout=61": read tcp 10.0.147.97:36458->20.150.42.4:443: read: connection reset by peer
> ```
> 
> This is at the end of the upload to Azure. Not sure if a real error but maybe we need an error check after commands in the script.

I was not able to reproduce it manually and can't see it when ran again, not sure what to do here...
> Then I see:
> 
> ```
>  🛃 Checking for smoke test file
> 
> [2021-04-30T18:54:56.536Z] ssh: connect to host 20.86.107.65 port 22: Connection refused
> 
> [2021-04-30T18:55:01.802Z] ssh: connect to host 20.86.107.65 port 22: Connection refused
> 
> [2021-04-30T18:55:07.069Z] Warning: Permanently added '20.86.107.65' (ECDSA) to the list of known hosts.
> 
> [2021-04-30T18:55:08.443Z] Smoke test passed! 🥳
> ```
> 
> this looks like a retry sequence and the wording can be updated to make it more clear that we're retrying to connect so we don't have to wonder if this is an error or not.

I added a message about retrying connection if that happens again, however it didn't appear now probably because we don't want spot instances anymore and those `B1s` spin up faster, don't know...

Anyway this should hopefully be everything for this PR.

ondrejbudai (approved):
Looks good, is this ready to be merged? RHEL 8.4 ostree test failure is expected and unrelated.

jrusz:
> Looks good, is this ready to be merged? RHEL 8.4 ostree test failure is expected and unrelated.

Thanks, I believe it is, but I'd like to hear from @atodorov if he agrees.

atodorov (approved):
LGTM

atodorov:
> > The test reports PASS but I see something which doesn't look well:
> > ```
> > 
> > [2021-04-30T18:48:51.367Z]                                                                                          -> github.com/Azure/azure-pipeline-go/pipeline.NewError, /builddir/build/BUILD/osbuild-composer-d50ef703d50e225d68c88a44eef81527feb9f943/_build/src/github.com/osbuild/osbuild-composer/vendor/github.com/Azure/azure-pipeline-go/pipeline/error.go:157
> > 
> > [2021-04-30T18:48:51.367Z]                                                                                          HTTP request failed
> > 
> > [2021-04-30T18:48:51.367Z]                                                                                          
> > 
> > [2021-04-30T18:48:51.367Z]                                                                                          Put "https://travisciuploads.blob.core.windows.net/travisciuploads/image-rhel_84-x86_64-PR-1358-44.vhd?comp=page&timeout=61": read tcp 10.0.147.97:36458->20.150.42.4:443: read: connection reset by peer
> > ```
> > 
> > 
> > This is at the end of the upload to Azure. Not sure if a real error but maybe we need an error check after commands in the script.
> 
> I was not able to reproduce it manually and can't see it when ran again, not sure what to do here...
> 

Indeed it is not present. Could have been an unhandled TCP close in one of the upstream libraries. We'll investigate if it happens again.

atodorov:
```
PUT https://travisciuploads.blob.core.windows.net/travisciuploads/image-rhel_84-x86_64-PR-1358-45.vhd?comp=page&timeout=61

azurerm_linux_virtual_machine.testterraformvm: Creation complete after 1m53s [id=/subscriptions/8d026bb1-2a65-454d-a88f-c896db94c4f8/resourceGroups/travis-ci-osbuild-composer/providers/Microsoft.Compute/virtualMachines/vm-rhel_84-x86_64-PR-1358-45]
```

I just want to make sure our config/storage bucket is indeed configured with the not so appropriate name of "travisciuploads" !

@ondrejbudai ^^^

ondrejbudai:
Yes, this is probably a leftover from the past.

PR #1373, opened 2021-04-23
GCP: Specify and randomize GCE region used for image import

thozza:
The GCP image import method currently use the Cloud Build API with Google's Daisy workflow. This workflow creates multiple GCE resources during its execution. Although the desired Region for the imported image is specified as a workflow argument, this has no effect on the GCE Zone used by the workflow for created resources. By default it seems to default to "us-central1-a" Zone. As a result, there are common cases of resources being exhausted in the default zone.

Add a method, which translates provided Google Storage Region to a GCE Region, which is needed mainly for multi and dual Storage Regions. Note that the status of the Region is not checked by the function as this would require a logic to handle situations when the API caller specified a Region, which is DOWN in the first place.

Add a method, which returns a list of available GCE Zones for a given GCE Region.

Modify the ComputeImageImport() method to translate the provided Google Storage Region to list of corresponding GCE Regions. If the provided Storage Region is not multi or dual Region, then the list contains only a single item, the provided Region. Then pick a random Region from the list. Subsequently get available GCE Zones within the Region and pick a random one for use by the workflow. Specify the GCE Zone to use as a build step argument.

This change should be completely transparent to the API user.

The change should help with mitigation of transient failures of image imports to GCP observed this week in our CI.

This pull request includes:

    - I don't think that this needs documenting in NEWS, but if others think otherwise, I'll happily add it. 

### DIFF
--- a/internal/cloud/gcp/compute.go
+++ b/internal/cloud/gcp/compute.go
@@ -2,7 +2,10 @@ package gcp
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
+	"math/big"
+	"strings"
 	"time"
 
 	cloudbuild "cloud.google.com/go/cloudbuild/apiv1"
@@ -60,8 +63,43 @@ func (g *GCP) ComputeImageImport(ctx context.Context, bucket, object, imageName,
 		"-timeout=7000s",
 		"-client_id=api",
 	}
+
 	if region != "" {
 		buildStepArgs = append(buildStepArgs, fmt.Sprintf("-storage_location=%s", region))
+
+		// Set the region to be used by the daisy workflow when creating resources
+		// If not specified, the workflow seems to always default to us-central1.
+
+		// The Region passed as the argument is a Google Storage Region, which can be a multi or dual region.
+		// Multi and Dual regions don't work with GCE API, therefore we need to get the list of GCE regions
+		// that they map to. If the passed Region is not a multi or dual Region, then the returned slice contains
+		// only the Region passed as an argument.
+		gceRegions, err := g.storageRegionToComputeRegions(ctx, region)
+		if err != nil {
+			return nil, fmt.Errorf("failed to translate Google Storage Region to GCE Region: %v", err)
+		}
+		// Pick a random GCE Region to be used by the image import workflow
+		gceRegionIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(gceRegions))))
+		if err != nil {
+			return nil, fmt.Errorf("failed to pick random GCE Region: %v", err)
+		}
+		// The expecation is that Google won't have more regions listed for multi/dual
+		// regions than what can potentially fit into int32.
+		gceRegion := gceRegions[int(gceRegionIndex.Int64())]
+
+		availableZones, err := g.computeZonesInRegion(ctx, gceRegion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get available GCE Zones within Region '%s': %v", region, err)
+		}
+		// Pick random zone from the list
+		gceZoneIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(availableZones))))
+		if err != nil {
+			return nil, fmt.Errorf("failed to pick random GCE Zone: %v", err)
+		}
+		// The expecation is that Google won't have more zones in a region than what can potentially fit into int32
+		zone := availableZones[int(gceZoneIndex.Int64())]
+
+		buildStepArgs = append(buildStepArgs, fmt.Sprintf("-zone=%s", zone))
 	}
 	if os != "" {
 		buildStepArgs = append(buildStepArgs, fmt.Sprintf("-os=%s", os))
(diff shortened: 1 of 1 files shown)


internal/cloud/gcp/compute.go
```diff
@@ -288,3 +315,108 @@ func (g *GCP) ComputeDiskDelete(ctx context.Context, zone, disk string) error {
...
+// since these can not be used with GCE API calls.
+//
+// Uses:
+//  - Compute Engine API
+func (g *GCP) storageRegionToComputeRegions(ctx context.Context, region string) ([]string, error) {
+	regionLower := strings.ToLower(region)
+
+	// Handle Dual-Regions
+	// https://cloud.google.com/storage/docs/locations#location-dr
+	if regionLower == "asia1" {
```
thozza, line 336: Personally I have doubts that it is a good idea to hardcode the list of zones in the code. Mainly because it will be impossible to change this e.g. when the osbuild-composer version containing the change will end up in a RHEL EUS release and Google decides to change the definition of any of the Dual-Regions. I would be curious what others think. I would be more happy with making this dynamic. Maybe return the same list of Regions for multi-region as for the Dual regions, as this is mostly about where are Google's daisy workflow creates temporary resources for the image import. So one could argue that as long it is not on the other continent, it should not be that big of an issue, but I don't have any data to support this.

internal/cloud/gcp/compute.go
```diff
@@ -60,8 +62,33 @@ func (g *GCP) ComputeImageImport(ctx context.Context, bucket, object, imageName,
 		"-timeout=7000s",
 		"-client_id=api",
 	}
+
 	if region != "" {
 		buildStepArgs = append(buildStepArgs, fmt.Sprintf("-storage_location=%s", region))
+
+		// Set the region to be used by the daisy workflow when creating resources
+		// If not specified, the workflow seems to always default to
```
ondrejbudai, line 70: I think this sentence misses its end.

internal/cloud/gcp/compute.go
```diff
@@ -60,8 +62,33 @@ func (g *GCP) ComputeImageImport(ctx context.Context, bucket, object, imageName,
...
+		// Multi and Dual regions don't work with GCE API, therefore we need to get the list of GCE regions
+		// that they map to. If the passed Region is not a multi or dual Region, then the returned slice contains
+		// only the Region passed as an argument.
+		gceRegions, err := g.storageRegionToComputeRegions(ctx, region)
+		if err != nil {
+			return nil, fmt.Errorf("failed to translate Google Storage Region to GCE Region: %v", err)
+		}
+
+		// Pick a random GCE Region to be used by the image import workflow
+		rand.Seed(time.Now().UnixNano())
```
ondrejbudai, line 82: I always prefer `crypto/rand` as a seed because it doesn't have a global state which is always weird. On the other hand, it's overkill for choosing a region. What do you think?

internal/cloud/gcp/compute.go
ondrejbudai: Isn't there an API?

internal/cloud/gcp/compute.go
```diff
@@ -288,3 +315,108 @@ func (g *GCP) ComputeDiskDelete(ctx context.Context, zone, disk string) error {
...
+		regionsMap[regionPrefix] = append(regionsMap[regionPrefix], regionObj.Name)
+	}
+
+	switch regionLower {
+	case "asia", "us":
+		return regionsMap[regionLower], nil
+	case "eu":
+		var euRegions []string
+		for _, euRegion := range regionsMap["europe"] {
+			// "europe-west2" (London) and "europe-west6" (Zurich) are excluded
```
ondrejbudai, line 374: why? Can you put here a link to docs?

internal/cloud/gcp/compute.go
```diff
@@ -288,3 +315,108 @@ func (g *GCP) ComputeDiskDelete(ctx context.Context, zone, disk string) error {
...
+			if euRegion != "europe-west2" && euRegion != "europe-west6" {
+				euRegions = append(euRegions, euRegion)
+			}
+		}
+		return euRegions, nil
+	default:
+		// This case should never happen, since the "default" case is handled above by
+		// if regionLower != "asia" && regionLower != "eu" && regionLower != "us"
+		return nil, fmt.Errorf("failed to translate Google Storage Regino '%s' to Compute Engine Region", regionLower)
```
ondrejbudai, line 383:
```suggestion
		return nil, fmt.Errorf("failed to translate Google Storage Region '%s' to Compute Engine Region", regionLower)
```

internal/cloud/gcp/compute.go
```diff
@@ -288,3 +315,108 @@ func (g *GCP) ComputeDiskDelete(ctx context.Context, zone, disk string) error {
...
+		var euRegions []string
+		for _, euRegion := range regionsMap["europe"] {
+			// "europe-west2" (London) and "europe-west6" (Zurich) are excluded
+			if euRegion != "europe-west2" && euRegion != "europe-west6" {
+				euRegions = append(euRegions, euRegion)
+			}
+		}
+		return euRegions, nil
+	default:
+		// This case should never happen, since the "default" case is handled above by
```
ondrejbudai, line 381: I would just `panic` if something can't happen. Not a strong opinion though.

ondrejbudai (commented):
Thanks for digging deep into this issue! I have some comments inline.

internal/cloud/gcp/compute.go
thozza: ahh yes, good catch!

internal/cloud/gcp/compute.go
thozza: It is the same link as on the line 351 (https://cloud.google.com/storage/docs/locations#location-mr). I can definitely put it here as well.

As for the reasoning, these regions are not in the European Union (sorry UK :smiley:)

internal/cloud/gcp/compute.go
thozza: Yeah, I was thinking about using panic here, I just don't like the idea of killing the whole process. I also don't have a strong preference :smile:

ondrejbudai: No worries then. :)

internal/cloud/gcp/compute.go
thozza: Unfortunately I was not able to find any relevant API.

There is an API for listing all Compute Engine regions and zones within each region and I'm using this API in the code. However the "dual" and "multi" Cloud Storage regions are defined on top of the GCE Regions, in addition to GCE Regions. The GCE API knows about its regions and the Storage API does not allow one to list regions at all.

internal/cloud/gcp/compute.go
thozza: I think that it makes sense to be consistent across the whole composer code base. Thanks.

internal/cloud/gcp/compute.go
ondrejbudai: Ah, that sucks, thanks for checking! :)

internal/cloud/gcp/compute.go
thozza: I added a link below the comment.

internal/cloud/gcp/compute.go
thozza: This is changed to `crypto/rand`.

thozza:
Thanks @ondrejbudai for the review and great comments. All changes are now reflected in the PR.

internal/cloud/gcp/compute.go
```diff
@@ -60,8 +63,43 @@ func (g *GCP) ComputeImageImport(ctx context.Context, bucket, object, imageName,
...
+		gceRegions, err := g.storageRegionToComputeRegions(ctx, region)
+		if err != nil {
+			return nil, fmt.Errorf("failed to translate Google Storage Region to GCE Region: %v", err)
+		}
+		// Pick a random GCE Region to be used by the image import workflow
+		gceRegionIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(gceRegions))))
+		if err != nil {
+			return nil, fmt.Errorf("failed to pick random GCE Region: %v", err)
+		}
+		// The expecation is that Google won't have more regions listed for multi/dual
```
ondrejbudai, line 86: Tell that to the inventors of IPv4. :stuck_out_tongue_winking_eye:

ondrejbudai (approved):
Amazing, thanks!

PR #1376, opened 2021-04-26
distribution: Add OCP template for clouddot

Gundersanne:
This is just an initial template, there will be some followups.

### DIFF
--- a/distribution/osbuild-composer-clouddot-template.yml
+++ b/distribution/osbuild-composer-clouddot-template.yml
@@ -1,0 +1,79 @@
+apiVersion: v1
+kind: Template
+labels:
+  app: osbuild-composer
+  template: osbuild-composer
+metadata:
+  annotations:
+    description: OCP template for osbuild-composer in cloud.redhat.com
+  name: osbuild-composer
+objects:
+
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    labels:
+      service: osbuild-composer
+    name: osbuild-composer
+  spec:
+    replicas: 1
+    selector:
+      matchLabels:
+        name: osbuild-composer
+    strategy:
+      # Update pods 1 at a time
+      type: RollingUpdate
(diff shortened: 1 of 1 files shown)


distribution/osbuild-composer-clouddot-template.yml
```diff
@@ -0,0 +1,79 @@
...
+objects:
+
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    labels:
+      service: osbuild-composer
+    name: osbuild-composer
+  spec:
+    replicas: 2
```
ondrejbudai, line 19: I think we want just 1 replica because osbulid-composer without a database isn't prepared to have multiple replicas.

distribution/osbuild-composer-clouddot-template.yml
```diff
@@ -0,0 +1,79 @@
...
+
+- apiVersion: v1
+  kind: Service
+  metadata:
+    labels:
+      service: osbuild-composer
+    name: osbuild-composer
+  spec:
+    ports:
+      - name: composer-api
```
ondrejbudai, line 62: I think we also want a port for workers because they will also live outside the OpenShift cluster.

Gundersanne: Indeed, I'm not 100% sure what it'll look like, but I think that'll probably end up being different openshift service.

The problem with composer exactly those 2 ports, as I don't think we'll get cloud.redhat.com:8700/..., so we might need to be clever with routing where `..:443/api/osbuild-composer/workers/....` gets redirected to  `:8700` in the container. We'll probably need a different service for that but not 100% sure.

ondrejbudai: okay, let's worry about that later. :) We might even want to merge these APIs to one port in composer.

ondrejbudai (approved):
Thanks!

PR #1381, opened 2021-04-28
Suport building for other distros/releases

bcl:
This pull request includes:

- [ ] 1st commit of any `rhbz#` related PR contains bug reproducer; CI reports FAIL or

(leaving the above alone for now)

What we have here is a bunch of changes that make it possible to build images for all the other supported distributions.
The blueprint now supports a `distro` field that tells it which distro to build for when a compose is started. If it is missing it builds for the host system's distro.

You can list supported distros with `curl --unix-socket /run/weldr/api.socket http://localhost:4000/api/v1/distros/list | python -m json.tool` -

```
{
    "distros": [
        "centos-8",
        "fedora-32",
        "fedora-33",
        "rhel-8",
        "rhel-84",
        "rhel-90"
    ]
}
```

Routes that could return different results based on distro selection now support an optional ?distro=DISTRONAME parameter.
This includes compose/types since image types are implemented per-distro there may be a different set available.

FYI - composer-cli has not been updated to support any of these new features. You can still use it to push a new blueprint though, it doesn't care what the content is.

Currently this doesn't limit what you can build. In production I think we want to limit building to similar releases of the host's distro (although, in my fedora-33 VM I can successfully build a centos-8 image). We need to figure out what these limits are, and they may be limited for support reasons more than technical reasons.

This does not include any changes to the sources API or TOML yet. The plan there is to default to supporting the host distro, or adding a `distros` list to the source to list the distros it supports. Possibly with wildcard? eg. `distros="fedora-*"`

Some of these commits still need cleanup and documentation. But this 'snapshot' of things builds, passes tests, and can build other distros, depsolve, etc.

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -1742,6 +1742,22 @@ func (api *API) blueprintsNewHandler(writer http.ResponseWriter, request *http.R
 		return
 	}
 
+	// Check the blueprint's distro to make sure it is valid
+	// or if it doesn't have one set it to the host distro.
+	if len(blueprint.Distro) > 0 {
+		_, ok := api.distroRepos[blueprint.Distro]
+		if !ok {
+			errors := responseError{
+				ID:  "BlueprintsError",
+				Msg: fmt.Sprintf("'%s' is not a valid distribution", blueprint.Distro),
+			}
+			statusResponseError(writer, http.StatusBadRequest, errors)
+			return
+		}
+	} else {
+		blueprint.Distro = api.hostDistroName
+	}
+
 	commitMsg := "Recipe " + blueprint.Name + ", version " + blueprint.Version + " saved."
 	err = api.store.PushBlueprint(blueprint, commitMsg)
 	if err != nil {
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -44,9 +44,8 @@ type API struct {
 	store   *store.Store
 	workers *worker.Server
 
-	rpmmd  rpmmd.RPMMD
-	distro distro.Distro
-	repos  []rpmmd.RepoConfig
+	rpmmd rpmmd.RPMMD
+	repos []rpmmd.RepoConfig
 
 	logger *log.Logger
 	router *httprouter.Router
@@ -108,7 +107,6 @@ func NewTestAPI(rpm rpmmd.RPMMD, arch distro.Arch, distro distro.Distro,
 	api := &API{
 		store:           store,
 		workers:         workers,
-		distro:          distro,
 		rpmmd:           rpm,
 		repos:           repos,
 		logger:          logger,
@@ -170,7 +168,6 @@ func New(repoPaths []string, stateDir string, rpm rpmmd.RPMMD,
 	api := &API{
 		store:           store,
 		workers:         workers,
-		distro:          hostDistro,
 		rpmmd:           rpm,
 		repos:           distroRepos[name][common.CurrentArch()],
 		logger:          logger,
@@ -191,6 +188,15 @@ func (api *API) arch(distro string) (distro.Arch, error) {
 		return nil, fmt.Errorf("Unsuppoered distro: %s", distro)
 	}
 	return d.GetArch(archName)
+}
+
+// platformID returns the module platform string for the selected distro
+func (api *API) platformID(distro string) (string, error) {
+	d := api.distros.GetDistro(distro)
+	if d == nil {
+		return "", fmt.Errorf("Unsupported distro: %s", distro)
+	}
+	return d.ModulePlatformID(), nil
 }
 
 func setupRouter(api *API) *API {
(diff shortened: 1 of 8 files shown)


internal/weldr/api.go
```diff
@@ -43,15 +44,16 @@ type API struct {
...
+	rpmmd rpmmd.RPMMD
 
 	logger *log.Logger
 	router *httprouter.Router
 
 	compatOutputDir string
+
+	hostDistroName string                                   // Name of the host distro
+	distros        *distroregistry.Registry                 // Supported distros
+	distroRepos    map[string]map[string][]rpmmd.RepoConfig // Repos for each distro
```
thozza, line 56: Out of curiosity, what's the motivation for keeping repositories for all architectures? Is it a potential future extension to allow specifying also architecture in the Blueprint and allowing building images on remote workers running on a different architecture?

bcl: Both that, and because the architecture code may differ between the distros. Each one has it's own code so in theory the could be different between distro/release.

thozza: Right, but to your last point, this is not actually anyhow reflected in the code. The architecture code for all distros is determined by calling `common.CurrentArch()` which effectively returns always the same arch code not based on what the distro uses.

internal/weldr/api.go
```diff
@@ -91,18 +93,114 @@ func (api *API) systemRepoNames() (names []string) {
...
+func (api *API) platformID(distro string) (string, error) {
+	d := api.distros.GetDistro(distro)
+	if d == nil {
+		return "", fmt.Errorf("Unsupported distro: %s", distro)
+	}
+	return d.ModulePlatformID(), nil
+}
+
+func (api *API) hostRepos() []rpmmd.RepoConfig {
+	return api.distroRepos[api.hostDistroName][common.CurrentArch()]
```
thozza, line 200: After reading the code, I think that this won't work if the `hostDistroName` will be `centos-stream-8` or `rhel-8-beta`. the `distroRepos` map does not reflect, that the same distro definition may be GA/Beta or non-Stream/Stream. In the end, the difference is IMO mainly in the repositories used for building the image, but this shortcut in distro definitions is actually making the problem here harder...

thozza (changes requested):

teg:
Thanks for all this work @bcl! Nice to see how quickly this came together.

Regarding the open question about which distros should be supported where, I think we have two options:

1. basic version: only the host distro on itself (Fedora on Fedora, RHEL on RHEL)
2. stretch goal: upstream builds downstream (Fedora, CentOS and RHEL on Fedora; CentOS and RHEL on CentOS; RHEL on RHEL)

We'd have to make sure we get the messaging right for option 2, but I think it could be very powerful.

@msehnout, @thozza and @ondrejbudai, we had a discussion about the subscription story, and my understanding was that subscribing non-RHEL hosts should work just fine, so as long as we do the integration right, the above will "just work". Did I get that right?

ondrejbudai:
Yes, you can subscribe Fedora and CentOS Stream. On my subscribed Fedora, I have a keypair in `291051867672977190.pem` but I'm not entirely sure which repos I can access with it. I know that when I'm running `ubi` in a `podman`, I can access RHEL repositories in the container so the keypair is definitely valid. As @thozza pointed out, subscriptions without Simple Content Access might make this more difficult.

thozza:
@teg @ondrejbudai let me check with the subscription-manager team to make sure we have accurate information, because I literally have no way to verify the behavior without SCA.

msehnout:
It is possible to subscribe Fedora and get RHEL content. For example using my subscription I can invoke:
```
$ curl --head --cacert /etc/rhsm/ca/redhat-uep.pem --cert /etc/pki/entitlement/6074731247270028676.pem --key /etc/pki/entitlement/6074731247270028676-key.pem https://cdn.redhat.com/content/dist/rhel/atomic/7/7Server/x86_64/cdk/3.3/os/repodata/repomd.xml
HTTP/1.1 200 OK
```
And get access to RHEL7 content.

internal/weldr/api.go
bcl: The host architecture name is from `common.CurrentArch()`, but if you look at the code in `internal/distro/*/distro.go there are different implementations for each distro+arch. `LoadRepositories()` already returns a map of the per-arch repos so we may as well keep them all, even if we don't use them yet.

internal/weldr/api.go
bcl: Ah, now I see what you mean. It would be cleaner if we didn't mangle names and every variant had it's own distro name. It's possible someone may want to build for beta vs. stream vs. a specific release.

thozza: Yeah, I fully agree. But given our current options of how distros are defined internally in composer, this would create additional copies for each distro (read as copies of Go code). I do believe that we should not do it before we have a better way of defining distros / arches / image types, including inheritance...

thozza:
Just a note to @msehnout comment, Martin tested it with employee subscription which has the SCA enabled. I'll post an update here once I have some information from the sub-man team regarding how this works without SCA.

internal/weldr/api.go
bcl: Will that kind of re-structuring be part of #1387? It's probably best to do that first and then I can rebase on it and adjust accordingly.

thozza: @bcl no, the restructuring of how are Distros defined won't be part of https://github.com/osbuild/osbuild-composer/pull/1387 but I included a change which will ensure that the `distros.FromHost()` returns a `distro.Distro` instance, which `Name()` returns the correctly "mangled" name. This is implemented by https://github.com/osbuild/osbuild-composer/pull/1387/commits/06077e5df2c54de5286d4fb26146ad6ed15c7a9b . I invite you to provide feedback if you have any.. :wink:

But I do believe that the distroregistry API and how it is used should not change after https://github.com/osbuild/osbuild-composer/pull/1387 is merged, even when we restructure how are Distros defined. I think that the abstractions should be already good enough to hide such work from Weldr's point of view.

bcl:
See #1427

PR #1384, opened 2021-05-04
[POC] Run integration tests in gitlab ci

jrusz:
This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,6 +8,7 @@ stages:
   after_script:
     - schutzbot/update_github_status.sh update
     - schutzbot/save_journal.sh
+    - schutzbot/run_cloud_cleaner.sh
   tags:
     - terraform
   artifacts:
@@ -80,6 +81,34 @@ OSTree:
       - RUNNER:
           - openstack/fedora-33-x86_64
           - openstack/rhel-8-x86_64
+
+Integration:
+  stage: test
+  extends: .terraform
+  variables:
+    EXTRA_REPO_PATH_SEGMENT: "gitlab/"
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/koji.sh
+    - /usr/libexec/tests/osbuild-composer/libvirt.sh
+    - /usr/libexec/tests/osbuild-composer/aws.sh
+    - /usr/libexec/tests/osbuild-composer/azure.sh
+    - /usr/libexec/tests/osbuild-composer/api.sh azure
+    - /usr/libexec/tests/osbuild-composer/api.sh aws
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/fedora-32-x86_64
+          - aws/fedora-33-x86_64
+          - aws/centos-stream-8-x86_64
+      - RUNNER:
+          - aws/rhel-8-x86_64
+          - aws/rhel-8.4-x86_64
+        INTERNAL_NETWORK: ["true"]
+      - RUNNER:
+          - aws/rhel-8-x86_64
+        INTERNAL_NETWORK: ["true"]
+        DISTRO_CODE: ["rhel_90"]
 
 finish:
   stage: finish
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -815,34 +815,41 @@ void run_tests(test_type) {
 
     if (test_type == 'integration') {
         // Run Koji tests.
-        sh (
-            label: "Koji tests",
-            script: "/usr/libexec/tests/osbuild-composer/koji.sh"
-        )
+        if (env.BUILD_CAUSE == 'cron') {
+            sh (
+                label: "Koji tests",
+                script: "/usr/libexec/tests/osbuild-composer/koji.sh"
+            )
 
-        // Run the libvirt test.
-        sh (
-            label: "Integration test: libvirt",
-            script: "/usr/libexec/tests/osbuild-composer/libvirt.sh"
-        )
+            // Run the libvirt test.
+            sh (
+                label: "Integration test: libvirt",
+                script: "/usr/libexec/tests/osbuild-composer/libvirt.sh"
+            )
 
-        // Run the AWS test.
-        sh (
-            label: "Integration test: AWS",
-            script: "/usr/libexec/tests/osbuild-composer/aws.sh"
-        )
+            // Run the AWS test.
+            sh (
+                label: "Integration test: AWS",
+                script: "/usr/libexec/tests/osbuild-composer/aws.sh"
+            )
 
-        // Run the Azure test.
-        sh (
-            label: "Integration test: AZURE",
-            script: "/usr/libexec/tests/osbuild-composer/azure.sh"
-        )
+            // Run the Azure test.
+            sh (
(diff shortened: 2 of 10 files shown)


.gitlab-ci.yml
```diff
@@ -32,49 +32,64 @@ RPM:
...
+#      - RUNNER:
+#          - openstack/fedora-33-x86_64
+#          - openstack/rhel-8-x86_64
+#
+# try just libvirt first
+Integration:
   stage: test
   extends: .terraform
   variables:
     EXTRA_REPO_PATH_SEGMENT: "gitlab/"
```
atodorov, line 89: This will need all of the environment variables from Jenkinsfile (note, TEST_TYPE doesn't seem to be used, DISTRO_CODE can be removed in favor of /etc/os-release). 

@ondrejbudai do you mind configuring some of the credentials in GitLab or tell us which ones are already configured ?

Some links I've found on the topic:
https://docs.gitlab.com/ee/ci/variables/#project-cicd-variables (Project or Group variables with masking)
https://docs.gitlab.com/ee/ci/secrets/
https://secrethub.io/docs/guides/gitlabci/

atodorov: Current job fails b/c of the missing DISTRO_CODE variable.

ondrejbudai: You can see all the configured credentials in the GitLab UI. I think we cannot remove `DISTRO_CODE` that easily because it's used to distinguish RHEL 9 tests (they run on RHEL 8).

jrusz: > Current job fails b/c of the missing DISTRO_CODE variable.

Yeah, I'm currently trying to figure out the best way to define variables specific to each runner like we had in Jenkinsfile for each stage...

atodorov: IMO the should either be specific to the job we're running (base vs integration, etc) or common for all jobs. There's no need for specific variables per runner b/c the test jobs on different OS/arch combo still need the same env variables.

jrusz: Yeah it seems DISTRO_CODE is the only one that is runner specific... It's not used in that many places so it could be replaced easily with something like ID+VERSION_ID from /etc/os-release like you suggested.. Defining variables per groups of jobs should be easy then, there is already one defined in the .gitlab-ci.yaml and secrets can be added in settings in the gitlab repository...

jrusz:
I got a failure for koji tests here https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1237992586#L2206 ... How could this have ever worked with 8.4? The DISTRO_CODE would be "rhel_84" and it would fail the same way because there's not such file under /repositores either.... Can you maybe shed some light on this @ondrejbudai ?

ondrejbudai:
DISTRO_CODE is set to rhel_8 on RHEL 8 GA tests.

jrusz:
I currently don't know what to do with RHEL 8.4 and Fedora-34 so I commented them out. I modified scripts that use DISTRO_CODE to initialize it in case it's not provided and then made a few small modifications to allow the tests to run in gitlab CI. Some of the modifications could probably be done better, so I'm open to suggestions. I didn't move vmware and gcp testing yet. Vmware is because I'm not yet sure how to make it run only on rhel-8 and gcp because it uses a file as credential and we can't mask that inside gitlab ci. I've also removed the integration tests from Jenkinsfile and was wondering whether the same should be done for other tests that already run in gitlab CI like base and ostree...

.gitlab-ci.yml
```diff
@@ -35,15 +36,15 @@ RPM:
           - aws/fedora-32-x86_64
           - aws/fedora-33-x86_64
           - aws/fedora-33-aarch64
-          - aws/fedora-34-x86_64
-          - aws/fedora-34-aarch64
+#          - aws/fedora-34-x86_64
+#          - aws/fedora-34-aarch64
```
atodorov, line 40: I think this is solved in #1393

.gitlab-ci.yml
```diff
@@ -35,15 +36,15 @@ RPM:
...
           - aws/rhel-8-x86_64
           - aws/rhel-8-aarch64
           - aws/centos-stream-8-x86_64
           - aws/centos-stream-8-aarch64
-      - RUNNER:
-          - aws/rhel-8.4-x86_64
-        INTERNAL_NETWORK: ["true"]
+#      - RUNNER:
+#          - aws/rhel-8.4-x86_64
+#        INTERNAL_NETWORK: ["true"]
```
atodorov, line 47: What is the actual problem that you are getting here ?

atodorov:
> I've also removed the integration tests from Jenkinsfile and was wondering whether the same should be done for other tests that already run in gitlab CI like base and ostree...

Bring those back b/c they will break the nightly pipeline execution. All of them should be removed once GitLab CI is configured to trigger y cron every day & has something similar to `BUILD_CAUSE = detect_build_cause()`.

.gitlab-ci.yml
jrusz: The missing 8.4 latest nightly symlink... https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1243772878 But I can see they brought it back now.. I need to ask someone what's happening.

atodorov (approved):

ondrejbudai (changes requested):
Looks good! The only remaining issue is that the integration tests are not run for RHEL 9.

ondrejbudai:
You should be able to just do:
```

  parallel:
    matrix:
      - RUNNER:
          - aws/fedora-32-x86_64
          - aws/fedora-33-x86_64
          - aws/centos-stream-8-x86_64
      - RUNNER:
          - aws/rhel-8-x86_64
          - aws/rhel-8.4-x86_64
        INTERNAL_NETWORK: ["true"]
      - RUNNER:
          -  aws/rhel-8-x86_64
         INTERNAL_NETWORK: ["true"]
         DISTRO_CODE: ["rhel_90"]
```

atodorov (approved):
LGTM, waiting for the latest executions to complete.

jrusz:
> You should be able to just do:
> 
> ```
> 
>   parallel:
>     matrix:
>       - RUNNER:
>           - aws/fedora-32-x86_64
>           - aws/fedora-33-x86_64
>           - aws/centos-stream-8-x86_64
>       - RUNNER:
>           - aws/rhel-8-x86_64
>           - aws/rhel-8.4-x86_64
>         INTERNAL_NETWORK: ["true"]
>       - RUNNER:
>           - aws/rhel-8-x86_64
>         INTERNAL_NETWORK: ["true"]
>         DISTRO_CODE: ["rhel_90"]
> ```

Thanks, that worked. Can we merge this now? Jenkins is stuck on rhel8 aarch64 base tests, the same test passed in gitlab ci... I've seen this several times already with the base tests on aarch64, I'll do a little manual investigation...

ondrejbudai:
> Jenkins is stuck on rhel8 aarch64 base tests, the same test passed in gitlab ci... I've seen this several times already with the base tests on aarch64, I'll do a little manual investigation...

It's probably this bug: https://bugs.launchpad.net/qemu/+bug/1805256
There's a known workaround but I haven't got to applying it yet.

atodorov:
My vote goes for merging this PR!

ondrejbudai (approved):
LGTM

PR #1387, opened 2021-05-05
Add RepoRegistry API and use it for Weldr API (+ bunch of unit tests changes)

thozza:
This pull request includes:

    - **TODO after the discussion on this PR settles.**

The main intention of this PR is to offload the work of loading and managing repo definitions from the composer and Weldr API code. A new internal package `reporegistry` is added for this purpose. There have been other changes done as a side effect of the work and most of them affects unit tests. More details below:

 - Extend the "Test Distro" implementation and definition to contain two architectures and make the second architecture contain two image types. Add New2() function returning another "Test Distro".
- Modify the `internal/store` unit tests to reflect changes done to the "Test Distro".
- Remove fedoratest distro and replace it with test_distro in all affected tests.
- Separate the loading of repo definitions from JSON file from `LoadRepositories()` to a standalone function `loadRepositoriesFromFile()`, to make it easy to reuse it in the future.
- Add unit tests for `LoadRepositories()` function.
- Extend the RepoConfig structure to contain new field ImageTypeTags. Extend also other structures and functions as needed, to support loading repository definitions, which use this new field. The idea is that a repository should be used for building all image types, unless it has some ImageTypeTags defined. In such case, it should be used only for building the specific image types, which names are specified in the new field.
- Add RepoRegistry as a higher-level API to load and manage repository definitions for each distribution. Currently it provides one method, which returns a set of repositories needed to build a given image type. The RepoRegistry uses the new ImageTypeTags field in the RepoConfig structure and returns all the needed repositories for the image type.
- Modify rpmmd unit tests and add unit tests for RepoRegistry.
- Add News entry describing the change done to RepoConfig and its JSON representation.
- Modify composer to use RepoRegistry, instead of loading the host repositories, when initializing WeldrAPI.
- Modify WeldrAPI to use RepoRegistry, instead of a map of repository definitions. Make sure that the RepoRegistry method specific to image type is used in Welder where appropriate. Specifically when depsolving a Blueprint, which is used to build a specific image type. Update Weldr API unit tests to reflect the change.
- Add a new method to RepoRegistry, allowing to get list of repositories, which should be used for building an image for a given architecture, without specifying the exact image type. Add relevant unit tests.
- Add `New*HostDistro()` to all distro definitions, accepting the name that should be returned by the distro's `Name()` method. This is useful mainly if the host distro is Beta or Stream variant of the distro.
- Change the distroregistry.Registry to contain host distro as a separate value set when creating it using `New()` function. This value is returned by `Registry.FromHost()` method. Determining the host distro is handled by the `NewDefault()` function. Move the distro name mangling to distroregistry package. Add relevant unit tests.


cmd/osbuild-composer/composer.go
```diff
@@ -99,15 +100,17 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener) err
...
 	}
 
-	repos, err := rpmmd.LoadRepositories(repoPaths, name)
+	rr, err := reporegistry.New(repoPaths)
 	if err != nil {
-		return fmt.Errorf("Error loading repositories for %s: %v", hostDistro.Name(), err)
+		return fmt.Errorf("error loading repositories definitions: %v", err)
 	}
 
+	// TODO: should we check here if repositories for the host distro and arch were loaded?
```
thozza, line 108: So the code previously failed here if the distro repo definitions were not found. Now it will fail later, when a blueprint is depsolved or compose created. Should we check this here and fail immediately?

internal/weldr/api.go
```diff
@@ -82,23 +83,26 @@ func (cs ComposeState) ToString() string {
 // systemRepoIDs returns a list of the system repos
 // NOTE: The system repos have no concept of id vs. name so the id is returned
 func (api *API) systemRepoNames() (names []string) {
-	for _, repo := range api.repos {
-		names = append(names, repo.Name)
+	repos, err := api.repoRegistry.ReposByArch(api.arch, false)
```
thozza, line 86: I would like to discuss, whether the image-type-specific repos should be treated as System Repos. Currently they are not and @bcl didn't want to treat them in this way. However as a result of not treating them this way, the user will be able to define sources with the same name as the image-type-specific repository. I'm not sure if this will be a problem or not.

internal/weldr/api.go
```diff
@@ -465,17 +469,26 @@ func (api *API) getSourceConfigs(params httprouter.Params) (map[string]store.Sou
 	sources := map[string]store.SourceConfig{}
 	errors := []responseError{}
 
+	repos, err := api.repoRegistry.ReposByArch(api.arch, false)
```
thozza, line 472: Also here, should the user be able to get config of a image-type-specific repo/source? Currently they won't be able to do so.

cmd/osbuild-composer/composer.go
```diff
@@ -99,15 +100,17 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener) err
 		name = "centos-stream-8"
 	}
 
-	repos, err := rpmmd.LoadRepositories(repoPaths, name)
+	rr, err := reporegistry.New(repoPaths)
```
ondrejbudai, line 103: hmm, `name` is no longer used. This needs to be addressed. Not sure how exactly though...

ondrejbudai (commented):
I love this, seems to make things much clearer. Few notes inline.

internal/reporegistry/reporegistry.go
```diff
@@ -0,0 +1,78 @@
...
+
+// ReposByImageType returns a slice of rpmmd.RepoConfig instances, which should be used for building the specific
+// image type. All repositories for the associated distribution and architecture, without any ImageTypeTags set,
+// are always part of the returned slice. In addition, if there are repositories tagged with the specific image
+// type name, these are added to the returned slice as well.
+func (r *RepoRegistry) ReposByImageType(imageType distro.ImageType) ([]rpmmd.RepoConfig, error) {
+	if imageType.Arch() == nil || reflect.ValueOf(imageType.Arch()).IsNil() {
+		return nil, fmt.Errorf("there is no architecture associated with the provided image type")
+	}
+	fmt.Printf("%#v\n", imageType.Arch())
```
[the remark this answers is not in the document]
thozza: ah, yes. Thanks!

cmd/osbuild-composer/composer.go
thozza: Yes, it's on my TODO for today. I'll have to move the "name mangling" to the Weldr API, similar as @bcl did it in his PR. I'll probably just copy his change 1:1 (Thanks @bcl :-D)

The thing is that the `Name()` method of the distro instance passed to Weldr API won't return the expected name, so Weldr would not be able to get the hostDistro repositories from the RepoRegistry.

internal/rpmmd/repository.go
```diff
@@ -245,6 +248,50 @@ func loadRepositoriesFromFile(filename string) (map[string][]RepoConfig, error)
 	return repoConfigs, nil
 }
 
+// LoadAllRepositories loads all repositories for given distros from the given list of paths.
+// Behavior is the same as with the LoadRepositories() method.
+func LoadAllRepositories(confPaths []string) (map[string]map[string][]RepoConfig, error) {
```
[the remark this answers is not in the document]
thozza: I really don't have a strong preference here. I Find it still readable enough, but I'm not against defining type aliases. Let me think about it.

internal/reporegistry/reporegistry.go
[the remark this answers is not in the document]
thozza: Removed.

cmd/osbuild-composer/composer.go
thozza: This is fixed by https://github.com/osbuild/osbuild-composer/pull/1387/commits/4885e2d3a728fd91615eed991d569d89d005e1bb

cmd/osbuild-composer/composer.go
bcl: For now I think it should keep the same behavior. I think @teg has mentioned that eventually we should have a fallback distro so that if it doesn't match exactly it will still run, but that's a separate thing to work on.

internal/weldr/api.go
bcl: I think those repos make sense as a separate namespace that users don't see and cannot touch.
The whole system repo thing is an artifact of how lorax-composer worked, combining the host repos with custom user repos -- the system repos came from /etc/yum.repos.d/ at startup, and the user repos were stored in lorax-composer. Now that we don't depend directly on the host repo files I think it makes sense to hide anything the user cannot change since there's nothing they can do with them anyway.

cmd/osbuild-composer/composer.go
thozza: OK, failing right away makes sense from my point of view for now, as it keeps the current behavior.

internal/weldr/api.go
thozza: @bcl hmm, interesting thought. Hiding anything the user can not change would actually imply also hiding the "system" repositories. Is that what you are suggesting?

bcl: yes, but I suppose we should ask someone working on the UI if they think it is important to show them.

internal/rpmmd/repository.go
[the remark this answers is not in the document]
thozza: I defined a type alias `rpmmd.DistrosRepoConfigs`

cmd/osbuild-composer/composer.go
thozza: I added the check.

internal/weldr/api.go
jkozol: Even though the user cannot change the default repos I think it is still good to show the user where the packages are being pulled from.

docs/news/unreleased/repo-conf-image-type-tags.md
```diff
@@ -0,0 +1,82 @@
+# Allow image type-specific repositories using Image Type Tags
+
+The schema of the repository definitions used by *Weldr API*, located in `/usr/share/osbuild-composer/repositories/` or `/etc/osbuild-composer/repositories` is extended with a new field called **`image_type_tags`** and is expected to be an array of strings representing specific image types.
+
+The behavior of how are defined repositories processed and used by  osbuild-composer* is extended in the following way:
```
msehnout, line 5:
```suggestion
The behavior of how are defined repositories processed and used by *osbuild-composer* is extended in the following way:
```

msehnout (approved):
Looks good to me! Nice work!

Unfortunately, I don't have an answer to your question about displaying sources. I find the whole concept of system repos vs sources confusing and as you mentioned, there are no clear boundaries.

One suggestion below:

ondrejbudai (approved):
The changes in distro names are interesting but nobody was able to come up with anything better so I guess let's roll with it for now. This is not a public API so we can always change it later if we change our minds.

ondrejbudai:
Forgot to write: Thanks a lot for this! :)

PR #1386, opened 2021-05-05
[WIP] Regression test for the UID bug

msehnout:
This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #1395, opened 2021-05-07
test cases: Introduce a directory for regression test cases

msehnout:
I'm about to propose 2 more regression test cases next week:
 * https://github.com/osbuild/osbuild-composer/pull/1386
 * a test case for the issue with Satellite repos

Therefore I thought it'd be a good idea to introduce a structure in the regression test cases and possibly a reasonable naming scheme. This PR is my proposal.

This pull request includes:

- N/A adequate testing for the new functionality or fixed issue
  - N/A create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news
  - N/A submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/
  - [X] Entry in the `test/README.md` file

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -191,6 +191,9 @@ install -m 0755 -vp tools/libvirt_test.sh                       %{buildroot}%{_l
 install -m 0755 -vd                                             %{buildroot}%{_libexecdir}/tests/osbuild-composer
 install -m 0755 -vp test/cases/*                                %{buildroot}%{_libexecdir}/tests/osbuild-composer/
 
+install -m 0755 -vd                                             %{buildroot}%{_libexecdir}/tests/osbuild-composer/regression-cases
+install -m 0755 -vp test/regression-cases/*                     %{buildroot}%{_libexecdir}/tests/osbuild-composer/regression-cases/
+
 install -m 0755 -vd                                             %{buildroot}%{_datadir}/tests/osbuild-composer/ansible
 install -m 0644 -vp test/data/ansible/*                         %{buildroot}%{_datadir}/tests/osbuild-composer/ansible/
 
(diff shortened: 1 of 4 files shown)


jrusz (commented):
I guess that's fine, I'd appreciate a link to the original bug \ somewhere in the test as a comment though, it would make it easier to understand what's exactly being tested...  In case there's more cases are they going to be like separate bash scripts? That could use some kind of common script to run them all so that we don't have to add them one by one to run in the CI...

msehnout:
> I guess that's fine, I'd appreciate a link to the original bug somewhere in the test as a comment though, it would make it easier to understand what's exactly being tested... In case there's more cases are they going to be like separate bash scripts? That could use some kind of common script to run them all so that we don't have to add them one by one to run in the CI...

The link makes sense, even though it can point to both Github issues and RH bugzilla (where some comments are private).

Yes, my plan is to introduce each of them as a separate bash script. The reason is that not all cases work on all distros. For example the only test we currently have only works for image types that exclude some package and this package is required by another one. If we wanted to make this work on Fedora, we'd have to use different packages than "redhat-lsb-core" and "nss". Which is doable of course.

This test (https://github.com/osbuild/osbuild-composer/pull/1386) will only work for RHEL8.4 as it is currently the only one that supports realtime kernel and can produce ostree commits.

Do you have any proposal regarding how the common script could work? I could imagine a script that would have a switch and it would list only the cases that are relevant to given distro and arch.

atodorov:
FTR I like #1406 better than this one.

msehnout:
Closing this in favor of https://github.com/osbuild/osbuild-composer/pull/1406

PR #1399, opened 2021-05-10
[TEST ONLY] GCE image type using reporegistry

thozza:
#1365 rebased on #1387 using `reporegistry`. For testing only.


ondrejbudai:
#1387 is merged so I don't think we need this anymore. Feel free to reopen if I'm wrong.

PR #1391, opened 2021-05-06
Add support for RHEL 8.5

achilleas-k:
This pull request includes:

- [x] adequate testing for the new functionality or fixed issue: _basic testing of arch listing_
- [x] adequate documentation informing people about the change such as
  - [x] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news: _not needed_
  - [x] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/: _not needed_

Cherry-picked commits from #1351 that add support for RHEL 8.5 without any defined image types.

This adds the basic functions and types for defining RHEL 8.5 image types using the new v2 manifest schema of osbuild.

Helper functions for defining stages and pipelines are added but are based on the Edge image types of the original PR, so they will probably need to be adjusted as new image types are added.

Each image type defines its own top-level `pipelines()` function now.  I kept the edge type pipeline definitions from the original previous PR, but they are "dead code" for now since they're not attached to any image type definition.  I can remove them from the PR and add them when the edge image types are added if anyone would prefer the distro definition to be cleaner.

With this PR, osbuild-composer can now run on RHEL 8.5, but since no image types are defined, there's not much you can do with it:
```
composer-cli --json compose types
{
    "types": null
}
```

PR #1401, opened 2021-05-10
RHEL 8.5: tar installer

achilleas-k:
This pull request includes:

Implements a `tar-installer`: an installer image that embeds an OS tarball (same as the `tar` image type) that's installed when the ISO is booted.

I had this implemented and ready for a while based on #1351.  Rebased now on #1391.  Opening now for early review but keeping as a draft until we #1391 is merged and to add tests.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -80,7 +80,7 @@ Prepare-rhel-internal:
   parallel:
     matrix:
       - RUNNER:
-          - aws/rhel-8.4-x86_64
+          - aws/rhel-8.5-x86_64
         INTERNAL_NETWORK: ["true"]
 
 Base:
@@ -107,6 +107,7 @@ Base:
           - aws/centos-stream-8-aarch64
       - RUNNER:
           - aws/rhel-8.4-x86_64
+          - aws/rhel-8.5-x86_64
           # see COMPOSER-918
           # - aws/rhel-8.4-aarch64
         INTERNAL_NETWORK: ["true"]
@@ -223,6 +224,20 @@ API:
           - aws/rhel-8-x86_64
         INTERNAL_NETWORK: ["true"]
         DISTRO_CODE: ["rhel_90"]
+
+Installer:
+  stage: test
+  extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[\S]+/'
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/installers.sh
+  parallel:
+    matrix:
+      - RUNNER:
+          - openstack/rhel-8.5-x86_64
 
 finish:
   stage: finish
--- a/internal/distro/rhel85/distro_test.go
+++ b/internal/distro/rhel85/distro_test.go
@@ -227,6 +227,7 @@ func TestArchitecture_ListImageTypes(t *testing.T) {
 				"edge-commit",
 				"edge-container",
 				"edge-installer",
+				"tar",
 				"tar-installer",
 			},
 		},
@@ -236,15 +237,16 @@ func TestArchitecture_ListImageTypes(t *testing.T) {
 				"edge-commit",
 				"edge-container",
 				"edge-installer",
+				"tar",
 			},
 		},
 		{
 			arch:     "ppc64le",
-			imgNames: []string{},
+			imgNames: []string{"tar"},
 		},
 		{
 			arch:     "s390x",
-			imgNames: []string{},
+			imgNames: []string{"tar"},
 		},
 	}
 
--- a/internal/distro/rhel85/distro_test.go
+++ b/internal/distro/rhel85/distro_test.go
@@ -227,6 +227,7 @@ func TestArchitecture_ListImageTypes(t *testing.T) {
 				"edge-commit",
 				"edge-container",
 				"edge-installer",
+				"tar-installer",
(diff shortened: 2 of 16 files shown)


internal/distro/rhel85/distro_test.go
```diff
@@ -30,20 +30,23 @@ func TestArchitecture_ListImageTypes(t *testing.T) {
 		rhelAdditionalImageTypes []string
 	}{
 		{
-			arch:     "x86_64",
-			imgNames: []string{},
+			arch: "x86_64",
+			imgNames: []string{
+				"tar",
+				"tar-installer",
+			},
```
atodorov, line 37: From what I can see this is verifying that we've got these 2 new image types. However I don't see any test which tries to build the new image types and makes sure nothing fails. `cmd/osbuild-image-tests/main_test.go` may be a good place to start. Alternatively a separate integration test would do as well.

achilleas-k: Thanks for the review, Alex.  I was thinking an integration test that builds the installer and auto-installs it, similar to the ostree integration tests, would be a good idea for the tar-installer.

The plain tar image type existed in the previous distros, so we could follow the same testing here.

achilleas-k:
Rebased on `main` and added an integration test that builds an installer, installs it and checks for installed packages (I based it on `ostree.sh`).

achilleas-k:
@atodorov or @ondrejbudai I haven't enabled the integration test in the `.gitlab-ci.yml` file yet.  Is the 8.5 runner ready?

teg:
Really nice @achilleas-k !

@jikortus is this something you could use/extend as part of your work?

atodorov (changes requested):
@achilleas-k you should be able to rebase to latest main and see your test jobs running on the 8.5 runners.

.gitlab-ci.yml
```diff
@@ -224,6 +226,22 @@ API:
...
         DISTRO_CODE: ["rhel_90"]
 
+Installer:
+  stage: test
+  extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[\S]+/'
+  variables:
+    EXTRA_REPO_PATH_SEGMENT: "gitlab/"
```
atodorov, line 236: This has been removed by @ondrejbudai in #1510. You can drop the EXTRA_REPO_PATH_SEGMENT variable.

.gitlab-ci.yml
```diff
@@ -81,6 +81,7 @@ Prepare-rhel-internal:
     matrix:
       - RUNNER:
           - aws/rhel-8.4-x86_64
+          - aws/rhel-8.5-x86_64
```
atodorov, line 84: This stage is used when testing the latest nightly builds so IMO use only 8.5, not both.

atodorov (changes requested):
Minor changes requested, otherwise looks good.

.gitlab-ci.yml
achilleas-k: Removed.  Thanks.

.gitlab-ci.yml
achilleas-k: Switched to 8.5 only.

achilleas-k:

achilleas-k:
Had to update the test manifests for the edge image types introduced in #1351.  The core stages that are shared between all image types didn't include the fix-bls stage before the installer type was added.

internal/distro/rhel85/distro_test.go
atodorov: @achilleas-k please open a ticket in JIRA with the `QE` prefix for this tar integration test if you haven't already so that we don't forget about it and make sure to plan for the work accordingly.

.gitlab-ci.yml
```diff
@@ -107,6 +107,7 @@ Base:
           - aws/centos-stream-8-aarch64
       - RUNNER:
           - aws/rhel-8.4-x86_64
+          - aws/rhel-8.5-x86_64
```
atodorov, line 110: FTR this is going to fail b/c of the other missing images. Once the CI job passes we should vet the tests which failed, there should be only a few of them which either use qcow2 or try to compare the list of available image types. Everything else should report PASS.

internal/distro/rhel85/distro_test.go
atodorov: ^^^ I see installers.sh. Closing this comment.

atodorov (approved):
LGTM with comments.

.gitlab-ci.yml
achilleas-k: What's the plan then?  Are we keeping this open until all the other image types land?  I don't mind either way.

teg: I guess the plan is up to you ;) Would it be a lot of churn to only enable the tests we expect to pass for this PR, and then enable the rest when the other image types are ready?

atodorov: @achilleas-k let's not keep PRs open longer than they need to. The fact that there's missing support for image types (which has the nice side effect of breaking our test suite) is not related to the work around tar installer.

ondrejbudai:
Please, don't merge PRs that are clearly untested by the CI. For this PR, none of the tests were actually run because the CI actually failed on the RPM build step. I hope this won't break `main`.

image

ondrejbudai:
Just a note why it failed on rpms: This was caused by the outage of RHSM servers.

atodorov:
My bad, sorry. I didn't inspect the GitLab results this time and relied on the last time I saw things were looking good.

PR #1403, opened 2021-05-11
tests: Define COMPOSE_URL/COMPOSE_ID in a single place. Fix #1397

### LINKED ISSUE #1397: tests: Define COMPOSE_URL via separate script
Context #1388 and #1351 and the existing prepare-rhel-internal.sh script. 

This variable is now being used in several places all of which need to define it if it doesn't already exist. The default logic should be moved to a small script which can then be called from the top-most level (GitLab or Jenkinsfile) to define this variable so that downstream scripts will be able to consume it.

Alternatively downstream scripts will just call the helper script to set the necessary values.

atodorov:
we don't want this defined multiple times in different places

This pull request includes:


ondrejbudai (dismissed):
looks good, thanks!

ondrejbudai (approved):

PR #1405, opened 2021-05-11
make composer work with multiple different subscriptions

msehnout:
Current status:
 * ready for review

This pull request includes:

Related:
https://github.com/osbuild/guides/pull/36
osbuild/osbuild#645


test/cases/satellite.sh
```diff
@@ -0,0 +1,220 @@
...
+  Require all denied
+</Directory>
+STOPHERE
+}
+
+# Provision the software under tet.
+/usr/libexec/osbuild-composer-test/provision.sh
+
+greenprint "Installing dependencies"
+sudo dnf install -y openssl httpd mod_ssl
```
ondrejbudai, line 73: This should go to the spec file. :)

test/cases/satellite.sh
```diff
@@ -0,0 +1,220 @@
...
+# Access log
+LogFormat "%h %l %u %t \"%r\" %>s %b" common
+CustomLog logs/access_log common
+
+# Never change this block
+<Directory />
+  AllowOverride None
+  Require all denied
+</Directory>
+STOPHERE
```
atodorov, line 66: Doesn't httpd come with a working config file? What's the difference here ?

test/cases/satellite.sh
```diff
@@ -0,0 +1,220 @@
...
+   SSLEngine on
+   SSLProxyEngine on
+   SSLCertificateFile ca/server.crt
+   SSLCertificateKeyFile ca/server.key
+   # And require the client to authenticate using a certificate issued by our custom CA
+   SSLVerifyClient require
+   SSLVerifyDepth 1
+   SSLCACertificateFile ca/ca.crt
+</VirtualHost>
+STOPHERE
```
atodorov, line 143: I think you should be able to drop this under `/etc/httpd/conf.d/` instead of overriding httpd.conf

test/cases/satellite.sh
```diff
@@ -0,0 +1,220 @@
...
+description = "A base system with fish"
+version = "0.0.1"
+
+[[packages]]
+name = "fish"
+EOF
+
+sudo composer-cli blueprints push "$BLUEPRINT_FILE"
+sudo composer-cli blueprints depsolve ${BLUEPRINT_NAME}
+sudo composer-cli --json compose start ${BLUEPRINT_NAME} qcow2
```
atodorov, line 220: missing newline at EOF.

test/cases/satellite.sh
```diff
@@ -0,0 +1,220 @@
...
+      "check_gpg": false,
+      "rhsm": true
+    }
+  ]
+}
```
atodorov, line 200: NOTE: this override will potentially conflict with other internal overrides that we do, especially when this test is enabled on RHEL, see #1318. 

This test script should be executed on a stand-alone runner so that it doesn't affect other executions afterwards.

schutzbot/Jenkinsfile
```diff
@@ -814,58 +814,60 @@ void run_tests(test_type) {
     }
 
     if (test_type == 'integration') {
+        sh (
+            label: "Satellite test",
+            script: "/usr/libexec/tests/osbuild-composer/satellite.sh"
+        )
```
atodorov, line 820: This also needs to be added into GitLab but is currently blocked by #1408 !  Also see https://github.com/osbuild/osbuild-composer/pull/1405#discussion_r630822032

msehnout:
Depends on: https://github.com/osbuild/osbuild/pull/645

internal/distro/fedora32/distro.go
```diff
@@ -17,6 +17,7 @@ import (
 )
 
 const defaultName = "fedora-32"
+const releaseVersion = "32"
```
ondrejbudai, line 20: I think we should reference `releaseVersion` in the other constants (`defaultName`, `modulePlatformID`, `ostreeRef`). Something for a follow-up.

internal/rpmmd/repository.go
```diff
@@ -385,20 +362,22 @@ func runDNF(dnfJsonPath string, command string, arguments interface{}, result in
...
-	RHSM        *RHSMSecrets
-	dnfJsonPath string
+	CacheDir      string
+	subscriptions *rhsm.Subscriptions
+	dnfJsonPath   string
 }
 
 func NewRPMMD(cacheDir, dnfJsonPath string) RPMMD {
+	// Expect that errors in loading are caused by missing subscriptions.
+	subscptions, _ := rhsm.LoadSystemSubscriptions()
```
ondrejbudai, line 372: We should at least log the error. Hiding it makes debugging difficult. Also `s/subscptions/subscriptions`. :)

internal/distro/fedora32/distro.go
```diff
@@ -267,6 +267,11 @@ func sources(packages []rpmmd.PackageSpec) *osbuild.Sources {
 		fileSource := osbuild.FileSource{
 			URL: pkg.RemoteLocation,
 		}
+		if pkg.Secrets == "org.osbuild.rhsm" {
```
ondrejbudai, line 270: Is there any usecase for Fedora and rhsm?

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -0,0 +1,259 @@
...
+      "check_gpg": false,
+      "rhsm": true
+    }
+  ]
+}
+STOPHERE
+
+# TODO: remove this once it is in released version of osbuild
```
ondrejbudai, line 186: Just so we don't forget: This needs to go away before we merge this.

ondrejbudai (commented):
Awesome stuff, thanks! Looking forward to shipping this!

internal/rhsm/secrets.go
```diff
@@ -0,0 +1,108 @@
...
+// client certificate for a specific repository.
+type RHSMSecrets struct {
+	SSLCACert     string
+	SSLClientKey  string
+	SSLClientCert string
+}
+
+// LoadSystemSubscriptions loads all the available subscriptions.
+func LoadSystemSubscriptions() (*Subscriptions, error) {
+	availableSubscriptionsFile := "/etc/yum.repos.d/redhat.repo"
```
teg, line 35: Is the special semantics of this file documented anywhere we could refer to? Might make sense to explain in some detail. Could people use other repo files for the same purpose, or is this special?

internal/rpmmd/repository.go
teg: We should probably treat the cases of no subscriptions gracefully and not return an error in that case (it is valid to have an unsubscribed system), and then only fail on real errors here.

We should, however, fail (with a helpful error code) in case someone tries to do something later on that requires subscriptions and we have none.

teg (commented):
Love this! Only very small nit picks/questions inline.

test/cases/regression-composer-works-behind-satellite.sh
msehnout: Sure, I hope osbuild gets a new release soon and I can remove this.

internal/distro/fedora32/distro.go
msehnout: Very unlikely, but it should be possible to setup private mirror (e.g. using Satellite or Pulp) with Fedora content and require subscription. Anyway it is useful for the test case I introduced.

internal/rpmmd/repository.go
msehnout: I implemented these error handling cases:
 1. The repo file is not there - print a message about it
 2. The repo file is there but it is not possible to open it or parse it - print a different message
 3. User attempts to use rhsm on a system with no subscriptions - suggest that the system must be subscribed

internal/rhsm/secrets.go
msehnout: It is not different from any other repo file. I wrote this fact into a comment.

teg: So the reason that it is special is that it is the file written by RHSM, right? I.e., it does not make sense for us to support the same logic for other repo files, as we are explicitly targeting RHSM?

teg (approved):
osbuild has been released, so this is ready to go in (after dropping the repo pinning from Schutzfile). Great work!

teg (dismissed):
This is great! Thanks for all the work you did on this!

test/cases/satellite.sh
msehnout: I've created a cleanup code at the end of the script. Do you think we should run it in a stand-alone machine anyway?

schutzbot/Jenkinsfile
msehnout: This is now outdated. The test case is executed from the regression.sh script which in turn is executed from Gitlab CI configuration.

test/cases/regression.sh
```diff
@@ -40,11 +42,26 @@ run_test_case () {
...
 /usr/libexec/osbuild-composer-test/provision.sh
 
-# Run each test case.
+# Run test cases common for all distros.
 for TEST_CASE in "${TEST_CASES[@]}"; do
     run_test_case ${TESTS_PATH}/"$TEST_CASE"
 done
 
+case "${ID}" in
+    "fedora")
```
atodorov, line 51: Can this test script be made more generic and also work on RHEL ?

atodorov (commented):
Added questions.

test/cases/regression.sh
msehnout: It should be possible, the problem is that the script creates a proxy for RPMrepo which works in a different way for RHEL and Fedora. Fedora repositories are hosted on AWS, but RHEL repos are hosted on PSI. I'm not sure we can make this change in time for the Z stream update.

atodorov (dismissed):

internal/rpmmd/repository.go
```diff
@@ -393,20 +371,28 @@ func runDNF(dnfJsonPath string, command string, arguments interface{}, result in
...
-	CacheDir    string
-	RHSM        *RHSMSecrets
-	dnfJsonPath string
+	CacheDir      string
+	subscriptions *rhsm.Subscriptions
+	dnfJsonPath   string
 }
 
 func NewRPMMD(cacheDir, dnfJsonPath string) RPMMD {
+	subscptions, err := rhsm.LoadSystemSubscriptions()
```
ondrejbudai, line 380: typo

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -24,22 +24,104 @@ function generate_certificates {
...
+    "fedora")
+        echo "Running on Fedora"
+        DISTRO_NAME="${ID}-${VERSION_ID}"
+        REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/${ID}-${VERSION_ID}.json"
+        REPO1_NAME="fedora"
+        REPO2_NAME="updates"
+        ;;
+    "rhel")
+        echo "Running on RHEL"
+        case "${VERSION_ID}" in
```
ondrejbudai, line 66: I suggest comparing only the major version against `${VERSION_ID%.*}`.

internal/rpmmd/repository.go
msehnout: Fixed, thanks 👍

test/cases/regression-composer-works-behind-satellite.sh
msehnout: Fixed, thanks 👍

teg (dismissed):
Thanks so much for getting this over the finishing line, this will be very much appreciated :)

ondrejbudai (dismissed):

ondrejbudai:
I rebased this PR to fix the conflict in `go.sum`.

teg (approved):

ondrejbudai (approved):

PR #1409, opened 2021-05-13
WIP: Testing rhel nightlies in gitlab ci

jrusz:
This pull request includes:


tools/define-compose-url.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
-curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.4/COMPOSE_ID > COMPOSE_ID
+# Small upgrade, this will get the latest nightly COMPOSE_ID with the specified VERSION
+curl -L "https://cts.engineering.redhat.com/api/1/composes/?release_version=${VERSION}&tag=nightly" | jq -r .items[0].compose_info.payload.compose.id > COMPOSE_ID
```
atodorov, line 5: Where does the `VERSION` variable come from? Seems like it is undefined to me?

atodorov: FTR this will return `null` once the nightly is removed (e.g. old enough). Maybe we should error out explicitly if such condition is met.   Which isn't much different than relying on the symlink URL in the old version and handling the errors gracefully. 

What's the purpose of this change? How does that help us solve our problems with missing symlink or missing nightly trees  ?

atodorov (commented):
Added several questions.

tools/define-compose-url.sh
jrusz: The `VERSION` variable comes from the scheduled pipeline settings, you can set additional env vars just for that run. The idea is to add stages that would only be ran as a scheduled nightly test. So when the nightlies will stop existing, we can just bump the variable version in gitlab.

This is still a WIP, I don't have everything figured out yet. I'm mostly trying if this approach will work for us and trying out some things along the way. This should hopefully point to the latest working nightly, but when the time comes where they're going to be old enough to not exist anymore and they also stop building them I don't yet know how to automatically figure that out...

jrusz: > What's the purpose of this change? How does that help us solve our problems with missing symlink or missing nightly trees ?

I guess that when the time when nightly builds are no longer available we'll have to manually bump the `VERSION` and probably also change the aws runners. So what this could do for us is this:

1. We'll always use latest finished and accepted nightly compose
2. Pointing to a newer minor build is as easy as changing the `VERSION` in gitlab pipeline schedules

Currently I can't think of a better way. In case this stops working because of missing nightly trees/symlink it means it's time to drop the current nightly testing and bump the version. Also it won't return `null` but the latest tagged nightly `COMPOSE_ID`. Of course that means that if the build is too old, the `COMPOSE_URL` won't exist anymore...

If you think that this won't help us I'll try to come up with something different..

atodorov: Few comments FTR:

> The VERSION variable comes from the scheduled pipeline settings, you can set additional env vars just for that run. > The idea is to add stages that would only be ran as a scheduled nightly test. 

That kind of defeats the idea of configuration as code!

> So when the nightlies will stop existing, we can just bump the variable version in gitlab.

We can do the same in git and have traceability of the changes so it doesn't really solve the problem just shifts the configuration from one place to another.

> I guess that when the time when nightly builds are no longer available we'll have to manually bump the VERSION and probably also change the aws runners. So what this could do for us is this:
>
>    We'll always use latest finished and accepted nightly compose
>    Pointing to a newer minor build is as easy as changing the VERSION in gitlab pipeline schedules
>
> Currently I can't think of a better way. In case this stops working because of missing nightly trees/symlink it means it's time to drop the current nightly testing and bump the version. Also it won't return null but the latest tagged nightly COMPOSE_ID. Of course that means that if the build is too old, the COMPOSE_URL won't exist anymore...

I don't think this is any different from what we have ATM. It just shifts the definitions/variables into a different place. 

With this PR let's focus on making the nightly pipeline run more or less in the same way it does now and solve the remaining problems separately.

jrusz: Yeah, you're right, I see it now.
I've put in some default values but I'd like to keep the option of specifying the `VERSION` and `TAG` from outside, I believe it can be useful if we want to run the tests on some other compose. 

Also one thing that's an improvement over using the latest symlink is that we're really using the latest compose that's finished and passed acceptance testing. I could see a newer compose in the 8.5 latest folder but that was actually unusable. I'm going to open a ticket for that so it won't happen again but using this internal CTS service sounds safer to me.

I'm going to try running the whole test suite using RHEL 8.4 runner, if that passes I think we should be in the same state as with Jenkins. Scheduled pipeline will be our cron trigger and the `prepare-rhel-internal` part will only run during the scheduled pipeline run. 

Also I think we should get 8.5 runner in aws, I'm going to work that out with Ondrej.

schutzbot/prepare-rhel-internal.sh
```diff
@@ -8,7 +8,7 @@ function greenprint {
 
 ARCH=$(uname -m)
 
-source ../tools/define-compose-url.sh
```
jrusz, line 11: Did this work for you in Jenkins @atodorov ? I had to change for Gitlab CI, so in case this was correct I need to add some if condition.

.gitlab-ci.yml
```diff
@@ -26,61 +27,77 @@ init:
 RPM:
   stage: rpmbuild
   extends: .terraform
+  only:
+    - schedules
```
atodorov, line 31: That should be the opposite, only on PR/branch push, not on schedules.

schutzbot/prepare-rhel-internal.sh
atodorov: It works in Jenkins b/c we execute from inside the `schutzbot` directory. Guess pwd is different for GitLab.

.gitlab-ci.yml
jrusz: Yes, I was just trying something out. It's possible to also use `except` which in this case sounds better.

schutzbot/prepare-rhel-internal.sh
jrusz: Alright, I solved this in the first way I could come up with just to keep Jenkins compatibility until we're certain we can use Gitlab CI for nightly testing. I guess I'll add a comment with this information too.

atodorov:
Question: is it going to be possible to manually start (restart) scheduled jobs but allow for editing of `COMPOSE_URL` like documented here https://github.com/osbuild/osbuild-composer/blob/main/test/README.md#cron-jobs-for-internal-builds-testing ?

This is also a vital part of the nightly testing pipeline.

jrusz:
> Question: is it going to be possible to manually start (restart) scheduled jobs but allow for editing of `COMPOSE_URL` like documented here https://github.com/osbuild/osbuild-composer/blob/main/test/README.md#cron-jobs-for-internal-builds-testing ?
> 
> This is also a vital part of the nightly testing pipeline.

Yes. If you go in Gitlab into CI/CD -> Schedules you can see one schedule that I created for testing. It's possible specify `COMPOSE_URL` there manually by editing the schedule and then running it. It might be better though to just create another "dummy" schedule which won't be triggered automatically and will serve this purpose.

atodorov:
> > Question: is it going to be possible to manually start (restart) scheduled jobs but allow for editing of `COMPOSE_URL` like documented here https://github.com/osbuild/osbuild-composer/blob/main/test/README.md#cron-jobs-for-internal-builds-testing ?
> > This is also a vital part of the nightly testing pipeline.
> 
> Yes. If you go in Gitlab into CI/CD -> Schedules you can see one schedule that I created for testing. It's possible specify `COMPOSE_URL` there manually by editing the schedule and then running it.

I see it now. That will work. README & images will need an update though.

How do I make it execute immediately though? Just set the interval to now+5 mins ?

> It might be better though to just create another "dummy" schedule which won't be triggered automatically and will serve this purpose.

It isn't better IMO. It's actually worse because these 2 schedules will start diverging over time. However I see that there's no "ephemeral edit" functionality like in Jenkins so it will probably better to change the instructions. Something like: 
"If you want to redefine COMPOSE_URL/schedule a new nightly/rel-eng pipeline by hand then go to this page, define a new schedule similar to XXX, run your jobs then remove your newly related schedule."

jrusz:
> > > Question: is it going to be possible to manually start (restart) scheduled jobs but allow for editing of `COMPOSE_URL` like documented here https://github.com/osbuild/osbuild-composer/blob/main/test/README.md#cron-jobs-for-internal-builds-testing ?
> > > This is also a vital part of the nightly testing pipeline.
> > 
> > 
> > Yes. If you go in Gitlab into CI/CD -> Schedules you can see one schedule that I created for testing. It's possible specify `COMPOSE_URL` there manually by editing the schedule and then running it.
> 
> I see it now. That will work. README & images will need an update though.

Good point, I'll update the README when I get this working.

> How do I make it execute immediately though? Just set the interval to now+5 mins ?

There is a play button right next to the edit button. I currently have it disabled and run it manually whenever I push new updates to test them. 
> 
> > It might be better though to just create another "dummy" schedule which won't be triggered automatically and will serve this purpose.
> 
> It isn't better IMO. It's actually worse because these 2 schedules will start diverging over time. However I see that there's no "ephemeral edit" functionality like in Jenkins so it will probably better to change the instructions. Something like:
> "If you want to redefine COMPOSE_URL/schedule a new nightly/rel-eng pipeline by hand then go to this page, define a new schedule similar to XXX, run your jobs then remove your newly related schedule."

Yeah that should work, I'll include something like then in the README.

.gitlab-ci.yml
```diff
@@ -88,9 +116,27 @@ Base:
...
+Base-nightly:
+  stage: test
+  extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule"'
+  variables:
+    EXTRA_REPO_PATH_SEGMENT: "gitlab/"
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/base_tests.sh
```
atodorov, line 128: I don't like the fact that the scripts section is already different from what is executed in `Base`. IMO extend from Base and redefine only rules and runner sections, similar to how `New OSTree:` does it.

.gitlab-ci.yml
```diff
@@ -151,6 +199,70 @@ Integration:
...
+    EXTRA_REPO_PATH_SEGMENT: "gitlab/"
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/api.sh
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/rhel-8.4-x86_64
+        INTERNAL_NETWORK: ["true"]
+
```
atodorov, line 265: hmmm, again, not liking the duplication and the possibility of testing to diverge over time and the necessity to maintain both job definitions. 

Is it possible to specify conditions per runner ? If that's possible we may say something like: 
- don't run Fedora/CentOS in schedules
- skip libvirt tests in schedules b/c of bugs, etc.
-

tools/define-compose-url.sh
atodorov: > I've put in some default values but I'd like to keep the option of specifying the VERSION and TAG from outside, I believe it can be useful if we want to run the tests on some other compose.

Again, how is `VERSION` & `TAG` any different from specifying the `COMPOSE_URL` directly?  It does the same thing but adds a few more steps in the process. If you want to test something different then you can specify the URL to SUT. 

IMO none of these changes are needed or beneficial at the moment.

test/README.md
```diff
@@ -365,3 +365,19 @@ then do the following:
...
+
+We're using internal CTS instace for detecting latest accepted nightly build. It's possible to
+manually specify which `VERSION` and `TAG` you want to find or just the whole `COMPOSE_URL` directly.
+These variables can all be changed in the pipeline schedule settings. It is advised to create
+a new schedule if you want to try something or change these values, set it as inactive and then
+just trigger it manually.
+
+   !['Create pipeline schedule'](./gitlab_ci_scheduled_pipeline.png)
+
+   !['Trigger scheduled pipeline'](./gitlab_ci_schedule_trigger.png)
```
atodorov, line 383: IMO the relevant sections for Jenkins should be removed and/or the wording & images updated to document the flow for GitLab instead. We don't want to have both of them anyway and Jenkins trigger has been removed as of today.

atodorov (changes requested):

.gitlab-ci.yml
```diff
@@ -88,68 +116,152 @@ Base:
...
-          - openstack/fedora-33-x86_64
-          # See COMPOSER-919
-          # - openstack/fedora-34-x86_64
-          - openstack/rhel-8-x86_64
-          - openstack/rhel-8.4-x86_64
-          # do we want centos as well ?
-          # do we want secondary architectures here ??
+  extends: Base
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER == "aws/rhel-8.4-x86_64"'
```
atodorov, line 123: Option 1) use a "contains" operator to match everything that contains "rhel" instead of specifying explicitly
Option 2) Redefine the runners explicitly and leave only rhel ones here.

jrusz:
Successful nightly pipeline run: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/321898047 

@atodorov could you take another look at this please? 

I've added the option to specify custom `REPO_URL` in the scheduled pipeline settings for cases where we would want to test scratch builds with the latest nightlies in the future. What this does is it installs all osbuild-composer packages though, not only test packages, meaning it won't use any osbuild-composer package from the nightly repo, but I guess it makes sense to install them all in the same version.

I've also added the possibility to either use osbuild from the nightly repository or the version specified in `Schutzfile` which in this case is needed because the version in the nightly repositories is too old. I'm not sure if my implementation is acceptable though..

Lastly I've made the `libvirt.sh` test run on Openstack because it seems there's the same issue on RHEL 8.4 as this one #1454 . It's probably not the nicest way to do this but I couldn't figure out a better way. It will still show `aws` in the job name but the runner will actually provision a machine in Openstack because I changed the `RUNNER` variable.

I hope the rest of the changes make sense.

.gitlab-ci.yml
jrusz: I've extended from both Integration and Base stages and that seems to work well.

test/README.md
jrusz: I've removed Jenkins part and left only Gitlab CI.

tools/define-compose-url.sh
jrusz: Right, I've removed both `VERSION` and `TAG` it's simpler to just provide the whole `COMPOSE_URL`.

.gitlab-ci.yml
```diff
@@ -165,6 +203,16 @@ Integration:
...
 
+Integration-nightly:
+  stage: test
+  extends: Integration
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER == "aws/rhel-8.4-x86_64" && $SCRIPT != "libvirt.sh"'
+    # Run libvirt test on openstack because it has kvm support
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER == "aws/rhel-8.4-x86_64" && $SCRIPT == "libvirt.sh"'
+      variables:
+        RUNNER: "openstack/rhel-8.4-x86_64"
```
atodorov, line 214: Looking at #1459 this seems like it needs to be an unconditional change in the Integration stage. I find it very weird if you are seeing bugs when running the nightly pipeline but not when running the regular one.

.gitlab-ci.yml
atodorov: ping ^^^

atodorov (changes requested):
Some changes requested

.gitlab-ci.yml
```diff
@@ -165,6 +203,16 @@ Integration:
...
+
 finish:
```
atodorov, line 216: Missing updates for the new API stage

.gitlab-ci.yml
jrusz: Alright so I added a regexp to match any RHEL-8.X-x86_64 runner... I hope that's what we want to achieve here. If we update the runners to 8.5 or switch back to openstack it should still work.

tools/define-compose-url.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 set -euo pipefail
 
-curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.4/COMPOSE_ID > COMPOSE_ID
+# jq is needed to get COMPOSE_ID
+sudo dnf -y install jq
+
+# This will get the latest finished compose that was tagged accordingly
+curl -kL "https://cts.engineering.redhat.com/api/1/composes/?release_version=8.5.0&tag=nightly" | jq -r .items[0].compose_info.payload.compose.id > COMPOSE_ID
```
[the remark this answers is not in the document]
jrusz: The reason I wanted to introduce this is because it should always point to the latest RTT_ACCEPTED nightly. It can happen that the latest 8.5 folder in internal download servers can contain an unfinished compose or otherwise broken or I encountered it in the past at least... But like you're pointing out it's probably not that reliable as the RHEL-8.5.0-20210622.n.0 was also RTT_ACCEPTED... 

I agree to just drop it and point to latest 8.5... Also I noticed a latest finished 8.5 symlink exists, might be a bit better.

IMO it would be best if we could just trigger on some UMB message that a new nightly suitable for testing has been created but I guess that can be solved later.

test/README.md
```diff
@@ -327,41 +327,18 @@ is is missing lots of details!
...
-
-3. In Jenkins Classic UI, left-hand sidebar click ***Replay*** button. This will allow you to
-   replay a Pipeline with a modified syntax
-4. Locate the  `detect_build_cause()` function near the bottom and modify it so that it
-   will `return "cron"`. This is required because when Pipelines are restarted manually
-   their build cause is **Replayed #xy**. See the images for reference:
-
-   !['Modify build cause'](./pipeline_replay_01.png)
-
-   !['Modify build cause'](./pipeline_replay_02.png)
```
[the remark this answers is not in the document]
jrusz: Thanks, should be removed now.

test/README.md
```diff
@@ -327,41 +327,18 @@ is is missing lots of details!
...
 
-By default we test against latest nightly builds. If you wish to test against other
-flavors (e.g. rel-eng) specify the `COMPOSE_URL` environment variable to point to
-the respective URL (stopping before the `/compose/` path). See the image below or
-`schutzbot/prepare-rhel-internal.sh` for more details.
+We're using internal CTS instace for detecting latest accepted nightly build. It's possible to
+manually overrida which `COMPOSE_URL` to use instead.
+These variables can all be changed in the pipeline schedule settings. It is advised to create
+a new schedule if you want to try something or change these values, set it as inactive and then
+just trigger it manually.
```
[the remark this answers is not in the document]
jrusz: Updated.

.gitlab-ci.yml
jrusz: Thanks for pointing this out. I've put some more time into investing why is this happening and it seems there's a bug in qemu-kvm because this issue appears after a newer version is used. I created rhbz#1975242 for it. I'm not sure if we should just switch to using Openstack until then or wait for some resolution :/ ... 

I guess I could also try to introduce a similar change in the regular pipeline so that only the libvirt test is ran on Openstack and do that for Centos stream as well as currently all of the Centos stream tests a re running in Openstack because of this bug.

.gitlab-ci.yml
jrusz: Added. I noticed a problem though, whenever someone adds a new test which will not be a part of tests package in the latest nightly, these tests will just fail. Current example is `api.sh aws.s3`

atodorov: That is a legitimate concern however it is off topic on the current PR. Open a ticket so we can deal with it appropriately.

tools/define-compose-url.sh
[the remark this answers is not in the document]
atodorov: +1 for the `-finished` symlink. I didn't know about it.

.gitlab-ci.yml
atodorov: Ahh, looks like one of the rarest times we were saved because the NVRs we use were too old. What is called the `rhel-8` runner is still in fact 8.3. Will probably have to switch at the moment when we update the runners. Not a concern for now.

.gitlab-ci.yml
```diff
@@ -93,9 +121,17 @@ Base:
       - "*.repo"
     when: always
 
+Base-nightly:
+  stage: test
+  extends: Base
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+\/rhel-8\..-x86_64/'
```
atodorov, line 128: Will fail when secondary arches are added.
Will fail when we add rhel-9 runners.

Why not use only "rhel" and possibly have a simpler regex?

atodorov (changes requested):
One last change requested

.gitlab-ci.yml
```diff
@@ -198,6 +243,12 @@ API:
         INTERNAL_NETWORK: ["true"]
         DISTRO_CODE: ["rhel_90"]
 
+API-nightly:
+  stage: test
+  extends: API
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+\/rhel-8\..-x86_64/'
+
```
atodorov, line 251: Grrr, found out this is missing the definitions for OStree jobs. 

Isn't there any better way to selectively execute a subset of the test matrix ? 

What would happen if you remove the section overrides and instead add the following rule to all sections:
```
- if: '$CI_PIPELINE_SOURCE != "schedule" || ($CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+\/rhel-8\..-x86_64/)'
```

**Note:** syntax may not be correct. What I want to say is "execute if not started by schedule or (started by schedule and the runner is RHEL)".

.gitlab-ci.yml
jrusz: > Will fail when secondary arches are added.

True but I thought that for nightly testing we're only concerned with x86_64. No problem to add it though.

> Will fail when we add rhel-9 runners.
>
> Why not use only "rhel" and possibly have a simpler regex?

If I just add all RHEL runners I'm afraid that there will be a duplication of test runs. What this whole PR currently achieves is that it runs the tests on a 8.4 runner with 8.5 nightly repos. If I also include regular rhel-8 runner, that will just mean it will also run on 8.3 with 8.5 nightly repos, unless I'm missing something. So that's why I made the regexp to match only 8.X which should be the current nightly. 

About rhel-9 runners, sure I can add them now, didn't think about them ahead..

.gitlab-ci.yml
jrusz: That's a pretty good idea, I'll try it. It should be possible to define several rules per and execute the jobs under different conditions. This means that we also want to run OStree tests for nightly right?

jrusz: Yup, works as expected.
Regular pipeline: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/326103978
Scheduled pipeline: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/326104181

Unless you can spot some other potential problem I'll squash the latest commit. What I can see is that it won't have 'nightly' in the name but we can deduce that from the fact that it's going to be a scheduled pipeline and also from the stages as it won't have rpbuild. And then I'm not sure if it's immediately obvious from the code what it's supposed to do but I guess that's just gitlab-ci.yml syntax... I can add a comment though

atodorov:
IMO the current code looks good. 

I've restarted the regular pipeline, b/c we need to see that pass before safely merging. However I see quite a few failures on the scheduled pipeline, not sure what to do about them.

jrusz:
> IMO the current code looks good.
> 
> I've restarted the regular pipeline, b/c we need to see that pass before safely merging. However I see quite a few failures on the scheduled pipeline, not sure what to do about them.

Thanks, I'll have to squash my last commit so that will trigger the regular pipeline again, just wanted to hear your opinion before that...
I'm not sure about the failures either, especially the Ostree ones... Libvirt failure is caused by a qemu bug most likely (rhbz#1975242) api aws.s3 and gen2 Azure is caused by the tests missing in the installed rpm, we decided to solve it later.

atodorov (approved):
LGTM, regular pipeline execution at https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/326820204

All jobs except openstack passing. I am going to merge this.

PR #1412, opened 2021-05-17
distribution: Add state directory

Gundersanne:
This should become a PersistentVolumeClaim in future, but let's get the
deployment running first.


teg (approved):
Makes sense. Thanks!

PR #1425, opened 2021-05-26
cloudapi: Add x-rh-identity header filter

Gundersanne:


cmd/osbuild-composer/config.go
```diff
@@ -15,6 +15,9 @@ type ComposerConfigFile struct {
 		AllowedDomains []string `toml:"allowed_domains"`
 		CA             string   `toml:"ca"`
 	} `toml:"worker"`
+	CloudAPI struct {
```
teg, line 18: I wonder, I think we never exposed this name, should we perhaps rename this to `ComposerAPI` or something like that (and similarly for the actual code)? "Cloud" sounds... wrong. @ondrejbudai @larskarlitski?

teg (changes requested):
This look very nice over all. Just please add a commit message :) Left one question inline.

cmd/osbuild-composer/config.go
larskarlitski: Yeah, I really like `ComposerAPI` as well. Thanks!

ondrejbudai: Agreed

Gundersanne:
@teg So this does replace all ssl auth on the integration tests with the identityfilter. Should I duplicate the tests instead? Or maybe just add an extra step for the rhel-8 and centos-8 ones?

teg:
> @teg So this does replace all ssl auth on the integration tests with the identityfilter. Should I duplicate the tests instead? Or maybe just add an extra step for the rhel-8 and centos-8 ones?

I think we do have to test both, but if you can just extend the existing things to avoid duplicating everything that's fine by me. I think only verifying that one call works is sufficient, it should all be the same, right?

teg:
/call/endpoint/

Gundersanne:
@teg only centos-8 is failing, probably due to https://bugzilla.redhat.com/show_bug.cgi?id=1961562

Anyway it's unrelated so if it looks good I'd like to merge this

teg (approved):
This is great. Yeah, I would too be in favour of merging this as is as the test failure does indeed seem clearly unrelated.

PR #1426, opened 2021-05-26
distro/fedora: run on unsupported version

msehnout:
Run on unsupported versions of Fedora. This is mainly useful for running osbuild-composer on Fedora 34 and 35 which is still impossible.

I can build and boot qcow2 image with this patch.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -75,10 +75,17 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *
 
 func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener) error {
 	archName := common.CurrentArch()
+	fallback := false
+	var fallback_name string
+	var fallback_repos map[string][]rpmmd.RepoConfig
 
 	hostDistro := c.distros.FromHost()
 	if hostDistro == nil {
-		return fmt.Errorf("host distro is not supported")
+		hostDistro, fallback_name, fallback_repos = c.distros.FallbackHost()
+		fallback = true
+		if hostDistro == nil {
+			return fmt.Errorf("host distro is not supported")
+		}
 	}
 
 	arch, err := hostDistro.GetArch(archName)
@@ -89,6 +96,9 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener) err
 	rr, err := reporegistry.New(repoPaths)
 	if err != nil {
 		return fmt.Errorf("error loading repository definitions: %v", err)
+	}
+	if fallback {
+		rr.AddRepos(fallback_name, fallback_repos)
 	}
 
 	// Check if repositories for the host distro and arch were loaded
--- a/internal/distro/fedora33/distro.go
+++ b/internal/distro/fedora33/distro.go
@@ -21,9 +21,11 @@ const modulePlatformID = "platform:f33"
 const ostreeRef = "fedora/33/%s/iot"
 
 type distribution struct {
-	name          string
-	arches        map[string]architecture
-	buildPackages []string
+	name             string
+	modulePlatformID string
+	ostreeRef        string
+	arches           map[string]architecture
+	buildPackages    []string
 }
 
 type architecture struct {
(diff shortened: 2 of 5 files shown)


bcl:
See #1427

bcl:
ah crap. Sorry.

msehnout:
It builds an image (f34) and the image is bootable. What do you think about the approach? @thozza , @ondrejbudai, @teg, and anyone else interested? It is not a super nice patch, but it does work and I expect something similar for the support of RHEL clones (but I'm not sure how to generate the repos there).

Would you like to see something like this merged into composer?
Should I implement support for Alma linux and Rocky linux as part of this PR?

cmd/osbuild-composer/composer.go
```diff
@@ -75,10 +75,17 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *
...
+	var fallback_repos map[string][]rpmmd.RepoConfig
 
 	hostDistro := c.distros.FromHost()
 	if hostDistro == nil {
-		return fmt.Errorf("host distro is not supported")
+		hostDistro, fallback_name, fallback_repos = c.distros.FallbackHost()
+		fallback = true
+		if hostDistro == nil {
+			return fmt.Errorf("host distro is not supported")
+		}
```
thozza, line 88: Personally I would prefer to move this logic to the `FromHost()` method and have it as the only exported method. The `FallbackHost()` could be called inside of the `FromHost()`. So in the end if the `FromHost()` returns a `distro.Distro`, it is either the supported or unsupported/fallback distro or `nil` and the distro is not supported at all.

internal/distro/fedora33/distro.go
```diff
@@ -586,14 +588,18 @@ func ostreeCommitAssembler(options distro.ImageOptions, arch distro.Arch) *osbui
...
+	return newDistro(defaultName, modulePlatformID, ostreeRef)
 }
 
 func NewHostDistro(name string) distro.Distro {
-	return newDistro(name)
+	return newDistro(name, modulePlatformID, ostreeRef)
+}
+
+func NewUnreleasedDistro(name, distroVersion string) distro.Distro {
+	return newDistro(name, "platform:f"+distroVersion, "fedora/"+distroVersion+"/%s/iot")
```
thozza, line 599: I would suggest to rather change the `NewHostDistro()` definition in all distros to accept the `name` and `distroVersion` than an additional Method. This way we could easily decide which Fedora / RHEL distro definition will be used as a fallback in Fedora and RHEL and construct it in the same way whether it is unsupported or not.

internal/distroregistry/distroregistry.go
```diff
@@ -126,3 +129,23 @@ func mangleHostDistroName(name string, isBeta, isStream bool) string {
...
+	}
+	distroVersion := parts[1]
+
+	if strings.HasPrefix(hostDistroName, "fedora-") {
+		log.Println("Warning: " + hostDistroName + " is not supported Fedora version. " +
+			"osbuild-composer will start with Fedora 33 image definitions but this might not work!")
+		r.hostDistro = fedora33.NewUnreleasedDistro(hostDistroName, distroVersion)
+	}
+	return r.hostDistro, hostDistroName, unsupported.GenerateFedoraRepositories(distroVersion)
+}
```
thozza, line 151: I would suggest to move this logic into `FromHost()` and not to export this as an additional method to prevent the need to implement additional logic by the caller to actually get the Host `distro.Distro`.

cmd/osbuild-composer/composer.go
```diff
@@ -90,6 +97,9 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener) err
 	if err != nil {
 		return fmt.Errorf("error loading repository definitions: %v", err)
 	}
+	if fallback {
+		rr.AddRepos(fallback_name, fallback_repos)
+	}
```
thozza, line 102: I'm wondering whether these fallback repositories could be added already when constructing the `RepoRegistry`, to not to make a special case out of them. But I don't have a specific idea how to do it... :smile:

thozza (commented):
Thanks for adding the support to build on "unsupported" Fedora versions! These are very interesting changes, I added a few comments / suggestions...

thozza:
> It builds an image (f34) and the image is bootable. What do you think about the approach? @thozza , @ondrejbudai, @teg, and anyone else interested? It is not a super nice patch, but it does work and I expect something similar for the support of RHEL clones (but I'm not sure how to generate the repos there).

So in case of RHEL, the repositories won't change with a minor release, could only if the release is EUS and EUS repositories should be used. But in case of osbuild-composer use cases, we should just use the repo definition that is used for "supported" distros.

I was also thinking whether it would make sense to have one Fedora and RHEL distro defined as "to be used for unsupported" distro in the distroregistry and add its instance to the list of supported distros, so that `FromHost()` can return it? These would be probably the latest Fedora / RHEL distro definitions and they would be constructed for the unsupported use case in the same way as they are for the "host distro" use case - by calling `NewHostDistro()`...

> 
> Would you like to see something like this merged into composer?

Something like this yes, but we should definitely have a discuss before that. :smile: 

> Should I implement support for Alma linux and Rocky linux as part of this PR?

I don't see any need to do so as part of this PR, but it is up to you...

ondrejbudai:
We discussed this and the result is:

-  We want the Weldr API to work even on Fedora versions that don't have explicit support by `osbuild-composer` (including rawhide).
- Creating repo URLs out of the thin air is a bit wonky. Instead, we will just ship repositories in a json format as we currently do but for all Fedora versions.
- If a repository URL is defined but its corresponding Fedora version isn't explicitly supported by composer, the newest supported Fedora definition is used. Currently, the newest supported Fedora is 33, therefore the Weldr API on Fedora 34 would build images from Fedora 34 repos using Fedora 33 repositories.
- It's highly recommended to add Fedora X repositories to osbuild-composer as soon as Fedora X-1 is branched out of it.

When this is implemented, we will revisit the topic of supporting RHEL rebuilds.

msehnout:
Replaced with #1438

PR #1427, opened 2021-05-26
Building for other distributions

bcl:
This pull request includes:

This replaces draft #1381


bcl:
Found some problems with `modules/info` including the host distro builds along with the selected distro so this still needs some work.

bcl:
Ok, so the host distro package 'bug' was actually me leaving a source repo active that was connected to the f33 mirror so chalk that up to user error.

I have now also successfully built a RHEL 8.3 qcow2 using a Fedora 33 VM with a subscription (after turning it off and on again so that osbuild-composer would load the certs) -- so everything seems to be working.

internal/weldr/api.go
```diff
@@ -1928,8 +1928,13 @@ func ostreeResolveRef(location, ref string) (string, error) {
 	return parent, nil
 }
 
-func (api *API) depsolveBlueprintForImageType(bp *blueprint.Blueprint, imageType distro.ImageType) (map[string][]rpmmd.PackageSpec, error) {
-	packageSets := imageType.PackageSets(*bp)
+func (api *API) depsolveBlueprintForImageType(bp blueprint.Blueprint, imageType distro.ImageType) (map[string][]rpmmd.PackageSpec, error) {
+	// Depsolve using the host distro if none has been specified
+	if len(bp.Distro) == 0 {
```
ondrejbudai, line 2086: I think I usually expect `len()` when working with slices. With strings, I would expect `bp.Distro == ""`, I think it's more idiomatic.

internal/weldr/api.go
```diff
@@ -2902,12 +2907,17 @@ func (api *API) allRepositories() ([]rpmmd.RepoConfig, error) {
...
 
-func (api *API) depsolveBlueprint(bp *blueprint.Blueprint) ([]rpmmd.PackageSpec, error) {
+func (api *API) depsolveBlueprint(bp blueprint.Blueprint) ([]rpmmd.PackageSpec, error) {
 	repos, err := api.allRepositories()
 	if err != nil {
 		return nil, err
 	}
 
+	// Depsolve using the host distro if none has been specified
+	if len(bp.Distro) == 0 {
```
ondrejbudai, line 3123: Same as above.

internal/common/helpers.go
```diff
@@ -23,3 +26,13 @@ func PanicOnError(err error) {
 		panic(err)
 	}
 }
+
+// IsStringInSlice returns true if the string is present, false if not
+// slice must be sorted
+func IsStringInSlice(slice []string, s string) bool {
```
ondrejbudai, line 32: oh, this actually always required a sorted slice. While doing refactorings, can you rename the method to `IsStringInSortedSlice` so it's obvious what it does? I think this is a big trap for someone who doesn't read comments.

internal/weldr/api.go
```diff
@@ -1940,22 +1955,31 @@ func ostreeResolveRef(location, ref string) (string, error) {
...
+// depsolveBlueprintForImageType handles depsolving the blueprint package list and
+// the packages required for the image type.
+// NOTE: The imageType *must* be from the same distribution as the blueprint.
 func (api *API) depsolveBlueprintForImageType(bp blueprint.Blueprint, imageType distro.ImageType) (map[string][]rpmmd.PackageSpec, error) {
 	// Depsolve using the host distro if none has been specified
 	if len(bp.Distro) == 0 {
 		bp.Distro = api.hostDistroName
 	}
 
+	if bp.Distro != imageType.Arch().Distro().Name() {
```
ondrejbudai, line 2090: I would honestly just panic here, this is a clear programming error.

docs/news/unreleased/other-distro-build.md
```diff
@@ -0,0 +1,145 @@
...
+                "enabled": true
+            }
+        ]
+    }
+
+
+## Unknown Distributions
+
+If an unknown distribution is selected the response from the API server will be
+a `DisroError`, like this:
```
ondrejbudai, line 134: `s/DisroError/DistroError`

ondrejbudai (commented):
Amazing work, I'm loving every bit of this PR, I'm looking forward to trying it myself.

I reviewed this very briefly tonight but it's getting a bit late and I'm tired so I will give it another shot later the next week

internal/weldr/api.go
bcl: I don't like panic unless it's absolutely necessary, this should never happen but if it does I don't see anything wrong with returning an error.

internal/common/helpers.go
bcl: Good idea!

bcl:
Rebased onto Release 30 to rerun all the tests.

docs/news/unreleased/other-distro-build.md
```diff
@@ -0,0 +1,145 @@
...
+sources, and the optional `?distro=` selection on API routes.
+
+eg. `curl --unix-socket /run/weldr/api.socket http://localhost/api/v1/distros/list`
+
+    {
+        "distros": [
+            "centos-8",
+            "fedora-32",
+            "fedora-33",
+            "rhel-8",
```
teg, line 21: Could we rename this `rhel-83`? The current naming is misleading (my bad), so would be a shame to expose it in an API that didn't yet expose it. We probably need to keep `rhel-8` as an alias in the koji/cloud APIs just in case it is being used still though.

internal/weldr/api.go
```diff
@@ -156,7 +177,8 @@ func New(repoPaths []string, stateDir string, rpm rpmmd.RPMMD, distros *distrore
 		logger:          logger,
 		compatOutputDir: compatOutputDir,
 		hostDistroName:  hostDistro.Name(),
-		distros:         distros,
+		distroRegistry:  dr,
+		distros:         validDistros(rr, dr, hostArch.Name(), logger),
```
teg, line 181: I like this approach! Allows us to easily select which distros to support where in the .spec file, by just shipping (or not) the repo files. (If I understood correctly).

teg (changes requested):
This is looking really great! I left one request about distro naming inline, and in addition to that I wonder if we should ship per-minor-version repository files? So far we had rhel-8.json, but that was due to ignorance on my side (didn't know we had repos for minor versions on the CDN), and for it to make sense to build selected minor versions I think we also need to point to the right minor-version repository. What do you think?

docs/news/unreleased/other-distro-build.md
bcl: If there is a valid distro named rhel-83 with a valid repository file it will, but right now osbuild-composer is aliasing those to rhel-8. The list of distributions is from DistroRegistry, with the ones without actual repo json files filtered out.
I also think that `rhel-8` may be useful as a "latest rhel-8 content" source since it points to the cdn url without a minor version.

In other words :) I think the code is fine, it's the repository data that needs to be updated to include repos for each minor release.

internal/weldr/api.go
bcl: Correct, also see my reply above.

bcl:
> What do you think?

Correct. Unless there is a distro file with a URL pointing to the minor release it doesn't know how to select it. So the ones we want to support should be added to the rhel8/distro code and repository json files.

docs/news/unreleased/other-distro-build.md
teg: I agree, the code is good as-is. However, as this is exposing these names for the first time through the weldr API, I wonder if this PR should also do the renaming so that it will be exposed "correctly"? We could add back a `rhel-8` if we want a generic one, but for now I think we should aim to just expose the distros we have with the most correct names.

teg:
Sorry for taking so long to get back to this. I would love to see this merged, I think the only outstanding thing is to change the repositories we ship for RHEL8 to point to the minor versions, so the right content will be used for each, rather than "the most recent". Do you agree we should include that in this PR too?

bcl:
> Do you agree we should include that in this PR too?

I'd rather see that as a separate PR. My thinking is that this PR adds the ability to select the distros, and that the actual distros available is a different problem to solve. And I'm anxious to be done with all these patches :)

teg:
> > Do you agree we should include that in this PR too?
> 
> I'd rather see that as a separate PR. My thinking is that this PR adds the ability to select the distros, and that the actual distros available is a different problem to solve. And I'm anxious to be done with all these patches :)
> 

ok, i think this is good to go then, let's get it landed and address the rest in follow ups

teg (approved):

bcl:
Feel free to merge when you are ready. I'm not sure how valid those Schutzbot failures are, the ones I've spot-checked seem to be unrelated.

ondrejbudai (changes requested):
All tests on RHEL 8.4 fail with the following message. I'm afraid we need to fix this before we merge this. :/

```
Error initializing weldr API: loaded repository definitions don't contain any for the host distro/arch: there are no repositories for distribution 'rhel-84' and architecture 'x86_64'
```

bcl:
> All tests on RHEL 8.4 fail with the following message.

I think I've fixed those by copying over the rhel-84.json and rhel-85.json to the override repo. There are still a few other failures, but I am totally lost as to how to debug them (and actually, I do not know where/how you found the error you pasted above).

ondrejbudai (approved):
Looks good! Several tests failed due to #1543 so I restarted them.

@bcl I know that the test suite can be very confusing if someone isn't working with it on a regular basis. May I ask what would help you so you can find these failures by yourself the next time? I'm more than happy to help you with it and make the process more streamlined. :-)

bcl:
> @bcl I know that the test suite can be very confusing if someone isn't working with it on a regular basis. May I ask what would help you so you can find these failures by yourself the next time? I'm more than happy to help you with it and make the process more streamlined. :-)

When they fail the first thing I look at is the end of the logs and if there's nothing obvious there I search for 'error' or 'failed' in the text. Failing that I skim through it looking for something looking like an error. So I guess a summary at the bottom would be ideal, or a short bit of guide text specific to the test that helps find failures.

PR #1428, opened 2021-05-27
tests: Update libvirt because of internal error: unknown feature amd-…

atodorov:
…sev-es

Related: rhbz#1961562

This pull request includes:

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -106,3 +106,7 @@ fi
 greenprint "Installing test packages for ${PROJECT}"
 # Note: installing only -tests to catch missing dependencies
 retry sudo dnf -y install "${PROJECT}-tests"
+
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Does this change actually solve any issue in our CI? We should already have the latest libvirt version installed...

atodorov:
> Does this change actually solve any issue in our CI? We should already have the latest libvirt version installed...

Indeed it doesn't. We've got the latest libvirt and there doesn't seem to be any updates ATM.  At this time I'm not sure if we want to take this patch in or just live with the fact that some jobs will be failing and we'll have to restart them until we move to newer OS releases in the future. 

Feel free to close if you don't want to take this in.

ondrejbudai:
Yeah, I don't think this brings in additional value. The latest libvirt should be already installed by `dnf install` several lines above the addition unless it's already preinstalled in the image which it isn't.

PR #1429, opened 2021-05-27
distribution: Add PVC to clouddot template

Gundersanne:
This will make the job queue persistent over redeployments.

---

Blocked on PersistentVolumeClaim resource usage permission


Gundersanne:
Got approved to use PersistentVolumeClaims in clouddot, removing blocked label.

teg (approved):
Nice!

PR #1431, opened 2021-05-27
Extend information gathered by the `image-info` tool

thozza:
This is a preparation for defining EC2 images for RHEL-8.5.0. These extensions to image-info tool represent modifications done to the official EC2 images currently produced as RHEL release. It is important to be able to analyse these aspects of images, before we define them in osbuild-composer, to ensure that the resulting images will be consistent with the current state.

- Read non-empty lines from /etc/hosts file and add them to the report.
- Read content of /etc/machine-id and add it to the report.
- Read uncommented key/values from /etc/systemd/logind.conf and add them to the report.
- Read all ifcfg-* files from /etc/sysconfig/network-scripts/ and add their values to the report.
- Read content of /etc/locale.conf and add it to the report.
- Read SELinux configuration from /etc/selinux/config and add it to the report.
- Inspect the filesystem tree for SELinux context mismatches and add them to the report.
- Read configuration files from /etc/modprobe.d/ and for now report only all blacklisted kernel modules.
- Read RHSM configuration from /etc/rhsm/rhsm.conf and add it to the report.
- Read cloud-init configuration from /etc/cloud/cloud.conf and add it to the report.
- Read all *.conf files from /etc/dracut.conf.d/ and add their content to the report.
- Read VC and X11 keyboard configuration and add it to the report.
- Read specific configuration directives from Chrony configuration and add them to the report. Specifically 'server', 'pool', 'peer' and 'leapsectz'.
- Read drop-in configurations for *.service unit files from /etc/systemd/system/ and add them to the report.
- Read all configuration files from /etc/tmpfiles.d/ and add them to the report.
- Read all configuration files from /etc/sysctl.d/ and add them to the report.
- Read the Tuned active profile and profile mode and add them to the report.
- Read all configuration files from /etc/security/limits.d and add them to the report.
- Read sudoers configuration from /etc/sudoers and files in /etc/sudoers.d/ and add uncommented lines to the report. No
  sophisticated parsing is done, because the configuration format grammar is too complicated for the purpose of image-info.
- Read udev rules configuration files from /etc/udev/rules.d/ and add them to the report.
- Read DNF configuration and defined vars and add them to the report.
- Read profile ID and enabled features used by authselect.
- Enable SELinux, extended attributes and POSIX ACLs support when unpacking 'tar' image type to prevent potential mismatches

Regenerate all image test cases to reflect changes in the image-info output. Modify the distro-arch-imagetype-map.json to cover all combinations currently covered by existing image test cases.

Add doc strings to all read_* functions.

Fix one occurrence of read_partition() call, which have not been modified as part of previous commit [1].

[1] 598c2b6

This pull request includes:


teg:
This is really great!

msehnout (commented):
Looks pretty cool! One suggestion though: The tool is becoming quite big, I think it'd be great if we had a proper doc string for each of those `read_sth` functions mentioning the expected output format. WDYT?

thozza:
> Looks pretty cool! One suggestion though: The tool is becoming quite big, I think it'd be great if we had a proper doc string for each of those `read_sth` functions mentioning the expected output format. WDYT?

What is your idea of the expected format? An example data structure returned from the function or just a basic definitions showcasing the structure? I can definitely add some examples from regenerated image test cases and I think it could be useful.

msehnout:
> > Looks pretty cool! One suggestion though: The tool is becoming quite big, I think it'd be great if we had a proper doc string for each of those `read_sth` functions mentioning the expected output format. WDYT?
> 
> What is your idea of the expected format? An example data structure returned from the function or just a basic definitions showcasing the structure? I can definitely add some examples from regenerated image test cases and I think it could be useful.

Basic definition or idea should be enough. We can look for specific examples in the test cases. But I'm fine with both.

tools/image-info
```diff
@@ -407,20 +424,499 @@ def read_sysconfig(tree):
...
+
+
+# Read all uncommented key/values set in /etc/systemd/logind.conf
+def read_logind_conf(tree):
+    result = {}
+
+    with contextlib.suppress(FileNotFoundError):
+        with open(f"{tree}/etc/systemd/logind.conf") as f:
+            parser = configparser.RawConfigParser()
+            # prevent conversion of the opion name to lowercase
```
ondrejbudai, line 467:
```suggestion
            # prevent conversion of the option name to lowercase
```

tools/image-info
```diff
@@ -407,20 +424,499 @@ def read_sysconfig(tree):
...
+        setfiles_re = re.compile(setfiles_pattern)
+
+        for line in output.splitlines():
+            line = line.strip()
+            if not line:
+                continue
+            match = setfiles_re.match(line)
+            # do not silently ignore changes of 'setfiles' output
+            if not match:
+                raise RuntimeError("could not match line '{}' with pattern '{}'".format(line, setfiles_pattern))
```
ondrejbudai, line 538: f-string? :)

ondrejbudai (commented):
This is amazing, thanks a lot for all the hard work!

tools/image-info
thozza: Oh yes, I'm just still not used to using it :smile:

thozza:
I added doc strings to all read_* functions and fixed one additional bug that I spotted in image-info. The PR should be ready for review...

msehnout (approved):
I glanced through the doc strings and they look good! 👍 I assume the code hasn't changed except for the one bug fix.

ondrejbudai (approved):
What an amazing PR! Thanks a lot for doing this all hard work!

tools/image-info
```diff
@@ -396,31 +668,989 @@ def read_rhsm(tree):
...
+    """
+    result = {}
+
+    # iterate through all *.conf files in /etc/dracut.conf.d/
+    files = glob.glob(f"{tree}/etc/dracut.conf.d/*.conf")
+    for file in files:
+        confname = os.path.basename(file)
+        config = {}
+        with open(file) as f:
+            # dracut configuration key/values delimiter is '=' or '+='
```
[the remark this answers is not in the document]
thozza: Based on what I read in `dracut.conf(5)`, this is really tied to the option. So the answer if it has different semantics is yes, "kind of". Options that add additional values to s list of already existing ones tend to use `+=` instead of `=` for the assignment. The usage is specific to the option name and you can not use both (AFAIK). So while I understand the implementation choice to make it explicit which options are "adding" values to some list, from the ones that are assigning the provided value. But it didn't seem to be that important for `image-info` use case.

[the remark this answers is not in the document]
ondrejbudai: Thanks for the explanation!

thozza:
> I glanced through the doc strings and they look good! +1 I assume the code hasn't changed except for the one bug fix.

Except of one additional change to use `f"..."` instead of `.format()`, that's correct.

ondrejbudai:
This PR is hitting #1453 and #1454 in the CI. Both tests are unrelated to this change so I think it's safe to merge this.

PR #1435, opened 2021-05-31
GitLab CI: 1 test script per runner

atodorov:


atodorov:
https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/312399551 - shows 48 jobs created for all of the Integration os/arch + script combos.

Notes:
- we need to use yaml anchors to avoid repetition of the scripts section
- Then specifying multiple runners separately, e.g. 2 or more RUNNER definitions each of them needs the SCRIPTS definition as well, otherwise the matrix is not what we expect.

@ondrejbudai, @jrusz - how do you feel about the current syntax? If you think that's acceptable I will modify the rest of the file to match.

ondrejbudai:
First of all: How much time does this save?

Otherwise, I think that this is acceptable, I would suggest passing just the script name and args in `INTEGRATION_TESTS` and hardcoding `/usr/libexec/tests/osbuild-composer/` in the `script` field.

jrusz:
Looks pretty good, it does what's needed without repeating code, so I think it's acceptable.

atodorov:
> First of all: How much time does this save?
> 

- The last job on main reported: 33 jobs for main in 118 minutes and 55 seconds (queued for 6 seconds)
- The job I quoted earlier says: 62 jobs for PR-1435 in 746 minutes and 55 seconds (queued for 11 seconds) 

and I'm not sure how to interpret that.  Looking at the individual integration test jobs they took about 15 mins on average.

I've removed the last commit which disabled some of the jobs to have a better picture of how much time it takes. 

> Otherwise, I think that this is acceptable, I would suggest passing just the script name and args in `INTEGRATION_TESTS` and hardcoding `/usr/libexec/tests/osbuild-composer/` in the `script` field.

:+1:

ondrejbudai:
>     * The last job on main reported: 33 jobs for main in 118 minutes and 55 seconds (queued for 6 seconds)
> 
>     * The job I quoted earlier says: 62 jobs for PR-1435 in 746 minutes and 55 seconds (queued for 11 seconds)

There are 3 jobs that timed out after 12 hours so they are probably messing up with the stats.

jrusz (approved):
Looks good, I don't see any issues with this change and I believe this will speed up the whole CI run. 

Something seems to be wrong with the azure test though...

atodorov:
> Something seems to be wrong with the azure test though...

That's unrelated to this particular PR but I've also noticed increased frequency of failure with Azure lately.

PR #1439, opened 2021-06-01
Hosted edge commit

achilleas-k:
This pull request includes:

This PR adds the ability to build edge commits through the Cloud API and upload them to an S3 bucket.  On successful upload, the service returns a [Presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html) that can be used to download the object.

Image builder counterpart: https://github.com/osbuild/image-builder/pull/193

One thing to note is the lifetime of the Presigned URL.  It's currently set to 24 hours, but I don't know if that's something we want to expose or adjust.

~~## Draft~~

~~This will remain a draft until #1351 is merged so I can rebase it.  Wanted to open the PR for visibility and to discuss any issues with the approach.~~

~~I'd also like to get the same treatment for the Edge installer image type (COMPOSER-926) calling it ready, but we can also discuss whether that could be moved to a separate PR.~~

## EDIT

No longer based on 8.5 work in #1351.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -219,7 +219,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				key = uuid.New().String()
 			}
 
-			_, err = a.Upload(path.Join(outputDirectory, options.Filename), options.Bucket, key)
+			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), options.Bucket, key)
 			if err != nil {
 				appendTargetError(osbuildJobResult, err)
 				return nil
@@ -259,7 +259,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			const azureMaxUploadGoroutines = 4
 			err = azureStorageClient.UploadPageBlob(
 				metadata,
-				path.Join(outputDirectory, options.Filename),
+				path.Join(outputDirectory, exportPath, options.Filename),
 				azureMaxUploadGoroutines,
 			)
 
@@ -280,7 +280,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			}
 
 			log.Printf("[GCP] 🚀 Uploading image to: %s/%s", options.Bucket, options.Object)
-			_, err = g.StorageObjectUpload(ctx, path.Join(outputDirectory, options.Filename),
+			_, err = g.StorageObjectUpload(ctx, path.Join(outputDirectory, exportPath, options.Filename),
 				options.Bucket, options.Object, map[string]string{gcp.MetadataKeyImageName: args.Targets[0].ImageName})
 			if err != nil {
 				appendTargetError(osbuildJobResult, err)
@@ -421,7 +421,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 					ContainerName:  storageContainer,
 					BlobName:       blobName,
 				},
-				path.Join(outputDirectory, options.Filename),
+				path.Join(outputDirectory, exportPath, options.Filename),
 				azure.DefaultUploadThreads,
 			)
 			if err != nil {
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -243,6 +243,33 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 			osbuildJobResult.Success = true
 			osbuildJobResult.UploadStatus = "success"
+		case *target.AWSS3TargetOptions:
+			a, err := awsupload.New(options.Region, options.AccessKeyID, options.SecretAccessKey)
+			if err != nil {
+				appendTargetError(osbuildJobResult, err)
+				return nil
+			}
+
+			key := options.Key
+			if key == "" {
+				key = uuid.New().String()
+			}
+
+			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), options.Bucket, key)
+			if err != nil {
+				appendTargetError(osbuildJobResult, err)
+				return nil
+			}
+			url, err := a.S3ObjectPresignedURL(options.Bucket, key)
+			if err != nil {
+				appendTargetError(osbuildJobResult, err)
+				return nil
+			}
+
+			osbuildJobResult.TargetResults = append(osbuildJobResult.TargetResults, target.NewAWSS3TargetResult(&target.AWSS3TargetResultOptions{URL: url}))
+
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
 		case *target.AzureTargetOptions:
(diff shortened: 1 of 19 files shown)


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -241,6 +241,32 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				Region: options.Region,
 			}))
 
+			osbuildJobResult.Success = true
+			osbuildJobResult.UploadStatus = "success"
+		case *target.AWSS3TargetOptions:
```
teg, line 246: Might even want to introduce `jobimpl-osbuild-s3.go` and make this a whole new job type, specific to S3.

achilleas-k: Not sure I follow exactly.  How much of a job type of its own will this be?  I can see how splitting it out of `target/aws_target.go` into its own `target/s3_target.go` would make sense, but I don't see separate jobs for the other targets.

teg: Ah, you need to inspect the contents of my brain that I didn't include here, sorry. We had an idea way back when to gradually deprecate the `osbuild` job type, and replace it with things in the style of `osbild-koji` and friends. Idea being that building/pushing is one job type and "registering" is another. And rather than having the target enum the arguments are just part of the job type.

That should probably be a separate discussion from this PR though, this is good as is, just wanted to mention it as something to keep in mind/discuss eventually.

achilleas-k: ah, that's a nice plan.  I don't know if it makes sense to even plant the seed of that here though.  I'm worried we might get sidetracked a bit too far.

teg: Yeah, fair enough. Let's ignore that for now!

teg:
Could we land this irrespective of the 8.5 work? As this is only exposed on the cloud API, it should not depend on anything to do with 8.5, I think?

achilleas-k:
> Could we land this irrespective of the 8.5 work? As this is only exposed on the cloud API, it should not depend on anything to do with 8.5, I think?

You're absolutely right.  I'll "un-rebase" before marking as ready.

achilleas-k:
Cleaned up and ready.

teg (approved):
Nice!

internal/ostree/ostree.go
```diff
@@ -0,0 +1,53 @@
...
+
+var ostreeRefRE = regexp.MustCompile(`^(?:[\w\d][-._\w\d]*\/)*[\w\d][-._\w\d]*$`)
+
+type OSTreeRequest struct {
+	URL    string `json:"url"`
+	Ref    string `json:"ref"`
+	Parent string `json:"parent"`
+}
+
+func VerifyRef(ref string) bool {
```
ondrejbudai, line 22: Can you add a small test for this function?

internal/upload/awsupload/awsupload.go
```diff
@@ -284,3 +284,17 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
...
 	return registerOutput.ImageId, nil
 }
+
+func (a *AWS) S3ObjectPresignedURL(bucket, objectKey string) (string, error) {
+	log.Printf("[AWS] 📋 Generating Presigned URL for S3 object %s/%s", bucket, objectKey)
+	req, _ := a.s3.GetObjectRequest(&s3.GetObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(objectKey),
+	})
+	url, err := req.Presign(24 * time.Hour)
```
ondrejbudai, line 294: I wonder if this should be configurable in the TargetOptions at some point. We definitely don't need to implement it now though. :)

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -241,6 +241,32 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+
+			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), options.Bucket, key)
+			if err != nil {
+				appendTargetError(osbuildJobResult, err)
+				return nil
+			}
+			url, err := a.S3ObjectPresignedURL(options.Bucket, key)
+			if err != nil {
+				appendTargetError(osbuildJobResult, err)
+			}
```
ondrejbudai, line 267: I think we should also `return nil` here. Otherwise, this will look like a success.

ondrejbudai (dismissed):
Two notes, one (imho) issue. Otherwise, this is great!

internal/upload/awsupload/awsupload.go
achilleas-k: I'll set it to the maximum until then instead.

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: Thanks for catching this.

achilleas-k:
Added a slightly unrelated change: Requests to the Koji API now read the exports list from the image type.  Since this PR adds the same for the cloud API, I think it's good to finally have all build jobs support this and use the export path when handling the final image.

teg:
Very nice! Should we add some upload test too?

achilleas-k:
> Very nice! Should we add some upload test too?

Added a test to `test/cases/api.sh` that builds an edge commit, downloads it using the Presigned URL, and checks if it contains the ostree ref that was used in the request.

teg (approved):
Nice! ❤️

test/cases/api.sh
```diff
@@ -53,8 +54,15 @@ case $CLOUD_PROVIDER in
   "$CLOUD_PROVIDER_AZURE")
     echo "Testing Azure"
     ;;
+  "$CLOUD_PROVIDER_AWS_S3")
+    echo "Testing S3 bucket upload"
+    if [[ $ID != "rhel" ]]; then
+        echo "Skipped. S3 upload test is only tested on RHEL."
```
ondrejbudai, line 60: A tiny nitpick: I would love to see here an explanation why this is tested only on RHEL (the only distro supporting rhel-edge-commit). Someone will surely stumble upon this.

ondrejbudai (approved):
Great work!

test/cases/api.sh
achilleas-k: Good thinking thanks.

achilleas-k:
Doing some in-CI print debugging because I cannot for the life of me figure out why the commit doesn't contain the ref at the path I'm checking.

achilleas-k:
> Doing some in-CI print debugging because I cannot for the life of me figure out why the commit doesn't contain the ref at the path I'm checking.

Found the issue in the request payload.  Should be fixed now :crossed_fingers:

achilleas-k:
After a quick discussion with @teg and @thozza, we agreed that it's better to move the ostree options into the `ImageRequest` part of the request data.

ondrejbudai (approved):

PR #1440, opened 2021-06-01
schutzbot: send notifications to slack

gicmo:
Send notifications about successful and failed builds to slack instead of Telegram.


atodorov:
> Send notifications about successful and failed builds to slack instead of Telegram.

Would you like to add the same for the GitLab pipeline ?

atodorov (approved):

PR #1442, opened 2021-06-03
tests: Image tests update for osbuild

jrusz:
We need to update `image_tests.sh` in order to be able to run it in Gitlab CI for osbuild. Originally we had the cloud credentials in Jenkins as files but that can't be masked in Gitlab CI. I don't know which approach is best here, whether to keep the current implementation and just print the variables one by one to a file to be used by the `TEST_CMD`, print them there one by one or I don't know, I'm open to suggestions.  

This pull request includes:

### DIFF
--- a/test/cases/image_tests.sh
+++ b/test/cases/image_tests.sh
@@ -6,6 +6,12 @@ source /etc/os-release
 ARCH=$(uname -m)
 DISTRO_CODE="${DISTRO_CODE:-${ID}_${VERSION_ID//./}}"
 
+if [[ -n "$CI_BUILD_ID" ]]; then
+    BUILD_ID="${BUILD_ID:-${CI_BUILD_ID}}"
+else
+    BUILD_ID="${BUILD_ID:-$(uuidgen)}"
+fi
+
 WORKING_DIRECTORY=/usr/libexec/osbuild-composer
 IMAGE_TEST_CASE_RUNNER=/usr/libexec/osbuild-composer-test/osbuild-image-tests
 IMAGE_TEST_CASES_PATH=/usr/share/tests/osbuild-composer/manifests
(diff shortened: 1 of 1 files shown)


atodorov:
> I don't know which approach is best here, whether to keep the current implementation and just print the variables one by one to a file to be used by the `TEST_CMD`, print them there one by one or I don't know, I'm open to suggestions.


The command which consumes the creds file is:

```
    TEST_CMD="env $(cat "$AWS_CREDS" "$AZURE_CREDS" "$OPENSTACK_CREDS" "$VCENTER_CREDS") BRANCH_NAME=${BRANCH_NAME-main} BUILD_ID=$BUILD_ID DISTRO_CODE=$DISTRO_CODE $TEST_RUNNER -test.v ${IMAGE_TEST_CASES_PATH}/${TEST_CASE_FILENAME}"
```

which to my understanding will simply define each line of the file as separate env variable. In other words it's like prefixing each var name with `export`.  And this is more or less what GitLab is doing as well so I think if we drop the intermediate file the rest of the test suite should be able to find the necessary variables b/c they should already be defined, no ?

jrusz:
> > I don't know which approach is best here, whether to keep the current implementation and just print the variables one by one to a file to be used by the `TEST_CMD`, print them there one by one or I don't know, I'm open to suggestions.
> 
> The command which consumes the creds file is:
> 
> ```
>     TEST_CMD="env $(cat "$AWS_CREDS" "$AZURE_CREDS" "$OPENSTACK_CREDS" "$VCENTER_CREDS") BRANCH_NAME=${BRANCH_NAME-main} BUILD_ID=$BUILD_ID DISTRO_CODE=$DISTRO_CODE $TEST_RUNNER -test.v ${IMAGE_TEST_CASES_PATH}/${TEST_CASE_FILENAME}"
> ```
> 
> which to my understanding will simply define each line of the file as separate env variable. In other words it's like prefixing each var name with `export`. And this is more or less what GitLab is doing as well so I think if we drop the intermediate file the rest of the test suite should be able to find the necessary variables b/c they should already be defined, no ?

Oh, yeah, I guess you're right, they are already defined as env vars, so it should work. I'll try it.

.gitlab-ci.yml
```diff
@@ -59,8 +59,8 @@ Image_test:
   parallel:
     matrix:
       - RUNNER:
-          - aws/fedora-33-x86_64
-          - aws/rhel-8-x86_64
+          - openstack/fedora-33-x86_64
+          - openstack/rhel-8-x86_64
```
atodorov, line 63: Why do we need this change?

test/cases/image_tests.sh
```diff
@@ -5,6 +5,12 @@ set -euo pipefail
 source /etc/os-release
 ARCH=$(uname -m)
 DISTRO_CODE="${DISTRO_CODE:-${ID}_${VERSION_ID//./}}"
+BUILD_ID="${BUILD_ID:-${CI_PIPELINE_ID}}"
+
+# Set DISTRO_CODE for rhel-8 runners
+if [[ "$CI_JOB_NAME" =~ .*rhel-8-.* ]]; then
+    DISTRO_CODE="rhel_8"
+fi
```
atodorov, line 13: I don't understand why we need this change ?

.gitlab-ci.yml
jrusz: I was just wondering if it runs faster in Openstack but it seems it does not. I'm going to change it back to aws runners. Also I may have wanted to try if it fixes the Openstack test failure but I guess that's just some credentials missing.

ondrejbudai: One of our priorities is to move as many jobs as possible out of Openstack.

test/cases/image_tests.sh
jrusz: When this test was ran in Jenkins it had `DISTRO_CODE` specified for each stage. Without this change, when it's ran on regular rhel-8 runner, which is currently 8.3 it will fail with `ls: cannot access 'rhel_83-x86_64*.json': No such file or directory` because there is no 8.3 manifest... What this does is if the runner is regular RHEL 8, it sets the `DISTRO_CODE` to `rhel_8` ... This is what I came up with, it should work when we update the runners to use RHEL 8.4 because they'll be still be called rhel 8 runners. So no hard coding for 8.3 and then having to change when we update...

jrusz:
I'm seeing some new failures that I reported in #1464 and Openstack is still failing with `No OpenStack credentials given` even though I added the same credentials that are in Jenkins. I can see the test is passing in when ran in Jenkins for osbuild repository and I didn't notice any change in the test since the tagged version that's being used there...

ondrejbudai:
`sudo` doesn't keep environment variables

jrusz:
> `sudo` doesn't keep environment variables

Thanks, `-E` helped... Now just the failures with IoT and edge, I have no idea what's happening there.

test/cases/image_tests.sh
thozza: This change and not having `BUILD_ID` or `CI_BUILD_ID` defined, e.g. if I simply just install `osbuild-composer-tests` from the S3 repo, makes is impossible to simply execute `image-tests.sh` locally

thozza: The same goes for `CI_JOB_NAME`. I would suggest to keep this test cases ideally working out of the box even outside of the CI environment...

jrusz: Ok so I removed that change, I guess it's better to define `DISTRO_CODE` in .gitlab-ci.yml file if it needs to be something different. Also I don't think this worked outside CI before as `BUILD_ID` was just set to `BUILD_ID` and I believe that comes from Jenkins. I changed it to some random uuid though in case it's not defined so it should work outside CI now.

jrusz:
Image tests for F-33 and RHEL-8 are passing https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/326855410

atodorov (changes requested):
LGTM now. Remove the first commit and we can merge.

atodorov (approved):
LGTM, @thozza do you have any other comments as to how this test script behaves?  Does it work for you including the current changes ?

thozza (approved):
These changes look good to me. I don't have any additional feedback at the moment.

jrusz:
All the failed jobs failed because of #1502 or PSI outage, I restarted them now.

ondrejbudai (approved):

PR #1438, opened 2021-06-01
distro: add fedora 34 and 35 aliases to f33

msehnout:
This PR is an alternative to https://github.com/osbuild/osbuild-composer/pull/1426, specifically a result of this discussion: https://github.com/osbuild/osbuild-composer/pull/1426#issuecomment-852196746

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #1444, opened 2021-06-03
Build RHEL 8.5 from 8.4 image type definitions

msehnout:
Depends on: #1438 

This pull request includes:


internal/distroregistry/distroregistry.go
```diff
@@ -24,7 +24,7 @@ var supportedDistros = []supportedDistro{
 	{rhel8.New, rhel8.NewHostDistro},
 	{rhel84.New, rhel84.NewHostDistro},
 	{rhel84.NewCentos, rhel84.NewCentosHostDistro},
-	{rhel85.New, rhel85.NewHostDistro},
+	{rhel85.New, rhel84.NewRhel85HostDistro}, // Replace once RHEL8.5 package actually have any image types
```
thozza, line 27: You should replace `rhel85.New` with `rhel84.NewRhel85`, because the distro returned by `rhel85.New()` does not have any image types defined. This change will allow us to not only build images on 8.5, but also 8.5 itself.

internal/distro/test_distro/distro.go
```diff
@@ -14,6 +14,7 @@ import (
 type TestDistro struct {
 	name             string
 	modulePlatformID string
+	ostreeRef        string
```
thozza, line 17: You should also define "ostreeRef" constants for "TestDistro" and "TestDistro2" and set them in `New()` and `New2()` when calling `newTestDistro()`. And as a result please extend the `newTestDistro()` to accept an "ostreeRef" value.

thozza (commented):
Looks Good! I added two comments. Also I'm wondering whether we should add CI testing for F34, F35 and RHEL-85 as part of this PR, what do you think @msehnout ?

ondrejbudai:
When will we remove the alias? When all 8.5 image types are implemented? @achilleas-k thoughts?

thozza:
> When will we remove the alias? When all 8.5 image types are implemented? @achilleas-k thoughts?

I would say that when at least some of them are defined. E.g. the KVM guest image as we need it for 8.5. I would not block ourselves on the AMI image and would also suggest to not to port its 8.4 version to 8.5 and rather use the one that will be based on the official RHEL images.

achilleas-k:
> > When will we remove the alias? When all 8.5 image types are implemented? @achilleas-k thoughts?
> 
> I would say that when at least some of them are defined. E.g. the KVM guest image as we need it for 8.5. I would not block ourselves on the AMI image and would also suggest to not to port its 8.4 version to 8.5 and rather use the one that will be based on the official RHEL images.

It does raise an interesting point.  For instance if we add some but not all image types and remove the alias, we're basically losing features (we lose support for certain image types).  It would also be weird to mix explicit with fallback support I think.

I think @thozza's idea is good.  We choose the ones we want to always support and remove the alias once those are done.

ondrejbudai:
> I would say that when at least some of them are defined. E.g. the KVM guest image as we need it for 8.5. I would not block ourselves on the AMI image and would also suggest to not to port its 8.4 version to 8.5 and rather use the one that will be based on the official RHEL images.

The critical usecase for guest images is the brew one. On the other hand, we happily use the 8.4 definition even for 8.5 and it just works. Thus, we are just fine there.

> It does raise an interesting point. For instance if we add some but not all image types and remove the alias, we're basically losing features (we lose support for certain image types). It would also be weird to mix explicit with fallback support I think.

Mixing explicit with fallback sounds like a lot of extra logic and I'm honestly afraid to implement this. I also do agree that losing features is weird.

My takeaway from this is that we should probably stick to the alias-duplicate-iterate model for 8.6:

1) During 8.5 development, let's create an alias for 8.6, so osbuild-composer works immediately after 8.6 is branched out. 

2) When the 8.6 development starts, the first PR should remove the 8.6 alias and duplicate the 8.5 definition into 8.6. Also, the alias for 8.7 should be created.

3) During the rest of the 8.6 development, improvements are iterated upon the duplicated definition.

This model allows us to always support everything. I know that starting 8.5 development from scratch allowed us to do a big refactoring from the start but it's imho important to realize that we are currently sacrificing some features for some time. I personally prefer doing smaller and safer iterations.

---

I wonder what to do in this stage with 8.5 though. If we introduce an alias for 8.5, it feels natural to me that the PR removing the alias should implement all image types. Why? Let's consider the other options:

1) A PR adds an image definition to 8.5 but doesn't remove the alias: The new definition is therefore not tested in the CI, that's weird.

2) A PR adds some (but not all) image definitions to 8.5 and removes the alias: We're losing features, feels weird to me.

Both options sound questionable to me. I would prefer to always have everything.

I wonder if duplicating 8.4 definitions into 8.5 and merging the work already in `main` from @achilleas-k is something that we should take as the next step. And of course iterate from there... Not sure how hard this can be.

achilleas-k:
I completely understand your concerns @ondrejbudai and I can't disagree.

> If we introduce an alias for 8.5, it feels natural to me that the PR removing the alias should implement all image types.

I think this is doable and we should decide to do it.  It'll be big and might require coordination, but if this is the ideal scenario, let's do this.

  > 1. A PR adds an image definition to 8.5 but doesn't remove the alias: The new definition is therefore not tested in the CI, that's weird.

If we don't agree that adding everything in a single PR is possible, maybe we could find a way to separately test in CI the new definitions while the alias still exists, but that would probably introduce extra, unnecessary work.

thozza:
What about introducing the non-alias RHEL-8.5 definitions under a different name in the meantime, e.g. `rhel-85-wip` until we are in the state to remove the alias? It would allow us to test new definitions in CI and still not to regress in 8.5 functionality-wise....

bcl:
I still need to look into this in more detail, but we're going to need to stop removing things -- once building for other release is actually working (PR #1427) users are going to expect to be able to build for previous releases. Currently I can build for rhel-8 but when trying to use rhel-84 it complains that there is no repository.

ondrejbudai:
> I still need to look into this in more detail, but we're going to need to stop removing things -- once building for other release is actually working (PR #1427) users are going to expect to be able to build for previous releases.

This PR doesn't propose dropping things for already released distros therefore I think this should cause any concerns for the cross-distro feature. :)

teg:
> I completely understand your concerns @ondrejbudai and I can't disagree.
> 
> > If we introduce an alias for 8.5, it feels natural to me that the PR removing the alias should implement all image types.
> 
> I think this is doable and we should decide to do it.  It'll be big and might require coordination, but if this is the ideal scenario, let's do this.

I agree with Ondrej's argument why this should be preferable. In the future we will hopefully be able to avoid such a blocking rewrite, but for now I guess we will have to make the most out of it.

Rather than the `wip` distro, we could possibly do the coordination work in a branch?

Maybe we need to have a meeting about this once the next osbuild is out to settle on a plan?

>   > 1. A PR adds an image definition to 8.5 but doesn't remove the alias: The new definition is therefore not tested in the CI, that's weird.
> 
> If we don't agree that adding everything in a single PR is possible, maybe we could find a way to separately test in CI the new definitions while the alias still exists, but that would probably introduce extra, unnecessary work.

ondrejbudai:
I discussed this with Tom and we agreed that at this point, it would make the 8.5 situation even worse. On the other hand, this is still a great idea and we definitely want this solution but for 8.6. This will be implemented as soon as 8.5 definitions are done. Thanks, Martin, for doing this!

PR #1449, opened 2021-06-07
tests: Use CI_BUILD_ID, not CI_PIPELINE_ID. Fixes #1448

### LINKED ISSUE #1448: the integration test split causes cloud cleaner to delete artifacts from other job
Now, we run each integration test in its own job. This causes the cloud cleaner to malfunction: GCP jobs are usually faster than Azure. Therefore, the cloud cleaner runs faster in the GCP job and it deletes both GCP and Azure artifacts which causes the Azure job to fail.

@atodorov can you look into this?

Many examples can be found in this pipeline: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/314796558/builds

atodorov:
PIPELINE_ID is the same across different jobs running in the same
pipeline while BUILD_ID is unique for every job.

Note: In the case where we have 1 test script/runner
CI_BUILD_ID == CI_JOB_ID

### DIFF
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 source /etc/os-release
 DISTRO_CODE="${DISTRO_CODE:-${ID}_${VERSION_ID//./}}"
 BRANCH_NAME="${BRANCH_NAME:-${CI_COMMIT_BRANCH}}"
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
This indeed fixed the Azure issue, thanks!

PR #1450, opened 2021-06-07
X rh identity header filter

Gundersanne:


cmd/osbuild-worker/main.go
```diff
@@ -90,12 +92,13 @@ func main() {
 		Azure *struct {
 			Credentials string `toml:"credentials"`
 		} `toml:"azure"`
+		OfflineToken string `toml:"offline_token"`
```
teg, line 95: Should this be an external file, like the other credentials? I think we might print this file to the logs.

internal/worker/api/api.go
```diff
@@ -3,3 +3,5 @@
 package api
 
 const BasePath = "/api/worker/v1"
+const CloudBasePath = "/api/composer-worker/v1"
+const OAuthURL="https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
```
teg, line 7: Should we make this configurable to work with staging/not only RH infra?

teg (commented):
Love this! thanks for picking it up!

cmd/osbuild-worker/main.go
Gundersanne: ahh yea, good point, as this is a secret

internal/worker/api/api.go
Gundersanne: absolutely

internal/worker/api/api.go
```diff
@@ -3,3 +3,5 @@
 package api
 
 const BasePath = "/api/worker/v1"
+const CloudBasePath = "/api/composer-worker/v1"
```
teg, line 6: I guess we should transition towards this being the only one, maybe expose both on the server side for now (in a follow-up)?

ondrejbudai: I would actually prefer both paths to be implemented even in this PR. I think this makes this transition simpler and it's actually trivial: All we have to do is to call `api.RegisterHandlers` in server twice.

internal/worker/server.go
```diff
@@ -46,13 +50,26 @@ type JobStatus struct {
 	Canceled bool
 }
 
+type IdentityHeader struct {
```
ondrejbudai, line 53: I suggest making this struct private. Actually, it could live inside the `VerifyIdentityHeader` method but both is fine with me.

cmd/osbuild-worker/main.go
```diff
@@ -28,6 +28,8 @@ type connectionConfig struct {
 	ClientCertFile string
 }
 
+
```
ondrejbudai, line 31: Extra newlines detected! :D

internal/worker/server.go
Gundersanne: We might wanna use it in handlers at some point. I'm not sure. But until then it can live inside the method

teg:
Does this change the worker api?

teg (dismissed):
Nice!

ondrejbudai (changes requested):
I think we should test the OAuth implementation. It would be very nice to have at least a unit test.

But more importantly: This is an amazing addition, can we please have a post in `docs/news/unreleased/`? Pretty please! :heart_eyes:

Gundersanne:
@ondrejbudai I made a unit test which mimics the oauth sso, proxy, and then also the worker server. Sadly I couldn't get it to work with the default fsjobqueue so I made a tiny mock queue.

When I try it with the regular queue it gets stuck in reflect.Select when trying to dequeue a job uuid, and I'm not sure why :/

teg:
> @ondrejbudai I made a unit test which mimics the oauth sso, proxy, and then also the worker server.

Nice!

> Sadly I couldn't get it to work with the default fsjobqueue so I made a tiny mock queue.
> 
> When I try it with the regular queue it gets stuck in reflect.Select when trying to dequeue a job uuid, and I'm not sure why :/

Hm, this is a bit disconcerting, no? I have no idea either, but probably something we need to figure out (in the unlikely event there is an actual regression here somewhere)?

teg (approved):
Yay!

ondrejbudai (approved):

Issue #1445, opened 2021-06-03
Add a way to add new RHEL channel as source for a build


pypingou:
**Describe the bug**
Currently only the baseos and appstream channels are available for RHEL images.
Adding a new channel using ``composer-cli sources add ...`` does not work as it gets a 403 error from the CDN, even if the system is properly subscribed and can pull just fine other packages from the CDN

**Environment**
 * OS version: RHEL 8.4
 * osbuild-composer version: osbuild-composer-28.4-1.el8.x86_64

**To Reproduce**

* Try adding a new source to image-builder that enables the OCP channel or the Code Ready Builder channel
* Run ``composer-cli compose depsolve <blueprint_name>``
* See it fail because of a 403 returned by the CDN

**Expected behavior**
Being able to add new channel via the ``sources new`` command.

**Additional context**
The issue can be circumvent by editing ``/usr/share/osbuild-composer/repositories/rhel-8.json`` manually.
It seems the issue is that the current format/UI doesn't support specifying ``rhsm=true`` when adding a new source.

msehnout:
Thanks for filling in this issue! You are absolutely correct that the sources cannot define RHSM, which is something we should fix: https://github.com/osbuild/osbuild-composer/blob/9a4b5763e9da6701d67d12b8c048d07a70b668de/internal/store/store.go#L45

PR #1452, opened 2021-06-09
weldr: accept rhsm parameter in sources

msehnout:
The system sources allow specification of the rhsm parameter, but it
isn't available in the sources configured over the Weldr API. This patch
implements support for it.

Fixes: #1445

This pull request includes:

### DIFF
--- a/internal/weldr/json.go
+++ b/internal/weldr/json.go
@@ -187,6 +187,7 @@ func NewSourceConfigV1(id string, s store.SourceConfig) SourceConfigV1 {
 	sc.CheckSSL = s.CheckSSL
 	sc.System = s.System
 	sc.Distros = s.Distros
+	sc.RHSM = s.RHSM
 
 	return sc
 }
@@ -203,6 +204,7 @@ type SourceConfigV1 struct {
 	Proxy    string   `json:"proxy,omitempty" toml:"proxy,omitempty"`
 	GPGUrls  []string `json:"gpgkey_urls,omitempty" toml:"gpgkey_urls,omitempty"`
 	Distros  []string `json:"distros,omitempty" toml:"distros,omitempty"`
+	RHSM     bool     `json:"rhsm" toml:"rhsm"`
 }
 
 // Key returns the key, .ID in this case
@@ -229,6 +231,7 @@ func (s SourceConfigV1) SourceConfig() (ssc store.SourceConfig) {
 	ssc.CheckGPG = s.CheckGPG
 	ssc.CheckSSL = s.CheckSSL
 	ssc.Distros = s.Distros
+	ssc.RHSM = s.RHSM
 
 	return ssc
 }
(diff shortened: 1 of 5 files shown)


teg (approved):
This makes sense to me! Solves a real problem of adding non standard repos from the RH CDN. Would be great if Brian could give it a sanity check too though.

internal/weldr/json.go
```diff
@@ -139,6 +139,7 @@ type SourceConfigV0 struct {
 	System   bool     `json:"system" toml:"system"`
 	Proxy    string   `json:"proxy,omitempty" toml:"proxy,omitempty"`
 	GPGUrls  []string `json:"gpgkey_urls,omitempty" toml:"gpgkey_urls,omitempty"`
+	RHSM     bool     `json:"rhsm" toml:"rhsm"`
```
bcl, line 142: Technically this should only be added to the V1 Source structs/json since V0 is 'finished'.

bcl (changes requested):
GH won't let you comment on other bits of code, so I'll put this here:

RHSM value also needs to be copied in other functions that convert to/from the store's structure and the JSON API structure. I *think* the complete list is:

* SourceConfigV1.SourceConfig()
* NewSourceConfigV1

internal/weldr/json.go
msehnout: Where is V0 used?

bcl: /api/v0/projects/source/* uses it.

msehnout: Sure, but how does one use it with composer-cli? I tried it on RHEL and it uses the V1 version.

Anyhow, I've made the changes you requested. Is it ok now?

bcl: composer-cli uses the v1 route, but the v0 routes and responses should be left alone since they are 'done'.

bcl (dismissed):
Looks good to me.

msehnout:
Added a test case and a news entry.

teg:
This is looking great to me! @atodorov do you have any thoughts on the testing of this? There is a nice unit test I think, but not sure what integration test (if any) it makes sense to hook into to verify this? The RT or Ansible repositories are candidates we could test against, but I think we don't currently test against the CDN in our CI, so this may not make sense?

atodorov (changes requested):
I don't see any changes related to the test suite/new test scenarios being added. At the very minimum just to make sure assignments of values continue working across the different functions. 

At the same time I see slightly decreased test coverage, for example in internal/store/json.go an if condition that was covered before is now not covered. Given there are no other changes in the PR that may be a result of the new field in the struct. Please check it out.

msehnout:
> I don't see any changes related to the test suite/new test scenarios being added. At the very minimum just to make sure assignments of values continue working across the different functions.
> 
> At the same time I see slightly decreased test coverage, for example in internal/store/json.go an if condition that was covered before is now not covered. Given there are no other changes in the PR that may be a result of the new field in the struct. Please check it out.

You are right, I thought this PR already contained a test case. Maybe I misremember.

Anyway, I've introduced one additional unit test and added a verification to the weldr client tests. Is it ok like this?

atodorov (approved):

PR #1455, opened 2021-06-09
prometheus: collect metrics from cloudapi

chloenayon:
Create prometheus package to provide infrastructure to gather metrics. Implement it 
in cloudapi and worker to collect information including number of total and successful 
compose requests. 

This pull request includes:


Gundersanne (changes requested):
Looking good, two items:

The ComposeSuccesses need some work.

And for testing we could call `/metrics` before and after a compose in the api test (and maybe koji?), and then make sure the relevant metrics increased?

osbuild-composer.spec
```diff
@@ -249,13 +250,13 @@ cd $PWD/_build/src/%{goipath}
 %endif
 
 %post
-%systemd_post osbuild-composer.service osbuild-composer.socket osbuild-composer-api.socket osbuild-remote-worker.socket
+%systemd_post osbuild-composer.service osbuild-composer.socket osbuild-composer-api.socket osbuild-remote-worker.socket
```
Gundersanne, line 253: Trailing whitespace, here and below

Gundersanne (commented):
So small linting error https://github.com/osbuild/osbuild-composer/pull/1455/checks?check_run_id=2797485029

Gundersanne (changes requested):
Thank you! The test needs some fixes but the placement and idea are correct.

test/cases/api.sh
```diff
@@ -499,6 +499,15 @@ OUTPUT=$(curl \
 
 COMPOSE_ID=$(echo "$OUTPUT" | jq -r '.id')
 
+METRICS_OUTPUT=$(curl \
+  --cacert ca-crt.pem \
+  --key client-key.pem \
+  --cert client-crt.pem \
+  https://localhost:4430/metrics)
+
+METRICS_LINE=$(echo "$METRICS_OUTPUT" | grep "^total_compose_requests")
```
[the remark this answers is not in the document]
chloenayon: Thank you for this especially, I was having trouble coming up with a way to get the number in one line

Gundersanne:
Hmm in the api test it tries calling metrics `curl --cacert ca-crt.pem --key client-key.pem --cert client-crt.pem https://localhost/metrics` but the credentials aren't entirely correct.

`curl --silent --show-error --cacert /etc/osbuild-composer/ca-crt.pem --key /etc/osbuild-composer/client-key.pem --cert /etc/osbuild-composer/client-crt.pem --header 'Content-Type: application/json' --request POST --data @/tmp/tmp.3pvnqah8Lh/request.json https://localhost/api/composer/v1/compose` is the compose request, they have absolute paths towards the ca/key/cert.

test/cases/api.sh
```diff
@@ -499,6 +499,14 @@ OUTPUT=$(curl \
 
 COMPOSE_ID=$(echo "$OUTPUT" | jq -r '.id')
 
+METRICS_OUTPUT=$(curl \
+  --cacert /etc/osbuild-composer/ca-crt.pem \
+  --key /etc/osbuild-composer/client-key.pem \
+  --cert /etc/osbuild-composer/client-crt.pem \
+  https://localhost/metrics)
+
+PREV_COMPOSES=$(echo "$METRICS_OUTPUT" | grep "^total_compose_requests" | cut -f2 -d' ')
```
Gundersanne, line 508: As the first `/metrics` curl happens after the first `/compose` request, the `total_compose_requests` is already 1; so the test later on will fail as `$((PREV_COMPOSES + 1))` is 2.

So just move the first /`metrics` call to before the first `/compose` call and you should be good. Mind if you also check the `"^total_successful_compose_requests"` ? Just for completeness, check it in the same way as PREV_COMPOSES, but then like PREV_SUCCESSFUL_COMPOSES or something.

test/cases/api.sh
```diff
@@ -531,6 +531,16 @@ esac
...
 #
 
+function collectMetrics(){
+    METRICS_OUTPUT=$(curl \
+                          --cacert /etc/osbuild-composer/ca-crt.pem \
+                          --key /etc/osbuild-composer/client-key.pem \
+                          --cert /etc/osbuild-composer/client-crt.pem \
+                          https://localhost/metrics)
+
+    "$1"="$(echo "$METRICS_OUTPUT" | grep "^total_compose_requests" | cut -f2 -d' ')"
```
Gundersanne, line 541: Bash functions are a little quirky, they return whatever gets written to stdout in that function. It is possible to do what you're doing here using `declare -n out="$1"` and then `out="whatever value"` but I'm not super familiar with `declare`.

```
function barme () {
    declare -n out=$1
    out="bar"
}

FOO="foo"
declare -n FOO
barme FOO
echo $FOO # == bar
``` 

But this is difficult to read I think. Just using stdout will make a lot more sense to most people.

```
function collectMetrics() {
...
echo "$METRICS_OUTPUT" |  ... |cut -f2 -d' '
}
COMPOSES="$(collectMetrics)"
# or just
test "$(collectMetrics)" = "1"
```

Gundersanne (approved):
thanks!

Issue #1454, opened 2021-06-09
libvirt test on CS8 cannot ssh into the VM


ondrejbudai:
No idea what's causing this, see the example:

https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1330953776

ondrejbudai:
Log from a CS8 UEFI VM:

```
[    0.000000] Linux version 4.18.0-277.el8.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 8.4.1 20200928 (Red Hat 8.4.1-1) (GCC)) #1 SMP Wed Feb 3 20:35:19 UTC 2021
[    0.000000] Command line: BOOT_IMAGE=(hd0,gpt3)/boot/vmlinuz-4.18.0-277.el8.x86_64 root=UUID=5028726e-075f-4504-b42f-26550789f4bc console=tty0 console=ttyS0,115200n8 no_timer_check net.ifnames=0 crashkernel=auto
[    0.000000] x86/fpu: x87 FPU will use FXSAVE
[    0.000000] BIOS-provided physical RAM map:
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009ffff] usable
[    0.000000] BIOS-e820: [mem 0x0000000000100000-0x000000003e8edfff] usable
[    0.000000] BIOS-e820: [mem 0x000000003e8ee000-0x000000003eaedfff] reserved
[    0.000000] BIOS-e820: [mem 0x000000003eaee000-0x000000003eb75fff] usable
[    0.000000] BIOS-e820: [mem 0x000000003eb76000-0x000000003eb7dfff] ACPI data
[    0.000000] BIOS-e820: [mem 0x000000003eb7e000-0x000000003ebfdfff] ACPI NVS
[    0.000000] BIOS-e820: [mem 0x000000003ebfe000-0x000000003effffff] usable
[    0.000000] BIOS-e820: [mem 0x000000003f000000-0x000000003fffffff] reserved
[    0.000000] BIOS-e820: [mem 0x00000000b0000000-0x00000000bfffffff] reserved
[    0.000000] NX (Execute Disable) protection: active
[    0.000000] extended physical RAM map:
[    0.000000] reserve setup_data: [mem 0x0000000000000000-0x000000000009ffff] usable
[    0.000000] reserve setup_data: [mem 0x0000000000100000-0x000000003d048017] usable
[...]

PR #1456, opened 2021-06-09
tests: fix journalctl killing once again

ondrejbudai:
3a8c6c8a introduced a new logic for killing journalctl. Unfortunately, it
doesn't work properly. In ostree tests, multiple journalctls are spawned
but there can be only one trap active at a time. This caused all but the last
journalctls to hang indefinitely. Unfortunately, hanging background processes
is something that causes the GitLab CI to hang indefinitely as well.

This commit modifies the logic a bit: The trap is still set. However, there's
also an explicit kill of journalctl after the compose is finished. After the
process is successfully killed, the trap is removed.

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -159,6 +159,10 @@ greenprint "💬 Getting compose log and metadata"
 get_compose_log "$COMPOSE_ID"
 get_compose_metadata "$COMPOSE_ID"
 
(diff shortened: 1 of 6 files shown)


teg (approved):

ondrejbudai:
#1454 appeared, otherwise we're green.

PR #1466, opened 2021-06-10
cloudapi: replace chi with echo

chloenayon:
standardize middleware frameworks 

This pull request includes:


Gundersanne (changes requested):
I really like this, we should standardize on something.

This needs a rebase and will need a rebase after https://github.com/osbuild/osbuild-composer/pull/1455 lands as well. Marking as blocked for now.

Gundersanne:
Should be unblocked now, this will be a bit hairy to rebase I think, if you need help let me know!

internal/cloudapi/server.go
```diff
@@ -639,7 +600,8 @@ func (server *Server) ComposeMetadata(w http.ResponseWriter, r *http.Request, id
 		resp.OstreeCommit = &commitMetadata.Compose.OSTreeCommit
 	}
 
-	if err := json.NewEncoder(w).Encode(resp); err != nil {
+	if err := json.NewEncoder(ctx.Response()).Encode(resp); err != nil {
 		panic("Failed to write response: " + err.Error())
 	}
+	return nil
```
Gundersanne, line 606: Ditto

internal/cloudapi/server.go
```diff
@@ -391,51 +375,44 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
...
-	w.Header().Set("Content-Type", "application/json; charset=utf-8")
-	w.WriteHeader(http.StatusCreated)
-	err = json.NewEncoder(w).Encode(response)
-	if err != nil {
-		panic("Failed to write response")
-	}
+
+	ctx.Response().Header().Set("Content-Type", "application/json; charset=utf-8")
+	ctx.Response().WriteHeader(http.StatusOK)
+	return json.NewEncoder(ctx.Response()).Encode(response)
```
Gundersanne, line 395: `return ctx.JSON(http.StatusOK, response)`

internal/cloudapi/server.go
```diff
@@ -50,94 +63,81 @@ func NewServer(workers *worker.Server, rpmMetadata rpmmd.RPMMD, distros *distror
...
+	contentType := ctx.Request().Header["Content-Type"]
 	if len(contentType) != 1 || contentType[0] != "application/json" {
-		http.Error(w, "Only 'application/json' content type is supported", http.StatusUnsupportedMediaType)
-		return
+		return echo.NewHTTPError(http.StatusUnsupportedMediaType, "Only 'application/json' content type is supported")
 	}
 
 	var request ComposeRequest
-	err := json.NewDecoder(r.Body).Decode(&request)
+	err := json.NewDecoder(ctx.Request().Body).Decode(&request)
```
Gundersanne, line 127: We should really do this in a binder, see [kojiapi for reference](https://github.com/osbuild/osbuild-composer/blob/main/internal/kojiapi/server.go#L490). But could also do this in a follow-up.

internal/cloudapi/server.go
```diff
@@ -50,94 +63,81 @@ func NewServer(workers *worker.Server, rpmMetadata rpmmd.RPMMD, distros *distror
...
 
 // Compose handles a new /compose POST request
-func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
-	contentType := r.Header["Content-Type"]
+func (h *apiHandlers) Compose(ctx echo.Context) error {
+	contentType := ctx.Request().Header["Content-Type"]
 	if len(contentType) != 1 || contentType[0] != "application/json" {
-		http.Error(w, "Only 'application/json' content type is supported", http.StatusUnsupportedMediaType)
-		return
+		return echo.NewHTTPError(http.StatusUnsupportedMediaType, "Only 'application/json' content type is supported")
```
Gundersanne, line 123: This would get simplified using the Binder too.

internal/cloudapi/server.go
```diff
@@ -50,94 +63,81 @@ func NewServer(workers *worker.Server, rpmMetadata rpmmd.RPMMD, distros *distror
...
+		e.Use(server.VerifyIdentityHeader)
 	}
-	r.Use(server.IncRequests)
-	r.Route(path, func(r chi.Router) {
-		HandlerFromMux(server, r)
-	})
+	handler := apiHandlers{
+		server: server,
+	}
+	RegisterHandlers(e.Group(path), &handler)
```
Gundersanne, line 75: You need to add the IncRequests middleware here again, in e.Group.

internal/cloudapi/server.go
```diff
@@ -37,6 +36,20 @@ type Server struct {
...
 
+const (
+	identityHeaderKey contextKey = iota
+)
+
+type identityHeader struct {
+	Identity struct {
+		AccountNumber string `json:"account_number"`
+	} `json:"identity"`
+}
```
Gundersanne, line 47: These can just stay in the VerifyIdentityHeader function, less diff noise that way.

Gundersanne (changes requested):
Thank you!

internal/cloudapi/server.go
chloenayon: Got it, thank you! I think I'll change this in a follow-up, at the moment I really want to get this PR in

internal/cloudapi/server.go
```diff
@@ -517,61 +483,47 @@ func composeStatusFromJobStatus(js *worker.JobStatus, result *worker.OSBuildJobR
...
-		return
+	if _, _, _, err = h.server.workers.Job(jobId, &job); err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, "Job %s not found: %s", id, err)
 	}
 
 	if status.Finished.IsZero() {
 		// job still running: empty response
-		if err := json.NewEncoder(w).Encode(new(ComposeMetadata)); err != nil {
+		if err := ctx.JSON(200, ComposeMetadata{}); err != nil {
 			panic("Failed to write response: " + err.Error())
```
Gundersanne, line 525: just `return ctx.JSON` here. If there's an error creating the response it'll generate a 500 error. The panic here can be removed, as it'll crash composer.

internal/cloudapi/server.go
```diff
@@ -639,7 +591,8 @@ func (server *Server) ComposeMetadata(w http.ResponseWriter, r *http.Request, id
 		resp.OstreeCommit = &commitMetadata.Compose.OSTreeCommit
 	}
 
-	if err := json.NewEncoder(w).Encode(resp); err != nil {
+	if err := ctx.JSON(200, resp); err != nil {
 		panic("Failed to write response: " + err.Error())
```
Gundersanne, line 595: Same, just 'return ctx.JSON' and remove the panic.

Gundersanne (changes requested):
Looking good, one final comment.

Gundersanne:
Ugh it seems I merged a bug a few days ago in the api tests, which is why they're failing.

internal/cloudapi/server.go
```diff
@@ -50,94 +53,87 @@ func NewServer(workers *worker.Server, rpmMetadata rpmmd.RPMMD, distros *distror
...
 		const identityHeaderKey contextKey = iota
 		type identityHeader struct {
 			Identity struct {
 				AccountNumber string `json:"account_number"`
 			} `json:"identity"`
 		}
-
-		idHeaderB64 := r.Header["X-Rh-Identity"]
+		idHeaderB64 := c.Request().Header.Get("X-Rh-Identity") //r.Header["X-Rh-Identity"]
 		if len(idHeaderB64) != 1 {
```
Gundersanne, line 79: Missed this one, since Header.Get() returns a string instead of a slice, len() != 1 won't work. `if idHeaderB64 == ""` => return error, otherwise continue.

internal/cloudapi/server.go
```diff
@@ -50,94 +53,87 @@ func NewServer(workers *worker.Server, rpmMetadata rpmmd.RPMMD, distros *distror
...
 			} `json:"identity"`
 		}
-
-		idHeaderB64 := r.Header["X-Rh-Identity"]
-		if len(idHeaderB64) != 1 {
-			http.Error(w, "Auth header is not present", http.StatusNotFound)
-			return
+		idHeaderB64 := c.Request().Header.Get("X-Rh-Identity")
+		if idHeaderB64 == "" {
+			return echo.NewHTTPError(http.StatusNotFound, "Auth header is not present")
```
ondrejbudai, line 80: Would be great to rework these statuses in a follow-up. For example, this is imho a good example of 401 Unauthorized, see https://http.cat/401 (and https://http.cat ) for more info.

internal/cloudapi/server.go
```diff
@@ -517,61 +483,45 @@ func composeStatusFromJobStatus(js *worker.JobStatus, result *worker.OSBuildJobR
...
+		return echo.NewHTTPError(http.StatusNotFound, "Job %s not found: %s", id, err)
 	}
 
 	if status.Finished.IsZero() {
 		// job still running: empty response
-		if err := json.NewEncoder(w).Encode(new(ComposeMetadata)); err != nil {
-			panic("Failed to write response: " + err.Error())
-		}
-		return
+		return ctx.JSON(200, ComposeMetadata{})
```
ondrejbudai, line 524: I suggest using `http.StatusOK` instead of `200`.

internal/cloudapi/server.go
```diff
@@ -639,7 +589,5 @@ func (server *Server) ComposeMetadata(w http.ResponseWriter, r *http.Request, id
 		resp.OstreeCommit = &commitMetadata.Compose.OSTreeCommit
 	}
 
-	if err := json.NewEncoder(w).Encode(resp); err != nil {
-		panic("Failed to write response: " + err.Error())
-	}
+	return ctx.JSON(200, resp)
```
ondrejbudai, line 592: I suggest using `http.StatusOK` instead of `200`.

ondrejbudai (commented):
This looks great! Two minor nitpicks and one general thought inline. :)

(I restarted two tests that failed for unrelated reasons.)

internal/cloudapi/server.go
Gundersanne: This is quirky I know, but CRC api guidelines require unauthorized to result in 404.

Gundersanne (approved):

Issue #1464, opened 2021-06-10
Image tests: Fedora IoT tests and  RHEL 8 edge+tar image tests are failing


jrusz:
It seems that after #1431 was merged, some of the image tests started failing. 
Namely `fedora_33-x86_64-fedora_iot_commit-boot.json` and `fedora_33-x86_64-fedora_iot_commit_debug-boot.json` which can be seen here:  https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1334367653 

It's possible the same goes for other Fedora distros but I haven't checked them.

I can see more failures for RHEL-8 here: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1334367654
Those failures are namely: `rhel_8-x86_64-rhel_edge_commit-boot.json` `rhel_8-x86_64-rhel_edge_commit_rt-boot.json` and `rhel_8-x86_64-tar-boot.json`

The RHEL failures all seem to be the same and related to SElinux? I think it's the same for Fedora but the error diff is much longer...

thozza:
So the issue is that for ostree-based images, the `image-info` bind-mounts `/usr/etc` to `/etc`. `setfiles` founds out that there is an association between files in `/usr/etc` and `/etc` and also finds conflicting specifications for these locations. The choice of the specification is not deterministic, thus can randomly produce different results at any time. The content of `/etc` in this case will always have incorrect SELinux labels for the location.

The ideal solution would be to analyze the SELinux ctx mismatches without the bind-mounted `/etc`. However for our use case, I propose much simpler fix/workaround - exclude `/etc` from the SELinux ctx mismatch analysis of ostree-based images. (FYI @gicmo @achilleas-k).

My proposed fix would be https://github.com/thozza/osbuild-composer/commit/8252153d5150b3d1caa93e22bf71d70ea4b240d4, which is currently being tested in https://github.com/osbuild/osbuild-composer/pull/1468

thozza:
Update: I think that I found cause of the remaining differences in the reported `image-info` output in the CI and when regenerating image test case in a VM. The reason seems to be that some labels are not known to the host when the image test case is generated inside a VM, while they are known to the host when running the image test cases in CI. There is at least the difference, that in the CI environment, the package `osbuild-composer-tests` is installed on the system, which pulls in additional packages, some of them shipping SELinux policy. Example is `container-selinux`. I'm now verifying various scenarios and will be making additional changes to the generator script and regenerating relevant test cases.

[...]

PR #1468, opened 2021-06-11
[TEST ONLY]Image tests update

thozza:
Testing image-info SELinux ctx fix/workaround for #1464


atodorov:
@thozza I still see failures of the test jobs that you left uncommented. What's the status here? 

Reminder that this issue is blocking a chain of about 5 PRs+tickets so we should fix it as soon as possible.

thozza:
> @thozza I still see failures of the test jobs that you left uncommented. What's the status here?

The status is that on Fri I regenerated the ostree-based images in VMs with the changed `image-info`, but I'm getting different results than what is produced in the CI. I need to investigate the issue further and determine what is the difference in the environment what am I missing.

> Reminder that this issue is blocking a chain of about 5 PRs+tickets so we should fix it as soon as possible.

Yes, I'm aware of it. Unfortunately regenerating image test cases takes quite a lot of time, as well as CI runs, so there is quite a big workaround.

thozza:
I hopefully resolved all the issues and regenerated test cases. If this PR passes in CI, I'll close it and open a new one with just my changes...

thozza:
There was still a failure on `rhel-8`, but it is a different issue. I suspect that the way `setfiles` is used by `image-info` on Fedora does not work on RHEL. Mainly due to logs like this:
```
🏃🏻 Running test: rhel_8-x86_64-rhel_edge_commit-boot.json
=== RUN   TestImages
=== RUN   TestImages/rhel_8-x86_64-rhel_edge_commit-boot.json
=== RUN   TestImages/rhel_8-x86_64-rhel_edge_commit-boot.json/image_info
usage:  setfiles [-diIDlmnpqvFW] [-e excludedir] [-r alt_root_path] spec_file pathname...
usage:  setfiles [-diIDlmnpqvFW] [-e excludedir] [-r alt_root_path] spec_file -f filename
usage:  setfiles -s [-diIDlmnpqvFW] spec_file
usage:  setfiles -c policyfile spec_file
ostree config get core.mode --repo=/var/tmp/tmpn_5_i9dg/root/repo
ostree refs --repo=/var/tmp/tmpn_5_i9dg/root/repo
ostree rev-parse rhel/8/x86_64/edge --repo=/var/tmp/tmpn_5_i9dg/root/repo
ostree checkout --force-copy efe76c8dce80b6aed9b7f5678324043c8c39b960f8384cba15c8ea3da26f012c /var/tmp/tmpjj2mj2pf/tree --repo=/var/tmp/tmpn_5_i9dg/root/repo
    main_test.go:106: 
...
```

I'm currently investigating this further...

thozza:
The issue is that usage of `setfiles` on RHEL-8 is different than on Fedora and in the upstream and unfortunately it does not allow passing path to binary policy using `-c`. This is completely crucial for our use case, because we need to check the SELinux context of the files in the image using the policy from the image. I filled https://bugzilla.redhat.com/show_bug.cgi?id=1973754, but this won't help with `rhel-8` and `rhel-84` testing. I have a few ideas that I could work, but I need to verify some of them first:

1. For the time when `image-info` runs `setfiles` we could bind-mount the policy file from the image to the host. `setfiles` should be consistently using a specific binary policy file present on the host, so this may workaround the limitation. The policy file should be unmounted right after the check is finished. However I have no idea if this may cause any issues on the host system.
2. Chroot to the image filesystem tree and run `setfiles` from the image (this would actually require `policycoreutils` RPM to be installed in the image, which may not be the case all the time and desired at all).
3. Patch `setfiles` for the purpose of testing RHEL-8 images in our CI.

If anyone has any other ideas, please don't hesitate to bring them up...

atodorov:
> this won't help with `rhel-8` and `rhel-84` testing. 

FTR `rhel-84` will become 8.5 very soon so we can start testing with an updated RPM including the backport you requested in Bugzilla. 

`rhel8` will become the 8.4 GA for which we still need a workaround (or request a z-stream update ?)

thozza:
@atodorov I added a new CLI option to `osbuild-image-tests` to ignore the `selinux/context-mismatch` part of the image-info report and enabled it for RHEL-8. Based on my preliminary testing, it works nicely as a workaround for image tests on RHEL-8.

thozza:
Image tests passed. I created #1485 with just my changes.

PR #1471, opened 2021-06-14
deploy-qemu: rewrite in python

msehnout:
The discussion about Bash syntax have reached the threshold where it is
necessary to rewrite the script in Python to make everyone's life
easier.

This pull request includes:

### DIFF
--- a/tools/deploy-qemu
+++ b/tools/deploy-qemu
@@ -1,73 +1,73 @@
-#!/bin/bash
+#!/usr/bin/env python3
 
-#
-# deploy-qemu IMAGE USERDATA
-#
-# Starts an ephemeral virtual machine in qemu, injecting configuration via
-# cloud-init. Stopping this script stops the VM and discards all data.
-#
-# IMAGE    -- An os image that can be booted by qemu and has cloud-init
-#             installed and enabled. No changes are made to this file.
-#
-# USERDATA -- A cloud-init user-data config file, or a directory of
-#             configuration as accepted by the `gen-user-data` tool.
-#
-# In addition, if the QEMU_EXTRA_ARGS environment variable is defined, it adds
-# its content as additional arguments to qemu.
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
 
-set -euo pipefail
 
-if [[ "$#" != 2 ]]; then
-  echo "usage: $0 IMAGE USERDATA"
-  exit 1
-fi
+HELP_TEXT="""deploy-qemu IMAGE USERDATA
 
-scriptdir=$(dirname "$0")
-image=$1
-userdata=$2
-read -ra qemu_extra_args <<< "${QEMU_EXTRA_ARGS:-}"
-workdir=$(mktemp -d "$scriptdir/qemu-tmp-XXXXXX")
-function cleanup() {
-  rm -rf "$workdir"
-}
-trap cleanup EXIT
+Starts an ephemeral virtual machine in qemu, injecting configuration via
+cloud-init. Stopping this script stops the VM and discards all data.
 
-mkdir "$workdir/cidata"
+IMAGE    -- An os image that can be booted by qemu and has cloud-init
+            installed and enabled. No changes are made to this file.
 
-if [ -d "$userdata" ]; then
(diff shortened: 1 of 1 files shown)


tools/deploy-qemu.py
```diff
@@ -0,0 +1,71 @@
...
+if len(sys.argv) != 3:
+    print(HELP_TEXT)
+    exit(1)
+
+script_dir = os.path.dirname(os.path.realpath(__file__))
+image = sys.argv[1]
+userdata = sys.argv[2]
+
+qemu_extra_args = os.getenv("QEMU_EXTRA_ARGS")
+qemu_extra_args = qemu_extra_args.split(' ') if qemu_extra_args is not None else []
```
larskarlitski, line 32: I think you can drop this line and modify the previous one to:

```python
qemu_extra_args = os.getenv("QEMU_EXTRA_ARGS", [])
```

tools/deploy-qemu.py
```diff
@@ -0,0 +1,71 @@
...
+
+    subprocess.run(["qemu-system-x86_64",
+        "-M", "accel=kvm:hvf",
+        "-m", "1024",
+        "-snapshot",
+        "-cpu", "host",
+        "-net", "nic,model=virtio",
+        "-net", "user,hostfwd=tcp::2222-:22,hostfwd=tcp::4430-:443",
+        "-cdrom", f"{workdir}/cloudinit.iso"] +
+        qemu_extra_args +
```
larskarlitski, line 70: I prefer think I'd have preferred the inline `*` syntax here, but it's a matter of taste:
```python3
["qemu-system-x86_64", ..., *qemu_extra_args, image]
```

tools/deploy-qemu.py
```diff
@@ -0,0 +1,71 @@
+import os
```
larskarlitski, line 1: A shebang would be nice.
```
#!/usr/bin/env python3
```

larskarlitski (changes requested):
Very nice! Less bash is always good. Thanks ❤️ 

Works as advertised, but please drop the `.py` extension so that the documentation stays up to date and people (ehm, me) aren't confused :)

tools/deploy-qemu.py
msehnout: It actually is more complicated. If you simply read the environment variable you could get sth like "-smp 2 -m 2048", but you need to split it to make it a list.

I've considered these two options, but rejected both:
 * This returns either a string or a list, so the processing will be difficult: `os.getenv("QEMU_EXTRA_ARGS", [])`
 * `os.getenv("QEMU_EXTRA_ARGS", "").split(' ')` returns `[""]` when QEMU_EXTRA_ARGS is empty

Eventually I decided to go with the solution in this patch.

tools/deploy-qemu.py
msehnout: Much better, I forgot this syntax exists 👍

tools/deploy-qemu.py
msehnout: Sure, but which one?!

FTR: we agreed `/usr/bin/env` is good enough for this kind of script.

tools/deploy-qemu.py
larskarlitski: Oh dear, didn't read it properly. Thanks, makes sense!

larskarlitski: Technically, we'd probably have to use `shlex` here, but we didn't do this in the bash version either.

larskarlitski (approved):

tools/deploy-qemu.py
msehnout: Oh wow, I had no idea `shlex` exists. But I hope it is just fine like this.

ondrejbudai (approved):
Tested on Fedora 34, works like a charm, thanks a lot for doing this!

msehnout:
Thanks for the reviews 👍 . CI haven't finished even though it has been more than 12 hours, so I'm going to merge the PR despite it being yellow. In doesn't matter anyway because the deploy-qemu script is not executed anywhere in the CI pipeline.

PR #1485, opened 2021-06-21
`image-info` changes related to reading SELinux labels unknown to the host

### LINKED ISSUE #1464: Image tests: Fedora IoT tests and  RHEL 8 edge+tar image tests are failing

thozza:
FIxes #1464

When `image-info` inspects ostree images, the `/usr/etc` is bind-mounted to `/etc`. This results in conflicting SELinux policy specification for these files and makes the outcome dependent on the `setfiles` build. All the files in `/etc` have mismatch in the expected vs. actual SELinux context.

Exclude `/etc` from the check of SELinux ctx mismatches in case the analysed tree is from an ostree-based image.

Sort the list returned `read_selinux_ctx_mismatch()` based on the item's `filename` key, to make the result consistent across runs.

`image-info` can not read SELinux labels from the images, which are not known to the host. This makes the report content depend on the host environment. As a temporary workaround, relabel the image-info script with osbuild_exec_t label to allow it to read unknown SELinux labels.

Modify documentation in `test/README.md` to explain the issue with `image-info` and unknown SELinux labels.

Modify the `generate-all-test-cases` to relabel `image-info` before generating test cases.

Modify the `image_tests.sh` to relabel `image-info` before running image test cases.

Add 'tar' image for 'rhel-8' on 's390x' back to the matrix of generated test cases, as it was removed by mistake. Regenerate the image test case. Remove 'tar' image from 'rhel-84' on 's390x' from the matrix of generated test cases, as it is not supported.

Regenerate all affected image test cases.

Add a new CLI option to `osbuild-image-tests` called `-skip-selinux-ctx-check` to workaround the limitation of `setfiles` on RHEL-8 [1]. If the option is passed to the binary, then the 'selinux/context-mismatch' part is removed from the "expected" and "actual" image-info report, before these two reports are compared.

Modify `image_tests.sh` to run `osbuild-image-tests` with `-skip-selinux-ctx-check` when run on RHEL-8.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1973754

This pull request includes:


ondrejbudai (approved):
Splendid!

teg (approved):

atodorov (approved):

atodorov:
CI reports 2 failures:

1) Is openstack failing to provision the system, which is a known infrastructure issue ATM
2) GCP ERROR: (gcloud.compute.instances.delete) You do not currently have an active account selected. which I don't think is related to the current PR. 

Everything else is PASSing so I'm going ahead and merging this.

thozza:
> CI reports 2 failures:
> 
> 1. Is openstack failing to provision the system, which is a known infrastructure issue ATM
> 2. GCP ERROR: (gcloud.compute.instances.delete) You do not currently have an active account selected. which I don't think is related to the current PR.
> 
> Everything else is PASSing so I'm going ahead and merging this.

The GCP error is actually hidden in the journal log, because as you can see in the job log, the compose actually failed. The reason is that the GCP zone that is used by the test has exhausted resources. I created a workaround, but you were faster with the merge 😄 .  Thus I created https://github.com/osbuild/osbuild-composer/pull/1489.

Thanks to all for a patience with this issue!

PR #1488, opened 2021-06-22
random fixes for tools/deploy-qemu

ondrejbudai:
Two fixes for `tools/deploy-qemu`:
- Previously, the temporary directory was created in the script dir. It just
imo created unnecessary and confusing files in my git checkout. Let's just
use the default temporary directory because we don't really have special
needs for it.
- exit is just a helper for the interactive shell, see:
  https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

### DIFF
--- a/tools/deploy-qemu
+++ b/tools/deploy-qemu
@@ -24,7 +24,7 @@ its content as additional arguments to qemu."""
 
 if len(sys.argv) != 3:
     print(HELP_TEXT)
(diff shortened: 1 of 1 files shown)


msehnout (approved):
Looks good to me! I'm going to merge this right away as it is not executed in our CI anyway.

PR #1489, opened 2021-06-22
Cloud API test: randomize used GCP zone from the region

thozza:
The `api.sh` test currently always defaults to `<REGION>-a` zone when creating instance using the built image. The resources in a zone may get exhausted and the solution is to use a different zone. Currently even a CI job retry won't help with mitigation of such error during a CI run.

Modify `api.sh` to pick random GCP zone for a given region when creating a compute instance. Use only GCP zones which are "UP".

The `cloud-cleaner` relied on the behavior of `api.sh` to always choose the "<REGION>-a" zone. Guessing the chosen zone in `cloud-cleaner` is not viable, but thankfully the instance name is by default unique for the whole GCP project. Modify `cloud-cleaner` to iterate over all available zones in the used region and try to delete the specific instance in each of them.

Make `ComputeZonesInRegion` method from the `internal/cloud/gcp` package exported and use it in `cloud-cleaner` for getting the list of available zones in a region.

Fix generation of predictable `TEST_ID` by `api.sh` test case in GitLab CI.

This pull request includes:

### DIFF
--- a/cmd/cloud-cleaner/main.go
+++ b/cmd/cloud-cleaner/main.go
@@ -27,8 +27,6 @@ func cleanupGCP(testID string, wg *sync.WaitGroup) {
 		log.Println("[GCP] Error: 'GCP_REGION' is not set in the environment.")
 		return
 	}
-	// api.sh test uses '--zone="$GCP_REGION-a"'
-	GCPZone := fmt.Sprintf("%s-a", GCPRegion)
 	GCPBucket, ok := os.LookupEnv("GCP_BUCKET")
 	if !ok {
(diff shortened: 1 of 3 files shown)


thozza:
I just realised that this will break `cloud-cleaner` ability to clean up GCP, because it also defaults to `<REGION>-a`. So the change needs to be reflected there as well.

thozza:
The fact that `cloud-cleaner` can successfully clean up after the `api.sh` script has been tested in https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1426944347

ondrejbudai (approved):

PR #1490, opened 2021-06-22
Cloud API: compose/id/metadata

achilleas-k:
This pull request includes:

_Followup to #1439_

Added a new Cloud API endpoint that returns metadata about a finished build.  Currently the metadata contains full package list (NEVRA) for the built image and, where available, the commit ID (hash) for the created Edge commit.

## Internal changes
- Utility functions for extracting and structuring the RPM metadata was moved to an internal package to be reused.
- Added a small utility function that reads the schema version from a raw Manifest.
- Metadata returned from stages was previously only stored for the RPM stage.  Now all metadata from all stages is stored.  Metadata from the `org.osbuild.ostree.commit` stage are also structured.

See individual commit messages for more details.

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -19,6 +19,7 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/distro"
 	"github.com/osbuild/osbuild-composer/internal/distroregistry"
+	"github.com/osbuild/osbuild-composer/internal/osbuild1"
 	"github.com/osbuild/osbuild-composer/internal/ostree"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/target"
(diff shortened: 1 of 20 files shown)


teg (dismissed):
Love it!

internal/distro/distro_test.go
```diff
@@ -16,3 +23,41 @@ func TestDistro_Manifest(t *testing.T) {
 		distroregistry.NewDefault(),
 	)
 }
+
+func TestDistro_Version(t *testing.T) {
+	pipelinePath := "../../test/data/manifests/"
```
ondrejbudai, line 28: I wonder if we want to base this test on the actual testing manifests. They change quite frequently and with the migration of 8.5 and 9.0 manifests to v2, this test will probably need a rewrite. Maybe we want to consider using a fixed mock manifest inlined in Go? I think we already have several of those.

achilleas-k: Yeah, that's definitely a better idea.

internal/cloudapi/server.go
```diff
@@ -521,3 +522,87 @@ func (server *Server) GetVersion(w http.ResponseWriter, r *http.Request) {
...
+
+	var rpms []rpmmd.RPM
+	if status.Finished.IsZero() {
+		// job still running: empty response
+		if err := json.NewEncoder(w).Encode(new(ComposeMetadata)); err != nil {
+			panic("Failed to write response: " + err.Error())
+		}
+		return
+	}
+	rpms = rpmmd.OSBuildStagesToRPMs(result.OSBuildOutput.Build.Stages)
```
achilleas-k, line 555: This is wrong.  It's not the PRMs for the build root that we want but the RPMs for the payload.

achilleas-k: I should also add a test for this.  A check to verify that a non-standard package specified in the request is included in the metadata.

achilleas-k: Fixed.

internal/distro/distro_test.go
achilleas-k: Done!

achilleas-k:
Changed the result parsing so that it now supports collecting metadata from v2 manifest results as well.  It gets a little bit messy, but I'd like to clean it up in a separate PR where we switch to using v2 results internally (see #1493).

I tested the change against #1351 locally and I'll rebase that onto this PR once it's merged.  I also added a verification in the `api.sh` script to check if `postgresql` (specified in the request) is included in the package list in the metadata.

teg (dismissed):

achilleas-k:
Don't believe GitLab CI report and click through instead.  Not sure why it's showing green, it hasn't properly started yet.

internal/distro/distro_test.go
```diff
@@ -16,3 +18,81 @@ func TestDistro_Manifest(t *testing.T) {
...
+)
+
+func TestDistro_Version(t *testing.T) {
+	assert := assert.New(t)
+	expectedVersion := "1"
+	for idx, rawManifest := range v1manifests {
+		manifest := distro.Manifest(rawManifest)
+		detectedVersion, err := manifest.Version()
+		assert.NoError(err, "Could not detect Manifest version for %d: %v", idx, err)
+		assert.Equal(expectedVersion, detectedVersion, "in manifest %d", idx)
```
atodorov, line 76: IMO in this case it would be better to use require.NoError (fails fast) instead of assert. Other than than the assertion functions are the same.

atodorov: FTR in general with Go I prefer the `require` package over the `assert` one but I see we've used both of them.

achilleas-k: I prefer letting it run through the rest of the tests even if one fails (where possible) since they're mostly independent and it's good to know if one or all of them are failing.  But it's not that important here.  Changed it to require.

achilleas-k:
I added a package verification step to all tests in `api.sh` but missed the fact that the Azure tests don't define any custom packages.

teg (approved):
Still great!

PR #1511, opened 2021-06-29
tests: make azure.sh easier to run locally

jrusz:
This package needs to be present otherwise the VM is not
responding. This change makes sure the package is present
in the image.

This pull request includes:

### DIFF
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -3,8 +3,8 @@ set -euo pipefail
 
 source /etc/os-release
 DISTRO_CODE="${DISTRO_CODE:-${ID}_${VERSION_ID//./}}"
-BRANCH_NAME="${BRANCH_NAME:-${CI_COMMIT_BRANCH}}"
-BUILD_ID="${BUILD_ID:-${CI_BUILD_ID}}"
+BRANCH_NAME="${CI_COMMIT_BRANCH:-'local'}"
+BUILD_ID="${CI_BUILD_ID:-$(uuidgen)}"
 HYPER_V_GEN="${HYPER_V_GEN:-V1}"
 
 # Colorful output.


atodorov (approved):
lgtm

ondrejbudai:
Why did this break now? Do we know what changed?

jrusz:
> Why did this break now? Do we know what changed?

Not yet, they probably removed dhcp-client from some group. Yuxin created rhbz#1977385 I hope we'll get the answer there.

I also added another small commit which makes it easier to run the test locally.

ondrejbudai:
Hmm, I'm confused, sorry. What exactly does this PR solve?

jrusz:
> Hmm, I'm confused, sorry. What exactly does this PR solve?

For one it makes it easier to run locally by adding defaults for some CI variables and then it makes sure `dhcp-client` package is present in the vhd image that gets build. There was an issue with it that Yuxin found, not sure what the state is now though, there's a BZ reference in the commit message...

teg:
Sounds to me that if the client is needed then it should be in the image definitions?

ondrejbudai:
Ah, I see two issues then: 

1) We don't run 8.5 tests in the CI yet so it doesn't make much sense to include this workaround just yet.
2) It would be better to add the package conditionally (on RHEL 8.5 only).

I think we need to ask `cloud-init` maintainers if they are going to change its deps or if we need to fix it in the image definitions.

achilleas-k:
For reference, I added it to the default packages for vhd in the RHEL 8.5 PR:
https://github.com/osbuild/osbuild-composer/pull/1536/files#diff-fe9ae752df6d857d5dfebfce62ea791d4951c30276264e0c84653a5501f69791R136

jrusz:
> Ah, I see two issues then:
> 
> 1. We don't run 8.5 tests in the CI yet so it doesn't make much sense to include this workaround just yet.

Well it won't hurt either, but I'll agree with whatever you think is better :)

> 2. It would be better to add the package conditionally (on RHEL 8.5 only).

You can see the blueprint also has `cloud-init` package added there, that's only because of Fedora, because the default image definition does not have it, only RHEL does (and I guess centos). So we already have a similar "workaround" in place.

ondrejbudai (dismissed):
I asked on both bzs if they're going to add the dependency on `dhcp-client` in `cloud-init`. If they aren't, we need to change the image definitions for 8.5 and 9.0 beta. If they are, we don't need this fix.

I'm fine with merging the second commit though. :)

teg (approved):

PR #1513, opened 2021-06-30
tests: Run some libvirt tests in openstack

jrusz:
This is just me trying to run just libvirt tests on Openstack.

There are two reasons for this:
1) TCG is not supported downstream meaning we can't expect booting qcow images without kvm to work.
2) We don't want to move all integration testing to Openstack as it's severely slowing down the CI as can be seen now when we moved Centos stream 8 integration testing.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -153,7 +153,6 @@ New OSTree:
 .INTEGRATION_TESTS: &INTEGRATION_TESTS
   SCRIPT:
     - koji.sh
-    - libvirt.sh
     - aws.sh
     - azure.sh
     - vmware.sh
@@ -174,14 +173,14 @@ Integration:
           - aws/fedora-33-x86_64
           # See COMPOSER-919
           # - aws/fedora-34-x86_64
-          - openstack/centos-stream-8-x86_64
+          - aws/centos-stream-8-x86_64
       - <<: *INTEGRATION_TESTS
         RUNNER:
           - aws/rhel-8-x86_64
           - aws/rhel-8.4-x86_64
         INTERNAL_NETWORK: ["true"]
       - SCRIPT:
-         - azure_hyperv_gen2.sh
+          - azure_hyperv_gen2.sh
         RUNNER:
           - aws/rhel-8.4-x86_64
         INTERNAL_NETWORK: ["true"]
@@ -190,6 +189,17 @@ Integration:
           - aws/rhel-8-x86_64
         INTERNAL_NETWORK: ["true"]
         DISTRO_CODE: ["rhel_90"]
+      - SCRIPT:
+          - libvirt.sh
+        RUNNER:
+          - aws/fedora-33-x86_64
+          - openstack/centos-stream-8-x86_64
+      - SCRIPT:
+          - libvirt.sh
+        RUNNER:
+          - aws/rhel-8-x86_64
+          - aws/rhel-8.4-x86_64
+        INTERNAL_NETWORK: ["true"]
 
 .API_TESTS: &API_TESTS
   TARGET:


jrusz:
Ok this might look horrible but actually works and I don't know how to better do it with what `.gitlab-ci.yml` syntax allows. Basically what this does now is it runs all the integration tests except `libvirt.sh` in combination with specific runners which is Centos-stream-8 now and RHEL-8.5 will need to be added in the future. It then runs the `libvirt.sh` test but changes the runner to openstack. I've also added a rule for scheduled `libvirt.sh` test because that's already consuming RHEL-8.5 content where it's failing. I made a separate block with the rules to not mess up the job stage too much. 

I've also been thinking about creating some sort of "common.yml" file for gitlab-ci where we would put such things and just include it to this main `.gitlab-ci.yml` file for better readability, but I'm not sure it's necessary now.

jrusz:
@atodorov @ondrejbudai Could you share your thoughts please?

.gitlab-ci.yml
```diff
@@ -150,6 +150,18 @@ New OSTree:
...
+.INTEGRATION_RULES: &INTEGRATION_RULES
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SCRIPT != "libvirt.sh" || ($CI_PIPELINE_SOURCE != "schedule" && $SCRIPT == "libvirt.sh" && $RUNNER != "aws/centos-stream-8-x86_64")'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SCRIPT == "libvirt.sh" && $RUNNER == "aws/centos-stream-8-x86_64"'
+      variables:
+        RUNNER: "openstack/centos-stream-8-x86_64"
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $SCRIPT == "libvirt.sh" && $RUNNER == "aws/rhel-8.4-x86_64"'
+      variables:
+        RUNNER: "openstack/rhel-8.4-x86_64"
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[\S]+/ && $SCRIPT != "libvirt.sh"'
```
atodorov, line 163: Why are these rules so complicated ? What are you trying to achieve here ?

Instead of using all of this complexity why not simply create another stage/matrix definition for the libvirt test jobs and define only the runners which are known to work?

atodorov (changes requested):

.gitlab-ci.yml
jrusz: You're right that is a lot simpler I just didn't want to make the matrix longer or duplicate some runners there so I was looking for a different way to achieve what I wanted which is run the libvirt test on openstack only in case of centos-stream-8 and rhel-8.5

jrusz:
So it seems we're not yet supporting building qcow images on rhel-8.5 so I changed it back to use 8.4 on aws which is passing in regular pipeline but will fail for nightly pipeline because that's consuming 8.5 content... 

I would just leave it at that for now and when we have 8.5 support just switch the runner to openstack.

.gitlab-ci.yml
```diff
@@ -174,14 +173,14 @@ Integration:
...
-          - openstack/centos-stream-8-x86_64
+          - aws/centos-stream-8-x86_64
       - <<: *INTEGRATION_TESTS
         RUNNER:
           - aws/rhel-8-x86_64
           - aws/rhel-8.4-x86_64
         INTERNAL_NETWORK: ["true"]
       - SCRIPT:
-         - azure_hyperv_gen2.sh
+          - azure_hyperv_gen2.sh
```
atodorov, line 183: accidental white space

atodorov (changes requested):
minor change requested.

atodorov (approved):

ondrejbudai (approved):

Issue #1461, opened 2021-06-10
RHEL 9 testing is wasteful


ondrejbudai:
We run only integration tests for RHEL 9 (these are running on RHEL 8 GA though). It doesn't make sense to run some of them though:

- `koji.sh` -indeed builds a RHEL 9 image
- `libvirt.sh` - actually builds a RHEL 8 image. Thus it doesn't make much sense.
- `azure.sh` - the test is terminated immediately after provisioning the machine.
- `api.sh`- the test is terminated immediately after provisioning the machine. This happens three times in one pipeline.
- `vmware.sh`- actually builds a RHEL 8 image. Thus it doesn't make much sense.

PR #1514, opened 2021-06-30
distro/rhel90: add s390utils-core to the build pipeline for s390x

ondrejbudai:
A recent update to crypto-policies changed its requirements from:

Recommends: grubby

to:

Recommends: (grubby if kernel)

We don't install kernel in the build pipeline, thus grubby was now no longer
installed. This caused also s390utils-core to not be installed on s390x.
s390utils-core has to be in the build pipeline though because we use
/usr/sbin/zipl from it to install the bootloader to s390x images.

Long story short, images for s390x currently cannot be build because
/usr/sbin/zipl is no longer in the build pipeline. This commit fixes that
by explicitly adding s390utils-core to the package list for s390x.

I verified this fix on an actual s390x machine. The test manifest was
regenerated but it's actually still the same because we use an older
snapshot that doesn't contain the newest crypto-policies change.

Kudos to Tomáš Hozza for helping me out!

This pull request includes:

### DIFF
--- a/internal/distro/rhel90/packages.go
+++ b/internal/distro/rhel90/packages.go
@@ -64,7 +64,9 @@ var packages = struct {
 			"grub2-ppc64le",
 			"grub2-ppc64le-modules",
 		},
-		s390x: nil,
+		s390x: []string{
+			"s390utils-core",
+		},
 	},
 	Qcow2: rpmmd.PackageSet{
 		Include: []string{


thozza (approved):
LGTM!

ondrejbudai:
The koji test succeeded, this is the only relevant one (see https://github.com/osbuild/osbuild-composer/issues/1461 ). Let's merge this so we can unblock composes.

PR #1516, opened 2021-07-01
CI: limit testing for RHEL 8.5

achilleas-k:
This PR is meant to limit the number of tests run for RHEL 8.5 to just the features and image types that are supported.

- Disable base tests on RHEL 8.5
The unit tests should be passing here but there others depend on image types that aren't supported.

This means that there will be untested features for RHEL 8.5, but that should be fixed soon with the main image types.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -107,7 +107,8 @@ Base:
           - aws/centos-stream-8-aarch64
       - RUNNER:
           - aws/rhel-8.4-x86_64
-          - aws/rhel-8.5-x86_64
+          # main image types not yet supported in 8.5
+          # - aws/rhel-8.5-x86_64
           # see COMPOSER-918
           # - aws/rhel-8.4-aarch64
         INTERNAL_NETWORK: ["true"]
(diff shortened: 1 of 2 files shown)


teg (dismissed):
Usually we would have enabled just the tests that are expected to pass to verify the new functionality, but this slipped in a bit early causing `main` to now fail CI. Let's merge this to get back in 🟢 and then reenable all tests as soon as we can.

atodorov (dismissed):

ondrejbudai (dismissed):

achilleas-k:
CI is currently stuck on the virt-install step.  Likely something wrong with the kickstart file, since the ISO contains one already for the live image and injecting one with virt-install probably doesn't override (though even if it does, it would remove the live img command)

ondrejbudai:
Should we just disable the test for now and fix it in a follow-up PR?

achilleas-k:
> Should we just disable the test for now and fix it in a follow-up PR?

My thoughts exactly.  Disabled.

ondrejbudai (approved):

atodorov:
> Should we just disable the test for now and fix it in a follow-up PR?

Can we pair this proposal with a QE ticket in Jira with the respective priority please?

PR #1522, opened 2021-07-02
docs: document the release process

ondrejbudai:
This commit adds RELEASING.md guide describing the process of making a new
upstream release and pushing it into Fedora and CentOS Stream 9.

Additionally, a new helper is added to the repository - update-distgit.py.
The purpose of this helper is to simplify work with the dist-git.

Random thoughts:
I decided to left off RHEL 8 because the guide would be full of internal
URLs and tools. I will probably write a similar guide for it and put it
into internal guides.

I decided to just reference the RHEL Developer Guide for CentOS Stream 9.
It's pretty well written and I didn't feel like duplicating the effort.

We should definitely switch to PR-based approach for Fedora and implement
at least some smoke tests. I believe that the current guide is good enough
and we can iterate later.

**One final note:** @gicmo, I requested a review from you because you're
usually the one releasing osbuild and the process is very similar so I think
it makes sense to cooperate. :) Others are of course encouraged to review
as well. ;)

This pull request includes:

- [x] adequate testing for the new functionality or fixed issue - not needed I think

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,166 @@
+# Making a new release
+
+This guide describes the process of releasing osbuild-composer both [upstream][upstream-git] and into [Fedora][fedora-distgit] and [CentOS Stream][centos-distgit].
+
+> This guide assumes that the number of the new release is stored in an environment variable called `VERSION`.
+
+## Upstream release
+
+Osbuild-composer release is a tagged commit that's merged into the `main` branch as any other contribution.
+
+### Making a release PR
+
+Firstly, a new branch from (up-to-date) main needs to be created:
+
+```
+git checkout main
+git pull
+git checkout -b release-version-$VERSION
+```
+
+All unreleased news have to copied over to a new directory:
+
+```
+mkdir docs/news/$VERSION
+mv docs/news/unreleased/* docs/news/$VERSION
+```
+
+Now, the `NEWS.md` file must be updated. A template for a new release can be generated using `make release`:
+
+```
+make release
+your-favourite-editor NEWS.md
+```
+
+The next step is to update the version number in the spec file. You can use `your-favourite-editor`, or this lovely `sed` script:
+
+```
+sed -i -E "s/(Version:\\s+)[0-9]+/\1$VERSION/" osbuild-composer.spec
+```
+
+At this moment, everything in the checkout is ready for making the release commit:
+
+```
+git add osbuild-composer.spec NEWS.md docs/news/unreleased docs/news/$VERSION
+git commit -s -m $VERSION -m "Release osbuild-composer $VERSION"
+```
+
+As the final step, push the branch into your fork and open a PR against the `main` branch. The PR now needs to be approved and merged.
+
+### After the release PR is merged
+
+Firstly, make sure to update the local checkout of the `main` branch:
+
+```
+git checkout main
+git pull
+```
+
+Tag the release and push the tag:
+
+```
+git tag -s -m 'osbuild-composer $VERSION' v$VERSION HEAD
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -0,0 +1,166 @@
+# Making a new release
+
+This guide describes the process of releasing osbuild-composer both [upstream][upstream-git] and into [Fedora][fedora-distgit] and [CentOS Stream][centos-distgit].
+
+> This guide assumes that the number of the new release is stored in an environment variable called `VERSION`.
+
+## Upstream release
+
+Osbuild-composer release is a tagged commit that's merged into the `main` branch as any other contributions.
```
achilleas-k, line 9:
```suggestion
Osbuild-composer release is a tagged commit that's merged into the `main` branch as any other contribution.
```

tools/update-distgit.py
```diff
@@ -0,0 +1,124 @@
...
+
+    commit_message = f"Update to {version}"
+    subprocess.check_call(["git", "commit", "-m", commit_message])
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+    parser.add_argument("--version", metavar="VERSION", type=str, help="version to be released to downstream",
+                        required=True)
+    parser.add_argument("--author", metavar="VERSION", type=str,
```
achilleas-k, line 108:
```suggestion
    parser.add_argument("--author", metavar="AUTHOR", type=str,
```

tools/update-distgit.py
```diff
@@ -0,0 +1,124 @@
...
+    downstream_spec_lines = downstream.splitlines()
+
+    # Find where changelog starts in both specfiles
+    changelog_start_in_up_spec = upstream_spec_lines.index("%changelog")
+    changelog_start_in_down_spec = downstream_spec_lines.index("%changelog")
+
+    # Create a new changelog entry
+    date = datetime.now().strftime("%a %b %d %Y")
+    changelog = f"""\
+* {date} {author} - {version}-1
```
achilleas-k, line 61: Since we don't handle rebuilds (`{version}-N`) in the script, maybe it should be mentioned in the guide that rebuilds (though rare) follow a different flow.  At least as a sidenote that we shouldn't look to the guide if we're doing a rebuild release.

achilleas-k (commented):
I love everything about this!  Thank you!!!

(very very minor comments)

atodorov:
Question: do we want to document some release criteria here? It's much more important in the context of downstream releases but why not document it and follow the same principles everywhere?

RELEASING.md
ondrejbudai: Fixed, thanks!

tools/update-distgit.py
ondrejbudai: Fixed, thanks!

tools/update-distgit.py
ondrejbudai: From my point of view, the guide is only for releasing a new upstream version and getting it into downstreams. Doing rebuilds is a topic that's relevant only for downstream and I think it doesn't make much sense to even mention it in upstream guide... In other words, we cannot write here everything about RPM packaging and I suggest drawing the line before rebuilds, they are rare and not advised.

ondrejbudai:
(rebased and fixed issues found by @achilleas-k, thanks!)

@msehnout, @thozza I wonder if you have something to say, I would like at least one more person to review this. :-)

gicmo (approved):
FWIW, this is awesome and looks good to me!

msehnout (approved):
Looks good to me! I used the script in the past and it worked fine.

PR #1523, opened 2021-07-02
worker: Move job tokens to the queue itself

Gundersanne:
This removes state from the worker server, as it no longer contains the
list of running jobs. Instead only the queue knows if jobs are running
or not.

This pull request includes:

  ~~- [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/~~

### DIFF
--- a/internal/jobqueue/fsjobqueue/fsjobqueue.go
+++ b/internal/jobqueue/fsjobqueue/fsjobqueue.go
@@ -40,6 +40,17 @@ type fsJobQueue struct {
 	// Maps job ids to the jobs that depend on it, if any of those
 	// dependants have not yet finished.
 	dependants map[uuid.UUID][]uuid.UUID
+
+	// Currently running jobs. Workers are not handed job ids, but
+	// independent tokens which serve as an indirection. This enables
+	// race-free uploading of artifacts and makes restarting composer more
+	// robust (workers from an old run cannot report results for jobs
+	// composer thinks are not running).
+	// This map maps these tokens to job ids. Artifacts are stored in
+	// `$STATE_DIRECTORY/artifacts/tmp/$TOKEN` while the worker is running,
+	// and renamed to `$STATE_DIRECTORY/artifacts/$JOB_ID` once the job is
+	// reported as done.
+	jobIdByToken map[uuid.UUID]uuid.UUID
 }
 
 // On-disk job struct. Contains all necessary (but non-redundant) information
@@ -47,6 +58,7 @@ type fsJobQueue struct {
 // (de)serialized on each access.
 type job struct {
 	Id           uuid.UUID       `json:"id"`
+	Token        uuid.UUID       `json:"token"`
 	Type         string          `json:"type"`
 	Args         json.RawMessage `json:"args,omitempty"`
 	Dependencies []uuid.UUID     `json:"dependencies"`
@@ -68,9 +80,10 @@ const channelSize = 100
 // loaded and rescheduled to run if necessary.
 func New(dir string) (*fsJobQueue, error) {
 	q := &fsJobQueue{
-		db:         jsondb.New(dir, 0600),
-		pending:    make(map[string]chan uuid.UUID),
-		dependants: make(map[uuid.UUID][]uuid.UUID),
+		db:           jsondb.New(dir, 0600),
+		pending:      make(map[string]chan uuid.UUID),
+		dependants:   make(map[uuid.UUID][]uuid.UUID),
+		jobIdByToken: make(map[uuid.UUID]uuid.UUID),
 	}
 
 	// Look for jobs that are still pending and build the dependant map.
@@ -78,15 +91,30 @@ func New(dir string) (*fsJobQueue, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error listing jobs: %v", err)
 	}
+
 	for _, id := range ids {
-		uuid, err := uuid.Parse(id)
+		jobId, err := uuid.Parse(id)
 		if err != nil {
 			return nil, fmt.Errorf("invalid job '%s' in db: %v", id, err)
 		}
-		j, err := q.readJob(uuid)
+		j, err := q.readJob(jobId)
 		if err != nil {
 			return nil, err
 		}
+
+		// If a job is running, and not cancelled, track the token
+		if !j.StartedAt.IsZero() && j.FinishedAt.IsZero() && !j.Canceled {
+			// Fail older running jobs which don't have a token stored
+			if j.Token == uuid.Nil {
+				err = q.FinishJob(j.Id, nil)
+				if err != nil {
+					return nil, fmt.Errorf("Error finishing job '%s' without a token: %v", j.Id, err)
+				}
+			} else {
+				q.jobIdByToken[j.Token] = j.Id
+			}
+		}
+
 		err = q.maybeEnqueue(j, true)
 		if err != nil {
 			return nil, err
@@ -102,6 +130,7 @@ func (q *fsJobQueue) Enqueue(jobType string, args interface{}, dependencies []uu
 
 	var j = job{
 		Id:           uuid.New(),
+		Token:        uuid.Nil,
 		Type:         jobType,
 		Dependencies: dependencies,
 		QueuedAt:     time.Now(),
@@ -140,13 +169,13 @@ func (q *fsJobQueue) Enqueue(jobType string, args interface{}, dependencies []uu
 	return j.Id, nil
 }
 
-func (q *fsJobQueue) Dequeue(ctx context.Context, jobTypes []string) (uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
+func (q *fsJobQueue) Dequeue(ctx context.Context, jobTypes []string) (uuid.UUID, uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
 	// Return early if the context is already canceled.
 	if err := ctx.Err(); err != nil {
-		return uuid.Nil, nil, "", nil, err
+		return uuid.Nil, uuid.Nil, nil, "", nil, err
 	}
 
 	// Filter q.pending by the `jobTypes`. Ignore those job types that this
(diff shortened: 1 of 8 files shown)


internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -300,6 +304,28 @@ func (q *fsJobQueue) Job(id uuid.UUID) (jobType string, args json.RawMessage, de
 	return
 }
 
+func (q *fsJobQueue) IdFromToken(token uuid.UUID) (id uuid.UUID, err error) {
+	ids, err := q.db.List()
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("error listing jobs: %v", err)
+	}
+
+	for _, id := range ids {
```
teg, line 313: This almost certainly does not matter, but linear search is making me a bit anxious ;) In principle the number of jobs could grow very large over time (we don't delete them when they finish), so reading them back in on every call might not be ideal (this may be mostly academic).

How about creating a second set of objects in addition to the jobs, that represent the "work"/"job slot" or whatever you want to call it. For now they could just be named after the token and contain the job id. That way lookup in both directions would be `O(1)`.

Gundersanne: Makes sense :)

Gundersanne: I think I can simplify it by keeping the token in the job struct, but just having a map of token -> id, populating that at the start when going through all jobs anyway, and then making sure that's threadsafe when reading or writing. No need to have another object on disk, as the fsqueue isn't meant to be used concurrently anyway, this should be fine I think?

teg: Yeah, that should be fine too!

internal/worker/server.go
```diff
@@ -267,47 +252,34 @@ func (s *Server) RequestJob(ctx context.Context, arch string, jobTypes []string)
...
+		case jobqueue.ErrNotExist:
+			return ErrInvalidToken
+		default:
+			return err
+		}
 	}
 
-	// Always delete the running job, even if there are errors finishing
-	// the job, because callers won't call this a second time on error.
-	delete(s.running, token)
```
teg, line 306: Is the token correctly invalidated in the new code? Maybe make a new version of `IdFromToken()` that pops the token so that `FinishJob()` is guaranteed to return `ErrInvalidToken` if it is called more than once?

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -253,6 +260,10 @@ func (q *fsJobQueue) FinishJob(id uuid.UUID, result interface{}) error {
 		return fmt.Errorf("error writing job %s: %v", id, err)
 	}
 
+	q.mapMu.Lock()
```
teg, line 263: I think the same applies to this patch, that we can just use `mu` instead of `mapMu`.

teg (changes requested):
This is really great! Will make everything much more stable. Had some comments about the locking, otherwise good!

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -40,13 +40,18 @@ type fsJobQueue struct {
 	// Maps job ids to the jobs that depend on it, if any of those
 	// dependants have not yet finished.
 	dependants map[uuid.UUID][]uuid.UUID
+
+	// maps token to jobID
+	mapMu        sync.RWMutex
```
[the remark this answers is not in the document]
Gundersanne: Ah yea I think I can repurpose `mu`. I used a new mutex, but a `RWMutex` to have a better concurrent reading/writing of maps, but we can just reuse `mu` as this will probably not be the bottleneck :D

internal/worker/server.go
Gundersanne: Ah yea, I just build the map regardless if it's running or not. I'll make sure to change this, and add a test for it to make sure :)

Gundersanne: Hm, `ErrInvalidToken` results in a 404 `not found`. Now the error will be `ErrJobNotRunning`, which also results in a 404 `not found` :D So it'll be hard to test, but i'll make sure to remove tokens once the job has finished or has been cancelled.

Gundersanne: I'll not be removing the token when it has been cancelled to preserve current behavior: If a job is cancelled you can ask if it's cancelled via GetJob as long as composer isn't restarted. As soon as it is restarted, the token will not be found as the token is gone.

internal/jobqueue/fsjobqueue/fsjobqueue.go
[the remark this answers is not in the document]
teg: And if it is, probably the db is the solution ;)

internal/worker/server.go
teg: Maybe add a todo about cleaning this up? Sounds like the current semantics is weird

Gundersanne: I'll make an issue. Tbh since it's 'internal' api between the workers and composer it's not hugely important. But I admit the behavior is a little weird.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -68,25 +81,38 @@ const channelSize = 100
...
-		j, err := q.readJob(uuid)
+		j, err := q.readJob(jobId)
 		if err != nil {
 			return nil, err
 		}
+
+		// If a job is running, and not cancelled, track the token
+		if !j.StartedAt.IsZero() && j.FinishedAt.IsZero() && !j.Canceled {
+			// Ignore older jobs which don't have a token stored
+			if j.Token != uuid.Nil {
```
teg, line 110: In case the token is not set the job cannot be updated and we should probably fail it on load, shouldn't we? Could be a follow up. Would let you drop the "doesn't work for old jobs" from the news entry ;)

ondrejbudai: +1 for finishing the job.

internal/jobqueue/jobqueue.go
```diff
@@ -62,6 +62,13 @@ type JobQueue interface {
 
 	// Job returns all the parameters that define a job (everything provided during Enqueue).
 	Job(id uuid.UUID) (jobType string, args json.RawMessage, dependencies []uuid.UUID, err error)
+
+	// Find job by token, this will return an error if the job hasn't been dequeued
+	IdFromToken(token uuid.UUID) (id uuid.UUID, err error)
+
+	// Get a list of tokens which haven't been updated in the specified time frame
+	Heartbeats(olderThan time.Duration) (tokens []uuid.UUID)
+	RefreshHeartbeat(token uuid.UUID)
```
ondrejbudai, line 71: Missing comment (:

internal/worker/server.go
```diff
@@ -124,6 +112,31 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
 	}
 }
 
+// This function should be started as a goroutine
+// Every 30 seconds it goes through all running jobs, removing any unresponsive ones.
+// It fails jobs which fail to check if they cancelled for more than 2 minutes.
+func (s *Server) WatchHeartbeats() {
+	for range time.Tick(time.Second * 30) {
+		for _, token := range s.jobs.Heartbeats(time.Second * 120) {
+			go func(token uuid.UUID) {
```
ondrejbudai, line 121: Why is this running in a separate go routine?

ondrejbudai (commented):
Very nice! Two comments inline.

internal/worker/server.go
Gundersanne: Well this should run throughout the lifetime of the server, as there isn't really a prompt to make it check for heartbeats.

ondrejbudai: Yeah, I understand this, this is the reason why the whole `WatchHeartbeats` method runs in its own goroutine. I wonder why we spawn a separate one for each timed-out job. Can we just have one for everything?

Gundersanne: Oh the inner one. Yea there's no reason for that to be there, I'll remove it :D Thanks for noticing

internal/worker/server.go
```diff
@@ -111,6 +112,30 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
+// This function should be started as a goroutine
+// Every 30 seconds it goes through all running jobs, removing any unresponsive ones.
+// It fails jobs which fail to check if they cancelled for more than 2 minutes.
+func (s *Server) WatchHeartbeats() {
+	//nolint:staticcheck // avoid SA1015, this is an endless function
+	for range time.Tick(time.Second * 30) {
+		for _, token := range s.jobs.Heartbeats(time.Second * 120) {
+			id, _ := s.jobs.IdFromToken(token)
+			log.Printf("Removing unresponsive job: %s\n", id)
+			result, err := json.Marshal(OSBuildJobResult{
```
teg, line 124: Not all jobs will be `osbuild` jobs, so don't think we can put an `OSBuildJobResult` there unconditionally.

internal/worker/server.go
```diff
@@ -111,6 +112,30 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
+				Success: false,
+			})
+			if err != nil {
+				log.Printf("Error finishing unresponsive job: %v", err)
+				return
+			}
+			err = s.FinishJob(token, result)
```
teg, line 131: In `New()` you pass `nil` instead of `result` when you fail the job. Is that ok (i.e., will any parsers of this assume it is non-`nil`?), if so, maybe just do that here too?

teg (changes requested):
Looking great, but found one more nit :)

internal/worker/server.go
Gundersanne: Ah yea, makes a lot more sense :)

teg (approved):
Love it!

PR #1524, opened 2021-07-02
CI: test installer

achilleas-k:
This pull request includes:

Keeping as draft to find the best way to test the installer.


atodorov (approved):
LGTM but it seems like this PR needs to be rebased.

achilleas-k:
The test script needs some work.  I'm having trouble getting the installation to work fully automatically.  I suspect it has to do with the ISO having a kickstart file already embedded and the addition of the other kickstart through `virt-install`.  From reading the docs+source it seemed like the new ks would override the existing one, but maybe I missed something.

atodorov:
@achilleas-k can you explain why do you need to have 2 kickstart files? I'm not sure I understand what's the point here.

achilleas-k:
> @achilleas-k can you explain why do you need to have 2 kickstart files? I'm not sure I understand what's the point here.

The `tar-installer` image type that we build already contains a kickstart file.  It contains a single `liveimg` line that points to the tarball on the ISO.

For testing, we would need to override that ks file to perform a fully non-interactive installation (primarily `autopart`) and also keep the original `liveimg` command.

achilleas-k:
Updated the test: The kickstart file is now modified by mounting the ISO, reading the existing content, and then using `mkksiso` to write a new one with the existing commands and some extra ones specifically for automating the installation and testing.  This should now work without needing to add kickstart-related directives to the `virt-install` command and brings the test closer to the original intended usage, since the original ks contents aren't overwritten entirely but wrapped in the new commands.

achilleas-k:
I think we're ready now.  Watching the tests closely :eyes:

ondrejbudai (approved):
Looks good, thanks!

ondrejbudai (approved):

PR #1529, opened 2021-07-05
test: Allow a failed dmesg imported by new RT kernel

henrywang:
Today, `ostree-ng.sh` failed on dmesg output checking. (https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1397215096). The failed dmesg is not caused by osbuild, but new RT kernel itself(See [bz#1979272](https://bugzilla.redhat.com/show_bug.cgi?id=1979272)). Allow this failed dmesg to make CI happy.

BTW, we have successful test yesterday with rt kernel `kernel-rt-4.18.0-319.rt7.100.el8.x86_64`, but failed today on rt kernel `kernel-rt-4.18.0-320.rt7.101.el8.x86_64`

### DIFF
--- a/test/data/ansible/check_ostree.yaml
+++ b/test/data/ansible/check_ostree.yaml
@@ -310,7 +310,7 @@
       block:
         - assert:
             that:
-              - result_dmesg_error.stdout == ""
+              - "'sha512_generic: module verification failed: signature and/or required key missing - tainting kernel' in result_dmesg_error.stdout or result_dmesg_error.stdout == ''"
             fail_msg: "found error or fail log in rt kernel test"
             success_msg: "no error or fail log found"
       always:


teg (dismissed):
Thanks! maybe we should not be so strict on dmsg?

henrywang:
> Thanks! maybe we should not be so strict on dmsg?

Thanks for review my PR. :+1: I have to use dmsg to check boot errors, like service error, kernel error, etc. That might be not related with osbuild*, but that might be the issue for Edge image.

atodorov (approved):

atodorov:
> That might be not related with osbuild*, but that might be the issue for Edge image. 

This is already going into the realm of checking the built image outside of what's needed for osbuild-composer. Perhaps a downstream test suite (or just a general Edge test suite) will be a more suitable place for this kind of checks.

PR #1538, opened 2021-07-08
Filesystem config

kingsleyzissou:
I am extending the weldr api to accept filesystem customisations. Any existing images will only support the `'/'` mountpoint.
Existing images will through structured errors if an unsupported filesystem configuration has been specified.

I have added
  * the ability to specify filesystem customisations in blueprints file, with `name`, `mountpoint` and `minSize` parameters
  * errors for OSTree types for rhel8 and testing for the rhel8 distro
  * errors for unsupported mountpoints on rhel8 (only '/' supported) and added testing
  
Any feedback on the approach taken, suggestions for improvement would be greatly appreciated. I will then extend the other distro types after this.

This pull request includes:

### DIFF
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -1,15 +1,16 @@
 package blueprint
 
 type Customizations struct {
-	Hostname *string                `json:"hostname,omitempty" toml:"hostname,omitempty"`
-	Kernel   *KernelCustomization   `json:"kernel,omitempty" toml:"kernel,omitempty"`
-	SSHKey   []SSHKeyCustomization  `json:"sshkey,omitempty" toml:"sshkey,omitempty"`
-	User     []UserCustomization    `json:"user,omitempty" toml:"user,omitempty"`
-	Group    []GroupCustomization   `json:"group,omitempty" toml:"group,omitempty"`
-	Timezone *TimezoneCustomization `json:"timezone,omitempty" toml:"timezone,omitempty"`
-	Locale   *LocaleCustomization   `json:"locale,omitempty" toml:"locale,omitempty"`
-	Firewall *FirewallCustomization `json:"firewall,omitempty" toml:"firewall,omitempty"`
-	Services *ServicesCustomization `json:"services,omitempty" toml:"services,omitempty"`
+	Hostname   *string                   `json:"hostname,omitempty" toml:"hostname,omitempty"`
+	Kernel     *KernelCustomization      `json:"kernel,omitempty" toml:"kernel,omitempty"`
+	SSHKey     []SSHKeyCustomization     `json:"sshkey,omitempty" toml:"sshkey,omitempty"`
+	User       []UserCustomization       `json:"user,omitempty" toml:"user,omitempty"`
+	Group      []GroupCustomization      `json:"group,omitempty" toml:"group,omitempty"`
+	Timezone   *TimezoneCustomization    `json:"timezone,omitempty" toml:"timezone,omitempty"`
+	Locale     *LocaleCustomization      `json:"locale,omitempty" toml:"locale,omitempty"`
+	Firewall   *FirewallCustomization    `json:"firewall,omitempty" toml:"firewall,omitempty"`
+	Services   *ServicesCustomization    `json:"services,omitempty" toml:"services,omitempty"`
+	Filesystem []FilesystemCustomization `json:"filesystem,omitempty" toml:"filesystem,omitempty"`
 }
 
 type KernelCustomization struct {
@@ -62,6 +63,11 @@ type FirewallServicesCustomization struct {
 type ServicesCustomization struct {
 	Enabled  []string `json:"enabled,omitempty" toml:"enabled,omitempty"`
 	Disabled []string `json:"disabled,omitempty" toml:"disabled,omitempty"`
+}
+
+type FilesystemCustomization struct {
+	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
+	MinSize    int    `json:"minsize,omitempty" toml:"size,omitempty"`
 }
 
 type CustomizationError struct {
@@ -186,4 +192,11 @@ func (c *Customizations) GetServices() *ServicesCustomization {
 	}
 
 	return c.Services
+}
+
+func (c *Customizations) GetFilesystems() []FilesystemCustomization {
+	if c == nil {
+		return nil
+	}
+	return c.Filesystem
 }
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -200,3 +200,19 @@ func (c *Customizations) GetFilesystems() []FilesystemCustomization {
 	}
 	return c.Filesystem
 }
+
+func (c *Customizations) GetFilesystemsMinSize() uint64 {
+	if c == nil {
+		return 0
+	}
+	agg := 0
+	for _, m := range c.Filesystem {
+		agg += m.MinSize
+	}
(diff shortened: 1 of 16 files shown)


internal/blueprint/customizations.go
```diff
@@ -66,9 +66,9 @@ type ServicesCustomization struct {
 }
```
teg, line 66: This commit should probably be squashed into the previous one :)

internal/blueprint/customizations.go
```diff
@@ -64,6 +65,12 @@ type ServicesCustomization struct {
 	Disabled []string `json:"disabled,omitempty" toml:"disabled,omitempty"`
 }
 
+type FilesystemCustomization struct {
+	Name       *string `json:"name,omitempty" toml:"name,omitempty"`
```
teg: Not sure we need `name`? Is that meant to be used as file system label? I'd vote for deferring that, but happy to be proven wrong :)

teg (changes requested):
This is really great! Thanks @kingsleyzissou! I left a couple of minor comments in-line. I think it would also make sense to extend the basic tests to verify that this works from the CLI. Something like `/` can be configured, and `/boot` fails.

internal/blueprint/customizations.go
kingsleyzissou: Oh yeah, hadn't pushed those changes, meant for it to be `label` but we can defer that

internal/blueprint/customizations.go
kingsleyzissou: @teg okay sure thing, I will extend those tests. Could you please point me in the correct direction for the CLI tests please? =)

internal/blueprint/customizations.go
teg: Probably exposing the label makes sense, but we may want to review if there are any restrictions we should enforce/keep in mind before adding it.

achilleas-k:
Looks good so far.  The minSize doesn't have any effect currently when it's valid.  I imagine that's coming up as well.

Not sure where it's best to handle the size.  The imageOptions that set the size are currently created outside the distro package in the API layer: https://github.com/osbuild/osbuild-composer/blob/78ef2470429d54f16248a12185d8ba13b443d9ee/internal/cloudapi/server.go#L191
but they call the `imageType.Size()` method to handle it, so I suppose that might be a good place to pass the customization as well.  Up to you, really.  There might be a simpler way I haven't considered.

internal/blueprint/customizations.go
```diff
@@ -187,3 +194,10 @@ func (c *Customizations) GetServices() *ServicesCustomization {
 
 	return c.Services
 }
+
+func (c *Customizations) GetFilesystem() []FilesystemCustomization {
```
ondrejbudai, line 198: Maybe `GetFilesystems` would be better since the method returns a slice?

ondrejbudai (commented):
I wonder how to handle old distributions. Do we want to backport filesystem customization into already released distributions? If not, we should just ignore a filesystem customization in these and implement it only in RHEL 8.5. @teg thoughts?

internal/blueprint/customizations.go
kingsleyzissou: Good point!

achilleas-k:
> I wonder how to handle old distributions. Do we want to backport filesystem customization into already released distributions? If not, we should just ignore a filesystem customization in these and implement it only in RHEL 8.5. @teg thoughts?

According to the description in COMPOSER-1019, for existing distributions, only  `/` can be specified with a `size` parameter and it works as an alias for the size argument/option.

teg:
@ondrejbudai I made a proposal in the jira ticket that i thought would be minimally intrusive, but this is definitely a grey area, so feedback welcome!

internal/blueprint/customizations.go
```diff
@@ -64,6 +65,12 @@ type ServicesCustomization struct {
 	Disabled []string `json:"disabled,omitempty" toml:"disabled,omitempty"`
 }
 
+type FilesystemCustomization struct {
+	Name       string `json:"name,omitempty" toml:"name,omitempty"`
+	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
+	MinSize    int    `json:"minsize,omitempty" toml:"size,omitempty"`
+}
```
bcl, line 72: I don't know how elaborate we want to get (hopefully not very...) but for reference this is what pykickstart supports:
https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#id46

teg: Thanks Brian! I think we'll aim for as minimal as possible for now and extend it when we must, but good to keep what kickstart does in mind.

ondrejbudai:
@teg @achilleas-k Ah, right, thanks for the pointer. Yeah, that sounds like a reasonable way to do it. :)

internal/distro/test_distro/distro.go
```diff
@@ -165,6 +166,19 @@ func (t *TestImageType) Exports() []string {
 }
 
 func (t *TestImageType) Manifest(b *blueprint.Customizations, options distro.ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, seed int64) (distro.Manifest, error) {
+	mountpoints := b.GetFilesystems()
+
+	invalidMountpoints := []string{}
```
kingsleyzissou, line 171: @ondrejbudai @teg wasn't too sure how to get around this for testing different distros. For now it's okay because the implementation for RHEL85 will come later, but would you have any tips on how I might be able to test 2 separate distros in the [`compose` tests](https://github.com/osbuild/osbuild-composer/blob/c18e2891d6ea699af4d480ef92578a7547214779/internal/client/compose_test.go#L522)? Is this even possible?

kingsleyzissou: Sorry was confusing myself here. You can ignore that comment above.

teg (dismissed):
This is great! This extends the API, but doesn't enable any new filesystem configurations, so I think this is good to go and further integration tests should be added when we extend distro definitions to take advantage of this.

ondrejbudai (dismissed):
Looks good to me so far, thanks!

I think that this PR would benefit from:
- squashing the 1st and 4th commit (the 4th commit is basically a small enhancement to the 1st one)
- making the commits' subject lines shorter. The subject should have max 72 characters. For body, it can be up to 80. If these limits are exceeded, Github rendering goes wild:

image

btw I really like the tests in this PR. :heart_eyes:

kingsleyzissou:
@teg, I wanted to extend the changes made to the other distros. Would you like me to do that in this PR or maybe do a separate one for the other distros?

@ondrejbudai thanks for the review, I appreciate the feedback and I'll make those changes this morning.

ondrejbudai:
I think it's best to do all of them in one big PR so we don't end up in a half-finished state in main.

achilleas-k (dismissed):
Great work!  Thank you.

teg:
> @teg, I wanted to extend the changes made to the other distros. Would you like me to do that in this PR or maybe do a separate one for the other distros?
> 
> @ondrejbudai thanks for the review, I appreciate the feedback and I'll make those changes this morning.

I agree with @ondrejbudai, we should extend the baseline (only `/` supported) functionality to all the distros in this PR, then merge that, and extend 8.5 in a follow-up PR.

kingsleyzissou:
@ondrejbudai CI has passed now, please could you review if you are happy with the changes you requested?

teg (approved):
Great stuff, thanks for this! Looks like @ondrejbudai's comments were addressed :)

PR #1539, opened 2021-07-09
Db

Gundersanne:
TODO:

- [x] make api tests use db queue
- [x] actually allow selecting the db queue rather than the fsqueue
- [x] add initcontainer for migration

---

This pull request includes:

##### (should we, or should we settle on not shipping this/supporting this for the releases into distros ?)


internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+func (q *dbJobQueue) JobStatus(id uuid.UUID) (result json.RawMessage, queued, started, finished time.Time, canceled bool, deps []uuid.UUID, err error) {
+	conn, err := q.pool.Acquire(context.Background())
+	if err != nil {
+		return
+	}
+	defer conn.Release()
+
+	var sp, fp *time.Time
+	var rp pgtype.JSON
+	err = conn.QueryRow(context.Background(), sqlQueryJobStatus, id).Scan(&rp, &queued, &sp, &fp, &canceled)
```
ondrejbudai, line 279: Doesn't it pass double pointers inside? (`**time.Time`) Is that what's desired? Feels weird.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+
+	dependencies, err := q.jobDependencies(ctx, conn, id)
+	if err != nil {
+		return uuid.Nil, uuid.Nil, nil, "", nil, fmt.Errorf("error querying the job's dependencies: %v", err)
+	}
+
+	return id, token, dependencies, jobType, args, nil
+}
+
+func (q *dbJobQueue) FinishJob(id uuid.UUID, result interface{}) error {
```
ondrejbudai, line 195: I think this method should send a notification because finishing one job can trigger another one.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("error connecting to database: %v", err)
+	}
+	defer conn.Release()
+
+	tx, err := conn.Begin(context.Background())
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("error starting database transaction: %v", err)
+	}
+	defer func() { _ = tx.Rollback(context.Background()) }()
```
ondrejbudai, line 115: I would honestly prefer logging even these rollback errors. I think that `Rollback` returns an error only for a closed transaction which should never happen but just in case.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+		return fmt.Errorf("error connecting to database: %v", err)
+	}
+	defer conn.Release()
+
+	// todo return errnotexist as well if it can't find it
+	// TODO check if job is not running first and return errnotrunning or errcanceled
+
+	var finished *time.Time
+	canceled := false
+	err = conn.QueryRow(context.Background(), sqlQueryJob, id).Scan(nil, nil, nil, &finished, &canceled)
```
ondrejbudai, line 207: I think we should put this query in a transaction too.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func (q *dbJobQueue) FinishJob(id uuid.UUID, result interface{}) error {
+	conn, err := q.pool.Acquire(context.Background())
+	if err != nil {
+		return fmt.Errorf("error connecting to database: %v", err)
+	}
+	defer conn.Release()
+
+	// todo return errnotexist as well if it can't find it
```
ondrejbudai, line 202: I think this is already done?

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+
+	rows, err := conn.Query(context.Background(), sqlQueryHeartbeats, olderThan.String())
+	if err != nil {
+		return
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var t uuid.UUID
+		_ = rows.Scan(&t) // TODO check err
```
ondrejbudai, line 354: TODO

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+
+// Reset the last heartbeat time to time.Now()
+func (q *dbJobQueue) RefreshHeartbeat(token uuid.UUID) {
+	conn, err := q.pool.Acquire(context.Background())
+	if err != nil {
+		return
+	}
+	defer conn.Release()
+
+	_, _ = conn.Exec(context.Background(), sqlRefreshHeartbeat, token)
```
ondrejbudai, line 369: Missing error check

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -0,0 +1,390 @@
...
+	rows, err := conn.Query(ctx, sqlQueryDependencies, id)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	dependencies := []uuid.UUID{}
+	for rows.Next() {
+		var d uuid.UUID
+		_ = rows.Scan(&d) // TODO check err
```
ondrejbudai, line 382: TODO

ondrejbudai (commented):
Very nice PR. Some comments inline. :)

cmd/osbuild-composer/composer.go
```diff
@@ -60,9 +63,47 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *
 
 	c.rpm = rpmmd.NewRPMMD(path.Join(c.cacheDir, "rpmmd"), "/usr/libexec/osbuild-composer/dnf-json")
 
-	jobs, err := fsjobqueue.New(queueDir)
-	if err != nil {
-		return nil, fmt.Errorf("cannot create jobqueue: %v", err)
+	var jobs jobqueue.JobQueue
+	if config.Worker.DBConfigDir != "" {
+		// Construct dburl from secrets; this is a directory
+		host, err := ioutil.ReadFile(path.Join(config.Worker.DBConfigDir, "host"))
```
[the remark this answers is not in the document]
Gundersanne: This is hairy, I know :( The problem is that the mapping of ocp/k8s secrets doesn't really mesh well with a single config file.

So the problem here is that the db secrets will come in the form of an openshift/k8s secret. You can map those to the environment, or mount them in a directory. In the latter case, a single file would be created per item in the mounted directory, which is what I went for at this point.


The best thing would be if we could map the secrets to ocp template variabls, and use the vars to construct the config with, but sadly that's not possible.

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: huh, I'll try without

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: Hm, this just selects stuff though, no?

Gundersanne: oh sorry, missed the lines below, it then proceeds to modify stuff based on the return value of that select, ok yea  agreed :)

Gundersanne: image
This is why I think, currently those are stored as NULL values rather than `time.IsZero() == true` values. The latter might be better?

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: Hm, do you think we should change the signature of these heartbeat functions to relay errors to the worker server so the server can decide what to do with them?

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: Jus checked, there really isn't a good mapping I'm afraid :/ Closest postgres constant is `allballs` :1

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: ugh I somehow messed this up and replied to the wrong comment: https://github.com/osbuild/osbuild-composer/pull/1539#discussion_r672175785

Gundersanne: Hm so if we change it, we'd need to have time.Time default-values stored, so I think the current form is fine from that perspective. 

Then again I do the opposite for uuids, where I do use `uuid.Nil`.

What do you prefer?

cmd/osbuild-composer/composer.go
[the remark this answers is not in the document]
ondrejbudai: It probably doesn't matter much. Env variables feel slightly more natural to me (when a container requires secrets, they are usually passed in using env variables, eg. postgres). On the other hand, files are very nice too, everything is a file in Unix and it just works, right?

internal/jobqueue/dbjobqueue/dbjobqueue.go
ondrejbudai: if time can be `nil`, this completely makes sense then. Maybe put there a comment? Something like "we're passing a double pointer here so the method can return nil".

internal/jobqueue/dbjobqueue/dbjobqueue.go
ondrejbudai: It would probably just log them because there's nothing else to do, right? Maybe we should just log the errors here and expand the signature only if need to.

Gundersanne: agreed ya

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: Will do!

cmd/osbuild-composer/composer.go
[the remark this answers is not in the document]
Gundersanne: yea the env is slightly more idiomatic for containers, but slightly less for composer. Like if you'd configure this in composer via the env you'd need to edit the systemd unit file? If the latter isn't too much of a pain for users we could just use the env.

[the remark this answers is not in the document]
teg: I would be fine with the container idiomatic thing. I know we shied away from this in the past, but maybe we are just making our lives harder than they need to be? cc @larskarlitski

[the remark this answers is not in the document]
Gundersanne: I changed it so it does both :)

test/cases/api.sh
```diff
@@ -641,6 +694,10 @@ test $((INIT_COMPOSES+1)) = "$SUBS_COMPOSES"
 # shellcheck disable=SC2024
 sudo jq -rM .args.manifest /var/lib/osbuild-composer/jobs/"${COMPOSE_ID}".json > "${ARTIFACTS}/manifest.json"
```
Gundersanne, line 695: @achilleas-k Hm, since composer is backed by a DB here, I assume we want a sql query here. Should I just dump the `args jsonb`?

Gundersanne: Wait, I guess args.Manifest only :) nvm,  think I found it :D

teg (approved):
This looks great! Thanks!

PR #1552, opened 2021-07-14
CI: scheduled cloud-cleaner for Azure

jrusz:
This pull request includes:

I'm trying to see if it's possible to trap cloud-cleaner during test execution

### DIFF
--- a/schutzbot/scheduled_cloud_cleaner.sh
+++ b/schutzbot/scheduled_cloud_cleaner.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Azure cleanup
+if ! hash az; then
+    # this installation method is taken from the official docs:
+    # https://docs.microsoft.com/cs-cz/cli/azure/install-azure-cli-linux?pivots=dnf
+    sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
+    echo -e "[azure-cli]
+name=Azure CLI
+baseurl=https://packages.microsoft.com/yumrepos/azure-cli
+enabled=1
+gpgcheck=1
+gpgkey=https://packages.microsoft.com/keys/microsoft.asc" | sudo tee /etc/yum.repos.d/azure-cli.repo
+
+  greenprint "Installing azure-cli"
+  sudo dnf install -y azure-cli
+  az version
+fi
+
+az login --service-principal --username "${AZURE_CLIENT_ID}" --password "${AZURE_CLIENT_SECRET}" --tenant "${AZURE_TENANT_ID}"
+
+# List all resources from AZURE_RESOURCE_GROUP
+RESOURCE_LIST=$(az resource list -g "$AZURE_RESOURCE_GROUP")
+RESOURCE_COUNT=$( echo "$RESOURCE_LIST" | jq .[].name | wc -l)
+
+# filter out resources older than X hours
+DELETE_TIME=$(date -d "- $HOURS_BACK hours" +%s)
+OLD_RESOURCE_LIST_NAMES=()
+for i in $(seq 0 $(("$RESOURCE_COUNT"-1))); do
+    RESOURCE_TIME=$(echo "$RESOURCE_LIST" | jq .[$i].createdTime | tr -d '"')
+    RESOURCE_TYPE=$(echo "$RESOURCE_LIST" | jq .[$i].type | tr -d '"')
+    RESOURCE_TIME_SECONDS=$(date -d "$RESOURCE_TIME" +%s)
+    if [[ "$RESOURCE_TIME_SECONDS" -lt "$DELETE_TIME" && "$RESOURCE_TYPE" != Microsoft.Storage/storageAccounts ]]; then
+        OLD_RESOURCE_LIST_NAMES+=("$(echo "$RESOURCE_LIST" | jq .["$i"].name | sed -e 's/^[^-]*-//' | tr -d '"')")
+    fi
+done
+
+#Exit early if no there are no resources to delete
+if [ ${#OLD_RESOURCE_LIST_NAMES[@]} == 0 ]; then
+    echo "Nothing to delete."
+    exit 0
+fi
+
+# Keep only unique resource names
+mapfile -t RESOURCE_TO_DELETE_LIST  < <(printf "%s\n" "${OLD_RESOURCE_LIST_NAMES[@]}" | sort -u)
+echo "${RESOURCE_TO_DELETE_LIST[@]}"
+
+TO_DELETE_COUNT=${#RESOURCE_TO_DELETE_LIST[@]}
+echo "There are resources from $TO_DELETE_COUNT test runs to delete."
+
+for i in $(seq 0 $(("$TO_DELETE_COUNT"-1))); do
+    echo "Running cloud-cleaner in Azure for resources with TEST_ID: ${RESOURCE_TO_DELETE_LIST[$i]}"
+    TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner
+done
+
+echo "Azure cleanup complete!"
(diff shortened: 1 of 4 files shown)


jrusz:
Apparently when a job get canceled the runner gets destroyed immediately so trapping cloud-cleaner doesn't seem to be an option. 

I came up with another way which should work but I'm not sure if it's robust enough or if it's not too complicated. I created just a simple first version to present an idea of how it would work. This would solve our problem with canceled pipelines and this is how it's supposed to work:

1. Every X hours send a request to get all canceled jobs in X hours
2. Parse all the required variables from the output
3. Run cloud-cleaner with the required variables for each canceled job

Of course this would filter out jobs that are only interacting with clouds. Also if this is to work properly we'll have to simplify cloud resources naming so that it's possible to put it together even if the original variables are gone meaning not using any uuidgen or something like that. I've added more comments inside the actual scipt.

I will appreciate any comments or suggestions how to make this work for us.

thozza:
I think that running the `cloud-cleaner` regularly is a good idea. The prerequisite is as you've mentioned to use predictable resource names. This is done e.g. in `api.sh` for GCP, where in CI it uses predictable `TEST_ID` and if run outside of CI environment, it uses random UUID. I'm all for using predictable `TEST_ID` as a baseline for predictable resource names, but I think that we should keep a fallback to using random UUID to keep the test case script working in local environments.

So if we can reasonably recreate all necessary environment variables for each canceled job, then this is great and we should definitely run `cloud-cleaner` for them!

In addition, I think that `cloud-cleaner` may need some modifications / fixes to make it work for all tests that create cloud resources, because AFAIK it properly cleans up after `api.sh` only for GCP. For the other clouds, it cleans up only after image tests which can boot the image (I guess, but I'm not sure).

internal/test/helpers.go
```diff
@@ -249,7 +248,5 @@ func GenerateCIArtifactName(prefix string) (string, error) {
 		return "", fmt.Errorf("The environment variables must specify BRANCH_NAME, BUILD_ID, and DISTRO_CODE")
 	}
 
-	arch := common.CurrentArch()
-
-	return fmt.Sprintf("%s%s-%s-%s-%s", prefix, distroCode, arch, branchName, buildId), nil
+	return fmt.Sprintf("%s%s-%s-%s", prefix, distroCode, branchName, buildId), nil
```
atodorov, line 251: won't this lead to conflicts once we start testing on multiple architectures ?

atodorov: Why not prefix all test artifacts with "testing_" and query the cloud environments directly like we did in the lorax days ?

atodorov:
> For the other clouds, it cleans up only after image tests which can boot the image (I guess, but I'm not sure).

For Azure at least (b/c this is most recent) you are correct. api.sh cleans up after itself but that won't happen if the job is cancelled.

internal/test/helpers.go
jrusz: What I did in this PR is that I use the Gitlab runner name as the `DISTRO_CODE`. That contains both the OS, version and architecture and it's easy to recreate for the cloud-cleaner to use. I pretty much merged the variables. It's probably not ideal but it worked nicely in this case. 

Prefixing all artifacts with "testing_" also sounds good, I'm just not sure how much of a change in cloud-cleaner code would that require...

thozza: Prefixing all artifacts with `testing_` won't work at least for GCP, because resource names are pretty much at the limit of their length. For GCP, I didn't use the content of the environment variables, but their hash to prevent issues with invalid characters in the resource name (e.g. `_` is not allowed). See https://github.com/osbuild/osbuild-composer/blob/c09638efe51d40b5ae1744fcbf71e5e22ff37dc9/test/cases/api.sh#L448

ondrejbudai: FTR I proposed using tags, I hope that all cloud envs do support them. AWS and Azure surely do, what about GCP @thozza?

thozza: GCP definitely has some features similar to tags, which we could use. AFAIK for Cloud Storage objects, we can set only custom metadata (key-value). We actually already set it to the desired image name in composer when we upload the built image to Storage, otherwise we could not possibly determine the association between storage objects and image names in `cloud-cleaner` https://github.com/osbuild/osbuild-composer/blob/b7df6ac6275a317d9741e981ac84872d353989c4/cmd/cloud-cleaner/main.go#L98

For GCP Compute resources, there are "tags", but these have a bit different use case and seem to be used only for networking. We can also set custom metadata, but probably the best approach would be to set ["Labels"](https://cloud.google.com/compute/docs/labeling-resources), which can be then used directly with `gcloud` or API calls to list only specific resources.

I must say that GCP is a bit confusing in the sense that each "product" (e.g. Cloud Storage vs. Compute Engine) have a different feature set WRT tags/labels/metadata.

jrusz:
Ok finally pushed an update. I tried to describe it in the commit message, hope it's clear what it does now. 

The reason why I'm not using tags for the Azure resources is because of this bug https://github.com/Azure/azure-cli/issues/19138 ... If I listed just tagged resources I could not tell when they were created and I didn't want to clean them all since they could be from a currently running job. It would be possible to just look for the tag in the resources I'm listing now but I don't see any added value currently as it's only listing resources from a resource group that's used for testing and if someone names their resource differently that the jobs do it won't clean them up anyway. I'm fine with adding the additional check for tags if you think it's going to be better though.

The plan is to create a scheduled pipeline which will run it once a day for now and see if that's enough. The time to look back can be set as env var and I'm thinking to currently set it to 3 hours as that's the timeout we have for jobs.

I'm also currently doing a test where I want to cancel api and integration azure jobs in the middle of creating resources and let this clean up. I'll post the results in the next comment.

jrusz:
This is now ready for review. I tested it here https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1507709881 . It cleaned up leftover resources from one of my failed previous runs. See my previous comment and commits for more info.

schutzbot/scheduled_cloud_cleaner.sh
```diff
@@ -0,0 +1,57 @@
...
+mapfile -t RESOURCE_TO_DELETE_LIST  < <(printf "%s\n" "${OLD_RESOURCE_LIST_NAMES[@]}" | sort -u)
+echo "${RESOURCE_TO_DELETE_LIST[@]}"
+
+TO_DELETE_COUNT=${#RESOURCE_TO_DELETE_LIST[@]}
+echo "$TO_DELETE_COUNT"
+
+for i in $(seq 0 $(("$TO_DELETE_COUNT"-1))); do
+    CLEANER_CMD="TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner"
+    echo "Running cloud-cleaner in Azure for resources with TEST_ID: ${RESOURCE_TO_DELETE_LIST[$i]}"
+    eval "$CLEANER_CMD"
```
atodorov, line 54: I think you should be able to just execute the command as:
`TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner`

and avoid using `eval`. I'm surprised Shellcheck doesn't complain about it.

atodorov (dismissed):
Small nitpick, otherwise looks good to me.

internal/test/helpers.go
atodorov: @ondrejbudai, @thozza can we mark this comment as resolved or do you have strong preference on using tags instead of naming conventions ?

schutzbot/scheduled_cloud_cleaner.sh
jrusz: Yeah that works, thanks!

internal/test/helpers.go
thozza: Hi @atodorov. I'm OK with using naming convention as long as it works with all cloud providers and existing test cases which are creating cloud resources.

jrusz:
Shall we merge this @atodorov ? It seems there are no complaints against this and I don't want to leave it open for too long.

atodorov (approved):

Issue #1477, opened 2021-06-17
CloudAPI: Invalid package set returns 500


teg:
**Describe the bug**
Passing an invalid compose request should return BadRequest, not InternalServerError. Requesting combinations of packages that cannot be installed is a user error.

**Environment**
 * osbuild-composer version 30

**To Reproduce**
Build an AMI for RHEL8.4 with `alsa-lib` in the package set (or any other excluded package). See the request failing with InternalServerError.

**Expected behavior**
The request fails with BadRequest.

**Additional context**
Fixing this would allow us more accurate metrics when measuring the stability of our API.

Reported by @Gundersanne

teg:
This might also be relevant fort he weldr API to make the error handling there more helpful for cockpit-composer, but I did not check. cc @jkozol

lavocatt:
At first we were not able to reproduce the issue on rhel-8. @Gundersanne came up with this solution:

```
{
    "distribution": "centos-8",
    "image_requests": [ { "architecture": "x86_64",
        "image_type": "ami",
        "repositories": [
            {
                "baseurl": "http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/",
                "rhsm": false
            }, {
            "baseurl": "http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/",
            "rhsm": false
            }, {
            "baseurl": "http://mirror.centos.org/centos/8-stream/extras/x86_64/os/",
            "rhsm": false
            }
        ],
        "upload_request": {
            "type": "aws.s3",
            "options": {
                "region": "truc",
                "s3": {
                    "access_key_id": "muche",
                    "secret_access_key": "bidule",
                    "bucket": "chose"
                }
            }
        }
    } ],
    "customizations": { "packages": [ "libcanberra" ] }
}
```

And we do get the desired output:

```
Failed to depsolve base packages for ami/x86_64/centos-8: DNF error occured: DepsolveError: There was a problem depsolving ['checkpolicy', 'chrony', 'cloud-init', 'cloud-init', 'cloud-utils-growpart', '@core', 'dhcp-client', 'gdisk', 'shim-x64', 'langpacks-en', 'net-tools', 'NetworkManager', 'redhat-release', 'redhat-release-eula', 'rsync', 'selinux-policy-targeted', 'tar', 'yum-utils', 'libcanberra', 'kernel', 'dracut-config-generic', 'grub2-pc', 'grub2-efi-x64']: 
[...]

PR #1563, opened 2021-07-16
CloudAPI: Invalid package set returns 400

lavocatt:
Previously a bad error code was returned, solves #1477.

This pull request includes:

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -182,7 +182,17 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 		for name, packages := range packageSets {
 			pkgs, _, err := server.rpmMetadata.Depsolve(packages, repositories, distribution.ModulePlatformID(), arch.Name())
 			if err != nil {
-				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), http.StatusInternalServerError)
+				var error_type int
+				switch err.(type) {
+				// Known DNF errors falls under BadRequest
+				case *rpmmd.DNFError:
+					error_type = http.StatusBadRequest
+				// All other kind of errors are internal server Errors.
+				// (json marshalling issues for instance)
+				case error:
+					error_type = http.StatusInternalServerError
+				}
+				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), error_type)
 				return
 			}
 			pkgSpecSets[name] = pkgs
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -953,4 +953,35 @@ curl \
         --header "x-rh-identity: $INVALIDAUTHSTRING" \
         http://localhost:443/api/composer/v1/version)" = "404" ]
 
+#
+# Make sure that requesting a non existing paquet returns a 400 error
+#
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    --header "x-rh-identity: $VALIDAUTHSTRING" \
+    -H "Content-Type: application/json" \
+    -d '{ "distribution": "centos-8", "image_requests": [ { "architecture": "x86_64", "image_type": "ami", "repositories": [ { "baseurl": "http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/", "rhsm": false }, { "baseurl": "http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/", "rhsm": false }, { "baseurl": "http://mirror.centos.org/centos/8-stream/extras/x86_64/os/", "rhsm": false } ], "upload_request": { "type": "aws.s3", "options": { "region": "somewhere", "s3": { "access_key_id": "thingy", "secret_access_key": "thing", "bucket": "thingything" } } } } ], "customizations": { "packages": [ "jesuisunpaquetquinexistepas_idonotexist" ] } }' \
+    http://localhost:443/api/composer/v1/compose)" = "400" ]
+
+#
+# Make sure that a request that makes the dnf-json crash returns a 500 error
+#
+sudo cp -f /usr/libexec/osbuild-composer/dnf-json /usr/libexec/osbuild-composer/dnf-json.bak
+sudo cat << EOF | sudo tee /usr/libexec/osbuild-composer/dnf-json
+#!/usr/bin/python3
+raise Exception()
+EOF
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    --header "x-rh-identity: $VALIDAUTHSTRING" \
+    -H "Content-Type: application/json" \
+    -d '{ "distribution": "centos-8", "image_requests": [ { "architecture": "x86_64", "image_type": "ami", "repositories": [ { "baseurl": "http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/", "rhsm": false }, { "baseurl": "http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/", "rhsm": false }, { "baseurl": "http://mirror.centos.org/centos/8-stream/extras/x86_64/os/", "rhsm": false } ], "upload_request": { "type": "aws.s3", "options": { "region": "somewhere", "s3": { "access_key_id": "thingy", "secret_access_key": "thing", "bucket": "thingything" } } } } ], "customizations": { "packages": [ "jesuisunpaquetquinexistepas_idonotexist" ] } }' \
+    http://localhost:443/api/composer/v1/compose)" = "500" ]
+
+sudo mv -f /usr/libexec/osbuild-composer/dnf-json.bak /usr/libexec/osbuild-composer/dnf-json
+
 exit 0


internal/cloudapi/server.go
```diff
@@ -182,7 +182,7 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 		for name, packages := range packageSets {
 			pkgs, _, err := server.rpmMetadata.Depsolve(packages, repositories, distribution.ModulePlatformID(), arch.Name())
 			if err != nil {
-				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), http.StatusInternalServerError)
+				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), http.StatusBadRequest)
```
teg, line 185: I wonder, could we distinguish between the different error types? I.e., distinguish between "conflicting packages"/"unknown package" (should return 400) on the one hand, and "file not found"/<genuine internal error> on the other (should be 500)? In the case of a 400 it would be ideal to return some structured error that we could eventually expose in some nice way to the user. I.e., an array of packages that were not found, or some structured information about conflicting requests.

The structured errors should probably be done in a follow-up PR, distinguishing between the error codes could be done here or in a follow-up, depending on how much work that would be, this is definitely already an improvement not he existing behaviour.

teg (dismissed):
This is great! I left some suggestions inline, up to you if you want to address those now or not. Only thing I'd like to ask though is to add a test that verifies the return code, like you did manually, just in CI.

internal/cloudapi/server.go
lavocatt: Thanks for the comment!

I agree with you. And it seems that there is an issue with my test case anyway.
I've got a weird error `Failed to depsolve base packages for ami/x86_64/rhel-8: unexpected end of JSON input`

Looking at my logs, I can see that the parsing of my input json fails.

```
Jul 16 16:05:50 fedora python3[682563]: detected unhandled Python exception in '/usr/libexec/osbuild-composer/dnf-json'
Jul 16 16:05:50 fedora osbuild-composer[682563]: Traceback (most recent call last):
Jul 16 16:05:50 fedora osbuild-composer[682563]:   File "/usr/libexec/osbuild-composer/dnf-json", line 117, in <module>
Jul 16 16:05:50 fedora osbuild-composer[682563]:     base = create_base(
Jul 16 16:05:50 fedora osbuild-composer[682563]:   File "/usr/libexec/osbuild-composer/dnf-json", line 70, in create_base
Jul 16 16:05:50 fedora osbuild-composer[682563]:     for repo in repos:
Jul 16 16:05:50 fedora osbuild-composer[682563]: TypeError: 'NoneType' object is not iterable
```

Something is off with either my test scenario or something else and I need to investigate.

Either way, in this specific configuration, the result should clearly be a 500 error and not a 400 as the server crashed.

If it is right for you, I'd like to explore answering the question within this PR as I think my actual solution is too simple to be considered the right answer. And correcting it may result in a PR with the feature you were talking about.

teg: Ah, indeed, an exception should be a 500. Good catch!

The problem is that your request does not contain any repositories, but we should either check for this and return 400, or we should handle it gracefully in `dns-json`.

> If it is right for you, I'd like to explore answering the question within this PR as I think my actual solution is too simple to be considered the right answer.

Sounds perfect!

internal/cloudapi/server.go
```diff
@@ -182,7 +182,22 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 		for name, packages := range packageSets {
 			pkgs, _, err := server.rpmMetadata.Depsolve(packages, repositories, distribution.ModulePlatformID(), arch.Name())
 			if err != nil {
-				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), http.StatusInternalServerError)
+				var error_type int
+				switch err.(type) {
+				// DNFException means that the dnf-json tool encountered a non
+				// expected error during its execution. This subtype is handled
+				// as a 500 error.
+				case *rpmmd.DNFException:
```
lavocatt, line 190: In fact, I'm starting to wonder if we really need the DNFException. Like if json unmarshalling crashed and led to an error, shouldn't it be enough to declare an Internal Server Error ?
Let me now if you prefer that I simplify.

dnf-json
```diff
@@ -104,97 +108,103 @@ def repo_checksums(base):
...
-        base = create_base(
-            repos,
-            module_platform_id,
-            persistdir,
-            cachedir,
-            arch
-        )
-    except dnf.exceptions.Error as e:
-        exit_with_dnf_error(
-            type(e).__name__,
```
Gundersanne, line 126: Out of curiosity, is `type(e).__name__` equal to `dnf.exceptions.Error` or just `Error` ?

internal/cloudapi/server.go
```diff
@@ -182,7 +182,22 @@ func (server *Server) Compose(w http.ResponseWriter, r *http.Request) {
 		for name, packages := range packageSets {
 			pkgs, _, err := server.rpmMetadata.Depsolve(packages, repositories, distribution.ModulePlatformID(), arch.Name())
 			if err != nil {
-				http.Error(w, fmt.Sprintf("Failed to depsolve base packages for %s/%s/%s: %s", ir.ImageType, ir.Architecture, request.Distribution, err), http.StatusInternalServerError)
+				var error_type int
+				switch err.(type) {
```
Gundersanne, line 186: I think here we should look into the `DNFError.Kind` field and check what kind of error we got. A `DownloadError` for instance is definitely a 500 imo, but a `DespolveError` should be a 400.

internal/rpmmd/repository.go
```diff
@@ -186,6 +186,13 @@ type DNFError struct {
 	Reason string `json:"reason"`
 }
 
+// DNFException inherits from DNFError. Represents an error due to a crash
+// on the python side. Its purpose is to give a way to the user of Depsolve to
+// differentiate the types of errors.
+type DNFException struct {
```
Gundersanne, line 192: Hmm, all DNFErrors are `dnf.exceptions`. We should filter what to do with the based on `DNFError.Kind` rather than the golang type.

internal/cloudapi/server.go
Gundersanne: Indeed I think it might make more sense to just have DNFError, I agree. And then filtering further based on specific dnf errors.

Gundersanne (changes requested):
Thank you! Definitely in the right direction I think.

So I think the distinction between Error and Exception which has been made is a bit quirky. They're all dnferrors.

If we just take the `DNFError.Kind` field to be always equal to the exception names [listed in the docs](https://dnf.readthedocs.io/en/latest/api_exceptions.html), we could just end up with something like:

```
_, _, err := rpmmd.Depsolve(...)

dnfErr, ok := err.(rpmmd.DNFError)
if ok && dnfErr.Kind == "DepsolveError" {
    return 400
} else {
    return 500
}
```
etc...

And like you said, if it crashes it should just be a 500, I agree.

dnf-json
lavocatt: I believe that the name of the exception is going to be the name of the actual instantiated class.
For this particular feature, It would have gave the same behaviour with only `except Exception as e:`

internal/cloudapi/server.go
lavocatt: With this solution I am aiming at a proposition which does not involve string parsing. But I have no reservation to go this direction.

The only thing is that the type of the error is not necessarily `*rpmmd.DNFError` at this point. The return type of `server.rpmMetadata.Depsolve(..)` is `error` so any kind of error is possible.
It means that to have a string parsing we will have to keep the distinction between the type `*rpmmd.DNFError` that contains a `kind` variable and the rest of the possible errors. I believe that any of those possible errors can be 500 anyway.

internal/rpmmd/repository.go
lavocatt: Yeah, I know, using type overloading to sort the exceptions was not the most straightforward approach, but this was mainly to avoid string parsing.
And in fact, in the end, this type does not proves itself this useful.

ondrejbudai:
I feel slightly lost in this PR so ignore me if I'm too confused.

I think that the return values of `dnf-json` are currently very primitive and can be expanded greatly. `dnf-json` should be imho simple and stupid. In the case of an error, it should just return a structured one - its type, a generic error message (never hurts), and type-specific data - e.g. a list of conflicting packages.

osbuild-composer should be the smart one - if `dnf-json` doesn't return `json` or returns a weird status code, that's surely 500. When it gets a nice structured error, it should examine it and return an appropriate error. I think that `rpmmd` should be only responsible for API translation (Go code <=> running python script) and the cloud server should be one to decide the mapping between an error type and HTTP error code.

Why should the API server be responsible for mapping the error codes? Because they might be different for different APIs. Imagine an error that was caused by failing to download metadata. In the case of the cloud API, that is probably 403 because it's the caller's responsibility to pass working repos. If the caller passed broken repos, it's not composer's fault, it's just a bad request. However, in the case of the weldr API, we define the repos inside composer so we might want to return 500 because the user did nothing wrong.

On the strings vs. "subtypes" debate: If we want to return more structured errors, subtypes are probably the only way. (I kinda imagine the `dnf-json`  error types to be similar to `osbuild` stages with their well-defined input schemas.) All Go types representing `dnf-json` error types should implement the `error` interface to make the code more idiomatic. `error.Is` and `error.As` methods can be used to do the mapping between `dnf-json` error type and HTTP error.

dnf-json
```diff
@@ -10,7 +10,6 @@ import tempfile
 
 DNF_ERROR_EXIT_CODE = 10
 
-
```
Gundersanne: newline noise

dnf-json
```diff
@@ -78,6 +77,9 @@ def exit_with_dnf_error(kind: str, reason: str):
     json.dump({"kind": kind, "reason": reason}, sys.stdout)
     sys.exit(DNF_ERROR_EXIT_CODE)
 
+def exit_with_dnf_exception(kind: str, reason: str):
```
Gundersanne: This isn't used anymore now no? Does the same thing as `exit_with_dnf_error`.

internal/cloudapi/server.go
Gundersanne: This makes sense I think, and if we'd ever want to expand it in future we'd just create types in go for those specific errors (like `rpmmd.DNFDepsolveError`) etc...

test/cases/api.sh
```diff
@@ -953,4 +953,30 @@ curl \
...
         http://localhost:443/api/composer/v1/version)" = "404" ]
 
+#
+# Make sure that a request that makes the dnf-json crash returns a 500 error
+#
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    -k --key /etc/osbuild-composer/client-key.pem \
```
Gundersanne, line 963: At this point osbuild-composer has been restarted to work with the rh-identity header (see line 933 and up). So https and --key/--cert won't work here. Either move above line 919 or change the request here to `http` with `--x-rh-identity: "$VALIDAUTHSTRING"`.

The header is something that gets added when composer is running behind the CRC gateway (which handles auth for us).

test/cases/api.sh
```diff
@@ -953,4 +953,30 @@ curl \
...
+    https://localhost/api/composer/v1/compose)" = "500" ]
+
+#
+# Make sure that requesting a non existing paquet returns a 400 error
+#
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    -k --key /etc/osbuild-composer/client-key.pem \
```
Gundersanne, line 976: ditto on the https/mTLS

test/cases/api.sh
```diff
@@ -953,4 +953,30 @@ curl \
...
+# Make sure that a request that makes the dnf-json crash returns a 500 error
+#
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    -k --key /etc/osbuild-composer/client-key.pem \
+    --cert /etc/osbuild-composer/client-crt.pem \
+    -H "Content-Type: application/json" \
+    -d '{"distribution":"centos-8","image_requests":[{"architecture":"x86_64","image_type":"ami"}]}' \
```
Gundersanne, line 966: I'm not sure about the 500 test case here. Since it's kinda weird it gets to trying depsolving in the first place when the request itself is invalid. So the ideal scenario should be 400 here (but let's improve input validation later).

How about having a valid request, but forcing dnf-json to crash? We could:
1. mv /usr/libexec/osbuild-composer/dnf-json /usr/libexec/osbuild-composer/dnf-json.bak
2. write a python3 executable to /usr/libexec/osbuild-composer/dnf-json which always crashes, just something like:
```
#!/usr/bin/python3
raise Exception
```
3. check that it returns 500 now
4. mv -f /usr/libexec/osbuild-composer/dnf-json.bak /usr/libexec/osbuild-composer/dnf-json

thoughts?

Gundersanne (changes requested):
Thanks!

Also replying to @ondrejbudai 

> I feel slightly lost in this PR so ignore me if I'm too confused.
> 
> I think that the return values of `dnf-json` are currently very primitive and can be expanded greatly. `dnf-json` should be imho simple and stupid. In the case of an error, it should just return a structured one - its type, a generic error message (never hurts), and type-specific data - e.g. a list of conflicting packages.
> 
> osbuild-composer should be the smart one - if `dnf-json` doesn't return `json` or returns a weird status code, that's surely 500. When it gets a nice structured error, it should examine it and return an appropriate error. I think that `rpmmd` should be only responsible for API translation (Go code <=> running python script) and the cloud server should be one to decide the mapping between an error type and HTTP error code.


> Why should the API server be responsible for mapping the error codes? Because they might be different for different APIs. Imagine an error that was caused by failing to download metadata. In the case of the cloud API, that is probably 403 because it's the caller's responsibility to pass working repos. If the caller passed broken repos, it's not composer's fault, it's just a bad request. However, in the case of the weldr API, we define the repos inside composer so we might want to return 500 because the user did nothing wrong.

Agreed, 

> On the strings vs. "subtypes" debate: If we want to return more structured errors, subtypes are probably the only way. (I kinda imagine the `dnf-json` error types to be similar to `osbuild` stages with their well-defined input schemas.) All Go types representing `dnf-json` error types should implement the `error` interface to make the code more idiomatic. `error.Is` and `error.As` methods can be used to do the mapping between `dnf-json` error type and HTTP error.

Right I'm not against subtypes. But then we'd still map the [python-dnf errors](https://dnf.readthedocs.io/en/latest/api_exceptions.html) to types in rpmmd. So you'd have rpmmd.DepsolveError, which inherits from DNFError.

Currently in this PR the api just makes the distinction between the DNFError or generic Error go types. Which is fine I think. If we need more complex parsing we'd just have to make more DNFError types. And I think you end up with string parsing regardless, but then in rpmmd if rpmmd maps it to go types.

test/cases/api.sh
lavocatt: I agree. Once we will have input validation this testcase will fail.
We can't mv the dnf-json because current user don't have access rights to the script. We need to find another way.

lavocatt: Ok, since the script has sudo access I've done the thing :)

test/cases/api.sh
```diff
@@ -953,4 +953,35 @@ curl \
...
+#!/usr/bin/python3
+raise Exception
+EOF
+[ "$(curl \
+    --silent \
+    --output /dev/null \
+    --write-out '%{http_code}' \
+    --header "x-rh-identity: $VALIDAUTHSTRING" \
+    -H "Content-Type: application/json" \
+    -d '{"distribution":"centos-8","image_requests":[{"architecture":"x86_64","image_type":"ami"}]}' \
```
Gundersanne, line 982: One last nitpick: can this be a valid request? That way we're sure the 500 is because of the dnf crash :)

lavocatt: Definitely !

Gundersanne (approved):
Nice, thank you! 

Let's let the CI do it's thing until green before merge.

PR #1565, opened 2021-07-19
replace hard coded ssh keys with generated one

jabia99:
This pull request includes:

Trying the basic functionality just with libvirt test, i'll extend it to the other tests if it works

### DIFF
--- a/test/cases/ostree-ng.sh
+++ b/test/cases/ostree-ng.sh
@@ -59,7 +59,6 @@ polkit.addRule(function(action, subject) {
 EOF
 
 # Set up variables.
-OSBUILD_COMPOSER_TEST_DATA=/usr/share/tests/osbuild-composer/
 OSTREE_REF="test/rhel/8/${ARCH}/edge"
 OS_VARIANT="rhel8-unknown"
 TEST_UUID=$(uuidgen)
@@ -84,7 +83,9 @@ COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
 # SSH setup.
 SSH_OPTIONS=(-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5)
-SSH_KEY=${OSBUILD_COMPOSER_TEST_DATA}keyring/id_rsa
+SSH_DATA_DIR=$(/usr/libexec/osbuild-composer-test/gen-ssh.sh)
+SSH_KEY=${SSH_DATA_DIR}/id_rsa
+SSH_KEY_PUB=$(cat "${SSH_KEY}".pub)
 
 case "${ID}-${VERSION_ID}" in
     "rhel-8.4")
@@ -279,7 +280,7 @@ name = "kernel-rt"
 name = "admin"
 description = "Administrator account"
 password = "\$6\$GRmb7S0p8vsYmXzH\$o0E020S.9JQGaHkszoog4ha4AQVs3sk8q0DvLjSMxoxHBKnB2FBXGQ/OkwZQfW/76ktHd0NX5nls2LPxPuUdl."
-key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC61wMCjOSHwbVb4VfVyl5sn497qW4PsdQ7Ty7aD6wDNZ/QjjULkDV/yW5WjDlDQ7UqFH0Sr7vywjqDizUAqK7zM5FsUKsUXWHWwg/ehKg8j9xKcMv11AkFoUoujtfAujnKODkk58XSA9whPr7qcw3vPrmog680pnMSzf9LC7J6kXfs6lkoKfBh9VnlxusCrw2yg0qI1fHAZBLPx7mW6+me71QZsS6sVz8v8KXyrXsKTdnF50FjzHcK9HXDBtSJS5wA3fkcRYymJe0o6WMWNdgSRVpoSiWaHHmFgdMUJaYoCfhXzyl7LtNb3Q+Sveg+tJK7JaRXBLMUllOlJ6ll5Hod root@localhost"
+key = "${SSH_KEY_PUB}"
 home = "/home/admin/"
 groups = ["wheel"]
 EOF
@@ -551,7 +552,7 @@ name = "kernel-rt"
 name = "admin"
 description = "Administrator account"
 password = "\$6\$GRmb7S0p8vsYmXzH\$o0E020S.9JQGaHkszoog4ha4AQVs3sk8q0DvLjSMxoxHBKnB2FBXGQ/OkwZQfW/76ktHd0NX5nls2LPxPuUdl."
-key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC61wMCjOSHwbVb4VfVyl5sn497qW4PsdQ7Ty7aD6wDNZ/QjjULkDV/yW5WjDlDQ7UqFH0Sr7vywjqDizUAqK7zM5FsUKsUXWHWwg/ehKg8j9xKcMv11AkFoUoujtfAujnKODkk58XSA9whPr7qcw3vPrmog680pnMSzf9LC7J6kXfs6lkoKfBh9VnlxusCrw2yg0qI1fHAZBLPx7mW6+me71QZsS6sVz8v8KXyrXsKTdnF50FjzHcK9HXDBtSJS5wA3fkcRYymJe0o6WMWNdgSRVpoSiWaHHmFgdMUJaYoCfhXzyl7LtNb3Q+Sveg+tJK7JaRXBLMUllOlJ6ll5Hod root@localhost"
+key = "${SSH_KEY_PUB}"
 home = "/home/admin/"
 groups = ["wheel"]
 EOF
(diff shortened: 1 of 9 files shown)


tools/libvirt_test.sh
```diff
@@ -82,6 +82,15 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
 COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
 COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
+# Create SSH key
+SSH_KEY=/tmp/id_rsa
+ssh-keygen -f "${SSH_KEY}" -N "" -q
+
+# Change cloud-init/user-data ssh key
+key=" -ssh-rsa $(sed "1d;38d;s/\n//g" "${SSH_KEY}"| tr -d '[:space:]')"
```
jrusz, line 90: The sed can be simlified and made a little better by `sed "$ d;1d" "${SSH_KEY}"` which will remove first and last lines only, without relying on it being exactly 38 lines.

tools/libvirt_test.sh
```diff
@@ -82,6 +82,15 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
...
+sed -i "/ssh-rsa/d" /usr/share/tests/osbuild-composer/cloud-init/user-data # In case there's a key for some reason
```
jrusz, line 91: You should remove it in the first place in the original file `test/data/cloud-intit/user-data` so you don't have to worry about this.

tools/libvirt_test.sh
```diff
@@ -82,6 +82,15 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
...
+echo "$key" >> /usr/share/tests/osbuild-composer/cloud-init/user-data
```
jrusz, line 92: That is just going to append the key at the end of the file which won't work. You need to put it under the `ssh_authorized_keys:` part of `user-data` file... 
I suggest moving the whole user part at the end and just amend it with the key ( I expect cloud-init doesn't care about the order of things there as long as the syntax is correct). Also don't forget the correct indent (adding spaces before it)

jrusz (changes requested):
Tests are currently failing, see the comments I added, that should help resolve the failures. 

Once that's done and the test is passing we can continue with next steps which would be:

1. Removing the actual ssh keys from repository and from specfile
2. Extending this to the other tests

tools/libvirt_test.sh
ondrejbudai: Writing into `/usr` always feels weird, can we do this in a temporary directory?

tools/libvirt_test.sh
```diff
@@ -82,11 +82,21 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
 COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
 COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
+WORKSPACE=/tmp/
```
jrusz, line 85: Do not override this variable, instead define it only if undefined. Example: `WORKSPACE="${WORKSPACE:-/tmp}"` and put it somewhere at the top where other vars are defined.

tools/libvirt_test.sh
```diff
@@ -82,11 +82,21 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
...
+ssh-keygen -f "${SSH_KEY}" -N "" -q -t rsa
+
+# Change cloud-init/user-data ssh key
+key=" - $(cat "${SSH_KEY}".pub)"
+cp ${OSBUILD_COMPOSER_TEST_DATA}cloud-init/user-data ${SSH_DATA}
+echo "$key" >> ${SSH_DATA}/user-data
+
 # Check for the smoke test file on the AWS instance that we start.
 smoke_test_check () {
     # Ensure the ssh key has restricted permissions.
```
jrusz, line 99: I suppose this comment no longer needs to be here..

jrusz (changes requested):
Looks good so far, tests are passing. Now you can continue expanding this to other test cases. You can also try creating some sort of common script that would take care of this and just use it in all the other tests like you suggested, I suppose that would be nice.

atodorov:
> Now you can continue expanding this to other test cases.

What's the scope of this PR? Should it cover all scripts or a single one? Depending on that you may want to merge and continue separately or keep working here.

jrusz:
> What's the scope of this PR? Should it cover all scripts or a single one? Depending on that you may want to merge and continue separately or keep working here.

During our discussions with @jabia99 we figured he's going to try it in one test first and then extend it to other tests because it's going to be pretty much the same. Since he's modifying the `user-data` for cloud-init that's used by several other tests they would fail the way they are now. So the scope is to cover all scripts in this PR.

There are some go test functions that make use of the ssh-key in the repository, those will not be changed in this PR, only the bash scripts.

jrusz (changes requested):
Why is there an empty file `test/cases/ostree` ? Just remove it please.

I added a few comments, just cosmetic changes requested. Otherwise looks good to me.

ondrejbudai (commented):
Very nice, thanks! :)

( @jrusz pointed out all minor issues. )

test/data/cloud-init/user-data
```diff
@@ -8,4 +8,4 @@ write_files:
     content: "c21va2UtdGVzdAo="
     encoding: "b64"
     owner: "root:root"
-    permissions: "0644"
\ No newline at end of file
+    permissions: "0644"
```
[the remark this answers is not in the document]
jabia99: I restored the file from main, but for some reason, the change it's still there (I think it has something to do with vim adding newline at the end of each line).

jrusz (approved):
LGTM. Nice work :)

test/cases/ostree-ng.sh
```diff
@@ -84,7 +83,9 @@ COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
 # SSH setup.
 SSH_OPTIONS=(-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5)
-SSH_KEY=${OSBUILD_COMPOSER_TEST_DATA}keyring/id_rsa
+SSH_DATA=$(/usr/libexec/osbuild-composer-test/gen-ssh.sh)
+SSH_KEY=${SSH_DATA}/id_rsa
+SSH_KEY_FILE=$(cat "${SSH_KEY}".pub)
```
atodorov, line 88: Swap variable names to actually match their contents. SSH_KEY_FILE is more appropriate for the file path, not for the file contents.

tools/libvirt_test.sh
```diff
@@ -82,11 +82,16 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
...
+
+SSH_DATA=$(/usr/libexec/osbuild-composer-test/gen-ssh.sh)
+SSH_KEY=${SSH_DATA}/id_rsa
+
 # Check for the smoke test file on the AWS instance that we start.
 smoke_test_check () {
-    # Ensure the ssh key has restricted permissions.
-    SSH_KEY=${OSBUILD_COMPOSER_TEST_DATA}keyring/id_rsa
-
+    # Ensure the ssh key has restricted permissionso.
```
atodorov, line 94: typo vs the original comment.

test/cases/ostree-ng.sh
```diff
@@ -84,7 +83,9 @@ COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
 # SSH setup.
 SSH_OPTIONS=(-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5)
-SSH_KEY=${OSBUILD_COMPOSER_TEST_DATA}keyring/id_rsa
+SSH_DATA=$(/usr/libexec/osbuild-composer-test/gen-ssh.sh)
```
atodorov, line 86: This should ideally be `SSH_DATA_DIR` but I'm not strongly opposed to the current name.

atodorov (changes requested):
A few nit picks, otherwise looks good.

tools/gen-ssh.sh
```diff
@@ -0,0 +1,27 @@
...
+# Create SSH key
+SSH_DATA_DIR="$(mktemp -d)"
+SSH_KEY=${SSH_DATA_DIR}/id_rsa
+ssh-keygen -f "${SSH_KEY}" -N "" -q -t rsa
+
+# Change cloud-init/user-data ssh key
+key=" - $(cat "${SSH_KEY}".pub)"
+# Temporary, will copy user data from cloud-init once
+# go test are updated
+tee "${SSH_DATA_dir}"/user-data > /dev/null << EOF
```
teg, line 12: This is case sensitive, so should be `SSH_DATA_DIR`, not `SSH_DATA_dir` :)

teg (changes requested):
Thanks for this! Just a small nit inline.

tools/gen-ssh.sh
jabia99: ups, thanks!

atodorov (approved):

teg (approved):

PR #1568, opened 2021-07-20
ci: upload rpms built in RHEL 8 CDN buildroot into rhel-8-cdn directory

ondrejbudai:
Let's explain how RPMs for RHEL are built:

We use a subscribed RHEL 8.x machine and mock build these on it. Mock
initializes its own buildroot based on the latest RHEL 8 CDN content, see[1].
This means that the minor version of the buildroot is independent of the minor
version of the host.

However, we currently upload RPMs to a directory whose name consists also of
the minor version of the host. Our hosts are currently running RHEL 8.3 so
the RPMs are uploaded into rhel-8.3 directory despite them being built in the
RHEL 8.4 buildroot (RHEL 8 CDN buildroot specifically). This means that
we cannot guarantee that they are installable on RHEL 8.3 which is weird.

This commit adds a special case for hosts that run on subscribed RHEL and
thus build RPMs in a buildroot constructed from RHEL CDN. These RPMs are
now uploaded into rhel-8-cdn directory. This change more accurately reflects
the way we build our RPMs and removes some confusion.

Also, we need to bump osbuild commit so we have a version that already has
the rhel-8-cdn change in it.

[1]: https://github.com/rpm-software-management/mock/blob/main/mock-core-configs/etc/mock/templates/rhel-8.tpl#L37

This pull request includes:

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -9,7 +9,7 @@
   "rhel-8.3": {
     "dependencies": {
       "osbuild": {
-        "commit": "eb74ddf2ef8c203b5d354dfb97adee65babdc147"
+        "commit": "9a33fabd5176ba2b9b944bb7272a5cdf9fb9a5cc"
       }
     },
     "dependants": {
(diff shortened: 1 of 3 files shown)


msehnout (dismissed):
Looks good to me, especially considering that the same change has already landed in osbuild repo and it is currently impossible to bump the pinning in Schutzfile without this change.

atodorov (approved):

PR #1585, opened 2021-07-23
Weldr API: introduce Image Type denylist for filtering exposed images

thozza:
Extend Weldr API to accept a list of denied image types, which should not be exposed via API for any supported distribution. This functionality will be needed to not expose image types which can't be successfully built outside of Red Hat VPN. Example of such images are the official RHEL EC2 images, which include RHUI client packages not available publicly.

Image Types are filters when listing available compose types and creating a new compose using Weldr API.

Extend osbuild-composer configuration to allow specifying the list of denied Image Types for Weldr API.

Add unit tests for implemented changes.

Add NEWS entry describing the newly introduced functionality.

**UPDATE:**
Changed the Image Type denylist in Weldr API from being applied to all distributions to being distribution-specific. Added support for using globing patterns as distribution and image type name.

This pull request includes:

  - [X] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news


bcl:
I'm not sure why this is needed. If the system is a subscribed RHEL system, even if it is CentOS, then building these image types should work, right? (Not that we are going to officially support this, but it's useful for testing and is currently possible).

I'd rather see the distro code figure out what's supported in the current environment instead of needing to manually adjust a config file.

cmd/osbuild-composer/testdata/test.toml
```diff
@@ -5,3 +5,6 @@ ca = "/etc/osbuild-composer/ca-crt.pem"
 [worker]
 allowed_domains = [ "osbuild.org" ]
 ca = "/etc/osbuild-composer/ca-crt.pem"
+
+[weldr_api]
+image_type_denylist = [ "qcow2", "vmdk" ]
```
bcl, line 10: Assuming there isn't a better way to do this (let the distro code do the filtering) this should be a mapping of distro name (or glob?) to image types. You may want to only disable `qcow2` on certain distros not on all of them.

thozza:
> I'm not sure why this is needed. If the system is a subscribed RHEL system, even if it is CentOS, then building these image types should work, right? (Not that we are going to officially support this, but it's useful for testing and is currently possible).

Unfortunately No. There are situations in which it won't work on subscribed system, regardless if it is RHEL, CentOS or Fedora. I should have probably explained the reason behind this change in a more detail.

I'm working on defining the official RHEL EC2 images, so that we can produce them as part of RHEL release. However these RHEL EC2 images include RHUI (Red Hat Update Infrastructure) client RPMs, which are not available publicly. Even if your system is subscribed, these RPMs won't be available via any CDN repository. They are unfortunately available only from internal repository behind Red Hat VPN. While this situation should be probably revisited and these RPMs should get published on CDN, it won't change for RHEL-8.5.

Note that the "problem" with RHEL EC2 images is not only about installation of a few additional RPMs, but also about applying different RHSM configuration when RHUI client RPMs are installed.

I originally planned define one EC2 image type shared by all use-cases (in reality represented by each composer API) and let the API caller to specify which subscription method should be used for the image. However this turned out to require more refactoring than originally anticipated and in order to make it with the change into 8.5, I agreed with @teg to define RHEL EC2 images as a separate image type. As a result, these image types will automatically show up in all composer APIs, however only Weldr API is problematic. Mainly because of the on-prem use case and the repo with RHUI client RPMs being private. Weldr API is a bit special compared to other APIs that we have, because the caller does not provide the complete list of repositories, which should be used for building the image. Cloud API is not used directly by any users, only via Image Builder. Image Builder explicitly limits the list of supported image types and provides the repo definitions for each image build. And lastly, the Koji API is actually the one which will be used to trigger builds of these RHEL EC2 images. The user who triggers the build has to provide all repo definitions used for building the image.

> 
> I'd rather see the distro code figure out what's supported in the current environment instead of needing to manually adjust a config file.

I think that the list of image types I used for unit testing the functionality may be a bit confusing as it may look like implying that we want to limit the list of supported image types per for specific distro or based on which distro you are running composer. However the main motivation is to have a mechanism to limit specific image types from being exposed via Weldr API (by default) and at the same time not limit composer to build them via other types of API.

So for the RHEL EC2 image types (which are not yet in composer) it can't be determined based on the current environment.

cmd/osbuild-composer/testdata/test.toml
thozza: I see what you mean and it sounds interesting. My motivation was to have a mechanism to block a specific image types (soon to be defined RHEL EC2 images with new name) from being exposed via Weldr API regardless of the distro.

Unless you feel strongly about your proposal, I would prefer to keep the logic simple for now and don't allow specifying distro name or globing.

ondrejbudai (commented):
Looks good to me, thanks! Brian has a point with the denylist not being distro-specific. We might want to make it distro-specific because things might change in the future... @thozza wdyt? It seems slightly painful, unfortunately. :/

thozza:
> Looks good to me, thanks! Brian has a point with the denylist not being distro-specific. We might want to make it distro-specific because things might change in the future... @thozza wdyt? It seems slightly painful, unfortunately. :/

No, I agree... I just wanted to save myself a bit of a stress 😄  I'll look at it, it may not be that hard in the end.

thozza:
I made the Image Type denylist distribution-specific. It is possible to filter an image type for any distributions using `*` as the distribution name. The same applies to matching image types, where `*` can be used as an image type name to filter all image types. However I'm not sure if the second case is actually useful. If not, I can remove it from the PR and keep only the ability to match any distribution name using `*`.

internal/weldr/api.go
```diff
@@ -385,13 +385,21 @@ func (api *API) openImageFile(composeId uuid.UUID, compose store.Compose) (io.Re
...
-func (api *API) checkImageTypeDenylist(imageType string) error {
-	for _, deniedImgType := range api.imageTypeDenylist {
-		if imageType == deniedImgType {
-			return fmt.Errorf("image type denied by configuration: %q", imageType)
+// checkImageTypeDenylist checks the given ImageType and Distro names against
+//the distro-specific ImageType Denylist provided to the API from configuration.
+// If the given ImageType is not allowed the method returns an `error`.
+// Otherwise `nil` is returned.
+func (api *API) checkImageTypeDenylist(distroName, imageType string) error {
+	anyImageType := "*"
```
bcl, line 393: I'd like to see this use actual globbing for the match. I can imagine uses where you'd want to exclude a image type from all RHEL 8 with 'rhel-8*' or all Fedora, etc. Looks like we're already using a glob library in rpmmd, `github.com/gobwas/glob` so that wouldn't add any new dependencies.

internal/weldr/api.go
```diff
@@ -385,13 +385,21 @@ func (api *API) openImageFile(composeId uuid.UUID, compose store.Compose) (io.Re
...
-// returns an `error`. Otherwise `nil` is returned.
-func (api *API) checkImageTypeDenylist(imageType string) error {
-	for _, deniedImgType := range api.imageTypeDenylist {
-		if imageType == deniedImgType {
-			return fmt.Errorf("image type denied by configuration: %q", imageType)
+// checkImageTypeDenylist checks the given ImageType and Distro names against
+//the distro-specific ImageType Denylist provided to the API from configuration.
+// If the given ImageType is not allowed the method returns an `error`.
+// Otherwise `nil` is returned.
+func (api *API) checkImageTypeDenylist(distroName, imageType string) error {
```
bcl, line 392: Personally I'd make this return a bool. It's not really an error, you're just asking a question. I'd also name it something like `isImageTypeAllowed`

internal/weldr/api.go
```diff
@@ -401,7 +409,7 @@ func (api *API) checkImageTypeDenylist(imageType string) error {
 // This is necessary because different distros support different image types, and the image
 // type may have a different package set than other distros.
 func (api *API) getImageType(distroName, imageType string) (distro.ImageType, error) {
-	if err := api.checkImageTypeDenylist(imageType); err != nil {
+	if err := api.checkImageTypeDenylist(distroName, imageType); err != nil {
 		return nil, err
```
bcl, line 413: If you choose to change the return type to `bool` still return an error here, just move the one from inside the current function.

Then in `composeHandler` I think it needs to include the error string from `getImageType` because we now have 3 ways to exit this and all of them show up as the same `Unknown compose type for architecture: %s` response which is going to start getting confusing (likely my fault for not doing that when I added a new error result).

bcl (changes requested):
Looks pretty good, except for those comments. I don't feel too strongly about error vs. bool but I do think globs would be useful, and the error response in composeHandler needs to be fixed.

internal/weldr/api.go
thozza: I considered such name and returning bool. I'm honestly not sure now why I decided to go with returning error. I'll change it.

internal/weldr/api.go
thozza: I agree regarding `composeHandler`. I decided to keep the API (response) changes minimal, by not propagating the error message from `getImageType()` to the API response, because there were already two distinct errors which were not propagated. But it definitely make sense to distinguish these situations by returning the appropriate error message.

internal/weldr/api.go
thozza: I think it is now obvious enough, that I didn't want to get into the "globbing business" 😄 . I'll have a look at the package that we already use in rpmmd...

bcl: I can understand not wanting to complicate things, but I think it will be a common thing to want to exclude it from a complete set of distros without needing to add new entries for each supported release.

internal/weldr/api.go
ondrejbudai: I would change the name to `isImageTypeInDenyList` and make it return `bool`. The current name feels confusing to me.

ondrejbudai (commented):
This looks good. I propose that: `checkImageTypeDenylist` is fixed in this PR, proper globbing is added in a follow-up.

ondrejbudai (approved):
Awesome, thanks for doing this! :)

internal/weldr/api.go
ondrejbudai: This is done.

internal/weldr/api.go
ondrejbudai: This is done.

internal/weldr/api.go
ondrejbudai: This is also done.

PR #1586, opened 2021-07-23
Worker: report stages status after RunOSBuild

### LINKED ISSUE #1584: Print stages result from osbuild inside the worker's journal
In the file cmd/osbuild-worker/jobimpl-osbuild.go we find that handles the call to RunOSBuild and gets the data.

```

	// Run osbuild and handle two kinds of errors
	osbuildJobResult.OSBuildOutput, err = RunOSBuild(args.Manifest, impl.Store, outputDirectory, exports, os.Stderr)
	// First handle the case when "running" osbuild failed
	if err != nil {
		return err
	}
	// Second handle the case when the build failed, but osbuild finished successfully
	if !osbuildJobResult.OSBuildOutput.Success {
		return nil
	}
```

I think we could improve the behavior of this code chunk by adding detail printing about the stages. What I have in mind is to output a stage per line in green color when the stage is successful and a red multiline chunk with the reason of the failure when the stage errors out.

lavocatt:
To help along with debugging, this commit makes the worker able to print
the status of the different stages with a oneliner for each successfull
stages and a detailed message for failed ones.

Sample output:
Jul 23[..]: Build stages results:
Jul 23[..]: org.osbuild.rpm  success
Jul 23[..]: org.osbuild.selinux  success
Jul 23[..]: Stages results:
Jul 23[..]: org.osbuild.rpm  success
Jul 23[..]: org.osbuild.fix-bls  success
Jul 23[..]: org.osbuild.fstab  success
Jul 23[..]: org.osbuild.grub2  success
Jul 23[..]: org.osbuild.locale  success
Jul 23[..]: org.osbuild.timezone  success
Jul 23[..]: org.osbuild.users failure:
Jul 23[..]:  [/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
Jul 23[..]: All rules containing unresolvable specifiers will be skipped.
Jul 23[..]: Failed to create file /sys/fs/selinux/checkreqprot: Read-only file system
Jul 23[..]: useradd: group 'toto' does not exist

Fixes #1584

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -96,6 +96,37 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	if err != nil {
 		return err
 	}
+
+	log.Println("Build stages results:")
+
+	// Include the build stages output inside the worker's logs.
+	for _, stage := range osbuildJobResult.OSBuildOutput.Build.Stages {
+		if stage.Success {
+			log.Println(stage.Name, " success")
+		} else {
+			log.Printf("%s failure:\n", stage.Name)
+			stageOutput := strings.Split(stage.Output, "\n")
+			for _, line := range stageOutput {
+				log.Printf("	%s", line)
+			}
+		}
+	}
+
+	log.Println("Stages results:")
+
+	// Include the stages output inside the worker's logs.
+	for _, stage := range osbuildJobResult.OSBuildOutput.Stages {
+		if stage.Success {
+			log.Println(stage.Name, " success")
+		} else {
+			log.Printf("%s failure:\n", stage.Name)
+			stageOutput := strings.Split(stage.Output, "\n")
+			for _, line := range stageOutput {
+				log.Printf("	%s", line)
+			}
+		}
+	}
+
 	// Second handle the case when the build failed, but osbuild finished successfully
 	if !osbuildJobResult.OSBuildOutput.Success {
 		return nil


achilleas-k (commented):
Very nice!  Would `log.Print` be more appropriate?  It looks like that's what we use to print other status messages.

lavocatt:
> Very nice! Would `log.Print` be more appropriate? It looks like that's what we use to print other status messages.

Thanks for the review. Yes it could be nicer indeed. But the default loger does not expose a stderr logging without quitting the execution. So I'll need to instantiate a new logger for the case where I want to log to stderr.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -96,6 +96,29 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	if err != nil {
 		return err
 	}
+
+	fmt.Println("Build stages results:")
+
+	// Include the build stages output inside the worker's logs.
+	for _, stage := range osbuildJobResult.OSBuildOutput.Build.Stages {
```
ondrejbudai, line 103: @achilleas-k This doesn't work for v2 manifests, does it?

Gundersanne: The worker only depends on osbuild1 though. Is a bunch of osbuild2 output ignored now when returning the jobresult, and just what's overlapped in json tags between v1 and v2 results?

achilleas-k: osbuild2 output gets unmarshalled into an osbuild1.Result object automatically.  @ondrejbudai remember this? :)

https://github.com/osbuild/osbuild-composer/blob/1abd9a11924cf20a2a3b9e2cbe173dae4b614005/internal/osbuild1/result.go#L53

achilleas-k: Sidenote: At some point (soon) I think we should change this to use v2 types primarily at which point we would need to change occurrences like this as well.

achilleas-k (dismissed):

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -37,6 +37,8 @@ func appendTargetError(res *worker.OSBuildJobResult, err error) {
 }
 
 func (impl *OSBuildJobImpl) Run(job worker.Job) error {
+	errorLog := log.New(os.Stderr, "", 0)
+	errorLog.SetFlags(log.Ldate | log.Ltime)
```
ondrejbudai, line 41: is this really needed? The standard logger is defined in this way:

```golang
const (
	LstdFlags     = Ldate | Ltime // initial values for the standard logger
)
var std = New(os.Stderr, "", LstdFlags)
```

Looks like we are instantiating the exact same thing.

See https://github.com/golang/go/blob/7ba8e796c91eaf4befcacc4d24127ae54475d6a5/src/log/log.go#L317

ondrejbudai (changes requested):

cmd/osbuild-worker/jobimpl-osbuild.go
ondrejbudai: Ah, I remember now, we need to rewrite that at some point but this PR is definitely a good step forward, thanks! :)

cmd/osbuild-worker/jobimpl-osbuild.go
lavocatt: Oh! Indeed! Imma revert this. Thanks.

ondrejbudai (approved):
Looks good, thank you for looking into this!

achilleas-k (approved):

teg (commented):
Too late to the party, but: love this! Going to help us a lot I think :)

PR #1593, opened 2021-07-27
Add RHEL 9.0 beta support

ondrejbudai:
This PR copies rhel-85 definitions into rhel-90 and enables all tests. Let's see whether Schutzbot likes it.

This pull request includes:


tools/provision.sh
```diff
@@ -6,9 +6,23 @@ source /etc/os-release
 # koji and ansible are not in RHEL repositories. Depending on them in the spec
 # file breaks RHEL gating (see OSCI-1541). Therefore, we need to enable epel
 # and install koji and ansible here.
-if [[ $ID == rhel || $ID == centos ]]; then
+if [[ ( $ID == rhel || $ID == centos ) ${VERSION_ID%.*} == 8 ]]; then
     sudo dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
     sudo dnf install -y koji ansible
+elif [[ ( $ID == rhel || $ID == centos ) ${VERSION_ID%.*} == 8 ]]; then
```
atodorov, line 12: Forgot `&&` maybe ?

ondrejbudai: Good catch, thanks.

ondrejbudai:
This is now ready to be reviewed!

Nice things:
- All image types (except for vmdk and vhd that suffers from an unworkaroundable bug, see [rhbz#1989841](https://bugzilla.redhat.com/show_bug.cgi?id=1989841)) are buildable! \o/
- qcow2 image type (the only image type supported for RHEL 9 prior this PR) was updated to use manifest v2, I checked the output from image-info and there are basically no differences.

Caveats:
- No tests are enabled, I filled a lot of bugs that we can squash one after each other, see #1630, #1631, #1632, #1633, #1634, #1635, and #1636.
- This means that images that we build might be unbootable, or buggy.

Why I chose this approach? Because if I try to fix all issues in the tests in this PR, it would be huge and it wouldn't allow us to parallelize the work on RHEL 9

ondrejbudai:
I've just manually verified that Weldr API indeed works on RHEL 9. :heart_eyes: 

image

internal/distro/rhel90/package_sets.go
```diff
@@ -113,7 +113,7 @@ func qcow2CommonPackageSet() rpmmd.PackageSet {
 			"@core", "authselect-compat", "chrony", "cloud-init",
 			"cloud-utils-growpart", "cockpit-system", "cockpit-ws",
 			"dhcp-client", "dnf", "dnf-utils", "dosfstools",
-			"insights-client", "NetworkManager", "net-tools", "nfs-utils",
```
teg: What previous definition do you mean?

internal/distro/rhel90/distro.go
```diff
@@ -460,10 +459,6 @@ func New() distro.Distro {
 	return newDistro(defaultName, modulePlatformID, ostreeRef)
 }
 
-func NewRHEL86() distro.Distro {
-	return newDistro(rhel86Name, modulePlatformID, ostreeRef)
```
teg: Couldn't this be 9.1 alias instead?

internal/distro/rhel90/pipelines.go
```diff
@@ -56,14 +56,9 @@ func qcow2Pipelines(t *imageType, customizations *blueprint.Customizations, opti
 }
 
 func prependKernelCmdlineStage(pipeline *osbuild.Pipeline, t *imageType, pt *disk.PartitionTable) *osbuild.Pipeline {
-	if t.arch.name == s390xArchName {
```
teg: Is this also an issue for 8.5, is it tracked anywhere?

cc @achilleas-k

Schutzfile
```diff
@@ -46,5 +46,12 @@
         "commit": "35de3093a7b52569512bdc61d2105febbb9b0c7e"
       }
     }
+  },
+  "rhel-9.0": {
+    "dependencies": {
+      "osbuild": {
+        "commit": "152b14a0d6c5deccfce2096308230f6d9067d23e"
```
teg, line 53: What version do we need? Why? If we are to release this into RHEL9 will we also make sure to coordinate with osbuild release? cc @gicmo

teg (approved):
Minor comments, but nothing that should block this. Usually we would not merge without tests, but considering timelines, and that this is currently completely non-functional, merging this first as a tech preview and fixing up tests afterwards is justified, IMHO.

internal/distro/rhel90/pipelines.go
ondrejbudai: Nah, this isn't an issue for 8.5. The change is in `/usr/lib/kernel/install.d/20-grub.install`, thus in `grub2-common`.

Schutzfile
ondrejbudai: Ah, I picked a very random one. osbuild 30 with 234997eeb3e0355fd67373849176a49130f9e0e7 should be enough.

internal/distro/rhel90/distro.go
ondrejbudai: Ah, right, a good thing for a follow-up.

internal/distro/rhel90/package_sets.go
ondrejbudai: The ones made by Jozef that were generating v1 manifests.

PR #1536, opened 2021-07-08
RHEL 8.5: main image types

achilleas-k:
This pull request includes:

- [x] adequate testing for the new functionality or fixed issue
- [x] adequate documentation informing people about the change such as
  - [x] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

---

Draft PR to get CI running base tests on RHEL 8.5.  Some failures are expected:
- AMI not implemented.  I intend to add the base image after #1525 is merged.
- Still generating test manifests for all non-x86_64 architectures.

---

TODO:
- [x] AMI base image:
  - [x] Update packages match official ks ([COMPOSER-840](https://issues.redhat.com/browse/COMPOSER-840))
- [x] Unit tests for new stages
- [ ] Test manifests for s390x
- [ ] Test manifests for ppc64le
- [x] Allow force-installing excluded packages when specified in the blueprint

Changes to the AMI image type and AMI variants listed in [COMPOSER-840](https://issues.redhat.com/browse/COMPOSER-840) will be added in a later PR.

PR #1594, opened 2021-07-27
Fix differences between 8.4 and 8.5 image definitions

achilleas-k:
This pull request includes:

This PR is the result of a short investigation into the differences between the image types in RHEL 8.4 and 8.5, based on the output of the `image-info` tool (as it currently exists in the test manifests).

Elements fixed in this PR:
- AMI package exclude list
~~We (well, mostly @thozza) are working towards updating the AMI definitions to match the official images (https://issues.redhat.com/browse/COMPOSER-837).  The AMI definition I added in #1536 was an initial step towards this, where I updated the package list from the 8.4 definition.  Unfortunately, I neglected to include some packages to the exclusion list.  That's been fixed now.~~
    - I reverted the change in the AMI package lists since @thozza is currently working on bringing those up to date separately.
- GRUB 2 saved entry
As a consequence of the new way the blueprint packages are handled (as a separate package set), the wrong package set was searched for to detect the kernel and add the saved GRUB 2 value.  Fixed now.

Below is a list of other differences and the reasons they appeared.

### New udev rule

Introduced in 8.5
```json
"/etc/udev/rules.d": {
  "70-persistent-ipoib.rules": []
}
```

Added by `rdma-core`.

**New package dependency for `libpcap`.**

## New rhsm options

- `auto_registration`
- `auto_registration_interval`

**Added in org.osbuild.rhsm stage**

### rhsm.conf

Removed from 8.5.

**This is expected**: Subscription manager was meant to be excluded (#893) but then was re-added (un-excluded) (#1231) to edge image types.  Now it's no longer installed even when not explicitly excluded.

### dnf

Removed from 8.5

**This is expected**: dependency of subman (see above).

### DIFF
--- a/internal/distro/rhel85/pipelines.go
+++ b/internal/distro/rhel85/pipelines.go
@@ -36,7 +36,7 @@ func qcow2Pipelines(t *imageType, customizations *blueprint.Customizations, opti
 	}
 	partitionTable := defaultPartitionTable(options, t.arch, rng)
 	treePipeline.AddStage(osbuild.NewFSTabStage(partitionTable.FSTabStageOptionsV2()))
-	treePipeline.AddStage(osbuild.NewGRUB2Stage(grub2StageOptions(&partitionTable, t.kernelOptions, customizations.GetKernel(), packageSetSpecs[osPkgsKey], t.arch.uefi, t.arch.legacy)))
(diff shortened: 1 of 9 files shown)


teg (approved):
Thanks for the fixes!

PR #1600, opened 2021-07-28
Filesystem implementation

kingsleyzissou:
This pull request includes:

- implementation for file system partitions for RHEL 8.5

### DIFF
--- a/internal/disk/customizations.go
+++ b/internal/disk/customizations.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
-	"github.com/osbuild/osbuild-composer/internal/distro"
 )
 
 const (
@@ -31,42 +30,41 @@ func CreatePartitionTable(
 	rng *rand.Rand,
 ) PartitionTable {
 
-	basePartitionTable.Size = imageSize
-	partitions := []Partition{}
-
 	if bootPartition := basePartitionTable.BootPartition(); bootPartition != nil {
 		// the boot partition UUID needs to be set since this
 		// needs to be randomly generated
 		bootPartition.Filesystem.UUID = uuid.Must(newRandomUUIDFromReader(rng)).String()
 	}
 
+	for _, m := range mountpoints {
+		if m.Mountpoint != "/" {
+			partitionSize := uint64(m.MinSize) / sectorSize
+			partition := basePartitionTable.createPartition(m.Mountpoint, partitionSize, rng)
+			basePartitionTable.Partitions = append(basePartitionTable.Partitions, partition)
+		}
+	}
+
+	if tableSize := basePartitionTable.getPartitionTableSize(); imageSize < tableSize {
+		imageSize = tableSize
+	}
+
+	basePartitionTable.Size = imageSize
+
 	// start point for all of the arches is
 	// 2048 sectors.
 	var start uint64 = basePartitionTable.updatePartitionStartPointOffsets(2048)
 
-	for _, m := range mountpoints {
-		if m.Mountpoint != "/" {
-			partitionSize := uint64(m.MinSize) / sectorSize
-			partition := createPartition(m.Mountpoint, partitionSize, start, archName, rng)
-			partitions = append(partitions, partition)
-			start += uint64(m.MinSize / sectorSize)
-		}
-	}
-
 	// treat the root partition as a special case
 	// by setting the size dynamically
 	rootPartition := basePartitionTable.RootPartition()
-	rootPartition.Start = start
 	rootPartition.Size = ((imageSize / sectorSize) - start - 100)
 	rootPartition.Filesystem.UUID = uuid.Must(newRandomUUIDFromReader(rng)).String()
-
 	basePartitionTable.updateRootPartition(*rootPartition)
-	basePartitionTable.Partitions = append(basePartitionTable.Partitions, partitions...)
 
 	return basePartitionTable
 }
 
-func createPartition(mountpoint string, size uint64, start uint64, archName string, rng *rand.Rand) Partition {
+func (pt *PartitionTable) createPartition(mountpoint string, size uint64, rng *rand.Rand) Partition {
 	filesystem := Filesystem{
 		Type:         "xfs",
 		UUID:         uuid.Must(newRandomUUIDFromReader(rng)).String(),
@@ -75,15 +73,13 @@ func createPartition(mountpoint string, size uint64, start uint64, archName stri
 		FSTabFreq:    0,
 		FSTabPassNo:  0,
 	}
-	if archName == distro.Ppc64leArchName || archName == distro.S390xArchName {
+	if pt.Type != "gpt" {
 		return Partition{
-			Start:      start,
 			Size:       size,
 			Filesystem: &filesystem,
 		}
 	}
 	return Partition{
-		Start:      start,
 		Size:       size,
 		Type:       FilesystemDataGUID,
 		UUID:       uuid.Must(newRandomUUIDFromReader(rng)).String(),
--- a/internal/disk/customizations.go
+++ b/internal/disk/customizations.go
@@ -1,0 +1,179 @@
+package disk
+
+import (
+	"io"
+	"math/rand"
+
+	"github.com/google/uuid"
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/distro"
+)
+
+const (
+	sectorSize            = 512
+	BIOSBootPartitionGUID = "21686148-6449-6E6F-744E-656564454649"
+	BIOSBootPartitionUUID = "FAC7F1FB-3E8D-4137-A512-961DE09A5549"
+
+	FilesystemDataGUID = "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
+	FilesystemDataUUID = "CB07C243-BC44-4717-853E-28852021225B"
+
+	EFISystemPartitionGUID = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+	EFISystemPartitionUUID = "68B2905B-DF3E-4FB3-80FA-49D1E773AA33"
+	EFIFilesystemUUID      = "7B77-95E7"
+
+	RootPartitionUUID = "6264D520-3FB9-423F-8AB8-7A0A8E3D3562"
+)
+
+func CreatePartitionTable(
+	mountpoints []blueprint.FilesystemCustomization,
+	imageOptions distro.ImageOptions,
+	arch distro.Arch,
+	basePartitionTable PartitionTable,
+	bootType distro.BootType,
+	ec2 bool,
+	rng *rand.Rand,
+) PartitionTable {
+	archName := arch.Name()
+
+	basePartitionTable.Size = imageOptions.Size
+	partitions := []Partition{}
+	var start uint64 = 2048
+
+	if archName == distro.X86_64ArchName {
+		biosBootPartition := createPartition("bios", 2048, start, archName, rng)
+		partitions = append(partitions, biosBootPartition)
+		start += biosBootPartition.Size
+		if bootType != distro.LegacyBootType {
+			bootEFIPartition := createPartition("/boot/efi", 204800, start, archName, rng)
+			partitions = append(partitions, bootEFIPartition)
+			start += bootEFIPartition.Size
+		}
+	} else if archName == distro.Aarch64ArchName {
+		if ec2 {
+			bootEFIParition := createPartition("/boot/efi", 409600, start, archName, rng)
+			partitions = append(partitions, bootEFIParition)
+			start += bootEFIParition.Size
+			bootPartition := createPartition("/boot", 1048576, start, archName, rng)
+			partitions = append(partitions, bootPartition)
+			start += bootPartition.Size
+		} else {
+			bootEFIPartition := createPartition("/boot/efi", 204800, start, archName, rng)
+			partitions = append(partitions, bootEFIPartition)
+			start += bootEFIPartition.Size
+		}
+	} else if archName == distro.Ppc64leArchName {
+		biosBootPartition := createPartition("bios", 8192, start, archName, rng)
+		partitions = append(partitions, biosBootPartition)
+		start += biosBootPartition.Size
+	}
+
+	for _, m := range mountpoints {
+		if m.Mountpoint != "/" {
+			partitionSize := uint64(m.MinSize) / sectorSize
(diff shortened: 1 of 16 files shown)


internal/distro/rhel85/distro.go
```diff
@@ -353,6 +355,15 @@ func (t *imageType) sources(packages []rpmmd.PackageSpec, ostreeCommits []ostree
 	return sources
 }
 
+func checkMountpoint(mountpoint string) bool {
```
teg, line 358: We should check for prefixes, so `/var/log` should be accepted, but not `/variable`.

kingsleyzissou: Thanks, been thinking how to do that and regex is probably the way to go here

teg: I wonder if Golang has any nice file path helper library?

ondrejbudai: Might be an overkill: https://github.com/spf13/afero

kingsleyzissou: Ahhh I think golang path will be enough for this actually. Just writing a couple of tests.

internal/distro/rhel85/partition_tables.go
```diff
@@ -5,230 +5,317 @@ import (
...
+}
+
+var ppc64lePartitionTable = disk.PartitionTable{
+	UUID: "0x14fc63d2",
+	Type: "dos",
+	Partitions: []disk.Partition{
+		{
+			Size:     8192,
+			Type:     "41",
+			Bootable: true,
```
achilleas-k, line 42: This changed in a recent PR and has a `Start` offset now.  Something to keep in mind when rebasing.

internal/distro/rhel85/partition_tables.go
```diff
@@ -5,230 +5,317 @@ import (
...
+		Type:  "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
+		UUID:  "68B2905B-DF3E-4FB3-80FA-49D1E773AA33",
+		Filesystem: &disk.Filesystem{
+			Type:         "xfs",
+			UUID:         "2AE383D1-F57C-4F04-A1BD-32FC12A578EA",
+			Mountpoint:   "/boot",
+			FSTabOptions: "defaults",
+			FSTabFreq:    0,
+			FSTabPassNo:  0,
+		},
```
achilleas-k, line 317: Might be cleaner if the common parts were set at first and then the three values that differ are set conditionally.  It would make it easier to identify how the two cases differ when following the code.

achilleas-k (commented):
Turning out nicely I think.  One small style comment (bit of a matter of taste I imagine) and another note to look out for when eventually rebasing.

internal/distro/rhel85/partition_tables.go
kingsleyzissou: Oh yeah this is a good point. Thanks for that!

internal/distro/rhel85/partition_tables.go
```diff
@@ -324,6 +238,86 @@ func createPartition(
...
+		}
+	}
+	return disk.Partition{
+		Start: start,
+		Size:  size,
+		Type:  "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
+		UUID:  "68B2905B-DF3E-4FB3-80FA-49D1E773AA33",
+		Filesystem: &disk.Filesystem{
+			Type:         "xfs",
+			UUID:         "2AE383D1-F57C-4F04-A1BD-32FC12A578EA",
```
ondrejbudai, line 312: Just a minor note: xfs partitions must have a random uuid because there's a global table in kernel.

kingsleyzissou: I had no idea. Thanks for that, I'll update it.

thozza (commented):
Very nice work, I like where this is going. I added some inline comments, but also have some more general ideas:

1. Most of the logic in `internal/distro/rhel85/partition_tables.go` could and should be reused across various distro definitions. Going forward at least for 8.5, 9.0 and ideally also any new Fedora definition. For this reason it would be good to consider moving some of the functions (e.g. `createPartitionTable()`) into `internal/disk` package or somewhere else and exporting them. This would require moving also some constant definitions outside of the distro definition into appropriate packages and exporting them, but this should not be really an issue.
2. It would be great if we could have a declarative format for specifying the per-architecture default partitioning for an image type. Now this is effectively captured in the `createPartitionTable()` and `ec2PartitionTable()` functions. My idea would be to have the defaults defined per image type and have only the `createPartitionTable()` which would merge the default partitioning with Blueprint customisations. Since most of the image types share the default partitioning, we could have it defined as shared struct literal. And for any image types which need something special, like the **ec2** images, for them we could simply define a custom partitioning per-architecture in the image type definition. This would also work nicely if the partitioning is defined only for a subset of architectures (**ec2** case), where the non-existence of the default partitioning for the architecture would result in an error. I understand that this may not be really trivial, however I do believe that it would make it easier to share code logic across various distro definitions and help us to get to a declarative representation of distros / arch / image types.

What do you think?

internal/distro/rhel85/partition_tables.go
```diff
@@ -5,231 +5,212 @@ import (
...
+func createPartition(mountpoint string, size uint64, start uint64, archName string, rng *rand.Rand) disk.Partition {
+	var filesystem disk.Filesystem
+	// /boot/efi mountpoint is a special case
+	// return early
+	if mountpoint == "/boot/efi" {
+		filesystem = createFilesystemDisk(mountpoint, "7B77-95E7")
+		return disk.Partition{
+			Start:      start,
+			Size:       size,
+			Type:       "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah I think that is a great suggestion! Thanks for that, I was toying with the idea, but didn't want to change things too much and then ended up changing almost everything. Moving thinks to `internal/disk` package makes a lot of sense.

internal/distro/rhel85/partition_tables.go
```diff
@@ -5,231 +5,212 @@ import (
...
-					},
-				},
-			},
+const sectorSize = 512
+
+var x86PartitionTable = disk.PartitionTable{
+	UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
+	Type: "gpt",
+	Partitions: []disk.Partition{
+		{
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah another very sensible suggestion! It definitely makes more sense to do things like you are suggesting. I'll sort that out.

kingsleyzissou:
Thanks for the feedback on this @thozza, very much appreciated. 

For point number 1, I completely agree with you on that and will make those changes. For point number 2 I may need to pick your brains a little bit on that, but I will maybe message you tomorrow.

internal/distro/rhel85/partition_tables.go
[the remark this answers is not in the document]
thozza: Each image type / architecture already defines its `BootType`. You can use that to map the existing partitioning schemes to each `BootType`. Feel free to reach out to me if you have any questions. I have been messing up with these parts recently so I may be able to save you some time to figure things out.

internal/distro/rhel85/partition_tables.go
```diff
@@ -5,231 +5,231 @@ import (
...
-					},
-				},
-			},
+const (
+	sectorSize = 512
+
+	biosBootPartitionType = "21686148-6449-6E6F-744E-656564454649"
+	biosBootPartitionUUID = "FAC7F1FB-3E8D-4137-A512-961DE09A5549"
+
+	bootPartitionType = "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
```
thozza, line 19: I would suggest to use variable names based on https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs, instead of how we use them in the code (IOW for which mountpoint).

internal/distro/rhel85/distro.go
```diff
@@ -212,7 +197,9 @@ type imageType struct {
 	// bootable image
 	bootable bool
 	// If set to a value, it is preferred over the architecture value
-	bootType BootType
+	bootType distro.BootType
+	// List of valid arches for the image type
+	validArches ValidArches
```
thozza, line 214: I have basically the same comment to the comment and structure member name. While it is true that any architectures not listed in the map are not supported by the image type **if** it requires a partition table, the name is misleading. Something more descriptive would be IMO better. E.g. `basePartitionTables` or `defaultPartitionTables`, or anything else that comes to your mind 😉

internal/disk/customizations.go
```diff
@@ -0,0 +1,179 @@
...
+	FilesystemDataUUID = "CB07C243-BC44-4717-853E-28852021225B"
+
+	EFISystemPartitionGUID = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+	EFISystemPartitionUUID = "68B2905B-DF3E-4FB3-80FA-49D1E773AA33"
+	EFIFilesystemUUID      = "7B77-95E7"
+
+	RootPartitionUUID = "6264D520-3FB9-423F-8AB8-7A0A8E3D3562"
+)
+
+func CreatePartitionTable(
```
thozza, line 27: The `disk` package should be IMO agnostic of existing image types and therefore its exported functions should not contain arguments and logic specific to an image type. My idea was that the provided `basePartitionTable` should contain the "default" partition table, including default partitioning layout, for the specific image type and architecture. This however means that the function may need to potentially recalculate some values - e.g. size of the root partition.

internal/distro/rhel85/partition_tables.go
```diff
@@ -1,244 +1,41 @@
...
+	distro.S390xArchName:   dosPartitionTable,
+}
 
-var x86PartitionTable = disk.PartitionTable{
-	UUID:       "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
-	Type:       "gpt",
-	Partitions: []disk.Partition{},
+var ec2ValidArches = map[string]disk.PartitionTable{
+	distro.X86_64ArchName:  gptPartitionTable,
+	distro.Aarch64ArchName: gptPartitionTable,
```
thozza: So my idea when previously making the suggestion was that `defaultArches` and `ec2ValidArches` would hold the default partitioning scheme, probably very similarly defined as it is done currently https://github.com/osbuild/osbuild-composer/blob/1e2ba4da64190b3ada6d651192ac1bb4d53cd60c/internal/distro/rhel85/partition_tables.go#L12-L145

Then the `disk.CreatePartitionTable()` would basically just combine it with the provided Blueprint customisations and produce a potentially different version of the provided partition table on the output.

My idea was also that we could potentially omit some values in the definition of default / base partitioning table and le the `disk.CreatePartitionTable()` finalise the partitioning table before it is usable. This could be the start offset (it should be possible to calculate it based on the provided list of partitions and their sizes). and UUIDs. This is however not really that important.

PS: I would also suggest to make the names of these variables more self-explanatory and based more on their usage (default per-arch partitioning scheme).

thozza (commented):
Very nice iteration ❤️ . I like it! Added some inline comments...

internal/disk/customizations.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if bootPartition := basePartitionTable.BootPartition(); bootPartition != nil {
+		// the boot partition UUID needs to be set since this
+		// needs to be randomly generated
+		bootPartition.Filesystem.UUID = uuid.Must(newRandomUUIDFromReader(rng)).String()
+	}
+
+	// start point for all of the arches is
+	// 2048 sectors.
+	var start uint64 = basePartitionTable.updatePartitionStartPointOffsets(2048)
```
thozza, line 46: It would be great if you manage to handle the root partition inside of `updatePartitionStartPointOffsets()` to not split the logic into two functions, especially when only one is method of the `PartitionTable` type.

thozza (commented):
Very nice! I love how much you simplified the `CreatePartitionTable()` function. Added some comments inline....

internal/distro/rhel85/distro.go
```diff
@@ -51,23 +53,9 @@ const (
...
 
-type BootType string
-
-const (
-	UnsetBootType  BootType = ""
-	LegacyBootType BootType = "legacy"
-	UEFIBootType   BootType = "uefi"
-	HybridBootType BootType = "hybrid"
-)
+var mountpointAllowList = []string{"/", "/var", "/var/*", "/home", "/opt", "/srv", "/usr"}
```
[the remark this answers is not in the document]
kingsleyzissou: Good point, I guess I could move that to the `distro` package with the other commonly used types when I add the support for `rhel9`

internal/disk/customizations.go
```diff
@@ -0,0 +1,104 @@
...
+			partitions = append(partitions, partition)
+			start += uint64(m.MinSize / sectorSize)
+		}
+	}
+
+	// treat the root partition as a special case
+	// by setting the size dynamically
+	rootPartition := basePartitionTable.RootPartition()
+	rootPartition.Start = start
+	rootPartition.Size = ((imageSize / sectorSize) - start - 100)
```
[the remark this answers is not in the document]
kingsleyzissou: You're right, it doesn't actually do a check. I can add a check and a test to cover this.

However, the image size for all the partitions has been taken into account, 
his was set in a previous pr. We add up all the partition sizes and and update the size here: 
https://github.com/osbuild/osbuild-composer/blob/42480f7b265e8619dfb42fc0f6aa246e5e9d3421/internal/weldr/api.go#L2283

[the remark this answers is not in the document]
kingsleyzissou: Actually, since the custom sizes are `minimum` sizes, maybe we can increase the partition table size rather than throwing an error, @teg what are your thoughts?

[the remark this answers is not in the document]
thozza: Thanks for pointing this out. Nevertheless it may be still a good idea to make sure that we don't create buggy partitioning table even though there is currently no code path when this could happen. This may become an issue once we allow customising the filesystem via other APIs.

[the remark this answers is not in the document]
teg: I agree, increase to make things fit. However, if this is handle elsewhere, so the code path would never be taken, throwing an error for now might make the most sense.

internal/distro/rhel85/distro.go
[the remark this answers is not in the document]
teg: FWIW, I think we should keep policy per-distro to make it easier to change with having to worry about side effects

teg (approved):
This is really great! Only thing I would suggest we discuss is the actual policy, but that can be done in a follow up. We should verify we covered all the mount points we wanted, and I think all but / should include the asterisk too to cover all submounts. If nothing else to make the UI easier.

PR #1602, opened 2021-07-29
[rhel-8.4.0] make composer work with multiple subscriptions

msehnout:
Not so WIP any more, but still requires thorough reviews as the code didn't fit very well on the old one so the rebase was a bit tricky.

This pull request includes:


internal/distro/fedora33/distro.go
```diff
@@ -238,6 +243,11 @@ func sources(packages []rpmmd.PackageSpec) *osbuild.Sources {
 		fileSource := osbuild.FileSource{
 			URL: pkg.RemoteLocation,
 		}
+		if pkg.Secrets == "org.osbuild.rhsm" {
+			fileSource.Secrets = &osbuild.Secret{
+				Name: "org.osbuild.rhsm",
+			}
+		}
```
thozza, line 250: Please explain why this code block is needed for Fedora distro and why it is not included in the `fedora32` distro definition.

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -0,0 +1,317 @@
...
+done
+
+sudo journalctl -xe --unit osbuild-composer
+sudo journalctl -xe --unit osbuild-worker
+
+# Did the compose finish with success?
+if [[ $COMPOSE_STATUS != FINISHED ]]; then
+    echo "Something went wrong with the compose. 😢"
+    exit 1
+fi
```
thozza, line 317: missing newline at the end of the file.

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -0,0 +1,317 @@
...
+    }
+  ]
+}
+STOPHERE
+
+sudo systemctl restart osbuild-composer
+sudo composer-cli status show
+
+BLUEPRINT_FILE=/tmp/bp.toml
+BLUEPRINT_NAME=zishy
```
thozza, line 266: Interesting name 😄

thozza (commented):
Overall this looks good. Nice work! I added a few comments.

internal/distro/fedora33/distro.go
msehnout: Because even Fedora systems can have subscriptions. E.g. to private Satellite repos or RHEL repos in case of cross-distro building.

test/cases/regression-composer-works-behind-satellite.sh
msehnout: `Fixing whitespaces, what else would I do ...`

test/cases/regression-composer-works-behind-satellite.sh
msehnout: Hehe, `fishy` image, but with zsh :-D

internal/distro/fedora33/distro.go
thozza: So why it is not add it also to F32? Because it is EOL?

msehnout: Sorry, I forgot about that part. Yes, it is EOL so I skipped it.

test/cases/regression-composer-works-behind-satellite.sh
msehnout: Do you consider this issue so pressing it needs a respin?

thozza: No, I can live with it. Please just make sure that the issue is not present on `main`.

thozza (approved):

PR #1603, opened 2021-07-29
RHEL 8.5: fixes to image definitions

achilleas-k:
This PR fixes up some small (but important) oversights in the image definitions added in #1536.
1. OS image tree pipeline did not define a build root and was instead being built on the host
2. SELinux stage is always the last stage in a pipeline
3. Updated test manifests to reflect 1 & 2


ondrejbudai (approved):

thozza (approved):
❤️

PR #1607, opened 2021-07-29
RHEL-8.5: redefine ami image and add RHEL ec2 image types

thozza:
Redefine RHEL-8.5 `ami` image based on the official RHEL ec2 images.
Add definitions of RHEL `ec2` and `ec2-ha` image types for RHEL-8.5.

This pull request includes:

  - unit tests and image test cases were added
    - will do it in a follow up PR next week.


internal/distro/rhel85/partition_tables.go
```diff
@@ -143,6 +143,94 @@ func defaultPartitionTable(imageOptions distro.ImageOptions, arch distro.Arch, r
 	panic("unknown arch: " + arch.Name())
 }
 
+func ec2PartitionTable(imageOptions distro.ImageOptions, arch distro.Arch, rng *rand.Rand) disk.PartitionTable {
+	var sectorSize uint64 = 512
+	if arch.Name() == "x86_64" {
+		return disk.PartitionTable{
+			Size: imageOptions.Size,
+			UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
+			Type: "gpt",
```
ondrejbudai, line 152: is it really `gpt`, non-uefi partition table? :O Not an expert, isn't that uncommon?

thozza: I must admit that I'm no expert either. I basically used values from image-info report for current RHEL ec2 images:
```json
  "partition-table": "gpt",
  "partition-table-id": "A03D2C64-F209-407F-B699-3E8519C9771E",
  "partitions": [
    {
      "bootable": false,
      "fstype": null,
      "label": null,
      "partuuid": "21D40DD5-AC59-41B9-A1A8-21C9FE02C3CB",
      "size": 1048576,
      "start": 1048576,
      "type": "21686148-6449-6E6F-744E-656564454649",
      "uuid": null
    },
    {
      "bootable": false,
      "fstype": "xfs",
      "label": null,
      "partuuid": "E6934A8A-B0C2-4FC3-8170-E872167C3446",
      "size": 6442450944,
      "start": 2097152,
      "type": "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
      "uuid": "6c1f2123-a6e2-4b97-8511-47913c094f88"
    }
  ],
```

I'm happy to investigate this further if you have any pointers or ideas what to look for.

internal/distro/rhel85/distro.go
```diff
@@ -252,8 +266,38 @@ func (t *imageType) PackageSets(bp blueprint.Blueprint) map[string]rpmmd.Package
...
+			addUEFIBootPkg = true
+		case HybridBootType:
+			addLegacyBootPkg = true
+			addUEFIBootPkg = true
+		default:
+			panic(fmt.Sprintf("unsupported boot type: %q", bt))
+		}
+
+		if addLegacyBootPkg {
+			if set, exists := distroSets[bootLegacyPkgsKey]; exists {
```
achilleas-k, line 285: I think this check is unnecessary because, if it doesn't exist, the `Append()` method gets an empty list and does nothing.  it's not an issue, just might be simpler to read without the check.

achilleas-k (commented):
Just looked at the boot package set handling and I like it.  One small comment.
But thanks for bringing this up, I think it's all cleaner now.

internal/distro/rhel85/distro.go
thozza: Thanks. I simplified it!

thozza:
There are still a few small discrepancies between the official RHEL ec2 image and these ec2 image definitions, but nothing critical IMO. These versions should be close enough to the official images to be good for QE testing...

achilleas-k (approved):
Great work.  Thanks for fixing up the mess with the inputs.  It's all so much cleaner now.

Small question: Are we keeping the `ami` image type around for backwards compatibility?  It's the same image as the ec2 one just not compressed, right?

thozza:
> Great work. Thanks for fixing up the mess with the inputs. It's all so much cleaner now.

Thanks. I'm glad that you like it. There are still more inputs and stages to fix 😄 

> Small question: Are we keeping the `ami` image type around for backwards compatibility? It's the same image as the ec2 one just not compressed, right?

The difference is the output format, but also the RHUI client packages and part of the subscription-manager configuration. The plan (not for 8.5) would be to merge them and make the `ami` be an alias for `ec2` and have a way to tell composer to compress the image. Nevertheless as you've noticed, I made the `ami` to share majority of the pipeline with `ec2` images so that the experience is very similar with both images and the switch is not so painful in the future.

internal/distro/rhel85/partition_tables.go
thozza: The partitioning is the same as on the official RHEL images and the image boots successfully. I'm marking this comment as resolved.

achilleas-k:
> > Great work. Thanks for fixing up the mess with the inputs. It's all so much cleaner now.
> 
> Thanks. I'm glad that you like it. There are still more inputs and stages to fix 

We'll get to them all in time, I'm sure :)

> > Small question: Are we keeping the `ami` image type around for backwards compatibility? It's the same image as the ec2 one just not compressed, right?
> 
> The difference is the output format, but also the RHUI client packages and part of the subscription-manager configuration. The plan (not for 8.5) would be to merge them and make the `ami` be an alias for `ec2` and have a way to tell composer to compress the image. Nevertheless as you've noticed, I made the `ami` to share majority of the pipeline with `ec2` images so that the experience is very similar with both images and the switch is not so painful in the future.

Perfect plan :+1:

internal/osbuild2/files_input.go
```diff
@@ -1,18 +1,102 @@
...
+// `"name:<pipeline_name>": {"file": "<filename>"}`
+type FilesInputReferencesPipeline map[string]FileReference
+
+func (*FilesInputReferencesPipeline) isFilesInputReferences() {}
+
+type FileReference struct {
+	File string `json:"file"`
+}
+
+func NewFilesInputReferencesPipeline(pieline, filename string) FilesInputReferences {
```
ondrejbudai, line 98: `pieline` sounds yummy, I like it!

internal/distro/rhel85/package_sets.go
```diff
@@ -64,15 +64,22 @@ func x8664InstallerBuildPackageSet() rpmmd.PackageSet {
 
 // BOOT PACKAGE SETS
 
-// x86_64 arch-specific boot package set
-func x8664BootPackageSet() rpmmd.PackageSet {
+// x86_64 Legacy arch-specific boot package set
+func x8664LegacyBootPackageSet() rpmmd.PackageSet {
 	return rpmmd.PackageSet{
-		Include: []string{"dracut-config-generic", "grub2-pc", "grub2-efi-x64", "shim-x64"},
+		Include: []string{"dracut-config-generic", "grub2-pc"},
```
ondrejbudai, line 70: isn't `dracut-config-generic` needed for all images? Otherwise, `dracut` builds `initramfs` specific for the host. Not sure where to put it though. Maybe in both legacy and uefi sets?

ondrejbudai (approved):
Two minor comments, otherwise this looks awesome!

internal/distro/rhel85/package_sets.go
thozza: Yeah, I was not sure. I can put it in both sets. Having a duplicated package should not affect the depsolved package set.

ondrejbudai: sgtm, it will be less confusing if we ever want to build uefi-only images.

ondrejbudai: Let's fix in a follow-up.

internal/osbuild2/files_input.go
ondrejbudai: Let's fix in a follow-up.

ondrejbudai (approved):

PR #1608, opened 2021-07-30
internal/osbuild2: add stages for raw images and grub iso

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

This pull request includes:

This PR implements all the needed stages to create raw images and the new grub.iso stage needed for the new edge simplified installer. I'll follow up with a PR to create the pipelines (I think it's easier to review like this)

### DIFF
--- a/internal/osbuild2/ostree_config_stage.go
+++ b/internal/osbuild2/ostree_config_stage.go
@@ -1,0 +1,27 @@
+package osbuild2
+
+type OSTreeConfigOptions struct {
+	Sysroot SysrootOptions `json:"sysroot"`
+}
+
+type SysrootOptions struct {
+	ReadOnly bool `json:"readonly"`
+}
+
+// Options for the org.osbuild.ostree.config stage.
+type OSTreeConfigStageOptions struct {
+	// Location of the ostree repo
+	Repo string `json:"repo"`
+
+	Config OSTreeConfigOptions `json:"config"`
+}
+
+func (OSTreeConfigStageOptions) isStageOptions() {}
+
+// A new org.osbuild.ostree.init stage to create an OSTree repository
+func NewOSTreeConfigStage(options *OSTreeConfigStageOptions) *Stage {
+	return &Stage{
+		Type:    "org.osbuild.ostree.config",
+		Options: options,
+	}
+}
--- a/internal/osbuild2/ostree_deploy_stage.go
+++ b/internal/osbuild2/ostree_deploy_stage.go
@@ -1,0 +1,28 @@
+package osbuild2
+
+// Options for the org.osbuild.ostree.deploy stage.
+type OSTreeDeployStageOptions struct {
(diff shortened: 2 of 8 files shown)


runcom:
@gicmo @achilleas-k ptal

internal/osbuild2/ostree_config_stage.go
```diff
@@ -0,0 +1,27 @@
+package osbuild2
+
+type ConfigOptions struct {
```
achilleas-k, line 3: We should probably rename this to something like `OSTreeConfig` to avoid future collisions with other config structs in the package and to make it easier to know which stage it belongs to.  The name can get awkward with all these structs but it's generally better to pseudo-namespace them I think.

internal/osbuild2/ostree_deploy_stage.go
```diff
@@ -0,0 +1,29 @@
...
+
+	OsName string `json:"osname"`
+
+	Ref string `json:"ref"`
+
+	Mounts []string `json:"mounts"`
+
+	Rootfs Rootfs `json:"rootfs"`
+
+	KernelOpts []string `json:"kernel_opts"`
```
achilleas-k, line 14: `Mounts` and `KernelOpts` should be marked `omitempty`.  It probably doesn't make a difference functionally but it can sometimes so we'd rather mirror the stage schema as accurately as possible for this.

(I'm sure there are a lot of cases where we don't do this though)

internal/osbuild2/ostree_fillvar_stage.go
```diff
@@ -0,0 +1,24 @@
+package osbuild2
+
+// Options for the org.osbuild.ostree.fillvar stage.
+type OSTreeFillvarStageOptions struct {
+
+	Deployment Deployment `json:"deployment"`
+}
+
+type Deployment struct {
```
achilleas-k, line 9: Same as with `ConfigOptions`, should probably rename this to `OSTreeDeployment`.

achilleas-k (commented):
Great work.  Thanks for this!

I have some very minor comments.

internal/osbuild2/ostree_fillvar_stage.go
runcom: fixed!

internal/osbuild2/ostree_deploy_stage.go
runcom: fixed!

internal/osbuild2/ostree_config_stage.go
runcom: fixed!

runcom:
comments addressed thanks @achilleas-k

teg (approved):
Nice stuff!

gicmo:
Obsoleted by https://github.com/osbuild/osbuild-composer/pull/1654

PR #1616, opened 2021-08-04
[RHEL-8.5] ec2 images followup

thozza:
This is a follow-up to #1607

Notable changes:
- image-info: mount partitions in correct order when analysing image
- composer: don't expose ec2 and ec2-ha RHEL-8.5 images via WeldrAPI
- docs/news: reflect EC2 image changes in the RHEL-8.5 images note
- rhel85: include dracut-config-generic in x86 UEFI boot package set

This pull request includes:

  - unit tests were extended / modified where it was appropriate
    - extended an existing NEWS entry related to RHEL-8.5 image types

### DIFF
--- a/cmd/osbuild-composer/config.go
+++ b/cmd/osbuild-composer/config.go
@@ -28,9 +28,11 @@ type ComposerConfigFile struct {
 	ComposerAPI struct {
 		IdentityFilter []string `toml:"identity_filter"`
 	} `toml:"composer_api"`
-	WeldrAPI struct {
-		DistroConfigs map[string]WeldrDistroConfig `toml:"distros"`
-	} `toml:"weldr_api"`
+	WeldrAPI WeldrAPIConfig `toml:"weldr_api"`
+}
+
+type WeldrAPIConfig struct {
+	DistroConfigs map[string]WeldrDistroConfig `toml:"distros"`
 }
 
 type WeldrDistroConfig struct {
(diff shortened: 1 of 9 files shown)


cmd/osbuild-composer/config.go
```diff
@@ -51,17 +53,35 @@ func (c *ComposerConfigFile) weldrDistrosImageTypeDenyList() map[string][]string
...
 }
 
+// GetDefaultConfig returns the default configuration of osbuild-composer
+// Defaults:
+// - 'ec2' and 'ec2-ha' image types on 'rhel-85' are not exposed via Weldr API
+func GetDefaultConfig() *ComposerConfigFile {
+	return &ComposerConfigFile{
+		WeldrAPI: WeldrAPIConfig{
+			map[string]WeldrDistroConfig{
+				"rhel-85": {
```
ondrejbudai, line 63: I wonder if we should make this more generic... `rhel-8*`? maybe even `rhel-*`?

ondrejbudai (approved):
Amazing! One comment is inline but I'm not sure if we want to do something about it. Up to you. ;)

cmd/osbuild-composer/config.go
thozza: Probably won't make much of a difference, but it will be one less thing to keep in mind when we will be adding support for 8.6. Ideally the `ami` should be an alias to `ec2` in the future and we should have a way to optionally compress the image or choose to install RHUI client instead of using RHSM. At that point we won't need this default configuration any more and will be able to expose the image type via all APIs, but we are not there yet. I'll make the change, probably for all RHEL version using `rhel-*`.

ondrejbudai (approved):

PR #1620, opened 2021-08-05
hack: remove podman from RHEL for Edge

gicmo:
The `podman` package seems to conflict with `docker-ce`, so in order to enable the testing of RHEL for Edge with a proprietary software that needs docker, remove the hard `podman` dependency.

DO NOT MERGE.

If we actually wanted to support a different container engine, the correct way to do it, would be to do what we did for the kernel, where we have an implicit default that can be overwritten by the blueprint.


henrywang:
@gicmo Requiring `podman` comes from PM's [`Edge MVP user stories`](https://docs.engineering.redhat.com/display/VIRTCLOUD/RHEL+for+Edge+Test+Plans). Removing `podman` will be a break change.

gicmo:
@henrywang This is not intended to be merged. It is for a demo of a client that wants to test RHEL for Edge with docker and podman is conflicting with docker.

ondrejbudai:
@gicmo can we close this?

PR #1621, opened 2021-08-05
RHEL repofile changes

ondrejbudai:
The first commit adds rhel-85 repository file. Since composer has gained the ability to cross-distro build images, we actually need a repo file for each RHEL minor version and this was forgotten when rhel-85 definition was added.

The second commit changes the rhel-8 repository file to point at 8.3 content. rhel-8 is in fact rhel-83. Building the rhel-8 distribution right now would mean that 8.3 definition would be used against 8.4 content which doesn't make much sense. Therefore, this commit fixes the minor version to 8.3.

This pull request includes:


teg (approved):
Thanks!

PR #1624, opened 2021-08-05
cloudapi: rework return values

### LINKED ISSUE #1606: cloudapi: Replace `panic`s with 500s
There's still several places where the handlers will panic instead of returning an InternalServerError. Calling a handler shouldn't crash composer.

chloenayon:
Change instances of panic on error to return StatusInternalServerError
instead. Additionally change return code 200s to http.StatusOK.

Fixes #1606

This pull request includes:

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -78,7 +78,7 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
 
 	err := json.NewDecoder(ctx.Request().Body).Decode(i)
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Cannot parse request body: %v", err))
+		return echo.NewHTTPError(http.StatusBadRequest, "Cannot parse request body: %s", err.Error())
 	}
 	return nil
 }
@@ -170,7 +170,7 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
 	// use the same seed for all images so we get the same IDs
 	bigSeed, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
 	if err != nil {
-		panic("cannot generate a manifest seed: " + err.Error())
+		return echo.NewHTTPError(http.StatusInternalServerError, "Cannot generate a manifest seed: %s", err.Error())
 	}
 	manifestSeed := bigSeed.Int64()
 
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
Look good, thanks!

thozza (dismissed):
Thanks for the fix! Changes look good to me.

Nitpick: It would be probably better to be consistent in how we format the message in `echo.NewHTTPError()` calls with the rest of the code. From a brief look, there seems to be one instance when we use `fmt.Snprintf()`:
https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L81

while in all other relevant cases we simply pass the arguments for the message format string as additional values:
https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L146
https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L180
https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L184
https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L216
...

My suggestion would be to stick to the dominant style - pass arguments of the message as additional values when calling `echo.NewHTTPError()`, instead of string concatenation. It would be also great if you could change https://github.com/osbuild/osbuild-composer/blob/a3ac31a4830604cf0d25f91236e092512cddab13/internal/cloudapi/server.go#L81 to the same style to make the whole file consistent.

internal/cloudapi/server.go
```diff
@@ -78,7 +78,7 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
 
 	err := json.NewDecoder(ctx.Request().Body).Decode(i)
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Cannot parse request body: %v", err))
+		return echo.NewHTTPError(http.StatusBadRequest, "Cannot parse request body: %s", err)
```
thozza, line 81: You should use `err.Error()` instead of `err` here, to make sure that the passed value type is `string`.

thozza (dismissed):
LGTM

thozza (approved):
Still good!

PR #1625, opened 2021-08-05
Oauth2

Gundersanne:
blocked on:
- [x] ocm-sdk landed in fedora-33/34/35/36


ondrejbudai:
I think you unknowingly opened a pandora's box. :D

https://github.com/dgrijalva/jwt-go is now archived. A community fork is here: https://github.com/golang-jwt/jwt
I guess we should switch to the latter one everywhere (echo is doing this in its latest version) but this has a small issue: It isn't shipped in Fedora yet. I created a [review request](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1989386) on Monday but there's no response so far.

What I'm basically saying is that this might be blocked for some time unless the jwt situation is resolved in Fedora.

Gundersanne:
> I think you unknowingly opened a pandora's box. :D
> 
> https://github.com/dgrijalva/jwt-go is now archived. A community fork is here: https://github.com/golang-jwt/jwt
> I guess we should switch to the latter one everywhere (echo is doing this in its latest version) but this has a small issue: It isn't shipped in Fedora yet. I created a [review request](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1989386) on Monday but there's no response so far.
> 
> What I'm basically saying is that this might be blocked for some time unless the jwt situation is resolved in Fedora.


Yea I know the dgrijalva is deprecated, I think some parts ocm-sdk are still pulling it in though :/ Still figuring out what's the best way forward here. Wonder if it's worht using  https://github.com/auth0/go-jwt-middleware instead of the api.openshift.com's platform default (since the ocm-sdk also isn't shipped in fedora).

teg (commented):
This seems really good!

However, could you expand on the commit messages a bit (or a lot)? I was going to ask you to talk me through it, but for the sake of everyone else following along having it written would be great.

The ideal would be a high level explanation of how the protocol works and would be used, as well as your implementation choices. Ideally enough so that someone who knows the project without knowing about jwt etc can understand and asses if the patches make sense without reading the code. And when they read the code they just have to verify that it does no more nor less than the commit message says.

cmd/osbuild-composer/main.go
```diff
@@ -21,7 +21,7 @@ var repositoryConfigs = []string{
 
 func main() {
 	var verbose bool
-	flag.BoolVar(&verbose, "v", false, "Print access log")
```
ondrejbudai, line 24: can you comment this change in the commit message?

cmd/osbuild-composer/composer.go
```diff
@@ -168,11 +181,26 @@ func (c *Composer) Start() error {
...
 		go func() {
+			handler := c.workers.Handler()
+			var err error
+			if c.config.Worker.EnableJWT {
+				handler, err = auth.BuildJWTAuthHandler(
+					c.config.Worker.JWTKeysURL,
+					c.config.Worker.JWTKeysCA,
+					c.config.Worker.JWTACLFile,
+					[]string{},
+					c.workers.Handler(),
```
ondrejbudai, line 192: Maybe let's use `handler` here in order to call the method only once?

cmd/osbuild-composer/config.go
```diff
@@ -100,6 +110,8 @@ func loadConfigFromEnv(intf interface{}) error {
 				continue
 			}
 			fieldV.SetString(confV)
+		case reflect.Bool:
+			fallthrough
```
ondrejbudai, line 114: I suggest using `continue` for consistency with the later cases.

cmd/osbuild-worker/main.go
```diff
@@ -148,32 +149,47 @@ func main() {
...
 
-		if config.Authentication.OAuthURL == "" {
-			log.Fatal("OAuth URL should be specified together with the offline token")
+			if config.Authentication.OAuthURL == "" {
+				log.Fatal("OAuth URL should be specified together with the offline token")
+			}
+			oAuthURL = &config.Authentication.OAuthURL
+
+			if strings.HasPrefix(address, "http") {
+				out, err := exec.Command("systemd-escape", "-u", address).Output()
```
ondrejbudai, line 169: Isn't there a way to escape this without using `systemd`?

(If not, we might need to tweak the spec file to require this executable.)

ondrejbudai (commented):
Quick review. :)

cmd/osbuild-composer/main.go
Gundersanne: Oh yea, that's the glog flag fiasco >.> will do!

cmd/osbuild-worker/main.go
Gundersanne: Ah good point, I'll look into it

Gundersanne:
I updated the commit message a bit more, will include some more stuff re: the glog flag stuff.

Also wrote https://github.com/Gundersanne/osbuild-composer/blob/oauth2/docs/news/unreleased/api-oauth2-support.md explaining the new config options a bit more.

cmd/osbuild-worker/main.go
Gundersanne: Hmm, systemd-escape is shipped as part of systemd; so we should already be covered since we already require systemd.
image

Gundersanne:
Damn I accidentally deleted my branch :woman_facepalming:  Reopened here https://github.com/osbuild/osbuild-composer/pull/1717 with some more information in the news document re: workflow.

PR #1627, opened 2021-08-06
osbuild: Simplify mounts and devices

achilleas-k:
Simplifying devices and mounts for stages in the osbuild2 package.  This represents the osbuild schema much more closely.  Device and Mount have their own types that aren't tied directly to a Stage type, so stage-specific mounts and devices weren't a good representation.

I also took the opportunity to abstract the creation of some stages that include mounts and devices from their final representation.  This makes the Stage creation a bit more restrictive, to match the schema, while also allowing the use of the Mount and Device interfaces.  For example, `mkfs` stages require only one device labelled by the key `device`, so it makes more sense for the creation function to accept only one device and have that function add the key itself, instead of relying on the caller to assign the key or relying on struct types for all use cases (which is what we were doing before).
I think the usefulness of this approach is especially apparent in the zipl.inst stage, where the stage accepts a number of devices and mounts, with arbitrary keys, but also requires one device with the key `disk`.  The new creation function for this stage takes a general Devices map, for the arbitrarily named devices, and a separate disk Device that is added to the stage by the creation function with the appropriate key.

I intended to make the same changes for the inputs, which have the same issues, but that requires bigger changes that should be done in a separate PR.

This pull request includes:

### DIFF
--- a/internal/distro/rhel90/pipelines.go
+++ b/internal/distro/rhel90/pipelines.go
@@ -873,10 +873,10 @@ func liveImagePipeline(inputPipelineName string, outputFilename string, pt *disk
 	p.Name = "image"
 	p.Build = "name:build"
 
-	loopback := osbuild.NewLoopbackDevice(&osbuild.LoopbackDeviceOptions{Filename: outputFilename})
 	p.AddStage(osbuild.NewTruncateStage(&osbuild.TruncateStageOptions{Filename: outputFilename, Size: fmt.Sprintf("%d", pt.Size)}))
-	sfOptions, sfDevices := sfdiskStageOptions(pt, loopback)
-	p.AddStage(osbuild.NewSfdiskStage(sfOptions, sfDevices))
+	sfOptions := sfdiskStageOptions(pt)
+	loopback := osbuild.NewLoopbackDevice(&osbuild.LoopbackDeviceOptions{Filename: outputFilename})
+	p.AddStage(osbuild.NewSfdiskStage(sfOptions, loopback))
 
 	for _, stage := range mkfsStages(pt, loopback) {
 		p.AddStage(stage)
@@ -934,28 +934,24 @@ func mkfsStages(pt *disk.PartitionTable, device *osbuild.Device) []*osbuild2.Sta
 				UUID:  p.Filesystem.UUID,
 				Label: p.Filesystem.Label,
 			}
-			devices := &osbuild.MkfsXfsStageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsXfsStage(options, devices)
+			stage = osbuild.NewMkfsXfsStage(options, stageDevice)
 		case "vfat":
 			options := &osbuild.MkfsFATStageOptions{
 				VolID: strings.Replace(p.Filesystem.UUID, "-", "", -1),
 			}
-			devices := &osbuild.MkfsFATStageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsFATStage(options, devices)
+			stage = osbuild.NewMkfsFATStage(options, stageDevice)
 		case "btrfs":
 			options := &osbuild.MkfsBtrfsStageOptions{
 				UUID:  p.Filesystem.UUID,
 				Label: p.Filesystem.Label,
 			}
-			devices := &osbuild.MkfsBtrfsStageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsBtrfsStage(options, devices)
+			stage = osbuild.NewMkfsBtrfsStage(options, stageDevice)
 		case "ext4":
 			options := &osbuild.MkfsExt4StageOptions{
 				UUID:  p.Filesystem.UUID,
 				Label: p.Filesystem.Label,
 			}
-			devices := &osbuild.MkfsExt4StageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsExt4Stage(options, devices)
+			stage = osbuild.NewMkfsExt4Stage(options, stageDevice)
 		default:
 			panic("unknown fs type " + p.Type)
 		}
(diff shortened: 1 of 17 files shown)


ondrejbudai (dismissed):
Very nice!

internal/osbuild2/device.go
```diff
@@ -1,8 +1,6 @@
 package osbuild2
 
-type Devices interface {
-	isStageDevices()
-}
```
thozza, line 5: Hmm. I'm probably a bit confused, but if Devices are very similar to Inputs, I would expect `Devices` to be kept as an interface type. Then specific device types would implement it, e.g. the `LoopbackDevice`. And lastly stages accepting the `LoopbackDevice` would have their own stage-specific interface type for devices, which would be then implemented by all Device types which are acceptable for the stage.

As I understand it with Inputs, the Inputs object properties are specific to the Input type. Devices seem to be a bit different and it seems that the `Devices` object can have only `device` property (https://github.com/osbuild/osbuild/blob/569697e9a9fff55c62c694a28120a26c6a7d3c7e/schemas/osbuild2.json#L19-L37). Thus defining `Devices` type as map is IMO too generic.

internal/osbuild2/mkfs_stages_test.go
```diff
@@ -21,25 +21,23 @@ func TestNewMkfsStage(t *testing.T) {
...
 		Label: "test",
 	}
-	btrfsDevices := &MkfsBtrfsStageDevices{Device: *device}
-	mkbtrfs := NewMkfsBtrfsStage(btrfsOptions, btrfsDevices)
+	mkbtrfs := NewMkfsBtrfsStage(btrfsOptions, device)
 	mkbtrfsExpected := &Stage{
 		Type:    "org.osbuild.mkfs.btrfs",
 		Options: btrfsOptions,
-		Devices: btrfsDevices,
+		Devices: Devices{"device": *device},
```
thozza, line 28: I think that it would be good if we had a constructor-like function for `Devices` (e.g. similar to https://github.com/osbuild/osbuild-composer/blob/4a8e5e8b19888face6adb3c7041d55de6d9da7e5/internal/osbuild2/xz_stage_test.go#L28). Personally I'm not a fan of literals with "magic" values such as `device`, because one has to has to know what key to use. Defining the key `device` as a constant would be also an improvement IMO. Especially when this literal with the key is used on 4 places here and in every "filesystem" stage constructor-function.

internal/osbuild2/mount.go
```diff
@@ -1,8 +1,6 @@
 package osbuild2
 
-type Mounts interface {
-	isStageMounts()
-}
```
thozza, line 5: I have conceptually the same comment here as with `Devices`...

internal/osbuild2/zipl_inst_stage.go
```diff
@@ -13,20 +13,17 @@ type ZiplInstStageOptions struct {
...
-func (ZiplInstStageMounts) isStageMounts() {}
-
-// Return a new zipl.inst stage. A device needs to be specified as 'disk' and root mountpoint must be provided
-func NewZiplInstStage(options *ZiplInstStageOptions, devices *CopyStageDevices, mounts *CopyStageMounts) *Stage {
+// Return a new zipl.inst stage. The 'disk' parameter must represent the
+// (entire) device that contains the /boot partition.
+func NewZiplInstStage(options *ZiplInstStageOptions, disk *Device, devices *Devices, mounts *Mounts) *Stage {
+	// create a new devices map and add the disk to it
+	devmap := map[string]Device(*devices)
+	devmap["disk"] = *disk
```
thozza, line 21: Shouldn't this be defined as a separate Device(s) type with property `disk`, instead of `device`?

thozza (commented):
Very nice work, everything is nicely simplified.

However I must say that I have a few conceptual questions, which I put inline in the code. Maybe I'm just confused and misunderstood the concept of Devices and Mounts in osbuild, but the refactoring of how Devices and Mounts are represented and passed to stages seems to me a bit weird. I'm trying to understand their concept by mapping them to Inputs in my head, which may be wrong.... 😕

internal/distro/rhel90/pipelines.go
```diff
@@ -934,28 +934,24 @@ func mkfsStages(pt *disk.PartitionTable, device *osbuild.Device) []*osbuild2.Sta
...
-			stage = osbuild.NewMkfsFATStage(options, devices)
+			stage = osbuild.NewMkfsFATStage(options, stageDevice)
 		case "btrfs":
 			options := &osbuild.MkfsBtrfsStageOptions{
 				UUID:  p.Filesystem.UUID,
 				Label: p.Filesystem.Label,
 			}
-			devices := &osbuild.MkfsBtrfsStageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsBtrfsStage(options, devices)
+			stage = osbuild.NewMkfsBtrfsStage(options, device)
```
gicmo, line 948:
```suggestion
			stage = osbuild.NewMkfsBtrfsStage(options, stageDevice)
```

internal/distro/rhel90/pipelines.go
```diff
@@ -934,28 +934,24 @@ func mkfsStages(pt *disk.PartitionTable, device *osbuild.Device) []*osbuild2.Sta
...
 		case "ext4":
 			options := &osbuild.MkfsExt4StageOptions{
 				UUID:  p.Filesystem.UUID,
 				Label: p.Filesystem.Label,
 			}
-			devices := &osbuild.MkfsExt4StageDevices{Device: *stageDevice}
-			stage = osbuild.NewMkfsExt4Stage(options, devices)
+			stage = osbuild.NewMkfsExt4Stage(options, device)
```
gicmo, line 954:
```suggestion
			stage = osbuild.NewMkfsExt4Stage(options, stageDevice)
```

gicmo (changes requested):

internal/osbuild2/device.go
gicmo: `device` in the schema is just the name for the definition. Devices in osbuild are indeed a `name: object` mapping with an arbitrary number of devices. Some stages (`mkfs*`) require there to be at least one known one (usually called `device`) which is the one they operate on.
I am a bit of an outsider to the go code but on a first glance the map looks fine to me.

gicmo: One other note: I think devices and mounts are simpler than inputs. The latter has `references`, which exact contents is  flexible, i.e. either `array` or `object`, and the contents of said `object` depends on the input type and origin: it is, e.g. different for `org.osbuild.files` input with `pipeline` vs. `org.osbuild.files` with `source` vs. `org.osbuild.tree`.

internal/osbuild2/zipl_inst_stage.go
gicmo: That looks fine to me. In the manifest, we need:

```
"devices": {
   "disk": {
     <a device, probably loopback>
   }
}
```
Maybe I am missing something though

thozza: Yeah, I should have resolved this comment. It was made before our session with Achilleas to sort out how to actually represent devices and various input types in composer. I didn't realise that the key under which the input device is expected is stage-specific...

internal/osbuild2/device.go
thozza: Yes, in the end we agreed that the stage devices and inputs should be represented as map. My comment was caused mostly by a confusion 😃

gicmo (approved):
I have fixed the tiny issues I found and I think this is good to go.

internal/osbuild2/mkfs_stages_test.go
teg: Using more constants makes sense, but let's leave that for future work.

teg (approved):
Nice cleanups @achilleas-k , and thanks for cleaning up the cleanups @gicmo ! I dismissed the last comment from @thozza, not because I disagree, but I think it was only a style issue that we could address in follow-ups (and this is a blocker for the release).

gicmo:
Rebased this onto `main`.

gicmo (dismissed):
Still good!

teg:
Yikes, I caused a conflict. Needs rebase.

gicmo:
> Yikes, I caused a conflict. Needs rebase.

done.

teg (approved):

PR #1641, opened 2021-08-12
cloudapi: change Organization customization to a string

### LINKED ISSUE #1582: Cloudapi: Use string for Organization in subscription customization
Currently the cloudapi isn't deployed anywhere where stringent requirements are necessary. Let's make sure we fix things like this before it is.

chloenayon:
Make Organization in the Subscription customization a string
instead of an int.

fixes #1582 

This pull request includes:

### DIFF
--- a/internal/cloudapi/server.go
+++ b/internal/cloudapi/server.go
@@ -221,7 +221,7 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
 		imageOptions := distro.ImageOptions{Size: imageType.Size(0)}
 		if request.Customizations != nil && request.Customizations.Subscription != nil {
 			imageOptions.Subscription = &distro.SubscriptionImageOptions{
-				Organization:  request.Customizations.Subscription.Organization,
+				Organization:  fmt.Sprintf("%d", request.Customizations.Subscription.Organization),
 				ActivationKey: request.Customizations.Subscription.ActivationKey,
 				ServerUrl:     request.Customizations.Subscription.ServerUrl,
 				BaseUrl:       request.Customizations.Subscription.BaseUrl,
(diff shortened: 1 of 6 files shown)


Gundersanne (changes requested):
Ah nice, good catch in the distro package, I didn't realize this issue went all the way into there.

Still needs a change in `test/cases/api.sh`; when building the request the organization value needs some quotes around it.

gicmo:
This needs a rebase on main to fix CI

Gundersanne (approved):
Very nice, thanks!

edit: just re-triggered CI.

internal/cloudapi/openapi.yml
```diff
@@ -466,8 +466,8 @@ components:
         - insights
       properties:
         organization:
-          type: integer
-          example: 2040324
+          type: string
+          example: '2040324'
```
Gundersanne, line 470: let's revert this to an intereger

test/cases/api.sh
```diff
@@ -416,7 +416,7 @@ if [[ "$ID" == "rhel" ]]; then
   SUBSCRIPTION_BLOCK=$(cat <<EndOfMessage
 ,
     "subscription": {
-      "organization": ${API_TEST_SUBSCRIPTION_ORG_ID:-},
+      "organization": "${API_TEST_SUBSCRIPTION_ORG_ID:-}",
```
Gundersanne, line 419: This can also be reverted.

Gundersanne (changes requested):
Sorry for doing this, but I think it's better to not break the openapi.yml, but keep the internal change to string of SubscriptionImageOptions.

After reconsidering this and introducing v2 soon in #1704, let's not give ourselves the headache of having to updating image-builder at the same time.

And then in server.go:
```
- Organization:  request.Customizations.Subscription.Organization,
+ Organization:  fmt.Sprintf("%d", request.Customizations.Subscription.Organization),
```

Sorry for the double work, my bad as I said something different in the issue :(

Gundersanne (approved):
lgtm!

PR #1647, opened 2021-08-17
debug: see how rhel-9 test are working

jrusz:
This pull request includes:

Looking into issues with running tests on RHEL-9.


jrusz:
I don't think it's the nicest way to make this work but I tried to duplicate as code as little as I could. I'm open to suggestions to make it look better but right now the functions that call `composer-cli --json ...` have another struct defined with the response as that has changed with `weldr-client` and then I added a small function to check whether `weldr-client` is installed and decide which result structure to use in the test based on that. Other than that it includes necessary updates to make testing on RHEL-9 runners work.

jrusz:
@bcl since you're the author of `weldr-client` could you take a look at this as well please? Maybe you'll have some suggestions for improvements :)

cmd/osbuild-composer-cli-tests/main_test.go
```diff
@@ -198,12 +204,24 @@ func startCompose(t *testing.T, name, outputType string) uuid.UUID {
...
+		err := json.Unmarshal(rawReply, &replyWeldr)
+		require.Nilf(t, err, "Unexpected reply: %v", err)
+		require.Truef(t, replyWeldr.Body.Status, "Unexpected status %v", replyWeldr.Body.Status)
+		return replyWeldr.Body.BuildID
+	} else {
+		err := json.Unmarshal(rawReply, &reply)
+		require.Nilf(t, err, "Unexpected reply: %v", err)
+		require.Truef(t, reply.Status, "Unexpected status %v", reply.Status)
+		return reply.BuildID
+	}
```
thozza, line 224: As mentioned above, I think this should be possible to simplify e.g. such as:
```golang
	// composer-cli response
	type reply struct {
		BuildID uuid.UUID `json:"build_id"`
		Status  bool      `json:"status"`
	}
	// weldr-client response
	var replyWeldr struct {
		Body reply `json:"body"`
	}
	rawReply := runComposerJSON(t, "compose", "start", name, outputType)
	var err error
	if checkWeldrClient(t) {
		err = json.Unmarshal(rawReply, &replyWeldr)
	} else {
		err = json.Unmarshal(rawReply, &replyWeldr.Body)
		
	}
	require.Nilf(t, err, "Unexpected reply: %v", err)
	require.Truef(t, replyWeldr.Body.Status, "Unexpected status %v", replyWeldr.Body.Status)
	return replyWeldr.Body.BuildID
```

tools/provision.sh
```diff
@@ -2,13 +2,25 @@
...
+        https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+    sudo rpm -Uvh /tmp/epel.rpm
     sudo dnf install -y koji ansible
+elif [[ $ID == rhel || $ID == centos ]] && [[ ${VERSION_ID%.*} == 9 ]]; then
+    # we have our own small epel for EL9, let's install it
+
+    # install Red Hat certificate, otherwise dnf copr fails
+    curl -LO --insecure https://hdn.corp.redhat.com/rhel8-csb/RPMS/noarch/redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm
+    sudo dnf install -y ./redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm dnf-plugins-core
+    sudo dnf copr enable -y copr.devel.redhat.com/osbuild-team/epel-el9 "rhel-9.dev-$ARCH"
```
thozza, line 21: Could you please explain why it is needed to pass `"rhel-9.dev-$ARCH"` to the `dnf copr` command? I would expect everything to work just fine even without it.

thozza (changes requested):
Thanks for our work. In principle, all changes look reasonable. I think some parts could be simplified / deduplicated and I found one unused function argument. Added comments in the code, but not to all of the places that could be simplified  as I hope it is obvious from my comments.

tools/provision.sh
jrusz: I don't have an explanation I just took it from this commit from Ondrej https://github.com/osbuild/osbuild-composer/commit/e1462e261db3583628c73d92705965101e000842

jrusz:
Great suggestions @thozza thanks a lot! I knew this could look better :D ... I'll just verify it works as expected and then run the full test suite a request another review :)

tools/provision.sh
thozza: hmm, not sure why Ondrej used it. I never saw it being used, that't the reason why I'm asking... I think it should not be needed, but can't hurt to keep it there if it works. My motivation was only that we would not need to define `ARCH` variable in that case 😄

jrusz: I can try removing it and see what happens :D ... so it would be just `"rhel-9.dev"` ?

thozza: @jrusz no, please try just `sudo dnf copr enable -y copr.devel.redhat.com/osbuild-team/epel-el9`

jrusz: Alright thanks. The base tests seemed to have passed, I just added this change and if everything passes I think we could be ready to merge this :)

thozza: I just tested the `dnf copr` command as I suggested and for some reason it does not work with that copr repo if the "chroot" is not specified. I have no idea why, as I never saw it being used, but I think that we need to revert back to the original stat. Sorry for the noise 😉

jrusz: Np, I reverted it. Let's see if the CI passes now with all the tests enabled :)

thozza (approved):
Everything looks good! 🥳

teg (approved):

PR #1650, opened 2021-08-18
Add proxy support to repo sources

bcl:
This adds 3 new fields to the repo sources toml that allows a proxy server to be used with the specific source. Currently it supports unauthenticated proxies, and basic auth username/password protected proxies.

This pull request includes:


msehnout (commented):
LGTM 👍 but I'd label this like a draft so that it isn't merged before the corresponding osbuild PR because without it, it won't work, right? Also the spec file will need a bump when osbuild gets released.

bcl:
Switched to draft. @msehnout how are releases that depend on osbuild features coordinated? And should the spec be updated in this PR or as part of the corresponding osbuild-composer release? I'd hoped support would get into osbuild 31 but it looks like it wasn't merged.

msehnout:
> Switched to draft. @msehnout how are releases that depend on osbuild features coordinated? And should the spec be updated in this PR or as part of the corresponding osbuild-composer release? I'd hoped support would get into osbuild 31 but it looks like it wasn't merged.

We don't have any specific process for the coordination, but I usually bump the spec file requirement right in the osbuild-composer PR (this one) because the feature wouldn't work without osbuild support. This one has the advantage that osbuild can include the proxy support and it works just fine with older osbuild-composer, so the only requirement is that osbuild hits downstream earlier or at the same time. It is unfortunate that osbuild has been released few hours ago, so the next planned release is in two weeks. I'd personally target the osbuild PR to `v32` and this PR to osbuild-composer `v33` (in ~3 weeks).

Alternative would be to release both at the same time and in Bodhi we would submit a signle update for both components so that composer doesn't get in sooner than osbuild.

PR #1651, opened 2021-08-20
ci: minor fixes for nightly testing

jrusz:
Updating rules for Installer and Regression testing and fixing failures
if certain variables are not set.

This pull request includes:

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -103,7 +103,7 @@ if [ -f "rhel8internal.repo" ]; then
     sudo mv rhel8internal.repo /etc/yum.repos.d/
     # Use osbuild from schutzfile if desired for testing custom osbuild-composer packages
     # specified by $REPO_URL in ENV and used in prepare-rhel-internal.sh
-    if [ "$SCHUTZ_OSBUILD" == 1 ]; then
+    if [ "${SCHUTZ_OSBUILD:=false}" == true ]; then
         sudo rm -f /etc/yum.repos.d/osbuild-composer.repo
     else
         sudo rm -f /etc/yum.repos.d/osbuild*.repo
--- a/schutzbot/prepare-rhel-internal.sh
+++ b/schutzbot/prepare-rhel-internal.sh
@@ -64,7 +64,7 @@ cp rhel-8.json rhel-8-beta.json
 
 JOB_NAME="${JOB_NAME:-${CI_JOB_ID}}"
 # Do not create tests repo if it's provided from ENV
-if [ -z "$REPO_URL" ]; then
+if [ -z "${REPO_URL+x}" ]; then
     greenprint "📦 Installing requirements"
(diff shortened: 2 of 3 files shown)


jrusz:
@atodorov could you please review and merge if you have no objections? These are fixes for my past oversights related to nightly testing. Currently have have both the nightly pipeline and scheduled cloud-cleaner pipeline failing because of this...

schutzbot/prepare-rhel-internal.sh
```diff
@@ -64,7 +64,7 @@ cp rhel-8.json rhel-8-beta.json
 
 JOB_NAME="${JOB_NAME:-${CI_JOB_ID}}"
 # Do not create tests repo if it's provided from ENV
-if [ -z "$REPO_URL" ]; then
+if [ -z "${REPO_URL+x}" ]; then
```
atodorov, line 67: I don't understand the `+x` part.

schutzbot/deploy.sh
```diff
@@ -103,7 +103,7 @@ if [ -f "rhel8internal.repo" ]; then
     sudo mv rhel8internal.repo /etc/yum.repos.d/
     # Use osbuild from schutzfile if desired for testing custom osbuild-composer packages
     # specified by $REPO_URL in ENV and used in prepare-rhel-internal.sh
-    if [ "$SCHUTZ_OSBUILD" == 1 ]; then
+    if [ "${SCHUTZ_OSBUILD:=false}" == true ]; then
```
atodorov, line 106: What provides this variable ? Surely whatever relies on it will not expect to change the behavior behind the accepted values. 

What's the reasoning for not comparing to `1` ?

atodorov (commented):
Added questions for items which I don't understand why they are needed.

schutzbot/prepare-rhel-internal.sh
jrusz: I got this from here https://stackoverflow.com/a/13864829  I've initially set this up so that we can use custom URL to a scratch build for the nightly testing pipeline. It worked fine until I've removed the variable because we no longer need it and want to use the rpm that's in the nightly compose and since we use `set -u` it fails on this as can be seen here: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1518775766#L523

schutzbot/deploy.sh
jrusz: This variable is also provided only in the scheduled pipeline settings. I've set it up so that we can use a pinned osbuild version in schutzfile if needed. This was the case when we were running the nightly testings using the osbuild-composer scratch build. osbuild was not in a new enough version in the repository so a pinned version was used instead. It should not be needed now and it again fails because of `set -u` ... There is no reason for changing it from `1` to  `true` but I had to set it up to some default value and using `true` `false` just seemed better....

atodorov (approved):

PR #1541, opened 2021-07-09
[Experimental] CI clean-ups

ondrejbudai:
Untangling CI in preparation for RHEL 9.

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] create a file in [news/unreleased](https://github.com/osbuild/osbuild-composer/tree/main/docs/news/unreleased) directory if this change should be mentioned in the release news
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #1657, opened 2021-08-23
[CI] remove rhn subscriptions before registering a new one

jabia99:
Sometimes CI  gets to an error related to RHSM: https://issues.redhat.com/browse/COMPOSER-1037

This PR intends to solve this issue.

This pull request includes:

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -52,6 +52,7 @@ ARCH=$(uname -m)
 
 if [[ $ID == "rhel" && $VERSION_ID == "8.3" && -n "${RHN_REGISTRATION_SCRIPT:-}" ]] && ! sudo subscription-manager status; then
     greenprint "Registering RHEL"
+    sudo subscription-manager remove --all
     sudo chmod +x "$RHN_REGISTRATION_SCRIPT"
     sudo "$RHN_REGISTRATION_SCRIPT"
 fi
--- a/schutzbot/mockbuild.sh
+++ b/schutzbot/mockbuild.sh
@@ -13,6 +13,7 @@ ARCH=$(uname -m)
 # Register RHEL if we are provided with a registration script.
 if [[ $ID == "rhel" && $VERSION_ID == "8.3" && -n "${RHN_REGISTRATION_SCRIPT:-}" ]] && ! sudo subscription-manager status; then
     greenprint "🪙 Registering RHEL instance"
+    sudo subscription-manager remove --all
     sudo chmod +x "$RHN_REGISTRATION_SCRIPT"
     sudo "$RHN_REGISTRATION_SCRIPT"
 fi


jrusz (commented):
Do you think this will be enough to solve the issue? I thought we need a retry logic... What do you think @atodorov I'm not sure about this... Also I feel like this code will be removed in https://github.com/osbuild/osbuild-composer/pull/1541 Ondrej wants to register the runners during provisioning basically so we don't have to worry about it in the code in repository... So maybe a change in the registration script itself that we have in CI variables would be better?

jabia99:
> Do you think this will be enough to solve the issue? I thought we need a retry logic... 

From what I have seen in the bug reports, the issue is that we are registering the machine twice ([1815172](https://bugzilla.redhat.com/show_bug.cgi?id=1815172#c10)). So I thought maybe this could be a solution. But I could add retry logic instead

> So maybe a change in the registration script itself that we have in CI variables would be better?

Probably, but I was afraid of changing the CI variables. I thought about merging this and if we don't get this error in a few weeks, add it to the CI variable

atodorov:
> Do you think this will be enough to solve the issue?

IDK if this will solve the issue but it is a quick & easy patch and will not hurt IMO. See comments though. I'd say give it a go first and let's see what happens.

schutzbot/deploy.sh
```diff
@@ -52,6 +52,7 @@ ARCH=$(uname -m)
 
 if [[ $ID == "rhel" && $VERSION_ID == "8.3" && -n "${RHN_REGISTRATION_SCRIPT:-}" ]] && ! sudo subscription-manager status; then
     greenprint "Registering RHEL"
+    sudo subscription-manager remove --all
```
atodorov, line 55: Is this going to remove all subscriptions on the account? Including ones from other systems? 

If yes this is prone to race conditions with other test jobs running in parallel which subscribe and then we remove their subscriptions before these systems were able to execute their dnf commands. That would lead to even more peculiar and hard to debug cases if it is really the case. 

Is there a way to remove all previous subscriptions only from the same system ?

atodorov (commented):
Added comments b/c I don't fully understand how subscription-manager works in our environment.

schutzbot/deploy.sh
jabia99: I've read a bit about how subscription manager works and I think it only removes the subscription from this system ([Removing a subscription](https://access.redhat.com/documentation/en-us/red_hat_subscription_management/1/html-single/rhsm/index#reg-cli-remove))

But I could be getting it wrong. Maybe just doing retry logic is faster (but not as clean imo)

jrusz:
> IDK if this will solve the issue but it is a quick & easy patch and will not hurt IMO. See comments though. I'd say give it a go first and let's see what happens.

True, so let's merge this and observe. According to the documentation it should only remove subscriptions from the system where it's ran on.

atodorov (approved):

PR #1659, opened 2021-08-23
distro/rhel85: pre-load uid/gid database for edge

gicmo:
When building RHEL for Edge commits and a parent together with an URL was specified, add a `org.osbuild.ostree.passwd` stage which then will pre-load the uid/gid database with the data from the parent commit. This ensures that uids and gids do not change for the "child" commit.

NB: needs osbuild fix for the `ostree.checkout` input: https://github.com/osbuild/osbuild/pull/775

This pull request includes:

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -24,12 +24,19 @@ type repository struct {
 	CheckGPG   bool   `json:"check_gpg,omitempty"`
 }
 
+type ostreeOptions struct {
+	Ref    string `json:"ref"`
+	Parent string `json:"parent"`
+	URL    string `json:"url"`
+}
+
 type composeRequest struct {
 	Distro       string              `json:"distro"`
 	Arch         string              `json:"arch"`
 	ImageType    string              `json:"image-type"`
 	Blueprint    blueprint.Blueprint `json:"blueprint"`
 	Repositories []repository        `json:"repositories"`
+	OSTree       ostreeOptions       `json:"ostree"`
 }
 
 func main() {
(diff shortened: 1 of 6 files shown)


gicmo:
I believe @henrywang wanted to add a test case for this, if I am not mistaken.

teg (commented):
I think we can easily test this. @henrywang had to do a work-around for the `-rt` tests I believe to avoid running into this bug, could just remove that again :) Otherwise: great!

henrywang:
@gicmo I'll push test update in this PR.
@teg Yeah, Just change the first blueprint with `normal` kernel configured and then upgrade to `rt`kernel. Easy to update.
To make my updated teat work, I have to pin `osbuild` to the latest commit which has ` stages: add org.osbuild.ostree.passwd` commit included. @gicmo @teg May I do that? Any conflict or concern about my pin? Thanks.

henrywang:
@gicmo I found the following error when build upgrade image:
```
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
Failed to create file /sys/fs/selinux/checkreqprot: Read-only file system
File "/run/osbuild/bin/org.osbuild.users", line 170, in <module>
   r = main(args["tree"], args["options"])
  File "/run/osbuild/bin/org.osbuild.users", line 163, in main
    add_ssh_key(tree, name, key)
  File "/run/osbuild/bin/org.osbuild.users", line 130, in add_ssh_key
    os.mkdir(ssh_dir, 0o700)
FileNotFoundError: [Errno 2] No such file or directory: '/run/osbuild/tree//home/admin//.ssh'
```

msehnout (commented):
It looks fine to me. I'm not sure what is causing the issue @henrywang found.

gicmo:
> It looks fine to me. I'm not sure what is causing the issue @henrywang found.

The reason is the following:

- A user, say `admin` is created via a blueprint in the _original_ commit
- User gets added to the user database
- Blueprint for the _update_ commit also includes said user
- `org.osbuild.ostree.passwd` will will pre-fill the user database with the info from the _original_ commit
- `org.osbuild.user` stage checks if the user exists via `getpwnam`, which it does and does not create it
   -  thus there is _no home directory_ is created and also does not exist for user `admin`
- `add_ssh_key` in `org.osbuild.user` will fail because the home dir is missing

In short: you can not add ssh keys to users already present in the parent commit. Probably should not touch them at all.

PR #1662, opened 2021-08-24
Some improvements for nightly pipeline

atodorov:
This pull request includes:


teg (approved):

teg (approved):

jrusz (commented):
LGTM, tests pass on both x86_64 and aarch64 architectures on RHEL-8.5 ... I just have one question, was https://github.com/osbuild/osbuild-composer/pull/1662/commits/4975bd37cffb5fa2de558935d7c188d21d37c3e8 supposed to fix the failing regression test? It's still failing but I think we can fix in in another PR... Link to latest scheduled pipeline run: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/363163253

atodorov:
> LGTM, tests pass on both x86_64 and aarch64 architectures on RHEL-8.5 ... I just have one question, was [4975bd3](https://github.com/osbuild/osbuild-composer/commit/4975bd37cffb5fa2de558935d7c188d21d37c3e8) supposed to fix the failing regression test? 

Yes, that's what I am trying to fix with this commit. 

> It's still failing but I think we can fix it in another PR... 

I'd like to get the fixes merges & tested ASAP b/c we need another rebase downstream. CI jobs are taking their time eitherway so I'm not sure which will be the fastest way to get this fixed.

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -65,17 +65,19 @@ case "${ID}" in
...
+                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}-beta.json"
                 else
                     DISTRO_NAME="rhel-8"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-8.json"
+                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
                 fi
-                REPO1_NAME="baseos"
-                REPO2_NAME="appstream"
+                REPO1_NAME="baseos-${ARCH}"
+                REPO2_NAME="appstream-${ARCH}"
```
jrusz, line 80: Won't this break regular pipelines? I think you renamed the repos only for the nightly pipeline...

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -65,17 +65,19 @@ case "${ID}" in
...
+                VERSION_SUFFIX=$(echo "${VERSION_ID}" | tr -d ".")
                 if grep beta /etc/os-release;
                 then
                     DISTRO_NAME="rhel-8-beta"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-8-beta.json"
+                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}-beta.json"
                 else
                     DISTRO_NAME="rhel-8"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-8.json"
+                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
```
jrusz, line 77: Now I'm wondering if this will work on our regular RHEL-8 runners. They will still expect `rhel-8.json` no?

jrusz (commented):
Added some questions.

test/cases/regression-composer-works-behind-satellite.sh
atodorov: Duh. We need more conditionals here :-(

jrusz:
I'm running the nightly pipeline with the rpms that got built in this PR, let's see how that looks first https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/365939717

jrusz:
The regression test is still failing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1566246101 ... It's still doing only
 ```
+ REPO1_NAME=baseos
+ REPO2_NAME=appstream
```
even though you have added the condition to change it when it's running on nightly. I don't understand why it does not do it, the same piece of code works if I try it locally so the condition looks ok to me and `NIGHTLY` var is set to `true` so it has to work... I guess this will require some more debugging...

Some notes:
1. Tests fail on aarch64 in the pipeline I linked above because I supplied it only the x86_64 rpms for this test run.
2. The change in regression tests is really included since the echo change on line 68 can be seen in the run.

atodorov:
> The regression test is still failing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1566246101 ... It's still doing only
> 
> ```
> + REPO1_NAME=baseos
> + REPO2_NAME=appstream
> ```
> 
> even though you have added the condition to change it when it's running on nightly. I don't understand why it does not do it, the same piece of code works if I try it locally so the condition looks ok to me 

FTR I've tried it locally yesterday too and it it did work for me as well. Will continue debugging then.

jrusz (approved):
LGTM. We decided to merge this and run the nightly pipeline again with the rebased RPMs. We were unable to reproduce current regression test failure locally.

PR #1687, opened 2021-08-26
cloudapi: add support for S3-compatible targets

chloenayon:
This pull request includes:

### DIFF
--- a/cmd/osbuild-service-maintenance/aws.go
+++ b/cmd/osbuild-service-maintenance/aws.go
@@ -11,8 +11,8 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/cloud/awscloud"
 )
 
-func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey, region string, cutoff time.Time) error {
-	a, err := awscloud.New(region, accessKeyID, accessKey, "")
+func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey, region, endpoint string, cutoff time.Time) error {
+	a, err := awscloud.New(region, endpoint, accessKeyID, accessKey, "")
 	if err != nil {
 		return err
 	}
--- a/cmd/osbuild-service-maintenance/main.go
+++ b/cmd/osbuild-service-maintenance/main.go
@@ -56,7 +56,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		logrus.Info("Cleaning up AWS")
-		err := AWSCleanup(maxCReqs, dryRun, conf.AWSAccessKeyID, conf.AWSSecretAccessKey, "us-east-1", cutoff)
+		err := AWSCleanup(maxCReqs, dryRun, conf.AWSAccessKeyID, conf.AWSSecretAccessKey, "us-east-1", "", cutoff)
 		if err != nil {
 			logrus.Errorf("AWS cleanup failed: %v", err)
 		}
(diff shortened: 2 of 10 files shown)


internal/cloudapi/server.go
```diff
@@ -328,9 +328,13 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
 			}
 
 			key := fmt.Sprintf("composer-api-%s", uuid.New().String())
+			if *awsS3UploadOptions.Endpoint == "" {
+				// reset to default? Find out what the behavior is
```
ondrejbudai, line 332: Let's just pass `nil` in this case?

internal/cloudapi/server.go
```diff
@@ -331,6 +331,7 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
 			t := target.NewAWSS3Target(&target.AWSS3TargetOptions{
 				Filename:        imageType.Filename(),
 				Region:          awsS3UploadOptions.Region,
+				Endpoint:        *awsS3UploadOptions.Endpoint,
```
Gundersanne, line 334: This will dereference `nil` if the endpoint isn't specified (which will be most of the time).  It'll need something similar to [the session token handling](https://github.com/osbuild/osbuild-composer/pull/1666/files#diff-b4cd771217e68c4d9daf38fb95f1526891f973015d00a94d071f5282617959bdR288-R312).

Gundersanne (changes requested):
Thank you!

The first commit can be dropped as that seems to be some noise from #1641. It'll also need a rebase back on main.

ondrejbudai:
Can you please rebase?

Gundersanne:
@ondrejbudai took the liberty of rebasing

ondrejbudai:
Looks good, two thoughts:

- a new entry in `docs/news/unreleased` would be nice
- can we test this somehow? At least a unit test with local https://github.com/minio/minio ?

teg (commented):
This is really cool. I wonder how we could get it integrated with the `docker-compose` setup in the demo so that local testing doesn't need S3 credentials (when testing non-cloud images, I mean). Also, this might need a bit of a rework after @lavocatt's patch for making credentials optional lands.

teg:
> This is really cool. I wonder how we could get it integrated with the `docker-compose` setup in the demo so that local testing doesn't need S3 credentials

To expand on that: wouldn't it be cool if we could have the full setup running locally, and when we build something like a qcow2 (not supported yet), or an edge image (not supported in the UI, so some imagination is needed ;) ), we would get a link to download the artefacts and in production it points to amazon S3, but for testing it points to `localhost`? Similar things should be possible with local container storage if and when we support pushing to quay.io.

Gundersanne:
So on the question of where to put this test. I'd add `cmd/aws-upload/main_test.go`, in the spec file compile that test with `test -c`, add it to the tests pkg; and then call the test in base_tests. We already start a postgres container in that test, just start a minio one as well.

ondrejbudai:
I like this! Can you please run `tools/prepare-source.sh`, and fix the shellcheck error?

cmd/aws-upload/main_test.go
```diff
@@ -0,0 +1,29 @@
+package main
```
Gundersanne, line 1: I think this file can be part of `cmd/osbuild-upload-aws` no?

Gundersanne (changes requested):
Thanks for the test!

A few comments, there's also some checks failing and the cloudapi/v1 changes can be droppe (it's unused).

ondrejbudai:
Done in https://github.com/osbuild/osbuild-composer/pull/2686

PR #1692, opened 2021-08-27
Add support for packit

ochosi:
This is based on the packit configuration of osbuild.

This pull request includes:

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,22 @@
+# https://packit.dev/docs/configuration/
+
+specfile_path: osbuild-composer.spec
+
+synced_files:
+    - osbuild-composer.spec
+    - .packit.yaml
+
+upstream_package_name: osbuild-composer
+downstream_package_name: osbuild-composer
+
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
Makes sense!

msehnout:
Is there any reason to use `fedora-rawhide` instead of `fedora-all`?

ochosi:
> Is there any reason to use `fedora-rawhide` instead of `fedora-all`?

I would just start with `rawhide` and then maybe go to pushing to dist-git there directly and then take care of `fedora-all`

teg:
I'm all for this!

Should we also consider going full source-git to make this even better integrated? I don't know what the blockers would be.

ochosi:
> I'm all for this!
> 
> Should we also consider going full source-git to make this even better integrated? I don't know what the blockers would be.

Funny you should mention this - I discussed this yesterday with Ondřej briefly when discussing the PR / packit. I personally don't know enough about the subject but with a few pointers I'm happy to educate myself :)

PR #1697, opened 2021-08-28
Pre-define an ostree remote for RFE raw images (including created via simplified installer)

gicmo:
Preset a remote called `rhel-edge` with the URL that was used to fetch the commit, so it can be used to fetch updates.

This pull request includes:

### DIFF
--- a/internal/distro/rhel85/pipelines.go
+++ b/internal/distro/rhel85/pipelines.go
@@ -1088,10 +1088,11 @@ func ostreeDeployPipeline(
 	p.Name = "image-tree"
 	p.Build = "name:build"
 	osname := "redhat"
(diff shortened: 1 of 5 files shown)


teg (approved):

teg (changes requested):
We need to discuss whether to land this for 8.5

teg:
This is definitely what we want long-term, but without automated updates, this doesn't really get us all the way. While it is more cumbersome, the user could just as well set the remote when doing the update. So I'm inclined to leave this for 8.6.

teg:
After further discussion, it was decided to not include this for 8.5, but reconsider for 8.6.

PR #1705, opened 2021-08-30
repositories: exclude downstream from upstream

teg:
We don't want to give the impression we support building Fedora on RHEL or
CentOS Stream, or CentOS Stream on RHEL, so drop these repositories from
the package.

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue (this is about dropping untested/unsupported functionality)

### DIFF
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -48,7 +48,7 @@ func TestFetchChecksum(t *testing.T) {
 // this should run cross-arch dependency solving N-1 times.
 func TestCrossArchDepsolve(t *testing.T) {
 	// Load repositories from the definition we provide in the RPM package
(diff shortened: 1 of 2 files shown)


teg:
cc @bcl

ondrejbudai:
We also don't want to build RHEL 9 on RHEL 8 because of possible kernel issues.

ondrejbudai:
I wonder where we should test this. Is pushing a blueprint enough, or do we have to start a compose?

(Also this needs an exception, do not merge until we have it.)

teg:
I suggest we add `test/cases/cross-distro.sh`, which only makes the call to enumerate the distros (using curl), and asserts that if we are on RHEL we only see RHEL, and if we are on RHEL8 we only see RHEL8.

teg:
Closing in favour of #1723

PR #1715, opened 2021-08-31
Filesystem submounts fix

kingsleyzissou:
This pull request includes:

Currently the filesystem configuration allows for mount points and their direct sub-directories to be configured as separate partitions (i.e. `/var` & `/var/log`). This pr expands this support to add an additional level to allow for grandchild-directories (i.e. `/var/log/audit`)
 

### DIFF
--- a/internal/distro/rhel85/distro.go
+++ b/internal/distro/rhel85/distro.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"path"
 	"sort"
+	"strings"
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/disk"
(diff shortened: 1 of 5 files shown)


ondrejbudai:
I wonder if we should allow arbitrary depth. Do we foresee any issues if we do that? cc @teg

kingsleyzissou:
@ondrejbudai good point, that would probably scale a lot better. Will wait to see what @teg suggests

teg:
Agreed, let's lift the limitation on depths. Really what we want to limit is the top-level mounts, what people do under the ones we allow we don't care about.

teg (changes requested):
We need to discuss whether rot land this for 8.5/9.0, but otherwise: great stuff!

ondrejbudai (approved):

ondrejbudai (approved):
Exception granted for both distros, merging.

PR #1717, opened 2021-09-01
Oauth2

Gundersanne:
Accidentally deleted the branch on https://github.com/osbuild/osbuild-composer/pull/1625, sorry for the mess.

See #1625 for previous reviews and comments. I added more information in https://github.com/Gundersanne/osbuild-composer/blob/oauth2/docs/news/unreleased/api-oauth2-support.md as a result.


internal/auth/context.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/golang-jwt/jwt"
+	"github.com/openshift-online/ocm-sdk-go/authentication"
+)
+
+const (
+	UsernameKey string = "username"
```
teg, line 13: Unused?

internal/auth/context.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// Username is stored in token claim with key 'sub'
+	claims, ok := userToken.Claims.(jwt.MapClaims)
+	if !ok {
+		err := fmt.Errorf("Unable to parse JWT token claims")
+		return nil, err
+	}
+
+	// TODO figure out how to unmarshal jwt.mapclaims into the struct to avoid all the
```
teg, line 46: Should this be done before merging?

internal/auth/context.go
```diff
@@ -0,0 +1,90 @@
...
+	//
+	//var accountAuth api.AuthPayload
+	//err := json.Unmarshal([]byte(claims), &accountAuth)
+	//if err != nil {
+	//	err := fmt.Errorf("Unable to parse JWT token claims")
+	//	return nil, err
+	//}
+
+	payload := &AuthPayload{}
+	// default to the values we expect from RHSSO
```
teg, line 57: What's the logic behind these fallbacks? Also, do we need to support all these fields?

teg (approved):
Looking great! Couple of comments, but nothing really worth blocking on.

Gundersanne:
> Looking great! Couple of comments, but nothing really worth blocking on.

Ah I copied context.go from the example service :D I need to rebase anyway so I'll go through it :)

Gundersanne:
@teg turns out I just copied it in without thinking, we're not actually using it. When we need to map claims (for logging purposes or authorization purposes or whatever) we'll just write an echo middleware which does the same thing. So good catch :D

teg (dismissed):

cmd/osbuild-mock-openid-provider/main.go
```diff
@@ -0,0 +1,148 @@
...
++TjE6w7KPLghlnSMlrGk1PZ1K8CHWYj1xymGYQDhLe0QGtBCZv0+wI1UjAI3xCMU
+n+5siufiI6HrtE0ZCUzxEArQMKYxnmZMnl9nuirBevOyFyhx39o2n1iARXTOBrtB
+OuWrC1D1EqGyuhS/Tc/Wf/lpKZDCTUwWrfBRxQL1IDKyh5bBWCc9bGoIvPu3Xfdh
+FMbK8VRsEw5jq3smKQc7mE4rO3jjJWIyLe09vtVZ3lMP301Ud+cawelTnKhJPMkY
+IJl8V39V6f44NjRBvsLoQjPIAVvm4PdvXdzZnqwexr8+vT0Vf3U/klxmbT+2yDgG
+uwIDAQAB
+-----END PUBLIC KEY-----
+`
+
+const jwtPrivateKeyPEM = `
```
ondrejbudai, line 28: Hmm, @jabia99 lately worked on removing all hard-coded test secrets from the repository. I wonder if we should also remove this one otherwise we're basically "regressing". Sorry, I know how painful is to generate keys. :(

jabia99: You can take a look at tools/gen-ssh and do something similar with this key. If you prefer, I could do it myself if you leave the hard coded key like it is.

Gundersanne:
Oof i did a bad rebase here, ignore for now

cmd/osbuild-mock-openid-provider/main.go
Gundersanne: [Just did it inline](https://github.com/osbuild/osbuild-composer/pull/1717/files#diff-8c9fc1f11806f990402d945e7a6a3bdcf8be429120a4b8c2ec18e9ff5251f02fR1177-R1178).

teg (approved):

teg (approved):

PR #1754, opened 2021-09-08
Use V2 results internally

### LINKED ISSUE #1493: OSBuild result parsing (v1 vs v2)
OSBuild outputs the results of pipelines and stages in different formats depending on the schema version of the input manifest (v1 or v2).  Currently, when composer runs an osbuild job using the new v2 manifest schema, the result is parsed and converted to the format of the v1 schema result.
https://github.com/osbuild/osbuild-composer/blob/6fac4dcb7011b6708802f52440955fd832b7eec1/internal/osbuild1/result.go#L195

As part of the transition, we should reverse this and instead convert output and results from v1 schemas to the v2 result format.

I'm opening this issue to track this part of the transition.

achilleas-k:
With the introduction of the new manifest schema (v2) the output from osbuild calls also changed (the `Result` type).  Up until now, osbuild-composer has been converting output from v2 jobs to v1 results.  This PR changes the handling of results so that v1 results are converted to v2 instead and the new v2 Result type is used everywhere internally.

This also changes the format of result data stored in the job queues.  The result value itself is opaque, but it is parsed in a few cases to extract information (e.g., return package lists for finished jobs).

Remaining tasks:
- [x] Test RPM list deduplication
- [x] Test reading from mixed-result job queues (both FS and DB)
- [x] Use pipeline names provided by image types to distinguish build from image pipelines

---

This pull request includes:

---

Closes #1493


internal/worker/json.go
```diff
@@ -75,6 +79,20 @@ type KojiFinalizeJobResult struct {
...
+type PipelineNames struct {
+	Build   []string `json:"build"`
+	Payload []string `json:"payload"`
+}
+
+// Returns a concatenated list of the pipeline names
+func (pn *PipelineNames) All() []string {
+	return append(pn.Build, pn.Payload...)
+}
+
```
Gundersanne, line 95: I think it actually makes sense to add these PipelineNames to the `*JobResults` which have `osbuild.Result`, because we're not actually interested in adding the v2 as new jobtypes (and thus new jobresults), we want to replace the old ones. The new jobs wouldn't do anything new.

Plus you'd make every user of internal/worker responsible to actually handle this v1/v2 distinction.

So rather than duplicating everything which has `osbuild.Result` in a `*JobResultV2`, let's just update those current Results, exactly as this PR does.

achilleas-k: yeah that would be ideal.  I'm going to work up a POC for that in osbuild and talk to @gicmo about it.

achilleas-k:
Rebased on main and copied cloud API changes to v2.

achilleas-k:
Marking ready for review.  I wanted to take the opportunity of working on some more useful osbuild output changes at the same time since we're changing the result format in composer.  It would save us from having to figure out backwards compatibility down the line.  But this is ready now and there are other priorities so it's probably better if we merge it now and deal with the compatibility wrangling down the line.

teg:
Would be great to get this in for the next re-deploy, do we have any known blockers? (apart from reviews)

gicmo:
I have not looked in-depth at this, but from a birds-eye view it totally makes sense, so FWIW: 👍

ondrejbudai (dismissed):
This is awesome! I'm worried that we cannot merge this now because we added a new distro (rhel90) in the meantime. Can you please rebase so we can get this in?

achilleas-k:
Rebased on main and added the required changes to RHEL 9.0.  The new changes were squashed into the old commits that were making the same changes.

achilleas-k:
@ondrejbudai we're green!

achilleas-k:
New conflicts.  Rebasing.

Gundersanne (approved):

PR #1762, opened 2021-09-09
rhel85/90: install docs in rpm to fix missing man pages

ondrejbudai:
When we firstly introduced the v2 manifests, we excluded docs from all rpm
stages because it doesn't make sense to include docs in edge images.

However, when we ported the other image types to v2, we left the flag on.
The side effect of --excludedocs is that we no longer install man pages into
the image. This the default behaviour of rpm and can be seen here:
https://github.com/rpm-software-management/rpm/blob/b88f43b9a3bf39a92ba8ca511d33fdc60f399c1e/macros.in#L555

This is actually quite a major regression in non-edge images, thus this commit
reverts this setting. It would be great to have the option not to install
docs in some rpm stages but as this is a difficult topic and we're already
past the deadline, we will consider this in the future.

I also regenerated all test manifests.

This pull request includes:


ondrejbudai:
As this needs an exception, I'm marking this as a draft for now.

thozza (approved):
Changes look good. I think that we should consider not installing Docs in some specific types of pipelines. E.g. the `build` pipeline most probably does not need Docs to be installed in it. The same goes for the installer or container pipelines, which are just a delivery mechanism for a different "payload" (ostree commit, etc.). However I'm not sure if this would save us any reasonable time or space when building images to justify the additional logic and code.

ondrejbudai:
I agree that we should revisit this later, this is mainly a quick fix so we fix this before the deadline.

PR #1800, opened 2021-09-17
Drop RELEASING.md and point to dev guide

ochosi:
Instead of keeping a howto on releasing in every repository we move the
documentation to a central place in our "developer guide".
See https://github.com/osbuild/guides/pull/45


teg (approved):

ondrejbudai (approved):

ondrejbudai:
Just a docs change, we don't need to wait for :banana: .

PR #1801, opened 2021-09-17
cloudapi: Improve logging

diaasami:
This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -24,6 +24,7 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/weldr"
 	"github.com/osbuild/osbuild-composer/internal/worker"
+	logrus "github.com/sirupsen/logrus"
 )
 
 type Composer struct {
@@ -62,6 +63,7 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *
 	}
 
 	c.distros = distroregistry.NewDefault()
+	logrus.Info(fmt.Sprintf("Loaded %d distros", len(c.distros.List())))
 
 	c.rpm = rpmmd.NewRPMMD(path.Join(c.cacheDir, "rpmmd"), "/usr/libexec/osbuild-composer/dnf-json")
 
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -103,7 +103,7 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener,
 }
 
 func (c *Composer) InitAPI(cert, key string, enableJWT bool, l net.Listener) error {
-	c.api = cloudapi.NewServer(c.logger, c.workers, c.rpm, c.distros)
+	c.api = cloudapi.NewServer(c.workers, c.rpm, c.distros)
 	c.koji = kojiapi.NewServer(c.logger, c.workers, c.rpm, c.distros)
 
 	clientAuth := tls.RequireAndVerifyClientCert
--- a/cmd/osbuild-composer/config.go
+++ b/cmd/osbuild-composer/config.go
@@ -35,6 +35,7 @@ type ComposerConfigFile struct {
 		JWTACLFile string `toml:"jwt_acl_file"`
 	} `toml:"composer_api"`
 	WeldrAPI WeldrAPIConfig `toml:"weldr_api"`
+	logLevel string         `toml:"log_level"`
 }
 
 type WeldrAPIConfig struct {
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/coreos/go-systemd/activation"
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -34,8 +35,17 @@ func main() {
 		if os.IsNotExist(err) {
 			config = &ComposerConfigFile{}
 		} else {
-			log.Fatalf("Error loading configuration: %v", err)
+			logrus.Fatalf("Error loading configuration: %v", err)
 		}
+	}
+
+	logrus.SetOutput(os.Stdout)
+	logLevel, err := logrus.ParseLevel(config.logLevel)
+
+	if err == nil {
+		logrus.SetLevel(logLevel)
+	} else {
+		logrus.Info("Failed to load loglevel from config:", err)
 	}
 
 	log.Println("Loaded configuration:")
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -48,46 +48,46 @@ func main() {
 		logrus.Info("Failed to load loglevel from config:", err)
 	}
 
-	log.Println("Loaded configuration:")
-	err = DumpConfig(config, log.Writer())
+	logrus.Info("Loaded configuration:")
+	err = DumpConfig(config, logrus.StandardLogger().WriterLevel(logrus.InfoLevel))
 	if err != nil {
-		log.Fatalf("Error printing configuration: %v", err)
+		logrus.Fatalf("Error printing configuration: %v", err)
 	}
 
 	stateDir, ok := os.LookupEnv("STATE_DIRECTORY")
 	if !ok {
-		log.Fatal("STATE_DIRECTORY is not set. Is the service file missing StateDirectory=?")
+		logrus.Fatal("STATE_DIRECTORY is not set. Is the service file missing StateDirectory=?")
 	}
 
 	cacheDir, ok := os.LookupEnv("CACHE_DIRECTORY")
 	if !ok {
-		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
+		logrus.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 
 	composer, err := NewComposer(config, stateDir, cacheDir, logger)
 	if err != nil {
-		log.Fatalf("%v", err)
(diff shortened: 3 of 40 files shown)


internal/cloudapi/v2/errors.go
```diff
@@ -219,7 +220,8 @@ func (s *Server) HTTPErrorHandler(echoError error, c echo.Context) {
 			apiErr := APIError(code, sec, c)
 
 			if sec.httpStatus == http.StatusInternalServerError {
-				c.Logger().Errorf("Internal server error. Code: %s, OperationId: %s", apiErr.Code, apiErr.OperationId)
+				logrus.Errorf("Internal server error. Code: %s, OperationId: %s, internalError: %v", apiErr.Code, apiErr.OperationId,
+					echoError.(*echo.HTTPError).Internal)
```
Gundersanne, line 224: hm this type assertion will fail in of ErrorNotHTTPError no?

internal/worker/server.go
```diff
@@ -74,10 +76,10 @@ func (s *Server) WatchHeartbeats() {
 	for range time.Tick(time.Second * 30) {
 		for _, token := range s.jobs.Heartbeats(time.Second * 120) {
 			id, _ := s.jobs.IdFromToken(token)
-			log.Printf("Removing unresponsive job: %s\n", id)
+			logrus.Info(fmt.Sprintf("Removing unresponsive job: %s\n", id))
```
Gundersanne, line 79: logrus.Infof

internal/worker/server.go
```diff
@@ -74,10 +76,10 @@ func (s *Server) WatchHeartbeats() {
...
 			err := s.FinishJob(token, nil)
 			if err != nil {
-				log.Printf("Error finishing unresponsive job: %v", err)
+				logrus.Error(fmt.Sprintf("Error finishing unresponsive job: %v", err))
```
Gundersanne, line 82: Errorf

internal/worker/server.go
```diff
@@ -270,7 +272,7 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
 	if s.artifactsDir != "" {
 		err := os.Rename(path.Join(s.artifactsDir, "tmp", token.String()), path.Join(s.artifactsDir, jobId.String()))
 		if err != nil {
-			log.Printf("Error moving artifacts for job%s: %v", jobId, err)
+			logrus.Error(fmt.Sprintf("Error moving artifacts for job %s: %v", jobId, err))
```
Gundersanne, line 275: Errorf

Gundersanne (commented):
Looking good!

Also  (I can't comment below the diffs which is annoying):
In the case of the HTTPErrorHandler not receiving an echo error (so it would return `ErrorNotHTTPError`, line 242), I'd also log that error before responsing with ErrorNotHTTPError. I mean at that point something went very wrong but still.

Gundersanne:
After some more discussion:

```
import ( 
        logrus "github.com/sirupsen/logrus"
)
```
will actually give you a global instance. When importing in the internal/* packages it's the same instance.

So let's configure this instance in `cmd/`'s main() (be it the worker, or composer). And then only use this instance in the `internal/*` packages. We won't need init() for this either I think.

The logging level should be configurable.

And maybe the  format in future. text (default), or json (might be useful for the service). But let's do that later.

cmd/osbuild-composer/main.go
```diff
@@ -19,6 +20,11 @@ var repositoryConfigs = []string{
 	"/usr/share/osbuild-composer",
 }
 
+func init() {
+	logrus.SetOutput(os.Stdout)
+	logrus.SetLevel(logrus.InfoLevel)
+}
+
```
Gundersanne, line 27: configuration in init() can be dropped

cmd/osbuild-composer/composer.go
```diff
@@ -43,6 +44,11 @@ type Composer struct {
 	weldrListener, localWorkerListener, workerListener, apiListener net.Listener
 }
 
+func init() {
+	logrus.SetOutput(os.Stdout)
+	logrus.SetLevel(logrus.InfoLevel)
```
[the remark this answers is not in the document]
Gundersanne: can be dropped

cmd/osbuild-composer/main.go
```diff
@@ -38,75 +44,75 @@ func main() {
 		}
 	}
 
-	log.Println("Loaded configuration:")
-	err = DumpConfig(config, log.Writer())
+	logrus.Info("Loaded configuration:")
```
Gundersanne, line 47: configure the logger here after reading the config.

internal/common/echo_logrus.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// EchoLogrusLogger extend logrus.Logger
+type EchoLogrusLogger struct {
+	*logrus.Logger
+}
+
+var commonLogger = &EchoLogrusLogger{
+	Logger: logrus.New(),
```
Gundersanne, line 17: do we have to instantiate a  new logger? can't we just use the static one? Becuase this logger will hve to be configured as well.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -88,6 +89,11 @@ type dbJobQueue struct {
 	pool *pgxpool.Pool
 }
 
+func init() {
+	logrus.SetOutput(os.Stdout)
+	logrus.SetLevel(logrus.InfoLevel)
+}
+
```
Gundersanne, line 96: can be dropped

internal/cloudapi/v2/errors.go
diaasami: I cannot be completely sure so I added a check

cmd/osbuild-composer/config.go
```diff
@@ -35,6 +35,9 @@ type ComposerConfigFile struct {
 		JWTACLFile string `toml:"jwt_acl_file"`
 	} `toml:"composer_api"`
 	WeldrAPI WeldrAPIConfig `toml:"weldr_api"`
+	Logging  struct {
+		logLevel string `toml:"log_level"`
```
Gundersanne, line 39: Maybe we can move this option just to the top-level struct? I know we don't have any options there yet but it doesn't really belong to any particular api. wdyt @teg

teg: Sounds good to me.

cmd/osbuild-composer/main.go
```diff
@@ -38,75 +39,84 @@ func main() {
 		}
 	}
 
-	log.Println("Loaded configuration:")
-	err = DumpConfig(config, log.Writer())
+	logrus.SetOutput(os.Stdout)
+	logLevel, err := logrus.ParseLevel(config.Logging.logLevel)
```
Gundersanne, line 43: Hm what will happen now if you don't explicitly specify the logging level? Maybe we could add the info level to `GetDefaultConfig`.

cmd/osbuild-composer/main.go
```diff
@@ -38,75 +39,84 @@ func main() {
...
+
+	if err == nil {
+		logrus.SetLevel(logLevel)
+	} else {
+		log.Println("Failed to set log level", err)
```
Gundersanne, line 48: As this would only happen with an invalid config, this can be logrus.Fatal imo.

internal/common/echo_logrus.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return log.OFF
+}
+
+func (l *EchoLogrusLogger) Output() io.Writer {
+	return l.Out
+}
+
+func (l *EchoLogrusLogger) SetOutput(w io.Writer) {
+	l.Out = w
```
Gundersanne, line 59: Hm does this mean theoretically we could change the app-wide logger via echo? Maybe this should be a non-op.

internal/common/echo_logrus.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (l *EchoLogrusLogger) Level() log.Lvl {
+	return toEchoLevel(l.Logger.Level)
+}
+
+func (l *EchoLogrusLogger) SetLevel(v log.Lvl) {
+	l.Logger.Level = toLogrusLevel(v)
```
Gundersanne, line 67: Same I don't think we should set the global logger via echo. All the configuring should happen in cmd/. So imo this should be a nop.

cmd/osbuild-composer/main.go
diaasami: This would also happen if this field is missing in the config, should we make it a `Warn` ?

cmd/osbuild-composer/main.go
diaasami: I found out that default log level in logrus is `Info`, if we don't specify it in the config, then nothing happens here since it will be an empty string and not parse-able.

cmd/osbuild-composer/main.go
Gundersanne: Well the default is info, and not configuring it should be allowed. So I'd just make it info.

cmd/osbuild-composer/main.go
Gundersanne: Right, it's just relying on the parsing I guess to decide the default then. In that case the `Failed to set log level` is really expected behaviour though, and should just be an info message.

Gundersanne (dismissed):
LGTM!

cmd/osbuild-composer/composer.go
```diff
@@ -62,6 +63,7 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string, logger *
 	}
 
 	c.distros = distroregistry.NewDefault()
+	logrus.Info(fmt.Sprintf("Loaded %d distros", len(c.distros.List())))
```
Gundersanne, line 66: `logrus.Infof` maybe in one of the followups.

Gundersanne (approved):

PR #1809, opened 2021-09-20
schutzbot: Update terraform sha

Gundersanne:


ondrejbudai:
I think we need the same fix also in openstack.

ondrejbudai (approved):

PR #1814, opened 2021-09-21
api/cloud: Move depsolving to workers

ondrejbudai:
Proof of concept of the simplest way to move depsolving to workers. Do not try this at home unless you are a trained professional.

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -18,6 +18,7 @@ import (
 	"github.com/BurntSushi/toml"
 
 	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/upload/azure"
 	"github.com/osbuild/osbuild-composer/internal/upload/koji"
 	"github.com/osbuild/osbuild-composer/internal/worker"
@@ -131,6 +132,7 @@ func main() {
 		log.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 	store := path.Join(cacheDirectory, "osbuild-store")
+	rpmmd_cache := path.Join(cacheDirectory, "rpmmd")
 	output := path.Join(cacheDirectory, "output")
 	_ = os.Mkdir(output, os.ModeDir)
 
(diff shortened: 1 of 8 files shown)


internal/cloudapi/v1/v1.go
```diff
@@ -157,25 +158,45 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
...
+				panic(err)
+			}
+			if status.Canceled {
+				panic("unexpected cancellation of a depsolve job")
+			}
+			if !status.Finished.IsZero() {
+				break
 			}
-			pkgSpecSets[name] = pkgs
+			time.Sleep(500 * time.Millisecond)
```
teg, line 184: Considering depsolving should take ~100ms once we no longer flush the caches, maybe  make the sleep 50ms rather than 500?

ondrejbudai: yep, that's a good point

cmd/osbuild-worker/main.go
```diff
@@ -237,6 +239,9 @@ func main() {
 		"koji-finalize": &KojiFinalizeJobImpl{
 			KojiServers: kojiServers,
 		},
+		"depsolve": &DepsolveJobImpl{
+			RPMMD: rpmmd.NewRPMMD(rpmmd_cache, "/usr/libexec/osbuild-composer/dnf-json"),
```
teg, line 243: The worker needs to pull in `dnf-json`. Put it in a sub-package that both composer and the worker depends on?

internal/worker/json.go
```diff
@@ -72,6 +73,19 @@ type KojiFinalizeJobResult struct {
...
+type DepsolveJob struct {
+	PackageSets      map[string]rpmmd.PackageSet `json:"package_sets"`
+	Repos            []rpmmd.RepoConfig          `json:"repos"`
+	ModulePlatformID string                      `json:"module_platform_id"`
+	Arch             string                      `json:"arch"`
+	Releasever       string                      `json:"releasever"`
+}
+
+type DepsolveJobResult struct {
+	PackageSpecs map[string][]rpmmd.PackageSpec `json:"package_specs"`
```
teg, line 92: I think using structs from another package for serialisation is an anti-pattern we should try to avoid (who even wrote this patch?), but I would not consider this a blocker, as this would anyway need more cleanup across the codebase.

thozza:
I must say that I find our split into sub-packages quite weird. What is the purpose of `osbuild-composer-core`? I would actually expect parts common to the composer and worker to be there, such as the `dnf-json` and the actual binary of `osbuild-composer` to be in the main package, which for unknown reason contains only `.service` files... Would be someone able to explain to be the motivation behind this?

PR #1820, opened 2021-09-22
`generate-test-cases`: small fix and drop `--with-customizations` option

thozza:
Generating image test case with all possible blueprint customizations is supported only for `qcow2` image type. Generating `qcow2` images with customizations was not a lot of fun. One had to use a special CLI option
for this case.

To streamline the generation of image test cases, drop the `--with-customizations` option from the `generate-test-cases` script and move the functionality to `format-request-map.json` by defining a new `qcow2-customize` test case. This is very similar to what `*edge-rt` test case already uses. This mean that when the test case for `qcow2` image type is being generated, actually two test cases will be generated. The filesystem customizations since RHEL-8.5 are handled through distro-specific `overrides`.

In addition:
- Ensure that the `supported_arches` from `format-request-map.json` does not leak to the generated image test case, since it is relevant only for the generator script.
- remove redundant overrides from `format-request-map.json`

This pull request includes:

### DIFF
--- a/tools/test-case-generators/distro-arch-imagetype-map.json
+++ b/tools/test-case-generators/distro-arch-imagetype-map.json
@@ -5,7 +5,6 @@
             "fedora-iot-commit",
             "openstack",
             "qcow2",
-            "qcow2-customize",
             "vhd",
             "vmdk"
         ],
@@ -18,7 +17,6 @@
             "ami",
             "openstack",
             "qcow2",
-            "qcow2-customize",
             "tar",
             "vhd",
             "vmdk"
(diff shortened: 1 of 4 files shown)


ondrejbudai (approved):
:heart: :heart_eyes: :heart: :heart_eyes: :heart: :heart_eyes: :heart: :heart_eyes: :heart: :heart_eyes: :heart:

ondrejbudai (approved):

kingsleyzissou (approved):
Wow! So much cleaner & better. Love it

PR #1827, opened 2021-09-23
apiv2: specify aws creds is optional

lavocatt:
If the aws secrets (secretkey + keyId) are empty and are specified
in the worker, the worker uses those credentials.

This pull request includes:


cmd/osbuild-worker/main.go
```diff
@@ -97,6 +98,10 @@ func main() {
 			OAuthURL         string `toml:"oauth_url"`
 			OfflineTokenPath string `toml:"offline_token"`
 		} `toml:"authentication"`
+		AWS *struct {
```
teg, line 101: I think this should be done like GCP and Azure, and simply take the path to the credentials file.

cmd/osbuild-worker/main.go
```diff
@@ -218,13 +223,23 @@ func main() {
 		}
 	}
 
+	var awsCredentials *credentials.Credentials
+	if config.AWS != nil {
+		awsCredentials = credentials.NewStaticCredentials(*config.AWS.AccessKeyId, *config.AWS.SecretAccessKey, "")
```
teg, line 228: I would use `NewSharedCredentials()`, and pass it the path to the file with the creds.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -36,6 +38,20 @@ func appendTargetError(res *worker.OSBuildJobResult, err error) {
...
+func (impl *OSBuildJobImpl) getAWS(region string, accessId string, secret string, token string) (*awsupload.AWS, error) {
+	if accessId != "" && secret != "" {
+		return awsupload.New(region, accessId, secret, token)
+	} else {
+		cred, err := impl.AWSCreds.Get()
+		if err != nil {
+			return nil, err
+		}
+		return awsupload.New(region, cred.AccessKeyID, cred.SecretAccessKey, cred.SessionToken)
+	}
```
teg, line 52: We should also support a third option (which we should actually be using in our workers): no credentials are passed in at all, and the AWS SDK will use the instance credentials, in case the VM is running in EC2. See: https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials

We may need to change the API of `awsupload.New()` to take a `creds` object as argument, rather than the individual pieces. And the fallback would then be to pass `nil`, which should (though I did not try) use instance credentials.

teg (changes requested):
Looking very nice! Thanks!

I have two suggestions:
 - Use a separate file to store the credentials, as is done for Azure and GCP
 - Use instance credentials when no creds are passed in the job and no creds have been provisioned in the configuration. This should be the default in the service.

cmd/osbuild-worker/main.go
ondrejbudai: +1, this struct is printed to stdout when a worker starts, thus we would leak credentials.

lavocatt: Good for me.

cmd/osbuild-worker/jobimpl-osbuild.go
lavocatt: I agree.
Well I think, in the end, we can have two "constructors" for our awsupload.AWS object, one that initialize it with individual pieces, and another that initializes it with a filename.

cmd/osbuild-worker/main.go
lavocatt: In the end, I removed the initialization of the credentials here. It was not making a lot of sense.

cmd/osbuild-worker/main.go
lavocatt: I'm wondering why we want to give the AWS credentials through the `toml` config file because, if we use the function to read credentials from files from the AWS library it reads them from the file pointed by `$AWS_SHARED_CREDENTIALS_FILE` or default to `$HOME/.aws/credentials`.

ondrejbudai: So how would this work? I assume something like this?
1) Creds from a target would have the highest priority
2) Then, the library function would be called and it looks into `AWS_SHARED_CREDENTIALS_FILE` and `$HOME/.aws/credentials`.


Am I right?

lavocatt: Exactly. Then the lib looks into `AWS_SHARED_CREDENTIALS_FILE` first, and if empty, looks in `$HOME/.aws/credentials`.

ondrejbudai: Nice, I like that.

ondrejbudai: btw does it work also with an attached IAM policy on a EC2 instance?

teg: For instance policies: if no creds are found using any of the methods described above, i believe we just need to pass `nil` as the creds object and the right thing should happen. I'm really not sure about that though, so something to check.

teg: By the way, only "problem" I could imagine with not specifying the file would be that it is inconsistent with how the others work. Maybe they can be made to work in similar ways too? Or maybe it does not matter. I would not block on it though.

.gitlab-ci.yml
```diff
@@ -291,6 +291,7 @@ API:
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/api.sh ${TARGET}
+    - /usr/libexec/tests/osbuild-composer/api_v2.sh ${TARGET}
```
Gundersanne, line 294: We don't need to run this for each target. I know they're skipped in the shell script but this will result in a VM being provisioned and immediately torn down. Just add it for 1 target (see centos defined below).

Gundersanne (changes requested):

PR #1829, opened 2021-09-23
cloudapi: use Recover middleware to handle panics

diaasami:
Recover middleware recovers from panics anywhere in the
chain & prints stack trace

This pull request includes:


teg (commented):
This sounds cool. But could you elaborate a bit in the commit message? What kind of things are we catching in this way, and what is the change in behaviour compared to the current one?

Gundersanne:
Ah @diaasami discussed this yesterday. 

So by default a panic will not crash the webserver of course, but the client will just get an empty reply. Using this, which simply calls `recover()`, it will make it to the errorhandler and return a 500 at least. And it will log the error using `echo.Logger` which with some minor work we can hook up to logrus, instead of stderr.

That said they'd probably get the `ErrorNotHTTPError` error code, which is true of course but not very informative? Is it worth just writing our own and returning a specific error, `ErrorRecoveredFromPanic` or something?

diaasami:
> That said they'd probably get the `ErrorNotHTTPError` error code, which is true of course but not very informative? Is it worth just writing our own and returning a specific error, `ErrorRecoveredFromPanic` or something?

I do not think an error without details and message would be helpful for the client, may be only for metrics.

Gundersanne:
> > That said they'd probably get the `ErrorNotHTTPError` error code, which is true of course but not very informative? Is it worth just writing our own and returning a specific error, `ErrorRecoveredFromPanic` or something?
> 
> I do not think an error without details and message would be helpful for the client, may be only for metrics.

Yea ErrorRecoveredFromPanic is about as useful as ErrorNotHTTPError, plus the details of those errors should really only be seen by us as who knows what they might contain.

teg:
I think some generic 500 makes sense and details in the logs. Don't think we really need to say it was a panic, but don't have a strong opinion.

Gundersanne:
Agreed, this is fine as is.

Gundersanne (approved):

PR #1833, opened 2021-09-24
distribution: Use After=network.target instead of multi-user

msehnout:
There is no need for osbuild-composer to wait until the whole multi-user
target is completed. It can be started earlier as it doesn't have any
dependencies in the target.

This can be a problem if there is a unit in the target that is not starting
and still is unrelated to osbuild-composer. There was a bug like this
with Plymouth where the service didn't finish and it was hanging. That
prevented osbuild-composer from starting and the user was left with
working SSH connection and shell, but composer-cli and systemctl start
osbuild-composer.service were both unresponsive.

Replace After=multi-user.target with After=network.target to start
osbuild-composer earlier.

This pull request includes:

### DIFF
--- a/distribution/osbuild-composer.service
+++ b/distribution/osbuild-composer.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=OSBuild Composer
-After=multi-user.target
+After=network.target
 
 # Weldr API needs a local worker by default.
 # Run `systemctl mask osbuild-worker@1.service`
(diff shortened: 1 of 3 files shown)


thozza (approved):
LGTM

teg (approved):
Thanks for figuring this out! Yeah, this was clearly a mistake, which I should have caught, being the systemd veteran on the team ;)

ondrejbudai (approved):

PR #1841, opened 2021-09-28
cloudapi: Replace upload types with image types

Gundersanne:
Because there's only a few combinations of upload types and image types
that make sense, enforce correct combinations by eliminating upload
types.

This pull request includes:


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -483,22 +483,30 @@ components:
...
           $ref: '#/components/schemas/OSTree'
-        upload_request:
-          $ref: '#/components/schemas/UploadRequest'
+        upload_options:
+          $ref: '#/components/schemas/UploadOptions'
+    ImageTypes:
+      type: string
+      enum:
+        - aws.ec2
+        - aws.s3
```
teg, line 505: I don't think we need to support S3 here, we should only ever produce AMIs in EC2.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -483,22 +483,30 @@ components:
...
+        - gcp
+        - azure
+        - rhel-edge-commit.aws.s3
+        - rhel-edge-installer.aws.s3
```
teg, line 509: I'm also unsure if we need to specify `aws.s3` here. Really this should take no options, we just want to upload a blob "somewhere" and get back a URL. Or am I oversimplifying?

Gundersanne: well I'm not sure, don't we want to have these as bootable images as well at some point? @achilleas-k idk what's best here, remove the `aws.s3` bit?

teg: The installer is meant to be downloaded, not to be booted in AWS. We should eventually have edge images in AWS, but they would be their own image type.
Oh, we should call these things `edge-commit` and `edge-installer`, without the `rhel` prefix, no? (Should work for centos too)

achilleas-k: I agree.  No suffix for blob types, I think.  The upload location shouldn't matter in the end.  The fact that it's s3 doesn't need to be visible and we could (theoretically) change the location at some point and it would (or should) make no difference.

Gundersanne: https://github.com/osbuild/osbuild-composer/blob/main/internal/distro/rhel84/distro.go#L869 hmm they're explicitly called rhel-edge-* :/

Gundersanne: let's do that mapping in crc then.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -598,10 +551,17 @@ components:
...
+      required:
+        - region
+      properties:
+        region:
+          type: string
+          example: 'eu-west-1'
+    GCPUploadOptions:
       type: object
       required:
-        - bucket
```
Gundersanne, line 604: @thozza Thoughts on dropping this and fixing/pinning the bucket in composer itself? We'll do the same for aws, just wanted to double-check that we can do this for GCP as well.

Gundersanne:
I added the uploadTypes back in, as they're needed in the uploadstatus.

internal/cloudapi/v2/openapi.v2.yml
thozza: What is the plan WRT sharing the built image with the customer in the nearest region? GCP has ~3 major geographic multi-regions, so we may need to decide which one to use for uploading the image and then importing it. I'm not sure how will composer determine the desired location. If there will be a way to specify it to composer, then yes, the logic to choose the appropriate storage bucket can be implemented in composer and based on its configuration. I just expect what we don't necessarily want to always default to North America region, although it may be good enough to start with.

thozza: Ah, so I see that the `region` is still part of the `GCPUploadOptions`. In this case we will need to implement the logic to choose the appropriate bucket in composer based on the region...

Gundersanne: We could do something like "bucket + '.' + region" ; so `image-builder.service.eu` or whatever.

thozza: How will this problem solved for AWS? Does it exist there at all? I'm not completely sure if choosing a bucket in the appropriate region matters too much for AWS.

Combining `bucket` and `region` into single string seems unnecessarily complicated to me. It definitely does not make sense to expose the `bucket` in `image-builder` API. As for the composer Cloud API, I would propose two options:

1. We keep the `bucket` in the request, instead of combining `bucket` and `region` into single string.
2. Remove the `bucket` from the request and implement a logic and configuration in composer, which will choose the appropriate bucket based on the requested region and have the code be shared by all cloud providers where this matters (my expectation is that it could be AWS and GCP at least?).
3. ... (open to suggestions)

Gundersanne: Ah it seems aws actually has the same restriction as gcp here actually, buckets need to be uniquely named, cross-region.

I don't think we should allow users to specify it here either, since we're the only ones who control these buckets, and if at one point a 2nd service other than our own depends on the cloudapi they can't really choose these anyway, we'd have to tell them what to use.

Gundersanne:
opened https://github.com/osbuild/osbuild-composer/issues/1854 to deal with the GCP region; we'll drop bucket at some point.

internal/cloudapi/v2/v2.go
```diff
@@ -427,6 +422,22 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
+func imageTypeFromApiImageType(it ImageTypes) string {
+	switch it {
+	case ImageTypes_aws:
+		return "ami"
+	case ImageTypes_gcp:
+		return "vhd"
+	case ImageTypes_azure:
+		return "vhd"
+	case ImageTypes_edge_commit:
+		return "rhel-edge-commit"
```
teg, line 434: Does this still work for 8.5? Did we add an alias, or did we actually rename it to `edge-commit`, etc? @achilleas-k ?

achilleas-k: The canonical name is `edge-commit` but it's also still registered under the alias `rhel-edge-commit`.

Gundersanne: Ah shall I change that?

achilleas-k: `edge-commit` and `edge-installer` wont work in 8.4 but `rhel-edge-*` work in 8.4 and 8.5 (and 8.6 if we decide to keep the aliases).  So maybe we should keep the rhel prefix internally.

teg: I would say expose it always as `edge-commit`, and then internally we can translate to `rhel-edge-commit` (which we can drop eventually if we no longer support 8.4).

@achilleas-k probably should keep the alias in 8.X, backwards compatibility and all that.

thozza: Just a note.... Images are never exposed as their alias, only using their main name. Aliases were added mainly with the backward compatibility in mind, so that you won't see them, when listing available image types, but if you use them to submit a compose, it will work just fine.

Architecture's `GetImageType()` checks also aliases
https://github.com/osbuild/osbuild-composer/blob/1e724be8e969b0d741f97d3bb7108e2218449de7/internal/distro/rhel85/distro.go#L124-L137

While `ListImageTypes()` does not

https://github.com/osbuild/osbuild-composer/blob/1e724be8e969b0d741f97d3bb7108e2218449de7/internal/distro/rhel85/distro.go#L115-L122

Gundersanne: So currently `edge-(commit|installer)` is exposed and translated to `rhel-edge-(commit|installer)`. This api should only ever support rhel-ga out of the rhels (so only rhel8.4 atm, maybe some day rhel 8.x and rhel 9.x at the same time), so I think  this all makes sense then? When rhel 8.5 goes ga we change the internal translation.

cmd/osbuild-composer/config.go
```diff
@@ -17,14 +17,19 @@ type ComposerConfigFile struct {
...
-	JWTACLFile     string   `toml:"jwt_acl_file"`
+	AllowedDomains []string  `toml:"allowed_domains"`
+	CA             string    `toml:"ca"`
+	EnableTLS      bool      `toml:"enable_tls"`
+	EnableMTLS     bool      `toml:"enable_mtls"`
+	EnableJWT      bool      `toml:"enable_jwt"`
+	JWTKeysURL     string    `toml:"jwt_keys_url"`
+	JWTKeysCA      string    `toml:"jwt_ca_file"`
+	JWTACLFile     string    `toml:"jwt_acl_file"`
+	AWS            AWSConfig `toml:"aws_config"`
```
thozza, line 28: I'm not sure that it makes sense to have the `AWS` configuration structure in the `KojiAPIConfig`, it seems like a bug to me. As far as I understand the change, the AWS configuration is relevant for the CloudAPI, not KojiAPI. It would probably make more sense to add a new `CloudAPiConfig` structure and have `AWS` moved to it.

cmd/osbuild-composer/composer.go
```diff
@@ -105,7 +105,7 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener,
 }
 
 func (c *Composer) InitAPI(cert, key string, enableTLS bool, enableMTLS bool, enableJWT bool, l net.Listener) error {
-	c.api = cloudapi.NewServer(c.workers, c.rpm, c.distros)
+	c.api = cloudapi.NewServer(c.workers, c.rpm, c.distros, c.config.Koji.AWS.Bucket)
```
thozza, line 108: Taking a configuration value from Koji API configuration to create a Cloud API server does not seem correct to me.

thozza (dismissed):
Changes look good to me, but I have a question about the newly added `AWS` configuration, which I believe should be resolved before merging.

cmd/osbuild-composer/config.go
Gundersanne: That is a great idea, thank you <3

teg (approved):
Still like this. @thozza raises a good point, but I agree we can address it separately.

PR #1850, opened 2021-09-29
Split jobs

Gundersanne:
- [ ] #1841


teg:
Nice!

I wonder about a couple of things though:
- should the region/accounts to share with be arguments of /build rather than /compose? That way you could rebuild/reupload/reshare an image without having to redepsolve
- similarly, should we have builds/<id> as children of a compose where the results of the build (ami id etc) can be found. That would allow us to rebuild an image in case it expired as we still have the manifests.

Backwards compatibility might be tricky in this case, but could do something like requiring the caller to call both compose and build in succession with no other calls in-between and not allow more than one build for now.

Gundersanne:
> Nice!
> 
> I wonder about a couple of things though:
> 
>     * should the region/accounts to share with be arguments of /build rather than /compose? That way you could rebuild/reupload/reshare an image without having to redepsolve
> 
>     * similarly, should we have builds/ as children of a compose where the results of the build (ami id etc) can be found. That would allow us to rebuild an image in case it expired as we still have the manifests.
> 
> 
> Backwards compatibility might be tricky in this case, but could do something like requiring the caller to call both compose and build in succession with no other calls in-between and not allow more than one build for now.



> Nice!
> 
> I wonder about a couple of things though:
> 
>     * should the region/accounts to share with be arguments of /build rather than /compose? That way you could rebuild/reupload/reshare an image without having to redepsolve

So duplicate upload_options to `/build` ?  (can't remove it from /compose rn because it would break ofc).

> 
>     * similarly, should we have builds/ as children of a compose where the results of the build (ami id etc) can be found. That would allow us to rebuild an image in case it expired as we still have the manifests.

The queryparam 'id' of `/composes/{id}/build` is the compose job. So doesn't it have access  by just getting the job status?

> 
> Backwards compatibility might be tricky in this case, but could do something like requiring the caller to call both compose and build in succession with no other calls in-between and not allow more than one build for now.

PR #1856, opened 2021-09-30
config: don't fail LoadConfig when file doesn't exist

achilleas-k:
When the config file doesn't exist, don't return because we need to keep the default and also load from env.

This pull request includes:


thozza (dismissed):
Looks good. Thanks a lot for spotting the problem and submitting a fix! I missed this use-case when I initially added the default configuration... 😕

ondrejbudai (dismissed):

achilleas-k:
Rebased for RHEL 9 CI fixes

thozza (approved):

PR #1857, opened 2021-09-30
Enable jwt disable tls

Gundersanne:


teg:
Could we make jwt and TLS independent of each other? Not entirely sure if that makes sense, but in principle you could combine them in any way, right?

teg:
mTLS and JWT should probably be mutually exclusive and error out if you enable both...

Gundersanne:
> mTLS and JWT should probably be mutually exclusive and error out if you enable both...

Is an extra option for mtls really required? Like currently there's 2 options for auth so...

test/cases/api.sh
```diff
@@ -1205,21 +1203,19 @@ TOKEN="$(curl localhost:8081/token | jq -r .access_token)"
 
 [ "$(curl \
         --silent \
-        --cacert /etc/osbuild-composer/ca-crt.pem \
         --output /dev/null \
         --write-out '%{http_code}' \
         --header "Authorization: Bearer $TOKEN" \
-        https://localhost/api/composer/v1/version)" = "200" ]
+        http://localhost:443/api/composer/v1/version)" = "200" ]
```
teg, line 1209: Not port 80?

cmd/osbuild-composer/testdata/test.toml
```diff
@@ -1,6 +1,9 @@
 [koji]
 allowed_domains = [ "osbuild.org" ]
 ca = "/etc/osbuild-composer/ca-crt.pem"
+enable_jwt = false
```
teg, line 4: Why did this change from `true` to `false`?

test/cases/api.sh
Gundersanne: We'd need a 2nd socket file for this. Happy to do that, we'd remove the tls config option.

cmd/osbuild-composer/testdata/test.toml
Gundersanne: because jwt + mtls (latter is true by default) is an invalid combo. I know it's just the config test but it seems weird tetsing an invalid config.

teg: ah, right!

test/cases/api.sh
teg: Right, and we want to drop these systemd sockets and just move to always use containers in our deployments/tests anyway, so I don't think it is worth adding a second socket which we will then drop again soon anyway... We should create a ticket to change the tests to actually test the containers we build and push to quay, but for now this is good enough I think.

teg (approved):

PR #1862, opened 2021-10-01
worker: use logrus for logging

diaasami:
This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"flag"
-	"log"
 	"os"
 
 	"github.com/coreos/go-systemd/activation"
@@ -25,9 +24,8 @@ func main() {
 	flag.BoolVar(&verbose, "verbose", false, "Print access log")
 	flag.Parse()
 
-	var logger *log.Logger
-	if verbose {
-		logger = log.New(os.Stdout, "", 0)
+	if !verbose {
+		logrus.Print("verbose flag is provided for backward compatibility only, current behavior is always printing the access log")
 	}
 
 	config, err := LoadConfig(configFile)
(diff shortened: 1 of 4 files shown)


cmd/osbuild-worker/main.go
```diff
@@ -8,14 +8,14 @@ import (
...
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
 	"path"
 	"strings"
 	"time"
 
 	"github.com/BurntSushi/toml"
+	log "github.com/sirupsen/logrus"
```
bcl, line 18: Is there a good reason to call this log instead of logrus? All the other files are using logrus so it could be confusing.

diaasami: Good point, I'll fix it, thanks

Gundersanne (changes requested):
Nice! The worker server is also an echo one, can we use the `EchoLogrusLogger` there as well?  And do the same magic with the errors as in the cloudapi/v2.

cmd/osbuild-composer/main.go
```diff
@@ -25,9 +24,15 @@ func main() {
 	flag.BoolVar(&verbose, "verbose", false, "Print access log")
 	flag.Parse()
 
-	var logger *log.Logger
-	if verbose {
-		logger = log.New(os.Stdout, "", 0)
+	verboseFound := false
+	flag.Visit(func(f *flag.Flag) {
```
Gundersanne, line 28: Why use `Visit` here? Since it's just a binary flag it's either true (specified) or false (not specified) no? And if specified display the info message.

diaasami: the flag being `false` means it's either default value or specified as false, no way to tell which, my initial idea was to issue the message if the flag is used (regardless of value)
But I think it's simpler to show the message when the behavior differs, so when verbose == false.

Gundersanne (changes requested):
Good as is, but could you squash the first 2 commits? Or move the stuff that changes composer's logging in `worker: Improve logging` to the `composer: logging` commit? Either is fine.

Gundersanne (approved):

PR #1863, opened 2021-10-01
rpmmd: use repo name as ID for DNF, instead of list index

thozza:
All of the repo definitions added after RHEL-8.3 consistently contain
'name' of each repository. However the rpmmd functions don't use it, but
instead use the index of the repository in the list of repositories.
This makes it a bit harder to determine the appropriate repository in
case of DNF failures such as:

```
2021/10/01 10:33:53 Failed to load subscriptions. osbuild-composer will
fail to build images if the configured repositories require them: no
matching key and certificate pair
Errors during downloading metadata for repository '3':
  - Status code: 403 for
    https://rpmrepo-storage.bucket.vpce-08d3201af28373567-o10c2v4q.s3.us-east-1.vpce.amazonaws.com/data/rhvpn/el8/sha256-518e7c455fb0bb23e32679c45a45cc269bdd5bd915b560b98bfd699aa4f100e0
(IP: 10.30.18.76)
panic: Could not depsolve: DNF error occured: RepoError: Error occurred
when setting up repo: Failed to download metadata for repo '3': Cannot
download repomd.xml: Cannot download repodata/repomd.xml: All mirrors
were tried

goroutine 1 [running]:
main.main()
        /home/thozza/devel/osbuild-composer/cmd/osbuild-pipeline/main.go:128
+0x1505
exit status 2
```

Modify `osbuild-pipeline` to use the 'name' from repository definitions.
In case it is not defined, fall back to use `repo-<idx>` value to
prevent `Name` being set to an empty string.

Modify methods in `rpmmd` to use the repo `Name`, instead of taking the
index as an argument.

Add 'name' to each repository definition in
`tools/test-case-generators/repos.json`, which missed it.

Errors from DNF now look like this:
```
2021/10/01 10:35:58 Failed to load subscriptions. osbuild-composer will
fail to build images if the configured repositories require them: no
matching key and certificate pair
Errors during downloading metadata for repository 'ha':
  - Status code: 403 for
    https://rpmrepo-storage.bucket.vpce-08d3201af28373567-o10c2v4q.s3.us-east-1.vpce.amazonaws.com/data/rhvpn/el8/sha256-518e7c455fb0bb23e32679c45a45cc269bdd5bd915b560b98bfd699aa4f100e0
(IP: 10.30.18.5)
panic: Could not depsolve: DNF error occured: RepoError: Error occurred
when setting up repo: Failed to download metadata for repo 'ha': Cannot
download repomd.xml: Cannot download repodata/repomd.xml: All mirrors
were tried

goroutine 1 [running]:
main.main()
        /home/thozza/devel/osbuild-composer/cmd/osbuild-pipeline/main.go:128
+0x1505
exit status 2
```

Signed-off-by: Tomas Hozza <thozza@redhat.com>

This pull request includes:


cmd/osbuild-pipeline/main.go
```diff
@@ -101,8 +102,13 @@ func main() {
 
 	repos := make([]rpmmd.RepoConfig, len(composeRequest.Repositories))
 	for i, repo := range composeRequest.Repositories {
+		repo_name := repo.Name
+		// ensure that the name is always set to a sane value
+		if repo_name == "" {
+			repo_name = fmt.Sprintf("repo-%d", i)
```
teg, line 108: Should we ensure the name is unique?

internal/rpmmd/repository.go
```diff
@@ -392,10 +392,9 @@ func NewRPMMD(cacheDir, dnfJsonPath string) RPMMD {
 	}
 }
 
-func (repo RepoConfig) toDNFRepoConfig(rpmmd *rpmmdImpl, i int, arch, releasever string) (dnfRepoConfig, error) {
-	id := strconv.Itoa(i)
+func (repo RepoConfig) toDNFRepoConfig(rpmmd *rpmmdImpl, arch, releasever string) (dnfRepoConfig, error) {
 	dnfRepo := dnfRepoConfig{
-		ID:             id,
+		ID:             repo.Name,
```
teg, line 397: Similarly to above, are we guaranteed that the Name will be unique? (and an appropriate identifier)

cmd/osbuild-pipeline/main.go
msehnout: Isn't it unique because the name contains the index of the repository?

internal/rpmmd/repository.go
msehnout, line 397: Isn't it a bit misleading that the "name" is assigned to "id" even though RPM repos have "name" option as well? Isn't the RPM repository name used in the error output in case it is available?

msehnout (commented):
Seems like a good improvement! I only have one question.

cmd/osbuild-pipeline/main.go
thozza: I think that Tom mainly means that the name provided in the repo definitions as an input may not be unique. I'm OK with adding check to ensure this on input (if DNF does not fail in this case, which I didn't test), but don't think that this should be necessarily taken as a valid case and workarounded in the code to use indexes instead. Alternatively a compromise would be to use the Repo name, but keep the index and combine these two values to make it fool-proof.

internal/rpmmd/repository.go
thozza: Good, point. I'll investigate this option further.

thozza:
Abandoning this PR in favor of #2023.

PR #1864, opened 2021-10-01
More cloud cleaning

Gundersanne:


schutzbot/scheduled_cloud_cleaner.sh
```diff
@@ -50,7 +49,41 @@ echo "There are resources from $TO_DELETE_COUNT test runs to delete."
...
+
+    mapfile -t BLOBS_TO_DELETE_LIST  < <(printf "%s\n" "${OLD_BLOBS_LIST_NAMES[@]}" | sort -u)
+    TO_DELETE_COUNT=${#BLOBS_TO_DELETE_LIST[@]}
+    if [ ${#OLD_BLOBS_LIST_NAMES[@]} == 0 ]; then
+        echo "No blobs delete in $STORAGE_ACCOUNT_NAME."
+    fi
+    echo "Deleting $TO_DELETE_COUNT blobs in $STORAGE_ACCOUNT_NAME."
+
+    for i2 in $(seq 0 $(("$TO_DELETE_COUNT"-1))); do
+        echo "Running cloud-cleaner in Azure for resources with TEST_ID: ${RESOURCE_TO_DELETE_LIST[$i2]}"
```
jrusz, line 83: I think you wanted to list `BLOBS_TO_DELETE_LIST` or append it the to `RESOURCE_TO_DELETE_LIST` since it didn't actually list the names of the images/blobs, just found out the count.

jrusz (commented):
This seems to be able to get the names of all the old images form all the storage accounts but it won't be able to remove them as is. If you call the `cmd/cloud-cleaner` with the name of the image to delete as `testID` it will still only look for it in the `travisciuploads` storage account because that's what we have in our CI variables and it sets it here: https://github.com/osbuild/osbuild-composer/blob/main/cmd/cloud-cleaner/main.go#L123 ... So when it calls the function here: https://github.com/osbuild/osbuild-composer/blob/main/cmd/cloud-cleaner/main.go#L136 it won't find it and delete it...

schutzbot/scheduled_cloud_cleaner.sh
```diff
@@ -50,7 +49,41 @@ echo "There are resources from $TO_DELETE_COUNT test runs to delete."
...
+        # TODO uncomment in actual test run
+        # AZURE_STORAGE_ACCOUNT="$STORAGE_ACCOUNT_NAME" TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner
```
Gundersanne, line 85: @jrusz right but I thought setting the AZURE_STORAGE_ACCOUNT here will overwrite whatever we get from the CI environment; and set it from travisuploadci to one of the scanned ones.

jrusz: Oh I see it now, well that should work then...

Gundersanne: If you think it's an ok approach overall I'll enable this here as a test run.

Gundersanne:
Hm this might be the wrong approach. @ondrejbudai wdyt of setting a [lifecycle policy](https://docs.microsoft.com/en-us/azure/storage/blobs/lifecycle-management-policy-configure) on storage accounts created by the image-builder service? After all we only say the images will be there for 14 days? So if the worker creates a storage account if none is given, have it attach a policy (could make this configurable too).

ondrejbudai:
Sounds good to me! :)

Gundersanne:
We would need to migrate to storage v2 but I don't think that should be a problem? I'll make an issue.

schutzbot/scheduled_cloud_cleaner.sh
```diff
@@ -50,7 +50,38 @@ echo "There are resources from $TO_DELETE_COUNT test runs to delete."
 
 for i in $(seq 0 $(("$TO_DELETE_COUNT"-1))); do
     echo "Running cloud-cleaner in Azure for resources with TEST_ID: ${RESOURCE_TO_DELETE_LIST[$i]}"
-    TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner
+    # TODO uncomment
+    # TEST_ID=${RESOURCE_TO_DELETE_LIST[$i]} /usr/libexec/osbuild-composer-test/cloud-cleaner
```
jrusz, line 54: Can we uncomment this please now? :)

jrusz (changes requested):
I tried to run it: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1645687907 ... Appears to be working as expected, just the warnings are annoying, can you try to get rid of them? Also added one inline comment.

Gundersanne:
@jrusz note that I also had a run here https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1643856368 which removed the bulk of the old data.

jrusz (approved):
Looks good now, latest run is much cleaner https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1645983644 ... I guess it's still complaining about the `az` commands on line 68 and 72 but it's no big deal :). Thanks for this!

PR #1865, opened 2021-10-01
Add gitleaks allowlist

achilleas-k:
Adding a gitleaks allowlist for test passwords and keys.

My fork will occasionally get notified of leaks from scanners which identify the hashed passwords in our test scripts.  With the allowlist, I'm hoping this will stop and then we can hopefully start getting alerts for real leaks only.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,10 @@
+[allowlist]
+  description = "Test keys and passwords that should not be reported as leaks"
+  regexes = [
+      '''\$6\$BhyxFBgrEFh0VrPJ\$MllG8auiU26x2pmzL4\.1maHzPHrA\.4gTdCvlATFp8HJU9UPee4zCS9BVl2HOzKaUYD\/zEm8r\/OF05F2icWB0K''',  # qcow2 test manifest user password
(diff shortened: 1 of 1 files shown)


.gitleaks.toml
```diff
@@ -0,0 +1,10 @@
+[allowlist]
+  description = "Test keys and passwords that should not be reported as leaks"
+  regexes = [
+      '''\$6\$BhyxFBgrEFh0VrPJ\$MllG8auiU26x2pmzL4\.1maHzPHrA\.4gTdCvlATFp8HJU9UPee4zCS9BVl2HOzKaUYD\/zEm8r\/OF05F2icWB0K''',  # qcow2 test manifest user password
+      '''\\\$6\\\$GRmb7S0p8vsYmXzH\\\$o0E020S\.9JQGaHkszoog4ha4AQVs3sk8q0DvLjSMxoxHBKnB2FBXGQ\/OkwZQfW\/76ktHd0NX5nls2LPxPuUdl\.''',  # hashed user password for ostree tests
```
thozza, line 5: why do you have 3 `\` here, while in the first regex you have only one?

achilleas-k: Excellent question: The hashed password in the second one is escaped because it's `echo`ed in bash scripts, whereas the first one appears in json in test manifests and isn't escaped.  So the second one is double-escaped in the regex here.

thozza (approved):

.gitleaks.toml
thozza: Makes sense. Thanks for the explanation!

ondrejbudai (approved):

ondrejbudai:
No need to test this change, admin-merging this.

PR #1869, opened 2021-10-04
Only run necessary CI jobs

diaasami:
This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -103,6 +103,9 @@ Base:
   extends: .terraform
   rules:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
+      changes:
+        - internal/**
+        - test/cases/base_tests.sh
     - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[0-9]-[^ga][\S]+/ && $NIGHTLY == "true"'
   script:
     - schutzbot/deploy.sh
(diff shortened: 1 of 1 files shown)


Gundersanne:
I love this idea, thoughts of what also should trigger tests:

-  `cmd/`  changes
- `schutzbot/` changes should trigger everything
- `vendor/` changes
- `distribution/` changes
- `container/` changes should trigger regression tests where we spin up the container (and in future the api tests)
- `*spec` changes
- `dnf-json` changes should trigger base tests or image-tests or something where we depsolve

diaasami:
After wondering why it's not working, and debugging, it seems this is not possible with the current setup :disappointed: 
> Use only:changes without pipelines for merge requests
> 
> Without pipelines for merge requests, pipelines run on branches or tags that don’t have an explicit association with a merge request. In this case, a previous SHA is used to calculate the diff, which is equivalent to git diff HEAD~. This can result in some unexpected behavior, including:
> 
>     When pushing a new branch or a new tag to GitLab, the policy always evaluates to true.
>     When pushing a new commit, the changed files are calculated by using the previous commit as the base SHA. 

Source: https://docs.gitlab.com/ee/ci/jobs/job_control.html#use-onlychanges-with-scheduled-pipelines

image

diaasami:
It can work but the effect is limited, the bigger the effect the more information about test<->code dependencies that we need to keep/update in .gitlab-ci, I'll close it for now and we can revisit it later if needed.

PR #1873, opened 2021-10-05
tests/ci: enable vmware.sh, cross-distro.sh and koji.sh on rhel-9

### LINKED ISSUE #1634: rhel 9: koji setup doesn't work in koji.sh
Something is wrong in the init script, not sure what.

```
+ timeout 10 bash -c 'until printf '\'''\'' 2>/dev/null >/dev/tcp/org.osbuild.koji.postgres/5432; do sleep 0.1; done'
ERROR:  relation "users" does not exist
LINE 1: insert into users (name, password, status, usertype) values ...
```

https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1477814552

jrusz:
This pull request includes:

I forgot to add the vmware.sh test to .gitlab-ci.yml file when I updated the test and cross-distro.sh should just work as is, so adding both in this commit.

Also fixes #1634 

### DIFF
--- a/tools/provision.sh
+++ b/tools/provision.sh
@@ -18,9 +18,8 @@ elif [[ $ID == rhel || $ID == centos ]] && [[ ${VERSION_ID%.*} == 9 ]]; then
     curl -LO --insecure https://hdn.corp.redhat.com/rhel8-csb/RPMS/noarch/redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm
     sudo dnf install -y ./redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm dnf-plugins-core
     sudo dnf copr enable -y copr.devel.redhat.com/osbuild-team/epel-el9 "rhel-9.dev-$ARCH"
-    # koji is not available yet apparently
     # jmespath required for json_query
-    sudo dnf install -y ansible-core python3-jmespath
+    sudo dnf install -y ansible-core koji python3-jmespath
 
     # json_query filter, used in our ansible playbooks, was moved to the
     # 'community.general' collection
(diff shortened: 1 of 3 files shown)


tools/provision.sh
```diff
@@ -18,9 +18,8 @@ elif [[ $ID == rhel || $ID == centos ]] && [[ ${VERSION_ID%.*} == 9 ]]; then
     curl -LO --insecure https://hdn.corp.redhat.com/rhel8-csb/RPMS/noarch/redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm
     sudo dnf install -y ./redhat-internal-cert-install-0.1-23.el7.csb.noarch.rpm dnf-plugins-core
     sudo dnf copr enable -y copr.devel.redhat.com/osbuild-team/epel-el9 "rhel-9.dev-$ARCH"
-    # koji is not available yet apparently
     # jmespath required for json_query
-    sudo dnf install -y ansible-core python3-jmespath
+    sudo dnf install -y ansible-core koji python3-jmespath
```
atodorov, line 22: shouldn't this be in koji.sh ? Why install it here if it is only needed for one test script only ?

atodorov (commented):
Added 1 comment, otherwise looks good.

tools/provision.sh
jrusz: I suppose we do this to have the test dependencies in one place? AFAIK ansible is only used in ostree and installer tests... I'm ok with moving it to the koji.sh script if you think that makes more sense, but I think it doesn't really matter...

atodorov: Bleeeh. At this point it doesn't matter. The correct course of action will be to adjust all of the dependencies but this is outside of the current scope. Merging as-is.

atodorov (approved):

PR #1881, opened 2021-10-06
ci: split nightly pipeline based on rhel version

jrusz:
This pull request includes:

Just some experiments before implementing nightly pipeline for RHEL-9

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,14 +84,15 @@ Prepare-rhel-internal:
   stage: prepare-rhel-internal
   extends: .terraform
   rules:
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $NIGHTLY == "true"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
   script:
     - schutzbot/prepare-rhel-internal.sh
   artifacts:
     paths:
-      - rhel-8.json
-      - rhel8internal.repo
-      - rhel-8-beta.json
+      - rhel-${RHEL_MAJOR}.json
+      - rhel${RHEL_MAJOR}internal.repo
+      - rhel-${RHEL_MAJOR}-beta.json
       - COMPOSE_ID
   parallel:
     matrix:
@@ -99,6 +100,7 @@ Prepare-rhel-internal:
           # NOTE: 1 runner prepares for all arches b/c subsequent jobs download
           # artifacts from all previous jobs and the last one wins
           - aws/rhel-8.6-nightly-x86_64
+          - aws/rhel-9.0-nightly-x86_64
         INTERNAL_NETWORK: ["true"]
 
 Base:
@@ -106,7 +108,8 @@ Base:
   extends: .terraform
   rules:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[0-9]-[^ga][\S]+/ && $NIGHTLY == "true"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/base_tests.sh
@@ -139,7 +142,8 @@ Regression:
   extends: .terraform
   rules:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[8-9]\.[0-9]-[^ga][\S]+/ && $NIGHTLY == "true"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/regression.sh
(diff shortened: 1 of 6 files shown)


atodorov:
> When I started working on this I expected to always have one set of RHEL-9 nightly runner and RHEL-8 nightly runners... Currently we have both 8.6 and 8.5 nightly and this won't work as expected anymore. Should I just reimplement it to be able to work even with two different RHEL-8/9.$MINOR nightly versions or can we just drop 8.5 nightly testing already? I guess there will always be some overlap, so maybe make it able to work with more nightly versions would be better?... @atodorov @ondrejbudai what do you think?

I think that the checks should be more robust instead of removing test runners. That should always improve the suite, right. However I also want to keep the code short and easy to maintain. So whether or not new changes will be appropriate we can say after we see them.  Do you have any idea in mind ?

jrusz:
I've rebased the PR since nightly testing on RHEL-8.5 was removed. Otherwise the idea stays the same, split the pipelines based on value of `$RHEL_VERSION` which is the same as `$VERSION_ID` so maybe I can rename it to reduce a avoid a potential confusion. I also ran the scheduled pipelines with rpms build in this PR again:

https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/400267700
https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/400278269

.gitlab-ci.yml
```diff
@@ -339,18 +339,21 @@ libvirt:
           - openstack/rhel-9.0-beta-nightly-x86_64
         INTERNAL_NETWORK: ["true"]
 
-RHEL 9 on 8:
```
ondrejbudai, line 334: RHEL 9 on 8 is still needed. RHEL 9 guest images and ec2 images are still built on RHEL 8.

We don't need to test this on nightlies though, this is not officially supported.

ondrejbudai (changes requested):

jrusz:
It should be good like this. Splits the RHEL-8 and RHEL-9 testing and adding a new nightly pipeline is as easy as adding another line of rules, a runner and a pipeline schedule.

ondrejbudai (approved):

jrusz:
I've encountered some problems with `ostree.sh` test failing on RHEL-8.4-ga and after further investigation I've discovered that we were basically skipping the test on centos-8 and fedora-33. Reason being that after sourcing the `tools/define-compose-url.sh` it has `exit 0` if not running on RHEL...
Some examples of the skipped runs:
https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1742811417
https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1742811409

I think that by changing the exit to return it should be fixed, testing it now.

jrusz:
This needs to be rebased once https://github.com/osbuild/osbuild-composer/pull/1949 gets merged because rhel-9-beta runners are no longer working and then it should be ready.

.gitlab-ci.yml
```diff
@@ -84,29 +84,32 @@ Prepare-rhel-internal:
...
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $VERSION_ID == "8.6"'
   script:
     - schutzbot/prepare-rhel-internal.sh
   artifacts:
     paths:
-      - rhel-8.json
-      - rhel8internal.repo
-      - rhel-8-beta.json
+      - rhel-${VERSION_ID}.json
+      - rhel${VERSION_ID}internal.repo
```
atodorov, line 94: The original `rhel8internal.repo` name is also used in all of config.json for the runners:
https://github.com/osbuild/gitlab-ci-terraform/blob/main/aws/rhel-8.6-nightly-x86_64/config.json

to make sure that the override logic afterwards will update this file accordingly and you don't end up pulling RPMs from a different repository than intended, see https://github.com/osbuild/gitlab-ci-terraform/pull/6 and deploy.sh, line 112.

The change in name breaks this behavior and you can potentially pull RPMs from a different repo/compose, which are a different version and not what you wanted to test. The risk is slim but not zero.

tools/define-compose-url.sh
```diff
@@ -2,19 +2,20 @@
 set -euo pipefail
 source /etc/os-release
 
-# This isn't needed when not running on RHEL
-if [[ $ID != rhel ]]; then
-  exit 0
+# This isn't needed when not running on RHEL or it's subscribed
+if [[ $ID != rhel ]] || sudo subscription-manager status; then
+  return 0
```
atodorov, line 7: This should be `exit 0` instead of `return 0`.

tools/define-compose-url.sh
```diff
@@ -2,19 +2,20 @@
...
-  COMPOSE_ID=$(curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-finished-RHEL-8.6/COMPOSE_ID)
+  COMPOSE_ID=$(curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-finished-RHEL-"${VERSION_ID}"/COMPOSE_ID)
 
   # default to a nightly tree but respect values passed from ENV so we can test rel-eng composes as well
   COMPOSE_URL="${COMPOSE_URL:-http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/$COMPOSE_ID}"
 
 elif [[ $ID == rhel && ${VERSION_ID%.*} == 9 ]]; then
-  COMPOSE_ID=$(curl -L http://download.devel.redhat.com/rhel-9/nightly/RHEL-9-Beta/latest-RHEL-9.0/COMPOSE_ID)
+  # NOTE: remove Beta from the url once Beta is out
+  COMPOSE_ID=$(curl -L http://download.devel.redhat.com/rhel-9/nightly/RHEL-9-Beta/latest-RHEL-"${VERSION_ID}"/COMPOSE_ID)
```
atodorov, line 18: I think Beta is already out, no ?

atodorov (changes requested):
Overall looks good, minor change requested. 

However I am concerned about the changes in the naming scheme of repo files. It does look cleaner this way but if we want to keep the current changes you need to patch all runner definitions as well. Otherwise the naming should include only the major version part of VERSION_ID.

tools/define-compose-url.sh
jrusz: Right, this is from before beta was not yet out, I'll update it, thanks!

tools/define-compose-url.sh
jrusz: `ostree.sh` test is sourcing this file and when this has `exit 0` it will not run the test at all on fedora and centos so I decided to just `return 0` so the test will run as expected... I'm open to suggestions though if you think this is not a good way to resolve this...

tools/define-compose-url.sh
```diff
@@ -2,19 +2,20 @@
 set -euo pipefail
 source /etc/os-release
 
-# This isn't needed when not running on RHEL
-if [[ $ID != rhel ]]; then
-  exit 0
+# This isn't needed when not running on RHEL or it's subscribed
+if [[ $ID != rhel ]] || sudo subscription-manager status; then
```
atodorov, line 6: How about the subscribed part? I'm wondering if we could end up in a situation where a subscribed test runner is trying to execute this script. Probably not, but in this case why do you need the check? 

If that's only because of ostree.sh IMO it makes sense to move the subscription check logic in ostree.sh b/c this file is used in other places as well.

.gitlab-ci.yml
jrusz: Thanks, I didn't notice that it was being used elsewhere. I guess it worked because during `prepare-rhel-internal.sh` the created repo gets `priority=1` so it would always be preferred over the others. I changed it to use just the major version in the filename to be sure though. 

This should be fine until we get to a state where we'll be testing two $major.$minor nightlies (eg. 8.6 and 8.7) but we can worry about that later, I'd prefer to have this PR resolved first.

tools/define-compose-url.sh
jrusz: Updated, we just don't have a `latest-finished` symlink for RHEL-9 yet though.

tools/define-compose-url.sh
jrusz: Yeah since it's only used in `ostree.sh` I moved the logic there, I'm not sure about whether to have the `return 0` or `exit 0` there... In case someone tries to sources it somewhere in the future his script could exit unexpectedly just like it was until now when `ostree.sh` was skipped on fedora and centos.

tools/define-compose-url.sh
atodorov: URL looks good but the comment above can be removed b/c it is no longer valid.

test/cases/ostree.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
-source /usr/libexec/osbuild-composer-test/define-compose-url.sh
+# Get compsoe url if it's running on unsubscried RHEL
```
atodorov, line 4: Small typo

tools/define-compose-url.sh
atodorov: `return 0` is in fact functionally correct. If that is not causing any other issues keep it. Strictly speaking the calling environment should be examining the exit code b/c in some situations COMPOSE_URL isn't going to be defined. However that's only used on RHEL test runners so I'm not expecting any errors.

atodorov (dismissed):
I made small comments. We can merge this PR once the tests report PASS and follow up on typos and comment removal in a separate PR though.

jrusz:
I had an unbound variable in `ostree.sh` so I had to fix that and so I also fixed the the comments and found out I forgot to remove one more `Beta` from a URL. I hope it's going to be right this time.

atodorov (approved):

PR #1896, opened 2021-10-08
composer: Don't dump sensitive fields from config

Gundersanne:


Gundersanne:
vmware tests are still failing; building in vsphere works but not the running. If only the vmware ones are failing I think we should admin-merge.

PR #1906, opened 2021-10-12
CentOS 8 update - Alias on RHEL 8.6

achilleas-k:
**CentOS 8** `centos-8` is now an alias to RHEL 8.6 `rhel-86` with appropriate changes.

I went through and found all the strings that are distro-specific and added them to the `distribution` object.  This saves us from having conditionals everywhere when adding values to stages that need different values.  It also helps to see all the things that make the distributions different in one place.

Differences between the two distros:
- `s390x` is not enabled on CentOS
- `ec2` image types are not enabled on CentOS (no `rhui-client`)
- RHSM configuration skipped for `ami` on CentOS

All other differences are handled by the aforementioned distro-wide strings.

Also:
- Regenerated all test manifests
- Enabled OSTree tests on CentOS 8

TODO:
- [x] Make `greenboot-failing-unit` available for ostree tests
- [x] Add `centos-8` to checks in newly enabled test scripts

----

This pull request includes:

### DIFF
--- a/docs/news/unreleased/centos-8-update.md
+++ b/docs/news/unreleased/centos-8-update.md
@@ -1,0 +1,7 @@
+# Update image definitions for CentOS Stream 8
+
+Image definitions for the CentOS Stream 8 distribution have been
+updated. CentOS Stream 8 (internally `centos-8`) is now an alias to the current
+RHEL 8.6 distribution, with appropriate modifications.
+
+All Edge image types are now supported in CentOS Stream 8.


achilleas-k:
The ostree tests are expected to fail now because we need `greenboot-failing-unit` to test ostree rollback and the package isn't available in public repositories.  Thanks @henrywang for the heads-up!

teg (approved):

gicmo (approved):
👍

henrywang:
@achilleas-k `--cache-only` works here.

ondrejbudai (dismissed):
Still amazing. :)

achilleas-k:
> I just now noticed how many times we (I?) restarted the simplified installer test on cs8. Probably something actually wrong with it and not just CI slowness. Checking.

Found the issue.  The ISO was using an incorrect isolabel and volume ID.  Should be working now.

PR #1907, opened 2021-10-12
customizations: add support for fs sizes in MB and GB

msehnout:
It is not convenient to specify filesystem sizes in bytes every time.
Add support for parsing string with mMgG specified.

This approach has a downside, that now the filesystem size is of type
String and therefore all of the references must be strings which changes
the original behavior.

Alternative approach is to define a custom type and enable parsing of
both number and string.

TODO: input validation in Weldr API

This pull request includes:

### DIFF
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -3,6 +3,8 @@ package blueprint
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
 )
 
 type Customizations struct {
@@ -73,7 +75,7 @@ type ServicesCustomization struct {
 
 type FilesystemCustomization struct {
 	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
-	MinSize    uint64 `json:"minsize,omitempty" toml:"size,omitempty"`
+	MinSize    string `json:"minsize,omitempty" toml:"size,omitempty"`
 }
 
 type CustomizationError struct {
@@ -260,7 +262,7 @@ func (c *Customizations) GetFilesystemsMinSize() uint64 {
 	}
 	var agg uint64
 	for _, m := range c.Filesystem {
-		agg += m.MinSize
+		agg += m.GetMinSize()
 	}
 	// This ensures that file system customization `size` is a multiple of
 	// sector size (512)
@@ -276,4 +278,14 @@ func (c *Customizations) GetInstallationDevice() string {
 		return defaultDevice
 	}
 	return c.InstallationDevice
+}
+
+func (fs *FilesystemCustomization) GetMinSize() uint64 {
+	min_size := common.FsSizeToUint64(fs.MinSize)
+	if min_size != nil {
+		return *min_size
+	}
+
+	// The string must be validated in the Weldr API
+	panic("Invalid filesystem size")
 }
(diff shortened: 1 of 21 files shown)


internal/common/helpers.go
```diff
@@ -36,3 +38,31 @@ func IsStringInSortedSlice(slice []string, s string) bool {
...
+	return_size := uint64(number)
+
+	mega_byte := regexp.MustCompile(`[[:digit:]]+\s*(m|M)(b|B)?`)
+	giga_byte := regexp.MustCompile(`[[:digit:]]+\s*(g|G)(b|B)?`)
+
+	if plain_number.MatchString(size) {
+		return &return_size
+	}
+	if mega_byte.MatchString(size) {
+		return_size *= 1000000
```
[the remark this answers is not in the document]
msehnout: Good point, it shouldn't if we follow this definition: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

internal/blueprint/customizations.go
```diff
@@ -277,3 +279,13 @@ func (c *Customizations) GetInstallationDevice() string {
 	}
 	return c.InstallationDevice
 }
+
+func (fs *FilesystemCustomization) GetMinSize() uint64 {
```
kingsleyzissou, line 283: Maybe we can add a test for this

internal/common/helpers.go
```diff
@@ -36,3 +38,31 @@ func IsStringInSortedSlice(slice []string, s string) bool {
...
+	// Read the number in the string
+	plain_number := regexp.MustCompile(`[[:digit:]]+`)
+	numbers_as_str := plain_number.FindAllString(size, 1)
+	number, err := strconv.ParseInt(numbers_as_str[0], 10, 64)
+	if err != nil {
+		return nil
+	}
+	return_size := uint64(number)
+
+	mega_byte := regexp.MustCompile(`[[:digit:]]+\s*(m|M)(b|B)?`)
```
ondrejbudai, line 52: I suggest closely following the metric prefixes: https://en.wikipedia.org/wiki/Metric_prefix

Thus accepting only `M` as mega (`m` is milli), and `G` as giga (`g` doesn't exist).

As an addition, we can also accept binary prefixes: https://en.wikipedia.org/wiki/Binary_prefix

I would also vote against accepting `b`, so we don't confuse users with bits vs. bytes.

msehnout: This was in the request I got, but I might have incorrectly understand it.

Regardless, I agree that strictly following the prefixes would be the best solution, even though it might differ from some other tools (I don't exactly know what syntax kickstarts allow).

ondrejbudai: I found something like this in the docs ( https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/sect-kickstart-syntax ):

> --size= - The size of the logical volume in MiB

I'm not a fan of this tbh.

btw e.g. `sfdisk` uses binary prefixes correctly (from `sfdisk(1)`): 

> The number may be followed by the multiplicative suffixes (KiB, MiB, GiB, TiB, PiB, EiB, ZiB and YiB), then it’s interpreted as size in bytes

teg: I think we can deviate from the request and do what we think makes more sense, so following the metric definitions sounds good to me.

PR #1953, opened 2021-10-28
cloudapi: add input validation

chloenayon:
Add logic to validate incoming requests, using openapi3. 
Currently lacks testing.

This pull request includes:


ochosi:
Hey! I removed myself from the reviewers because I don't know enough about how openapi3filter should be integrated into composer tbh.
@Gundersanne Feel free to re-assign to someone else :)

Gundersanne:
Hm in general I'm not sure if we want this for composer. All of composer's clients are highly trusted and whitelisted, it's not really userfacing.

And I know at least of one instance where this will cause an issue. ComposeRequest also implements `      - $ref: '#/components/schemas/ObjectReference'` which is a mistake I made since request bodies sohldn't carry that, but currently you can just leave those fields blank and it'll work fine. So that would have to be removed here too.

ondrejbudai:
I think that more data validation never hurts, I'm all for adding the filter. If the schema is wrong in some cases, let's fix it.

Gundersanne:
Oki, opened a PR to mitigate the spec issue. After that it should be fine.

- [ ] https://github.com/osbuild/osbuild-composer/pull/2042

Gundersanne:
Let's rebase this?

internal/cloudapi/v2/v2.go
```diff
@@ -92,6 +112,30 @@ func (s *Server) IncRequests(next echo.HandlerFunc) echo.HandlerFunc {
 	}
 }
 
+func (s *Server) ValidateRequest(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		request := c.Request()
+
+		route, params, err := s.router.FindRoute(request)
+		if err != nil {
+			return err
```
Gundersanne, line 121: These errors shold be defined in https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/errors.go.

internal/cloudapi/v2/v2.go
```diff
@@ -92,6 +112,30 @@ func (s *Server) IncRequests(next echo.HandlerFunc) echo.HandlerFunc {
...
+
+		input := &openapi3filter.RequestValidationInput{
+			Request:    request,
+			PathParams: params,
+			Route:      route,
+		}
+
+		ctx := request.Context()
+		if err := openapi3filter.ValidateRequest(ctx, input); err != nil {
+			return err
```
Gundersanne, line 132: Same here, but I wonder if it's possible we can get 'why' it's a 400 from the error here. The advantage of checking stuff in the handlers themselves is that we can return different self-defined error codes. If we can get a single useful string maybe we could attach it to the return error message.

internal/cloudapi/v2/v2.go
chloenayon: ErrorFailedToOpenAPISpec makes sense for GetSwagger, but what would be appropriate for validate/new router errors?

chloenayon:
I've updated to openapi v0.88.0 due to some breaking changes (I should probably add this to the commit message) but it turns out deepmap/oapi-codegen is still using v0.80.0 - hence the undefined errors

Gundersanne:
So the errors here are actually the spec itself being wrong. The objects ComposeLogs and ComposeManifests need `items: ...` defined. Unsure what those should be however.

ondrejbudai:
It would be nice to merge the following PR firstly, otherwise we will surely run into some issues with our integration tests when validation is turned on: https://github.com/osbuild/osbuild-composer/pull/2490

ondrejbudai:
#2637 should fix the mockbuild for Fedoras! :partying_face:

ondrejbudai:
I rebased, this should hopefully fix mockbuild on Fedora.

internal/cloudapi/v2/v2_test.go
```diff
@@ -159,11 +159,11 @@ func TestCompose(t *testing.T) {
...
 	{
-		"href": "/api/image-builder-composer/v2/errors/4",
-		"id": "4",
+		"href": "/api/image-builder-composer/v2/errors/30",
+		"id": "30",
 		"kind": "Error",
-		"code": "IMAGE-BUILDER-COMPOSER-4",
-		"reason": "Unsupported distribution"
+		"code": "IMAGE-BUILDER-COMPOSER-30",
+		"reason": "Request could not be validated"
```
ondrejbudai, line 166: This is suboptimal as it makes the error more generic. I would still prefer merging this and dealing with it in a follow-up (we should probably somehow propagate the whole validation error from kin-openapi to the response).

ondrejbudai:
@Gundersanne Can you review? I think that both me and @chloenayon are finnaly happy with it. :)

Note that this might break image-builder and koji-osbuild because it makes the validation stricter (we have to do some API test adjustments - not only in this PR). If something like this happens, we should fix that in API clients because they are probably not following the API spec correctly.

internal/cloudapi/v2/v2_test.go
Gundersanne: Agreed, how the filter generates the error: https://github.com/getkin/kin-openapi/blob/master/openapi3filter/errors.go#L12-L39.

Gundersanne (approved):
A few nitpicks regarding format. I don't think format is actually checked and is purely informative.

PR #1959, opened 2021-10-29
[WIP] Regression test fail

jabia99:
Find root cause for regression test failing on 8.6 runners


ondrejbudai:
I think this is now fixed. Feel free to reopen if I'm wrong.

PR #1969, opened 2021-11-02
tests: skip Openstack image boot testing on aarch64

jrusz:
We don't have aarch64 machines in Openstack so fallback to qemu when running on aarch64.

This pull request includes:

### DIFF
--- a/cmd/osbuild-image-tests/main_test.go
+++ b/cmd/osbuild-image-tests/main_test.go
@@ -389,9 +389,13 @@ func testBootUsingAzure(t *testing.T, imagePath string) {
 
 func testBootUsingOpenStack(t *testing.T, imagePath string) {
 	creds, err := openstack.AuthOptionsFromEnv()
+	currentArch := common.CurrentArch()
 
-	// if no credentials are given, fall back to qemu
-	if (creds == gophercloud.AuthOptions{}) {
+	// skip on aarch64 because we don't have aarch64 openstack or kvm machines
+	if currentArch == "aarch64" {
+		t.Skip("Openstack boot test is skipped on aarch64.")
+		// if no credentials are given, fall back to qemu
+	} else if (creds == gophercloud.AuthOptions{}) {
 		log.Print("No OpenStack credentials given, falling back to booting using qemu")
 		testBootUsingQemu(t, imagePath)
 		return


jrusz:
I changed this to just skipping the booting part of aarch64 Openstack images because we don't have aarch64 kvm or Openstack machines.

cmd/osbuild-image-tests/main_test.go
```diff
@@ -389,12 +389,16 @@ func testBootUsingAzure(t *testing.T, imagePath string) {
...
 	creds, err := openstack.AuthOptionsFromEnv()
+	currentArch := common.CurrentArch()
 
 	// if no credentials are given, fall back to qemu
 	if (creds == gophercloud.AuthOptions{}) {
 		log.Print("No OpenStack credentials given, falling back to booting using qemu")
 		testBootUsingQemu(t, imagePath)
 		return
+		// skip on aarch64 because we don't have aarch64 openstack or kvm machines
+	} else if currentArch == "aarch64" {
```
ondrejbudai, line 400: I would move this check above the previous check. If there are no OpenStack creds, qemu boot will still be tried.

jrusz: Right, that wouldn't work locally, I swapped it, thanks!

cmd/osbuild-image-tests/main_test.go
```diff
@@ -389,9 +389,13 @@ func testBootUsingAzure(t *testing.T, imagePath string) {
 
 func testBootUsingOpenStack(t *testing.T, imagePath string) {
 	creds, err := openstack.AuthOptionsFromEnv()
+	currentArch := common.CurrentArch()
 
-	// if no credentials are given, fall back to qemu
-	if (creds == gophercloud.AuthOptions{}) {
+	// skip on aarch64 because we don't have aarch64 openstack or kvm machines
+	if currentArch == "aarch64" {
+		t.Skip("Openstack boot test is skipped on aarch64.")
```
atodorov, line 396: Wouldn't `t.Skip()` terminate the existing test and not allow the else branch to execute? I have the feeling it will (haven't tested it though).  If you want to preserve the fallback functionality then a log message instead of Skip will work.

@ondrejbudai ^^^ do you have experience with how .Skip works?

atodorov (changes requested):

cmd/osbuild-image-tests/main_test.go
ondrejbudai: I'm confused. Which `else` branch?  If this branch is executed, the else branch is never executed.

jrusz: Well terminating the boot test when it's running on `aarch64` is what I want to achieve here. So it is expected that the `else` branch does not execute on `aarch64`.

atodorov: > If this branch is executed, the else branch is never executed.

Correct.

and if you call t.Skip() then the code on lines 397-400 won't be executed. The original description of this issue says "fallback to qemu when running on aarch64." which will not happen IMO. As written this test will be skipped entirely if on aarch64 without attempting to fallback to qemu regardless if we have configured cloud credentials or not. 

I am under the impression that we want the fallback to still happen.

jrusz: Now that I think about it I'm surprised it worked before I swapped the conditions https://gitlab.com/osbuild/ci/osbuild/-/jobs/1783867425#L1917 since the credentials are present in the CI and it should have tried to boot locally... I'll run the tests again once the rpms are built for this commit.

jrusz: > The original description of this issue says "fallback to qemu when running on aarch64." which will not happen IMO. As written this test will be skipped entirely if on aarch64 without attempting to fallback to qemu regardless if we have configured cloud credentials or not.
> 
> I am under the impression that we want the fallback to still happen.

I changed the name of the PR, commit and added a comment that I decided to change it to skipping on aarch64 because we have neither Openstack aarch64 machines nor aarch64 machines with kvm support. I tried to make it work without kvm but ran into several problems with qemu and after some time decided that it's not worth it and I'll just skip the boot test since everything else works as is on aarch64.

ondrejbudai (approved):

jrusz:
Here's the osbuild test run https://gitlab.com/osbuild/ci/osbuild/-/pipelines/411514193 ... Currently only rhel-9 is failing but that's another issue which is tracked already.

atodorov (approved):

PR #1980, opened 2021-11-08
composer: Add worker openapi spec endpoint to auth excludes

Gundersanne:
Also remove metrics from the excludes for the composer api. This
endpoint would not be called via the gateway so it's not relevant here.


ondrejbudai (dismissed):

teg (approved):
Thanks!

teg (approved):

ondrejbudai (approved):

PR #1985, opened 2021-11-09
spec: bump osbuild version to 41

msehnout:
This will be needed for new stages which are available only in osbuild
41.

This pull request includes:

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -326,8 +326,8 @@ The core osbuild-composer binary. This is suitable both for spawning in containe
 Summary:    The worker for osbuild-composer
 Requires:   systemd
 Requires:   qemu-img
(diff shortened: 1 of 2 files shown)


osbuild-composer.spec
```diff
@@ -326,7 +326,7 @@ The core osbuild-composer binary. This is suitable both for spawning in containe
 Summary:    The worker for osbuild-composer
 Requires:   systemd
 Requires:   qemu-img
-Requires:   osbuild >= 37
+Requires:   osbuild >= 41
 Requires:   osbuild-ostree >= 37
```
ondrejbudai, line 330: I think I would prefer these two to stay in sync. Maybe let's take this opportunity and factor them out?

ondrejbudai (changes requested):

PR #1984, opened 2021-11-09
osbuild2: new stage authconfig 

msehnout:
This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #1983, opened 2021-11-09
osbuild2: support 'install' command in the modprobe stage and rework data validation

thozza:
- osbuild2: support 'install' command in the modprobe stage
  - Related to osbuild/osbuild#867
- osbuild2/modprobe: extend and rework stage options validation
  -  Related to issue #1785.

This pull request includes:

- [x] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #1986, opened 2021-11-09
osbuild2: new stage yum config

msehnout:

This pull request includes:


ondrejbudai (approved):
Needs a rebase though.

internal/osbuild2/yum_config_stage.go
```diff
@@ -0,0 +1,45 @@
...
+type YumConfigConfig struct {
+	HttpCaching string `json:"http_caching,omitempty"`
+}
+
+type YumConfigPlugins struct {
+	Langpacks YumConfigPluginsLangpacks `json:"langpacks,omitempty"`
+}
+
+type YumConfigPluginsLangpacks struct {
+	Locales []string `json:"locales"`
```
thozza, line 14: `locales` must not be empty (https://github.com/osbuild/osbuild/blob/40bc2161274abac895ce74a323fada30674b3d2f/stages/org.osbuild.yum.config#L45). Feel free to disregard if you considered it, but I wonder whether it could / should be tested in `YumConfigStageOptions`'s `validate()` method.

thozza (commented):
In general this looks very good. I added one suggestion. In addition I wonder whether you considered unit testing the options validation?

internal/osbuild2/yum_config_stage.go
msehnout: Sure, do you think it is ok to just check for length > 0 or would you like to see a regex matching those values as well (in such case I'd change the schema as well)?

msehnout:
@thozza I've updated the PR according to the suggestions, can you please take a look again?

internal/osbuild2/yum_config_stage.go
```diff
@@ -0,0 +1,62 @@
...
+	if o.Config != nil && o.Config.HttpCaching != nil {
+		valid := false
+		allowed_http_caching_values := []string{"all", "packages", "lazy:packages", "none"}
+		for _, v := range allowed_http_caching_values {
+			if v == *o.Config.HttpCaching {
+				valid = true
+			}
+		}
+		if !valid {
+			err = fmt.Errorf("yum config parameter http_caching does not allow %s as a value", *o.Config.HttpCaching)
```
thozza, line 38: Nitpick: you could just return the `fmt.Errorf()` here as other stages tend to do:

- https://github.com/osbuild/osbuild-composer/blob/af9cca1b508f505c9c748993a92e777f29397928/internal/osbuild2/lvm2_create_stage.go#L27
- https://github.com/osbuild/osbuild-composer/pull/1983/files#diff-ecc3ff1d8a00b4f3d4a42406ff0b92e8c73bc66f6f36e26a274eb95493d02c2bR25 (still a PR)

internal/osbuild2/yum_config_stage.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+
+	if o.Plugins != nil && o.Plugins.Langpacks != nil && len(o.Plugins.Langpacks.Locales) < 1 {
+		append := ""
+		if err != nil {
+			append = fmt.Sprintf(", %s", err.Error())
+		}
+		err = fmt.Errorf("locales must contain at least one element, %s", append)
```
thozza, line 47: You could return the error right away here as well.

internal/osbuild2/yum_config_stage.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	return err
```
thozza, line 50: And here you could return `nil`. I find it easier to understand what value is returned, than when a variable is used. I would argue that the important part here is that the validation failed, than to capture all the issues, especially when the code panics when it does. But I understand that capturing all errors adds some value. I don't have a strong opinion about this, but think that we should be consistent across the codebase.

internal/osbuild2/yum_config_stage_test.go
```diff
@@ -0,0 +1,113 @@
...
+					HttpCaching: common.StringToPtr("all"),
+				},
+			},
+			true,
+		},
+	}
+	for _, test := range tests {
+		err := test.options.validate()
+		if test.valid {
+			require.NoError(t, err)
```
thozza, line 108: You could also test that calling the `NewYumConfigStage()` with the stage options panics or not, but not a blocker... (I used this approach in my PR - https://github.com/osbuild/osbuild-composer/pull/1983/files#diff-32275eea0486196031af6472d9e23733d1f8d2eaa44153efa84a06d4c39bf899R71)

thozza (approved):
The code looks good. I added some nitpick / consistency comments, but I don't see them as blockers.

thozza:
https://github.com/osbuild/osbuild-composer/pull/1984#issuecomment-967048573 applies here as well.

internal/osbuild2/yum_config_stage.go
msehnout: Hmm, I guess I'm fine either way, but this seemed like a better solution to me because you can see right away all the issues and you can possibly save yourself from doing 2 updates instead of a single one.

msehnout: Please have look in the new PR. All requested changes are there.

msehnout:
Thanks for the reviews, I've merged my PRs into a single one. Please have a look at #2006.

PR #1989, opened 2021-11-10
Logging improvements

diaasami:
This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -87,7 +87,7 @@ func WatchJob(ctx context.Context, job worker.Job) {
 // Requests and runs 1 job of specified type(s)
 // Returning an error here will result in the worker backing off for a while and retrying
 func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls map[string]JobImplementation) error {
-	logrus.Info("Waiting for a new job...")
+	logrus.Debug("Waiting for a new job...")
 	job, err := client.RequestJob(acceptedJobTypes, common.CurrentArch())
 	if err == worker.ErrClientRequestJobTimeout {
 		logrus.Debugf("Requesting job timed out: %v", err)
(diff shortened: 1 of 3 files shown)


teg (dismissed):

cmd/osbuild-worker/main.go
```diff
@@ -90,7 +90,7 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 	logrus.Info("Waiting for a new job...")
 	job, err := client.RequestJob(acceptedJobTypes, common.CurrentArch())
 	if err == worker.ErrClientRequestJobTimeout {
-		logrus.Debugf("Requesting job timed out: %v", err)
+		logrus.Infof("Requesting job timed out: %v", err)
```
Gundersanne, line 93: This is actually expected behaviour, but it's ok to have it here.

Gundersanne (dismissed):
lgtm!

cmd/osbuild-worker/main.go
ondrejbudai: Hmm, isn't this going to spam journal? If it is, I suggest reverting this change.

ondrejbudai (changes requested):
See my comment inline. I think that we don't print debug logs by default, right?

cmd/osbuild-worker/main.go
Gundersanne: every 20s yea, and you're right we don't have great filtering mechanisms yet.

ondrejbudai: @diaasami WHile you are on it, can you also change https://github.com/osbuild/osbuild-composer/blob/ab3712fd928d07c9f40651e253262494ae19d843/cmd/osbuild-worker/main.go#L90 to debug? This message is like 99,9% of our logs and it actually isn't that interesting.

diaasami:
> See my comment inline. I think that we don't print debug logs by default, right?

No we are not.

Gundersanne (approved):

Gundersanne (approved):

ondrejbudai (approved):
Thanks a lot!

PR #1991, opened 2021-11-10
osbuild2: new stage pwquality conf

msehnout:

This pull request includes:

### DIFF
--- a/internal/osbuild2/pwquality_conf_stage.go
+++ b/internal/osbuild2/pwquality_conf_stage.go
@@ -1,0 +1,23 @@
+package osbuild2
+
+type PwqualityConfConfig struct {
+	Minlen   *int `json:"minlen,omitempty"`
+	Dcredit  *int `json:"dcredit,omitempty"`
+	Ucredit  *int `json:"ucredit,omitempty"`
+	Lcredit  *int `json:"lcredit,omitempty"`
+	Ocredit  *int `json:"ocredit,omitempty"`
+	Minclass *int `json:"minclass,omitempty"`
+}
+
+type PwqualityConfStageOptions struct {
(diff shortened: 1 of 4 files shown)


internal/osbuild2/pwquality_conf_stage.go
```diff
@@ -0,0 +1,23 @@
+package osbuild2
+
+type PwqualityConfConfig struct {
+	Minlen   int `json:"minlen"`
+	Dcredit  int `json:"dcredit"`
+	Ucredit  int `json:"ucredit"`
+	Lcredit  int `json:"lcredit"`
+	Ocredit  int `json:"ocredit"`
+	Minclass int `json:"minclass"`
```
thozza, line 9: All of these values are optional should be marked as 'omitempty` and most probably also defined as pointer values, because the "zero" value for the type would otherwise end up in the stage options and would be defined in the pwquality configuration.

thozza (changes requested):
In general the code looks good, although there is one part that I commented on that should be IMO changed. In addition you should add case for the new stage into https://github.com/osbuild/osbuild-composer/blob/main/internal/osbuild2/stage.go#L36 and also to the relevant unit tests file.

msehnout:
@thozza I've updated the PR according to the suggestions, can you please take a look again?

thozza (approved):
LGTM 👍

ondrejbudai:
Needs a rebase. :(

thozza:
> Needs a rebase. :(

Yeah :( https://github.com/osbuild/osbuild-composer/pull/1984#issuecomment-967048573 applies here as well...

msehnout:
Thanks for the reviews, I've merged my PRs into a single one. Please have a look at #2006.

PR #1992, opened 2021-11-10
osbuild2: new stage sshd config

msehnout:

This pull request includes:

### DIFF
--- a/internal/osbuild2/sshd_config_stage.go
+++ b/internal/osbuild2/sshd_config_stage.go
@@ -1,0 +1,20 @@
+package osbuild2
+
+type SshdConfigConfig struct {
+	PasswordAuthentication          *bool `json:"PasswordAuthentication,omitempty"`
+	ChallengeResponseAuthentication *bool `json:"ChallengeResponseAuthentication,omitempty"`
+	ClientAliveInterval             *int  `json:"ClientAliveInterval,omitempty"`
+}
+
+type SshdConfigStageOptions struct {
+	Config SshdConfigConfig `json:"config"`
+}
+
+func (SshdConfigStageOptions) isStageOptions() {}
+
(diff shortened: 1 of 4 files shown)


internal/osbuild2/sshd_config_stage.go
```diff
@@ -0,0 +1,20 @@
+package osbuild2
+
+type SshdConfigConfig struct {
+	PasswordAuthentication          bool `json:"PasswordAuthentication"`
+	ChallengeResponseAuthentication bool `json:"ChallengeResponseAuthentication"`
+	ClientAliveInterval             int  `json:"ClientAliveInterval"`
```
thozza, line 6: All of these values should be marked as 'omitempty` and most probably also defined as pointer values, because the "zero" value for the type would otherwise end up in the stage options and would be defined in the SSHD configuration.

thozza (changes requested):
In general the code looks good, although there is one part that I commented on that should be IMO changed. In addition you should add case for the new stage into https://github.com/osbuild/osbuild-composer/blob/main/internal/osbuild2/stage.go#L36 and also  to the relevant unit tests file.

internal/osbuild2/sshd_config_stage.go
msehnout: No problem, what could go wrong if all the variables are pointers :-)

msehnout:
> In general the code looks good, although there is one part that I commented on that should be IMO changed. In addition you should add case for the new stage into https://github.com/osbuild/osbuild-composer/blob/main/internal/osbuild2/stage.go#L36 and also to the relevant unit tests file.

Thanks for the review! I've fixed the PR, can you take a look again?

thozza (approved):
Thanks for the fixes, it all looks perfect! 👍

PR #1999, opened 2021-11-11
worker: Introduce manifest-id-only job

Gundersanne:


Gundersanne:
@ondrejbudai https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1774028584 seems to work ; the "data" type just needs to be swapped out for something obscure and needs to be dequeued by id.

internal/worker/json.go
```diff
@@ -94,6 +94,14 @@ type DepsolveJobResult struct {
 	ErrorType    ErrorType                      `json:"error_type"`
 }
 
+type DataJob struct {
```
ondrejbudai, line 97: I would honestly go rather with specialized jobs rather than one generic one. For this case, I would just name it ManifestJob. It wouldn't have any arg fields, and just have `Manifest` as the only field of the result.

Gundersanne: Agreed, would you make the type only dequeue-able by ID? As in the worker server would throw an error when you try to dequeue this job by type?

ondrejbudai: Makes sense!

Gundersanne:
I split out the cloudapi part for a later revision. That way we won't break the stage deployment if we merge this, and can update the workers before merging the followup https://github.com/Gundersanne/osbuild-composer/tree/v2-manifestjob

Gundersanne:
ugh added the cloudapi stuff back in :facepalm:

ondrejbudai (approved):

ondrejbudai (approved):
Thanks a lot!

Gundersanne:
> Thanks a lot!

Thank you for the tricky queue stuff!

PR #2001, opened 2021-11-12
test: Use YAML as Ansible output format

henrywang:
To make Ansible output human-readable, use YAML as output format.


achilleas-k (approved):
This is great.  We were looking for a good replacement for the old DEBUG output format that's now unavailable.  Thank you!

PR #2019, opened 2021-11-16
RHEL-9.0: unify the default partitioning scheme used by all non-EDGE images

thozza:
Modify the default partitioning scheme to use:
- /boot/efi - 200 MB
- /boot - 500 MB

and make the default partitioning scheme to be used by all EC2 images. As a consequence get rid of the EC2-specific partitioning scheme.

- RHEL-9.0: add separate `/boot` partition to default partition table
- RHEL-9.0: increase the size of `/boot/efi` partition to 200 MB
- RHEL-9.0: add / fix BIOS boot partition size comment
- RHEL-9.0: use the default partitioning scheme for all EC2 images
- Regenerate image test cases

Related to https://bugzilla.redhat.com/show_bug.cgi?id=2022805

This pull request includes:


achilleas-k (approved):
I love this!!
1. Thanks for fixing the stuff I blindly threw in the distro definition.
2. Having a unified partitioning scheme will make things much simpler.

gicmo:
Thanks for this! I will have a look on Monday.

gicmo:
I think that schema could also be good for Edge. There we currently have:
```
ESP/EFI: 260096 sectors // 127 MB (would become 200 MB)
/boot:   786432 sectors // 384 MB (would become 500 MB)
```

So we would enlarge both partitions and I think that should be fine! @nullr0ute do you agree?

One thing I am wondering, esp. for Edge, if 200 MB for the ESP/EFI is big enough for firmware updates. Having a separate `/boot` I think it should be, but @hughsie will know best.

hughsie:
> if 200 MB for the ESP/EFI is big enough for firmware updates

That depends on if:

 * You're going to be doing UEFI UpdateCapsule on edge hardware
 * What the size of the capsule updates are (e.g. 16MB, 32MB)
 * How many kernels are installed, on the assumption you're sharing /boot and /boot/efi on the same physical partition

But all things considered, I think 200MB is probably fine as you're not going to have N distros installed, along with Windows 10.

gicmo:
> > if 200 MB for the ESP/EFI is big enough for firmware updates
> 
> That depends on if:
> 
> * You're going to be doing UEFI UpdateCapsule on edge hardware
> * What the size of the capsule updates are (e.g. 16MB, 32MB)

This is something for @nullr0ute (-:

> * How many kernels are installed, on the assumption you're sharing /boot and /boot/efi on the same physical partition

`/boot` will indeed be separate. Lennart is arguing for only having an ESP but we are not using one on PPC64LE or S390x and since some images types are going to be LVM (also customers really seem to want it) we are going to keep a separate `/boot`. Since the whole idea is to have one unified partitioning scheme we decided to keep `/boot` and EFI/ESP separate for now, so kernels won't be on the ESP/EFI.

> But all things considered, I think 200MB is probably fine as you're not going to have N distros installed, along with Windows 10.

Thanks for chiming in (so quickly)!

thozza:
@gicmo I'm happy to make the changes also for EDGE images, once there is a consensus. Nevertheless I would suggest to do it in a follow-up PR and I'm happy to make the changes. The reason for my suggestion to get this in sooner, rather than later is that it would be beneficial for QE and Layered Products such as RHOSP to get some images with these changes for testing sooner...

gicmo:
> @gicmo I'm happy to make the changes also for EDGE images, once there is a consensus. Nevertheless I would suggest to do it in a follow-up PR and I'm happy to make the changes. The reason for my suggestion to get this in sooner, rather than later is that it would be beneficial for QE and Layered Products such as RHOSP to get some images with these changes for testing sooner...

Makes sense! Works for me!

Issue #2020, opened 2021-11-16
Depsolving slowly inrecreasing disk usage under /var


lavocatt:
**Describe the bug**
`
/var/cache/osbuild-composer/rpmmd/` and `/var/cache/osbuild-worker/rpmmd/` are the two folders where the data is written.

**To Reproduce**
Let a worker depsolve for some days.

**Expected behavior**
The space taken should not grow.

lavocatt:
## hypothesis

`dnf-json` is relying on `libdnf` for fetching and updating the cache as well as for solving the package dependencies.
The cache path used for a repo is composed of the `id` of the repo and the `sha-256` of the repo base url (if given, `sha-256` of the id otherwise)

See from this chunk from `libdnf/repo/Repo.cpp`:

```cpp
std::string Repo::Impl::getHash() const
{
    std::string tmp;
    if (conf->metalink().empty() || (tmp=conf->metalink().getValue()).empty()) {
        if (conf->mirrorlist().empty() || (tmp=conf->mirrorlist().getValue()).empty()) {
            if (!conf->baseurl().getValue().empty())
                tmp = conf->baseurl().getValue()[0];
            if (tmp.empty())
                tmp = id;
        }
    }

    auto chksumObj = solv_chksum_create(REPOKEY_TYPE_SHA256);
    solv_chksum_add(chksumObj, tmp.c_str(), tmp.length());
    int chksumLen;
    auto chksum = solv_chksum_get(chksumObj, &chksumLen);
    static constexpr int USE_CHECKSUM_BYTES = 8;
    if (chksumLen < USE_CHECKSUM_BYTES) {
        solv_chksum_free(chksumObj, nullptr);
        throw Exception(_("getCachedir(): Computation of SHA256 failed"));
    }
    char chksumCStr[USE_CHECKSUM_BYTES * 2 + 1];
    solv_bin2hex(chksum, USE_CHECKSUM_BYTES, chksumCStr);
    solv_chksum_free(chksumObj, nullptr);

    return id + "-" + chksumCStr;
}

std::string Repo::Impl::getCachedir() const
{
    if (conf->basecachedir().empty()) {
        throw Exception(tfm::format(_("repo '%s': 'basecachedir' is not set"), id));
    }
    auto repodir(conf->basecachedir().getValue());
    if (repodir.back() != '/')
        repodir.push_back('/');
    return repodir + getHash();
}
```

[...]

PR #2023, opened 2021-11-16
Depsolving: avoid leaking disk space

lavocatt:
This commit removes the ID generation for dnfRepoConfig and replaces it
with the hash of the baseURL.
Solves #2020

This pull request includes:

### DIFF
--- a/internal/rpmmd/repository.go
+++ b/internal/rpmmd/repository.go
@@ -1,6 +1,7 @@
 package rpmmd
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -9,7 +10,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
@@ -395,10 +395,22 @@ func NewRPMMD(cacheDir, dnfJsonPath string) RPMMD {
 	}
 }
 
-func (repo RepoConfig) toDNFRepoConfig(rpmmd *rpmmdImpl, i int, arch, releasever string) (dnfRepoConfig, error) {
-	id := strconv.Itoa(i)
+func (repo RepoConfig) toDNFRepoConfig(rpmmd *rpmmdImpl, arch, releasever string) (string, dnfRepoConfig, error) {
+	/*
+		Generates an Id based on the content of the repo instead of using an arbitrary number. Because libdnf will use
+		this ID in the name of the folder cache, using an arbitrary number can lead to disk usage leaks.
+	*/
+	var data string
+	if repo.BaseURL != "" {
+		data = repo.BaseURL
+	} else if repo.Metalink != "" {
+		data = repo.Metalink
+	} else {
+		data = repo.MirrorList
+	}
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(data)))
 	dnfRepo := dnfRepoConfig{
-		ID:             id,
+		ID:             hash,
 		BaseURL:        repo.BaseURL,
 		Metalink:       repo.Metalink,
 		MirrorList:     repo.MirrorList,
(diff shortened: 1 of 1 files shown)


lavocatt:
This would not work yet because some requests might need an ID (the one where there is no base_url for the repos)

thozza:
So the ID is used as an identification of the repo in log messages in case there are some errors encountered while using it. Using hash instead of a numeric value may make it impossible to identify the problematic repo. I think that at least repo name should be set to make this possible. Some information about what I'm talking about is in #1863

ondrejbudai:
I think that this might indeed help in some cases and I agree with @thozza's concerns.

I have a feeling that I, unfortunately, omitted one fact, @lavocatt, sorry about that: The main issue with the leaking is in the internal service - brew calls us using the Koji API and we get a set of brand new baseurls each time. So the hash would be different there and thus, we still have the leaking issue.

But I'm still happy to merge this if we settle on id==name, that sounds promising - although we need to check whether all repos have names, if not, we need a fallback.

teg:
> But I'm still happy to merge this if we settle on id==name, that sounds promising - although we need to check whether all repos have names, if not, we need a fallback.

I agree with @thozza's point about making this debuggable. I'm still a bit concerned about reusing name as id, just because we would need to enforce:
 - name always exists
 - name is unique
 - name is suitable as an id

All of these things might always be true, but if have to add fallbacks we might not gain all that much. Would it be possible to keep id's opaque but either add a set a human-readable string to be printed in the logs (in case libdnfs supports that), or just print the id to name mapping in the logs too (as suggested by @lavocatt) so we could at least find out what it means when needed.

Apart from that: would love to see this land :)

thozza:
My understanding is that a DNF repo can have both `ID` and `name` set do different values. We currently set only `ID` and my suggestion as well as @msehnout's suggestion in the past was to set `ID` to whatever we need, but also set the repo's `name` to a human readable value and then use it in logs...

From DNF documentation about `name`: A human-readable name of the repository. Defaults to the `ID` of the repository. https://dnf.readthedocs.io/en/latest/conf_ref.html#repo-options

teg:
> My understanding is that a DNF repo can have both `ID` and `name` set do different values. We currently set only `ID` and my suggestion as well as @msehnout's suggestion in the past was to set `ID` to whatever we need, but also set the repo's `name` to a human readable value and then use it in logs...

Ah, perfect. Sorry, I had misunderstood you in the past. That makes perfect sense then. Could you make that part of this PR @lavocatt (as otherwise the new IDs would be even harder to debug) ?

lavocatt:
> > My understanding is that a DNF repo can have both `ID` and `name` set do different values. We currently set only `ID` and my suggestion as well as @msehnout's suggestion in the past was to set `ID` to whatever we need, but also set the repo's `name` to a human readable value and then use it in logs...
> 
> Ah, perfect. Sorry, I had misunderstood you in the past. That makes perfect sense then. Could you make that part of this PR @lavocatt (as otherwise the new IDs would be even harder to debug) ?

Of course! If I get it right, I'll keep the ID as it is in the PR and I'll add a requirement in the API to provide a human readable name for each repo. The I'll log the association name/repo to ease the debug ?

thozza:
@lavocatt I think that `dnf-json` already logs errors if there are any. It would be just good to verify that the `name` is used by DNF in these log messages, if provided. I think that it is important just to ensure that the `name` is set every time the list of repositories is passed to `dnf-json`. I'm not sure if it makes sense to make the repo `name` mandatory parameter of the Cloud API. Personally I think that this is not necessary and we can have some reasonable fallback. Using `name` is probably more important for the on-premise use case with Weldr API and we are already setting `name` for all repo definitions that we use by default.

teg:
I think you can set the name in the libdnf api and it will log correctly.

lavocatt:
I am closing this one because it is no longer relevant since https://github.com/osbuild/osbuild-composer/pull/2119.

PR #2024, opened 2021-11-16
debug: See if we can run some tests in aws instead of Openstack

jrusz:
We have a ton of jobs waiting for free Openstack runners, see if we can
run some of them in aws instead

This pull request includes:


ondrejbudai:
No tests that use libvirt (ostree, installer) can't be run on AWS was because it doesn't support KVM.

jrusz:
> No tests that use libvirt (ostree, installer) can be run on AWS was because it doesn't support KVM.

Yeah I though as much but seeing how many jobs are currently in queue waiting for Openstack runners I just wanted to try it and see for myself. I conclude that it either doesn't work as expected and also takes a long time. Closing as there is clearly no benefit.

PR #2026, opened 2021-11-16
Switch api tests to v2 & manifest job in api v2

Gundersanne:


Gundersanne:
@teg I wonder if we want some sort of hard timeout for the manifest job; I can imagine just like 100 manifest goroutines spinning endlessly. Or at least start printing warning messages after a certain amount of time so we're aware.

Ideally we'd capture this in metrics as well but we're only starting with per-job-type metrics.

ondrejbudai (dismissed):
Amazing, thanks!

Gundersanne:
ergh had a little error in subscriptions again

ondrejbudai (approved):

PR #2022, opened 2021-11-16
osbuild: check if result objects are nil in Write()

achilleas-k:
Before dereferencing the method receiver in Write(), check if the object is nil.

The result writer in osbuild1 (v1 results) isn't used anymore I think, but we should cover it until it's removed.

Fixes #2002

Issue #2002, opened 2021-11-12
weldr api crash in composelogshandler


Gundersanne:
**Describe the bug**
nil point deref

```
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: 2021/11/11 10:27:05 http: panic serving @: runtime error: invalid memory address or nil pointer dereference
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: goroutine 499 [running]:
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: net/http.(*conn).serve.func1(0xc0001841e0)
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]:         /usr/lib/golang/src/net/http/server.go:1804 +0x153
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: panic(0x561bd0fa6420, 0x561bd1483130)
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]:         /usr/lib/golang/src/runtime/panic.go:971 +0x499
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: github.com/osbuild/osbuild-composer/internal/osbuild1.(*Result).Write(0x0, 0x561bd1099938, 0xc000202600, 0x561bd109b898, 0xc000224380)
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]:         /builddir/build/BUILD/osbuild-composer-33.2/_build/src/github.com/osbuild/osbuild-composer/internal/osbuild1/result.go:87 +0x3c
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: github.com/osbuild/osbuild-composer/internal/weldr.(*API).composeLogsHandler(0xc000125680, 0x561bd10a7d90, 0xc000224380, 0xc000252e00, 0xc000104d00, 0x2, 0x4)
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]:         /builddir/build/BUILD/osbuild-composer-33.2/_build/src/github.com/osbuild/osbuild-composer/internal/weldr/api.go:2990 +0x965
Nov 11 10:27:05 localhost.localdomain osbuild-composer[15133]: github.com/julienschmidt/httprouter.(*Router).ServeHTTP(0xc00007c540, 0x561bd10a7d90, 0xc000224380, 0xc000252e00)
[...]

PR #2029, opened 2021-11-17
osbuild: check if result objects are nil in Write()

achilleas-k:
Cherry picked fix from #2022.

osbuild2 result does not have a `Write()` method in 8.5 so only osbuild1 needs the fix.

### DIFF
--- a/internal/osbuild1/result.go
+++ b/internal/osbuild1/result.go
@@ -84,8 +84,9 @@ func (result *StageResult) UnmarshalJSON(data []byte) error {
 }
 
 func (cr *Result) Write(writer io.Writer) error {
-	if cr.Build == nil && len(cr.Stages) == 0 && cr.Assembler == nil {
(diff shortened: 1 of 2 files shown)


ondrejbudai (changes requested):
oh... this is against 8.5... What's the plan here? Do we want to backport to z-stream? What doesn't actually work?

Blocking for now, if we want to backport, we need a bug firstly with the right acks.

Gundersanne:
The bug was reported on slack; https://github.com/osbuild/osbuild-composer/issues/2002

achilleas-k:
The idea was to backport, yes.  I opened the PR while we were chatting with Sanne and was going to write the bz for it but got held up by other things.

I think it's worth backporting.  This isn't very difficult to trigger and even if it's just for on-prem, it's still annoying.

Gundersanne:
This needs changes to the tests I think :/  I'm guessing 8.5 nightlies don't exist anymore

achilleas-k:
Closing.  Stale and irrelevant.

PR #2031, opened 2021-11-18
OCI support

rgolangh:
- OCI: vendor dependencies
- Add rpm build requierments
- Add build requirement in README.md
- Add support for OCI upload provider

Related-to: https://github.com/osbuild/cockpit-composer/pull/1374
Related-to: #2149 

ECOPROJECT-373


rgolangh:
@teg this is ready for review, although the parameters for the upload still need refinement.

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// GrantImageAccess grants the user the needed permissions to launch an instance
+// from the imageID in the designated compartment
+func (c Client) GrantImageAccess(imageID string, user string, compartmentID string) error {
+	// TBD
+	// should we support granting access to a group? oci have an api to grant
+	// access to groups or any-user (all users in tenancy) rather than specific user.
+	// also, not all regions support identity domains, i.e they'll need differnt method to
+	// allow access.
```
achilleas-k, line 137: How much work would it be to support both?  Would it need separate options, or could we have a single ID of some sort that would work for either a user or a group based on its type?

achilleas-k (commented):
Looking good.  I have some minor comments and questions below.
To expand on the access issue:
I think if there's a simple way to support both single users and all users in a tenancy we should support it.  If it's tricky to do, I suppose it depends on what we think the most common use case would be.  What do users of OCI expect to do most of the time?

teg:
Thanks for your work on this Roy! I guess in addition to the comments from Achilleas we need to hook this up with our CI tests?

rgolangh:
> Thanks for your work on this Roy! I guess in addition to the comments from Achilleas we need to hook this up with our CI tests?

Will gladly add it.
- we need a paid OCI account I guess as well - what is usually done to get an access to one? (my trial is about to expire)
- any code pointers or general guidance on adding it are wellcomed

cmd/osbuild-upload-oci/main.go
```diff
@@ -0,0 +1,108 @@
...
+
+	var uploader oci.Client
+	var err error
+
+	if objectName == "" {
+		fmt.Sprintf(
+			"%s-%s-%d",
+			fileName[strings.LastIndex(fileName, "/")+1:],
+			"image-builder-upload",
+			time.Now().Unix())
```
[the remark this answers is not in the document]
rgolangh: thanks for spotting this. I'm using 'cobra' now to compose the CLI command so now everything is easier and bit easier to get.

cmd/osbuild-upload-oci/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		panic(fmt.Errorf("failed to upload the image: %v", err))
+	}
+
+	imageID, err := uploader.Create(objectName, bucketName, bucketNamespace)
+	if err != nil {
+		panic(fmt.Errorf("failed to import the image from the bucket: %w", err))
+	}
+
+	fmt.Printf("Image(%s) was upladed and created successfully with the name %s under the bucket %s\n", imageID, objectName, bucketName)
```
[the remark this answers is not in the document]
rgolangh: done

cmd/osbuild-upload-oci/main.go
```diff
@@ -0,0 +1,108 @@
...
+func validateFlags() {
+	if bucketName == "" {
+		panic("please provide a valid [-bucketName name]")
+	}
+	if bucketNamespace == "" {
+		panic("please provide a valid [-bucketNamespace name]")
+	}
+	if fileName == "" {
+		panic("please specify [-fileName path] to upload")
+	}
```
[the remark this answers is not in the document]
rgolangh: resolved now that I'm using cobra

cmd/osbuild-upload-oci/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if objectName == "" {
+		fmt.Sprintf(
+			"%s-%s-%d",
+			fileName[strings.LastIndex(fileName, "/")+1:],
+			"image-builder-upload",
+			time.Now().Unix())
+	}
+	if privateKey != "" {
+		if tenancy == "" || region == "" || userID == "" {
+			panic(fmt.Sprintf("When suppling a private key the following args are mandatory as well: -tenancy, -region, and -user-id"))
```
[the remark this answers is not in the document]
rgolangh: > Looking good. I have some minor comments and questions below. To expand on the access issue: I think if there's a simple way to support both single users and all users in a tenancy we should support it. If it's tricky to do, I suppose it depends on what we think the most common use case would be. What do users of OCI expect to do most of the time?

The oci syntax for an IAM policy accepts a keyword of ''any-user" or a group id.  See 'Subject' here https://docs.oracle.com/en-us/iaas/Content/Identity/Concepts/policysyntax.htm

internal/upload/oci/upload.go
rgolangh: single ID for a user is not supported. Mayeb if ID is supplied then use that as the group ID, otherwise any-user?

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,140 @@
...
+			ObjectStorageClient:                 &c.storageClient,
+			EnableMultipartChecksumVerification: common.Bool(true),
+		},
+		FilePath: file.Name(),
+	}
+
+	ctx := context.Background()
+	resp, err := uploadManager.UploadFile(ctx, req)
+
+	if err != nil && resp.IsResumable() {
```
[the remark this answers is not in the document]
rgolangh: Since I don't think a user can explicitly set that then I'm not sure the error would add something. 
Reading the sdk code I see that if an upload fits into a single upload part (each part is 50Mb) then it is not resumable

rgolangh:
@achilleas-k  help needed here - rpm build for fedora fails because oracle/oci-sdk-go couldn't not be resolved by the rpm macro 'golang' - this does work for rhel.

ondrejbudai:
In Fedora, we cannot rely on vendored dependencies, the packaging rules forbid that. In order to proceed, we need to package https://github.com/oracle/oci-go-sdk in Fedora. I did a quick test and the results are promising: https://copr.fedorainfracloud.org/coprs/obudai/golang-github-oracle-oci-sdk/build/3122336/ All dependencies of oci-go-sdk are actually packaged in Fedora so we just need to package this one. The bad news is that `rpmbuild` fails on one file which is something we need to investigate.

Gundersanne:
> In Fedora, we cannot rely on vendored dependencies, the packaging rules forbid that. In order to proceed, we need to package https://github.com/oracle/oci-go-sdk in Fedora. I did a quick test and the results are promising: https://copr.fedorainfracloud.org/coprs/obudai/golang-github-oracle-oci-sdk/build/3122336/ All dependencies of oci-go-sdk are actually packaged in Fedora so we just need to package this one. The bad news is that `rpmbuild` fails on one file which is something we need to investigate.

I think cmd/genver can just be excluded; it's just a tool to update common/version.go.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -661,6 +677,47 @@ components:
...
+          example: 'aabbccdd'
+        image_name:
+          type: string
+          example: 'name-of-image'
+        private_key:
+          type: string
+          example: 'ssh-key-...'
+        fingerprint:
+          type: string
+          example: 'ssh-key-fingerprint...'
```
achilleas-k, line 715: These examples aren't necessary, but if we keep them, they should probably be valid but obviously fake examples of a key and fingerprint

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,234 @@
...
+
+	imageID, err := c.create(objectName, bucketName, namespace, compartmentID, imageName)
+	if err != nil {
+		return "", fmt.Errorf("failed to create a custom image using object '%s' bucket '%s' in namespace '%s': %w",
+			objectName,
+			bucketName,
+			namespace,
+			err)
+	}
+	err = c.GrantImageAccess(imageID, userID, compartmentID)
```
achilleas-k, line 84: I wonder if it would be better to expose `Create()` as public and use both `Create()` and `GrantImageAccess()` separately from `Upload()`, like we do with AWS.  It will make the external uploading and registering more granular (which could be better or worse), but it also makes it possible to have object uploads to OCI in the future, like we have with S3.

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,234 @@
...
+			return "", fmt.Errorf("failed to fetch the work request for creating the image: %w", err)
+		}
+		switch r.Status {
+		case workrequests.WorkRequestStatusSucceeded:
+			return *createImageResponse.Id, nil
+		case workrequests.WorkRequestStatusCanceled, workrequests.WorkRequestStatusFailed:
+			return "", fmt.Errorf("the work request for creating an image is in status %s", r.Status)
+		}
+		log.Printf("work request - status: %s, percent completed: %f\n", r.Status, *r.PercentComplete)
+		time.Sleep(10 * time.Second)
```
achilleas-k, line 136: Perhaps it would be better to reduce the interval to 1 second (or less even) and remove the progress printing so as to not flood the log.

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// GrantImageAccess grants the user the needed permissions to launch an instance
+// from the imageID in the designated compartment
+func (c Client) GrantImageAccess(imageID string, user string, compartmentID string) error {
+	// TBD
+	// should we support granting access to a group? oci have an api to grant
+	// access to groups or any-user (all users in tenancy) rather than specific user.
+	// also, not all regions support identity domains, i.e. they'll need different method to
+	// allow access.
```
achilleas-k, line 175: I don't know what the best approach is here.  We'd have to look into what's generally considered a sane default for OCI, I guess.

test/data/composer/osbuild-composer.toml
```diff
@@ -1,7 +1,9 @@
 [koji]
-allowed_domains = [ "localhost", "client.osbuild.org" ]
+allowed_domains = [ "localhost", "client.osbuild.org", "composer", "worker", "worker.osbuild.org", "composer.osbuild.org" ]
 ca = "/etc/osbuild-composer/ca-crt.pem"
 
 [worker]
-allowed_domains = [ "localhost", "worker.osbuild.org" ]
+allowed_domains = [ "localhost", "worker" ]
 ca = "/etc/osbuild-composer/ca-crt.pem"
+
```
achilleas-k, line 8: Was this changed for your local testing or do we need it for the changes here?

achilleas-k (commented):
This is all still looking good.  Other than the dependency packaging for Fedora, it seems we also need to figure out the `GrantImageAccess()` question and we're done.

The comments below are mostly minor things (typos and small recommendations or questions).

A more important note about changes to old distros:
We generally don't modify distro definitions for distribution versions that have already been released (e.g., RHEL 8.4 and 8.5).  For one, we wont make releases into those distro versions anymore (except in cases of critical bugs).

Has this been discussed?  Do we want to support RHEL 8.4 and RHEL 8.5 in OCI?  If it's part of the requirements for this, we might need to discuss how it should be handled (if we haven't already).

rgolangh:
> A more important note about changes to old distros: We generally don't modify distro definitions for distribution versions that have already been released (e.g., RHEL 8.4 and 8.5). For one, we wont make releases into those distro versions anymore (except in cases of critical bugs).
> 
> Has this been discussed? Do we want to support RHEL 8.4 and RHEL 8.5 in OCI? If it's part of the requirements for this, we might need to discuss how it should be handled (if we haven't already).

no, not discussed yet. 
Do you mean that the support for OCI in the osbuild will only appear in newer versions of RHEL 8.6?
Does an osbuild running on 8.6 supports producing images for lower versions of RHEL?

test/data/composer/osbuild-composer.toml
rgolangh: I'll revert this, that's a mistake

internal/upload/oci/upload.go
rgolangh: First, I read my code again and decide to rename create -> createImage.
For having more options to upload, I think that instead of exposing the low level methods, we can
make the oci.Client use a different Uploader implementation. And we can build it using 2 differnt constructor functions:
```
NewCustomImageUploader() (Client, error)

NewBucketImageUploader() (Client, error)
```

internal/upload/oci/upload.go
rgolangh: ack

internal/cloudapi/v2/openapi.v2.yml
rgolangh: ack, removing

internal/upload/oci/upload.go
```diff
@@ -0,0 +1,234 @@
...
+		switch r.Status {
+		case workrequests.WorkRequestStatusSucceeded:
+			return *createImageResponse.Id, nil
+		case workrequests.WorkRequestStatusCanceled, workrequests.WorkRequestStatusFailed:
+			return "", fmt.Errorf("the work request for creating an image is in status %s", r.Status)
+		}
+		log.Printf("work request - status: %s, percent completed: %f\n", r.Status, *r.PercentComplete)
+		time.Sleep(10 * time.Second)
+	}
+	return *createImageResponse.Id, nil
```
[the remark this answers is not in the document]
rgolangh: spot on, removing

internal/upload/oci/upload.go
rgolangh: any thoughts on having debug or other log levels? This is a valuable info to pass.
we can print the state only at fixed intervals, every 10/20 seconds and still loop each sec

rgolangh: anyhow, meanwhile I reduced the interval to 1 sec

internal/cloudapi/v2/openapi.v2.yml
rgolangh: > This is all still looking good. Other than the dependency packaging for Fedora, it seems we also need to figure out the `GrantImageAccess()` question and we're done.
> 
> The comments below are mostly minor things (typos and small recommendations or questions).
> 
> A more important note about changes to old distros: We generally don't modify distro definitions for distribution versions that have already been released (e.g., RHEL 8.4 and 8.5). For one, we wont make releases into those distro versions anymore (except in cases of critical bugs).
> 
> Has this been discussed? Do we want to support RHEL 8.4 and RHEL 8.5 in OCI? If it's part of the requirements for this, we might need to discuss how it should be handled (if we haven't already).

I'm dropping `distro/rhel84`

achilleas-k:
> > A more important note about changes to old distros: We generally don't modify distro definitions for distribution versions that have already been released (e.g., RHEL 8.4 and 8.5). For one, we wont make releases into those distro versions anymore (except in cases of critical bugs).
> > Has this been discussed? Do we want to support RHEL 8.4 and RHEL 8.5 in OCI? If it's part of the requirements for this, we might need to discuss how it should be handled (if we haven't already).
> 
> no, not discussed yet. Do you mean that the support for OCI in the osbuild will only appear in newer versions of RHEL 8.6? Does an osbuild running on 8.6 supports producing images for lower versions of RHEL?

Discussed in chat but adding info here as well:
- osbuild-composer does support producing images for lower versions of RHEL, but only if those distro versions implement the image type.
- We can add the image type to older versions, but only if it's absolutely necessary and explicitly required.

We decided to add OCI support only on newer versions and revisit the requirement for older version support later.

cmd/osbuild-upload-oci/main.go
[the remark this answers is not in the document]
achilleas-k: Don't think it's necessary to allow setting the name.

test/data/composer/osbuild-composer.toml
achilleas-k: Seems you missed this? :)

internal/upload/oci/upload.go
achilleas-k: That would also work.  So the `BucketImageUploader` would just upload an image to a bucket as a blob, and the `CustomImageUploader` would do the registration as well?
They'd both be reusing the bucket upload.

I think that it might obfuscate what's going on when someone's reading the caller code in the worker.  I'm more in favour of sort of self-documenting things.  The way we have it for AWS, which I think makes the flow clearer is:
```go
			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), options.Bucket, key)
			if err != nil {
				...
			}
			ami, err := a.Register(args.Targets[0].ImageName, options.Bucket, key, options.ShareWithAccounts, common.CurrentArch())
			if err != nil {
				...
			}
			if ami == nil {
				...
			}
```

and I think that makes it very easy to trace the whole thing (though admittedly, I cleaned up the error handling code).

test/data/composer/osbuild-composer.toml
Gundersanne: I think this is the only thing left :)

Gundersanne (changes requested):
So this PR doesn't add any tests but I'm happy to get this in as #2149 exists.

The only thing left is a nitpick regarding some test data. Some static analysis checks failing. 

And I'd drop the changes to the cloudapi/v2 for now. We cannot accept secrets from users, either a sharing mechanism where we upload to our account and share with a user's account (in which case secrets are parsed in the workers themselves) or an s3 download link are the options for now.

Gundersanne (approved):
@achilleas-k this lgtm, wdyt?

I think we can do the hosted version in a followup as we will need to have a sharing mechanism. OCI  doesn't support sharing custom images directly so we will probably end up with a presigned url that they can access and then make a custom image out of.

Let's do the CI for this in a follow-up as well?  See #2149

achilleas-k (dismissed):
LGTM but there's a code check error about using a bad rng.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -500,6 +505,49 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+			log.Print("[OCI] ⬆ Uploading the image")
+			file, err := os.Open(path.Join(outputDirectory, exportPath, options.FileName))
+			if err != nil {
+				appendTargetError(osbuildJobResult, fmt.Errorf("failed to create an OCI uploder: %w", err))
+				return nil
+			}
+			defer file.Close()
+			rand.Seed(time.Now().Unix())
+			imageID, err := ociClient.Upload(
+				fmt.Sprintf("osbuild-upload-%d", rand.Int63()),
```
Gundersanne, line 534: can we just use options.Filename here?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -500,6 +506,49 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+			log.Print("[OCI] ⬆ Uploading the image")
+			file, err := os.Open(path.Join(outputDirectory, exportPath, options.FileName))
+			if err != nil {
+				appendTargetError(osbuildJobResult, fmt.Errorf("failed to create an OCI uploder: %w", err))
+				return nil
+			}
+			defer file.Close()
+			i, _ := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+			imageID, err := ociClient.Upload(
+			fmt.Sprintf("osbuild-upload-%d", i),
```
Gundersanne, line 535: go fmt

Gundersanne (dismissed):
lgtm

achilleas-k (dismissed):
LGTM

I'm considering this PR done.  It's blocked on unrelated CI issues for now, so we'll have to rebase and reapprove when they're fixed.

Gundersanne (approved):

PR #2038, opened 2021-11-18
Test all manifests with depsolved package sets 

thozza:
- Introduce a new test case checking manifests of all supported combinations of image types against manifests which are part of image test cases, including depsolving of default image's package sets. Depsolving is important to catch unintentional changes in image's package sets. Run the test case in CI. all manifests are verified on a single runner within a single job, because the depsolving should not care about the system distribution. However if needed, the test case could be split to verify manifests of a specific distribution on the same distribution. The single CI job currently take a little over 12 minutes to finish (the test itself takes approximately 419 seconds).
- Fix one issue in the `rhel-86` `image-installer` image type, found by the new test case.
- Fix one issue in the `rhel-90` `ec2-sap` image type, found by the new test case.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -137,6 +137,27 @@ Base:
       - "*.repo"
     when: always
 
+Manifests:
+  stage: test
+  extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule"'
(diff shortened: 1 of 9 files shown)


achilleas-k (approved):
This is perfect, I love it!  Thank you for fixing my bugs :)

This is definitely going to help catch a lot of bugs related to package changes in the future.

PR #2040, opened 2021-11-19
templates: bump max postgres connections to 10

ondrejbudai:
By default, pgxpool.Pool has 4 connections (or number of cpus if higher).
Currently, we have 3 replicas, that means max 3*4=12 DB connections.

The dequeue operation is actually blocking - when a worker is waiting for
a job, one connection is blocked. My theory is that with 16 workers, we just
don't have enough connections that causes all sorts of weird slowdowns.

This commit bumps the number of connection from one replica to 10, therefore
we should be at 30 connections in total.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -15,6 +15,8 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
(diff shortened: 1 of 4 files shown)


Gundersanne (approved):

ondrejbudai:
Just ostree tests pending, this change has nothing to do with them.

PR #2042, opened 2021-11-19
cloudapi/v2: No ObjectReference in request bodies

Gundersanne:
Because of the way json encoding works in golang we haven't run into any
issues. But if we add automated validation based on the spec it will
break, the image-builder-crc client for instance doesn't specify these
fields.


teg (dismissed):

ondrejbudai:
Needs a rebase, otherwise looks good. :)

teg (approved):

ondrejbudai (approved):

teg (approved):

Issue #1785, opened 2021-09-15
Unify osbuild stage option validation


achilleas-k:
There are currently two ways in which we validate the stage options when there are restrictions on their keys or values:
- In a custom `MarshalJSON()`
- In the `NewXStage()` ctor

Original comment by @ondrejbudai follows:

I think I prefer the validation in `NewXStage` method because then it's not possible to create a stage with invalid options. I say: Fail fast!

Problems (and my proposed solutions):

- `NewXStage()` method panics: Either we can just live with that (usually it's just a programming error so it's fine to panic), or we should just return also `error` from this method.
- People can always use `XStageOptions{}` directly instead of using `NewXStage()`: Make `XStageOptions` private: `xStageOptions`. Some adjustments need to be done to the code but I think it might be worth it since the code is imho safer.

_Originally posted by @ondrejbudai in https://github.com/osbuild/osbuild-composer/pull/1768#discussion_r708049732_

thozza:
In addition, it would probably make sense to implement validation in stages according to their schema, which currently don't have it implemented...

PR #2051, opened 2021-11-22
osbuild2: update cloud-init stage with new options

msehnout:
Related to: https://github.com/osbuild/osbuild/pull/866/

This pull request includes:

### DIFF
--- a/internal/osbuild2/cloud_init_stage.go
+++ b/internal/osbuild2/cloud_init_stage.go
@@ -1,7 +1,6 @@
 package osbuild2
 
 import (
-	"encoding/json"
 	"fmt"
 )
 
@@ -13,6 +12,9 @@ type CloudInitStageOptions struct {
 func (CloudInitStageOptions) isStageOptions() {}
 
 func NewCloudInitStage(options *CloudInitStageOptions) *Stage {
+	if err := options.Config.validate(); err != nil {
+		panic(err)
+	}
 	return &Stage{
 		Type:    "org.osbuild.cloud-init",
 		Options: options,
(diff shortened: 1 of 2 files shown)


achilleas-k (commented):
We had a discussion recently about doing the option validation in the `NewXStage()` constructor instead of the `MarshalJSON()` method.
https://github.com/osbuild/osbuild-composer/issues/1785

In this case, I think it would probably also save us the trouble of having a custom marshaller, since there aren't any weird union values in the schema.

msehnout:
> We had a discussion recently about doing the option validation in the `NewXStage()` constructor instead of the `MarshalJSON()` method. #1785

Yeah, I read that one. Actually that's why I created all those `validate()` functions, so that I can call it in the `NewXStage`. Just forgot to actually call it :facepalm:

msehnout:
@achilleas-k Please have a look again. I wasn't sure whether to leave the `validate()` calls in the MarshalJSON functions or not. Right now the constructor as well as the marshalers call the functions, but it is a lot of boilerplate code. If we are ok with removing the validation from marshalers, we can remove a lot of lines and call the validation recursively (I didn't want to call `validate()` on internal variables in all the structures, because MarshalJSON calls MarshalJSON for every internal variable and in turn it calls `validate()`)

achilleas-k:
Looks good.  Having a nested/recursive `validate()` call that starts from the `NewStage()` is much nicer I think.  Let's clean the custom marshallers, which should also remove the need for the type aliases.  The whole thing should look much cleaner then.

msehnout:
I agree. The PR is updated and simplified now :-)

achilleas-k (approved):
Looks great.  Thanks!

PR #2052, opened 2021-11-22
tests/nightly: Re-enable satellite regression test on nightly composes

jrusz:
I might have made a mistake when I tested this last time. Trying again.

This pull request includes:


jrusz:
Nightly pipeline run with latest changes: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/416531988

jrusz:
Regression test in nightly pipeline passed https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1820611505

tools/define-compose-url.sh
```diff
@@ -8,16 +8,16 @@ if [[ $ID != rhel ]]; then
 fi
 
 if [[ $ID == rhel && ${VERSION_ID%.*} == 8 ]]; then
-  COMPOSE_ID=$(curl -L http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-finished-RHEL-"${VERSION_ID}"/COMPOSE_ID)
+  COMPOSE_ID=$(curl -L http://download.eng.bos.redhat.com/rhel-8/nightly/RHEL-8/latest-finished-RHEL-"${VERSION_ID}"/COMPOSE_ID)
```
atodorov, line 11: What is the reason for this change? 

AFAIK .devel. will redirect to the nearest mirror which could be eng.bos or anything else depending on where we provision.

atodorov (commented):
Looks good. Added one comment to understand the reason for changing some of the URLs.

tools/define-compose-url.sh
jrusz: >  What is the reason for this change?

There can be a time where other mirrors are not synced with bos yet (which I believe has happened when investigating the failures on nightlies) and by hard-coding this to bos we can be sure the compose is always there. But now that I think about it I'm not so sure about the benefit. If failure due to what I just said happens we can just restart the pipeline at a later time and if we see failures because of this we can always make this change later.. So would you prefer to keep devel there?

atodorov (approved):

PR #2061, opened 2021-11-24
ci: make some jobs interruptible

jrusz:
Jobs that don't interact with clouds can be canceled at any time without
the risk of leaving unused resources in the cloud. This enables the use
of "automatic cancellation of redundant pipelines" which means that if
and update is pushed to an open PR the current running pipeline is
canceled. This is done by adding an "interruptible" flag to the jobs.
Default value is false so only jobs that have it explicitly set to true
will be canceled.

It should help reduce stress on the CI in some cases.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,6 +36,7 @@ init:
     - shell
   script:
     - schutzbot/update_github_status.sh start
+  interruptible: true
 
 RPM:
   stage: rpmbuild
@@ -44,6 +45,7 @@ RPM:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
   script:
     - sh "schutzbot/mockbuild.sh"
+  interruptible: true
   after_script:
     - schutzbot/update_github_status.sh update
     - schutzbot/save_journal.sh
@@ -75,6 +77,7 @@ Container:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
   script:
     - sh "schutzbot/containerbuild.sh"
+  interruptible: true
   parallel:
     matrix:
       - RUNNER:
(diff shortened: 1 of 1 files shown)


jabia99:
We now have scheduled cloud cleaner for azure and aws. Do you think we could add the "interruptible" flag to script specific to this clouds? or should we wait to extend scheduled cloud cleaner to all the clouds?

jrusz:
> We now have scheduled cloud cleaner for azure and aws. Do you think we could add the "interruptible" flag to scripts that are specific to this clouds? or should we wait to extend scheduled cloud cleaner to all the clouds?

We would have to split Integration an API tests because of this and I think it would make the `gitlab-ci.yml` messy. I would just wait until we can apply it to all of them...

ondrejbudai:
Did you test this change? I'm curious to see some interrupted pipelines.

jrusz:
> Did you test this change? I'm curious to see some interrupted pipelines.

Yes I did in osbuild repo and also here, but it probably only works if the pipeline is started normally like without the error that we get where we have to cancel it and make it run again. 

It might not be visible but this https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/415633146 got canceled by me force pushing and then running this https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/415634708 which I then canceled on purpose.

Here is an example in osbuild: I think that this https://gitlab.com/osbuild/ci/osbuild/-/pipelines/415369457 got canceled in favor of this https://gitlab.com/osbuild/ci/osbuild/-/pipelines/415397990

jrusz:
I'm in no hurry with this so if you want to see it in action feel free to just ammend, force push and do it again and watch the result.

ondrejbudai (approved):

PR #2062, opened 2021-11-24
dnf json as a service

lavocatt:
This pull request includes:

### DIFF
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -5,8 +5,8 @@
 package main
 
 import (
-	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"testing"
@@ -23,6 +23,11 @@ import (
 
 func TestFetchChecksum(t *testing.T) {
 	dir, err := test.SetUpTemporaryRepository()
+	fs := http.FileServer(http.Dir(dir))
+	go func() {
+		err := http.ListenAndServe(":9000", fs)
+		assert.Nilf(t, err, "Could not start the http server: %v", err)
+	}()
 	defer func(dir string) {
 		err := test.TearDownTemporaryRepository(dir)
 		assert.Nil(t, err, "Failed to clean up temporary repository.")
@@ -31,13 +36,13 @@ func TestFetchChecksum(t *testing.T) {
 
 	repoCfg := rpmmd.RepoConfig{
 		Name:      "repo",
-		BaseURL:   fmt.Sprintf("file://%s", dir),
+		BaseURL:   "http://localhost:9000",
 		IgnoreSSL: true,
 	}
 
 	// use a fullpath to dnf-json, this allows this test to have an arbitrary
 	// working directory
-	rpmMetadata := rpmmd.NewRPMMD(path.Join(dir, "rpmmd"), "/usr/libexec/osbuild-composer/dnf-json")
+	rpmMetadata := rpmmd.NewRPMMD(path.Join(dir, "rpmmd"))
 	_, c, err := rpmMetadata.FetchMetadata([]rpmmd.RepoConfig{repoCfg}, "platform:f31", "x86_64", "31")
 	assert.Nilf(t, err, "Failed to fetch checksum: %v", err)
 	assert.NotEqual(t, "", c["repo"], "The checksum is empty")
@@ -64,7 +69,7 @@ func TestCrossArchDepsolve(t *testing.T) {
 
 			// use a fullpath to dnf-json, this allows this test to have an arbitrary
 			// working directory
-			rpm := rpmmd.NewRPMMD(dir, "/usr/libexec/osbuild-composer/dnf-json")
+			rpm := rpmmd.NewRPMMD(dir)
 
 			repos, err := rpmmd.LoadRepositories([]string{repoDir}, distroStruct.Name())
 			require.NoErrorf(t, err, "Failed to LoadRepositories %v", distroStruct.Name())
--- a/distribution/osbuild-dnf-json.service
+++ b/distribution/osbuild-dnf-json.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=OSbuild dnf-json service
+Requires=osbuild-dnf-json.socket
+
+[Service]
+Type=simple
+PrivateTmp=true
+ExecStart=/usr/libexec/osbuild-composer/dnf-json
+Restart=always
+RestartSec=1s
+CacheDirectory=osbuild-dnf-json
+# systemd >= 240 sets this, but osbuild-worker runs on earlier versions
+Environment="CACHE_DIRECTORY=/var/cache/osbuild-dnf-json"
--- a/distribution/osbuild-dnf-json.socket
+++ b/distribution/osbuild-dnf-json.socket
@@ -1,0 +1,9 @@
+[Unit]
+Description=OSbuild dnf-json socket
+
+[Socket]
+Service=osbuild-dnf-json.service
+ListenStream=/run/osbuild-dnf-json/api.sock
+
+[Install]
+WantedBy=sockets.target
--- a/distribution/osbuild-worker@.service
+++ b/distribution/osbuild-worker@.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=OSBuild Composer Worker (%i)
-Requires=osbuild-local-worker.socket
+Requires=osbuild-local-worker.socket osbuild-dnf-json.socket
 After=network.target osbuild-local-worker.socket
 
 [Service]
--- a/dnf-json
+++ b/dnf-json
@@ -10,6 +10,7 @@ import socketserver
 import logging
 import sys
 from http.server import BaseHTTPRequestHandler
+import pathlib
 
 import dnf
 import hawkey
@@ -254,15 +255,21 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
 log.info("Starting the dnf-json server")
 
 LISTEN_FDS = int(os.environ.get("LISTEN_FDS", 0))
+SOCK_PATH = "/run/osbuild-dnf-json/"
+SOCK_NAME = "api.sock"
 
 class SystemDActivationSocketServer(socketserver.ForkingMixIn, socketserver.UnixStreamServer):
     def server_bind(self):
         log.debug("service bind")
-        assert LISTEN_FDS == 1
-        log.debug("rebind socket")
-        log.debug("address_family: %d ", self.address_family)
-        log.debug("socket_type: %d ", self.socket_type)
-        self.socket = socket.fromfd(3, self.address_family, self.socket_type)
+        if LISTEN_FDS == 0:
+            log.debug("create new socket")
+            socketserver.UnixStreamServer.server_bind(self)
+        else:
+            log.debug("rebind socket")
+            log.debug("address_family: %d ", self.address_family)
+            log.debug("socket_type: %d ", self.socket_type)
+            self.socket = socket.fromfd(3, self.address_family, self.socket_type)
 
-server = SystemDActivationSocketServer('', DnfJsonRequestHandler)
+pathlib.Path(SOCK_PATH).mkdir(parents=True, exist_ok=True)
+server = SystemDActivationSocketServer(f"{SOCK_PATH}{SOCK_NAME}", DnfJsonRequestHandler)
 server.serve_forever()
--- a/dnf-json
+++ b/dnf-json
@@ -8,130 +8,183 @@ import os
 import socket
 import socketserver
 import logging
+import sys
 from http.server import BaseHTTPRequestHandler
 
 import dnf
 import hawkey
-from systemd.journal import JournalHandler
 
+formatter = logging.Formatter(
+        fmt="%(asctime)s %(name)s.%(levelname)s: %(message)s",
+        datefmt="%Y.%m.%d %H:%M:%S"
+        )
+handler = logging.StreamHandler(stream=sys.stdout)
+handler.setFormatter(formatter)
 
 log = logging.getLogger('dnf-json')
-log.addHandler(JournalHandler())
+log.addHandler(handler)
 log.setLevel(logging.INFO)
 
-DNF_ERROR_EXIT_CODE = 10
+class Solver():
 
-def timestamp_to_rfc3339(timestamp, package):
-    d = datetime.datetime.utcfromtimestamp(package.buildtime)
-    return d.strftime('%Y-%m-%dT%H:%M:%SZ')
+    def __init__(self, repos, module_platform_id, persistdir, cachedir, arch):
+        self.base = dnf.Base()
 
-def dnfrepo(desc, parent_conf=None):
-    """Makes a dnf.repo.Repo out of a JSON repository description"""
+        # Enable fastestmirror to ensure we choose the fastest mirrors for
+        # downloading metadata (when depsolving) and downloading packages.
+        self.base.conf.fastestmirror = True
 
-    repo = dnf.repo.Repo(desc["id"], parent_conf)
+        # We use the same cachedir for multiple architectures. Unfortunately,
+        # this is something that doesn't work well in certain situations
+        # with zchunk:
+        # Imagine that we already have cache for arch1. Then, we use dnf-json
+        # to depsolve for arch2. If ZChunk is enabled and available (that's
+        # the case for Fedora), dnf will try to download only differences
+        # between arch1 and arch2 metadata. But, as these are completely
+        # different, dnf must basically redownload everything.
+        # For downloding deltas, zchunk uses HTTP range requests. Unfortunately,
+        # if the mirror doesn't support multi range requests, then zchunk will
+        # download one small segment per a request. Because we need to update
+        # the whole metadata (10s of MB), this can be extremely slow in some cases.
+        # I think that we can come up with a better fix but let's just disable
+        # zchunk for now. As we are already downloading a lot of data when
+        # building images, I don't care if we download even more.
+        self.base.conf.zchunk = False
 
-    if "baseurl" in desc:
-        repo.baseurl = desc["baseurl"]
-    elif "metalink" in desc:
-        repo.metalink = desc["metalink"]
-    elif "mirrorlist" in desc:
-        repo.mirrorlist = desc["mirrorlist"]
-    else:
-        assert False
+        # Try another mirror if it takes longer than 5 seconds to connect.
+        self.base.conf.timeout = 5
 
-    if desc.get("ignoressl", False):
-        repo.sslverify = False
-    if "sslcacert" in desc:
-        repo.sslcacert = desc["sslcacert"]
-    if "sslclientkey" in desc:
-        repo.sslclientkey = desc["sslclientkey"]
-    if "sslclientcert" in desc:
-        repo.sslclientcert = desc["sslclientcert"]
+        # Set the rest of the dnf configuration.
+        self.base.conf.module_platform_id = module_platform_id
+        self.base.conf.config_file_path = "/dev/null"
+        self.base.conf.persistdir = persistdir
+        self.base.conf.cachedir = cachedir
+        self.base.conf.substitutions['arch'] = arch
+        self.base.conf.substitutions['basearch'] = dnf.rpm.basearch(arch)
 
-    # In dnf, the default metadata expiration time is 48 hours. However,
-    # some repositories never expire the metadata, and others expire it much
-    # sooner than that. We therefore allow this to be configured. If nothing
-    # is provided we error on the side of checking if we should invalidate
-    # the cache. If cache invalidation is not necessary, the overhead of
-    # checking is in the hundreds of milliseconds. In order to avoid this
-    # overhead accumulating for API calls that consist of several dnf calls,
-    # we set the expiration to a short time period, rather than 0.
-    repo.metadata_expire = desc.get("metadata_expire", "20s")
+        for repo in repos:
+            self.base.repos.add(self._dnfrepo(repo, self.base.conf))
+        self.base.fill_sack(load_system_repo=False)
 
-    return repo
+    def _dnfrepo(self, desc, parent_conf=None):
+        """Makes a dnf.repo.Repo out of a JSON repository description"""
 
+        repo = dnf.repo.Repo(desc["id"], parent_conf)
 
-def create_base(repos, module_platform_id, persistdir, cachedir, arch):
-    base = dnf.Base()
-
-    # Enable fastestmirror to ensure we choose the fastest mirrors for
-    # downloading metadata (when depsolving) and downloading packages.
-    base.conf.fastestmirror = True
-
-    # We use the same cachedir for multiple architectures. Unfortunately,
-    # this is something that doesn't work well in certain situations
-    # with zchunk:
-    # Imagine that we already have cache for arch1. Then, we use dnf-json
-    # to depsolve for arch2. If ZChunk is enabled and available (that's
-    # the case for Fedora), dnf will try to download only differences
-    # between arch1 and arch2 metadata. But, as these are completely
-    # different, dnf must basically redownload everything.
-    # For downloding deltas, zchunk uses HTTP range requests. Unfortunately,
-    # if the mirror doesn't support multi range requests, then zchunk will
-    # download one small segment per a request. Because we need to update
-    # the whole metadata (10s of MB), this can be extremely slow in some cases.
-    # I think that we can come up with a better fix but let's just disable
-    # zchunk for now. As we are already downloading a lot of data when
-    # building images, I don't care if we download even more.
-    base.conf.zchunk = False
-
-    # Try another mirror if it takes longer than 5 seconds to connect.
-    base.conf.timeout = 5
-
-    # Set the rest of the dnf configuration.
-    base.conf.module_platform_id = module_platform_id
-    base.conf.config_file_path = "/dev/null"
-    base.conf.persistdir = persistdir
-    base.conf.cachedir = cachedir
-    base.conf.substitutions['arch'] = arch
-    base.conf.substitutions['basearch'] = dnf.rpm.basearch(arch)
-
-    for repo in repos:
-        base.repos.add(dnfrepo(repo, base.conf))
-
-    base.fill_sack(load_system_repo=False)
-    return base
-
-def repo_checksums(base):
-    checksums = {}
-    for repo in base.repos.iter_enabled():
-        # Uses the same algorithm as libdnf to find cache dir:
-        #   https://github.com/rpm-software-management/libdnf/blob/master/libdnf/repo/Repo.cpp#L1288
-        if repo.metalink:
-            url = repo.metalink
-        elif repo.mirrorlist:
-            url = repo.mirrorlist
-        elif repo.baseurl:
-            url = repo.baseurl[0]
+        if "baseurl" in desc:
+            repo.baseurl = desc["baseurl"]
+        elif "metalink" in desc:
+            repo.metalink = desc["metalink"]
+        elif "mirrorlist" in desc:
+            repo.mirrorlist = desc["mirrorlist"]
         else:
             assert False
 
-        digest = hashlib.sha256(url.encode()).hexdigest()[:16]
+        if desc.get("ignoressl", False):
+            repo.sslverify = False
+        if "sslcacert" in desc:
+            repo.sslcacert = desc["sslcacert"]
+        if "sslclientkey" in desc:
+            repo.sslclientkey = desc["sslclientkey"]
+        if "sslclientcert" in desc:
+            repo.sslclientcert = desc["sslclientcert"]
 
-        repomd_file = f"{repo.id}-{digest}/repodata/repomd.xml"
-        with open(f"{base.conf.cachedir}/{repomd_file}", "rb") as f:
-            repomd = f.read()
+        # In dnf, the default metadata expiration time is 48 hours. However,
+        # some repositories never expire the metadata, and others expire it much
+        # sooner than that. We therefore allow this to be configured. If nothing
+        # is provided we error on the side of checking if we should invalidate
+        # the cache. If cache invalidation is not necessary, the overhead of
+        # checking is in the hundreds of milliseconds. In order to avoid this
+        # overhead accumulating for API calls that consist of several dnf calls,
+        # we set the expiration to a short time period, rather than 0.
+        repo.metadata_expire = desc.get("metadata_expire", "20s")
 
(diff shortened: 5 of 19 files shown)


distribution/dnf-json.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=OSBuild dng-json service
+After=network.target
```
teg, line 3: I don't think you need to order after the network target, but maybe I am missing something?

dnf-json
```diff
@@ -1,21 +1,30 @@
...
-DNF_ERROR_EXIT_CODE = 10
 
+log = logging.getLogger('dnf-json')
+log.addHandler(JournalHandler())
+log.setLevel(logging.INFO)
+
+DNF_ERROR_EXIT_CODE = 10
 
-def timestamp_to_rfc3339(timestamp):
+def timestamp_to_rfc3339(timestamp, package):
```
teg: This change looks right, but how did this work before?

teg: If this is a bug fix, maybe making it a separate commit would be the correct thing to do.

dnf-json
```diff
@@ -123,98 +126,147 @@ def repo_checksums(base):
...
-                "DepsolveError",
-                (
-                    "There was a problem depsolving "
-                    f"{arguments['package-specs']}: {e}"
+class DnfJsonRequestHandler(BaseHTTPRequestHandler):
+
+    def _send(self):
+        self.client_address=('',)
+
+    def exit_with_dnf_error(self, kind: str, reason: str):
```
teg: Should probably be `return_with_` rather than `exit_with_`, as we no longer exit the binary?

dnf-json
```diff
@@ -123,98 +126,147 @@ def repo_checksums(base):
...
+        self._send()
+        self.send_response(500)
```
teg, line 193: This preserves the behaviour (I think), so it is the right thing for the first version. But I would love to see this improved to return structured 400 errors in case of user error to explain all the different ways something might have gone wrong (conflicting requests, invalid repository, ...).

dnf-json
```diff
@@ -123,98 +126,147 @@ def repo_checksums(base):
...
+                    packages.append({
+                        "name": package.name,
+                        "summary": package.summary,
+                        "description": package.description,
+                        "url": package.url,
+                        "epoch": package.epoch,
+                        "version": package.version,
+                        "release": package.release,
+                        "arch": package.arch,
+                        "buildtime": timestamp_to_rfc3339(package.buildtime, package),
```
teg: Why do we need to pass both `package` and `package.buildtime`?

dnf-json
```diff
@@ -123,98 +126,147 @@ def repo_checksums(base):
...
+
+LISTEN_FDS = int(os.environ.get("LISTEN_FDS", 0))
+LISTEN_PID = os.environ.get("LISTEN_PID", None) or os.getpid()
+
+log.debug("env %d and %d", LISTEN_FDS, LISTEN_PID)
+
+class SystemDActivationSocketServer(socketserver.UnixStreamServer):
+    def server_bind(self):
+        log.debug("service bind")
+        if LISTEN_FDS == 0:
```
teg, line 269: I think we may want to only support LISTEN_FDS=1 so we avoid confusion due to both this dameon and systemd "owning" the same socket.

osbuild-composer.spec
```diff
@@ -307,6 +307,7 @@ cd $PWD/_build/src/%{goipath}
 %doc README.md
 %{_mandir}/man7/%{name}.7*
 %{_unitdir}/osbuild-composer.service
+%{_unitdir}/dnf-json.socket
```
teg, line 310: `osbuild-dnf-json.socket`

distribution/dnf-json.service
```diff
@@ -0,0 +1,19 @@
...
+Type=simple
+PrivateTmp=true
+ExecStart=/usr/libexec/osbuild-composer/dnf-json
+Restart=always
+RestartSec=1s
+CacheDirectory=osbuild-dnf-json
+# systemd >= 240 sets this, but osbuild-worker runs on earlier versions
+Environment="CACHE_DIRECTORY=/var/cache/osbuild-dnf-json"
+StandardOutput=journal
+StandardError=journal
```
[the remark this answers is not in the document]
lavocatt: Indeed. I commited them by mistake ^^

distribution/dnf-json.service
lavocatt: That's copy and paste from `osbuild-remote-worker@.service` and `osbuild-worker@.service` I did not think about this too much. Should I remove this ?

dnf-json
lavocatt: I the previous version the variable package was in the global scope thus accessible by the function. Since I moved the declaration to inside a class, the function was no longer able to access it, so I added it as a parameter.

dnf-json
lavocatt: It is solved by the next commit.

dnf-json
lavocatt: Yes it does preserve the behaviour because we check in composer the value of the HTTP code. Before we were checking the value of the return code of DNF.

At the moment we already forward the reason why to the end user:

```
 time curl -k --key /etc/osbuild-composer/client-key.pem --cert /etc/osbuild-composer/client-crt.pem -H "Content-Type: application/json" -d @../tests/1477.json https://localhost/api/composer/v1/compose -i 
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=UTF-8
Date: Wed, 24 Nov 2021 17:02:46 GMT
Content-Length: 225

{"message":"Failed to depsolve requested package set: DNF error occured: MarkingErrors: Error occurred when marking packages for installation: Problems in request:\nmissing packages: jesuisunpaquetquinexistepas_idonotexist"}

real	0m0.754s
user	0m0.033s
sys	0m0.009s
```

distribution/dnf-json.service
teg: I would remove it. This would make sure NetworkManager is started before the service (doesn't mean anything as the network might still be down) and stopped after it (which might be important to avoid network connection being dropped). However, we are not exposed to the network, so this should not be necessary. I would drop it. Unless someone corrects me ;) @gicmo @ondrejbudai ?

PS
I think it should not be in `osbuild-worker@.service` either, but in `-remote-` it makes sense.

dnf-json
teg: I think this is right for now, but in the future we should make this a structured object so we can show the error message nicely in the UI. And we should distinguish between 400 (user made a mistake) and 500 (we made a mistake). Something like:
```
{
  "type": "MissingPackage",
  "payload": {
    "packages": [
      "jesuisunpaquetquinexistepas_idonotexist"
    ]
  }
}
```

dnf-json
```diff
@@ -159,98 +209,46 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
-                })
+                )
+            except dnf.exceptions.Error as e:
+                self.response_with_dnf_error(
+                    type(e).__name__,
+                    f"Error occurred when setting up repo: {e}")
 
 log.info("Starting the dnf-json server")
 
 SOCKET_FILE_PATH = "/run/osbuild-dnf-json/api.sock"
```
teg, line 251: Nitpick: I don't think this is used any more?

teg (approved):
Looks great to me! Maybe a review from @ondrejbudai and/or @Gundersanne would be great too, as this is quite delicate stuff.

distribution/osbuild-dnf-json.service
```diff
@@ -0,0 +1,17 @@
...
+PrivateTmp=true
+ExecStart=/usr/libexec/osbuild-composer/dnf-json
+Restart=always
+RestartSec=1s
+CacheDirectory=osbuild-dnf-json
+# systemd >= 240 sets this, but osbuild-worker runs on earlier versions
+Environment="CACHE_DIRECTORY=/var/cache/osbuild-dnf-json"
+
+[Install]
+WantedBy=osbuild-composer.service
```
Gundersanne, line 17: Should be wanted by the workers too though, no @teg ? Or we include `Wants=` or `Require=` in workers and composer rather than WantedBy here? not sure

dnf-json
Gundersanne: Since depsolve is a job now  and asynchronous (as in, the request succeeds if the job was queued). So it wouldn't result in 400s/500s. But of course we still need a good way of distinguishing errors yea.

internal/cloudapi/v1/v1.go
```diff
@@ -175,9 +175,9 @@ func (h *apiHandlers) Compose(ctx echo.Context) error {
 
 		if depsolveResults.Error != "" {
 			if depsolveResults.ErrorType == worker.DepsolveErrorType {
-				return echo.NewHTTPError(http.StatusBadRequest, "Failed to depsolve requested package set: %s", depsolveResults.Error)
+				return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Failed to depsolve requested package set: %s", depsolveResults.Error))
```
Gundersanne, line 178: Why add `fmt.Sprintf` ? Also cloudapi/v1 is deprecated (everything uses v2 now). It's also untested. I'd be ok with dropping any changes to cloudapi/v1 (in fact we sohuld just remove cloudapi/v1 soon, separate PR of course).

Gundersanne (commented):
Ah this looks so neat! I have a few comments/questions.

distribution/osbuild-dnf-json.service
teg: Good point! I didn't think of this. I don't think we need an install section at all, we should just require the socket from the service that depend on it. The service would be started on demand.

ondrejbudai: Agreed with both of you, I suggest dropping the `Install` section and just directly using `Requires`, or `Wanted`. For now, I would just put `Requires` everywhere, it's imho the safest option.

distribution/osbuild-dnf-json.socket
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=OSbuild dnf-json socket
```
ondrejbudai, line 2: The tiniest nitpick ever: One unit has `OSBuild` in description, the other one has `OSbuild`.

internal/rpmmd/repository.go
```diff
@@ -322,39 +324,31 @@ func runDNF(dnfJsonPath string, command string, arguments interface{}, result in
...
 	}
 
-	err = cmd.Start()
+	bpost, err := json.Marshal(call)
 	if err != nil {
 		return err
 	}
 
-	err = json.NewEncoder(stdin).Encode(call)
+	response, err := httpc.Post("http://unix", "application/json", strings.NewReader(string(bpost)))
```
ondrejbudai, line 340: `bytes.NewReader` would avoid a conversion

internal/cloudapi/v1/v1.go
ondrejbudai: Agreed.

ondrejbudai (approved):
This is awesome! :-) `dnf-json` looks much cleaner after this change. I suggested several changes, up to you if you want to address them now, or in a follow-up.

internal/cloudapi/v1/v1.go
lavocatt: Oh yeah.. I really have to stop using v1 for my tests!

distribution/osbuild-dnf-json.service
lavocatt: So should I make the worker require something ?

teg: The worker should require the `.socket`, and that should be all that's needed, I think.

dnf-json
lavocatt: Ok so I was not testing it with the v2. Now I'm doing it and I can see that indeed there's no way for a user to get feedback, at least through composer's api.

```
time curl -k --key /etc/osbuild-composer/client-key.pem --cert /etc/osbuild-composer/client-crt.pem -H "Content-Type: application/json" -d @../tests/compose_request_v2.json https://localhost/api/image-builder-composer/v2/compose -i 
HTTP/1.1 201 Created
Content-Type: application/json; charset=UTF-8
Date: Thu, 25 Nov 2021 14:44:54 GMT
Content-Length: 113

{"href":"/api/image-builder-composer/v2/compose","kind":"ComposeId","id":"2989cd1f-5647-4783-ae58-6a2770cf89e1"}

real	0m0.043s
user	0m0.031s
sys	0m0.005s
┌─(✓)─[tlavocat@fedora]─[16% <─ ~/dev/tests]
└──╼> time curl -k --key /etc/osbuild-composer/client-key.pem --cert /etc/osbuild-composer/client-crt.pem -H "Content-Type: application/json" -d @../tests/compose_request_v2.json https://localhost/api/image-builder-composer/v2/composes/2989cd1f-5647-4783-ae58-6a2770cf89e1 -i 
HTTP/1.1 404 Not Found
Content-Type: application/json; charset=UTF-8
Date: Thu, 25 Nov 2021 14:52:49 GMT
Content-Length: 201

{"href":"/api/image-builder-composer/v2/errors/21","id":"21","kind":"Error","code":"IMAGE-BUILDER-COMPOSER-21","operation_id":"21PoEE2eQFJMusCbAfSNVy9rXKv","reason":"Requested resource doesn't exist"}

real	0m0.041s
user	0m0.026s
sys	0m0.011s
```
How can we improve that ?

distribution/osbuild-dnf-json.service
lavocatt: Hmm, when I add the dependency to the `.socket` systemd refuses to start the process because it's missing a dependency

lavocatt: And also, if the osbuil-dnf-json.service does not require its socket, the activation stops working.

ondrejbudai: > Hmm, when I add the dependency to the .socket systemd refuses to start the process because it's missing a dependency

Can you post diff of your changes and the message from log?

> And also, if the osbuil-dnf-json.service does not require its socket, the activation stops working.

I think that the dnf-json.service always has to require its socket. That shouldn't be changed.

teg: Both `osbuild-worker*.service` and `osbuild-dnf-json.service` should have `Requires=osbuild-dnf-json.socket`. I think that should be all that's needed.

lavocatt: Do you think that `osbuild-remote-worker@.service` service too ? It is unclear because it does not have any requirements now.

teg: Yeah, anything that will connect to the socket needs to `Require` it.

distribution/osbuild-worker@.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=OSBuild Composer Worker (%i)
-Requires=osbuild-local-worker.socket
+Requires=osbuild-local-worker.socket osbuild-dnf-json.socket
```
Gundersanne, line 3: Also needs to be added in `osbuild-remote-worker@.service` and `osbuild-composer.service`; they all require dnf-json.

lavocatt:
Unfortunately I discovered that I have memory leaks to take care :/

thozza (commented):
This looks very interesting and I like where this is going! However I have a major issue with the current version.

Please correct me if I'm wrong, but I think that running `dnf-json` only as a service will break our tooling used to generate image test cases. Specifically `osbuild-pipeline` is not run as a service and it uses RPMMD with `dnf-json` to depsolve image package sets: https://github.com/osbuild/osbuild-composer/blob/8136e0f069efbae0d68cb1b6ad18913df7492ada/cmd/osbuild-pipeline/main.go#L121
https://github.com/osbuild/osbuild-composer/blob/8136e0f069efbae0d68cb1b6ad18913df7492ada/tools/test-case-generators/generate-test-cases#L66-L70
 
The issue is currently noticeable in the failing CI Manifest tests because they use the same approach as generators of image test cases.

Running `dnf-json` as a service probably makes sense, however personally I think that the generation of image test cases should be kept simple (executing a single command). This could mean that all tooling is modified to spin up `dnf-json` service in the background and stop it once it is not needed any more, or that `dnf-json` preserves the ability to be run as a command, not only as a service. I don't have a strong preference, but I think that this needs to be solved before the PR can be merged.

internal/rpmmd/repository.go
```diff
@@ -322,39 +325,31 @@ func runDNF(dnfJsonPath string, command string, arguments interface{}, result in
 		arguments,
 	}
 
-	cmd := exec.Command(dnfJsonPath)
```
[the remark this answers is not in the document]
lavocatt: > This looks very interesting and I like where this is going! However I have a major issue with the current version.
> [...]
> Running `dnf-json` as a service probably makes sense, however personally I think that the generation of image test cases should be kept simple (executing a single command). This could mean that all tooling is modified to spin up `dnf-json` service in the background and stop it once it is not needed any more, or that `dnf-json` preserves the ability to be run as a command, not only as a service. I don't have a strong preference, but I think that this needs to be solved before the PR can be merged.

Thanks a lot! Yes you are right. I'm gonna change the tooling so that it uses the dameon version.

lavocatt:
> The issue is currently noticeable in the failing CI Manifest tests because they use the same approach as generators of image test cases.

The issue in the manifest test case is due to a memory leak in the daemon.

```
Nov 30 08:40:04 ip-10-30-18-53.us-east-1.aws.redhat.com kernel: Out of memory: Killed process 19731 (dnf-json) total-vm:3905420kB, anon-rss:3607436kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:7272kB oom_score_adj:0

```

lavocatt:
> Please correct me if I'm wrong, but I think that running `dnf-json` only as a service will break our tooling used to generate image test cases. Specifically `osbuild-pipeline` is not run as a service and it uses RPMMD with `dnf-json` to depsolve image package sets:

@thozza I've corrected the issue. I removed the dnfJsonPath from the api of RPMMD.
It works well also since I solved the memory leak issue (see https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1832645625). 

For clients of the Rpmmd api the behaviour is the same as before, because repository.go hides the complexity of calling the daemon through the unix socket.

SystemD is responsible for starting the process and I don't think that we should shut it down. Maybe we can add a timeout that closes the daemon if no communication was made to it for along time.

lavocatt:
My main issue now is that `create_repo_c` is creating a repository with access rights that does not allow the `dnf-json` daemon to access the files. I'll probably have to `chmod` stuff, but it seems hacky. I'll try to figure out something better.

ondrejbudai:
Isn't this a SELinux issue? IIRC SELinux prevents root processes to access home directories. A solution might be to run an ad-hoc HTTP server, I suggest something like `python -m http.server`, or a simple Go thing.

thozza:
> For clients of the Rpmmd api the behaviour is the same as before, because repository.go hides the complexity of calling the daemon through the unix socket.
> 
> SystemD is responsible for starting the process and I don't think that we should shut it down. Maybe we can add a timeout that closes the daemon if no communication was made to it for along time.

Right, this will work really nicely except for the case when generating image test cases...

The approach expects that RPMs which ship the `dnf-json` as a service are installed on the system when the API is used, e.g. when image test cases are being generated. This is in general not the case when generating image test cases, because thanks to `osbuild-pipeline` being run using `go run ...` we just need to generate manifests using the appropriate source code, but we don't need to have the RPMs built from it also installed on the system.

For CI testing, this should not be an issue, since the RPMs built from the tested source code are always installed on the system under test.

If you generate image test cases, e.g. using `./tools/test-case-generators/generate-all-test-cases ... remote ...`, the script by default installs latest versions of `osbuild` and `osbuild-composer` available in the distribution repositories. There is an option to define additional repos and these will have a higher priority. However there is a pretty unfortunate downside with `osbuild-composer`... we don't build RPMs for all architectures in our CI, for which we need to generate image test cases and test them (specifically `s390x` and `ppc64le`). This would be easily solved if we could use COPR repo which would contain the latest upstream builds (e.g. as it is done by PackIt for `osbuild` repository). However we don't have this yet for `osbuild-composer` and until the `osbuild-composer` RPMs with `dnf-json` as a service would get to Fedora 34/35 updates stable repo, generating image test cases for all architectures from the git `main` would be extremely harder.

Therefore if we go this way (having `dnf-json` only as a service), I propose that we first enable RPM builds for all architectures for Fedora using PackIt in osbuild-composer repository.

lavocatt:
> Isn't this a SELinux issue? IIRC SELinux prevents root processes to access home directories. A solution might be to run an ad-hoc HTTP server, I suggest something like `python -m http.server`, or a simple Go thing.

In fact, it's because `dnf-json` has a private `/tmp`, so when it tries to read the files at the path generated by the test it fails.
Using an `http.FileServer` in go makes the trick, good idea!

Gundersanne:
So dnf-json as a daemon does break composer as a standalone container for anything that expects the depsolving to happen in composer itself (so koji and the weldr api). This is the reason that the old-worker-new-composer regression test is broken as well, since there the composer PR version is running in a container and exposing the weldr api, but of course the dnf-json daemon isn't running in the container because we don't have systemd in the container.

We could expand `containers/osbuild-composer/entrypoint.py` to include dnf-json now, and invoke it the same way as we do weldr etc....

Gundersanne:
So I finally managed to get the regression test to work https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1841343046, but with a hack https://github.com/osbuild/osbuild-composer/pull/2085/commits/3612c14c2579d169ca3da6107df5c19006511c86#diff-0ffbc497a008289d8efbb1d7edc7781d4e0c3a1e72ab964bd3a3a9401dadae86R257 not sure if there's a better way.

The fds are shared between the container entrpoint process and all the child processes (so osbuild-composer which has a variable number of fds, and dnf-json). So we somehow need to get the correct fd into dnf-json, which should of course default to 3.

dnf-json
lavocatt: In the end we need to have this thing. Because sometime we want to be able to start dnf-json without systemd. This is the case when running in a container and for running the test cases generation.

thozza:
I had a look at https://github.com/osbuild/osbuild-composer/pull/2062/commits/8f39dd5c08291bf1973664647c7c593313fbad44 and think it won’t work with regenerating image test cases using `./tools/test-case-generators/generate-all-test-cases`. The reason is that the `dnf-json` installed on the system won’t be the one which creates a socket (until that version of composer gets to the distribution updates repos) and thus the RPMMD API from the “newer” sources copied over to the host won’t be able to connect to it.

I think that one solution could be to have a fallback in the RPMMD API to execute `dnf-json` as an executable if it is not listening on a socket, but I think this would be a bad approach as the code would be relevant for a limited time. So the alternative is IMO only to install the osbuild-composer version which includes `dnf-json` able to run as a service on the remote hosts when regenerating test cases. There are two possibilities:

1. Use PackIt and RPMs built in COPR when regenerating test cases by passing the COPR repo to the script. Passing the repo would be needed only for a limited time until the osbuild-composer version with `dnf-json` as a service would get to Fedora updates repo. 
2. Build the osbuild-composer RPMs on the remote host from copied sources and install them before regenerating test cases.

I'm in favor of (1), since having PackIt building composer RPMs for all arches would be definitely beneficial. (2) seems to me as adding additional potentially fragile step to generating test cases, but it could be a valid option. (2) would also require more coding as (1) is mostly about enabling PackIt and creating a YAML config for it.

ondrejbudai:
Note that COPR doesn't support s390x for el8 and both ppc64le and s390x for el9. For Fedora it supports all arches we need but that means we are stuck with Fedora for regenerating the test cases.

thozza:
> Note that COPR doesn't support s390x for el8 and both ppc64le and s390x for el9. For Fedora it supports all arches we need but that means we are stuck with Fedora for regenerating the test cases.

True.

We can ask COPR devels to enable these architectures, but s390x will be always only emulated 😞 .

But to be honest, I have been regenerating all image test cases on Fedora 34 for the past "forever" and it works great. There are sometimes issues which needed a workaround, but these show up in CI right away when a PR with regenerated test cases is proposed. So from my point of view, using only F34/F35 for regenerating test cases is not an issue...

thozza:
I implemented a workaround in the `generate-all-test-cases` script, which will allow the user to build the composer RPMs from sources on the runner and install them. Once merged, we should have a workaround for the transition period until composer RPMs with this change will get to the distribution repositories.

For more information see PR#2098

osbuild-composer.spec
```diff
@@ -307,6 +307,7 @@ cd $PWD/_build/src/%{goipath}
 %doc README.md
 %{_mandir}/man7/%{name}.7*
 %{_unitdir}/osbuild-composer.service
+%{_unitdir}/osbuild-dnf-json.socket
```
thozza, line 310: This is a packaging bug. The `.socket` unit should be part of the `dnf-json` sub-package, as the `.service` file is.

thozza (commented):
In general, this looks very good. I found one nitpick and a packaging bug. I'm still testing the `generate-all-test-cases` with this PR...

cmd/osbuild-dnf-json-tests/main_test.go
```diff
@@ -31,13 +36,13 @@ func TestFetchChecksum(t *testing.T) {
 
 	repoCfg := rpmmd.RepoConfig{
 		Name:      "repo",
-		BaseURL:   fmt.Sprintf("file://%s", dir),
+		BaseURL:   "http://localhost:9000",
```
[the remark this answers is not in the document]
lavocatt: The thing is, because the `dnf-json` service has its own private `/tmp` it could not access the files written at `dir` location. So I chose to keep the private `/tmp `and go this way for the test. I think the likelihood of a failure for the 9000 port is low, but I can enhance the log message in case the http server doesn't want to start to help a future debug.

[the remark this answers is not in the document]
thozza: Thanks for explanation. this is fine!

osbuild-composer.spec
lavocatt: I don't understand how `.spec` is structured yet. I reproduced what was already done in the file for the sub-package `worker` and applied it to `dnf-json`. Why is the `.socket` unit not with the sub-package `worker`, is this a mistake too ? Should I correct the two ?

thozza: No, those are fine. This is a bit misleading, but let me explain it.

In general the `.socket` unit starts a specific `.service` unit. The worker socket units are actually starting `osbuild-composer.service`, because it is the server's socket and the worker actually connects to it. So in this case this is correct. Worker does not listen on any socket, it connects to composer. In `dnf-json` case, the socket starts the service with the same base name and therefore it should be in the same binary RPM. In general everything would work just fine even now, because due to RPM dependencies the `dnf-json` sub-package will be pulled in when the main composer RPM is installed. But if you would install just the `dnf-json` sub-package, you would not be able to connect to it because systemd would not listen on the appropriate socket and would not start the service.

thozza:
I tested generating image test cases with the proposed branch and `dnf-json` service does not get started (I didn't expect it to, since there is no code in `internal/rpmmd/repository.go` which would start it. The generation fails with:
```
panic: Could not depsolve: Post "http://unix": dial unix /run/osbuild-dnf-json/api.sock: connect: no such file or directory

goroutine 1 [running]:
main.main()
        /root/tmp.VhHjpMVJIn/sources/cmd/osbuild-pipeline/main.go:127 +0x11c6
exit status 2
```

I think that the solution could be to simply extend the `tools/test-case-generators/generate-test-cases` script to check if the `osbuild-dnf-json.socket` is active and if not, start it. This should not be an issue, since the script must be run with root privileges as it runs `osbuild` and `image-info`...

osbuild-composer.spec
lavocatt: Thanks a lot for the explanation!

test/cases/generation.sh
```diff
@@ -0,0 +1,42 @@
...
+    setroubleshoot \
+    container-selinux \
+    lvm2
+
+# stop dnf-json socket
+sudo systemctl stop osbuild-dnf-json.socket
+
+# test the test cases generation
+WORKDIR=$(mktemp -d)
+mkdir -p "$WORKDIR"
```
thozza, line 34: This creates a temporary directory in `/tmp`, which is on `tmpfs` and based on the logs from the CI, there is not enough space for osbuild to store all of its data in the store directory. We usually use `/var/tmp`, which is on the disk and not in the RAM. I think that should solve the CI failure.

test/cases/generation.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# test the test case generation when dnf-json socket is stopped
+sudo ./tools/test-case-generators/generate-test-cases\
```
thozza, line 37: You should add
```
OSBUILD_LABEL=$(matchpathcon -n $(which osbuild))
chcon $OSBUILD_LABEL tools/image-info
```
before calling the generator script.

For more information, check the "important note" in https://github.com/osbuild/osbuild-composer/tree/main/test#image-tests

tools/test-case-generators/generate-test-cases
```diff
@@ -63,6 +67,16 @@ class TestCaseGenerator:
     def get_test_case(self, no_image_info, store):
         compose_request = json.dumps(self.test_case["compose-request"])
 
+        command = shlex.split(
+                "systemctl is-active --quiet osbuild-dnf-json.socket")
+        if not is_subprocess_succeeding(command):
+            print("osbuild-dnf-json.socket is not running, starting it up")
+            command = shlex.split(
+                "systemctl start osbuild-dnf-json.socket --quiet")
```
thozza, line 75: The same comment as above about splitting the string.

thozza (commented):
Thanks for adding the test case for generating an image test case. 😍  I think that once the path for the osbuild store is changed in the test case, the CI could be green and the PR should be good to go...

test/cases/generation.sh
```diff
@@ -0,0 +1,44 @@
...
+    go \
+    ranger \
+    openssl \
+    rpm-build \
+    aws \
+    qemu \
+    setools-console \
+    setroubleshoot \
+    container-selinux \
+    lvm2
```
thozza, line 27: We discussed this and I hoped that you'll eventually trim the list of packages only to those required for this test case to pass as there are definitely some unneeded packages. I'm not insisting that you do this in this PR, a follow up is fine by me. However if you are going to rebase this once more, you may as well clean it up.

I *think* that only `go` should be needed from the list. In addition, there is no reason to make dnf quiet by `-q` as we will at least see the installed package versions.

thozza (approved):
LGTM, thanks!

test/cases/generation.sh
lavocatt: Thanks @thozza Indeed we discussed it and I forgot to trim it. I'll do it in a follow up since I'm able to finally merge this PR. I'll keep you posted.

PR #2067, opened 2021-11-25
rpmmd: Reload subscriptions

Gundersanne:


cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -30,6 +32,16 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 		return err
 	}
 
+	for _, r := range args.Repos {
```
ondrejbudai, line 35: Why a loop?

I would even suggest making this dead simple and just creating a new RPPMD object for each job.

Gundersanne: well i thought only refresh in case the repos require rhsm

ondrejbudai (approved):

teg (approved):
makes sense to me!

PR #2068, opened 2021-11-25
Build a worker AMI using Schutzbot

ondrejbudai:
This PR takes the latest commit from https://github.com/osbuild/image-builder-packer (private repository) and integrates it in osbuild-composer repository. The packer configuration is then adjusted in several commits. Finally, the last commit adds a new step to Schutzbot - building an AMI that contains a worker built by that same pipeline.

This allows us to retire the image-builder-packer while making its content fully public and open-source, and to iterate faster because updating the workers will no longer require the boring "*update image-builder-packer*" step.

This pull request includes:


templates/packer/README.md
```diff
@@ -0,0 +1,6 @@
+# imagebuilder-packer
```
teg, line 1: This we can probably drop?

templates/packer/ansible/roles/common/files/monit_scripts/pozorbot_alert
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Send alerts to pozorbot.
```
teg, line 2: Is this something we could trigger from cloudwatch rather than from the instance?

templates/packer/ansible/roles/common/files/osbuild-composer-override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
```
teg, line 1: This we can drop.

templates/packer/ansible/roles/common/tasks/packages.yml
```diff
@@ -0,0 +1,76 @@
...
+      - jq
+      - osbuild-composer-worker
+      - unzip
+      - monit
+      - vector
+  register: result
+  retries: 5
+  until: result is success
+
+- name: Download AWS CLI installer
```
teg, line 53: If we need this, is it not available from a yum repo?

templates/packer/ansible/roles/common/tasks/packages.yml
```diff
@@ -0,0 +1,76 @@
...
+  yum_repository:
+    name: "osbuild"
+    description: "osbuild commit {{ OSBUILD_COMMIT }}"
+    baseurl: "http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com/osbuild/rhel-8-cdn/x86_64/{{ OSBUILD_COMMIT }}"
+    enabled: yes
+    gpgcheck: no
+    priority: "5"
+
+# We need EPEL for monit
+- name: Add EPEL
```
teg, line 22: Would be great if we can avoid non-RHEL content... I would say Amazon content would be ok, or anything that we have support for, but we should err on the side of caution.

Could monit and vector be replaced by https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Install-CloudWatch-Agent.html ?

templates/packer/ansible/roles/common/tasks/systemd.yml
```diff
@@ -0,0 +1,11 @@
+---
```
teg, line 1: No longer used.

teg (dismissed):
Thank you! I love this!
I did leave a lot of nitpicks, but I would be happy to merge this to reach parity and switch to using it and then iterate on that.

teg:
packer can't find jq

templates/packer/ansible/roles/common/tasks/packages.yml
Gundersanne: officially this is the way. Also v2 (which this link points to) is fully standalone.

templates/packer/ansible/roles/common/tasks/packages.yml
Gundersanne: Hm maybe as a standalone project/followup? looks pretty neat tho

templates/packer/ansible/roles/common/files/monit_scripts/verify_worker_connection
```diff
@@ -1,28 +0,0 @@
...
-
-# Exit now if this is not a worker.
-if [[ -z "${COMPOSER_HOST}" ]]; then
-  echo "Not a worker. Skipping check."
-  exit 2
-fi
-
-# Test a connection to composer.
-CONNECTION_TEST=$(
-  curl -s --connect-timeout 5 \
```
Gundersanne: so theoretically this would be pretty easy, we could just curl api.(stage.)openshift.com/....composer/openapi every now and then. No auth required, if that's up the service is up.

teg: Could it be enough to simply make sure we log sufficiently when the worker fails?

Gundersanne: Yea, that already happens, /jobs request will log `Requesting job failed` (lvl error).

templates/packer/README.md
ondrejbudai: Done.

templates/packer/ansible/roles/common/files/osbuild-composer-override.conf
ondrejbudai: Of course, I thought this was for worker but nah, ofc it wasn't.

templates/packer/ansible/roles/common/tasks/systemd.yml
ondrejbudai: Removed.

templates/packer/ansible/roles/common/tasks/packages.yml
ondrejbudai: Yeah, I would honestly follow the official docs when possible.

Gundersanne (dismissed):
\o/

ondrejbudai:
About monit and vector: I would love to get rid of both of them at some point. I suggest just merging this firstly, and do any changes in a follow-up.

Ideally, we should split logging and metrics&alerting:

I hope that logging can be done using the official CloudWatch Agent but last time I tried, it wasn't possible to just send the whole journal. I will investigate the options though. Vector is cool but I agree that having no 3rd party software is very nice.

For metrics&alerting, I'm not sure yet, this is something that we need to discuss with AppSRE (how to get metrics from workers to Prometheus).

ondrejbudai:
I had a minor issue in the playbook, it should be now fixed.

teg (approved):

PR #2073, opened 2021-11-29
enable gosec tool

jabia99:
Gosec is a security tool for golang. It's already enabled in image-builder. This PR will enable this tool and solve the vulnerability issues it finds.

Gosec available rules: https://github.com/securego/gosec#available-rules 

This pull request includes:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,10 +3,15 @@ linters-settings:
     disable:
       - shadow # default value recommended by golangci
       - composites
+  
+linters:
+  enable:
+    - gosec
 
 run:
   build-tags:
     - integration
+  timeout: 5m
 
 issues:
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
--- a/cmd/osbuild-auth-tests/main_test.go
+++ b/cmd/osbuild-auth-tests/main_test.go
@@ -42,6 +42,7 @@ func createTLSConfig(config *connectionConfig) (*tls.Config, error) {
 	return &tls.Config{
 		RootCAs:      roots,
 		Certificates: []tls.Certificate{cert},
+		MinVersion:   tls.VersionTLS12,
 	}, nil
 }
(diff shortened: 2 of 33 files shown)


.golangci.yml
```diff
@@ -3,14 +3,27 @@ linters-settings:
...
+  timeout: 5m
 
 issues:
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
   max-issues-per-linter: 0
 
   # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
   max-same-issues: 0
+
+  exclude-rules:
```
jabia99, line 23: "excludes:" under linters-settings didn't work for some reason ([gosec config](https://golangci-lint.run/usage/linters/#gosec)). So i had to use text based exclusion

ondrejbudai: I would honestly prefer doing local excludes over global excludes. Feels safer.

.golangci.yml
```diff
@@ -3,14 +3,27 @@ linters-settings:
...
+    - text: "G404:" # Ignore weak random number generator lint. We do not need strong randomness here.
+      linters:
+        - gosec
+    - text: "G601:" # Ignore Implicit memory aliasing in for loop
```
ondrejbudai, line 27: Did you check that loops that break this rule aren't buggy? They feel suspicious to me.

ondrejbudai (changes requested):
Please, can you do local excludes? I think it's very valuable to have all the rules enabled for new code.

.golangci.yml
jabia99: I checked some of them but not all. I seemed to me that unintended bugs would be really noticeable in this case.

jabia99: I will do local excludes in every case and review the loops

atodorov (approved):

PR #2081, opened 2021-11-30
Cloud API: Support more image types as S3 objects

achilleas-k:
## More blobby image types in the cloud

This PR adds support for the following image types to the cloud API:
- qcow2
- vmdk
- image-installer

All three are uploaded to the configured s3 bucket and a presigned URL is provided to the user for download (blobby image types).

## Tests

The test script api.sh has been changed to take an image type as argument instead of a target (cloud service).  All supported image types are built, except the `edge-installer`.  Only some blobby image types are verified.

### qcow2 and vmdk

The files are downloaded and run through the `image-info` tool.  The image info is checked to verify that the users specified in the request were created and the package (postgresql) is installed.

### edge-commit

Same as before this PR: The tarball is downloaded and extracted and the commit ID is verified.

## Test runners

I removed the Fedora and CentOS 8 runners for the API test.  I don't know if it's worth testing on those platforms.  I think it's a bit redundant since the API doesn't really do anything different between distros and the actual image builds for each distro are tested elsewhere.  The important part to test with respect to the API I think is the propagation of compose request options down to the manifest generation and the retrieval of accurate info from the API like build status and metadata.

I'm keeping the RHEL 8.4 variant because that does have different distro behaviour caused by the change in osbuild result formats.  It's probably also redundant to test both on 8.6 and 9.0, but I kept them for now.

------

This pull request includes:


internal/cloudapi/v2/v2.go
```diff
@@ -363,6 +370,7 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 
 		irTarget = t
 	default:
```
teg, line 372: Is this a left-over? Should use the logger if we want this, no?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -507,11 +507,14 @@ components:
...
       enum:
         - aws
-        - gcp
         - azure
         - edge-commit
         - edge-container
         - edge-installer
+        - gcp
+        - image-installer
+        - qcow2
```
teg, line 516: Call this `guest-image`?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -507,11 +507,14 @@ components:
...
+        - vmdk
```
teg, line 517: This is for vSphere? Call in `vsphere`?

teg (changes requested):
Looking great. Though, could we name the image types after their usecase/footprint, rather than file type? I made some suggestions, but maybe @gicmo and @thozza also have thoughts?

internal/cloudapi/v2/openapi.v2.yml
thozza: In RHEL it is called `rhel-guest-image` and since we dropped the `rhel-` prefix from images, the suggestion looks good to me.

internal/cloudapi/v2/v2.go
achilleas-k: oh no, sorry.  Leftover print debugging indeed.

internal/cloudapi/v2/openapi.v2.yml
Gundersanne: agreed, should refer to the targetplatform rather than the on disk format of the image

internal/cloudapi/v2/v2.go
```diff
@@ -165,6 +165,7 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 	}
 	imageType, err := arch.GetImageType(imageTypeFromApiImageType(ir.ImageType))
 	if err != nil {
+		println("image type from api image")
```
Gundersanne, line 168: `ctx.Logger()` or `logrus.` if you want log statement; though I'm guessing this is leftover debugging.

Also, there's a `println` in go?!

achilleas-k:
Thanks for the reviews.
I changed the names based on Tom's suggestions (in code and the gitlab-ci config) and cleaned up the embarrassing leftover `print`s.

test/cases/api.sh
```diff
@@ -704,7 +737,7 @@ sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SEL
 # NOTE: The rest of the job data can contain sensitive information
 #
 # Suppressing shellcheck.  See https://github.com/koalaman/shellcheck/wiki/SC2024#exceptions
-sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SELECT result->>'Manifest' FROM jobs" | sudo tee "${ARTIFACTS}/manifest.json"
+sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SELECT result FROM jobs WHERE type='manifest-id-only'" | .data | sudo tee "${ARTIFACTS}/manifest.json"
```
thozza, line 740: This line contains an error `... WHERE type='manifest-id-only'" | .data | sudo ...` producing `/usr/libexec/tests/osbuild-composer/api.sh: line 740: .data: command not found`

thozza (commented):
There is one error / typo in the `api.sh` on line 740 which makes the test case always fail in CI.

Otherwise this PR is pure awesomeness 😍  I love your modifications to `api.sh`, it is much better now and easy to extend for other S3-based image types. Also I like that you used `image-info` 😅

test/cases/api.sh
```diff
@@ -703,8 +736,10 @@ sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SEL
 # Save the Manifest from the osbuild-composer store
 # NOTE: The rest of the job data can contain sensitive information
 #
-# Suppressing shellcheck.  See https://github.com/koalaman/shellcheck/wiki/SC2024#exceptions
-sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SELECT result->>'Manifest' FROM jobs" | sudo tee "${ARTIFACTS}/manifest.json"
+# Disable -x for this command to avoid spewing the whole result and manifest into the log
+set +x
+sudo podman exec osbuild-composer-db psql -U postgres -d osbuildcomposer -c "SELECT result FROM jobs WHERE type='manifest-id-only'" | jq .data | sudo tee "${ARTIFACTS}/manifest.json"
```
Gundersanne, line 741: straight up piping to `jq` won't work because you have all the SQL stuff around it.

Maybe 

`SELECT result->>'data' FROM jobs WHERE type='manifest-id-only' | sudo tee.. `

It still won't b proper json though; or maybe using `psql -t` ; `-t` is supposed to surpress the psql header/footer.

ondrejbudai:
I'm not sure what to do about the test matrix, I don't see much value in testing `edge-*`, `guest-image` and `image-installer` image types again. I personally think that testing of uploading to S3 just once is fine.

test/cases/api.sh
achilleas-k: yeah, that was a bit silly.  I decided instead to just dump the result into the artifacts.

Gundersanne:
Still needs a rebase because of the 8.5 runners :( sorry

achilleas-k:
> I'm not sure what to do about the test matrix, I don't see much value in testing `edge-*`, `guest-image` and `image-installer` image types again. I personally think that testing of uploading to S3 just once is fine.

I think I agree.  There's a couple of things I feel should be covered by tests when it comes to the API:
1. **Options/customisations are properly propagated to the manifest for creating the image.**  This isn't covered by, for example, the weldr build tests because the cloud API options follow different code paths.  However, one image type should be enough to show that options are getting through.  Ideally, we could run a test that makes a request, doesn't actually build the artifact or do the upload, and instead produces a manifest that can be directly inspected for customisataions.
2. **All image types that should be supported by the API are enabled and work.**  This is covered by unit tests.  I think.

Building the `edge-commit` is good for checking that we get the right commit ID back from the metadata.  But it's not easy to check if it gets the right user options.

We could keep it as is for now for a while and then I can convert to the manifest checking test.  Perhaps a whole separate script that runs on all image types but doesn't build them.  Or I can slim it down now.

thozza (dismissed):

teg (approved):

.gitlab-ci.yml
```diff
@@ -304,11 +304,15 @@ Integration:
...
+  IMAGE_TYPE:
     - aws
+    - azure
+    - edge-commit
+    - edge-container
     - gcp
-    - aws.s3
+    - image-installer
+    - guest-image
+    - vsphere
```
Gundersanne, line 315: The point that Ondřej made is still valid I think, isn't it enough to just test 1 aws.s3 target?

achilleas-k:
Removed the new image types from the test matrix as discussed.
I kept the support for the image types in the test script in case we want to adapt them in the future.

thozza (approved):
Still awesome 😉

gicmo:
fwiw, lgtm

Gundersanne (approved):
poifect

ondrejbudai (approved):

teg (approved):

PR #2093, opened 2021-12-03
Prepare CI to build -tests RPM for downstream testing from source

atodorov:
I had just realized that building osbuild-composer-tests rpm from main is sometimes going to result in a different NVR than we have downstream (this will be more prominent toward the end of a minor release cycle).  Because we've got 

```
Requires:   %{name} = %{version}-%{release}
```

installing this RPM inside our test systems will fail with broken dependencies. Another possibility is that this will result in silently pulling a newer osbuild-composer RPM from the S3 repos instead of pulling the RPM from the nightly tree.

### DIFF
--- a/schutzbot/mockbuild.sh
+++ b/schutzbot/mockbuild.sh
@@ -99,10 +99,15 @@ sudo mock -r "$MOCK_CONFIG" --buildsrpm \
   --sources "./osbuild-composer-${COMMIT}.tar.gz" \
   --resultdir ./srpm
 
+if [ "${NIGHTLY:=false}" == "true" ]; then
+    RELAX_REQUIRES="--with=relax_requires"
+fi
+
 greenprint "🎁 Building RPMs"
 sudo mock -r "$MOCK_CONFIG" \
     --define "commit ${COMMIT}" \
     --with=tests \
+    ${RELAX_REQUIRES:+"$RELAX_REQUIRES"} \
     --resultdir "$REPO_DIR" \
     srpm/*.src.rpm
 
@@ -111,6 +116,12 @@ sudo chown -R "$USER" "${REPO_DIR%%/*}"
 
 greenprint "🧹 Remove logs from mock build"
 rm "${REPO_DIR}"/*.log
+
+# leave only -tests RPM to minimize interference when installing
+# osbuild-composer.rpm from distro repositories
+if [ "${NIGHTLY:=false}" == "true" ]; then
+    find "${REPO_DIR}" -type f -not -name "osbuild-composer-tests*.rpm" -exec rm -f "{}" \;
+fi
 
 # Create a repo of the built RPMs.
 greenprint "⛓️ Creating dnf repository"
--- a/schutzbot/mockbuild.sh
+++ b/schutzbot/mockbuild.sh
@@ -38,6 +38,9 @@ fi
 # Relative path of the repository – used for constructing both the local and
 # remote paths below, so that they're consistent.
 REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}
+if [ "${NIGHTLY:=false}" == "true" ]; then
+    REPO_PATH=nightly/${REPO_PATH}
+fi
 
 # Directory to hold the RPMs temporarily before we upload them.
 REPO_DIR=repo/${REPO_PATH}
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -238,6 +238,7 @@ function cleanupAzure() {
   # since this function can be called at any time, ensure that we don't expand unbound variables
   AZURE_CMD="${AZURE_CMD:-}"
   AZURE_IMAGE_NAME="${AZURE_IMAGE_NAME:-}"
+  AZURE_INSTANCE_NAME="${AZURE_INSTANCE_NAME:-}"
 
   # do not run clean-up if the image name is not yet defined
   if [[ -n "$AZURE_CMD" && -n "$AZURE_IMAGE_NAME" ]]; then
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -498,6 +498,8 @@ SCHEDULED_CLOUD_CLEANER:
 SonarQube:
   stage: test
   extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule"'
   script:
     - schutzbot/sonarqube.sh
   variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -380,8 +380,8 @@ API:
   extends: .terraform
   rules:
     - if: '$CI_PIPELINE_SOURCE != "schedule"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
+    # note: cloud API is not supported for on-prem installations so
+    # don't run this test case for nightly trees
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/api.sh ${IMAGE_TYPE}
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,6 +38,15 @@ init:
   script:
     - schutzbot/update_github_status.sh start
 
+
+.RPM_RUNNERS_RHEL: &RPM_RUNNERS_RHEL
+  RUNNER:
+    - aws/rhel-8.6-nightly-x86_64
+    - aws/rhel-8.6-nightly-aarch64
+    - aws/rhel-9.0-nightly-x86_64
+    - aws/rhel-9.0-nightly-aarch64
+  INTERNAL_NETWORK: ["true"]
+
 RPM:
   stage: rpmbuild
   extends: .terraform
@@ -63,12 +72,26 @@ RPM:
           - aws/rhel-8.5-ga-aarch64
           - aws/centos-stream-8-x86_64
           - aws/centos-stream-8-aarch64
-      - RUNNER:
-          - aws/rhel-8.6-nightly-x86_64
-          - aws/rhel-8.6-nightly-aarch64
-          - aws/rhel-9.0-nightly-x86_64
-          - aws/rhel-9.0-nightly-aarch64
-        INTERNAL_NETWORK: ["true"]
+      - <<: *RPM_RUNNERS_RHEL
+
+Build -tests RPM for RHEL:
+  stage: rpmbuild
+  extends: .terraform
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
+  script:
+    - sh "schutzbot/mockbuild.sh"
+  interruptible: true
+  after_script:
+    - schutzbot/update_github_status.sh update
+    - schutzbot/save_journal.sh
+  artifacts:
+    paths:
+      - journal-log.gpg
+  parallel:
+    matrix:
+      - <<: *RPM_RUNNERS_RHEL
 
 Container:
   stage: rpmbuild
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -408,6 +408,7 @@ Requires:   libvirt-daemon-driver-storage-disk
 Requires:   libvirt-daemon-kvm
 Requires:   qemu-img
 Requires:   qemu-kvm
+Requires:   rpmdevtools
 Requires:   virt-install
 Requires:   expect
 Requires:   python3-lxml
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -2,6 +2,11 @@
 # Pass --with tests to rpmbuild to override
 %bcond_with tests
 
+# When --with relax_requires is specified osbuild-composer-tests
+# will require osbuild-composer only by name, excluding version/release
+# This is used internally during nightly pipeline testing!
+%bcond_with relax_requires
+
 %global goipath         github.com/osbuild/osbuild-composer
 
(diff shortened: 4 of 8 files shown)


ondrejbudai:
FYI I found this when finding the cause of this issue: https://unix.stackexchange.com/questions/415990/how-can-i-expand-a-quoted-variable-to-nothing-if-its-empty

atodorov:
```
ðŸƒðŸ» Running test: osbuild-dnf-json-tests
=== RUN   TestFetchChecksum
    main_test.go:47: 
        	Error Trace:	main_test.go:47
        	Error:      	Expected nil, but got: &url.Error{Op:"Post", URL:"http://unix", Err:(*net.OpError)(0xc000064640)}
        	Test:       	TestFetchChecksum
        	Messages:   	Failed to fetch checksum: Post "http://unix": dial unix /run/osbuild-dnf-json/api.sock: connect: no such file or directory
    main_test.go:48: 
        	Error Trace:	main_test.go:48
        	Error:      	Should not be: ""
        	Test:       	TestFetchChecksum
        	Messages:   	The checksum is empty
--- FAIL: TestFetchChecksum (0.02s)
```

@ondrejbudai ^^^ any idea what may be causing this ?

atodorov:
Nightly pipeline results:
https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/444458924

Gundersanne (commented):
Maybe I'm not well placed to review this, I do have a question. Won't this overwrite existing existing test rpms in the rpmrepo s3 bucket whenever the nightly pipeline runs?

atodorov:
> Maybe I'm not well placed to review this, I do have a question. Won't this overwrite existing existing test rpms in the rpmrepo s3 bucket whenever the nightly pipeline runs?
> 
> edit: actually s3cmd put doesn't overwrite files, so the resulting test rpms are just lost?

I don't follow here but we've got 
`REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}`

and given that we'll be building the RPMs on the same distro/arch/commit combo nothing should be lost. The RPMs will be overwritten with newly built ones which should be functionally the same as the old ones (minus the dependency on osbuild-composer-tests). 

@Gundersanne  does this answer your question?

jrusz (changes requested):
I've tried a test run with a schedule pointing to this PR to test it and it ended up installing the `osbuild-composer` package from the S3 repository instead of the nightly one. This is the run https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/449524874 

In the mockbuild stage it said that the repository already exists so it skipped it. I guess this then resulted with the rpms being built with the `relax-requires` and that caused the `osbuild-composer` package to get installed from the S3 repository even though the nightly has higher priority? Or did DNF just figure out there is a newer version somewhere else and didn't respect the priority? We need to investigate this.

jrusz:
Yeah I think it's because when the regular pipeline ran for this PR it created the S3 repository with `osbuild-composer` rpms without the `relax-requires` flag and when the nightly pipeline ran the mockbuild stage there figured the repository already exists so it got skipped and then when `deploy.sh` is ran it will install `osbuild-composer-tests` package without the flag resulting in pulling the `osbuild-composer` dependence from the S3 repo even though the nightly should have priority. 

I suggest we create different S3 repository for the nightly testing, put something like `-nightly` in the path or anything to distinguish it.

atodorov:
> Yeah I think it's because when the regular pipeline ran for this PR it created the S3 repository with `osbuild-composer` rpms without the `relax-requires` flag and when the nightly pipeline ran the mockbuild stage there figured the repository already exists so it got skipped and then when `deploy.sh` is ran it will install `osbuild-composer-tests` package without the flag resulting in pulling the `osbuild-composer` dependence from the S3 repo even though the nightly should have priority.


I think this is correct. I've seen the issue a few times during development but I was usually running the scheduled job first and then the regular job and I was thinking that DNF messes up priorities somehow. Pushed a change which should resolve this, let's see.

schutzbot/mockbuild.sh
```diff
@@ -37,7 +37,7 @@ fi
 
 # Relative path of the repository – used for constructing both the local and
 # remote paths below, so that they're consistent.
-REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}
+REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}/${NIGHTLY:=false}
```
jrusz, line 40: This will add either `true` or `false` to the repo path, is that something we want or is this just for testing now?

jrusz: Yeah the tests don't seem to like that... What about adding a check whether it's nightly or not to line 50 and in case `NIGHTLY == true` print a message that this is nightly run and override the repository?

atodorov: > This will add either true or false to the repo path, is that something we want or is this just for testing now?

Does it hurt anything? I don't think so. Updated the URL to be a bit more verbose.

> Yeah the tests don't seem to like that... 

b/c I forgot to update the path in the .repo file. It has nothing to do with the actual URL itself.

> What about adding a check whether it's nightly or not to line 50 and in case NIGHTLY == true print a message that this is nightly run and override the repository?

Which is exactly the same thing as the change above: override the repository URL based on which pipeline we're using for execution.

schutzbot/mockbuild.sh
```diff
@@ -37,7 +37,7 @@ fi
 
 # Relative path of the repository – used for constructing both the local and
 # remote paths below, so that they're consistent.
-REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}
+REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}/nightly-${NIGHTLY:=false}
```
Gundersanne, line 40: Can we use ` REPO_PATH=osbuild-composer/${DISTRO_VERSION}/${ARCH}/${COMMIT}` as a default and ` REPO_PATH=osbuild-composer/nightly/${DISTRO_VERSION}/${ARCH}/${COMMIT}` for the nightlies? There might be references to these repos in other places and those would all break.

Gundersanne: Or `osbuild-composer-nightlies/${DIST....`

Gundersanne (dismissed):
lgtm

jrusz (changes requested):
When ran in a nightly schedule it fails on setting up `osbuild` repository. It puts the `/nightly` in the path but such a repository does not exist. Example: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1991876773

I suggest adding a condition to exclude adding the `/nightly` in repo path for `osbuild` in the `deploy.sh` script. Or don't use the `osbuild` repo in nightly testing at all, we want to run the tests with the one coming from nightly trees anyway right?

jrusz:
Ugh, this is dumb, there are actually two variables.... `project` and `PROJECT` ... `PROJECT` is set on line 10 in `deploy.sh` and is apparently always `osbuild-composer` and `project` is set on line 39... Your new condition is checking for `PROJECT` variable while it should be checking for `project`... Failed scheduled run https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1995499313

jrusz (dismissed):
Amazing, the scheduled pipeline works as expected, for example the missing upgrade test is now run and passing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1996443926 . LGTM!

atodorov:
> Amazing, the scheduled pipeline works as expected, for example the missing upgrade test is now run and passing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1996443926 . LGTM!

Any idea why this is different from https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/452468375 ? 

Manifests and Upgrade sections are missing.

jrusz:
> > Amazing, the scheduled pipeline works as expected, for example the missing upgrade test is now run and passing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1996443926 . LGTM!
> 
> Any idea why this is different from https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/452468375 ?
> 
> Manifests and Upgrade sections are missing.

We agreed that Upgrade test from RHEL-8 to RHEL-9 will only run during RHEL-9 pipeline so that's fine and idk about the Manifest test. Looking at the `.gitlab-ci.yml` file it's being only ran on rhel-9 runner

atodorov:
> > > Amazing, the scheduled pipeline works as expected, for example the missing upgrade test is now run and passing https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1996443926 . LGTM!
> > 
> > 
> > Any idea why this is different from https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/452468375 ?
> > Manifests and Upgrade sections are missing.
> 
> We agreed that Upgrade test from RHEL-8 to RHEL-9 will only run during RHEL-9 pipeline so that's fine and idk about the Manifest test. Looking at the `.gitlab-ci.yml` file it's being only ran on rhel-9 runner


FTR the job which you are posting results to is running against 9.0 while mine is running against 8.6. Hence the difference in results.

atodorov:
@Gundersanne this is good to be merged. The remaining failing tests seem unrelated.

jrusz:
I think I figured out what's the problem with api tests on RHEL-8.6 ... RHEL-8.6 is still using `osbuild-composer-40` while RHEL-9.0 uses `osbuild-composer-42`. When tested locally on RHEL-8.6 I got pretty much the same failure as seen in CI, it's just taking longer to fail but journal shows the same errors. So we either "waive" it or get `osbuild-composer-42` merged into RHEL-8.6 ASAP...

atodorov:
> I think I figured out what's the problem with api tests on RHEL-8.6 ... RHEL-8.6 is still using `osbuild-composer-40` while RHEL-9.0 uses `osbuild-composer-42`. When tested locally on RHEL-8.6 I got pretty much the same failure as seen in CI, it's just taking longer to fail but journal shows the same errors. So we either "waive" it or get `osbuild-composer-42` merged into RHEL-8.6 ASAP...

This doesn't actually state what the problem is. Did we introduce version specific code in `api.sh` ? If yes, which part of the code exactly is that ? There's a mechanism for conditional execution, I just need to know which parts.

jrusz:
> > I think I figured out what's the problem with api tests on RHEL-8.6 ... RHEL-8.6 is still using `osbuild-composer-40` while RHEL-9.0 uses `osbuild-composer-42`. When tested locally on RHEL-8.6 I got pretty much the same failure as seen in CI, it's just taking longer to fail but journal shows the same errors. So we either "waive" it or get `osbuild-composer-42` merged into RHEL-8.6 ASAP...
> 
> This doesn't actually state what the problem is. Did we introduce version specific code in `api.sh` ? If yes, which part of the code exactly is that ? There's a mechanism for conditional execution, I just need to know which parts.

Looking at the difference between `v40` and `v42` I don't see anything version specific in `api.sh` but there is added code which  appears to now be part of regular execution... 

There are 97 commits between those two versions I suppose some behavior has changed but I don't think I can find out exactly what, that's why I suggested we need a newer version because that seems to just work...

This also led me to a question, what if we introduce some new `osbuild-composer` functionality upstream, add tests for it, the functionality won't yet be available downstream but the test will. This will just cause the nightly pipeline to fail. But I suppose we can solve this once such a problem appears, just something I wanted to mention.

atodorov:
> This also led me to a question, what if we introduce some new `osbuild-composer` functionality upstream, add tests for it, the functionality won't yet be available downstream but the test will. This will just cause the nightly pipeline to fail. But I suppose we can solve this once such a problem appears, just something I wanted to mention.

This scenario should be solved similar as in 5f8f38237356278bc5d41ed35610208f38e2a53c

test/cases/api.sh
```diff
@@ -508,21 +510,35 @@ fi
...
+EXTRA_PACKAGES_JSON=""
+EXTRA_PACKAGE_NAMES=""
+if nvrGreaterOrEqual "osbuild-composer" "41"; then
+  PAYLOAD_REPOSITORIES_JSON=$(cat <<EndOfMessage
+    "payload_repositories": [
+      {
+        "baseurl": "${PAYLOAD_REPO_URL}"
+      }
+    ],
+EndOfMessage
```
Gundersanne, line 523: I don't like the approach of having version conditionals all over our tests. Either skip the api tests entirely in nightly pipelines (the service-facing api isn't supported on-prem anyways). Or somehow install the exact test rpms belonging to the exact nightly version.

Gundersanne (changes requested):

test/cases/api.sh
atodorov: > Or somehow install the exact test rpms belonging to the exact nightly version.

That's a discussion we've had two months ago and decided to take the current approach to unblock nightly tree testing and make it easier to add to the test suite without the need to go through an approval process and the need to rebase downstream RPMs in order to pick up the new tests.

It's either having to rebase RPMs in order to include new tests or having test conditionals. This PR alone introduces 2 of them. 


I am not in favor of removing tests without the appropriate analysis, even less so delegating this task to a different testing & CI pipeline (IQE). 

The API tests themselves were under debate long time ago but that didn't come to a definitive conclusion. It is a valid point that we need to review these tests and do 2 things: 
- make it clear what is the subject under test (b/c from your comment above it seems like these tests should have never been introduced in the repository or that they are testing multiple items at the same time or both)
-  decide which part goes where and split the work in appropriate chunks.

However this discussion should not be blocking making further progress on the existing nightly pipeline b/c  we risk reaching 8.6 GA without being able to properly test downstream.

Gundersanne: Is it impossible to get the test rpms which match the exact commit in the nightlies? Because doing it this will make our tests unmaintainable very quickly.

atodorov: > Is it impossible to get the test rpms which match the exact commit in the nightlies? Because doing it like this will make our tests unmaintainable very quickly.

We're trying to solve a different problem here. 

We already have osbuild-composer-tests which is built from the same sources as the osbuild-composer.rpm which goes into nightly trees. That's how things work before this PR. And without this PR we have a chicken-and-egg problem b/c every new test needs to first get into a new version released upstream, then rebased and released downstream and any changes which may be necessary to accommodate the nightly pipeline are usually discovered afterwards and need to go through the standard bugzilla approval process. 

That's why we've taken the approach to disconnect the test suite from the software under test when it comes to running the nightly CI pipelines.

Gundersanne (commented):
`0d7a0be2bc57ad3d2f32bc75f9530394d174aaf3` can be dropped no? Otherwise looks good.

atodorov:
> `0d7a0be2bc57ad3d2f32bc75f9530394d174aaf3` can be dropped no?

Not yet b/c on 8.6 we still have v40.

Gundersanne:
> `0d7a0be2bc57ad3d2f32bc75f9530394d174aaf3`

Right but we're not running the api.sh tests in the nightly pipeline  after https://github.com/osbuild/osbuild-composer/pull/2093/commits/287c115c8db2ab91b6a1591b606ac110c4db4d2a

Gundersanne (approved):
lgtm! thank you

PR #2095, opened 2021-12-03
composer: Only set queue and artifact dir for fsqueue

Gundersanne:
When backed by a DB, composer has no need of a queue or artifact directory.
    
This also addresses "Error moving artifacts for job" logging noise when using a dbqueue.

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -53,14 +53,13 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Compos
 		cacheDir: cacheDir,
 	}
 
-	queueDir, err := c.ensureStateDirectory("jobs", 0700)
-	if err != nil {
-		return nil, err
-	}
-
-	artifactsDir, err := c.ensureStateDirectory("artifacts", 0755)
-	if err != nil {
-		return nil, err
+	var err error
+	artifactsDir := ""
+	if config.Worker.EnableArtifacts {
+		artifactsDir, err = c.ensureStateDirectory("artifacts", 0755)
+		if err != nil {
+			return nil, err
+		}
 	}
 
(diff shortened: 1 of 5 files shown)


teg (commented):
Makes sense, but feels like a layering violation to make it conditional on no DB. For the jobs dir this obviously makes sense if we have the fs backend, but for artefacts, how is the logic done that disables them for the db backend?

Gundersanne:
> Makes sense, but feels like a layering violation to make it conditional on no DB. For the jobs dir this obviously makes sense if we have the fs backend, but for artefacts, how is the logic done that disables them for the db backend?

Yea you're right, it really is service vs on prem. at least the tests pass so that's ok, i'll figure out how to make this better

teg (commented):
this will disable artifacts dir by default, that's not backwards compatible, is it?

Gundersanne:
> this will disable artifacts dir by default, that's not backwards compatible, is it?

should default to https://github.com/osbuild/osbuild-composer/pull/2095/files#diff-93349275b33e6f3c25d6ae24f48c50991dd6ca9db01e5c5c7106f76cfe4a1582R96

teg:
> > this will disable artifacts dir by default, that's not backwards compatible, is it?
> 
> should default to https://github.com/osbuild/osbuild-composer/pull/2095/files#diff-93349275b33e6f3c25d6ae24f48c50991dd6ca9db01e5c5c7106f76cfe4a1582R96

Ah, I can't read. Indeed, that's ok.

> But not sure if I like specifying a relative directory as a config option :|

Make it a boolean? But use it to enable/disable artifacts handling as a whole? Now it is conditional on the weldr api.

teg (dismissed):
Looks good!

ondrejbudai:
Needs a rebase

teg (dismissed):
Perfect!

teg (approved):

PR #2099, opened 2021-12-06
tests: Fix several permission and koji failures

jrusz:
This pull request includes:

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -118,11 +118,11 @@ get_compose_metadata () {
 
     # Find the tarball and extract it.
     TARBALL=$(basename "$(find . -maxdepth 1 -type f -name "*-metadata.tar")")
-    tar -xf "$TARBALL"
-    rm -f "$TARBALL"
+    sudo tar -xf "$TARBALL"
+    sudo rm -f "$TARBALL"
 
     # Move the JSON file into place.
-    cat "${COMPOSE_ID}".json | jq -M '.' | tee "$METADATA_FILE" > /dev/null
+    sudo cat "${COMPOSE_ID}".json | jq -M '.' | tee "$METADATA_FILE" > /dev/null
 }
 
 # Get the console screenshot from the AWS instance.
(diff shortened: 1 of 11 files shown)


henrywang:
@jrusz This permission issue should be related with new `weldr-cli`, which replaced `composer-cli` in RHEL 8.6 latest nightly compose.
`sudo composer-cli compose metadata "$COMPOSE_ID" > /dev/null` generates metadata tar file with `root` user permission.
My fix on `ostree.sh`:
```
diff --git i/ostree.sh w/ostree.sh
index 4b9328c..edc1b65 100755
--- i/ostree.sh
+++ w/ostree.sh
@@ -225,11 +225,11 @@ get_compose_metadata () {

     # Find the tarball and extract it.
     TARBALL=$(basename "$(find . -maxdepth 1 -type f -name "*-metadata.tar")")
-    tar -xf "$TARBALL" -C "${TEMPDIR}"
-    rm -f "$TARBALL"
+    sudo tar -xf "$TARBALL" -C "${TEMPDIR}"
+    sudo rm -f "$TARBALL"

     # Move the JSON file into place.
-    cat "${TEMPDIR}"/"${COMPOSE_ID}".json | jq -M '.' | tee "$METADATA_FILE" > /dev/null
+    sudo cat "${TEMPDIR}"/"${COMPOSE_ID}".json | jq -M '.' | tee "$METADATA_FILE" > /dev/null
 }

 # Build ostree image.
```
This fix also works with the other shell script which needs generate metadata file.

jrusz:
Hi @henrywang ,
this was part of my investigation about what happened and it is indeed because of a new `weldr-client`. We didn't see this in RHEL-9 yet because it's using an older version than the one that just got into RHEL-8.6. The permission change is intended as the logs and metadata can contain sensitive information apparently. The fix you're suggesting is probably the easiest fix and is what I'm going to do, I'll update all the affected test scripts in this PR.

ondrejbudai (dismissed):
Thanks for looking into this!

jrusz:
Apparently we also have issues with koji :/ ... It seems that it started happening from version 1.27 as the last passing test I could find is using 1.26 but I don't see any obvious change in the changelog that could have caused it.

Also it appears that one more sudo is needed in `ostree-ng.sh` for `skopeo copy`

atodorov (dismissed):
LGTM but it sounds like it needs  a bit more work.

ondrejbudai (dismissed):

ondrejbudai (approved):

PR #2101, opened 2021-12-06
distro/depsolve/cloudapi: Add 3rd-party repository support.

ldjebran:
Allow 3rd-party repositories to be supported and custom packages installed.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -11,12 +11,17 @@ type DepsolveJobImpl struct {
 	RPMMDCache string
 }
 
-func (impl *DepsolveJobImpl) depsolve(packageSets map[string]rpmmd.PackageSet, repos []rpmmd.RepoConfig, modulePlatformID, arch, releasever string) (map[string][]rpmmd.PackageSpec, error) {
+func (impl *DepsolveJobImpl) depsolve(packageSets map[string]rpmmd.PackageSet, repos []rpmmd.RepoConfig, modulePlatformID, arch, releasever string, packageSetsRepositories map[string][]rpmmd.RepoConfig) (map[string][]rpmmd.PackageSpec, error) {
 	rpmMD := rpmmd.NewRPMMD(impl.RPMMDCache)
 
 	packageSpecs := make(map[string][]rpmmd.PackageSpec)
 	for name, packageSet := range packageSets {
-		packageSpec, _, err := rpmMD.Depsolve(packageSet, repos, modulePlatformID, arch, releasever)
+		repositories := make([]rpmmd.RepoConfig, len(repos))
+		copy(repositories, repos)
+		if packageSetRepositories, ok := packageSetsRepositories[name]; ok {
+			repositories = append(repositories, packageSetRepositories...)
+		}
+		packageSpec, _, err := rpmMD.Depsolve(packageSet, repositories, modulePlatformID, arch, releasever)
 		if err != nil {
 			return nil, err
 		}
@@ -33,7 +38,7 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	}
 
 	var result worker.DepsolveJobResult
-	result.PackageSpecs, err = impl.depsolve(args.PackageSets, args.Repos, args.ModulePlatformID, args.Arch, args.Releasever)
+	result.PackageSpecs, err = impl.depsolve(args.PackageSets, args.Repos, args.ModulePlatformID, args.Arch, args.Releasever, args.PackageSetsRepositories)
 	if err != nil {
 		switch err.(type) {
 		case *rpmmd.DNFError:
--- a/internal/cloudapi/v2/openapi.v2.yml
+++ b/internal/cloudapi/v2/openapi.v2.yml
@@ -531,6 +531,12 @@ components:
           type: string
           format: url
           example: 'https://mirrors.fedoraproject.org/metalink?repo=fedora-32&arch=x86_64'
+        gpg_key:
+          type: string
+        check_gpg:
+          type: boolean
+        ignore_ssl:
+          type: boolean
     UploadOptions:
       oneOf:
       - $ref: '#/components/schemas/AWSEC2UploadOptions'
@@ -669,6 +675,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/User'
+        payload_repositories:
+          type: array
+          items:
+            $ref: '#/components/schemas/Repository'
     OSTree:
       type: object
       properties:
--- a/internal/cloudapi/v2/v2.go
+++ b/internal/cloudapi/v2/v2.go
@@ -151,6 +151,11 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 		}
 	}
 
+	var payloadRepositories []Repository
+	if request.Customizations != nil && request.Customizations.PayloadRepositories != nil {
+		payloadRepositories = *request.Customizations.PayloadRepositories
+	}
+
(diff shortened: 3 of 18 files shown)


internal/distro/distro.go
```diff
@@ -108,6 +108,9 @@ type ImageType interface {
 	// Returns the names of the pipelines that create the image.
 	PayloadPipelines() []string
 
+	// Returns the package set names safe to install custom packages via custom repositories.
+	PackageSetKeys() []string
```
Gundersanne, line 112: The name `PackageSetKeys` is a bit too generic imo, but coming up with a good name here is hard.  Since we call it Build/Payload, maybe we can call this `PayloadPackageSets() []string` @achilleas-k wdyt?

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -11,12 +11,18 @@ type DepsolveJobImpl struct {
...
 
 	packageSpecs := make(map[string][]rpmmd.PackageSpec)
 	for name, packageSet := range packageSets {
-		packageSpec, _, err := rpmMD.Depsolve(packageSet, repos, modulePlatformID, arch, releasever)
+		repositories := repos
+		if packageSetRepositories, ok := packageSetsRepositories[name]; ok {
+			repositories = make([]rpmmd.RepoConfig, 0, len(repos)+len(packageSetRepositories))
+			repositories = append(repositories, repos...)
+			repositories = append(repositories, packageSetRepositories...)
+		}
```
Gundersanne, line 24: I think moving the `make` call here would make it a bit more readable:

```
repositories := make([]rpmmd.RepoConfig, len(repos))
copy(repositories, repos)
if packageSetRepositories, ok := packageSetsRepositories[name]; ok {
	repositories = append(repositories, packageSetsRepositories[name])
}
```

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -669,6 +677,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/User'
+        repositories:
```
Gundersanne, line 680: Hm can we add a description detailing that this is about payload/extra repositories? Now there will be two places where you can specify repos in the request and it should be obvious that these are non-essential. Optionally rename to `payload_repositories` or something, wdyt @teg naming is hard .-.

internal/cloudapi/v2/v2.go
```diff
@@ -182,13 +187,31 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
+	for _, packageSetKey := range packageSetKeys {
+		packageSetsRepositories[packageSetKey] = make([]rpmmd.RepoConfig, len(rcRepositories))
+		for j, repo := range rcRepositories {
+			packageSetsRepositories[packageSetKey][j].CheckGPG = repo.CheckGpg
+			packageSetsRepositories[packageSetKey][j].IgnoreSSL = repo.IgnoreSsl
+
+			if repo.Baseurl != nil {
+				packageSetsRepositories[packageSetKey][j].BaseURL = *repo.Baseurl
+			} else {
+				return echo.NewHTTPError(http.StatusBadRequest, "BaseUrl Must be specify")
```
Gundersanne, line 202: Please create a new error for this after https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/errors.go#L38 and after https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/errors.go#L97 as a 400, `ErrorNoBaseURLInPayloadRepository` or something. And then return it as `return HTTPError(ErrorNoBaseURLInPayloadRepository)`. 

For the message "BaseURL must be specified for payload repositories".

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -531,6 +533,12 @@ components:
           type: string
           format: url
           example: 'https://mirrors.fedoraproject.org/metalink?repo=fedora-32&arch=x86_64'
+        pgp_key:
```
Gundersanne, line 536: gpg_key, we also use that internally so let's just make it consistent.

Gundersanne (commented):
Thanks, some small comments.

To test this I think it's best if we create a small local repo in `test/cases/api.sh`, and add a package from there we know only exists in that repo. But I'm happy to add that.

Maybe a unit test as well though? https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/v2_test.go#L372

cmd/osbuild-worker/jobimpl-depsolve.go
ldjebran: My intention here was to make a copy of repos and allocate a new map only when it's needed, as on 90% of cases we may not have custom repos. what do you think?

Gundersanne: it's true that the usual path is without custom repos, then again this won't have any significant performance impact either way. It's not very important either way.

ldjebran:
@Gundersanne Many thanks for your review applied all your recommendations.

Pending test/cases/api.sh test with custom repository and package to install.

internal/worker/json.go
```diff
@@ -94,11 +94,12 @@ func (pn *PipelineNames) All() []string {
...
-	Repos            []rpmmd.RepoConfig          `json:"repos"`
-	ModulePlatformID string                      `json:"module_platform_id"`
-	Arch             string                      `json:"arch"`
-	Releasever       string                      `json:"releasever"`
+	PackageSets             map[string]rpmmd.PackageSet   `json:"package_sets"`
+	Repos                   []rpmmd.RepoConfig            `json:"repos"`
+	ModulePlatformID        string                        `json:"module_platform_id"`
+	Arch                    string                        `json:"arch"`
+	Releasever              string                        `json:"releasever"`
+	PackageSetsRepositories map[string][]rpmmd.RepoConfig `json:"package_sets_repositories"`
```
Gundersanne, line 102: Let's add `json:"package_sets_repositories,omitempty"` for the PackageSetsRepositories. Technically not required but makes it cleaner.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -531,6 +531,12 @@ components:
           type: string
           format: url
           example: 'https://mirrors.fedoraproject.org/metalink?repo=fedora-32&arch=x86_64'
+        pgpkey:
```
Gundersanne, line 534: gpg_key

Gundersanne (changes requested):
Thanks for the test! Looks good.

One more comment and one nitpick.

Can you also squash the two commits? History of a PR/'update' commits shouldn't make it to main. Also see https://www.osbuild.org/guides/developer-guide/workflow.html#workflow

ldjebran:
@Gundersanne changes done, many thanks

ldjebran:
Changes agreed offline done, changed names to gpg_key and ignore_ssl

Gundersanne:
Last commit on https://github.com/Gundersanne/osbuild-composer/tree/2101-integration-test  (https://github.com/Gundersanne/osbuild-composer/commit/6167a4e385bace8b4fae681e7b4fa7fc1206112e) has an integration test which adds a repo and installs a package from it locally.

Gundersanne (commented):
LGTM, but since I wrote the integration test and I want someone else to read the api spec at least I'll let someone else ack it.

Gundersanne:
Rebased to pick up new terraform sha

ldjebran:
rebased and resolved conflicts after PR merged https://github.com/osbuild/osbuild-composer/pull/2062

achilleas-k (approved):
LGTM!

PR #2102, opened 2021-12-06
spec: build all binaries with PIE

achilleas-k:
RHEL gating requires all binaries to be built with PIE.

Just build everything with PIE unconditionally.

Verified with `annocheck`.


gicmo (dismissed):
🥧

ondrejbudai (approved):

teg (approved):

PR #2108, opened 2021-12-08
Job errors

kingsleyzissou:
This pull request includes:

- wrap user errors and upload errors in custom errors
- generate error codes to differentiate between errors
- errors can then be reported back to the UI

Job metrics with success/error rates will be implemented in a future PR.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -33,13 +33,16 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	}
 
 	var result worker.DepsolveJobResult
+	result.ResultCode = worker.JobSuccess
 	result.PackageSpecs, err = impl.depsolve(args.PackageSets, args.Repos, args.ModulePlatformID, args.Arch, args.Releasever)
 	if err != nil {
 		switch err.(type) {
 		case *rpmmd.DNFError:
 			result.ErrorType = worker.DepsolveErrorType
+			result.ResultCode = worker.DNFError
 		case error:
 			result.ErrorType = worker.OtherErrorType
+			result.ResultCode = worker.DepsolveError
 		}
 		result.Error = err.Error()
 	}
--- a/cmd/osbuild-worker/jobimpl-koji-finalize.go
+++ b/cmd/osbuild-worker/jobimpl-koji-finalize.go
@@ -111,6 +111,7 @@ func (impl *KojiFinalizeJobImpl) Run(job worker.Job) error {
 		var result worker.KojiFinalizeJobResult
 		if err != nil {
 			result.KojiError = err.Error()
+			result.ResultCode = worker.KojiFinializeError
 		}
 		err = job.Update(&result)
 		if err != nil {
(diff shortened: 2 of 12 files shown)


Gundersanne (commented):
So in general for this I wonder if it makes sense to wait until we have structured error output from osbuild. And then base whatever we standerdize on, on that.

kingsleyzissou:
Yeah that's a fair point. This does feel like quite a bit of a workaround

teg:
I vote, as always, to not block but put in a workaround for now and revisit it once we have the new output.

Of course that means we need to make sure this doesn't paint us into a corner, but that we have a reasonable path to perfection (I didn't review this yet).

Thoughts @achilleas-k @gicmo ?

Gundersanne:
Right, the only problem is that the objects in `internal/worker/json.go` are implicit api. So any changes there would need to be compatible, and why not just decide on the format from `osbuild --json` first? Doesn't have to be hugely complex either just an object which makes sense to return from both the future osbuild http api and as it is now with the --json flag.

We could make the errors similar to what we have for api.openshift.com with standerdized error codes but could also make it completely different.

gicmo:
> Right, the only problem is that the objects in `internal/worker/json.go` are implicit api. So any future changes there would need to be compatible, and why not just decide on the format from `osbuild --json` first? Doesn't have to be hugely complex either just an object which makes sense to return from both the future osbuild http api and as it is now with the --json flag.
> 
> We could make the errors similar to what we have for api.openshift.com with standerdized error codes but could also make it completely different.

Happy to take any suggestions for an error reporting format (bonus if it is some RFC) but sure, why not the existing one from api.openshift.com. 
We could osbuild have return some generic failed error stage and just always return that until we get more detailed errors. Maybe a build error if the build fails (our current `success: False`).

achilleas-k:
> We could make the errors similar to what we have for api.openshift.com with standardized error codes

I think this is a good idea.  I think it would make our lives easier to have a standard format across all components.

gicmo:
So `api.openshift.com` seems to use:

```
{
  "kind": "string",
  "id": 0,
  "href": "string",
  "code": "string",
  "reason": "string",
  "details": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  }
}
```

where `kind` is `Error` and `id` is `Numeric identifier of the error.` and `code` is `Globally unique code of the error, composed of the unique identifier of the API and the numeric identifier of the error. For example, for if the numeric identifier of the error is 93 and the identifier of the API is clusters_mgmt then the code will be CLUSTERS-MGMT-93`.

That is generic enough so that we could use `kind` also to indicate `result`, `log` or `status`. Only thing is `details` which here seems to be key, value but the schema indicates that it can just be anything.

`osbuild`'s validation code has some structured error:

```
        return {
            "type": FAILED_TYPEURI,
            "title": FAILED_TITLE,
            "success": False,
            "errors": errors
        }
```
Which I think I got from [RFC7807](https://datatracker.ietf.org/doc/html/rfc7807).
Where `errors` here is an array of precise information about validation errors.

And of course osbuild has the build failures like:
```
   "type": "error",
    "success": False,
    "error": {
        "type": "org.osbuild.error.stage",
        "details": {
            "stage": {
                "id": failed["id"],
                "type": failed["name"],
                "output": failed["output"],
                "error": failed["error"]
            }
        }
    }
}
```

Gundersanne:
I honestly like both, and I think that it's entirely up to the code owners of `osbuild` :D

For osbuild an RFC compliant one might make more sense?

PR #2112, opened 2021-12-09
[DO NOT MERGE] rhel 8.5.0 debug rhsm

msehnout:
I just need the RPMs in S3 and CI to run.


ondrejbudai:
If this indeed fixes the CI on the 8.5 branch, I'm happy to merge this without the logging commit. Thanks a lot for looking into this, @msehnout. cc @achilleas-k

msehnout:
It doesn't fix it completely, it fixes just the RPM build :-/.

I decided not to spend any more time fixing the rest because it just gets outdated so quickly.

PR #2113, opened 2021-12-09
osbuild2: Expand dnf_config stage

Gundersanne:


internal/osbuild2/dnf_config_stage.go
```diff
@@ -1,22 +1,49 @@
...
+	}
+}
+
+func (o DNFConfigStageOptions) validate() error {
+	if o.Config != nil && o.Config.Main != nil {
+		valid := false
+		allowedIPR := []string{"4", "IPv4", "6", "IPv6"}
+		for _, v := range allowedIPR {
+			if o.Config.Main.IPResolve == v {
+				valid = true
```
achilleas-k, line 30: A very minor improvement, but we could `break` out of the loop as soon as it's found here.
We could also just return directly, but that might trip up potential future extensions to the function if we're not careful when adding them.

achilleas-k (approved):
LGTM

Minor comment, but I don't mind merging as is.

internal/osbuild2/dnf_config_stage.go
```diff
@@ -1,22 +1,49 @@
...
+		Config:    config,
+	}
+}
+
+func (o DNFConfigStageOptions) validate() error {
+	if o.Config != nil && o.Config.Main != nil {
+		valid := false
+		allowedIPR := []string{"4", "IPv4", "6", "IPv6"}
+		for _, v := range allowedIPR {
+			if o.Config.Main.IPResolve == v {
```
thozza, line 29: This is problematic. The structure member is of type `string`. If we ever extend the stage to allow more options in the `main` section, one can decide to set that option, but not the `IPResolve` option. In such case its value will be empty string. This check will always fail on the default string type value and the produced error message will be a bit misleading. In addition, the stage schema does not prevent specifying empty `config` and `main` properties.

You could probably add `""` to the list of valid values or make the member a pointer to `string`. The second option probably does not make much sense, since the empty string is not a valid stage schema value so we can use it as indication that the value was not set..

internal/osbuild2/dnf_config_stage_test.go
```diff
@@ -29,3 +39,86 @@ func TestNewDNFConfigStage(t *testing.T) {
...
+		},
+		{
+			DNFConfigStageOptions{
+				Variables: variables,
+				Config: &DNFConfig{
+					Main: nil,
+				},
+			},
+			true,
+		},
```
thozza, line 93:
```go
{
	DNFConfigStageOptions{
		Variables: variables,
		Config: &DNFConfig{
			Main: &DNFConfigMain{	},
		},
	},
	true,
}
```
Should be added as a valid case.

thozza (commented):
Very nice work! I think that the validation check is a bit problematic, but otherwise this looks perfect 🥳

thozza (dismissed):
Looks great! Thanks.

achilleas-k (dismissed):
Still looks good :)

achilleas-k (approved):

thozza (approved):
I thought that we already merged this 😱

PR #2119, opened 2021-12-13
dnf json cache cleaner

lavocatt:
This pull request includes:

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -11,20 +11,82 @@ import logging
 import sys
 from http.server import BaseHTTPRequestHandler
 import pathlib
+from multiprocessing import Lock
+import queue
+import shutil
+from datetime import datetime, timedelta
 
 import dnf
 import hawkey
+import pickle
 
+# Logging setup (to systemd if available)
 formatter = logging.Formatter(
         fmt="%(asctime)s %(name)s.%(levelname)s: %(message)s",
         datefmt="%Y.%m.%d %H:%M:%S"
         )
 handler = logging.StreamHandler(stream=sys.stdout)
 handler.setFormatter(formatter)
-
 log = logging.getLogger('dnf-json')
 log.addHandler(handler)
 log.setLevel(logging.INFO)
+
+# Synchronisation necessary for the multiprocess request handling.
+process_lock = Lock()
+
+class CacheState():
+    """
+    A CacheState keeps track of the cache folders.
+    CacheState objects can update the list of used cache folders, which
+    will associate a date object to the name of the folder.
+    CacheState objects can ask to clean the unused cache folders.
+    Meaning that the folders having reach a certain timeout will be deleted.
+    """
+
+    def __init__(self, cache_dir, cache_timeout, folder_dict={}):
+        """
+        cache_dir the place where the state is stored
+        folder_dict a dict containing the existing list of cache folders
+        cache_timeout a timedelta before a cache folder can be deleted
+        """
+        self.cache_dir = cache_dir
+        self.folder_dict = folder_dict
+        self.cache_timeout = cache_timeout
+
+    def update_used(self, folder):
+        """
+        Associate a datetime.now() to the folders given as parameters
+        """
+        log.debug("Folder %s was used", folder)
+        self.folder_dict[folder] = datetime.now()
+
+    def clean_unused(self):
+        """
+        Delete the folders having reach the timeout
+        """
+        log.info("clean unused folders")
+        now = datetime.now()
+        list_folder_to_delete = []
+        for folder, then in self.folder_dict.items():
+            delta = now - then
+            log.debug("delete %s if %s > than %s", folder, delta, self.cache_timeout)
+            if delta > self.cache_timeout:
+                list_folder_to_delete.append(folder)
+        for folder in list_folder_to_delete:
+            del self.folder_dict[folder]
+            shutil.rmtree(folder)
+
+    @staticmethod
+    def load_cache_state_from_disk(cache_dir):
+        try:
+            with open(os.path.join(cache_dir,"cache_state.pkl"), "rb") as inp:
+                return pickle.load(inp)
+        except FileNotFoundError:
+            return CacheState(cache_dir, timedelta(hours=24))
+
+    def store_on_disk(self):
+        with open(os.path.join(self.cache_dir, "cache_state.pkl"), "wb") as outp:
+            return pickle.dump(self, outp)
 
 class Solver():
 
@@ -127,7 +189,7 @@ class Solver():
         return checksums
 
     def _timestamp_to_rfc3339(self, timestamp):
-        d = datetime.datetime.utcfromtimestamp(timestamp)
+        d = datetime.utcfromtimestamp(timestamp)
         return d.strftime('%Y-%m-%dT%H:%M:%SZ')
 
     def dump(self):
@@ -181,6 +243,26 @@ class Solver():
         }
 
 class DnfJsonRequestHandler(BaseHTTPRequestHandler):
+    """
+    Answers Http requests to depsolve or dump packages.
+    """
+
+
+    def init_cache_folder_list(self, repos):
+        cache_folders = []
+        for repo in repos:
+            if "baseurl" in repo:
+                url = repo["baseurl"]
+            elif "metalink" in repo:
+                url = repo["metalink"]
+            elif "mirrorlist" in repo:
+                url = repo["mirrorlist"]
+            else:
+                assert False
+            digest = hashlib.sha256(url.encode()).hexdigest()[:16]
+            repoid = repo["id"]
+            cache_folders.append(f"{self.cache_dir}/{repoid}-{digest}")
+        return cache_folders
 
     def _send(self):
         self.client_address=('',)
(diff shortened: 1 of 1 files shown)


dnf-json
```diff
@@ -182,6 +216,22 @@ class Solver():
 
 class DnfJsonRequestHandler(BaseHTTPRequestHandler):
 
+    cache_folders = []
```
ondrejbudai, line 219: An instance of `DnfJsonRequestHandler` is created per a request, right? If it isn't, this variable is IMHO not thread-safe.

dnf-json
```diff
@@ -182,6 +216,22 @@ class Solver():
...
+
+    def init_cache_folder_list(self, repos):
+        for repo in repos:
```
ondrejbudai, line 222: list comprehension and a nested convert function from `repo` to `f"{repoid}-{digest}" might be slightly more pythonic I think.

dnf-json
```diff
@@ -11,10 +11,17 @@ import logging
 import sys
 from http.server import BaseHTTPRequestHandler
 import pathlib
+from multiprocessing import Value, Queue, JoinableQueue
```
ondrejbudai, line 14: `JoinableQueue` seems to be unused.

dnf-json
```diff
@@ -200,64 +250,129 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
+        # update the caches utilisation from the list of used folders
+        cache_state.update_used(self.cache_folders)
+        # clean all the folders unused since a timeout duration
+        cache_state.clean_unused(self.cachedir, CACHE_TIMEOUT)
+        # store cache state for next iteration
+        self.store_cache_state_to_disk(cache_state)
+
+    def load_cache_state_from_disk(self):
+        try:
+            with open(f"{self.cachedir}/cache_state.pkl", "rb") as inp:
```
ondrejbudai, line 359: `path.join` might be better

dnf-json
```diff
@@ -200,64 +250,129 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
+        # load the old save from previous run
+        cache_state = self.load_cache_state_from_disk()
+        # update the caches utilisation from the list of used folders
+        cache_state.update_used(self.cache_folders)
+        # clean all the folders unused since a timeout duration
+        cache_state.clean_unused(self.cachedir, CACHE_TIMEOUT)
+        # store cache state for next iteration
+        self.store_cache_state_to_disk(cache_state)
+
+    def load_cache_state_from_disk(self):
```
ondrejbudai, line 357: can we make this a static method of `CacheState`?

dnf-json
```diff
@@ -200,64 +250,129 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
+        try:
+            with open(f"{self.cachedir}/cache_state.pkl", "rb") as inp:
+                return pickle.load(inp)
+        except FileNotFoundError as e:
+            return CacheState()
+
+    def store_cache_state_to_disk(self, cache_state):
```
ondrejbudai, line 364: I think this can also be a method of `CacheState`.

dnf-json
```diff
@@ -200,64 +250,129 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
+                process_count.value -= 1
+                # Only the last process alive is making the cleaning.
+                # This is because we don't want to delete directories that might
+                # be used by a live dnf instance.
+                if process_count.value == 0:
+                    self.clean_cache()
+                else:
+                    # If the process is not the last, it sends it's used folders
+                    # to the last one.
+                    self.send_cache_folder_used()
```
ondrejbudai, line 332: I think the code might be easier to understand if we always call this method. IOW I would put this line above `if process_count.value == 0:`. Note that `clean_cache` must be slightly adjusted (`self.cache_folders` can't be reused in this case which is IMHO good.)

ondrejbudai (changes requested):
Nice! This should solve our big pain point in the internal deployment. I put some comments inline. To sum them up, I have some questions about `multiprocessing` and I'm making some suggestions to make the code slightly easier to comprehend. Don't get me wrong, this PR is awesome. :heart_eyes:

dnf-json
lavocatt: This could be done even without adjusting the `clean_cache` method. Because it will fetch all the caches send through the queue and eventually it's own. But yes, maybe cleaner without having the process sending itself its own stuff.
But yeah, I've been thinking about for some time now. The thing is there can be some kind of starvation where in the worst case scenario the cache is never cleaned. IOW (you taught me this one today :smile: ) if there's always a new process answering a request arriving, then we never reach `0` in the counter, then the `clean_cache` is never called.
What I want to do is the following:
* Having another process whose work is to receive name of cache folders, update the file on disk and clean stuff.
* Make this process clean the cache at fix intervals, let's say every 50 requests.
* Make the other processes hold when the cleaning occurs.

dnf-json
lavocatt: Yes.

dnf-json
lavocatt: Yes, this make sense.

dnf-json
lavocatt: Okay, will do.

dnf-json
lavocatt: Yes that's right. But that's the forking library that is taking care of this.

dnf-json
```diff
@@ -200,64 +250,129 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
...
 
 log.info("Starting the dnf-json server")
 
 LISTEN_FDS = int(os.environ.get("LISTEN_FDS", 0))
 SOCK_PATH = "/run/osbuild-dnf-json/"
 SOCK_NAME = "api.sock"
 
+# The dnf-json web server has to use forks to serve the requests. Because the
+# dnf library is leaking memory in its Cpp side.
 class SystemDActivationSocketServer(socketserver.ForkingMixIn, socketserver.UnixStreamServer):
```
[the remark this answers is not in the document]
lavocatt: Hey! Thanks for the awesome review! So we unfortunately have to use forks here with `ForkingMixIn` Because using threads leads to loss of ram each time a request is answered. To answer the question, yes `ForkingMixIn` uses forks.

dnf-json
ondrejbudai: I wonder if locking per a cache folder rather than per a whole cache can solve this issue. Then the cleaner process can just process unlocked directories and ignore locked ones (and it doesn't actually make sense to clean these because they are being used).

ondrejbudai: Hmm, I wonder if we can just use `mtime`, or `atime` on these directories & per-directory locking to decide whether they can be deleted...

lavocatt: We can't rely change the locking mechanism as it is handled by the `dnf` library. However we could have multiple cache directories allowing multiple requests to be answered at the same time. I've changer the whole thing to fit my previous plan. I just need to find where to declare that dnf-json know requires this lib: https://pypi.org/project/prwlock/

lavocatt:
With this new solution performance is very similar than without any locks as long as only one process depsolves at a time. When many processes are accessing the depsolving in parallel, perf is worse than without locks. But I think that is a marginal situation as we have only one worker process on each machine accessing it's own dnf-json daemon.

dnf-json
```diff
@@ -11,21 +11,83 @@ import logging
...
+        folder_dict a dict containing the existing list of cache folders
+        cache_timeout a timedelta before a cache folder can be deleted
+        """
+        self.cache_dir = cache_dir
+        self.folder_dict = folder_dict
+        self.cache_timeout = cache_timeout
+
+    def update_used(self, folder):
+        """
+        Associate a dateime.now() to the folders given as parameters
```
ondrejbudai, line 58:
```suggestion
        Associate a datetime.now() to the folders given as parameters
```

dnf-json
```diff
@@ -11,21 +11,83 @@ import logging
...
+            del self.folder_dict[folder]
+            shutil.rmtree(folder)
+
+    @staticmethod
+    def load_cache_state_from_disk(cache_dir):
+        try:
+            with open(os.path.join(cache_dir,"cache_state.pkl"), "rb") as inp:
+                return pickle.load(inp)
+        except FileNotFoundError:
+            return CacheState(cache_dir, timedelta(hours=1))
```
ondrejbudai, line 85: Isn't 1 hour too little? The cache isn't that big, I would go with 24 hours, or even more.

dnf-json
```diff
@@ -181,6 +243,25 @@ class Solver():
         }
 
 class DnfJsonRequestHandler(BaseHTTPRequestHandler):
+    """
+    Answers Http requests to depsolve or dump packages.
+    """
+
+    cache_folders = []
```
ondrejbudai, line 250: Do we even need this as an instance variable? It might be simpler to just have this variable locally in `do _POST`.

ondrejbudai (commented):
Looks very nice, I can feel this being merged soon. :)

dnf-json
lavocatt: It depends on the proportion of the onetimer repos in the requests we get. One hour will not delete repos that are frequently used anyway but will flush quickly the others. How often should we clean the cache now ?

dnf-json
lavocatt: No we don't indeed :)

dnf-json
lavocatt: I bumped it to 24H :)

dnf-json
lavocatt: Fixed.

ondrejbudai (dismissed):

ondrejbudai:
Thanks a lot for doing this! This should help tremendously with keeping the internal service happy (until we can sunset it).

ondrejbudai (approved):
Still good.

PR #2124, opened 2021-12-14
debug: just trying something in ci

jrusz:
Just trying if connection to runner persists after reboot.


jrusz:
@atodorov it just dies right after I hit reboot without even trying to connect back https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1882924115  
I'll just go back to making it work with libvirt.

PR #2132, opened 2021-12-16
test: make test more resilient

jabia99:
During manual cleanup of unused resources, the storage account can get
removed. The current storage account is not possible to remove
manually, but adding this check to make it more resilient in future
scenarios.

This pull request includes:

### DIFF
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -178,6 +178,17 @@ name = "cloud-init"
 enabled = ["sshd", "cloud-init", "cloud-init-local", "cloud-config", "cloud-final"]
 EOF
 
+# Make sure the specified storage account exists
+if [ "$(az resource list --name "$AZURE_STORAGE_ACCOUNT")" == "[]" ]; then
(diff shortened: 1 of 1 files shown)


jrusz (changes requested):
The idea behind the task was to create the storage account if it's removed not fail the test. so instead of the exit just use azure cli to create the storage account/storage container please. 

To test it you can temporarily change the variable to something else and see if it creates it and the test passes.

jabia99:
> The idea behind the task was to create the storage account if it's removed not fail the test. so instead of the exit just use azure cli to create the storage account/storage container please.

I misunderstood the ticket. I added this functionality and I think it's working. (I tested the feature using the same code locally)

ondrejbudai (commented):
I will let @jrusz decide on this but FWIW I like this. :)

jrusz (approved):
Sorry I forgot about this, got dragged into other stuff. It looks good and if you say it works, let's merge it :)

PR #2137, opened 2021-12-16
Tests-generation: start dnf-json without systemd

lavocatt:
If needed start the dnf-json dameon directly without using systemd. This
can allow to use the test case generation even without the packages
installed on the system.

This pull request includes:

### DIFF
--- a/test/cases/generation.sh
+++ b/test/cases/generation.sh
@@ -15,11 +15,26 @@ function greenprint {
 # install requirements
 sudo dnf -y install go
 
+# test the test cases generation when systemd starts dnf-json
 # stop dnf-json socket
 sudo systemctl stop osbuild-dnf-json.socket
 
-# test the test cases generation
 WORKDIR=$(mktemp -d -p /var/tmp)
+
+OSBUILD_LABEL=$(matchpathcon -n "$(which osbuild)")
+chcon "$OSBUILD_LABEL" tools/image-info
+
+# test the test case generation when dnf-json socket is stopped
+sudo ./tools/test-case-generators/generate-test-cases\
+    --output test/data/manifests\
+    --arch x86_64\
+    --distro rhel-8\
+    --image-type qcow2\
+    --store "$WORKDIR"
+
+# test the test cases generation without systemd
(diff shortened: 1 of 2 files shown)


lavocatt:
I would have loved to extend the test for this, but i can't find a way to disable the dnf-json.socket so that systemd won't actually start it again with the command "systemct start dnf-json.socket"

achilleas-k (approved):
Ideally we would have a context manager that can kill the process even if the script fails, but since it's a dev/test tool I think it's fine as is.

lavocatt:
> Ideally we would have a context manager that can kill the process even if the script fails, but since it's a dev/test tool I think it's fine as is.

Actually I've been doing some random ctrl+c through the execution and the orphan process of dnf-json gets killed every time well.

thozza:
> I would have loved to extend the test for this, but i can't find a way to disable the dnf-json.socket so that systemd won't actually start it again with the command "systemct start dnf-json.socket"

You could mask the socket unit using `systemctl mask dnf-json.socket`

thozza (approved):
Thanks for the change!

I agree with @achilleas-k that a context manager would be the ideal solution, but for a development tool, this is good enough as well.

I added comment if you want to test that the socket unit does not get started.

lavocatt:
> > I would have loved to extend the test for this, but i can't find a way to disable the dnf-json.socket so that systemd won't actually start it again with the command "systemct start dnf-json.socket"
> 
> You could mask the socket unit using `systemctl mask dnf-json.socket`

Unfortunately mask won't work here. I need to find a way to do that manually by deleting files by hand.

achilleas-k:
> Unfortunately mask won't work here. I need to find a way to do that manually by deleting files by hand.

I think a good test is by running the `generate-all-test-cases` script on a system where `osbuild-composer` isn't installed (so just from a source tree).  Would that work?

thozza:
I found an easier solution for this problem and submitted it as PR #2138

lavocatt:
Cool that you came up with something better! I'm closing this one then.

PR #2139, opened 2021-12-17
RHEL-86/90: refactoring of `osPipeline` and its variants

thozza:
### Goals
- Reduce `osPipeline` variants (`ec2BaseTreePipeline`, `osPipeline` and `ostreeTreePipeline`) to a single `osPipeline`, which handles all the potential OS customizations. Nevertheless there are still other pipeline functions which handle pieces of customizations (mostly related to EDGE images).
- Do not hardcode any default configuration values in the `osPipeline` to prevent us from having to introduce yet another variant of the function in case a new image type needs different defaults (e.g. timezone or keyboard layout).

### Approach

- Introduce a new data structure `ImageConfig`, which holds the default configuration values for a specific image. There are however still some configuration values left in the `imageType` structure, which could be moved to the `ImageConfig` structure.
- The `ImageConfig` is defined on the distribution level, providing defaults for the whole distro.
- Specific image types can define their own `ImageConfig` which extends and/or overrides options set in the distribution-wide default configuration. The way this works is that any options not set on the Image Type-level are inherited from the distribution defaults. For options which are set in both, the distribution and image type, the image type option is preferred.
- The `osPipeline` becomes a generic function (which could be even shared among distributions), which simply adds appropriate stages to the OS pipeline based on the default Image Type configuration and user-provided BP customizations.

### Changes

- The patch set first gradually modifies the RHEL-90 pipelines with the goal to replace `ec2BaseTreePipeline` and `ostreeTreePipeline` by `osPipeline`.
- Configuration values previously hard-coded in `ec2BaseTreePipeline`, `osPipeline` and `ostreeTreePipeline` are moved to `ImageConfig` structure instances defined in the distribution and respective image types.
- Image test cases are regenerated in every commit that could affect them, so that it is clear which commit changed what.
- As part of the pipeline functions unification, the subscription customizations and RHSM configuration stages are now applied only to RHEL variant of the distribution (meaning not applied on CentOS Stream). In addition, no `subscription-manager` packages are installed any more on CentOS Stream images. This reflects upstream CentOS Stream image definitions. Details are explained in the respective commits.
- As the last step, the pipeline refactoring done for RHEL-90 is applied as a whole to RHEL-86.

### Notable points

- While there are some changes in the image manifests, these are mainly related to reordering of stages in the `os` pipeline resulting from the pipeline functions consolidation. These changes have no effect on the resulting image that is produced, which is apparent from no changes in the `image-info` report.
- The only changes in image test cases, which resulted in changes in the `image-info` report, are related to CentOS Stream images due to the subscriptions and RHSM configuration being applied only on RHEL and `subscription-manager` packages not being installed by default on CentOS Stream.
- The huge number of added lines in this PR comes from two added image test cases, which were previously forgotten to be generated:
  - `centos_8-ppc64le-qcow2-boot.json`
  - `centos_8-ppc64le-qcow2_customize-boot.json`

### Ideas

- Having the image configuration represented as a data structure (`ImageConfig`) can prove to be handy to represent a specific variant of any image. E.g. we could have a default definition of what it means for an image to be a SAP image (configuration-wise) and potentially easily apply it to an existing image type configuration. There may be definitely differences in the needed configuration for a specific footprint, but even then it should be easier to share the common parts.
- In an ideal case, we could end up with just generic functions defined in `<distro>/pipelines.go`, which could be eventually shared among all the distro definitions. The core of what is a distro would become the content, configuration and export format of images, which could be hopefully represented just as a data, not as as a code.

This pull request includes:

### DIFF
--- a/internal/distro/image_config.go
+++ b/internal/distro/image_config.go
@@ -1,0 +1,48 @@
+package distro
+
+import "github.com/osbuild/osbuild-composer/internal/osbuild2"
+
+// ImageConfig represents a (default) configuration applied to the image
+type ImageConfig struct {
+	Timezone            string
+	TimeSynchronization *osbuild2.ChronyStageOptions
+	Locale              string
+	Keyboard            *osbuild2.KeymapStageOptions
+	EnabledServices     []string
+	DisabledServices    []string
+	DefaultTarget       string
+	Sysconfig           []*osbuild2.SysconfigStageOptions
+}
+
+// InheritFrom inherits unset values from the provided parent configuration and
+// returns a new structure instance, which is a result of the inheritance.
+func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
+	finalConfig := ImageConfig(*c)
+	if parentConfig != nil {
+		if finalConfig.Timezone == "" {
+			finalConfig.Timezone = parentConfig.Timezone
+		}
+		if finalConfig.TimeSynchronization == nil {
+			finalConfig.TimeSynchronization = parentConfig.TimeSynchronization
+		}
+		if finalConfig.Locale == "" {
+			finalConfig.Locale = parentConfig.Locale
+		}
+		if finalConfig.Keyboard == nil {
+			finalConfig.Keyboard = parentConfig.Keyboard
+		}
+		if finalConfig.EnabledServices == nil {
+			finalConfig.EnabledServices = parentConfig.EnabledServices
+		}
+		if finalConfig.DisabledServices == nil {
+			finalConfig.DisabledServices = parentConfig.DisabledServices
+		}
+		if finalConfig.DefaultTarget == "" {
+			finalConfig.DefaultTarget = parentConfig.DefaultTarget
+		}
+		if finalConfig.Sysconfig == nil {
+			finalConfig.Sysconfig = parentConfig.Sysconfig
+		}
+	}
+	return &finalConfig
+}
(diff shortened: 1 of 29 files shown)


achilleas-k:
I *really* like where this is going :+1:

ondrejbudai (commented):
I love that this opens a path for so many customizations. If we just expose ImageConfig in the API and make it override the image type's config, so many properties can be easily customized by composer's users. I love it. Note that I greatly oversimplified this problem but you know what I mean.

I like this PR very much, thanks for it, you have my approval. As I'm not an expert in this field, I would also like @achilleas-k's opinion when he returns from his PTO. :)

thozza:
Thanks @ondrejbudai ! I would definitely wait for @achilleas-k to review, as well as @gicmo as he had some opinions about this.

Yesterday I realized that we could eventually have list of generic pipelines in composer (such as `osPipeline` but also `liveImagePipeline`, `xzArchivePipeline`, ...) and have a specific configuration structure for each implemented pipeline. This way we could represent all image definitions just as a data (list of pipelines and their configurations) and all of the `*Pipeline` functions would be just generic and would not hard-code any values. Some configuration would be probably common for multiple pipeline configurations, and this could be represented by embedding a common structure in the pipeline-specific one (IOW using composition).

I should probably draft this idea somewhere, but wanted to record it at least here before the EOY...

achilleas-k:
Quick update: I had a very brief look today but I'd like to have a closer look still.

achilleas-k (approved):
This is all great.  Thank you!!

PR #2141, opened 2021-12-17
LUKS & LVM support and file system refactoring

gicmo:
This PR does refactors the current file system code to allow it to model complex disk layouts that include LUKS container and LVM. For this it introduces a simple entity data model that can be used to navigate and operate on such complex layouts in a very generic way. 

TODO:
  - [x] More documentation
  - [x] More tests
  - [ ] Kernel command line options for LUKS in non-ostree deploy pipelines


internal/distro/rhel86/partition_tables.go
```diff
@@ -155,7 +155,7 @@ var ec2BasePartitionTables = distro.BasePartitionTableMap{
 				},
 			},
 			{
-				Size: 1048676,
```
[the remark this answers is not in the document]
gicmo: I am restoring it to `1048576`, which is precisely `512 MiB`.  104**85**76 became 104**86**76 during the 8.5 to 8.6 fork. The later is `512 MiB` and a bit (and also unaligned).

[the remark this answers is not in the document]
gicmo: Another indicator, btw, that image tests alone are not sufficient. The new value happy got encoded into them even though it should be the same across 8.5, 8.6, 9.0beta, 9.0. Maybe really need some kinda of linter, or better pay attention to the image test's content (which is probably too hard).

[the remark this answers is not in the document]
thozza: OK, I see it now. Actually 9.0 Beta has the same value as 8.5 so there was really some bug introduced in 8.6. Image test cases are definitely hard to review as part of PRs, but not impossible. I'm not sure how much people pay attention to them when they regenerate them. A linter would be definitely an improvement 😉

kingsleyzissou:
Wow this is already such an enormous improvement and has cleaned up the partition creation stuff.
The byte/sector size stuff really was a mess - sorry about that!

internal/disk/disk.go
```diff
@@ -47,9 +60,57 @@ type Filesystem struct {
...
+
+		SectorSize:   pt.SectorSize,
+		ExtraPadding: pt.ExtraPadding,
+	}
+}
+
 // Converts PartitionTable to osbuild.QEMUAssemblerOptions that encode
 // the same partition table.
-func (pt PartitionTable) QEMUAssemblerOptions() osbuild.QEMUAssemblerOptions {
+func (pt *PartitionTable) QEMUAssemblerOptions() osbuild.QEMUAssemblerOptions {
```
ondrejbudai, line 113: I like to use a non-pointer in these cases because it guarantees immutability.

gicmo: Ah fair point, but since this struct has reference members, it is not _really_ immutability, and there was actually a bug around this in PartitionTable. I also personally think it is surprising if some methods are by-value and some by-ref, but I can adjust if this is what the rest of the code does.

thozza: +1 for keeping all methods of the same type (by-value or by-type).

internal/distro/rhel86/pipelines.go
```diff
@@ -996,16 +996,23 @@ func bootloaderConfigStage(t *imageType, partitionTable disk.PartitionTable, ker
...
 
-	options := grub2StageOptions(partitionTable.RootFilesystem(), partitionTable.BootFilesystem(), kernelOptions, kernel, kernelVer, uefi, legacy, t.arch.distro.vendor, install)
+	options := osbuild.GenGrub2StageOptions(&partitionTable, kernelOptions, kernel, kernelVer, uefi, legacy, t.arch.distro.vendor, false)
 	options.Greenboot = greenboot
 
+	// before unifying the org.osbuild.grub2 stage option generator, we didn't
+	// set the following for RHEL 8.5, so we need to revert here to maintain
+	// the old behaviour
+	if uefi {
+		options.UEFI.Unified = false
```
teg, line 1010: Also for 8.6?

gicmo: Unified grub config is a RHEL 9 thing.

internal/disk/luks.go
```diff
@@ -59,3 +64,9 @@ func (lc *LUKSContainer) Clone() Entity {
 		Payload: lc.Payload,
 	}
 }
+
+func (lc *LUKSContainer) SetUUID(rng *rand.Rand) {
```
teg: Should this be `GenUUID()`?

gicmo: it should! I think I pushed a fix for this...

gicmo: Should have been fixed in commit 77aa4381e.

gicmo: Ah the fixup slipped into the wrong commit, I will fix.

ondrejbudai:
Needs rebase (to fix EL9 mockbuild).

gicmo:
I am marking this as ready for review. There are still two TODO list items open but they will be additive to the existing code and not change it anymore.

achilleas-k:
Manifest and image info updates coming later today

internal/disk/disk.go
```diff
@@ -15,19 +15,17 @@ import (
...
 }
 
 type Partition struct {
-	Start    uint64
-	Size     uint64
-	Type     string
-	Bootable bool
+	Start    uint64 // Start of the partition in sectors
+	Size     uint64 // Size of the partition in sectors
+	Type     string // `gpt` or `dos`
```
teg: Why do both the partitions and the partition table have a `Type` field? Will they ever differ?

gicmo: I am not sure that is right, actually. Let me check

internal/disk/disk.go
```diff
@@ -14,6 +14,11 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/osbuild2"
 )
 
+const (
+	// Default sector size in bytes
+	DefaultSectorSize = 512
```
teg, line 29: Could this ever change per-distro? I.e., will we see a future where the default is 4096 for some but not all distros?

gicmo: Maybe, it will be moved to the `PartitionTable` in a subsequent commit, which will make it more flexible.

teg: Ah, ignore this comment.

internal/disk/disk.go
gicmo: No, the comment is wrong, this is not `gpt` or `dos` but the partition type, Linux data, EFI, etcpp

internal/disk/disk.go
```diff
@@ -18,6 +30,63 @@ const (
...
+
+	// GetFSTabOptions returns options for mounting the entity.
+	GetFSTabOptions() FSTabOptions
+}
+
+// A VolumeContainer is a container that is able to create new volumes.
+//
+// CreateVolume creates a new volume with the given mount point and
+// size and returns it.
+type VolumeContainer interface {
```
teg, line 100: As discussed, I found the concept of `Volume` a bit confusing here. Maybe more documentation could fix that, but renaming to `MountableContainer` and `CreateMountable()` would also work, and I think that would make it self explanatory.

teg (dismissed):
Looks great! Left minor comments, but nothing worth blocking on.

internal/disk/disk.go
teg: Ah, got it. Makes sense then.

internal/disk/disk.go
```diff
@@ -1,228 +1,148 @@
...
-func (fs Filesystem) QEMUFilesystem() osbuild.QEMUFilesystem {
-	return osbuild.QEMUFilesystem{
-		Type:       fs.Type,
-		UUID:       fs.UUID,
-		Label:      fs.Label,
-		Mountpoint: fs.Mountpoint,
-	}
+// NewRandomVolIDFromReader creates a random 32 bit hex string to use as a
+// volume ID for FAT filesystems
+func NewRandomVolIDFromReader(r io.Reader) (string, error) {
```
thozza, line 144: I wonder whether this function really needs to be exported. The function to generate UUID is not exported and the fact that some filesystems use VolumeID and some UUID could be handled by `Filesystem.GenUUID()` based on the FS type. Especially when it is used only in a single place in `internal/distro/rhel90/pipelines.go`.

gicmo: Good point! I think I like the idea of just doing it in the filesystem itself. I do wonder if it can wait to be done in a follow up though.

thozza (commented):
Very nice work! It must have been quite an effort to rework all of partitioning code and extend it. Respect 🙇‍♂️ 

I added a "few" inline comments, but to sum it up:

- I have concerns about mixing sizes in bytes and sectors when using `disk.Partition`.
- I think that some disk-related methods could be made more generic and not hardcode things like e.g. filesystem with XFS, etc. The question is whether it is worth it and makes sense, but I thought that it could, given the extent of refactoring and goal to make the code generic.
- A bit of a nitpick, but I think that consistency is important. You used `Gen...StageOptions()` name for constructors, instead of `New...StageOptions()` as it is currently used across the code base. Using the `Gen...` name for functions / methods which generate e.g. slice of stages which do something more advanced probably makes sense. However I think that we should keep naming consistency for constructors.

gicmo:
> Very nice work! It must have been quite an effort to rework all of partitioning code and extend it. Respect 🙇‍♂️
> 
> I added a "few" inline comments, but to sum it up:
> 
> * I have concerns about mixing sizes in bytes and sectors when using `disk.Partition`.

Inside `disk` all the declarations should actually be bytes now and there should not be a mix anymore. Any left-over, I think, would be a mistake. In `osbuild2` it will need to be sectors because that is what `org.osbuild.sfdisk` wants.

> * I think that some disk-related methods could be made more generic and not hardcode things like e.g. filesystem with XFS, etc. The question is whether it is worth it and makes sense, but I thought that it could, given the extent of refactoring and goal to make the code generic.

I think that is actually a good idea. Just passing `Filesystems` around does not fully work, because of the case of `Btrfs`, where you can't have a different filesystem. But I think we should think about this and extend it in the future.

> * A bit of a nitpick, but I think that consistency is important. You used `Gen...StageOptions()` name for constructors, instead of `New...StageOptions()` as it is currently used across the code base. Using the `Gen...` name for functions / methods which generate e.g. slice of stages which do something more advanced probably makes sense. However I think that we should keep naming consistency for constructors.

I think our idea was `New...` is typically used in Go to create single objects, i.e. as you also call them, they are `constructors`. All our `Gen...` methods are not object constructors but helper functions to generate an array of stages. My personal initial intuition was to actually have them take a `Pipeline` parameter and just add the stages to the pipeline instead of returning them, but @achilleas-k preferred to return them because that was done in other places so we stuck with that.

internal/distro/rhel86/pipelines.go
```diff
@@ -692,14 +691,14 @@ func simplifiedInstallerBootISOTreePipeline(archivePipelineName, kver string) *o
...
 	pt := disk.PartitionTable{
-		Size: 20971520,
+		Size:       20971520,
+		SectorSize: 512,
 		Partitions: []disk.Partition{
 			{
 				Start: 0,
-				Size:  20971520 / sectorSize,
-				Filesystem: &disk.Filesystem{
+				Size:  20971520,
```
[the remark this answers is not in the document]
achilleas-k: This is a mistake in the docstring.  We initially added docstrings for all fields, then changed everything to be in bytes internally and only be converted to sectors when writing out (where necessary).

internal/disk/partition_table.go
```diff
@@ -0,0 +1,456 @@
...
+
+// Returns if the partition table contains a filesystem with the given
+// mount point.
+func (pt *PartitionTable) ContainsMountpoint(mountpoint string) bool {
+	return len(entityPath(pt, mountpoint)) > 0
+}
+
+// Create a new filesystem within the partition table at the given mountpoint
+// with the given minimum size in bytes.
+func (pt *PartitionTable) CreateFilesystem(mountpoint string, size uint64) error {
```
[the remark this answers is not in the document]
gicmo: I was wondering how to make this more generic as well. Problem is `btrfs` where it doesn't really make sense to take a `Filesystem`.

internal/disk/partition_table.go
```diff
@@ -0,0 +1,456 @@
...
+
+func (pt *PartitionTable) EnsureSize(s uint64) bool {
+	if s > pt.Size {
+		pt.Size = s
+		return true
+	}
+	return false
+}
+
+func (pt *PartitionTable) CreateVolume(mountpoint string, size uint64) (Entity, error) {
```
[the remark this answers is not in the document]
gicmo: Yeah, this is an artefact of the re-factoring work.

internal/disk/partition.go
```diff
@@ -0,0 +1,71 @@
+package disk
+
+import (
+	"fmt"
+)
+
+type Partition struct {
+	Start    uint64 // Start of the partition in sectors
+	Size     uint64 // Size of the partition in sectors
```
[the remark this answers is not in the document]
gicmo: The doc-string is actually wrong after the commit where we convert all sizes in `disks` to bytes. This needs to be fixed.

internal/disk/lvm.go
```diff
@@ -0,0 +1,130 @@
...
+
+	lv := LVMLogicalVolume{
+		Size:    size,
+		Payload: &filesystem,
+	}
+
+	vg.LogicalVolumes = append(vg.LogicalVolumes, lv)
+
+	return &vg.LogicalVolumes[len(vg.LogicalVolumes)-1], nil
+}
```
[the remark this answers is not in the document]
gicmo: Yeah, currently this is indeed a limitation. It is very much oriented on what current blueprints allow. The idea is that `CreateVolume` is a high level interface where depending on what "volume container" (`PartitionTable`, `btrfs`, `LVM`) is currently use the individual implementation does the right thing: create the partition and file system or create a sub-volume (for btrfs). If we make this more generic we would need to have another function that, depending on the "volume container" would return the correct `Entity` (`BtrfsSubVolume`, `Filesystem`). I think that would probably be more generic but also more complex and I think for now this is fine, but we might indeed have to re-visit this once we allow for different file system types or more complex layouts.

thozza:
> > Very nice work! It must have been quite an effort to rework all of partitioning code and extend it. Respect 🙇‍♂️
> > I added a "few" inline comments, but to sum it up:
> > 
> > * I have concerns about mixing sizes in bytes and sectors when using `disk.Partition`.
> 
> Inside `disk` all the declarations should actually be bytes now and there should not be a mix anymore. Any left-over, I think, would be a mistake. In `osbuild2` it will need to be sectors because that is what `org.osbuild.sfdisk` wants.

Yes, this is the case from what I remember. Nevertheless I would find it less confusing (also in case of debugging an issue related to partitions) to use a different structure type in `osbuild2`.

> > * I think that some disk-related methods could be made more generic and not hardcode things like e.g. filesystem with XFS, etc. The question is whether it is worth it and makes sense, but I thought that it could, given the extent of refactoring and goal to make the code generic.
> 
> I think that is actually a good idea. Just passing `Filesystems` around does not fully work, because of the case of `Btrfs`, where you can't have a different filesystem. But I think we should think about this and extend it in the future.
> 

Fair enough. However as I commented inline, some Entities like LVM logical volume could contain another container inside it, not just a filesystem.

> > * A bit of a nitpick, but I think that consistency is important. You used `Gen...StageOptions()` name for constructors, instead of `New...StageOptions()` as it is currently used across the code base. Using the `Gen...` name for functions / methods which generate e.g. slice of stages which do something more advanced probably makes sense. However I think that we should keep naming consistency for constructors.
> 
> I think our idea was `New...` is typically used in Go to create single objects, i.e. as you also call them, they are `constructors`. All our `Gen...` methods are not object constructors but helper functions to generate an array of stages. My personal initial intuition was to actually have them take a `Pipeline` parameter and just add the stages to the pipeline instead of returning them, but @achilleas-k preferred to return them because that was done in other places so we stuck with that.

My comment is specifically related to functions / methods which create a single object - stage options. I think that these should be treated as constructors. This is the case for other `osbuild2` stages, which were not touched by your PR. But I'll double-check your changes once more.

achilleas-k:
> > * A bit of a nitpick, but I think that consistency is important. You used `Gen...StageOptions()` name for constructors, instead of `New...StageOptions()` as it is currently used across the code base. Using the `Gen...` name for functions / methods which generate e.g. slice of stages which do something more advanced probably makes sense. However I think that we should keep naming consistency for constructors.
> 
> I think our idea was `New...` is typically used in Go to create single objects, i.e. as you also call them, they are `constructors`. All our `Gen...` methods are not object constructors but helper functions to generate an array of stages. My personal initial intuition was to actually have them take a `Pipeline` parameter and just add the stages to the pipeline instead of returning them, but @achilleas-k preferred to return them because that was done in other places so we stuck with that.

To add to this a bit:

There's two aspects here.  One is what @gicmo mentioned: I argued against calling a function `NewX()` when it wasn't returning a variable of type `X`.  This covers cases like `GenCopyFSTreeOptions()` or `GenImageKernelOptions()`, or the cases where a slice of `Stage` is returned.

The other aspect is cases like `GenZiplInstStageOptions()`, which do return the type mentioned in the function name.  I felt that given the arguments and what the functions do in these cases, it's a bit more than a constructor and doing much more than just initialising a struct type with some values.  I don't feel strongly about this though, that's still a constructor at the end of the day, I just wanted to differentiate it from the "dumb" constructors we have for stages and the like.  Maybe it's a distinction without a difference though and we should just use `NewX()` for anything that takes some arguments and creates a new `X`.

internal/disk/partition.go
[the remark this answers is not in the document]
achilleas-k: Seems like a leftover docstring.  All internal values should be in bytes now.

internal/osbuild2/disk.go
```diff
@@ -0,0 +1,81 @@
...
+
+// sfdiskStageOptions creates the options and devices properties for an
+// org.osbuild.sfdisk stage based on a partition table description
+func sfdiskStageOptions(pt *disk.PartitionTable) *SfdiskStageOptions {
+	partitions := make([]Partition, len(pt.Partitions))
+	for idx, p := range pt.Partitions {
+		partitions[idx] = Partition{
+			Bootable: p.Bootable,
+			Start:    pt.BytesToSectors(p.Start),
+			Size:     pt.BytesToSectors(p.Size),
```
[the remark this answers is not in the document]
gicmo: I am not sure I 100% understand. `disk.Partition` is bytes, `osbuild2.Partition`, defined in `sfdisk_stage.go` is sectors (as `org.osbuild.sfdisk` requires).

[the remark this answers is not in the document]
thozza: OK. I missed that the `Partition` structure used here is defined in `osbuild2` as it has the same name as the one defined in `disk` package`. Also doc strings in the `disk` package contributed to my confusion. In this case everything is correct and the mistake is on my side. Although as we discussed, renaming the `osbuild2` `Partition` to something like `SfdiskPartition` would be easier to comprehend (at least for me 😄 ).

internal/disk/partition_table.go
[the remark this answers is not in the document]
thozza: What about accepting an interface type?

thozza:
> To add to this a bit:
> 
> There's two aspects here. One is what @gicmo mentioned: I argued against calling a function `NewX()` when it wasn't returning a variable of type `X`. This covers cases like `GenCopyFSTreeOptions()` or `GenImageKernelOptions()`, or the cases where a slice of `Stage` is returned.
> 
> The other aspect is cases like `GenZiplInstStageOptions()`, which do return the type mentioned in the function name. I felt that given the arguments and what the functions do in these cases, it's a bit more than a constructor and doing much more than just initialising a struct type with some values. I don't feel strongly about this though, that's still a constructor at the end of the day, I just wanted to differentiate it from the "dumb" constructors we have for stages and the like. Maybe it's a distinction without a difference though and we should just use `NewX()` for anything that takes some arguments and creates a new `X`.

Thanks for the explanation and additional context. The source of my opinions may be that coming from other programming languages, I don't view constructors necessarily as functions which have to be simple or "dumb".

achilleas-k:
> Thanks for the explanation and additional context. The source of my opinions may be that coming from other programming languages, I don't view constructors necessarily as functions which have to be simple or "dumb".

Yeah, I'm not convinced by my argument either tbh.  Renaming :)

internal/disk/partition.go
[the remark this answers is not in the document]
achilleas-k: Fixed

internal/disk/lvm.go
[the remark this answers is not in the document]
thozza: I think that naming may be an issue here as @teg commented on the interface definition. Due to naming, I ended up having an incorrect assumption that this method should create a new Volume in the sense of a LVM LV...

internal/disk/disk.go
teg: This is clearly a candidate for follow-up work. Marking as resolved.

internal/osbuild2/disk.go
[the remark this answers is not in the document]
teg: Agreed with the naming suggestion. Though for a follow-up.

internal/distro/rhel8/distro.go
```diff
@@ -318,6 +325,10 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 	for _, m := range mountpoints {
 		if m.Mountpoint != "/" {
 			invalidMountpoints = append(invalidMountpoints, m.Mountpoint)
+		} else {
+			if m.MinSize > imageSize {
+				imageSize = m.MinSize
+			}
```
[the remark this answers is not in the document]
teg: Wasn't the point to move this logic away from the APIs and into the distro definitions? I.e., shouldn't we rather (in a follow-up) remove the `weldr` code? @achilleas-k @gicmo ? (I think this is not a blocker, as the functionality is idempotent).

internal/distro/rhel84/distro.go
```diff
@@ -367,6 +373,10 @@ func (t *imageType) pipeline(c *blueprint.Customizations, options distro.ImageOp
 	for _, m := range mountpoints {
 		if m.Mountpoint != "/" {
 			invalidMountpoints = append(invalidMountpoints, m.Mountpoint)
+		} else {
+			if m.MinSize > imageSize {
+				imageSize = m.MinSize
+			}
```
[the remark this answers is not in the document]
teg: Ditto.

internal/disk/partition_table.go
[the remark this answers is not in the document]
teg: Let's defer this discussion to a follow-up.

internal/disk/partition_table.go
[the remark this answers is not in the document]
teg: Let's address that in a follow-up.

teg (approved):
Thanks for all the code and all the reviews! Let's get this landed and do any further clean-ups in follow ups, as there does not seem to be any disagreement of the design/direction.

teg:
CentOS Stream 9 tests are known to be broken unrelated to this PR (sadly), so I'll override CI and do an admin merge.

achilleas-k:
:rocket:

gicmo:
🚀

PR #2143, opened 2021-12-20
Tools: add 'no-image-info' option to image test case requests

thozza:
Add a new option `no-image-info`, which can be specified for an image
type in `format-request-map.json` in case the image can not be analyzed
using `image-info` and thus it should not be run as part of generating
the image test case.

This is the case for e.g. the `image-installer` image type in RHEL-8.6,
which effectively breaks generating all image test cases for RHEL-8.6
due to `image-info` failing to analyze the resulting image.

This change will allow us to consistently generate image test cases
(manifests only) for all image types, which were so far not covered by
the image testing.

Modify the `generate-test-cases` script to recognize the `no-image-info`
option in the `format-request-map.json`

Regenerate `image-installer` test cases as they can not be analyzed using
`image-info`.

This pull request includes:


msehnout (approved):
Looks reasonable to me :-)

ondrejbudai (approved):

PR #2144, opened 2021-12-20
tools: AppSRE packer build

Gundersanne:


templates/packer/ansible/roles/common/tasks/rpmbuild.yml
```diff
@@ -0,0 +1,89 @@
...
+
+- name: Install build tools
+  package:
+    name:
+      - rpm-build
+      - dnf-plugins-core
+    state: present
+
+- name: Run hack for go-rpm macros
+  command: "sed -i  's/compiler(go-compiler)/golang/g' /usr/lib/rpm/macros.d/macros.go-srpm "
```
Gundersanne, line 39: @ondrejbudai I would really like to avoid this, how do you build composer on rhel properly? Or should I switch this to using mockbuild?

teg: i think mock is the closest we get to "proper rhel"

thozza: All production RPMs should be ideally built using mock. Otherwise the build will be affected by packages installed on the system where the RPM was built. Mock sets up a clean buildroot and builds RPMs inside it (uses chroot or systemd-nspawn). This is also how Brew builds RPMs.

Gundersanne: Well we do contorl the host system here, official rhel cloud image, but fair enough, swapping out with mock.

osbuild-composer.spec
```diff
@@ -28,8 +28,11 @@ License:        ASL 2.0
 URL:            %{gourl}
 Source0:        %{gosource}
 
-
+%if 0%{?packerbuild}
```
ondrejbudai, line 31: This doesn't seem to be needed anymore.

distribution/Dockerfile-ubi-packer
```diff
@@ -0,0 +1,7 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+RUN curl --location --output /etc/yum.repos.d/hashicorp.repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
+RUN microdnf install python3 rust cargo python3-devel openssl-devel
```
ondrejbudai, line 4: Hmm, why do we need rust?

distribution/Dockerfile-ubi-packer
```diff
@@ -0,0 +1,7 @@
...
+RUN pip3 install setuptools-rust
+RUN pip3 install ansible
```
ondrejbudai, line 6: Subscribing the container and using Red Hat Ansible might be better. This is using `docker` so subscribing the container should be totally possible (with `podman`, more hacks are needed because it tries to be smart about subscriptions).

templates/packer/ansible/roles/common/tasks/rpmbuild.yml
```diff
@@ -0,0 +1,95 @@
...
+    /tmp/rpmbuild/RPMS/osbuild-tools.rpm
+    /tmp/rpmbuild/RPMS/python3-osbuild.rpm
+    /tmp/rpmbuild/RPMS/osbuild-composer-core.rpm
+    /tmp/rpmbuild/RPMS/osbuild-composer-dnf-json.rpm
+    /tmp/rpmbuild/RPMS/osbuild-composer-worker.rpm
+
+- name: Cleanup build tools
+  package:
+    name:
+      - rpm-build
```
ondrejbudai, line 89: Missing `mock` here.

templates/packer/ansible/roles/common/tasks/rpmbuild.yml
```diff
@@ -0,0 +1,95 @@
...
+  shell: >-
+      mock
+      --rebuild
+      --define "commit {{ COMPOSER_COMMIT }}"
+      --define "_rpmfilename %%{NAME}.rpm"
+      --resultdir /tmp/rpmbuild/RPMS
+      /tmp/rpmbuild/SRPMS/osbuild-composer-*.src.rpm
+
+- name: Install artifacts
+  command: >-
```
ondrejbudai, line 75: I would use something different than `command` preferably. Can we use `dnf` to install local rpms?

templates/packer/ansible/roles/common/tasks/rpmbuild.yml
```diff
@@ -0,0 +1,95 @@
...
+    dnf install -y
+    /tmp/rpmbuild/RPMS/osbuild.rpm
```
ondrejbudai, line 77: I have two ideas:

1) don't install `osbuild-tools` here.
2) put everything in a local repository using `createrepo_c`, add a repofile to `/etc/yum.repos.d` (with an appropriate priority) and install just `osbuild-composer-worker`, everything else will be pulled using deps.

ondrejbudai (commented):
Nice! I put several comments inline. My biggest concern is that we are reusing the same VM for both building and running a production service. It just feels like too much. I have two ideas:

1) Run a dedicated EC2 instance just for building RPMs.
2) Run `mock` directly in the AppSRE environment in a privileged container (I just tried this locally, it perfectly works. The question is whether we can run privileged containers, I'm not sure if this is a VM, or not.

distribution/Dockerfile-ubi-packer
Gundersanne: To get ansible to install properly :/ Getting ansible in an unsubscribed ubi container is hard, installing it from epel doesn't work because there's missing dependencies.

distribution/Dockerfile-ubi-packer
Gundersanne: This depends on the host afaik . I guess you can get around the automatic secrets mapping with mounting over the right things (/run/secrets?). I think `docker` does this automatic subscription mapping as well though, at least if the host is RHEL.

teg:
> Nice! I put several comments inline. My biggest concern is that we are reusing the same VM for both building and running a production service. It just feels like too much. I have two ideas:
> 
> 1) Run a dedicated EC2 instance just for building RPMs.
> 2) Run `mock` directly in the AppSRE environment in a privileged container (I just tried this locally, it perfectly works. The question is whether we can run privileged containers, I'm not sure if this is a VM, or not.

I would love it if 2 works... Agreed that doing it on the production system would not be nice. Did we consider copr or something like that? 1 is basically running our own copr, isn't it?

Gundersanne:
> > Nice! I put several comments inline. My biggest concern is that we are reusing the same VM for both building and running a production service. It just feels like too much. I have two ideas:
> > 
> > 1. Run a dedicated EC2 instance just for building RPMs.
> > 2. Run `mock` directly in the AppSRE environment in a privileged container (I just tried this locally, it perfectly works. The question is whether we can run privileged containers, I'm not sure if this is a VM, or not.
> 
> I would love it if 2 works... Agreed that doing it on the production system would not be nice. Did we consider copr or something like that? 1 is basically running our own copr, isn't it?

Copr is another option, the annoying this is the token needs refreshing every 180 days but other than that it would work of course. 2 will only work from a privileged container, mock is not installed, unsure if we have access to those, ppl from AppSRE were unsure too.

Gundersanne:
So mock is not an option; I guess it's ec2 or copr.

Gundersanne:
Went for the ec2 option. We don't have mock and unsure if we should rely on rhcopr, plus the expiring token thing even if the cadence is slow.

teg:
"we don't have mock"? doesn't this use mock?

Gundersanne:
> "we don't have mock"? doesn't this use mock?

This uses mock in an ec2 machine yea, we don't have mock on the jenkins runner however. So first this starts an ec2 machine based on the same ami as the final image is based on, builds the rpms in mock, fetches them, does the packer build, pushes the rpm during the packer build and installs them.

Gundersanne:
Hm this breaks the packer test, i'll look into that. Unsure if we can/should keep that.

teg (approved):

PR #2147, opened 2021-12-21
Maintenance enable clouds

Gundersanne:
note that this is still dry-running so even if it's broken it's ok :)

### DIFF
--- a/cmd/osbuild-service-maintenance/main.go
+++ b/cmd/osbuild-service-maintenance/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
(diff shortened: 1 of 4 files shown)


cmd/osbuild-service-maintenance/main.go
```diff
@@ -66,11 +67,25 @@ func main() {
 	go func() {
 		defer wg.Done()
 		logrus.Info("Cleaning up GCP")
-		if conf.GoogleApplicationCreds == "" {
-			logrus.Error("GCP credentials not specified")
+		var gcpConf GCPCredentialsConfig
+		err := LoadConfigFromEnv(&conf)
```
ondrejbudai, line 71:
```suggestion
		err := LoadConfigFromEnv(&gcpConf)
```

I think?

achilleas-k (approved):
Looks good!

thozza (commented):
Too late but looked good 😄

PR #2152, opened 2021-12-22
worker/api: align error handler with cloudapi

diaasami:
This pull request includes:

### DIFF
--- a/internal/cloudapi/v2/errors.go
+++ b/internal/cloudapi/v2/errors.go
@@ -218,18 +218,17 @@ func apiErrorFromEchoError(echoError *echo.HTTPError) ServiceErrorCode {
 
 // Convert an echo error into an AOC compliant one so we send a correct json error response
 func (s *Server) HTTPErrorHandler(echoError error, c echo.Context) {
-	doResponse := func(code ServiceErrorCode, c echo.Context) {
+	doResponse := func(code ServiceErrorCode, c echo.Context, internal error) {
 		if !c.Response().Committed {
 			var err error
 			sec := find(code)
 			apiErr := APIError(code, sec, c)
 
 			if sec.httpStatus == http.StatusInternalServerError {
-				internalError, ok := echoError.(*echo.HTTPError)
 				errMsg := fmt.Sprintf("Internal server error. Code: %s, OperationId: %s", apiErr.Code, apiErr.OperationId)
 
-				if ok {
-					errMsg += fmt.Sprintf(", InternalError: %v", internalError)
+				if internal != nil {
+					errMsg += fmt.Sprintf(", InternalError: %v", internal)
 				}
 
 				c.Logger().Error(errMsg)
@@ -251,7 +250,7 @@ func (s *Server) HTTPErrorHandler(echoError error, c echo.Context) {
 	he, ok := echoError.(*echo.HTTPError)
 	if !ok {
 		c.Logger().Errorf("ErrorNotHTTPError %v", echoError)
-		doResponse(ErrorNotHTTPError, c)
+		doResponse(ErrorNotHTTPError, c, echoError)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


internal/worker/api/errors.go
```diff
@@ -148,18 +148,30 @@ func apiErrorFromEchoError(echoError *echo.HTTPError) ServiceErrorCode {
...
 			} else {
-				c.Logger().Infof("%s, Code: %s, OperationId: %s", internal,
-					apiErr.Code, apiErr.OperationId)
+				msg := fmt.Sprintf("%d Code: %s, OperationId: %s", c.Response().Status, apiErr.Code, apiErr.OperationId)
+
+				if internalErrorOk {
+					msg += fmt.Sprintf(", InternalError: %v", internalError)
+				}
+
+				c.Logger().Warn(msg)
```
diaasami, line 174: I thought `Warn` makes more sense because these requests are coming from the worker, which is controlled by us, so non-500 errors should be flagged and looked at.

internal/worker/api/errors.go
```diff
@@ -148,18 +148,30 @@ func apiErrorFromEchoError(echoError *echo.HTTPError) ServiceErrorCode {
...
 // Convert an echo error into an AOC compliant one so we send a correct json error response
 func HTTPErrorHandler(echoError error, c echo.Context) {
-	doResponse := func(code ServiceErrorCode, c echo.Context, internal error) {
+	doResponse := func(code ServiceErrorCode, c echo.Context) {
 		if !c.Response().Committed {
 			var err error
 			sec := find(code)
 			apiErr := APIError(code, sec, c)
 
+			internalError, internalErrorOk := echoError.(*echo.HTTPError)
```
Gundersanne, line 157: Hm now there's a double cast here, cloudapi does it too and I think it's much cleaner if `doResponse` just gets the internal error as an argument in case it needs to log.

Gundersanne (changes requested):
I think the double cast can be avoided by supplying the internalError to the doResponse handler as it is now.

Also we should only log he.Internal if the error is a 500. That said for the workerapi the warning makes sense on normal errors, if it doesn't get too noisy. We'll always have those errors in the worker logs too.

---

I think here you can achieve the same by just supplying `he.Internal` to doResponse, and I wouldn't mind the cloudapi being changed to behave the same way. The double casting makes it hard to read imo.

internal/worker/api/errors.go
```diff
@@ -168,22 +180,22 @@ func HTTPErrorHandler(echoError error, c echo.Context) {
...
 
 	sec, ok := he.Message.(ServiceErrorCode)
 	if !ok {
 		// No service code was set, so Echo threw this error
-		doResponse(apiErrorFromEchoError(he), c, he.Internal)
+		doResponse(apiErrorFromEchoError(he), c)
 		return
 	}
-	doResponse(sec, c, nil)
+	doResponse(sec, c)
```
Gundersanne, line 200: How about changing this to `doResponse(sec, c, he.Internal)` and keeping doResponse's signature as is?

internal/cloudapi/v2/errors.go
```diff
@@ -265,8 +264,8 @@ func (s *Server) HTTPErrorHandler(echoError error, c echo.Context) {
 	sec, ok := he.Message.(ServiceErrorCode)
 	if !ok {
 		// No service code was set, so Echo threw this error
-		doResponse(apiErrorFromEchoError(he), c)
+		doResponse(apiErrorFromEchoError(he), c, nil)
```
Gundersanne, line 267: Why nil and not he.Internal as well?

diaasami: oh right, echo can also set `Internal`, fixing.

Gundersanne (approved):
LGTM, thanks!

Gundersanne (dismissed):
Probably will need more rebases as other fixes roll in, but still LGTM :)

Gundersanne (approved):

PR #2155, opened 2021-12-22
Introduce Google GCE image type (RHEL, CentOS Stream) and support importing to GCP on premise

thozza:
- Introduce `gce` and `gce-rhui` image types for RHEL-9.0/8.6/8.5/8.4 and CentOS Stream 8/9
- Rework GCE image importing to use different API, which allows to directly insert the image into GCP
- Support importing images to GCP via Weldr API (on-premise)

### RHEL-8 (BYOS/RHUI) & RHEL-9 (BYOS) image differences compared to Google's image

* Used partition table is the same as for other image types. This means that there is a separate `/boot` partition and sizes of `/boot` and `/boot/efi` are the default ones (bigger than what Google uses). Also the BIOS boot partition is created, although the image is built as UEFI-only (no `grub2-pc*` packages are installed).

* SSH client is not configured. The reason is that while Google's kickstart was setting `ServerAliveInterval 420`, it did it using `sed` and since this option does not exist in the default SSH client configuration file, it had no effect. So technically the SSH client is not configured on images built by Google either.

* `google-compute-engine` configuration in `/etc/default/instance_configs.cfg.distro` is NOT created.

* `dhclient` is not configured in `/etc/dhcp/dhclient.conf`, because `NetworkManager` uses its own internal DHCP client implementation since RHEL-8, so `dhclient` is not used by default.

* GCP repository GPG keys are not imported on el9/c9s and GPG signatures are not checked, due to https://issuetracker.google.com/issues/223626963

* Packages:
  * `subscriptin-manager` is installed by default, even when the RHUI client is installed.
  * the `dracut-config-generic` package is installed, because it is standard part of the `x86_64` UEFI boot package set in distro definitions.
  * the `glibc-all-langpacks` package is installed on RHEL-8 images due to the way osbuild-composer depsolves package sets. This can not be workarounded at this moment (see https://github.com/osbuild/osbuild-composer/issues/2386).
  * any RPM dependencies of the packages mentioned above are naturally installed as well.

*  There is no RHEL-9 GCE RHUI image at this moment, since there are no Google RHUI client RPMs for RHEL-9 yet.

### Depends on

- [x] https://github.com/osbuild/osbuild-composer/pull/2381
- [x] https://github.com/osbuild/osbuild/pull/984
- [x] https://github.com/osbuild/image-builder/pull/331

This pull request includes:


thozza:
@jrusz FYI, this PR adds support for GCP to Weldr API and also adds the GCP integration test, which QE was considering in the past.

test/cases/api.sh
```diff
@@ -1153,15 +1150,15 @@ function verifyInGCP() {
...
+    --labels=gitlab-ci-test=true
+
   HOST=$($GCP_CMD compute instances describe "$GCP_INSTANCE_NAME" --zone="$GCP_ZONE" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')
 
   echo "⏱ Waiting for GCP instance to respond to ssh"
   _instanceWaitSSH "$HOST"
 
   # Verify image
-  _ssh="ssh -oStrictHostKeyChecking=no -i $GCP_SSH_KEY $SSH_USER@$HOST"
+  _ssh="$GCP_CMD compute ssh --strict-host-key-checking=no --ssh-key-file=$GCP_SSH_KEY --zone=$GCP_ZONE --quiet $SSH_USER@$GCP_INSTANCE_NAME --"
```
Gundersanne, line 1161: :heart_eyes:

Gundersanne (approved):
Just amazing

ondrejbudai (commented):
FWIW, I love this. :heart_eyes:

internal/cloud/gcp/compute.go
```diff
@@ -18,6 +18,43 @@ import (
...
+	{Type: common.StringToPtr(computepb.GuestOsFeature_SEV_CAPABLE.String())},
+	{Type: common.StringToPtr(computepb.GuestOsFeature_GVNIC.String())},
+}
+
+// GuestOsFeaturesByDistro returns the the list of Guest OS Features, which
+// should be used when importing an image of the specified distribution.
+//
+// In case the provided distribution does not have any specific Guest OS
+// Features list defined, nil is returned.
+func GuestOsFeaturesByDistro(distroName string) []*computepb.GuestOsFeature {
```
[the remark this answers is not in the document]
thozza: I agree, not sure where to put it. Since this depends on the GCP SDK as well as distro data...

achilleas-k (approved):
I have nothing to add over what's already been discussed so that's a :+1: from me.
Great work!

test/cases/api.sh
```diff
@@ -557,7 +557,7 @@ function createReqFileAWS() {
   "image_request": {
       "architecture": "$ARCH",
       "image_type": "${IMAGE_TYPE}",
-      "repositories": $(jq ".\"$ARCH\"" /usr/share/tests/osbuild-composer/repositories/"$DISTRO".json),
+      "repositories": $(jq ".\"$ARCH\" | .[] | select((has(\"image_type_tags\") | not) or (.\"image_type_tags\" | index(\"${IMAGE_TYPE}\")))" /usr/share/tests/osbuild-composer/repositories/"$DISTRO".json | jq -s .),
```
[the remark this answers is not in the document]
thozza: oh yeah. It took me some time to figure out 😄

thozza:
Incorporated changes related to RHSM configuration on BYOS and RHUI image types, after a discussion with the RHSM team. Rebased on top of the `main`.

image-types/rhel8/google-gce.md
```diff
@@ -0,0 +1,1907 @@
...
+* Used partition table is the same as for other image types. This means that there is a separate `/boot` partition and sizes of `/boot` and `/boot/efi` are the default ones (bigger than what Google uses). Also the BIOS boot partition is created, although the image is built as UEFI-only (no `grub2-pc*` packages are installed).
+
+* SSH client is not configured. The reason is that while Google's kickstart was setting `ServerAliveInterval 420`, it did it using `sed` and since this option does not exist in the default SSH client configuration file, it had no effect. So technically the SSH client is not configured on images built by Google either.
+
+* `google-compute-engine` configuration in `/etc/default/instance_configs.cfg.distro` is NOT created.
+
+* `dhclient` is not configured in `/etc/dhcp/dhclient.conf`, because `NetworkManager` uses its own internal DHCP client implementation since RHEL-8, so `dhclient` is not used by default.
+
+* Packages:
+  * `subscriptin-manager` is installed by default, even when the RHUI client is installed.
```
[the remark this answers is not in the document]
thozza: fixed

thozza:
I messed up when rebasing the last time and editing the `.gitlab-ci.yml`. I tried to split the integration test (Weldr) template into Fedora and RHEL, because GCE image type is not supported on Fedora yet. However my change resulted in Fedora not being tested as part of the integration test at all. 

Just as a side note, the reason was that the `extends` keyword is not able to merge arrays, but instead uses the last array when merging. To fix this, I had to use YAML anchors instead.

This should be now fixed...

thozza:
Resolved merge conflicts and rebased on top of `main`

internal/distro/rhel90/distro.go
```diff
@@ -1315,6 +1315,146 @@ func newDistro(distroName string) distro.Distro {
...
+					osbuild.NewModprobeConfigCmdBlacklist("floppy"),
+				},
+			},
+		},
+	}
+
+	gceImgType := imageType{
+		name:     "gce",
+		filename: "image.tar.gz",
+		mimeType: "application/x-tar",
```
achilleas-k, line 1441: Should the mimetype be for gzipped tar?  I don't even know if it matters tbh.

internal/distro/rhel90/distro_test.go
```diff
@@ -171,6 +171,14 @@ func TestFilenameFromType(t *testing.T) {
 				mimeType: "application/x-iso9660-image",
 			},
 		},
+		{
+			name: "gce",
+			args: args{"gce"},
+			want: wantResult{
+				filename: "image.tar.gz",
+				mimeType: "application/x-tar",
```
achilleas-k, line 179: Same as above.

internal/distro/rhel90/package_sets.go
```diff
@@ -705,6 +705,88 @@ func rhelEc2SapPackageSet(t *imageType) rpmmd.PackageSet {
...
+	// Some excluded packages are part of the @core group package set returned
+	// by coreOsCommonPackageSet(). Ensure that the conflicting packages are
+	// returned from the list of `Include` packages.
+	return ps.ResolveConflictsExclude()
+}
+
+// GCE BYOS image
+func gcePackageSet(t *imageType) rpmmd.PackageSet {
+	return gceCommonPackageSet(t)
+}
```
achilleas-k, line 788: Will the `gcePackageSet` change at some point?  Do we need this?

tools/test-case-generators/distro-arch-imagetype-map.json
```diff
@@ -1,4 +1,5 @@
 {
+<<<<<<< HEAD
```
achilleas-k, line 2: Leftover conflict markers.

achilleas-k (changes requested):
Looks good overall.  A few nits here and there, but also saw some merge conflict markers.

tools/test-case-generators/distro-arch-imagetype-map.json
thozza: Result of the gazillion rebases I had to do 😭  Thanks!

internal/distro/rhel90/package_sets.go
thozza: No, but there will be a RHUI variant of the package set which will inherit from `gceCommonPackageSet`, similar as there is in e.g. 8.6:
```go
// GCE BYOS image
func gcePackageSet(t *imageType) rpmmd.PackageSet {
	return gceCommonPackageSet(t)
}

// GCE RHUI image
func gceRhuiPackageSet(t *imageType) rpmmd.PackageSet {
	return rpmmd.PackageSet{
		Include: []string{
			"google-rhui-client-rhel8",
		},
	}.Append(gceCommonPackageSet(t))
}
```

The reason why it does not exist yet is that Google still didn't built a RHUI client packages for RHEL-9.

My reasoning for this piece of code was to:

1. Make a logical split between the BYOS and RHUI images package set.
2. Make the GCE code added by this PR ideally identical in all distro definitions.

I could just use the `gceCommonPackageSet` directly for the BYOS image or rename it to `gcePackageSet`. In such case, the RHUI image would be inheriting the BYOS package set and adding some packages on top of it.

I'm happy to rework this if you find it confusing and / or useless...

internal/distro/rhel90/distro.go
thozza: thanks, I'll change it to `application/gzip` based on [RFC6713](https://www.rfc-editor.org/rfc/rfc6713)

internal/osbuild2/tar_stage.go
```diff
@@ -29,13 +29,19 @@ type TarStageOptions struct {
...
 	// Enable support for SELinux contexts
-	SELinux bool `json:"selinux,omitempty"`
+	// Use pointer type because the option defaults to `true` and as a result,
+	// `false` value can not be omitted.
+	SELinux *bool `json:"selinux,omitempty"`
 
 	// Enable support for extended attributes
-	Xattrs bool `json:"xattrs,omitempty"`
+	// Use pointer type because the option defaults to `true` and as a result,
+	// `false` value can not be omitted.
```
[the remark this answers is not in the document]
thozza: hmm... a good point. 🤔  I'll reword them a bit...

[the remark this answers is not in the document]
thozza: I ended up removing these added comments. If anyone needs additional information, they can use blame and get to the relevant commit message which explains the change.

thozza:
@achilleas-k I addressed all of your comments except for the one related to `gcePackageSet`.

internal/distro/rhel90/package_sets.go
achilleas-k: > My reasoning for this piece of code was to:
> 
> 1. Make a logical split between the BYOS and RHUI images package set.
> 2. Make the GCE code added by this PR ideally identical in all distro definitions.
> 
> ...
> 
> I'm happy to rework this if you find it confusing and / or useless...

Sounds good.  Makes sense.  Let's keep it.

achilleas-k:
Restarted failed tests.  They _seemed_ to be network hiccups, but who knows.

achilleas-k (approved):
All good!!
Big PR is ready as far as I'm concerned. :)

PR #2159, opened 2022-01-03
test: split test script regression behind satellite

jabia99:
The test script regression-composer-works-behind-satellite.sh should be split in two parts to make it more visible and also because it contains two scenarios that are independent.

This pull request includes:

### DIFF
--- a/test/cases/regression-composer-works-behind-satellite-fallback.sh
+++ b/test/cases/regression-composer-works-behind-satellite-fallback.sh
@@ -33,32 +33,6 @@ function generate_certificates {
     sudo openssl x509 -req -in client.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out client.crt -days 365 -sha256
 }
 
-function cleanup {
-    # Make the cleanup function best effort
-    set +eu
-
-    greenprint "Display httpd logs"
-    cat /var/log/httpd/access_log
-    cat /var/log/httpd/error_log
-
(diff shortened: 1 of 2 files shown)


test/cases/regression-composer-works-behind-satellite-fallback.sh
```diff
@@ -0,0 +1,367 @@
...
+    sudo journalctl -xe --unit "${WORKER_UNIT}"
+
+    # Did the compose finish with success?
+    if [[ $COMPOSE_STATUS != FINISHED ]]; then
+        echo "Something went wrong with the compose. 😢"
+        exit 1
+    fi
+}
+
+# Part two, try that the fallback mechanism work
```
jrusz, line 323: It's not actually part two anymore so I'd change the comment a bit. Also I think it's worth keeping the check for `redhat.repo` file here and remove it if it's present. The system where this test would be ran can have `redhat.repo` present by default.

jrusz (changes requested):
Just added one small comment, I think we need to keep the function to check for `redhat.repo` and also put it back in the cleanup. Otherwise I think it's good.

test/cases/regression-composer-works-behind-satellite-fallback.sh
jabia99: thanks so much!

jrusz (approved):
LGTM

PR #2161, opened 2022-01-04
Packer improvements

ondrejbudai:
Three commits:
- bump the RHEL version to 8.5
- move configuration scripts from terraform to packer
- remove a broken monit check

The most important is the move of configuration. Previously, we had a huge cloud-init user-data configuration. With this change, all this configuration now lives inside the image and cloud-init must only provide a static file with several values. This should greatly help with testability.

The new image was tested manually and it's awesome. I also added some docs.

This pull request includes:

### DIFF
--- a/templates/packer/README.md
+++ b/templates/packer/README.md
@@ -1,0 +1,95 @@
+# osbuild-composer Packer configuration
+
+This directory contains a packer configuration for building osbuild-composer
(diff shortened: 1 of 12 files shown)


Gundersanne (approved):
Nice, this moves a lot more stuff upstream, lgtm!

PR #2166, opened 2022-01-04
Worker errors

kingsleyzissou:
This PR introduces structured errors for the workers as previously discussed. The structure
of the errors is:

```
{
  kind: Error, 
  id: VALIDATION_ERROR_NUMBER, 
  details: { 
    issues: [{...}, {...}] 
  }
}
```

### DIFF
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -1,0 +1,39 @@
+package clienterrors
+
+const (
+	ErrorNoDynamicArgs       ClientErrorCode = 1
+	ErrorInvalidTargetConfig ClientErrorCode = 2
+	ErrorSharingTarget       ClientErrorCode = 3
+	ErrorInvalidTarget       ClientErrorCode = 4
+
+	ErrorDNFError             ClientErrorCode = 1001
+	ErrorDepsolveJob          ClientErrorCode = 1002
+	ErrorDepsolveDependency   ClientErrorCode = 1003
+	ErrorParsingDynamicArgs   ClientErrorCode = 1004
+	ErrorManifestGeneration   ClientErrorCode = 1005
+	ErrorManifestDependency   ClientErrorCode = 1006
+	ErrorBuildJob             ClientErrorCode = 1007
+	ErrorUploadingImage       ClientErrorCode = 1008
+	ErrorRegisteringImage     ClientErrorCode = 1009
+	ErrorKojiFailedDependency ClientErrorCode = 10010
+	ErrorKojiBuild            ClientErrorCode = 1011
+	ErrorKojiInit             ClientErrorCode = 1012
+	ErrorKojiFinalize         ClientErrorCode = 1013
+	ErrorInvalidConfig        ClientErrorCode = 1014
+)
+
+type ClientErrorCode int
+
+type Error struct {
+	ID      ClientErrorCode `json:"id"`
+	Reason  string          `json:"reason"`
+	Details interface{}     `json:"details"`
+}
+
+func WorkerClientError(code ClientErrorCode, reason string, details ...interface{}) *Error {
+	return &Error{
+		ID:      code,
+		Reason:  reason,
+		Details: details,
+	}
+}
--- a/internal/worker/json.go
+++ b/internal/worker/json.go
@@ -7,6 +7,7 @@ import (
 	osbuild "github.com/osbuild/osbuild-composer/internal/osbuild2"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/target"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
 )
 
 //
@@ -27,6 +28,7 @@ type OSBuildJobResult struct {
 	OSBuildOutput *osbuild.Result        `json:"osbuild_output,omitempty"`
 	TargetResults []*target.TargetResult `json:"target_results,omitempty"`
 	TargetErrors  []string               `json:"target_errors,omitempty"`
+	JobError      *clienterrors.Error    `json:"job_error,omitempty"`
 	UploadStatus  string                 `json:"upload_status"`
 	PipelineNames *PipelineNames         `json:"pipeline_names,omitempty"`
 }
@@ -39,9 +41,10 @@ type KojiInitJob struct {
 }
 
 type KojiInitJobResult struct {
-	BuildID   uint64 `json:"build_id"`
-	Token     string `json:"token"`
-	KojiError string `json:"koji_error"`
+	BuildID   uint64              `json:"build_id"`
+	Token     string              `json:"token"`
+	KojiError string              `json:"koji_error"`
+	JobError  *clienterrors.Error `json:"job_error,omitempty"`
 }
 
 type OSBuildKojiJob struct {
@@ -55,13 +58,14 @@ type OSBuildKojiJob struct {
 }
 
 type OSBuildKojiJobResult struct {
-	HostOS        string          `json:"host_os"`
-	Arch          string          `json:"arch"`
-	OSBuildOutput *osbuild.Result `json:"osbuild_output"`
-	PipelineNames *PipelineNames  `json:"pipeline_names,omitempty"`
-	ImageHash     string          `json:"image_hash"`
-	ImageSize     uint64          `json:"image_size"`
-	KojiError     string          `json:"koji_error"`
+	HostOS        string              `json:"host_os"`
+	Arch          string              `json:"arch"`
+	OSBuildOutput *osbuild.Result     `json:"osbuild_output"`
+	PipelineNames *PipelineNames      `json:"pipeline_names,omitempty"`
+	ImageHash     string              `json:"image_hash"`
+	ImageSize     uint64              `json:"image_size"`
+	KojiError     string              `json:"koji_error"`
+	JobError      *clienterrors.Error `json:"job_error,omitempty"`
 }
 
 type KojiFinalizeJob struct {
@@ -76,7 +80,8 @@ type KojiFinalizeJob struct {
 }
 
 type KojiFinalizeJobResult struct {
-	KojiError string `json:"koji_error"`
+	KojiError string              `json:"koji_error"`
+	JobError  *clienterrors.Error `json:"job_error,omitempty"`
 }
 
 // PipelineNames is used to provide two pieces of information related to a job:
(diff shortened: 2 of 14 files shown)


internal/worker/clienterrors/errors.go
```diff
@@ -0,0 +1,116 @@
...
+
+type ServiceErrorCode int
+
+type serviceError struct {
+	code       ServiceErrorCode
+	httpStatus int
+	reason     string
+}
+
+type serviceErrors []serviceError
```
achilleas-k, line 49: If this is a `map[ServiceErrorCode]ServiceError` instead of a slice, we wouldn't need the `find()` function looping through all the errors (do we need it though?)

internal/worker/json.go
```diff
@@ -23,12 +24,14 @@ type OSBuildJob struct {
...
-	TargetResults []*target.TargetResult `json:"target_results,omitempty"`
-	TargetErrors  []string               `json:"target_errors,omitempty"`
-	UploadStatus  string                 `json:"upload_status"`
-	PipelineNames *PipelineNames         `json:"pipeline_names,omitempty"`
+	Success         bool                   `json:"success"`
+	OSBuildOutput   *osbuild.Result        `json:"osbuild_output,omitempty"`
+	TargetResults   []*target.TargetResult `json:"target_results,omitempty"`
+	TargetErrors    []string               `json:"target_errors,omitempty"`
+	TargetJobErrors []clienterrors.Error   `json:"target_job_errors,omitempty"`
+	JobError        *clienterrors.Error    `json:"job_error"`
```
achilleas-k, line 32: Is this for backwards compatibility?  Adding fields to the existing result object instead of modifying?
It's getting a bit messy with the double result checking of the result in the worker, but maybe this is necessary...

achilleas-k (commented):
Love the structured error messages.  It's getting a bit messy with the duplication of result/error checking in the worker.  Not sure how to best handle it if we want to be backwards compatible though.

Maybe we could only use the new errors from now on in the worker (so keep the old field empty) and in composer try to read both, so we keep backwards compatibility in composer instead.

ondrejbudai:
I have the same feelings as Achilleas - it's very nice but also slightly messy with the duplicated code. On the other hand, I would like to redo the worker API at one point, so maybe let's keep it dirty for now and clean it up in v2.

internal/worker/json.go
kingsleyzissou: Yeah my thoughts there were to not break backwards compatibility, but happy enough to modify it instead, if we are happy to do that.

kingsleyzissou: Or like you say, return an empty result and check in composer instead.

internal/worker/clienterrors/errors.go
kingsleyzissou: I don't think we do, I just copied what we were doing in other parts of the code. I'll change it to `map[ServiceErrorCode]ServiceError`. I'll get these changes done now

internal/worker/json.go
kingsleyzissou: It looks like the only time we read the results object in composer for errors is in the Koji api  i.e. `if initResult.KojiError != ""`) For the rest it's only ever happening in the workers.

Gundersanne (commented):
Nice!

The backward compatibility is a bit messy but unavoidable i think. We could get around that a bit if we can fill the new error fields based on the old ones in results when retrieving a result from the worker server and contain all the mess within worker/v1 which we could then more easily drop (just have a separate code for that and stick the errors into details or something).

Something we also need to consider is how composerapis (weldr/cloud/kojiapi) will use/translate these errors into an error meaningful to return to composerapi clients (as a composestatus). Probably we'll just end up extracting the string/details into something readable and sending that along, and not check the specific codes for now? The codes might be useful for metrics perhaps? Would any part of composer other than the workerclient and workerserver need the clienterrorcodes?

internal/worker/clienterrors/errors.go
Gundersanne: Ah yea I implemented the other parts using find(); it's a bit awkward and this code is duplicated a bunch. We could consider unifying some of it in common.

Regardless this slice is just internal to the errors.go file, so if it's easier to use a map then please do.

internal/worker/clienterrors/errors.go
```diff
@@ -0,0 +1,116 @@
...
+	ErrorNotHTTPError         ServiceErrorCode = 10001
+	ErrorServiceErrorNotFound ServiceErrorCode = 10002
+	ErrorMalformedOperationID ServiceErrorCode = 10003
+)
+
+type ServiceErrorCode int
+
+type serviceError struct {
+	code       ServiceErrorCode
+	httpStatus int
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah good point, I guess the idea behind it was to make it easier down the line to differentiate between 4xx and 5xx errors for the metrics and thought that was cleaner than just putting in a string with "4xx" or "5xx", but that could be an alternative

kingsleyzissou:
> The backward compatibility is a bit messy but unavoidable i think. We could get around that a bit if we can fill the new error fields based on the old ones in results when retrieving a result from the worker server and contain all the mess within worker/v1 which we could then more easily drop (just have a separate code for that and stick the errors into details or something).
Yeah I thought about doing that, but it gets a little tricky discerning what type of error we encountered especially in the osbuild jobs

> The codes might be useful for metrics perhaps? Would any part of composer other than the workerclient and workerserver need the clienterrorcodes?
Yeah exactly, the codes would be useful for the metrics. I'm not sure they are needed anywhere else other than the worker client and worker server.

teg:
> Something we also need to consider is how composerapis (weldr/cloud/kojiapi) will use/translate these errors into an error meaningful to return to composerapi clients (as a composestatus). Probably we'll just end up extracting the string/details into something readable and sending that along, and not check the specific codes for now?

I had hoped we could make the errors structured to the extent possible. That way the highest level can turn them into human readable strings in whatever way fits their context (with translations etc).

achilleas-k (commented):
Much cleaner.  I think this covers everything.

kingsleyzissou:
Cool, yeah if we're happy with that I can squash the commit and then add some tests.

I'll also do some more digging to see if we can check if this covers all the cases for backwards compatibility.

internal/worker/clienterrors/errors.go
```diff
@@ -0,0 +1,84 @@
...
+	ErrorKojiFailedDependency: {ErrorKojiFailedDependency, "Koji job has a failed dependency"},
+	ErrorKojiBuild:            {ErrorKojiBuild, "Koji build error"},
+	ErrorKojiInit:             {ErrorKojiInit, "Koji init error"},
+	ErrorKojiFinalize:         {ErrorKojiFinalize, "Koji finalize error"},
+	ErrorInvalidConfig:        {ErrorInvalidConfig, "Configuration error"},
+	ErrorClientErrorNotFound:  {ErrorClientErrorNotFound, "Error does not exist"},
+}
+
+type Error struct {
+	Kind    string          `json:"kind"`
```
Gundersanne, line 63: tbh I think for the client we can drop the `Kind` field as well.

internal/worker/clienterrors/errors.go
```diff
@@ -0,0 +1,84 @@
...
+	ID      ClientErrorCode `json:"id"`
+	Code    string          `json:"code"`
```
Gundersanne, line 65: Same with code, I think in this case just the `id` is enough.

Gundersanne (commented):
Hm I tried going through it again and it does end up feeling a little artificial since the reasons are just always set by the client and not sure if we'll ever reconstruct the errors by code in composer.

I'm almost inclined to just simplify this a lot, just standerdize on client errors being format 
```
type Error struct {
	ID           ClientErrorCode `json:"id"`
	Reason  string                  `json:"reason"`
	Details    interface{}          `json:"details,omitempty"`

}
```

and ending up with https://gist.github.com/Gundersanne/83c54a103657e57d7b16388fe3d9169f. I think the main thing is that we use an errorobject consistent across all job types with an error code that's stable.

internal/worker/clienterrors/errors.go
kingsleyzissou: I agree with this, it was starting to seem very redundant.

Gundersanne (commented):
Nice, I think this is getting pretty close. ig someone else should review as well.

I do think it might be more useful to approach the backward compatibility in a different way to centralize all the mess inside of the worker server. It would be nice if immediately after this patch: worker clients can use the new field of setting the error for a job, the apis can use the new JobError field, and the worker server handles any backfilling of JobError when reading old results (like we do for the osbuildjobresult success flag). worker.Server.JobStatus() would get expanded quite a bit with a lot of type casting (and some heuristics for old errors -> new errors, or we can just have a specific clienterrorcode if an error was converted). But all the mess would be contained + thrown away when/if we switch to worker v2. Curious to see what others think though.

internal/worker/json.go
```diff
@@ -27,6 +28,7 @@ type OSBuildJobResult struct {
 	OSBuildOutput *osbuild.Result        `json:"osbuild_output,omitempty"`
 	TargetResults []*target.TargetResult `json:"target_results,omitempty"`
 	TargetErrors  []string               `json:"target_errors,omitempty"`
+	JobError      *clienterrors.Error    `json:"job_error,omitempty"`
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah I really like that approach. We actually do need that for when we go ahead and implement the metrics, so I'll add this in the meantime while we wait for more comments.

internal/worker/clienterrors/errors.go
```diff
@@ -0,0 +1,39 @@
...
+	ErrorManifestGeneration   ClientErrorCode = 1005
+	ErrorManifestDependency   ClientErrorCode = 1006
+	ErrorBuildJob             ClientErrorCode = 1007
+	ErrorUploadingImage       ClientErrorCode = 1008
+	ErrorRegisteringImage     ClientErrorCode = 1009
+	ErrorKojiFailedDependency ClientErrorCode = 10010
+	ErrorKojiBuild            ClientErrorCode = 1011
+	ErrorKojiInit             ClientErrorCode = 1012
+	ErrorKojiFinalize         ClientErrorCode = 1013
+	ErrorInvalidConfig        ClientErrorCode = 1014
```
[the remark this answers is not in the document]
kingsleyzissou: I think it might be useful to split the errors between errors caused by us and errors because of user input at a later stage. So I can move the `ErrorDepsolveDependency` to the top section. Do you see any others that might need to be switched around?

kingsleyzissou:
I'm closing this in favour of https://github.com/osbuild/osbuild-composer/pull/2192

PR #2174, opened 2022-01-10
Packer fixes

Gundersanne:


teg (approved):

Gundersanne:
@teg packer is green here now; PACKER_IMAGE_USERS is also defined now so I think after this packer on appsre should be green and usable.

teg (approved):

PR #2176, opened 2022-01-11
Use lz4 to compress the rootfs

lavocatt:
This pull request includes:


thozza:
Hi.

You changed only the image test cases, but not the actual image definitions (in `internal/distro/`). Unit tests are failing because the manifest returned by composer does not match the manifest in image test case after you changed them.

In addition, these changes should not be most probably applied to old releases, only to RHEL 9.0 and 8.6, which are still in development.

internal/distro/rhel90beta/stage_options.go
```diff
@@ -281,8 +277,7 @@ func bootISOMonoStageOptions(kernelVer string, arch string) *osbuild.BootISOMono
 		RootFS: osbuild.RootFS{
 			Size: 9216,
 			Compression: osbuild.FSCompression{
-				Method:  "xz",
-				Options: comprOptions,
+				Method: "lz4",
```
thozza, line 280: RHEL-90 Beta definition should not be changed as it was already released.

lavocatt: Oh yes! I'll change that :) Thanks

lavocatt:
Well, it seems that none of our images are able to boot with lz4 images. I've tried rhel9 and centos stream 8 without success. I think this will require some more investigation.

thozza:
@lavocatt is this PR still relevant? Looks like it should be a draft or perhaps even closed if not relevant any more (can be reopen at any time).

PR #2179, opened 2022-01-12
Tracing: deactivate iostat by default

lavocatt:
When iostat runs it can prevent the script from exiting from a failure.
Until we have a better solution for this, prevent iostat to start.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -265,10 +265,8 @@ New OSTree:
   stage: test
   extends: OSTree
   script:
-    - schutzbot/start_iostats.sh
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/ostree-ng.sh
-    - schutzbot/stop_iostats.sh
   parallel:
     matrix:
       - RUNNER:
@@ -280,10 +278,8 @@ OSTree simplified installer:
   stage: test
   extends: OSTree
   script:
-    - schutzbot/start_iostats.sh
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/ostree-simplified-installer.sh
-    - schutzbot/stop_iostats.sh
   parallel:
     matrix:
       - RUNNER:
@@ -295,10 +291,8 @@ OSTree raw image:
   stage: test
   extends: OSTree
   script:
-    - schutzbot/start_iostats.sh
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/ostree-raw-image.sh
-    - schutzbot/stop_iostats.sh
   parallel:
     matrix:
       - RUNNER:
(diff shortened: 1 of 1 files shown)


lavocatt:
I integrated this commit also in this pr https://github.com/osbuild/osbuild-composer/pull/2176 and some jobs hanged. So iostat might not be the faulty part.

achilleas-k:
> I integrated this commit also in this pr #2176 and some jobs hanged. So iostat might not be the faulty part.

Interesting.  I wonder what else might have changed to be causing the issue.

lavocatt:
On slack @jrusz mentioned that "only the Installer job was stuck and that's because of the journal watcher not being killed properly on exit. It's missing this https://github.com/osbuild/osbuild-composer/commit/bc1a7dc30c4f1729d23bf028f44e35f74ed7d7e6"

I'm gonna try that, thanks @jrusz

jrusz:
@lavocatt Why did you decide to close this PR?

lavocatt:
Oh yes, because I wanted to apply the patch in an other one. Sorry, I should name why I'm closing stuff..

jrusz:
Oh ok, I wanted to ask about this, because IMO it's currently making the CI unusable, lots of jobs waiting for the 4 hour timeout so it would be nice if we can solve this sooner rather than later.

lavocatt:
Yes you are right, In the end I'll keep this PR opened to not forget the subject

jrusz (changes requested):
Since we're deactivating it, could you please remove it everywhere? I can see more instances of the `start_iostats.sh` and `stop_iostats.sh` eg line 48,50 ... And also please remove `iostats.json` from the artifacts too :)

jrusz:
To clarify I have no problem with keeping it and running it, the problem is with how it's being ran now. If the main script fails, the iostat script running in background is preventing the CI from failing and it waits for timeout which results in dozens of jobs being stuck waiting for timeout and IMO making the CI unusable.

Since I saw you remove it from some of the jobs I thought why not remove it from all the jobs? I'm talking about the `.gitlab-ci.yml` file, the scripts themselves can be left alone.

The other option that I would go for is to run this iostat script from within the main script of each job, for example by sourcing it and then having a `trap` to kill it on exit. Similar to how we do it with the journal watcher. This could maybe be added to `tools/set-env-vars.sh` I think because it's sourced in every script AFAIK. I can try creating a PoC of what I mean, I'm not 100% sure it would work as I think it would but at least I could figure something out.

jrusz:
Made one more test, now I'm 100% sure iostats are freezing CI https://github.com/osbuild/osbuild-composer/pull/2198 so we need to do something about it.

lavocatt:
> Since I saw you remove it from some of the jobs I thought why not remove it from all the jobs? I'm talking about the .gitlab-ci.yml file, the scripts themselves can be left alone.

That makes sense. cleaning the gitlab yml file and keeping the file is nice.

I'd be also happy with the other solution, it looks neat.

PR #2187, opened 2022-01-12
dnf-json: Add remote_location to dump package.

ldjebran:


teg (commented):
Could you explain in the commit message what this will be used for and why it is needed?

ldjebran:
@teg thanks for reviewing this PR. 

PR comment updated.

dnf-json
```diff
@@ -200,6 +200,7 @@ class Solver():
                 "summary": package.summary,
                 "description": package.description,
                 "url": package.url,
+                "remote_location": package.remote_location(),
```
teg, line 203: In the `depsolve()` call we use `repo_id` to match packages to repositories. Should we do the same thing here?

ldjebran: Thanks @teg for the hint, absolutely we can use that to identify back the package repository, especially we are providing that repository ids.

Updated this PR to reflect that.

teg: Thanks!

gicmo:
```
echo "The dump function will be used to search packages and as we are implementing third party repositories, adding the repo_id of the package will allow us to identify the package repository used." | wc -c

192
```

We normally try to stay within the usual 72 char limit for the body.

ldjebran:
> ```
> echo "The dump function will be used to search packages and as we are implementing third party repositories, adding the repo_id of the package will allow us to identify the package repository used." | wc -c
> 
> 192
> ```
> 
> We normally try to stay within the usual 72 char limit for the body.

How adding repo_id may be related to the restriction (as the already returned info may be far greater than the 72 char )? 
PS: I understand that some bytes will be used when receiving that info, but think it's not that critical.
The repo_id is critically needed to analyse from which repo the package is coming rom.

gicmo:
My comment was about the commit message.

ondrejbudai (approved):

PR #2191, opened 2022-01-13
tools/appsre-build-worker-packer: Run entire workflow in container

Gundersanne:
Avoids the issue with artifacts that jenkins can't clean up.


ondrejbudai (approved):
Let's try this. I think I would like to see this cleaned up a little but it can be done in a follow-up.

PR #2192, opened 2022-01-13
[WIP]: Worker errors alt backwards compatibility

kingsleyzissou:
This pull request includes:

Alternative implementation of errors and handling backwards compatibility inside composer.


internal/worker/server.go
```diff
@@ -289,6 +361,15 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
 		}
 	}
 
+	// var jobResult OSBuildJobResult
+	_, _, err = s.JobStatus(jobId, &result)
+	if err != nil {
+		return fmt.Errorf("error finding job status: %v", err)
+	}
+
+	_ = getStatusCode(result)
```
Gundersanne, line 370: Hm not sure if we should store the status code in the jobqueue. The error will always be in the jobresult and is always able to be interpreted by composer

internal/worker/server.go
```diff
@@ -131,11 +133,68 @@ func (s *Server) JobStatus(id uuid.UUID, result interface{}) (*JobStatus, []uuid
...
+				JobError: clienterrors.WorkerClientError(clienterrors.ErrorOldResultCompatible, r.Error),
+			}
+		}
+	}
+
+	if r, ok := result.(*DepsolveJobResult); ok {
+		if r.JobResult.JobError == nil && r.Error != "" {
+			var jobErr *clienterrors.Error
+			if r.ErrorType == DepsolveErrorType {
+				jobErr = clienterrors.WorkerClientError(clienterrors.ErrorDNFError, r.Error)
```
thozza, line 172: Based on the if statement, shouldn't this branch use the `clienterrors.ErrorDepsolveJob` instead of the else branch?

internal/worker/server.go
kingsleyzissou: Oh the idea wasn't to store it in the job queue, that's just where we are collecting the metrics:
https://github.com/osbuild/osbuild-composer/blob/main/internal/jobqueue/dbjobqueue/dbjobqueue.go#L366-L367

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -42,11 +43,14 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	if err != nil {
 		switch err.(type) {
 		case *rpmmd.DNFError:
-			result.ErrorType = worker.DepsolveErrorType
+			result.JobResult = worker.JobResult{
```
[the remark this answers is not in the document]
kingsleyzissou: Mind blown. Okay that kind of makes the one commit pretty redundant. I'll clean that up.

internal/worker/server.go
Gundersanne: ah I wonder if that could move to the worker server when finishing the job up.

kingsleyzissou: Yeah we can try that. I'll move this to a separate PR and do that after this one lands

internal/worker/server.go
```diff
@@ -131,11 +133,68 @@ func (s *Server) JobStatus(id uuid.UUID, result interface{}) (*JobStatus, []uuid
...
+		}
+	}
+
+	if r, ok := result.(*OSBuildKojiJobResult); ok {
+		if r.JobResult.JobError == nil && r.KojiError != "" {
+			r.JobResult = JobResult{
+				JobError: clienterrors.WorkerClientError(clienterrors.ErrorOldResultCompatible, r.KojiError),
+			}
+		}
+	}
```
achilleas-k, line 158: This is more of a personal choice, but I find that case switching with types is more readable and easier to manager:
```go
	switch r := result.(type) {
	case *KojiInitJobResult:
		if r.JobResult.JobError == nil && r.KojiError != "" {
			r.JobResult = JobResult{
				JobError: clienterrors.WorkerClientError(clienterrors.ErrorOldResultCompatible, r.KojiError),
			}
		}
	case *KojiFinalizeJobResult:
		if r.JobResult.JobError == nil && r.KojiError != "" {
			r.JobResult = JobResult{
				JobError: clienterrors.WorkerClientError(clienterrors.ErrorOldResultCompatible, r.KojiError),
			}
		}
	case *OSBuildKojiJobResult:
		...
	}
```

You get the `r` variable cast to the compared type immediately, it avoids needing to check `ok` every time, it bundles all the type checks into one block, and it makes it easy to define a `default` case (if necessary).

Practically and functionally it's the same though, so I don't mind keeping it this way.  I just find this easier to manager personally.

achilleas-k (commented):
Looks good in general.  Outside of the general discussions, I have a couple of minor comments (nitpicks and style choices).

internal/weldr/compose_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	jobResult := worker.OSBuildJobResult{TargetErrors: []string{"Upload error"}}
+	rawResult, err := json.Marshal(jobResult)
+	require.NoError(t, err)
+	err = api.workers.FinishJob(token, rawResult)
+	require.NoError(t, err)
+
+	jobStatus, _, err := api.workers.JobStatus(jobId, &jobResult)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
kingsleyzissou: Uff good catch. Thanks!

internal/worker/server.go
kingsleyzissou: Yeah happy to make that change, I agree that it would make things more readable and easier to manage.

internal/worker/server.go
kingsleyzissou: @thozza, do you have a suggestion for the code in the `else` branch here?
I think the same thing would apply for this bit of code too:
https://github.com/osbuild/osbuild-composer/pull/2192/files#diff-cdbd73affc81d7f2669bf94f8b6fecae6502db28431663452e9fac613728bff1L488-R509

kingsleyzissou: And I guess here too:
https://github.com/osbuild/osbuild-composer/pull/2192/files#diff-e0ac8df111d90787f24679627e7beea4774877480640e1494d296a4b2096b1b4R44-R49

thozza: I mean the `clienterrors.ErrorDNFError` in the `else` branch looked reasonable to me. 

To your other questions, I guess you should distinguish if the actual depsolving of the package set failed, or if there is some other DNF error and return the relevant error type. Using `clienterrors.ErrorDNFError` for non-depsolving errors even in the depsolving job seems OK to me.

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -42,11 +43,10 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	if err != nil {
 		switch err.(type) {
 		case *rpmmd.DNFError:
-			result.ErrorType = worker.DepsolveErrorType
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorDepsolveJob, err.Error())
```
Gundersanne, line 46: So the DepsolveError is a DNFError, see https://dnf.readthedocs.io/en/latest/api_exceptions.html.

So ErrorDNFDepsolveError seems more apprioriate here? This is the user input failure we don't need to pay attention to.

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -42,11 +43,10 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
...
 		case error:
-			result.ErrorType = worker.OtherErrorType
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorDNFError, err.Error())
```
Gundersanne, line 48: In the general case something went wrong, so a generic ErrorDepsolveJob here would indicate a legit failure that something is up and dnf-json is crashing somewhere it shouldn't.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -63,6 +75,9 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	// In all cases it is necessary to report result back to osbuild-composer worker API.
 	defer func() {
+
```
Gundersanne, line 78: nitpick, but no prepended newline.

internal/cloudapi/v2/v2.go
```diff
@@ -474,34 +474,44 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
 			}
-			jobResult.Error = "Error in depsolve job dependency"
+			jobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorDepsolveDependency, "Error in depsolve job dependency")
 			return
 		}
 
 		manifest, err := imageType.Manifest(b, options, repos, depsolveResults.PackageSpecs, seed)
 		if err != nil {
-			jobResult.Error = "Error generating manifest"
+			reason := "Error generating manifest"
```
Gundersanne, line 508: nitpick, but no need to the reason  variable here, just pass it directly.

Gundersanne (changes requested):
Looks good! some final comments

internal/worker/server.go
Gundersanne: I seemed to have caused a bunch of confusion with https://github.com/osbuild/osbuild-composer/pull/2192#discussion_r786743109

So how about https://github.com/Gundersanne/osbuild-composer/commit/b6bb949bfcbb502d41b9a969d9b37f249152ecab.

rpmmd classifies anything that came from dnf-json (where the http call didn't succeed and returned normally) as a "DNFError". Crrently anything classified as a DNFError is considered bad user input (though I'm not entirely convinced this should be the case).

If dnf-json crashes I'm not actually sure where it'll fail, will httpc.Post fail? Regardless, if it's not a "DNFError" it's an RPMMDError. And then let's drop the confusing ErrorDepsolveJob.

I think user input failures are definitely ErrorDNFDepsolveError, corresponding with https://dnf.readthedocs.io/en/latest/api_exceptions.html#dnf.exceptions.DepsolveError ; unsure about the MarkingErrors.

Gundersanne: We could just not change the behaviour in this PR and keep the distinction DNFError vs RPMMDError only, and then if it's a DNFError we regard it as user input failure always like we do now. And ameliorate the behavior in some future work.

cmd/osbuild-worker/jobimpl-depsolve.go
Gundersanne: see https://github.com/osbuild/osbuild-composer/pull/2192#discussion_r786789901

Gundersanne (approved):
Very nice, thank you!!

Gundersanne (approved):

PR #2193, opened 2022-01-14
tools/provision: community.general ansible collection for RHEL 8.6

achilleas-k:
Required for the `json_query` which we use in our test playbooks.  Also requires python3-jmespath.

This is the same change that we made for RHEL 9 ab5d560b1c02e79166b377c344a78e9f89c91901 and it is now required for RHEL 8.6 as well.

The conditions are a bit weird, but it's the cleanest way I could figure out to compare the versions properly to catch future versions as well.

### DIFF
--- a/test/data/ansible/check_ostree.yaml
+++ b/test/data/ansible/check_ostree.yaml
@@ -43,12 +43,12 @@
 
     # case: check ostree commit correctly updated
     - name: get deployed ostree commit
-      command: rpm-ostree status --json
+      shell: rpm-ostree status --json | jq -r '.deployments[0].checksum'
       register: result_commit
 
     - name: make a json result
       set_fact:
-        deploy_commit: "{{ result_commit.stdout | from_json | json_query('deployments[0].checksum') }}"
+        deploy_commit: "{{ result_commit.stdout }}"
 
     - name: check commit deployed and built
       block:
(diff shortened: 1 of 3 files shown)


henrywang (dismissed):
LGTM. Thanks!

Gundersanne:
RHEL 8.6 nightlies are borked, python3-jmespath installs into python3.6 packages; while the interpreter itself is python3.8 :( 

pip3.8 install jmespath as a workaround works

achilleas-k:
> RHEL 8.6 nightlies are borked, python3-jmespath installs into python3.6 packages; while the interpreter itself is python3.8 :(
> 
> pip3.8 install jmespath as a workaround works

**discussed in chat**
This only works when the ansible interpreter happens to be python 3.8.  On any system where 3.8 isn't available it would fail.  It also means it would break as soon a ansible switches to 3.9.

I've instead removed all (4) cases where we use `json_query` in our playbooks and instead filter the output using `jq` on the shell.

achilleas-k:
Is this PR too early?  It seems the change to use `ansible-core` in 8.6 aren't working because it's not available in 8.6 yet, but we know it's coming..?

henrywang:
> Is this PR too early? It seems the change to use `ansible-core` in 8.6 aren't working because it's not available in 8.6 yet, but we know it's coming, yes?

It's there in RHEL 8.6 nightly. http://download-node-02.eng.bos.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.6.0/compose/AppStream/x86_64/os/Packages/ansible-core-2.12.1-1.el8.x86_64.rpm

achilleas-k:
Hm... so it depends when this will land in the EPEL repo because that's where we install it from for the tests

Gundersanne:
Included in #2244

PR #2194, opened 2022-01-14
Dnf json

Gundersanne:
The goal is to be able to deploy dnf-json as a standalone service.

- cachedir should be able to be set by configuration rather than passed as argument
- dnf-json (at least the container) should be able to accept http connections

### DIFF
--- a/containers/osbuild-composer/entrypoint.py
+++ b/containers/osbuild-composer/entrypoint.py
@@ -108,6 +108,13 @@ class Cli(contextlib.AbstractContextManager):
             dest="dnf_json",
             help="Disable dnf-json",
         )
+        self._parser.add_argument(
+            "--dnf-json-port",
+            type=int,
+            default=0,
+            dest="dnf_json_port",
+            help="Specify the port dnf-json should listen on",
+        )
 
         self._parser.set_defaults(
             builtin_worker=False,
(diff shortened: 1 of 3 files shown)


dnf-json
```diff
@@ -305,10 +312,15 @@ class DnfJsonRequestHandler(BaseHTTPRequestHandler):
                 arguments = call["arguments"]
                 repos = arguments.get("repos", {})
                 arch = arguments["arch"]
-                self.cache_dir = arguments["cachedir"]
-                cache_state = CacheState.load_cache_state_from_disk(self.cache_dir)
                 module_platform_id = arguments["module_platform_id"]
 
+                # If dnf-json is run as a service, we don't want users to be able to set the
+                # cache
+                self.cache_dir = os.environ.get("OVERWRITE_CACHE_DIR", arguments.get("cachedir", ""))
```
lavocatt, line 319: BTW, maybe you know why, but I'm wondering, why do we want this to be a parameter in the first place ? What is the upside of having the cache being different for different requests ?

lavocatt (commented):
This looks nice! Thanks for adding this.

dnf-json
lavocatt: I think I do remember having a conversation at some point with Ondrej about the cache folders. Something about the fact that requests sharing the cache folder but having different target architecture are making dnf download everything twice, which slows the process. But I'm not sure if I remember correctly.

Gundersanne: Ah then we probably want a cache dir per architecture. I do tihnk it's better if dnf-json sets it's own cache rather than users specifying it too I agree. Not sure if there's an argument against it.

Gundersanne: Ah i think the idea is to be able to set the cache dir in 1 place (composer's CACHE_DIRECTORY), and have that be valid everywhere. CACHE_DIRECTORY is also set in dnf-json's systemd unit file but then not used anywhere. It's all a little confusing tbh, but I think we should keep the ability to set it  in the request.

lavocatt (approved):
LGTM!

containers/osbuild-composer/entrypoint.py
```diff
@@ -286,14 +305,20 @@ def run(self):
             if self.args.dnf_json:
                 proc_dnf_json = self._spawn_dnf_json()
 
-            proc_composer = self._spawn_composer(sockets)
+            if any([self.args.weldr_api, self.args.composer_api, self.args.local_worker_api, self.args.remote_worker_api]):
```
lavocatt, line 308: These modifications are made to allow dnf_json to be started without composer ?

lavocatt (commented):
That looks neat. Just one question to be sure I understand this all :)

containers/osbuild-composer/entrypoint.py
Gundersanne: Yea exactly. We'd only wanna run dnf-json in some cases.

lavocatt (approved):

PR #2195, opened 2022-01-17
templates/packer: Correct priority for worker rpms

Gundersanne:
Lower priority means higher, currently the images built through AppSRE's
infra install the worker from epel.


Gundersanne:
Verified that the correct/newer rpms are now installed on an ami prepped with this playbook.

PR #2198, opened 2022-01-18
CI: Updates to ensure smoother running

jrusz:
This pull request includes:


jrusz:
Changing the way how iostats are started and stopped because the previous way was causing CI jobs to hang on failure and wait for timeout. @atodorov could you please also take a look if you see any problem with this implementation?

atodorov (approved):

lavocatt (approved):
Thanks a lot for doing this.

jrusz:
There was this line in almost every script which removes all traps on EXIT `trap - EXIT` which removed even the trap for the iostats and so even passing jobs got stuck. I removed the line I don't think it will cause any problems but let's see what the CI thinks.

lavocatt (approved):
Yes that makes sense to not remove traps in the scripts.

jrusz:
This approach won't work that easily.. The `set-env-variables.sh` is not only called in `deploy.sh` so it starts and stop there, then does kinda the same in the test script and then each test script runs 'provision.sh' which also sources 'set-env-variables.sh' ... We'll have to put it somewhere else or figure out something different.

jrusz:
I have a suggestion. Let's just remove iostats alltogether from the CI temporarily (just the `.gitlab-ci.yml`), merge this so people can work with CI reliably and then figure out where to put the iostats in a different PR. What do you think @lavocatt ?

lavocatt:
Having everyone able to work effectively is more important. Let's figure out later a way to keep the iostat thing.

lavocatt (approved):

jrusz:
There are some unrelated issues with ostree tests on 8.6 and 9.0 that also affect main, we'll have to solve those first or get approval to force merge this...

PR #2199, opened 2022-01-18
templates/packer: Make cdn host check less sensitive

Gundersanne:


ochosi (approved):
<3

Gundersanne:
before merging i'll check the ami to see how it behaves.

Gundersanne:
Verified, admin merging as it's only related to the worker ami

PR #2202, opened 2022-01-18
templates/dashboards: Fixed grafana uids

Gundersanne:
This way we get a nice URL `.../d/image-builder-(composer|worker)`.


Gundersanne:
Admin merging because this only related to the grafana dashboards.

PR #2205, opened 2022-01-19
tests/api: Make cleanup best effort

jrusz:
We should not fail on missing variables or errors during cleanup. The
test can fail in any stage and the cleanup should be run full each time.

Current example where this causes problem can be seen here: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1990711774 
The test fails before the `S3_URL` is known and the cleanup function never reaches the part where background processes are killed and Gitlab CI gets stuck when there are processes running in background. This should ensure that the function is always ran in full.

This pull request includes:

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -194,12 +194,10 @@ function cleanupAWS() {
   AWS_SNAPSHOT_ID="${AWS_SNAPSHOT_ID:-}"
 
   if [ -n "$AWS_CMD" ]; then
-    set +e
     $AWS_CMD ec2 terminate-instances --instance-ids "$AWS_INSTANCE_ID"
     $AWS_CMD ec2 deregister-image --image-id "$AMI_IMAGE_ID"
     $AWS_CMD ec2 delete-snapshot --snapshot-id "$AWS_SNAPSHOT_ID"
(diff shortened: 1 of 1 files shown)


test/cases/api.sh
```diff
@@ -194,12 +194,10 @@ function cleanupAWS() {
   AWS_SNAPSHOT_ID="${AWS_SNAPSHOT_ID:-}"
 
   if [ -n "$AWS_CMD" ]; then
-    set +e
```
Gundersanne, line 197: This will do the opposite of what you want I think? `set -e` means fast-failing.

Gundersanne (approved):

test/cases/api.sh
jrusz: I've removed it from the individual functions and put it to the function calling these functions.

Gundersanne: Yup only noticed after, looks good :)

achilleas-k (approved):

jrusz:
It works, it fails instead of getting stuck :) https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/1992828358

jrusz:
I'm afraid this will required admin merge again @Gundersanne ... The failed jobs are related to the ostree issues, this PR aims to prevent the `api.sh edge-commit` test from getting stuck which will be beneficial to the CI.

PR #2214, opened 2022-01-22
cloupapi/v2: add koji support

teg:
Extend the cloud API to support koji.

The aim is to replace the current koji API with this one, once it reaches feature parity, and the clients have been redeployed to support this instead.

This will allow koji to work against composer on api.openshift.com, as the current koji API is incompatible with that. It also allows us to have one fewer APIs to maintain.

The API is extended in a backwards compatible way (I hope!), which makes it slightly ugly in places, but probably fine to do it like this rather than introduce a v3 (which we should do eventually anyway when we have a better idea of how it will all end up).

We must ensure that all the same image types are accessible through the new API as were accessible through the existing one.

This pull request includes:


cmd/osbuild-worker/jobimpl-osbuild-koji.go
```diff
@@ -86,6 +86,21 @@ func (impl *OSBuildKojiJobImpl) Run(job worker.Job) error {
 		return err
 	}
 
+	// In case the manifest is empty, try to get it from dynamic args
+	if len(args.Manifest) == 0 && job.NDynamicArgs() > 1 {
```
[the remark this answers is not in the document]
gicmo: (I am just widely guessing since I have a job failing with an empty manifest)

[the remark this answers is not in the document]
teg: I think I fixed this. The order of the arguments was wrong

cmd/osbuild-worker/jobimpl-osbuild-koji.go
```diff
@@ -86,6 +86,28 @@ func (impl *OSBuildKojiJobImpl) Run(job worker.Job) error {
...
+	if len(args.Manifest) == 0 {
+		if job.NDynamicArgs() > 1 {
+			var manifestJR worker.ManifestJobByIDResult
+			err = job.DynamicArgs(1, &manifestJR)
+			if err != nil {
+				return err
+			}
+
+			// skip the job if the manifest generation failed
+			if manifestJR.Error != "" {
```
kingsleyzissou, line 99: You will probably have to check for `JobErrors` now since https://github.com/osbuild/osbuild-composer/pull/2192 was merged

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -144,6 +144,72 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  '/compose/{id}/logs':
```
gicmo, line 147: Other routes start with `/composes`

teg: Fixed!

cmd/osbuild-worker/jobimpl-osbuild-koji.go
teg: Rebased on this. Please have a look though :)

kingsleyzissou: I think you got them all except this one :p

teg: LOL. oops.

teg: Done ;)

Gundersanne (commented):
Some small comments.

I'm not convinced of the type-safe jobstatus change. We were going to exploit this by passing just `JobResult` and getting joberrors without needing to know the type. Also the signature could just as well be
```
OSBuildJobStatus(id uuid.UUID) (result *OSBuildJobResult, *JobStatus, []uuid.UUID, error)
``` 
then.

cmd/osbuild-worker/jobimpl-koji-finalize.go
```diff
@@ -233,7 +233,9 @@ func hasFailedDependency(kojiInitResult worker.KojiInitJobResult, osbuildKojiRes
 	}
 
 	for _, r := range osbuildKojiResults {
-		if !r.OSBuildOutput.Success || r.JobError != nil {
+		// No `OSBuildOutput` imples failure: either osbuild crashed or
```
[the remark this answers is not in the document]
gicmo: If only I could spell ...

internal/cloudapi/v2/v2.go
```diff
@@ -411,16 +431,69 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
+
+		initID, err := h.server.workers.EnqueueKojiInit(&worker.KojiInitJob{
+			Server:  request.Koji.Server,
+			Name:    request.Koji.Name,
+			Version: request.Koji.Version,
+			Release: request.Koji.Release,
+		})
+		if err != nil {
+			// This is a programming error.
+			panic(err)
```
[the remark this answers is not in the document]
teg: Oops. This was all fixed up in the final commit.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -351,12 +351,15 @@ components:
       - type: object
         required:
           - status
-          - image_status
```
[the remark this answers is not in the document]
teg: Fair point. My thinking was that non-koi composes would still in practice always return this field, but no need to try to be clever about it, let's just return it always.

cmd/osbuild-worker/jobimpl-koji-finalize.go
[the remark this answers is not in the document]
teg: Fixed :)

teg:
@Gundersanne I agree with the suggested change to the internal `*JobStatus()` API, but let's leave that for a follow-up? Otherwise: addressed all comments.

internal/cloudapi/v2/v2.go
```diff
@@ -127,6 +128,34 @@ func (h *apiHandlers) GetError(ctx echo.Context, id string) error {
 	return ctx.JSON(http.StatusOK, apiError)
 }
 
+// splitExtension returns the extension of the given file. If there's
+// a multipart extension (e.g. file.tar.gz), it returns all parts (e.g.
+// .tar.gz). If there's no extension in the input, it returns an empty
+// string. If the filename starts with dot, the part before the second dot
+// is not considered as an extension.
+func splitExtension(filename string) string {
```
gicmo, line 136: Is there a subtle difference to `path.Ext()` that makes this necessary? if so, that could be part of the documentation.

teg: this was carried over from the koji api unchanged

Gundersanne: path.Ext only retusn stuff after the last '.'; so it would return .gz

Gundersanne (approved):

achilleas-k (commented):
Discussed in chat with @gicmo 

There might be a gap in the error handling.  It seems `JobError` is never set for a failed `KojiOSBuildJob`.
Compare https://github.com/osbuild/osbuild-composer/blob/c173d88061042ff04cdbdee13436d4a0d77bcf9c/cmd/osbuild-worker/jobimpl-osbuild.go#L175-L179
with
https://github.com/osbuild/osbuild-composer/blob/c173d88061042ff04cdbdee13436d4a0d77bcf9c/cmd/osbuild-worker/jobimpl-osbuild-koji.go#L100-L103

Neither of those were modified by this PR, but it might be worth checking.

It's _possibly_ fine because we check the statuses of all the dependencies of the finalize job here: https://github.com/osbuild/osbuild-composer/blob/c173d88061042ff04cdbdee13436d4a0d77bcf9c/cmd/osbuild-worker/jobimpl-koji-finalize.go#L111

but then this part of this PR will never have any effect
https://github.com/osbuild/osbuild-composer/pull/2214/files#diff-e9b0577660bfdf5be86ff316154728805fbbc038187e047e9e85f8290da3c473R163-R165

internal/cloudapi/v2/v2.go
```diff
@@ -729,33 +729,37 @@ func (h *apiHandlers) GetComposeStatus(ctx echo.Context, id string) error {
...
 			},
-			Status: composeStatusFromKojiJobStatus(finalizeStatus, &initResult, []worker.OSBuildKojiJobResult{buildResult}, &result),
-			ImageStatuses: &[]ImageStatus{
-				{
-					Status: imageStatusFromKojiJobStatus(buildJobStatus, &initResult, &buildResult),
-				},
-			},
-			KojiStatus: &KojiStatus{},
+			Status:        composeStatusFromKojiJobStatus(finalizeStatus, &initResult, buildJobResults, &result),
+			ImageStatus:   buildJobStatuses[0], // backwards compatibility
```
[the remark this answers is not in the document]
teg: That's the opposite of @Gundersanne comment on the first version ;) It would technically speaking be a breaking change, so I changed it to keep `ImageStatus` a mandatory field.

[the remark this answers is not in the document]
Gundersanne: I think non-breaking changes should become a habit :)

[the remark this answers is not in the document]
achilleas-k: haha sorry then :D

teg:
> There might be a gap in the error handling. It seems `JobError` is never set for a failed `KojiOSBuildJob`. Compare

You are right, and our backwards compatibility for `OSBuildJob` is also incomplete (does not handle the case when osbuild failed to run at all, so output is nil). I fixed it in the worker API, but we should also improve the actual job implementation for osbuild-koji to be on par (or maybe share code?) with the osbuild one.

Could you have a look at that separately @kingsleyzissou, just to make sure I didn't miss any cases?

achilleas-k (approved):
SPLENDID

PR #2225, opened 2022-01-27
distro/rhel87: copy all definitions from rhel86

dhodovsk:
Adjust strings from 86, 8.6, 8-6 to 87, 8.7, 8-7
Diff can be seen here: http://pastebin.test.redhat.com/1024612 

heavily inspired by https://github.com/osbuild/osbuild-composer/commit/513b7f94b9cc5c84c0b8c0b9ff8a445f25fc66c7?diff=split 

This pull request includes:

### DIFF
--- a/internal/distro/rhel87/distro.go
+++ b/internal/distro/rhel87/distro.go
@@ -1,0 +1,1277 @@
+package rhel87
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand"
+	"path"
+	"sort"
+	"strings"
+
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/disk"
+	"github.com/osbuild/osbuild-composer/internal/distro"
+	osbuild "github.com/osbuild/osbuild-composer/internal/osbuild2"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+)
+
+const (
+	// package set names
+
+	// build package set name
+	buildPkgsKey = "build"
+
+	// main/common os image package set name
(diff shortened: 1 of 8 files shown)


thozza:
Thank you a lot for your contribution and kickstarting this work...

IMO the following things should be part of this PR before we merge it:

- the new distro definition needs to be added to the distroregistry - https://github.com/osbuild/osbuild-composer/blob/main/internal/distroregistry/distroregistry.go
- image test cases should be generated for 8.7 (howto https://github.com/osbuild/osbuild-composer/tree/main/test#image-tests).
- we need to create RPMRepo snapshots for 8.7 and create appropriate repo definitions for testing and image test case generator tools.
- ideally enable CI testing on 8.7

I'm not saying that Dominika should do it as most of these things require knowledge of our processes. However without this the code is not tested and not used so there is no point in merging it.

thozza:
Since #2270 adds alias for RHEL-8.7, I propose to close this PR. The reason is that 8.6 definition is still being worked on and there is no point in copying it now into 8.7 and later double check and ensure that all changes from 8.6 were really moved to 8.7 definition. Instead we should copy/fork the 8.6 definition once it is "frozen" to not create additional work for us and prevent regressions.

dhodovsk:
I am all for closing this PR :upside_down_face: As stated in the description it was heavily inspired by what has been done for rhel-86 and as such, it is just a midless copy of 8.6 with most of the "6" changed to "7" (not in x86_64 :grin: ). I did it only in hope of unblocking new 8.7. builds but we now use rhel-86 distro and it seems to be sufficient for the early phases.

ondrejbudai:
Even though this gets closed, big THANK YOU for opening the PR and forcing us to do some actions. :D

thozza:
Definitely, BIG Thank You @dhodovsk for submitting the PR!

PR #2228, opened 2022-01-27
distro/rhel86: fix ec2 boot partition for arm64

gicmo:
It should be `1048576` (exactly 512 MiB), like it is for all other distributions. It somehow got mingled in when the distribution was forked off from 8.5/9.0 beta (1048676 to 1048576 strongly suggests a `sed` command was involved, so we blame that).


achilleas-k:
> (1048676 to 1048576 strongly suggests a `sed` command was involved, so we blame that)

Who would ever do something like that?? >_>

achilleas-k (approved):

PR #2233, opened 2022-01-28
distro/rhel90: no uuids in dos partition table

### LINKED ISSUE #2232: RHEL9: ppc64le fails to boot
Cannot boot our PPC64LE image. Manifest taken from the test directory.

Screenshot 2022-01-27 at 23 03 34

NB: the missing entry for `msdos2` which would be the boot partition.

gicmo:
The unification of the partition table also introduced uuids and types in uuid form for partition tables in dos layout, sill used on PPC64LE and s390x. The `org.osbuild.sfdisk` stage did work with that but produced a `/boot` partition with the wrong type, which grub2 refused to read from and thus prevented boot. Fix this by removing uuids from the dos partition tables.

Closes #2232


achilleas-k (approved):
LGTM :)

PR #2235, opened 2022-01-28
EC2: Disable password based authentication

gicmo:
Disable loging in via password authentication since this is an
official Amazon marketplace requirement:

>  Linux-based AMIs must not allow SSH password authentication.
  Disable password authentication via your sshd_config file by
  setting PasswordAuthentication to NO.

 See section "Security policies" from https://docs.aws.amazon.com/marketplace/latest/userguide/product-and-ami-policies.html

### DIFF
--- a/internal/distro/image_config.go
+++ b/internal/distro/image_config.go
@@ -36,6 +36,7 @@ type ImageConfig struct {
 	PamLimitsConf []*osbuild2.PamLimitsConfStageOptions
 	Sysctld       []*osbuild2.SysctldStageOptions
 	DNFConfig     []*osbuild2.DNFConfigStageOptions
+	SshdConfig    *osbuild2.SshdConfigStageOptions
 }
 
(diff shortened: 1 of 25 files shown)


thozza (approved):
Changes look good to me. Nevertheless I have a few questions / comments:

1. I'm not sure if it is OK for us to change the RHEL-8.5 definition, since it is already released.
2. Did you consult the changes to RHEL images with the Virt team? I think that these should be OK, but I wonder if anyone (e.g. SAP) has any specific requirements in this regard...
3. I wonder whether we should apply some additional logic in case the user specifies a user account with a password to be created via a blueprint customization?

Also it would be nice to extend `image-info` to check the SSHD configuration. Nevertheless it is somewhere on my TODO list... 😆

teg:
For completeness: this is in practice a noop, as @thozza pointed out. However, if we were to upload this image to the Amazon Marketplace, it would be rejected by their scanner as they check the ssh configuration.

PR #2238, opened 2022-01-30
Skip CI for draft PRs or WIP label

rgolangh:
Save resources and needless CI runs by skipping the trigger-gitlab job if a PR is 'draft' (can be set in the pull request page at any time) or by setting a label "WIP".

Aside from saving resources, WIP PRs that don't trigger CI are a good message to contributors to keep testing it locally and to only get the project resources after serious testing been made on their side.

### DIFF
--- a/.github/workflows/trigger-gitlab.yml
+++ b/.github/workflows/trigger-gitlab.yml
@@ -40,6 +40,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SCHUTZBOT_SSH_KEY: ${{ secrets.SCHUTZBOT_SSH_KEY }}
+      SKIP_CI: ${{ github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP') }}
     steps:
       - name: Clone repository
         uses: actions/checkout@v2.4.0
@@ -60,4 +61,5 @@ jobs:
             git checkout -b PR-${{ github.event.pull_request.number }}
           fi
 
-          git push -f ci
+          [[ "${SKIP_CI}" == true ]] && PUSH_OPTION="-o ci.skip" || PUSH_OPTION=""
+          git push -f ci "${PUSH_OPTION}"


jrusz:
What if I have a draft or WIP PR and I just want to see if it works before changing it to "ready for review", can I still trigger the CI at least manually?

jrusz:
Gitlab CI can also be skipped if you add anywhere in the commit `[skip-ci]` .. It will still push the PR to the gitlab repository but the pipeline will not be run. It's then easy to just trigger it manually if you need to.. However if we skip this whole Github action it won't even push the PR to the Gitlab repository so we wouldn't be able to trigger it manually I think...

rgolangh:
> Gitlab CI can also be skipped if you add anywhere in the commit [skip-ci] .
Fair point. Can it be tweaked to include 'WIP' so this would be more intuitive to other contributers coming from prow (openshift, kubernetes and so on)? if yes then this PR is really not needed.

rgolangh:
> > Gitlab CI can also be skipped if you add anywhere in the commit [skip-ci] .
> > Fair point. Can it be tweaked to include 'WIP' so this would be more intuitive to other contributers coming from prow (openshift, kubernetes and so on)? if yes then this PR is really not needed.

Having second thoughts though - if you want to trigger CI you must change your commit, (commit amend) and by that changing the history. The advantage with github label or draft state is that it doesn't touch the change

jrusz:
> Having second thoughts though - if you want to trigger CI you must change your commit, (commit amend) and by that changing the history. The advantage with github label or draft state is that it doesn't touch the change

I found this: https://docs.gitlab.com/ee/user/project/push_options.html#push-options-for-gitlab-cicd ... Can you add a condition to the GH action similar to what you already did but for the push? If it's draft or has WIP add the `-o ci.skip` to the `git push` command.

rgolangh:
that's much cleaner, thanks. I'll try to see if I can add that

thozza:
Personally I'm not a huge fan of skipping the CI for Draft PRs. I use them for work that I want to be tested in CI, but it is not necessarily complete yet. Nevertheless I understand that people may create a Draft PR and then not think about the consequences every time they push to their branch.

jrusz:
> Personally I'm not a huge fan of skipping the CI for Draft PRs. I use them for work that I want to be tested in CI, but it is not necessarily complete yet. Nevertheless I understand that people may create a Draft PR and then not think about the consequences every time they push to their branch.

You should still be able to see your pipeline when you go to Gitlab and run it manually if you want to test your work in progress...

ondrejbudai (approved):
Looks good, we can even invent labels like "force-test" which would enable the CI on draft PRs. That would basically make the CI opt-in for drafts.

thozza (approved):

PR #2243, opened 2022-01-31
Increase koji timeout

diaasami:
to avoid brew jobs failing with "dial tcp: i/o timeout"

This pull request includes:

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -2,6 +2,8 @@ package koji
 
 import (
 	"bytes"
+	"time"
+
 	// koji uses MD5 hashes
 	/* #nosec G501 */
 	"crypto/md5"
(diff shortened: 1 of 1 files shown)


Gundersanne (commented):
Do we set a timeout on the client somewhere? By default the client shouldn't timeout requests. So this PR adds a timeout rather than having it on 0.

teg:
> Do we set a timeout on the client somewhere? By default the client shouldn't timeout requests. So this PR adds a timeout rather than having it on 0.

Could it be that another timeout deeper down in the stack is not 0 by default? https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

diaasami:
Last 50 builds are green.
According to @achilleas-k, It seems like a temporary infra instability and no need for a long-term solution.
So closing the PR.

achilleas-k:
Thanks for the quick response Diaa.  And sorry the whole thing was probably not necessary in the end :)

PR #2248, opened 2022-02-02
gitlab-ci: make every stage interruptible

jabia99:
scheduled cloud cleaner is now fully functioning. This means we can
auto-cancel whole pipelines on new commits.

This PR adds the tag "interruptible" to every stage.

This pull request includes:


jrusz (changes requested):
I think that at this point it would be better to just add `interrputible: true` to the `.base` from which every other stage is extended and in case a new job is added which should not be interrupted it will have to be explicitly defined as `interruptible: false`

jrusz (approved):
LGTM

PR #2252, opened 2022-02-03
service-maintenance: Skip db cleanup

Gundersanne:
Let's enable the cloud cleanup first, and then move on the db.


ondrejbudai (approved):
Makes sense!

PR #2260, opened 2022-02-04
Worker error validation

kingsleyzissou:
This pull request includes adds additional checks to ensure there aren't any gaps in the
error handling for `OSBuildJob` and `KojiOSBuildJob` jobs.

The osbuild jobs were mostly fine the output is not nil from the start:
https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker/jobimpl-osbuild.go#L74-L76

But could get overridden here:
https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker/jobimpl-osbuild.go#L146

The osbuild worker now explicitly checks for nil `OsbuildOutput`.

Similarly, a function was added to validate koji build results and update the job with a job error if the worker exits
early due to an error.

The PR adds a few extra error cases too.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild-koji.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild-koji.go
@@ -56,12 +56,36 @@ func (impl *OSBuildKojiJobImpl) kojiUpload(file *os.File, server, directory, fil
 	return k.Upload(file, directory, filename)
 }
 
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
Good stuff.  I always suspect there are error cases we don't catch or nil dereferences we're not checking that just haven't happened to bite us yet, so it's nice to see this issue getting attention :)

Thanks!

teg (approved):

PR #2265, opened 2022-02-07
Support associating repositories to package sets

achilleas-k:
Some pipelines require packages from more than one compose, e.g. the 7.9 pipelines will require a 8.x compose since we need an 8 build root in addition to a 7 build root.  We therefore need to be able to pass in specific repositories for specific package sets. 

This PR adds an option in the API that allows a repository to be associated with one or more package sets.  The valid package sets are determined by the image type.

---

The base functionality for this PR is done, but I'd like to add some tests before finalising.  I've had this ready for a few days now and haven't had time to add the tests, so I'm opening this draft PR to make it visible (and to not forget about it).

---

This pull request includes:


Gundersanne (commented):
As a general note, we don't want to allow people to set the build package set do we? Or is this needed for koji?

Like in general we don't want people specifying arbitrary repositories for build package sets, of course this can still be guaranteed by image-builder-crc never passing the 'build' tag.

achilleas-k:
Yes, we don't want to allow users to specify repositories for any package set, so this functionality should never be exposed to Image Builder.  We need this for building RHEL 7 images because the QEMU tooling we need isn't available in RHEL 7, so we'll need to have a RHEL 8 build root to get those tools.

That all said, we could use this tagging to deprecate the existing user repo customisation and have image builder automatically tag any repos it gets from the user with `blueprint`, but this is just an aside.

thozza:
Very interesting change (and problem to solve)... 🤔 

@achilleas-k would you mind sharing the specific functionality that is missing on RHEL-7? Does it miss specific packages or it contains too old version of some tools? I'm asking because I'm wondering whether it would not be easier and more feasible to add the relevant packages to RHEL-7 (optional channel) or actually have our repository with RHEL-7 rebuild of specific packages, which would be passed for the image build.

I'm sure that you gave this a lot of thought, I just see a few issues with this approach. E.g. should the API then be able to list package set names? Because package set names and pipeline names, while most of the time named reasonably consistently, are mostly literal strings defined for each image type. Would not this be solved by e.g. not using build pipeline for RHEL-7 images and always building them on RHEL-8 system?

achilleas-k:
> @achilleas-k would you mind sharing the specific functionality that is missing on RHEL-7? Does it miss specific packages or it contains too old version of some tools? I'm asking because I'm wondering whether it would not be easier and more feasible to add the relevant packages to RHEL-7 (optional channel) or actually have our repository with RHEL-7 rebuild of specific packages, which would be passed for the image build.

The biggest (if not only) missing feature is support for Azure disk images.  There's a mention of `cp` also being an issue in some of our design docs, but I'm told that's not insurmountable.

> I'm sure that you gave this a lot of thought, I just see a few issues with this approach. E.g. should the API then be able to list package set names? Because package set names and pipeline names, while most of the time named reasonably consistently, are mostly literal strings defined for each image type. Would not this be solved by e.g. not using build pipeline for RHEL-7 images and always building them on RHEL-8 system?

I have thought about it quite a bit and you're absolutely right, we should have a way to get the package set / pipeline names from composer for a given image type.  We shouldn't rely on the consistency of the names which are, at the end of the day, arbitrary strings.

About the more general problem though, @gicmo has thought about it a lot more extensively so maybe he can chime in.  I imagine there might be some issues with building a RHEL 7 system in a RHEL 8 build root (or a RHEL 8 host without an isolated build root), but I can't think of them off the top of my head.

achilleas-k:
Marking as ready.  There are unit tests for the new repository splitting but we should probably add an integration test somewhere.  I didn't want to add yet another test case to our pipelines, so I'm asking for feedback: What's a good existing test to add some extra repos to test this feature?

gicmo (approved):
👍 ❤️

gicmo:
Packit is funny but only `fedora-36-*`, I guess a side effect of branching? Everything else is green. Will merge.

PR #2282, opened 2022-02-09
Relax TCP timeouts for koji connections

diaasami:
See COMPOSER-1354 and linked tickets

This pull request includes:

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -2,9 +2,13 @@ package koji
 
 import (
 	"bytes"
+	"net"
+	"time"
+
 	// koji uses MD5 hashes
 	/* #nosec G501 */
 	"crypto/md5"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


internal/upload/koji/koji.go
```diff
@@ -420,3 +424,24 @@ func GSSAPICredentialsFromEnv() (*GSSAPICredentials, error) {
 		KeyTab:    keyTab,
 	}, nil
 }
+
+func CreateKojiTransport(relaxTimeout uint) *http.Transport {
```
teg, line 428: Should we just fold this into `kojiNew()`? We now take a generic `http.Transport` there, but if the comments in this function is to be believed we can't really pass anything, maybe just hard-code the transport setup?

I think @ondrejbudai did this originally, so might have thoughts?

ondrejbudai: I think I wanted to make the `koji` library more flexible but I guess it isn't actually needed so I have no problems with just folding the code in.

internal/upload/koji/koji.go
```diff
@@ -420,3 +424,24 @@ func GSSAPICredentialsFromEnv() (*GSSAPICredentials, error) {
...
+		Renegotiation: tls.RenegotiateOnceAsClient,
+		MinVersion:    tls.VersionTLS12,
+	}
+
+	// Relax timeouts a bit
+	if relaxTimeout > 0 {
+		transport.TLSHandshakeTimeout *= time.Duration(relaxTimeout)
+		transport.DialContext = (&net.Dialer{
+			Timeout:   30 * time.Second * time.Duration(relaxTimeout),
+			KeepAlive: 30 * time.Second,
```
ondrejbudai, line 442: Is there a reason to send the keep-alive probes every 30 seconds instead of every 15 seconds (default value)?

diaasami: I thought that's the default https://go.dev/src/net/http/transport.go#L47
mmm making it 15 would increase the traffic but we can try.

teg: Ah, I missed this. Makes sense to me to leave this one untouched, unless it gives us problems?

ondrejbudai: Ahhh, such a mess. Yeah, let's leave it untouched and that means 30 seconds.

I accidentally read this documentation (https://pkg.go.dev/net#Dialer) which mentions that the default value for `net.Dialer` is 15 seconds but `DefaultTransport` actually sets it to 30 seconds because one default value is apparently not enough. :shrug:

ondrejbudai (approved):

PR #2284, opened 2022-02-10
Pre-define an ostree remote for RFE raw images (including created via simplified installer)

runcom:
I'm re-bringing this up from https://github.com/osbuild/osbuild-composer/pull/1697 as that didn't make 85 and even if we still don't have automatic updates this is pretty important for the admin workflow who's used to just serve the update and `rpm-ostree upgrade` - it should be pretty safe to land this (I don't remember what was the risk)

From the old PR:

> Preset a remote called rhel-edge with the URL that was used to fetch the commit, so it can be used to fetch updates.

This pull request includes:

### DIFF
--- a/internal/distro/rhel86/pipelines.go
+++ b/internal/distro/rhel86/pipelines.go
@@ -790,10 +790,11 @@ func ostreeDeployPipeline(
 	p.Name = "image-tree"
 	p.Build = "name:build"
 	osname := "redhat"
+	remote := "rhel-edge"
 
 	p.AddStage(osbuild.OSTreeInitFsStage())
 	p.AddStage(osbuild.NewOSTreePullStage(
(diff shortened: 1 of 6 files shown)


gicmo:
Thanks for bringing this back, it was on my TODO, but I didn't get around yet! ❤️ 

> (I don't remember what was the risk)

We were just very late for 8.5, i.e. would have had to ask for an exception.

runcom:
> We were just very late for 8.5, i.e. would have had to ask for an exception.

awesome, so we can still aim to have this in the upcoming releases (tbh the manual workaround is just different from i.e. RHEL 8.4 and anaconda, so this would be very cool to have)

runcom:
any reason the rpm jobs are queued but show success in schutbot? :thinking:

PR #2285, opened 2022-02-10
cloudapi: expose filesystem customizations

chloenayon:
This pull request includes:


chloenayon:
@jkozol

ondrejbudai (commented):
Looks nice, needs a rebase though.

teg (commented):
Thanks for this! Looking good :)

teg:
Hm, the packit issue is surprising, I don't think we changed anything to do with this (certainly not in this PR). Though I could obviously have missed something with our copr setup.

@TomasTomecek, do you know if this might be a packit/infrastructure issue?

ochosi:
> I think that's Fedora branching. Packit is reminding us to include the new version. :)

Indeed, might make sense to grant packit admin access to the copr?

ondrejbudai:
I don't understand though why packit is suggesting removing s390x so I'm bit reluctant to give packit admin access...

TomasTomecek:
> I don't understand though why packit is suggesting removing s390x so I'm bit reluctant to give packit admin access...

This is actually a bug in how Packit handles Copr projects and chroots (or the way Copr handles them). This gets exposed mostly around branching. Details: https://github.com/packit/packit/issues/1303

Let me try to fix the settings.

teg:
In addition to the packit problem, there seems to be some type issue: `/home/runner/work/osbuild-composer/osbuild-composer/internal/cloudapi/v2/v2.go:214:18: cannot use f.Mountpoint (variable of type *string) as string value in struct literal`

teg:
Ah, looks like packit is working as expected now, but tripping up over the bug mentioned above

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -804,6 +804,19 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Repository'
+        filesystem:
+          type: array
+          items:
+            $ref: '#/components/schemas/Filesystem'
+    Filesystem:
+      type: object
```
teg, line 812: I suggest making both `mountpoint` and `min_size` (see below) required:
```
required:
  - mountpoint
  - min_size
```

That would firstly fix the compile error in the PR currently (by making these optional they end up as pointers, so we would at least need to dereference them before assigning them to the blueprint customisations, but then we would also need to handle the case where they are not set), but I also think it is more correct:

The mountpoint should always be mandatory, as without it, I don't think a filesystem entry (of the kind we support) make sense. Since the size is actually the minimum size, there is no harm in making that mandatory too, the caller could just set it to 0 and it should have the same effect as not setting it at all.

teg (changes requested):
I had a look to figure out why this did not compile, and as a result came up with a couple of suggestions (hopefully including a fix).

internal/cloudapi/v2/v2.go
```diff
@@ -205,6 +205,22 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 		}
 	}
 
+	if request.Customizations != nil && request.Customizations.Filesystem != nil {
+		var fsCustomizations []blueprint.FilesystemCustomization
+		for _, f := range *request.Customizations.Filesystem {
+
+			fsCustomizations = append(fsCustomizations,
+				blueprint.FilesystemCustomization{
+					Mountpoint: f.Mountpoint,
```
teg, line 214: The reason this does not compile is that `f.Mountpoint` is `*string`, but `Mountpoint` expects `string`.

Either we need to change `(v2.Filesystem).Mountpoint` to `string` too, or we must dereference it like `Mountpoint: *f.Mountpoint`. However, if we do that, we better check that it is actually set first. As mentioned in the earlier comments, I think the correct solution is to change `(v2.Filesystem).Mountpoint` to `string` (by making it `required` in the open api spec).

internal/cloudapi/v2/v2.go
```diff
@@ -205,6 +205,22 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
+					MinSize:    f.Size,
```
teg, line 215: Same problem as above, `f.Size` is `*int`, but `MinSize` expects `uint`.

teg: Ah, actually it expects `uint64`. So to make this work you'll need to cast it: `uint64(f.MinSize)`.

This would be bad in case `MinSize` was passed in as a negative number (it would be turned into a huge positive one). Though, we need to guard against too large filesystem sizes anyway, so this might be acceptable. I would leave it as is, but maybe in a follow-up you could add a minimum size (of 0) to the openapi spec, then if we validate the requests against the spec we'll be fine.

chloenayon: makes sense, will do!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -818,6 +818,22 @@ components:
...
+      required:
+        - mountpoint
+        - min_size
+      properties:
+        mountpoint:
+          type: string
+          example: '/var'
+        min_size:
```
Gundersanne, line 834: You can specify a type to use using the `x-go-type` field. So `x-go-type: uint64` will remove the need to do any conversion and any theoretical size limit is honored.

teg: Oh, cool! Let's do that in a follow-up?

teg (approved):
Looks great! Thanks!

test/cases/api.sh
```diff
@@ -512,6 +512,12 @@ function createReqFileAWS() {
 {
   "distribution": "$DISTRO",
   "customizations": {
+    "filesystem": [
+      {
+        "mountpoint": "/var",
+        "min_size": 1024
```
teg, line 518: I think we don't guard against `/var` being too small, so I think the build will actually fail on this value. We are asking for a 1kB filesystem, which isn't going to work. I think this should be fixed in the filesystem code cc @gicmo @kingsleyzissou

teg: I took the liberty of pushing a fix for this. The errors in CI were really not all that helpful/clear, so I thought I'd quickly verify that this indeed is the issue.

kingsleyzissou:
I suspect the `edge-commit` api tests might be failing because we are trying to set filesystem mountpoints on edge images.

test/cases/api.sh
kingsleyzissou: I think `125MB` might still be a bit too small and that's why the one test is still failing, it might be worthwhile bumping that up. We can then add validation as a follow-up to the luks/lvm PR once that lands.

teg: Oops, I was a bit optimistic there.

teg: Fixed this.

teg:
> I suspect the `edge-commit` api tests might be failing because we are trying to set filesystem mountpoints on edge images.

Good catch! I dropped that from the tests.

kingsleyzissou (dismissed):

achilleas-k (approved):

teg:
CI fails on issue #2322.

We don't target koji for centos stream, and this is broken also in `main`, so let's not block on this.

PR #2286, opened 2022-02-10
ci: skip CI for draft and WIP PRs 

jrusz:
This pull request includes:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,4 +23,14 @@ test/README.md and add these additional items:
 - [ ] QE has approved reproducer/new tests and
 - [ ] Subsequent commits provide bug fixes without modifying the reproducer; CI reports PASS and
 - [ ] QE approves this PR; RHBZ status is set to `MODIFIED + Verified=Tested`
+
+Information regarding our GitLab pipeline (Schutzbot):
+
+CI will not be ran automatically if WIP label is applied or the PR is in DRAFT state, instead only
+a link will be provided to the pipeline which can then be triggered manually if desired. To run the
+CI automatically either switch the PR to ready or apply WIP+test label.
+
+Outside contributors need manual approval from one of the osbuild-composer maintainers.
+
+Schutzbot will only be triggered if all Tests jobs in GitHub workflow succeed.
 -->
--- a/.github/workflows/trigger-gitlab.yml
+++ b/.github/workflows/trigger-gitlab.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SCHUTZBOT_SSH_KEY: ${{ secrets.SCHUTZBOT_SSH_KEY }}
-      SKIP_CI: ${{ github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP') }}
+      SKIP_CI: ${{ (github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP')) && !contains(github.event.pull_request.labels.*.name, 'WIP+test') }}
     steps:
       - name: Report status
         uses: haya14busa/action-workflow_run-status@v1
@@ -59,5 +59,5 @@ jobs:
           touch ~/.ssh/known_hosts
           ssh-keyscan -t rsa gitlab.com >> ~/.ssh/known_hosts
           git remote add ci git@gitlab.com:osbuild/ci/osbuild-composer.git
-          [[ "${SKIP_CI}" == true ]] && PUSH_OPTION="-o ci.skip" || PUSH_OPTION=""
+          [[ "${SKIP_CI}" == true ]] && PUSH_OPTION='-o ci.variable="SKIP_CI=true"' || PUSH_OPTION=""
           git push -f ${PUSH_OPTION} ci
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,6 +40,21 @@ init:
   script:
     - schutzbot/update_github_status.sh start
 
+.build_rules:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "false"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "true"'
+      when: manual
+
+.upstream_rules:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+
+.nightly_rules:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-8.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "8"'
+
 
 .RPM_RUNNERS_RHEL: &RPM_RUNNERS_RHEL
   RUNNER:
@@ -56,7 +71,7 @@ RPM:
   stage: rpmbuild
   extends: .terraform
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - !reference [.build_rules, rules]
   script:
     - sh "schutzbot/mockbuild.sh"
   after_script:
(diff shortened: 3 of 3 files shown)


jrusz:
> What's the purpose of this PR? Doesn't the current method work?

It's improving it. The way it is now it won't show you the link to the pipeline on the PR and you have to go to the Gitalb CI page and manually trigger the pipeline run. With this the `init` stage will always run so you'll get the link on the PR status and the rest of the jobs will wait for manual triggering in case it's desired... 

I just hope it won't cause problems with limits/quota and I'm not sure yet if it's going to wait forever..

Also it's reducing rules duplication.

jrusz:
I'm simulating the workflow by adding a CI varible in Gitlab `SKIP_CI=true` the result is that it's waiting for the user to trigger it. Unless this causes more problems with CI limits I think it's a bit better, what do you think @thozza ?

Also @ondrejbudai is it necessary for the CI to get triggered after Tests/Checks are done? I liked it more when it was triggered pretty much right away, now I have to wait for 10 minutes before I start seeing something :/

thozza:
> I'm simulating the workflow by adding a CI varible in Gitlab `SKIP_CI=true` the result is that it's waiting for the user to trigger it. Unless this causes more problems with CI limits I think it's a bit better, what do you think @thozza ?

I checked the Gitlab pipeline and I think that this is great. A link in the Github checks list to Gitlab CI and there, one has to manually run it. Looks good from my PoW.

I know that some people would like to keep running the CI pipeline automatically on their Draft PR. Do you think that it would be possible to have some kind of a keyword in the PR description or a label set, which would accomplish this?

jrusz:
> I know that some people would like to keep running the CI pipeline automatically on their Draft PR. Do you think that it would be possible to have some kind of a keyword in the PR description or a label set, which would accomplish this?

I guess a label would be the easiest and that would mean just another condition in the GH workflow and rule update in `.gitlab-ci.yml` ... Easy to do imo but idk if people will remember to use it.

thozza:
> > I know that some people would like to keep running the CI pipeline automatically on their Draft PR. Do you think that it would be possible to have some kind of a keyword in the PR description or a label set, which would accomplish this?
> 
> I guess a label would be the easiest and that would mean just another condition in the GH workflow and rule update in `.gitlab-ci.yml` ... Easy to do imo but idk if people will remember to use it.

We can announce it on the `composer-eng` list once the workflow is implemented. And people who want to use it will have to learn it the hard way I guess 😄  Also we can modify the PR template to mention it.

jrusz:
Ok I changed the condition for `SKIP_CI` I just need someone to tell me if that is going to work, idk how else to test it than to merge it :sweat_smile: ... I added the label and we can announce it and change the template like @thozza suggested.

Overall I understand now the use of running the CI after the Checks are done and it makes sense to me now to wait for them first. This PR then just adds more convenience to running the Gitlab CI pipelines because it gives the link to the PR status and waits for manual execution in case of `WIP` label and draft PRs. In case it's draft and you want to run CI a new label must be applied `WIP+test` (didn't think of any better name, feel free to suggest better one). Also it's reducing the rules duplication in `.gitlab-ci.yml` file.

atodorov (commented):
+1 for reducing the duplcation inside the yaml file. We can take this separately.

I'm not sure how much value the skipping option does provide. I don't want to remember to open my PRs as a Draft or add labels to them b/c this is just some extra thing to remember in my head. On the other hand if my jobs weren't executed b/c the PR was opened as Draft then I can update the status very quickly just so I can satisfy the bot rules which essentially defeats the purpose of having any rules in the first place.

OTOH if PRs are opened as non-draft, not WIP then there's no change from what I see so what's the point then?

Given that this adds friction to the process and to the rest of the team I would vote against it, even if the friction seems to be rather small.

jrusz:
@ondrejbudai could take another look at this please?

.github/workflows/trigger-gitlab.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SCHUTZBOT_SSH_KEY: ${{ secrets.SCHUTZBOT_SSH_KEY }}
-      SKIP_CI: ${{ github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP') }}
+      SKIP_CI: ${{ github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP') && !contains(github.event.pull_request.labels.*.name, 'WIP+test') }}
```
ondrejbudai, line 16: Can you add parentheses to the condition? It's hard to parse whether `cond1 || cond2 && cond3` is `(cond1 || cond2) && cond3`, or `cond1 || (cond2 && cond3)`.

.gitlab-ci.yml
```diff
@@ -40,6 +40,21 @@ init:
   script:
     - schutzbot/update_github_status.sh start
 
+.rpmbuild_rules:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "false"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "true"'
+      when: manual
+
+.default_rules:
```
ondrejbudai, line 49: `upstream_rules` might be more descriptive, WDYT?

.gitlab-ci.yml
```diff
@@ -40,6 +40,21 @@ init:
   script:
     - schutzbot/update_github_status.sh start
 
+.rpmbuild_rules:
```
ondrejbudai, line 43: `build_rules` maybe? Because it's also used for the container build.

.gitlab-ci.yml
```diff
@@ -502,6 +505,10 @@ SonarQube:
 
 finish:
   stage: finish
+  rules:
```
ondrejbudai, line 508: is this really needed? Shouldn't the finish job just run always?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,4 +23,15 @@ test/README.md and add these additional items:
 - [ ] QE has approved reproducer/new tests and
 - [ ] Subsequent commits provide bug fixes without modifying the reproducer; CI reports PASS and
 - [ ] QE approves this PR; RHBZ status is set to `MODIFIED + Verified=Tested`
+
+Information regarding our CI:
```
ondrejbudai, line 27: CI sounds confusing because github actions are technically a part of our CI. Let's maybe use the term "GitLab pipeline" in this description? One mention of `Schutzbot` would be also nice to clear any confusion.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,4 +23,15 @@ test/README.md and add these additional items:
...
+
+CI will not be ran automatically if WIP label is applied or the PR is in DRAFT state, instead only
+a link will be provided to the pipeline which can then be triggered manually if desired. To run the
+CI automatically either switch the PR to ready or apply WIP+test label.
+
+Outside contributors need manual approval from one of the osbuild team members or request to
```
ondrejbudai, line 33: "friends of osbuilders" is just an implementation detail, it IMHO doesn't belong to the PR template.

> Outside contributors need manual approval to run the GitLab pipeline from one of the osbuild-composer maintainers.

is enough.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,4 +23,15 @@ test/README.md and add these additional items:
...
+be added to "friends of osbuilders" team for the CI to trigger without approval.
+
+Also CI is triggered after Tests actions are done for the PR which include Lint, Unit tests and ShellCheck.
```
ondrejbudai, line 36: I suggest not enumerating all tests. It can get outdated very quickly (it actually already is, we have 5 checks now). Mentioning that all Tests jobs in the github workflow must succeed in order to the gitlab pipeline to get triggered is enough.

ondrejbudai (changes requested):
Looks nice! Minor naming, docs, and parentheses nits inline. :)

.github/workflows/trigger-gitlab.yml
jrusz: ` SKIP_CI: ${{ (github.event.pull_request.draft == true || contains(github.event.pull_request.labels.*.name, 'WIP')) && !contains(github.event.pull_request.labels.*.name, 'WIP+test') }}` Is that correct? I'm not sure in case of github actions.

ondrejbudai: Yeah, that sounds correct. I even wonder if we should support the `WIP` label, marking a PR as a draft should be enough. No strong opinion though.

.gitlab-ci.yml
jrusz: I need to check it, I remember that it happened to me once that everything was skipped and only init and finish ran which reported pass to the PR.

jrusz: Yeah it's not necessary, I once removed the `SKIP_CI` variable from Gitlab and it didn't run any tests, just init and finish which reported pass on the PR. This should not happen though if we keep the variable in CI.. It's currently set to `false`.

ondrejbudai (dismissed):
Nice!

thozza (approved):
Thnaks a lot!

jrusz:
@ondrejbudai could you merge this please? I forgot about this cause I though it was merged already :D ... The failures were expected at that time...

PR #2287, opened 2022-02-10
Better implementation for TCP timeout relaxation

diaasami:
Relates-To COMPOSER-1354

This pull request includes:


internal/upload/koji/koji.go
```diff
@@ -127,7 +126,17 @@ type loginReply struct {
...
 
-func newKoji(server string, transport http.RoundTripper, reply loginReply) (*Koji, error) {
+func newKoji(server string, transport http.RoundTripper, reply loginReply, relaxFactor uint) (*Koji, error) {
+	// Relax timeouts
+	t := transport.(*http.Transport)
+	if relaxFactor > 0 {
+		t.TLSHandshakeTimeout *= time.Duration(relaxFactor)
+		t.DialContext = (&net.Dialer{
+			Timeout:   30 * time.Second * time.Duration(relaxFactor),
+			KeepAlive: 30 * time.Second,
```
teg, line 136: This might be me being confused about golang, or how this particular library works, but if you just don't set `KeepAlive` at all, won't it stay the default value (which in this case is `30`, so doesn't really matter, just trying to understand)?

ondrejbudai: If you set `KeepAlive` to 0, it's actually 15. But if you use `DefaultTransport`, `KeepAlive` is actually set to 30. I don't know why `DefaultTransport` sets non-default `KeepAlive`.

teg: Oh... 🤯 I'll be quiet.

PR #2291, opened 2022-02-10
internal/cloud: Allow aws creds from defaults

Gundersanne:
Defaults according to https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config:
Defaults to a chain of credential providers to search for credentials in
environment variables, shared credential file, and EC2 Instance Roles.

If nothing is specified fall back to whatever instance role.


teg (dismissed):
Always wanted this, very nice!

ondrejbudai (approved):
Agreed with Tom. Why didn't we do it sooner? 😅

teg (approved):

PR #2293, opened 2022-02-11
cmd/osbuild-service-maintenance: GCP deletes by image name

croissanne:


thozza (approved):
Sorry that I didn't realize that during the original review... 🙄

PR #2298, opened 2022-02-14
cloudapi/v2: add error object to ImageStatus

kingsleyzissou:
Add an error object to the ComposeStatus.ImageStatus. The error object contains a human-readable error reason
and optional details in the case of an error.

This pull request includes:


teg (commented):
I would love for the error object to be primarily structured data, so that it would be the frontend's job to generate a human readable string from it. The reason being that the frontend would ideally have the right context for knowing precisely what information to expose to the user and at what level of detail. It could even expose it differently in different contexts. If we pass it like a human readable string we are sort of locked in.

kingsleyzissou:
@teg that hadn't crossed my mind. I can make those changes. Do you thing returning the error code would be enough or is there other data you suggest returning?

teg:
> @teg that hadn't crossed my mind. I can make those changes. Do you thing returning the error code would be enough or is there other data you suggest returning?

I would say a good starting point would be enough structured data to recreate the human readable strings, then we can extend it over time?

Gundersanne:
I actually kinda like the idea of the human readable string that we can just propagate all the way through :) But why not both? Have the reason and details in there, but just also include the code? The code is meant to be stable after all.

kingsleyzissou:
@teg after thinking about Sanne's comment for a bit. If we were to handle the error in the frontend and create the human readable string would we not have to copy across some of the logic from the worker to translate the error codes?

For now, I've added the code to the error object but haven't removed the reason just yet. I guess we could still expand the object to add additional info similar to what we do here for example:
https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/errors.go#L167-L176
But those other fields are probably more than we need any way.

internal/cloudapi/v2/v2.go
```diff
@@ -822,6 +824,17 @@ func (h *apiHandlers) GetComposeStatus(ctx echo.Context, id string) error {
 	}
 }
 
+func composeStatusErrorFromJobError(jobError *clienterrors.Error) *ComposeStatusError {
+	if jobError == nil {
+		return nil
+	}
+	return &ComposeStatusError{
```
kingsleyzissou, line 831: It would probably be better to re-use the jobError object here

Gundersanne: Eh, most of these apis is just transforming one go object into the one defined by the api generator.

Gundersanne:
@teg further thoughts on this? I actually though the single string (reason) returned in the ComposeStatusError would be easy to present in the UI. It's a lot easier to present a single string than to present structured data depending on the error code. We can still do the latter using the code + details for specific codes, but we can always show the single string.

That said we should probably improve the strings, "osbuild build failed" isn't a very good error string :| But out of context for this PR.

teg:
> @teg further thoughts on this?

Sorry for dropping the ball on this.

I think it is fine as it is, though I think in the future we should discuss amending it with structured data so that consumers could choose to ignore the string.

That requires more discussion though, and no need to block on that.

Gundersanne (dismissed):

teg (approved):

PR #2306, opened 2022-02-15
devcontainer: adapt to composer

teg:
Make the devcontainer more complete for osbuild-composer by installing more dependencies and plugins. This enables golang and github integration by default.

Also improve on `docker-compose` so that if you enter the dev-container from `docker compose up` it is possible to use `composer-cli` as well as `generate-test-cases`.

A follow-up would spin up the whole docker compose from VSCode, but I could not get that to work cleanly, so for now these are separate.


.devcontainer/devcontainer.json
```diff
@@ -1,17 +1,12 @@
...
+  "name": "composer",
   "build": {
     "dockerfile": "Dockerfile",
     "context": ".."
   },
-  "mounts": [
-    "source=osbuild-share,target=/root/.local/share,type=volume"
-  ],
-  "runArgs": [
-    "--privileged"
```
ochosi, line 11: I'm afraid this leads to the workspace not being readable in the container (SE Linux and stuff, see also https://github.com/microsoft/vscode-remote-release/issues/1333).

Maybe someone else can try this out on native Linux as I presume you cannot reproduce on MacOS.

ochosi (changes requested):
:heart:

.devcontainer/devcontainer.json
gicmo: I also use `osbuild-pipeline` and `osbuild` directly in the container and the latter needs `--privileged`.

teg: Oops! Sorry for missing both these use cases (especially the non-macos one ;) ). Fixed now.

ochosi (approved):
Works like a charm!

teg:
I dropped the pre-commit hooks as that was giving me more problems than it was solving.

gicmo (dismissed):
LGTM

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
+version: '2.4'
+services:
+  devel:
+    image: local/osbuild-composer-devel
+    build:
+      dockerfile: ./.devcontainer/Dockerfile
```
achilleas-k, line 6: Not sure when this changed but the `context` key seems to be required for some older versions of docker compose.
On Arch with `Docker Compose version 2.2.3` this works.
On F35 with `docker-compose version 1.29.2` the file is considered invalid and should have `context: .` under `build`.

achilleas-k (changes requested):

docker-compose.yml
teg: Will fix! Thanks :)

teg: Fixed.

achilleas-k (approved):

PR #2314, opened 2022-02-16
Deduplicate some code in distro definitions

thozza:
This patch set deduplicates all code in `stage_inputs.go` of each distro definition and a few additional functions, which would have to be otherwise copied again to `rhel-84` due to `gce` image type.

This pull request includes:


thozza:
I think that this PR is ready for review, since the Manifests test passed. The rest is just a matter of persistent re-trying 😄

internal/rpmmd/repository.go
```diff
@@ -203,6 +203,23 @@ func (re *RepositoryError) Error() string {
 	return re.msg
 }
 
+func GetVerStrFromPackageSpecList(pkgs []PackageSpec, packageName string) (string, error) {
```
ondrejbudai, line 206: what about:

```golang
func (pkgs []PackageSpec) FindByName(name string) *PackageSpace {
    for _, pkg := range pkgs {
        if pkg.Name == packageName {
	    return &pkg
	}
    }
    return nil
}

func (p PackageSpec) NRA() string {
    return fmt.Sprintf("%s-%s.%s", p.Version, p.Release, p.Arch)
}

```

The following check:
```golang
p := pkgs.FindByName("osbuild")
if p == nil {
    panic("osbuild is not there, wow")
}
```
might be even omitted because it will panic nevertheless. :shrug: Or maybe we should return an error instead.

ondrejbudai (commented):
Nice! Others should review as well though as I'm not that familiar with image definitions.

internal/rpmmd/repository.go
ondrejbudai: `FindByNameOrPanic` might be another solution.

thozza: Thanks for the review.

Unfortunately the first part of your suggestion is not a valid Go code, since `[]PackageSpec` can not be used as a receiver type. I would have to define a type alias for it. I considered this, but decided to not do it as it would require changing the type in too many places. So I decided to define it as a function. However I like the idea of defining `VRA()` as `PackageSpec` method... 

WRT checking the error and panic. It felt too much to add gazillion of `if` statements to check returned error into every place where the function which previously panicked used to be called. Also it felt wrong to me to not provide function variant which returns `error` instead of always panicking. So I decided to implement a wrapper which checks the error and panics if there is some.

I'm happy to rework this a bit based on your suggestion.

ondrejbudai: > `[]PackageSpec` can not be used as a receiver type.

ofc :facepalm: sorry! Then we would have to go with a method similar to yours. Well, whatever works for you, I need to study more Go apparently. :shrug:

internal/rpmmd/repository.go
```diff
@@ -203,6 +203,23 @@ func (re *RepositoryError) Error() string {
...
+	for _, pkg := range pkgs {
+		if pkg.Name == packageName {
+			return fmt.Sprintf("%s-%s.%s", pkg.Version, pkg.Release, pkg.Arch), nil
+		}
+	}
+	return "", fmt.Errorf("package %q not found in the PackageSpec list", packageName)
+}
+
+func GetVerStrFromPackageSpecListPanic(pkgs []PackageSpec, packageName string) string {
```
achilleas-k, line 215: Indeed a very awkward and long function name.  I don't know if I have a better idea but we can always change it later since it's internal.

achilleas-k (approved):
Looks good.  Nice green/red ratio :)

PR #2328, opened 2022-02-21
schutzbot: keep runners alive when users are logged on

achilleas-k:
For troubleshooting in CI, if a user is logged on to a runner, keep it alive until they are logged off.

Not sure if there is a better way to do this.  I'm open to suggestions.

### DIFF
--- a/schutzbot/update_github_status.sh
+++ b/schutzbot/update_github_status.sh
@@ -6,6 +6,12 @@ if [[ "$CI_PIPELINE_SOURCE" == "schedule" ]]; then
     exit 0
 fi
 
+# if a user is logged in to the runner, wait until they're done
+while (( $(who -s | wc -l)  > 0 )); do
+    echo "Waiting for user(s) to log off"
+    sleep 30
+done
+
 if [[ $1 == "start" ]]; then
   GITHUB_NEW_STATE="pending"
   GITHUB_NEW_DESC="I'm currently testing this commit, be patient."


ondrejbudai (dismissed):
good idea!

jrusz (dismissed):
Looks good, definitely useful, but AFAIK we don't have access to the ssh key used on the runners...

achilleas-k:
> Looks good, definitely useful, but AFAIK we don't have access to the ssh key used on the runners...

My thinking was that sleeping in the `update_github_status.sh` would keep the EC2 and Openstack runners/instances alive as well, but I'm likely wrong.  The instances will stop once the script exits, before the `finish` stage is called, right?

achilleas-k:
> Just realized one thing, it won't work if you cancel the job or if it timeouts. Idk if that is a concern though.

I think it's fine to not cover those cases.  I'm thinking of use cases where we want to keep grab some log information from the instance before it's cleaned up.

jrusz:
> > Looks good, definitely useful, but AFAIK we don't have access to the ssh key used on the runners...
> 
> My thinking was that sleeping in the `update_github_status.sh` would keep the EC2 and Openstack runners/instances alive as well, but I'm likely wrong. The instances will stop once the script exits, before the `finish` stage is called, right?

The `update_github_status.sh` is ran in the `after_script` section which is ran for every job even if the main script fails but it's not ran if you cancel the job or if it timeouts.

ondrejbudai:
> Looks good, definitely useful, but AFAIK we don't have access to the ssh key used on the runners...

We do: https://github.com/osbuild/osbuild-composer/blob/main/schutzbot/team_ssh_keys.txt

jrusz:
> > Looks good, definitely useful, but AFAIK we don't have access to the ssh key used on the runners...
> 
> We do: https://github.com/osbuild/osbuild-composer/blob/main/schutzbot/team_ssh_keys.txt
> 
> See https://github.com/osbuild/osbuild-composer/blob/main/schutzbot/deploy.sh#L87

Ah I didn't know about this, I'll open a PR later, thanks!

achilleas-k:
[IT WORKS](https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2118092577) \o/
```
Running after_script
Running after script...
$ schutzbot/update_github_status.sh update
Waiting for user(s) to log off
```

achilleas-k:
Rebased on main.  We still want this, right?

ondrejbudai:
Indeed!

ondrejbudai (approved):

PR #2340, opened 2022-02-22
internal/cloudapi: Allow bp.Customizations being nil

Gundersanne:
Customizations being nil isn't the same as an empty object. The
edge-installer types require customizations equalling nil.

This partially reverts bb486bcccb5abf6b96e9e8b76c7226a693d08172.

### DIFF
--- a/internal/cloudapi/v2/v2.go
+++ b/internal/cloudapi/v2/v2.go
@@ -173,7 +173,6 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 	if err != nil {
 		return HTTPErrorWithInternal(ErrorFailedToInitializeBlueprint, err)
 	}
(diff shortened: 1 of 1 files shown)


teg:
Is this ready for review/to go in?

ondrejbudai:
btw why didn't we catch this? image-installer isn't built in `api.sh`, right?

teg:
> btw why didn't we catch this? image-installer isn't built in `api.sh`, right?

I asked the same, turns out it is only tested in image-builder. Room for improvement...

chloenayon:
Good catch, thank you!

Gundersanne:
Waiting on https://github.com/osbuild/osbuild-composer/pull/2343 to unbreak CI + rebase. Unless we're happy with admin-merging this.

PR #2342, opened 2022-02-22
RHEL 9.0: Fix customisation of Kernel command line options

### LINKED ISSUE #2129: [rhel 9, F33+] Customisation of kernel arguments is broken
Starting with Fedora 33[1] the kernel command line options in the BLS snippets is static. Previously the BLS snippets contained a `$kernelopts` variable, set in the `grubenv` file and thus setting (or changing) the kernel command line was done by changing `grubenv`. Starting with F33 this changed and the BLS snippets are not static, i.e. the kernel command line gets generated during the generation of the BLS snippets. The locations that are read are, `/usr/lib/kernel/cmdline`, `/etc/kernel/cmdline` and finally `proc/cmdline` (and that order).
We do already include the `org.osbuild.kernel-cmdline` stage in [RHEL9](https://github.com/osbuild/osbuild-composer/blob/cf73edd2d2248bbb5e561a8beeab694857be2132/internal/distro/rhel90/distro.go#L277) but we do not make use of the customisations `kernel.Append`. In face we do so for the `org.osbuild.grub2` stage [here](https://github.com/osbuild/osbuild-composer/blob/cf73edd2d2248bbb5e561a8beeab694857be2132/internal/distro/rhel90/distro.go#L537) which is wrong, since they will just get ignored.

Same seems true for [F33+](https://github.com/osbuild/osbuild-composer/blob/cf73edd2d2248bbb5e561a8beeab694857be2132/internal/distro/fedora33/distro.go#L347)

[1] https://src.fedoraproject.org/rpms/grub2/c/4a742183a39f344a7685bccdc76d5e64dea3766a?branch=f33

achilleas-k:
Resolves #2129.

Needs manifest and image info upates.

-----

This pull request includes:

### DIFF
--- a/internal/osbuild2/grub2_stage.go
+++ b/internal/osbuild2/grub2_stage.go
@@ -3,6 +3,7 @@ package osbuild2
 import (
 	"github.com/google/uuid"
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/common"
 	"github.com/osbuild/osbuild-composer/internal/disk"
 )
 
@@ -22,6 +23,7 @@ type GRUB2StageOptions struct {
 	UEFI               *GRUB2UEFI `json:"uefi,omitempty"`
 	SavedEntry         string     `json:"saved_entry,omitempty"`
 	Greenboot          bool       `json:"greenboot,omitempty"`
+	WriteCmdLine       *bool      `json:"write_cmdline,omitempty"`
 }
 
 type GRUB2UEFI struct {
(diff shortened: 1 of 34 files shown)


internal/osbuild2/grub2_stage.go
```diff
@@ -83,3 +85,42 @@ func NewGrub2StageOptions(pt *disk.PartitionTable,
...
+	if bootFs != nil {
+		bootFsUUID := uuid.MustParse(bootFs.GetFSSpec().UUID)
+		stageOptions.BootFilesystemUUID = &bootFsUUID
+	}
+
+	if uefi {
+		stageOptions.UEFI = &GRUB2UEFI{
+			Vendor:  vendor,
+			Install: install,
+			Unified: legacy == "",
```
teg, line 117: Why isn't this always `true`?

gicmo: We tried to minimise the changes in the manifest, since `UEFI` + `Legcay` is hybrid boot, which implies `unified` (same code path in osbuild), so in the case setting `Unified` is redundant.
Having said that, `Unified: true` is how we initially wrote the code and if we regenerate the manfifests anyway might as well always set it (even for the hybrid case).

achilleas-k: It should be.
Though it doesn't matter (functionally) in the cases where we use it, semantically it makes sense to have it pinned to true.  I just kept it that way to minimise the changes in the manifests and image info.

teg: ah, ok. just tripped me up due to the name of the function. a comment would solve it too

gicmo: I think it is either a comment or `Unified: True` and I am now swaying for the latter.

achilleas-k: Done

gicmo (commented):
I went through the image infos as well and they look good.

teg (approved):
Thanks!

thozza (approved):
LGTM

gicmo:
All green except the koji integration test on CS9, which is expected to fail!
Screenshot 2022-02-23 at 22 40 08

PR #2346, opened 2022-02-23
test: add tests purpose

jabia99:
Most test scripts don't have any documentation regarding it's purpose.
Although it can be guessed by the code, there's value in adding this
small comment.

[skip-ci]

This pull request includes:

### DIFF
--- a/test/cases/base_tests.sh
+++ b/test/cases/base_tests.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# Script that executes different high level sanity tests writen in GO. 
+#
+
 WORKING_DIRECTORY=/usr/libexec/osbuild-composer
 TESTS_PATH=/usr/libexec/osbuild-composer-test
 mkdir --parents /tmp/logs
--- a/test/cases/generation.sh
+++ b/test/cases/generation.sh
@@ -1,4 +1,11 @@
 #!/usr/bin/bash
+
+#
+# The objective of this script is to provide special requirement needed
+# for generate-test-cases. For example, disable the dnf-json daemon
+# and then execute generate-test-cases.
+#
+
 set -euxo pipefail
 
 # Provision the software under test.
--- a/test/cases/installers.sh
+++ b/test/cases/installers.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# This test builds a image-installer iso which is then used for OS installation. 
+# To do so, it creates an image-installer iso, then modifies the kickstart so it 
+# can be installed automatically, installs the VM with virt-install waits for it 
+# to boot and then run some test using ansible to check if everything is as expected
+#
+
 # Provision the software under test.
 /usr/libexec/osbuild-composer-test/provision.sh
 
--- a/test/cases/libvirt.sh
+++ b/test/cases/libvirt.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# Helper script that executes `tools/libvirt_test.sh` with the appropiate image type and boot type
+#
+
 # Get OS data.
 source /usr/libexec/osbuild-composer-test/set-env-variables.sh
 
--- a/test/cases/regression-composer-works-behind-satellite.sh
+++ b/test/cases/regression-composer-works-behind-satellite.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+#
+# osbuild-composer can work with multiple subscriptions on the host system.
+# test this by simulating a custom repository with custom certificates and 
+# verifies that an image can be built using those.
+#
+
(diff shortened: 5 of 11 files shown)


jabia99:
I'll divide this task in two different PR's. Please add any suggestions you have.

test/cases/base_tests.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# Script that servers as a base for some tests
```
jrusz, line 5: What do you mean "base for some tests"? Look at line 16 this runs various high level and sanity tests including cli tests which are written in GO. Check also some logs from a run in Gitlab CI to see how it looks.

test/cases/generation.sh
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/bash
+
+#
+# The objective of this script is to provide special requirement needed
+# for generate-test-cases. For example, disable the dnf-json daemon
+# and then execute generate-test-cases.
```
jrusz, line 6: Not exactly sure what this test does but looking at the history I found:
```
Test that the test cases generation will properly start the
dnf-json.socket when needed.
```
So I would guess it's doing just that, not sure if more uses are expected.

test/cases/installers.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# tests the correct installation fo selected packaes. To do so:
+# Builds a tar-installer, installs the OS, boots it, and checks if the
+# packages selected in the blueprint are installed.
```
jrusz, line 7: Uhm, no... This test builds a `image-installer` iso which is then used for OS installation. The packages there are necessary for some additional checks I would assume. So what this test does is creates an `image-installer` iso, then modifies the kickstart so it can be installed automatically, installs the VM with `virt-install` waits for it to boot and then run some test using ansible to check if everything is as expected. Check the code and ask if you don't understand any parts.

test/cases/libvirt.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# This script executes tools/libvirt_test.sh with the appropiate image type and boot type
```
jrusz, line 5: Well this is more like a helper script that runs `libvirt` test which is just creating a qcow2 image (openstack is basically the same, AFAIK there is no real difference), it then modifies them with cloud-init and verify that they boot.. Also it's testing UEFI where supported.

test/cases/vmware.sh
```diff
@@ -1,4 +1,11 @@
 #!/bin/bash
+
+#
+# Test osbuild-composer 'upload to vmware' functionality. To do so, create and
+# push a blueprint with composer cli. Then, create an instance in vSphere
+# from the uploaded image.
```
jrusz, line 6: Same comment as for `aws.sh`

tools/libvirt_test.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euo pipefail
 
+#
+# tests that guest images are buildable through the weldr API and boot-tests them using libvirt
```
jrusz, line 5: I'd say either "using the weldr API" or just "using composer-cli".. Idk if it's immediately obvious what "boot-tests them" is so I'd just say "and verifies they boot with cloud-init using libvirt"

test/cases/regression-composer-works-behind-satellite.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+#
+# osbuild-composer can work with multiple subscriptions on the host system.
+# test this functionality by generating the certificates and creating
+# an image
```
jrusz, line 6: This is not precise.. It's simulating a custom repository with custom certificates and verifies that an image can be built using those. Check this commit https://github.com/osbuild/osbuild-composer/commit/2d75b25ffe1d34717b808dc2a85bd704ea9bd0a8

jrusz (changes requested):
Some of the descriptions are not correct, I've added comments where changes are needed, feel free to ask any questions if there's something you don't understand. Also don't be afraid to ask the authors of the tests if you're unsure, they should know best what is the tests purpose.

test/cases/generation.sh
jabia99: I asked @lavocatt, and this is what I understood. If it's not clear enough we can think of another way to rephrase it. I think `generate-test-cases` needs dnf-json to be disabled on start so it can enable it. This script just disables this service

test/cases/libvirt.sh
jabia99: I'm trying to create this comments a bit "future proof" so they don't get outdated fast. Do you think the purpose is not clear or do you want to add more details?

jrusz (dismissed):
LGTM.

jrusz (approved):

PR #2350, opened 2022-02-23
Manifests test: ensure that every image type has test coverage

thozza:
- Extend the manifests test to ensure that each an every image type of each architecture and each distribution is covered by at least one image test case.
- Fetch the matrix of supported `distribution` x `architecture` x `image type` combinations directly from composer and deprecate the originally used JSON file with this information.
- Add repo definitions used by image tests for RHEL-8.7 and Fedora 36 (rawhide).

TODO:

- [x] Add RHEL-87/91 and F36 image test cases
- [x] Add missing image test cases for other distros

This pull request includes:


internal/distro/rhel90/distro.go
```diff
@@ -95,7 +95,11 @@ var distroMap = map[string]distribution{
 		vendor:             "redhat",
 		ostreeRefTmpl:      "rhel/9/%s/edge",
 		isolabelTmpl:       "RHEL-9-1-0-BaseOS-%s",
+<<<<<<< HEAD
 		runner:             "org.osbuild.rhel91",
+=======
+		runner:             "org.osbuild.rhel90",
+>>>>>>> b5e67fcf (Distro: alias RHEL-9.1 to RHEL-9.0)
```
gicmo: That looks wrong.

internal/distro/rhel90/distro.go
```diff
@@ -638,6 +642,10 @@ func NewHostDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 	return newDistro("rhel-90")
 }
 
+<<<<<<< HEAD
+=======
+// New creates a new distro object, defining the supported architectures and image types
+>>>>>>> b5e67fcf (Distro: alias RHEL-9.1 to RHEL-9.0)
```
gicmo: That also looks wrong.

tools/test-case-generators/generate-all-test-cases
```diff
@@ -1290,9 +1290,15 @@ class ManifestTestCaseMatrixGenerator(BaseTestCaseMatrixGenerator):
...
                     try:
+<<<<<<< HEAD
                         subprocess.check_call([f"{script_dir}/generate-test-cases", "--distro", f"{distro_name}", "--arch",
                             f"{arch_name}", "--image-types", f"{image_type}", "--store", "/dev/null", "--output", f"{self.output}",
                             "--keep-image-info"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+=======
+                        subprocess.run([f"{script_dir}/generate-test-cases", "--distro", f"{distro_name}", "--arch",
+                            f"{arch_name}", "--image-types", f"{image_type}", "--store", "/dev/null", "--output", f"{self.output}",
+                            "--keep-image-info"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True, encoding="utf-8")
+>>>>>>> 2f618e3b (generate-all-test-cases: add `manifests` command)
```
gicmo: This looks wrong.

internal/distro/rhel84/distro.go
```diff
@@ -1669,7 +1669,6 @@ func newDistro(name, modulePlatformID, ostreeRef string, isCentos bool) distro.D
 		uefi: false,
 	}
 	s390x.addImageTypes(
-		tarImgType,
```
gicmo, line 1672:
```
Building `tar` image for `s390x` on RHEL-84 ends with panic:
"s390x image must have a partition table, this is a programming error"
```

A tar image should not need a partition table, so this error message is funny. Removing it surely will fix the issue but I wonder why we are seeing this in the first place.

internal/distro/rhel90/distro.go
```diff
@@ -95,7 +95,7 @@ var distroMap = map[string]distribution{
 		vendor:             "redhat",
 		ostreeRefTmpl:      "rhel/9/%s/edge",
 		isolabelTmpl:       "RHEL-9-1-0-BaseOS-%s",
-		runner:             "org.osbuild.rhel91",
```
thozza, line 98: This one should not change.

internal/distro/rhel90/distro.go
```diff
@@ -638,6 +638,7 @@ func NewHostDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 	return newDistro("rhel-90")
 }
 
+// New creates a new distro object, defining the supported architectures and image types
```
thozza, line 641: probably a leftover from rebasing

internal/distro/rhel90/distro.go
jabia99: Then this commit will be empty, right?

thozza: yes, it should have been dropped from the PR 😉

tools/test-case-generators/format-request-map.json
```diff
@@ -513,7 +513,14 @@
...
     "no-image-info": true,
-    "overrides": {}
+    "overrides": {
+      "rhel-85": {
+        "blueprint": {}
+      },
+      "rhel-90-beta": {
+        "blueprint": {}
+      }
+    }
```
thozza, line 523: Is this really needed for 8.5 and 9.0 beta? IIRC, the support for user customizations was backported to 8.5 and 9.o beta via https://github.com/osbuild/osbuild-composer/pull/2516

tools/test-case-generators/format-request-map.json
```diff
@@ -810,6 +841,81 @@
           }
         }
       },
+      "fedora-36": {
```
thozza, line 844: I wonder whether this is needed (and a separate section for each Fedora version).  Was the test case generation failing without this?

tools/test-case-generators/format-request-map.json
```diff
@@ -885,6 +991,81 @@
           }
         }
       },
+      "rhel-84": {
```
thozza, line 994: I have the same comment as for the `fedora-36` override... Was this needed?

thozza (commented):
Since I'm a co-author, I won't approve.

Thanks for finishing my PR! I really appreciate your work. I added some comments, but don't consider any of them a blocker for merging. Happy to address them as a follow-up.

tools/test-case-generators/format-request-map.json
jabia99: Yes, it was failing with this error:
`The following custom mountpoints are not supported [“/opt”]`

tools/test-case-generators/format-request-map.json
jabia99: Same error as in fedora 36

tools/test-case-generators/format-request-map.json
jabia99: The problem was the fdo settings are not supported for those versions. I got the next error:
`panic: boot ISO image type "edge-simplified-installer" contains unsupported blueprint customizations: 'FDO' is not allowed`

tools/test-case-generators/format-request-map.json
thozza: this should not be an issue any more -> https://github.com/osbuild/osbuild-composer/blob/809afbd0ad0d33aaf3f6998ee46e49af22582b83/internal/distro/fedora/distro.go#L49

jabia99: Cool, I just deleted it and updated the manifests

tools/test-case-generators/generate-all-test-cases
```diff
@@ -1290,9 +1290,9 @@ class ManifestTestCaseMatrixGenerator(BaseTestCaseMatrixGenerator):
                     script_dir = os.path.dirname(__file__)
                     log.info("Generating manifest for %s-%s-%s", distro_name, arch_name, image_type)
                     try:
-                        subprocess.check_call([f"{script_dir}/generate-test-cases", "--distro", f"{distro_name}", "--arch",
+                        subprocess.run([f"{script_dir}/generate-test-cases", "--distro", f"{distro_name}", "--arch",
                             f"{arch_name}", "--image-types", f"{image_type}", "--store", "/dev/null", "--output", f"{self.output}",
-                            "--keep-image-info"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+                            "--keep-image-info"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True, encoding="utf-8")
```
achilleas-k, line 1295: The commit message for this change seems to have been copied from 8ee90334bab71920884914963aef6e20c3e26a9d and doesn't match the change.

achilleas-k (commented):
Looks good.  There's one commit message that doesn't match the change though.

tools/test-case-generators/generate-all-test-cases
thozza: I cherry-picked that change and submitted is as a separate PR, which has been already merged. I think that the change makes sense, because if the generation fails, there won't be any output. However, the commit message should be definitely adjusted.

jabia99: I think at some point this was part of that commit. @thozza, can you tell me a better commit message pls?

ondrejbudai (commented):
Looks good to me (apart from the bad commit message, see above).

tools/test-case-generators/generate-all-test-cases
jabia99: I'll change it. But I have no idea of the context of this change. What would be an appropriate message?

thozza: something like "Capture stdout and stderr output when running `generate-test-cases` in the `manifests` command. This is helpful for debugging test case generation failures."

jabia99: Thanks! done

achilleas-k (approved):
Awesome.  Thank you.

PR #2352, opened 2022-02-23
Implement HTTP retries for koji jobs

### LINKED ISSUE #2335: Koji client should use retries
**Describe the bug**
As connections are flaky it makes sense to retry xmlrpc calls in case of network error. ``internal/upload/koji/koji.go`` use simple ``http.RoundTripper`` transport instead of something with retry/backoff policy (according to my reading, correct me if I'm wrong). 

**Expected behavior**
Even temporary unavailability of koji server shouldn't fail the import or similar tasks.

**Additional context**
There are tens of thousands of connections (just chunked upload of results will cause tons of calls) so even a very small fraction of failing calls will fail the final task as it can't be completed. Retries are suggested way to handle network issues or temporary server unavailability for koji clients.

diaasami:
and log number of retries for metrics

Fixes #2335

This pull request includes:


ondrejbudai (approved):
Maybe a quick unit-test against an http server that drops each second packet? I'm not sure how difficult that would be.

teg (approved):
Looks great! We may also cover the non-upload API calls, but seems much less critical, so can totally do that in a follow-up.

ondrejbudai (changes requested):
I think you also need to add the new dependency to the spec file to fix Fedora build.

Otherwise, this looks good.

internal/upload/koji/koji.go
```diff
@@ -304,16 +309,47 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
...
 	u.RawQuery = q.Encode()
 
-	client := http.Client{Transport: k.transport}
+	retries := uint(0)
+
+	countingCheckRetry := func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+		retries++
+		shouldRetry, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+
+		if !shouldRetry && err != nil {
```
ondrejbudai, line 318: These 3 conditions are non-trivial, can you please add a comment describing the decision logic?

internal/upload/koji/koji.go
```diff
@@ -20,8 +23,10 @@ import (
 	"os"
 
 	"github.com/kolo/xmlrpc"
+	logrus "github.com/sirupsen/logrus"
```
ondrejbudai, line 26: Just a nitpick, is the `logrus` alias really needed? (maybe it is and I just don't know Go because my IDE is doing this for me :D )

ondrejbudai (changes requested):
Looking nice but I would really like to see some commentary in the code.

internal/upload/koji/koji.go
diaasami: I have added some comments to explain what's going on.
I can add more, let me know what you think. The conditions are the same as in DefaultRetryPolicy (except the first one), both trying to make sure it's the right error before taking a decision to retry or not.

gicmo:
See also https://pagure.io/koji/pull-request/3269

ondrejbudai (changes requested):
Awesome, one small nit: can you squash the commits?

ondrejbudai (dismissed):
Very nice, thanks a lot for looking into this.

ondrejbudai (approved):
Thanks!

Issue #2348, opened 2022-02-23
RHEL 9.0: Auto convert to LVM layout if there are customisations for new mount points


gicmo:
If the users request a new mountpoint we should automatically switch to LVM layout, but only on RHEL 9.0.

PR #2354, opened 2022-02-23
RHEL9.0: convert layout to LVM on filesystem customisations

gicmo:
Whenever we create a new mountpoint due to a user customization, ensure the layout uses LVM, i.e. convert plain layouts to it, if needed. This does not apply to rpm-ostree based systems, e.g. the simplified installer since they will be using LUKS in 9.0.

Closes: #2348


internal/disk/lvm.go
```diff
@@ -128,3 +132,12 @@ func (lv *LVMLogicalVolume) EnsureSize(s uint64) bool {
...
 	return false
 }
+
+// lvname returns a name for a logical volume based on the mountpoint.
+func lvname(path string) string {
+	if path == "/" {
+		return "rootlv"
+	}
+
+	return strings.ReplaceAll(path, "/", "") + "lv"
```
teg, line 142: Does this need to be unique? The mapping is not injective, so if uniqueness is required this won't work.

internal/osbuild2/device.go
```diff
@@ -67,7 +67,9 @@ func GenDeviceCreationStages(pt *disk.PartitionTable, filename string) []*Stage
 			volumes := make([]LogicalVolume, len(ent.LogicalVolumes))
 			for idx, lv := range ent.LogicalVolumes {
 				volumes[idx].Name = lv.Name
-				volumes[idx].Size = fmt.Sprintf("%d", lv.Size)
+				// NB: we need to specify the size in bytes, since lvcreate
+				// defaults to megabytes
```
teg, line 71: Commit message is wrong, but comment is right (bytes v megabytes).

internal/distro/rhel90/distro.go
```diff
@@ -382,7 +400,9 @@ func (t *imageType) getPartitionTable(
 
 	imageSize := t.Size(options.Size)
 
-	return disk.NewPartitionTable(&basePartitionTable, mountpoints, imageSize, false, rng)
+	lvmify := !t.rpmOstree
```
teg, line 403: Hm, do we do LVM unconditionally for non-ostree? Or am I misunderstanding the logic?

teg: Ah, no, all good. `lvmify` only has an effect in case `mountpoints` is non-empty.

teg:
Looks good to me, though did we intentionally only do this for RHEL9, or did we intend/communicate that we would also do it for RHEL8? I genuinely forgot.

internal/disk/lvm.go
gicmo: Yeah, indeed. Good point.

internal/osbuild2/device.go
gicmo: Will fix.

gicmo:
> Looks good to me, though did we intentionally only do this for RHEL9, or did we intend/communicate that we would also do it for RHEL8? I genuinely forgot.

I think the same reasoning why we go with the change for 9 also applies to 8, but I think we can do that in a follow up. Let me clean this one up for now.

internal/disk/lvm.go
```diff
@@ -128,3 +132,13 @@ func (lv *LVMLogicalVolume) EnsureSize(s uint64) bool {
...
 }
+
+// lvname returns a name for a logical volume based on the mountpoint.
+func lvname(path string) string {
+	if path == "/" {
+		return "rootlv"
+	}
+
+	path = strings.TrimLeft(path, "/")
+	return strings.ReplaceAll(path, "/", "_") + "lv"
```
teg, line 143: `/` and `/root` will end up with the same volume name. As will `/home/foo_bar` and `/home/foo/bar`.

teg (changes requested):
All good except for the name generation algorithm.

internal/disk/lvm.go
gicmo: Yeah, I think we would need to solve this at a level higher and attach a suffix if the name already exists. `/` is not allowed in the name and any substation we do will end up with the same problem, no?

teg (dismissed):
Thanks!

teg (dismissed):
Looks great! Thanks!

achilleas-k (dismissed):
Awesome stuff.  LGTM

achilleas-k (approved):

PR #2358, opened 2022-02-24
Add Azure Marketplace images

gicmo:
Add support for building images for the Azure marketplace: add a new image type "azure-rhui" that can be used to build images tailored to the Azure marketplace.
Add two sample manifests for 8.5 and 8.6, but note that even the 8.5 is using the 8.6 distro definitions. Also no image-info is included since `image-info` cannot (yet) handle LVM setups and the azure marketplace images use the LVM setup.


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -630,6 +630,7 @@ components:
         - aws-ha-rhui
         - aws-sap-rhui
         - azure
+        - azure-rhui
```
teg, line 633: We need to extend koji api too, as it is still in use by brew.

internal/distro/rhel86/distro.go
```diff
@@ -843,6 +843,121 @@ func newDistro(distroName string) distro.Distro {
 		basePartitionTables: defaultBasePartitionTables,
 	}
 
+	azureImgType := imageType{
+		name:     "azure",
```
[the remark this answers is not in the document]
gicmo: Yeah, why not. I tried to mimic ec2 and ami but this is a bit confusing so having it called the same everywhere is probably the better idea.

internal/cloudapi/v2/openapi.v2.yml
teg: Scratch this, it works automagically.

teg (approved):
Needs a PR description, but otherwise looking great!

teg (approved):

internal/distro/rhel86/distro.go
```diff
@@ -843,6 +843,121 @@ func newDistro(distroName string) distro.Distro {
 		basePartitionTables: defaultBasePartitionTables,
 	}
 
+	azureRhuiImgType := imageType{
+		name:     "azure-rhui",
+		filename: "disk.vhd",
+		mimeType: "application/x-vhd",
+		packageSets: map[string]packageSetFunc{
+			buildPkgsKey: ec2BuildPackageSet,
```
achilleas-k, line 851: I guess we might rename this since it's not just for ec2 anymore.
But later.

achilleas-k (approved):
LGTM

PR #2359, opened 2022-02-24
Set selinux to permissive mode for installers

achilleas-k:
This PR sets selinux to permissive mode in all installers in RHEL 8.6 and RHEL 9.0 (see [rhbz#2031939](https://bugzilla.redhat.com/show_bug.cgi?id=2031939)).

I also added configs to the `format-request-map` to be able to generate test manifests for `edge-installer` and `edge-simplified-installer` and updated the Go test to use the ostree parameters from the compose request in the test file to verify.


teg:
I could not find where we do the selinux labelling of the installer trees... How does this work correctly if the trees are unlabelled and selinux is enforcing?

thozza (dismissed):
Everything looks good. Thanks for adding the configs to `format-request-map`. I was adding it as well in one PR that I have in progress, but didn't do the necessary changes to pass OSTree parameters. So I'm glad that you did it 🙂

thozza:
> I could not find where we do the selinux labelling of the installer trees... How does this work correctly if the trees are unlabelled and selinux is enforcing?

My gut feeling is that we need to have the selinux labeling stage only in pipelines which are creating some configuration files to ensure that those are correctly labeled. But I'm not 100% sure...

teg:
> > I could not find where we do the selinux labelling of the installer trees... How does this work correctly if the trees are unlabelled and selinux is enforcing?
> 
> My gut feeling is that we need to have the selinux labeling stage only in pipelines which are creating some configuration files to ensure that those are correctly labeled. But I'm not 100% sure...

Ah, I thought we completely disabled SELinux from the rpm stage so you'd end up with a completely unlabelled tree if you don't relabel. But I could very well be wrong on that.

thozza:
> > > I could not find where we do the selinux labelling of the installer trees... How does this work correctly if the trees are unlabelled and selinux is enforcing?
> > 
> > 
> > My gut feeling is that we need to have the selinux labeling stage only in pipelines which are creating some configuration files to ensure that those are correctly labeled. But I'm not 100% sure...
> 
> Ah, I thought we completely disabled SELinux from the rpm stage so you'd end up with a completely unlabelled tree if you don't relabel. But I could very well be wrong on that.

I don't see any mentions of selinux in the osbuild rpm stage. And this comment in the selinux stage supports my assumption - https://github.com/osbuild/osbuild/blob/97905f6e0c27ab968694e1771a94249143a18b0a/stages/org.osbuild.selinux#L17-L19

achilleas-k:
Dropped the change from `edge-simplified-installer`

teg (dismissed):

internal/distro/rhel90/pipelines.go
```diff
@@ -805,6 +805,7 @@ func simplifiedInstallerTreePipeline(repos []rpmmd.RepoConfig, packages []rpmmd.
 		dracutStageOptions.Install = []string{"/fdo_diun_pub_key_root_certs.pem"}
 	}
 	p.AddStage(osbuild.NewDracutStage(dracutStageOptions))
+
```
gicmo, line 808: Omg, hope David doesn't see that!

gicmo (commented):
Is `permissive` also what the standard installer uses? I was thinking if the concern is AVC messages in the logs, wouldn't they also happen for `permissive`?

internal/distro/rhel90/pipelines.go
achilleas-k: See what? >_>

gicmo: sneaky!

achilleas-k:
> Is `permissive` also what the standard installer uses? 

Yes (according to linked bug and my own verification).

> I was thinking if the concern is AVC messages in the logs, wouldn't they also happen for `permissive`?

I think the main concern is being compatible with the official ISO.

teg:
We seem to have CS9 issues again. I did not dig deeply, so please check first, but I would be ok with ignoring CS9 as a matter of principle if we have reason to believe the bug is in the compose.

We really need to test against snapshots 👼 

cc @ochosi

gicmo:
> We seem to have CS9 issues again. I did not dig deeply, so please check first, but I would be ok with ignoring CS9 as a matter of principle if we have reason to believe the bug is in the compose.

Something is up with libvirt and UEFI:
`ERROR    Libvirt version does not support UEFI.`

achilleas-k:
It's the bug flavour of the hour :roll_eyes:

gicmo:
GREEN, but I merged the `C.UTF-8` PR and it now collides I think. Needs a rebase on main.

gicmo:
I have rebased it on main and re-ran `./tools/test-case-generators/generate-test-cases --store /tmp/store --output test/data/manifests --distro {centos-9,rhel-90} --arch {aarch64,x86_64} --image-types edge-simplified-installer --keep-image-info`

gicmo (approved):
💿

PR #2361, opened 2022-02-25
cmd/osbuild-worker: dnf-json returns MarkingErrors (plural)

Gundersanne:


kingsleyzissou (approved):
Thanks!

ondrejbudai (approved):
Good catch! We should probably refactor the `dnf-json` API at some point to have a proper OpenAPI spec. That would have helped in this case I think.

PR #2369, opened 2022-02-27
PartitionTable clone fix

achilleas-k:
In LUKSContainer and LVMLogicalVolume we neglected to clone the Payload which means we would modify the base PartitionTable when manipulating the clone.

Added test that collects all entities before and after cloning the test partition tables and compares for duplicates between the lists.

### DIFF
--- a/internal/disk/disk_test.go
+++ b/internal/disk/disk_test.go
@@ -1,6 +1,7 @@
 package disk
 
 import (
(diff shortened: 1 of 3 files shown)


gicmo (approved):
❤️ Thanks a lot!

teg (dismissed):

PR #2373, opened 2022-02-28
rhel86: automatically convert to LVM on fs customizations

gicmo:
Whenever we create a new mountpoint due to a user customization, ensure the layout uses LVM, i.e. convert plain layouts to it, if needed. It uses the existing lvm-ification code but enhances it so that we also create a `/boot` partition in case it does not yet exist.
Adjust the existing tests that assumed we can not create more than 4 partitions on mbr layouts, since that is now not true anymore.

### DIFF
--- a/internal/disk/disk_test.go
+++ b/internal/disk/disk_test.go
@@ -124,6 +124,45 @@ var testPartitionTables = map[string]PartitionTable{
 		},
 	},
 
+	"plain-noboot": {
+		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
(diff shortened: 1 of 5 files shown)


achilleas-k (approved):
Looks good.  Should we add a small unit test that checks for a boot partition on the PT when it's lvmified?

gicmo:
> Looks good. Should we add a small unit test that checks for a boot partition on the PT when it's lvmified?

good idea, maybe also specify min size for the root partition

achilleas-k (approved):

PR #2376, opened 2022-03-01
Update snapshots to 20220301

schutzbot:
Job(s) succeeded: 146
Job(s) failed: 1
Workflow run: https://github.com/osbuild/rpmrepo/actions/runs/1913593578


teg:
Would be nice if we could make this the default, i.e. only pin repos in one place and generate the test data from the schutzfile

gicmo:
> Would be nice if we could make this the default, i.e. only pin repos in one place and generate the test data from the schutzfile

I argued for this but @achilleas-k mentioned there were reasons not too, but he couldn't remember them exactly.

achilleas-k:
> > Would be nice if we could make this the default, i.e. only pin repos in one place and generate the test data from the schutzfile
> 
> I argued for this but @achilleas-k mentioned there were reasons not too, but he couldn't remember them exactly.

I believe one of the reasons was manifest / test case generation.  We'd have to be recreating all of them every time we update the Schutzfile.  Which brings us to automating it (again).

I've searched through the conversations that lead us here and can't see any other issues being brought up, though I'd consult with others before finalising.

thozza:
> > > Would be nice if we could make this the default, i.e. only pin repos in one place and generate the test data from the schutzfile
> > 
> > 
> > I argued for this but @achilleas-k mentioned there were reasons not too, but he couldn't remember them exactly.
> 
> I believe one of the reasons was manifest / test case generation. We'd have to be recreating all of them every time we update the Schutzfile. Which brings us to automating it (again).
> 
> I've searched through the conversations that lead us here and can't see any other issues being brought up, though I'd consult with others before finalising.

The prerequisite would have to be that image-info report is not part of the image test any more, otherwise we would have to rebuild all images on all architectures, which is currently not automated (mainly provisioning). 

Updating repos used for image tests (manifests) IMO does not bring much value and would generate just too much noise. The purpose of image tests is to ensure that manifests do not change for the same image request (distro + arch + image type + customizations + repos). If there is an intentional change of image definition or pipelines, then it makes sense to intentionally regenerate image test cases and if needed. However updating the used repos with the image definition itself can generate so much changes in the image test, that it makes it very hard to reasonably review the diff.

runcom:
This is good to go now 🎉

achilleas-k:
Rebasing to get #2385 in to make the CS9 Koji test green.

achilleas-k (approved):

gicmo:
Everything green but the installer.

PR #2387, opened 2022-03-03
`rhel86` various fixes for the `azure-rhui` image

gicmo:
- Sync the package set with the existing market place image
- Update sysconfig to set `kernel-core` as default kernel
- Remove `authconfig` call
- Import GPG keys from the tree(!) via a new `org.osbuild.rpm` stage option.

NB: the last point needs https://github.com/osbuild/osbuild/pull/985


internal/distro/image_config.go
```diff
@@ -20,6 +20,8 @@ type ImageConfig struct {
 	DefaultTarget       string
 	Sysconfig           []*osbuild2.SysconfigStageOptions
 
+	GPGKeyFiles []string
```
thozza, line 23: I would suggest to add "RPM" to the variable name to make it obvious what type of GPG keys these are.

In addition, you need to extend the `InheritFrom` method due to this new member.

thozza (changes requested):
The `InheritFrom` method needs to be extended, otherwise this looks good to me.

internal/distro/image_config.go
gicmo: Good catch! We should enhance that to use reflection in a future version (and maybe switch it do clone and override ;) ;)

thozza: _nods_

thozza:
@gicmo one more note, I think that the osbuild version dependency in the SPEC should be bumped with this PR, because it depends on a specific osbuild version (the one next one - not released yet)

gicmo:
@thozza I did not do that because I was wondering if a downstream patch for osbuild is enough of if we can deploy `main`. I am shaky on those details.

thozza:
> @thozza I did not do that because I was wondering if a downstream patch for osbuild is enough of if we can deploy `main`. I am shaky on those details.

If we don't do it now, we will IMO most probably forget to do it at all. If we would deploy from `main`, we would use a CI build and that already uses the bumped version (https://gitlab.com/osbuild/ci/osbuild/-/jobs/2162857403#L1028) so this is not a blocker.

gicmo:
> > @thozza I did not do that because I was wondering if a downstream patch for osbuild is enough of if we can deploy `main`. I am shaky on those details.
> 
> Yeah, this is a bit fuzzy...
> 
> If we don't do it now, we will IMO most probably forget to do it at all. If we would deploy from `main`, we would use a CI build and that already uses the bumped version (https://gitlab.com/osbuild/ci/osbuild/-/jobs/2162857403#L1028) so this is not a blocker.

Excellent point! Will fix as well!

internal/distro/image_config.go
gicmo: I added a comment for now, can always rename it later if it collides with something.

thozza (approved):
LGTM

internal/distro/image_config.go
thozza: It probably won't collide, I would just find "RPM" in the variable name easier to understand when one uses the structure, compared to a comment in its definition. But I agree with you and think that the current state is OK to merge.

thozza (dismissed):

ondrejbudai (approved):

PR #2389, opened 2022-03-04
nightly: update GPG key in prepare-rhel-internal & fix some tests

jrusz:
This pull request includes:


jrusz:
SHA-1 key is no longer supported. Updated keys used in `prepare-rhel-internal.sh` with a new key from `redhat-release` package.

thozza (approved):

thozza (approved):

test/cases/image_tests.sh
```diff
@@ -85,8 +87,10 @@ cd $WORKING_DIRECTORY
 OSBUILD_LABEL=$(matchpathcon -n "$(which osbuild)")
 sudo chcon "$OSBUILD_LABEL" /usr/libexec/osbuild-composer-test/image-info
 
+get_test_cases
 # Run each test case.
-for TEST_CASE in $(get_test_cases); do
+for TEST_CASE in "${TEST_CASES[@]}"; do
```
atodorov, line 92: I don't understand why not keep using the function like before? Surely the function itself can return an appropriate result, no ?

atodorov (changes requested):
> These test cases were never meant to be built, they exist only for the
purpose of manifest testing.

I don't understand this message in commit log. Can you make it more clear. Where are these json files used and why we don't want them consumed in image_tests.sh ?

thozza:
> > These test cases were never meant to be built, they exist only for the
> > purpose of manifest testing.
> 
> I don't understand this message in commit log. Can you make it more clear. Where are these json files used and why we don't want them consumed in image_tests.sh ?

I'll try to provide some additional context.

Image tests (manifests) in `osbuild-composer` have multiple purposes. We use manifests in `osbuild-composer` unit and CI tests to ensure that the image definitions don't change unintentionally. The expectation is that if we provide the same input to composer (repositories and blueprint), we will always get the same manifest for a given image type / arch / distribution.

In addition to this, we use these manifest in `osbuild` CI and try to actually build images using them. However some image types (e.g. the edge installer ones) require additional external infrastructure to be built. E.g. setting up a HTTP ostree repository with a commit, which can be pulled while building the image. Since the setup is a bit complicated, we decided to at least generate image tests for these images in composer with dummy values. These are good enough for composer to generate the manifest, so that we can keep checking that the definition does not change unintentionally. However since the values are dummy (e.g. a non-existing URL to ostree repo), the manifest can not be really successfully built using `osbuild`.

Due to the fact that `osbuild` and `osbuild-composer` share the same image test manifests, we need to filter out manifests with intentionally dummy values in `osbuild` CI in order for it to pass.

test/cases/image_tests.sh
```diff
@@ -87,6 +93,7 @@ sudo chcon "$OSBUILD_LABEL" /usr/libexec/osbuild-composer-test/image-info
 
 # Run each test case.
 for TEST_CASE in $(get_test_cases); do
+    # skip these until a better solution is implemented
```
atodorov, line 96: Nitpick - this comment is probably related to line 52.

jrusz:
Regular pipeline has expected failures: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/485914852

For RHEL-8.6 an image test failure fill be fixed in another PR and for RHEL-9.0 the Installer and Upgrade tests failures are known, reported in BZ and should be fixed by their relevant teams.

RHEL-8.6 nightly pipeline: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/485916940 
RHEL-9.0 nightly pipeline: https://gitlab.com/osbuild/ci/osbuild-composer/-/pipelines/485915247

jrusz:
I've pushed again to get more logs for Upgrade tests and to remove on redundant comment.

atodorov (approved):

ondrejbudai (approved):

Issue #2386, opened 2022-03-03
distro: depsolving blueprint packages separately pulls excluded packages into a vanilla image


thozza:
**Describe the bug**
Since RHEL-8.5 (https://github.com/osbuild/osbuild-composer/commit/cba720e63fff2fcc4952794d017c89ba3a0d18da),  we depsolve blueprint packages separately. Even if the package list in the blueprint is empty, we always add the kernel package name into it, so even for vanilla image builds. 
https://github.com/osbuild/osbuild-composer/blob/c79f51942bd494c1222812424a327638219d58f2/internal/distro/rhel86/distro.go#L352-L358

This gets depsolved and ends up pulling gazillion of packages due to transitive RPM dependencies of the kernel package. We then merge the list with the depsolved os package set.
https://github.com/osbuild/osbuild-composer/blob/c79f51942bd494c1222812424a327638219d58f2/internal/distro/rhel86/pipelines.go#L368

The final package set installed in a vanilla image (without any blueprint customizations or additional packages) contains even packages explicitly excluded in the default image package set.

Example of the result of depsolving `{Include:["kernel"] Exclude:[]}` `PackageSet` for RHEL-8.5: http://pastebin.test.redhat.com/1033784

As a result, it is not possible to have an image in RHEL-8 which would not contain `glibc-all-langpacks`, which has 25 MB.

**Environment**
 * OS version (`/etc/os-release` and `/etc/redhat-release`): any, RHEL images
 * osbuild-composer version (`rpm -qi osbuild-composer)`: latest `main`

**Expected behavior**
Packages explicitly excluded in the image's default package set should not be installed in the image built using an empty blueprint (not containing any additional packages to be installed or customizations).

thozza:
fixed via #2563

PR #2392, opened 2022-03-05
Support for fedora-iot-commit with Fedora aarch64

jordigilh:
This pull request includes:

This PR adds the `fedora-iot-commit` for aarch64 in fedora. As part of the effort to include the new type, the code has been refactored to align with the existing design in the RHEL distributions, mainly the usage of a `map[string]rmmd.PackageSet` structure to store the different packages (build and OS packages) before merging them into a single list for Included and Excluded.

The list of packages align with the ones listed here:
https://pagure.io/fedora-iot/ostree/blob/1c0a8aa02f10e1a9b2a8c9802d8bd7028fca5752/f/fedora-iot-base.json

@gicmo @achilleas-k can you take a look when you have some spare time? I'd like to hear your input on this.

### DIFF
--- a/internal/distro/fedora/distro.go
+++ b/internal/distro/fedora/distro.go
@@ -1,4 +1,4 @@
-package fedora33
+package fedora
 
 import (
 	"encoding/json"
--- a/internal/distro/fedora/distro.go
+++ b/internal/distro/fedora/distro.go
@@ -58,7 +58,8 @@ type architecture struct {
 	bootloaderPackages []string
 	legacy             string
 	bootType           distro.BootType
-	imageTypes         map[string]imageType
+	imageTypes         map[string]distro.ImageType
+	imageTypeAliases   map[string]string
 }
 
 type packageSetFunc func(t *imageType) rpmmd.PackageSet
@@ -66,10 +67,10 @@ type packageSetFunc func(t *imageType) rpmmd.PackageSet
 type imageType struct {
 	arch             *architecture
 	name             string
+	nameAliases      []string
 	filename         string
 	mimeType         string
 	packageSets      map[string]packageSetFunc
-	excludedPackages map[string]packageSetFunc
 	enabledServices  []string
 	disabledServices []string
 	kernelOptions    string
@@ -140,32 +141,37 @@ func (a *architecture) ListImageTypes() []string {
 	return formats
 }
 
-func (a *architecture) GetImageType(imageType string) (distro.ImageType, error) {
-	t, exists := a.imageTypes[imageType]
+func (a *architecture) GetImageType(name string) (distro.ImageType, error) {
+	t, exists := a.imageTypes[name]
 	if !exists {
-		return nil, errors.New("invalid image type: " + imageType)
+		aliasForName, exists := a.imageTypeAliases[name]
+		if !exists {
+			return nil, errors.New("invalid image type: " + name)
+		}
+		t, exists = a.imageTypes[aliasForName]
+		if !exists {
+			panic(fmt.Sprintf("image type '%s' is an alias to a non-existing image type '%s'", name, aliasForName))
+		}
 	}
-
-	return &t, nil
+	return t, nil
 }
 
-func (a *architecture) setImageTypes(imageTypes ...imageType) {
-	a.imageTypes = map[string]imageType{}
-	for _, it := range imageTypes {
-		a.imageTypes[it.name] = imageType{
-			arch:             a,
-			name:             it.name,
-			filename:         it.filename,
-			mimeType:         it.mimeType,
-			packageSets:      it.packageSets,
-			excludedPackages: it.excludedPackages,
-			enabledServices:  it.enabledServices,
-			disabledServices: it.disabledServices,
-			kernelOptions:    it.kernelOptions,
-			bootable:         it.bootable,
-			rpmOstree:        it.rpmOstree,
-			defaultSize:      it.defaultSize,
-			assembler:        it.assembler,
+func (a *architecture) addImageTypes(imageTypes ...imageType) {
+	if a.imageTypes == nil {
+		a.imageTypes = map[string]distro.ImageType{}
+	}
+	for idx := range imageTypes {
+		it := imageTypes[idx]
+		it.arch = a
+		a.imageTypes[it.name] = &it
+		for _, alias := range it.nameAliases {
+			if a.imageTypeAliases == nil {
+				a.imageTypeAliases = map[string]string{}
+			}
+			if existingAliasFor, exists := a.imageTypeAliases[alias]; exists {
+				panic(fmt.Sprintf("image type alias '%s' for '%s' is already defined for another image type '%s'", alias, it.name, existingAliasFor))
+			}
+			a.imageTypeAliases[alias] = it.name
 		}
 	}
 }
@@ -749,9 +755,10 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 		"parsec", "dbus-parsec"}
 
 	iotImgType := imageType{
-		name:     "fedora-iot-commit",
-		filename: "commit.tar",
-		mimeType: "application/x-tar",
+		name:        "fedora-iot-commit",
+		nameAliases: []string{"iot-commit", "edge-commit"},
+		filename:    "commit.tar",
+		mimeType:    "application/x-tar",
 		packageSets: map[string]packageSetFunc{
 			buildPkgsKey: iotBuildPackageSet,
 			osPkgsKey:    iotCommitPackageSet,
@@ -906,7 +913,7 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 		},
 	}
 
-	x86_64.setImageTypes(
+	x86_64.addImageTypes(
 		iotImgType,
 		amiImgType,
 		qcow2ImageType,
@@ -916,7 +923,7 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
 		ociImageType,
 	)
 
-	aarch64.setImageTypes(
+	aarch64.addImageTypes(
 		iotImgType,
 		amiImgType,
 		qcow2ImageType,
--- a/tools/test-case-generators/format-request-map.json
+++ b/tools/test-case-generators/format-request-map.json
@@ -171,7 +171,10 @@
         }
       }
     },
-    "overrides": {}
+    "overrides": {},
+    "supported_arches": [
+      "x86_64"
+    ]
   },
   "openstack": {
     "boot": {
--- a/cmd/osbuild-dnf-json-tests/main_test.go
+++ b/cmd/osbuild-dnf-json-tests/main_test.go
@@ -16,7 +16,7 @@ import (
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/distro"
-	fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/test"
 )
--- a/cmd/osbuild-store-dump/main.go
+++ b/cmd/osbuild-store-dump/main.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
 	"github.com/osbuild/osbuild-composer/internal/distro"
-	fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
+	"github.com/osbuild/osbuild-composer/internal/distro/fedora"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/store"
 	"github.com/osbuild/osbuild-composer/internal/target"
--- a/internal/blueprint/customizations_test.go
+++ b/internal/blueprint/customizations_test.go
@@ -19,7 +19,7 @@ func TestCheckAllowed(t *testing.T) {
 	GID := 321
 
 	expectedUsers := []UserCustomization{
-		UserCustomization{
+		{
 			Name:        "John",
 			Description: &Desc,
 			Password:    &Pass,
@@ -80,7 +80,7 @@ func TestGetKernel(t *testing.T) {
 func TestSSHKey(t *testing.T) {
 
 	expectedSSHKeys := []SSHKeyCustomization{
-		SSHKeyCustomization{
+		{
 			User: "test-user",
 			Key:  "test-key",
 		},
@@ -111,7 +111,7 @@ func TestGetUsers(t *testing.T) {
 	GID := 321
 
 	expectedUsers := []UserCustomization{
-		UserCustomization{
+		{
 			Name:        "John",
 			Description: &Desc,
 			Password:    &Pass,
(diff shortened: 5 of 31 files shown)


teg:
Thanks for the contribution, this is very relevant, love that it is getting attention!

@nullr0ute could you have a look?

achilleas-k:
Thanks for the great work Jordi!

@gicmo & @thozza 
How do we want to handle the package name and Fedora versions?  Should we just rename it to `fedora35` and drop F33?  Or maybe keep the F33 one around frozen for F34 until it's EOL (for another 2-3 months) and make this PR into F35 and soon F36?

thozza:
> @gicmo & @thozza How do we want to handle the package name and Fedora versions? Should we just rename it to `fedora35` and drop F33? Or maybe keep the F33 one around frozen for F34 until it's EOL (for another 2-3 months) and make this PR into F35 and soon F36?

I propose to rename it simply to `fedora`. I already did the preparation to make it super easy (in https://github.com/osbuild/osbuild-composer/commit/b9efe82bd76cc66d3d192df71ba565cc14ef1e44). The package is already imported as `fedora` in all of the places.

```bash
$ grep -r "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
cmd/osbuild-dnf-json-tests/main_test.go:        fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
cmd/osbuild-store-dump/main.go: fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
internal/distro/fedora33/distro_test.go:        fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
internal/distroregistry/distroregistry.go:      fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
internal/store/json_test.go:    fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
```

It is great that the Fedora distro is getting some attention. On the other hand, this will introduce the #2386 bug in Fedora. We could maybe test to try again merging the default and blueprint package set and resolve any conflicting `Include` / `Exclude` requests...

teg:
Should this be named `iot-commit` (without the fedora prefix), in the same way we are moving from `rhel-edge-commit` to `edge-commit`? Or maybe go all the way and just call it `edge-commit` in fedora too, as it is the upstream for edge. We can of course label it differently in the UI.

Peter, what are your thoughts on naming of these things?

jordigilh:
> Should this be named `iot-commit` (without the fedora prefix), in the same way we are moving from `rhel-edge-commit` to `edge-commit`? Or maybe go all the way and just call it `edge-commit` in fedora too, as it is the upstream for edge. We can of course label it differently in the UI.
> 
> Peter, what are your thoughts on naming of these things?

I've added the code for supporting aliases (based on rhel code), and also defined 2 aliases for `fedora-iot-commit`: `iot-commit` and `edge-commit`. Let me know if you'd prefer to limit to 1 alias at this time.

jordigilh:
> > @gicmo & @thozza How do we want to handle the package name and Fedora versions? Should we just rename it to `fedora35` and drop F33? Or maybe keep the F33 one around frozen for F34 until it's EOL (for another 2-3 months) and make this PR into F35 and soon F36?
> 
> I propose to rename it simply to `fedora`. I already did the preparation to make it super easy (in [b9efe82](https://github.com/osbuild/osbuild-composer/commit/b9efe82bd76cc66d3d192df71ba565cc14ef1e44)). The package is already imported as `fedora` in all of the places.
> 
> ```shell
> $ grep -r "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> cmd/osbuild-dnf-json-tests/main_test.go:        fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> cmd/osbuild-store-dump/main.go: fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> internal/distro/fedora33/distro_test.go:        fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> internal/distroregistry/distroregistry.go:      fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> internal/store/json_test.go:    fedora "github.com/osbuild/osbuild-composer/internal/distro/fedora33"
> ```
> 
> It is great that the Fedora distro is getting some attention. On the other hand, this will introduce the #2386 bug in Fedora. We could maybe test to try again merging the default and blueprint package set and resolve any conflicting `Include` / `Exclude` requests...

Renamed package `fedora33` and `fedora33-test` to be `fedora` and `fedora-test` respectively. Tomorrow I'll take a look at the bug #2386 but if you can drop me any guidance that would help me to address it quickly :smile:

thozza:
> Renamed package `fedora33` and `fedora33-test` to be `fedora` and `fedora-test` respectively. Tomorrow I'll take a look at the bug #2386 but if you can drop me any guidance that would help me to address it quickly 😄

Thanks.

So in principle the default image package set should be depsolved together with the blueprint packages, as it is currently done in the `fedora33` distro. So the process would be:

1. merge image packages and blueprint packages into a single package list.
2. remove any packages explicitly marked as `Include` from the `Exclude` list before depsolving.
3. depsolve the package list.

So you should keep a modified variant of:

- https://github.com/osbuild/osbuild-composer/blob/37181eb995e5286d065e0e38321410150e498539/internal/distro/fedora33/distro.go#L68-L79
- https://github.com/osbuild/osbuild-composer/blob/37181eb995e5286d065e0e38321410150e498539/internal/distro/fedora33/distro.go#L209-L218

jordigilh:
> > Renamed package `fedora33` and `fedora33-test` to be `fedora` and `fedora-test` respectively. Tomorrow I'll take a look at the bug #2386 but if you can drop me any guidance that would help me to address it quickly smile
> 
> Thanks.
> 
> So in principle the default image package set should be depsolved together with the blueprint packages, as it is currently done in the `fedora33` distro. So the process would be:
> 
>     1. merge image packages and blueprint packages into a single package list.
> 
>     2. remove any packages explicitly marked as `Include` from the `Exclude` list before depsolving.
> 
>     3. depsolve the package list.
> 
> 
> So you should keep a modified variant of:
> 
>     * https://github.com/osbuild/osbuild-composer/blob/37181eb995e5286d065e0e38321410150e498539/internal/distro/fedora33/distro.go#L68-L79
> 
>     * https://github.com/osbuild/osbuild-composer/blob/37181eb995e5286d065e0e38321410150e498539/internal/distro/fedora33/distro.go#L209-L218

Done! I now realize I removed that part without a second thought. Sorry about that :disappointed:

jordigilh:
@thozza is it ok to rename 
https://github.com/osbuild/osbuild-composer/blob/55b78bbd36f70b29876a2a89286904348354e94b/internal/distro/fedora33/distro.go#L345
to `org.osbuild.fedora`?

thozza:
> @thozza is it ok to rename
> 
> https://github.com/osbuild/osbuild-composer/blob/55b78bbd36f70b29876a2a89286904348354e94b/internal/distro/fedora33/distro.go#L345
> 
> 
> to `org.osbuild.fedora`?

This is a name of the runner that `osbuild` should use to build the image. I think that it would make sense to rename it to `org.osbuild.fedora`, but it would have to be first defined in https://github.com/osbuild/osbuild/tree/main/runners.

@gicmo do you have anything against having a version-less fedora runner defined in osbuild?

achilleas-k:
> So in principle the default image package set should be depsolved together with the blueprint packages, as it is currently done in the `fedora33` distro. So the process would be:
> 
> 1. merge image packages and blueprint packages into a single package list.
> 2. remove any packages explicitly marked as `Include` from the `Exclude` list before depsolving.
> 3. depsolve the package list.

We had this solution for a while before and it had other issues: dependencies of a selected package wont get removed from the exclude list and will fail.  We have a regression test right now that will fail if we do it this way.  Do we want to go back to this before moving to the solution we discussed already (no excludes and no recommended packages)?

gicmo:
> @gicmo do you have anything against having a version-less fedora runner defined in osbuild?

I think I would keep them versioned, so that we still have the option to do something based on a specific distro version but solve this in osbuild where we fallback to the latest one.

thozza:
> We had this solution for a while before and it had other issues: dependencies of a selected package wont get removed from the exclude list and will fail. We have a regression test right now that will fail if we do it this way. Do we want to go back to this before moving to the solution we discussed already (no excludes and no weak dependencies)?

Hmm... Good point on the other issues 😕  I guess we should try the "no excludes and no weak dependencies", but I'm not sure what would be the best compromise to unblock this PR... 🤷

achilleas-k:
> > We had this solution for a while before and it had other issues: dependencies of a selected package wont get removed from the exclude list and will fail. We have a regression test right now that will fail if we do it this way. Do we want to go back to this before moving to the solution we discussed already (no excludes and no weak dependencies)?
> 
> Hmm... Good point on the other issues confused I guess we should try the "no excludes and no weak dependencies", but I'm not sure what would be the best compromise to unblock this PR... shrug

Yeah at the end of the day, for the current solutions, it's one or the other, each with its own downsides.  I wrote the "dual depsolve" one because I felt it was preferable to have some (perhaps unwanted) extra packages than to block blueprint packages, but maybe that has side effects I didn't consider.

jordigilh:
@achilleas-k @thozza @teg what is the status of this PR? Is it just pending @nullr0ute 's review? Is there anything else you'd like me to address?

internal/distro/fedora/distro.go
```diff
@@ -196,49 +211,73 @@ func (t *imageType) PartitionType() string {
...
 		// removePackage is fine if the package doesn't exist
 		excludedPackages = removePackage(excludedPackages, pkg)
 	}
+	mergedSets[osPkgsKey] = rpmmd.PackageSet{Include: mergedSets[osPkgsKey].Include, Exclude: excludedPackages}
 
-	return packages, excludedPackages
-}
+	// Add kernel package if none defined
+	kc := bp.Customizations.GetKernel()
+	mergedSets[osPkgsKey] = mergedSets[osPkgsKey].Append(rpmmd.PackageSet{Include: []string{kc.Name}})
```
thozza, line 278: I have to apologize, but after a discussion within our team, we agreed that the preferred approach for now is to use the same approach as used in RHEL-8.6/9.0 distro definitions. This means introducing #2386 , but we think that it is acceptable. So this means reverting my previous suggestion and depsolving the "os" and "blueprint" packages separately. Sorry for the noise 🙄

internal/distro/fedora/distro.go
```diff
@@ -196,49 +211,73 @@ func (t *imageType) PartitionType() string {
...
+
 	if t.bootable {
-		packages = append(packages, t.arch.bootloaderPackages...)
+		switch t.arch.Name() {
+		case distro.X86_64ArchName:
+			mergedSets[osPkgsKey] = mergedSets[osPkgsKey].Append(rpmmd.PackageSet{Include: t.arch.bootloaderPackages})
+		case distro.Aarch64ArchName:
+			mergedSets[osPkgsKey] = mergedSets[osPkgsKey].Append(rpmmd.PackageSet{Include: t.arch.bootloaderPackages})
+		}
+
```
thozza, line 251: Since you introduced the `packageSetFunc()` and separated package sets into a `package_sets.go`, I think that it makes sense to go all the way and adopt the RHEL-8.6/9.0 approach completely. Which would mean e.g  substituting this section with https://github.com/osbuild/osbuild-composer/blob/b2d18166de636613f27b42c55aa2679461e9928b/internal/distro/rhel90/package_sets.go#L193-L238

IMO this will also make it easier to rework the Fedora distro definition to manifests V2...

internal/distro/fedora/distro.go
```diff
@@ -706,295 +734,186 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
...
 		bootloaderPackages: []string{
-			"dracut-config-generic",
-			"efibootmgr",
-			"grub2-efi-aa64",
-			"grub2-tools",
-			"shim-aa64",
-		},
-		uefi: true,
+			"dracut-config-generic", "efibootmgr", "grub2-efi-aa64",
+			"grub2-tools", "shim-aa64"},
```
thozza, line 746: Please keep one package name on a line.

internal/distro/fedora/distro.go
```diff
@@ -16,63 +16,74 @@ import (
...
 	modulePlatformID string
 	ostreeRefTmpl    string
 	arches           map[string]architecture
-	buildPackages    []string
 }
 
 type architecture struct {
 	distro             *distribution
 	name               string
 	bootloaderPackages []string
```
thozza, line 58: This should be also removed in favor of using the `packageSetFunc()` and be moved to the `package_sets.go` as I commented in the `PackageSets()` method.

internal/distro/fedora/distro.go
```diff
@@ -706,295 +734,186 @@ func newDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
...
+			"grub2-tools", "shim-aa64"},
+		bootType: distro.UEFIBootType,
 	}
 
+	iotServices := []string{
+		"NetworkManager.service", "firewalld.service", "rngd.service", "sshd.service",
+		"zezere_ignition.timer", "zezere_ignition_banner.service",
+		"greenboot-grub2-set-counter", "greenboot-grub2-set-success", "greenboot-healthcheck", "greenboot-rpm-ostree-grub2-check-fallback",
+		"greenboot-status", "greenboot-task-runner", "redboot-auto-reboot", "redboot-task-runner",
+		"parsec", "dbus-parsec"}
```
thozza, line 755: Please use one service name on a line. This will make it easier to review change diffs in the future. Thanks.

tools/test-case-generators/format-request-map.json
```diff
@@ -171,7 +171,10 @@
         }
       }
     },
-    "overrides": {}
+    "overrides": {},
+    "supported_arches": [
+      "x86_64"
```
thozza, line 176: Out of curiosity, what is the reason for not testing the `fedroa-iot-commit` image type with kernel customization on `aarch64`?

thozza (commented):
Thanks again for your contribution and sorry for the waiting. I did a thorough review and added comments to section which should be IMO changed. I'll make sure to work with you on a timely manner to get this PR merged soon...

tools/test-case-generators/format-request-map.json
jordigilh: On fedora35 at least, there doesn't seem to be packages for `kernel-debug` or `kernel-rt` in aarch64:
```bash
$> dnf list kernel-rt
Last metadata expiration check: 0:19:52 ago on Wed 16 Mar 2022 11:48:01 AM EDT.
Error: No matching Packages to list
```
```bash
$> dnf list kernel-debug
Last metadata expiration check: 0:19:34 ago on Wed 16 Mar 2022 11:48:01 AM EDT.
Error: No matching Packages to list
```
There's only `kernel`:
```bash
$> dnf list kernel
Last metadata expiration check: 0:21:26 ago on Wed 16 Mar 2022 11:48:01 AM EDT.
Installed Packages
kernel.aarch64 
```

internal/distro/fedora/distro.go
jordigilh: Understood.

thozza:
> @thozza Since we also need `fedora-iot-container` and `fedora-iot-installer` as image types for aarch64, I'm working on migrating fedora to V2, using rhel-90 as the code base. I think it makes more sense to work on the migration rather than continue with V1 in Fedora. I'm considering closing this PR in favor of the migration I just mentioned. WDYT?

@jordigilh yeah, I think that this approach makes sense. From my PoV ideally reimplementing the current images using V2 first and then adding support for new image types. I think that the benefit will be that the work on additional image types will be much easier after the complete move to V2.

If you would need any help or are not sure about some parts, feel free to reach out 😉

tools/test-case-generators/format-request-map.json
thozza: Thanks for the explanation!

jordigilh:
Closing this PR in favor of a new one to migrate Fedora to osbuild V2.

runcom:
for reference, the new PR is https://github.com/osbuild/osbuild-composer/pull/2461

PR #2393, opened 2022-03-06
dbjobqueue: reduce the number of needed connections

ondrejbudai:
Previously, all dequeuers (goroutines waiting for a job to be dequeued) were
listening for new messages on postgres channel jobs (LISTEN jobs). This didn't
scale well as each dequeuer required to have its own DB connection and the
number of DB connections is hard-limited in the pool's config.

I changed the logic to work somewhat differently: dbjobqueue.New() now spawns
a goroutine that listens on the postgres channel. If there's a new message,
the goroutine just wakes up all dequeuers using a standard go channel.
Go channels are cheap so this should scale much better.

A test was added that confirms that 100 dequeuers are not a big deal now. This
test failed when I tried to run it on the previous commit. I tried even 1000
locally and it was still fine.

This pull request includes:

### DIFF
--- a/internal/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/internal/jobqueue/dbjobqueue/dbjobqueue.go
@@ -6,14 +6,15 @@
 package dbjobqueue
 
 import (
+	"container/list"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/jackc/pgconn"
 	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
@@ -124,7 +125,51 @@ const (
 )
 
 type DBJobQueue struct {
-	pool *pgxpool.Pool
+	pool         *pgxpool.Pool
+	dequeuers    *dequeuers
+	stopListener func()
+}
+
+// thread-safe list of dequeuers
+type dequeuers struct {
+	list  *list.List
+	mutex sync.Mutex
+}
+
+func newDequeuers() *dequeuers {
+	return &dequeuers{
+		list: list.New(),
+	}
+}
+
+func (d *dequeuers) pushBack(c chan struct{}) *list.Element {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	return d.list.PushBack(c)
+}
+
+func (d *dequeuers) remove(e *list.Element) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	d.list.Remove(e)
+}
+
+func (d *dequeuers) notifyAll() {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	cur := d.list.Front()
+	for cur != nil {
+		listenerChan := cur.Value.(chan struct{})
+
+		// notify in a non-blocking way
+		select {
+		case listenerChan <- struct{}{}:
+		default:
+		}
+		cur = cur.Next()
+	}
 }
 
 // Create a new DBJobQueue object for `url`.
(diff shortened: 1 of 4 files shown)


ondrejbudai:
This is now ready after I did some clean-ups. It will conflict with #2344 though. I would prefer to merge #2344 firstly as this PR will be simpler to rebase. :)

ondrejbudai:
Rebased on `main`, this can go in.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -193,44 +267,46 @@ func (q *DBJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []
 		return uuid.Nil, uuid.Nil, nil, "", nil, jobqueue.ErrDequeueTimeout
 	}
 
-	conn, err := q.pool.Acquire(ctx)
-	if err != nil {
-		return uuid.Nil, uuid.Nil, nil, "", nil, fmt.Errorf("error connecting to database: %v", err)
-	}
+	// add ourselves as a dequeuer
+	q.dequeuersLock.Lock()
+	c := make(chan struct{}, 1)
```
teg, line 272: You don't need to lock the call to `make()`. I think it might be easier to follow the code if you pull out
```
q.dequeuersLock.Lock()
defer q.dequeuersLock.Unlock()
return q.dequeuers.PushBack(c)
```
into a helper function. And the same with all the other places we hold `dequeuersLock`.

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -246,6 +322,21 @@ func (q *DBJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []
...
+// running the sqlDequeue query
+func (q *DBJobQueue) dequeueMaybe(ctx context.Context, token uuid.UUID, jobTypes []string, channels []string) (id uuid.UUID, jobType string, args json.RawMessage, err error) {
+	var conn *pgxpool.Conn
+	conn, err = q.pool.Acquire(ctx)
+	if err != nil {
+		return
+	}
+	defer conn.Release()
+
+	err = conn.QueryRow(ctx, sqlDequeue, token, jobTypes, channels).Scan(&id, &jobType, &args)
```
teg, line 336: Should we not handle the error?

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -134,10 +142,76 @@ func New(url string) (*DBJobQueue, error) {
...
+		stopListener: make(chan struct{}),
+	}
+
+	go q.listen()
+
+	return q, nil
+}
+
+func (q *DBJobQueue) listen() {
+	ctx := context.Background()
```
teg, line 157: If this was passed in from the caller, could we not avoid having to create `cancelContext` below?

teg (commented):
I very much like the overall idea.

When all the listeners are now woken up to call `dequeueMaybe()` can they exhaust the connection pool, or `pgxpool` do some locking under the hood to make sure we only make the a certain number of connections at once?

internal/jobqueue/dbjobqueue/dbjobqueue.go
ondrejbudai: Indeed, I created a helper struct, thanks, great suggestion.

internal/jobqueue/dbjobqueue/dbjobqueue.go
ondrejbudai: `err` is returned via named return values.

internal/jobqueue/dbjobqueue/dbjobqueue.go
ondrejbudai: Thanks for the suggestion, I changed the behavior to create the context in `New()`, pass it into `listen()` and store the cancel function in `DbJobQueue`. Simpler to follow with one less goroutine. :)

Gundersanne (approved):

internal/jobqueue/dbjobqueue/dbjobqueue.go
teg: fancy!

teg (approved):
Thanks!

PR #2398, opened 2022-03-07
templates/packer: Drop image_users variable

Gundersanne:
Declaration of who to share the image with are now defined in app-interface.

### DIFF
--- a/templates/packer/composer.pkr.hcl
+++ b/templates/packer/composer.pkr.hcl
@@ -37,9 +37,6 @@ source "amazon-ebs" "image_builder" {
   # Set a name for the resulting AMI.
   ami_name = "${var.image_name}"
 
-  # Share the resulting AMI with accounts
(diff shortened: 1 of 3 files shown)


ondrejbudai:
This basically means that we would lose access to these images on the 93 account, right? I'm not sure if that's a good idea though because we would basically lose access to our prod images if we need to ever debug them.

I also wanted to reuse the app-sre builder to build ourselves Fedora images but they will surely be run on 93 temporarily so I would have to invent a workaround if sharing is not possible.

Gundersanne:
> This basically means that we would lose access to these images on the 93 account, right? I'm not sure if that's a good idea though because we would basically lose access to our prod images if we need to ever debug them.
> 
> I also wanted to reuse the app-sre builder to build ourselves Fedora images but they will surely be run on 93 temporarily so I would have to invent a workaround if sharing is not possible.

Let's keep the sharing then.

Issue #1474, opened 2021-06-15
[CI] List all installed RPMs on the system running the test case


thozza:
Job logs in Gitlab CI list only NVRs of packages which were explicitly installed as part of the job, but there is no information about packages which were already installed on the image used for CI run.

I would like to suggest, that we run `rpm -qa` as part of each job and log the output. This should be done ideally after all setup steps were finished and before the particular test case is executed. Having this kind of information will make it easier to debug potential CI failures and properly reproduce the environment.

jrusz:
fixed in https://github.com/osbuild/osbuild-composer/pull/2403

PR #2403, opened 2022-03-08
ci: run ci_details.sh in before_script

jrusz:
This is a nice script showing potentially useful details about the
runner so let's execute it at the begining of each job.

This pull request includes:


thozza:
This would resolve #1474

ondrejbudai:
Let me play the devil's advocate: This is just more lines in our already hard-to-parse huge logs.

Artifact it, maybe?

jrusz:
> Let me play the devil's advocate: This is just more lines in our already hard-to-parse huge logs.
> 
> Artifact it, maybe?

Sure, in that case how about I also remove any `set -x` from tests? I think those are the main cause of huge logs.. Also the CI trigger that just got merged doesn't work, the condition is `false` even though I expect it to be `true` cause this is draft PR :/

jrusz:
I changed my mind, I'll just remove it. I'm trying to fix a problem which doesn't really exist, I just find it confusing that some tests have `set -x` and others don't like based on what do we decide? I'll put it as a topic for next Composer QE.

schutzbot/ci_details.sh
```diff
@@ -37,13 +37,6 @@ echo "List of installed packages:"
 rpm -qa | sort
 echo "------------------------------------------------------------------------------"
 
-if rpm --quiet -q python36; then
-    echo -e "\n FAIL: python36 is installed, see #794 ..."
-    exit 1
-else
-    echo -e "\n PASS: python36 not insalled"
-fi
```
atodorov, line 45: This check was targeting a specific issue and we don't want to have python36 (not sure if that applies to other versions) pre-installed into CI job runners.  This should be removed unless we know better.

atodorov (changes requested):

schutzbot/ci_details.sh
jrusz: Uhh I'm not sure I understand now. I removed it because it sounded obsolete, should I keep it that way or bring it back?

atodorov: Bring it back, b/c at this point in time we don't know if this snippet needs to be removed or is still necessary.

atodorov (approved):

PR #2412, opened 2022-03-09
Add tests for blueprints without explicit definition for /

atodorov:
Related: RHBZ#2049500

This pull request includes:


jabia99:
Just so I understand. Why create a new test? Wouldn't it make more sense to add it in `filesystem.sh`?

atodorov:
> Just so I understand. Why create a new test? Wouldn't it make more sense to add it in `filesystem.sh`?

The root cause of the problem is that `/` mount point wasn't specified explicitly. `filesystems.sh` already specifies `/` because it needs to check that you can customize it. Hence the need for a new test.

jabia99:
> The root cause of the problem is that `/` mount point wasn't specified explicitly. `filesystems.sh` already specifies `/` because it needs to check that you can customize it. Hence the need for a new test.

I think I didn't explain myself. I mean adding a new case inside `filesystem.sh`. Inside it we have a [success case](https://github.com/osbuild/osbuild-composer/blob/663f1dcbee8d30c1fbe8430f28c7d8a4221b57cb/test/cases/filesystem.sh#L133) and a [fail case](https://github.com/osbuild/osbuild-composer/blob/663f1dcbee8d30c1fbe8430f28c7d8a4221b57cb/test/cases/filesystem.sh#L201), I think we could add this test as a new case. In the end it's probably the same amount of lines, but it might make more sense here as it is related

atodorov:
@gicmo, @ondrejbudai this new test/blueprint seems to pass on 9.0 but fails on 8.6:

```
: time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././srv': No space left on device" jobId=b41489c7-9198->
: time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././tmp': No space left on device" jobId=b41489c7-9198->
: time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././var': No space left on device" jobId=b41489c7-9198->
: time="2022-03-09T14:28:17Z" level=info msg="    Traceback (most recent call last):" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
: time="2022-03-09T14:28:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.copy\", line 140, in <module>" jobId=b41489c7-9198-4df2-87ec-1d746>
: time="2022-03-09T14:28:17Z" level=info msg="        r = main(_args, _args[\"options\"])" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
: time="2022-03-09T14:28:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.copy\", line 133, in main" jobId=b41489c7-9198-4df2-87ec-1d746fa0b>
: time="2022-03-09T14:28:17Z" level=info msg="        check=True)" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
: time="2022-03-09T14:28:17Z" level=info msg="      File \"/usr/lib64/python3.6/subprocess.py\", line 438, in run" jobId=b41489c7-9198-4df2-87ec-1d746fa0b>
: time="2022-03-09T14:28:17Z" level=info msg="        output=stdout, stderr=stderr)" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
: time="2022-03-09T14:28:17Z" level=info msg="    subprocess.CalledProcessError: Command '['cp', '-a', '--reflink=auto', '/run/osbuild/inputs/root-tree/.'>
: time="2022-03-09T14:28:17Z" level=info msg="    " jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
```


Not sure what to do here. Can you review and tell me if the reproducer looks fine to you (I've copied the blueprint from the original customer support case with some modifications).  

Then tell me why there's a difference in behavior and if that requires more patches or exclusion of this BZ from 8.6 ?

gicmo:
> @gicmo, @ondrejbudai this new test/blueprint seems to pass on 9.0 but fails on 8.6:
> 
> ```
> : time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././srv': No space left on device" jobId=b41489c7-9198->
> : time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././tmp': No space left on device" jobId=b41489c7-9198->
> : time="2022-03-09T14:28:17Z" level=info msg="    cp: cannot create directory '/run/osbuild/mounts/././var': No space left on device" jobId=b41489c7-9198->
> : time="2022-03-09T14:28:17Z" level=info msg="    Traceback (most recent call last):" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
> : time="2022-03-09T14:28:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.copy\", line 140, in <module>" jobId=b41489c7-9198-4df2-87ec-1d746>
> : time="2022-03-09T14:28:17Z" level=info msg="        r = main(_args, _args[\"options\"])" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
> : time="2022-03-09T14:28:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.copy\", line 133, in main" jobId=b41489c7-9198-4df2-87ec-1d746fa0b>
> : time="2022-03-09T14:28:17Z" level=info msg="        check=True)" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
> : time="2022-03-09T14:28:17Z" level=info msg="      File \"/usr/lib64/python3.6/subprocess.py\", line 438, in run" jobId=b41489c7-9198-4df2-87ec-1d746fa0b>
> : time="2022-03-09T14:28:17Z" level=info msg="        output=stdout, stderr=stderr)" jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
> : time="2022-03-09T14:28:17Z" level=info msg="    subprocess.CalledProcessError: Command '['cp', '-a', '--reflink=auto', '/run/osbuild/inputs/root-tree/.'>
> : time="2022-03-09T14:28:17Z" level=info msg="    " jobId=b41489c7-9198-4df2-87ec-1d746fa0b616
> ```
> 
> Not sure what to do here. Can you review and tell me if the reproducer looks fine to you (I've copied the blueprint from the original customer support case with some modifications).
> 
> Then tell me why there's a difference in behavior and if that requires more patches or exclusion of this BZ from 8.6 ?

`No space left on device`. I wonder if it is the image itself or the underlying disc of the worker itself (seems likely due to the 8.6/9.0 difference). Do you have the manifest here?

atodorov:
https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2181893488

jrusz:
I wonder if some of the regression tests fail to clean up properly after themselves... I'd add `df -h` between test executions in  https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression.sh#L62 to rule out runner disk space problem.

atodorov:
Adding a `df -h` before and after each test reveals the following: 

1) The used space on disk slightly increases between each test. However the overall increase between the first test and the last one is around 1GB on a 40GB disk.

2) When the newly added regression test is executed we see the following:

on 9.0 the script passes:

```
***** DEBUG: before running regression-no-explicit-rootfs-definition.sh
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        1.9G     0  1.9G   0% /dev
tmpfs           1.9G  4.0K  1.9G   1% /dev/shm
tmpfs           768M   20M  748M   3% /run
/dev/nvme0n1p2   40G  4.8G   36G  12% /
tmpfs           384M     0  384M   0% /run/user/1000
```

on 8.6 it fails:
```
***** DEBUG: before running regression-no-explicit-rootfs-definition.sh
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        1.8G     0  1.8G   0% /dev
tmpfs           1.8G  4.0K  1.8G   1% /dev/shm
tmpfs           1.8G  8.5M  1.8G   1% /run
tmpfs           1.8G     0  1.8G   0% /sys/fs/cgroup
/dev/nvme0n1p2   40G  6.0G   35G  15% /
tmpfs           366M     0  366M   0% /run/user/1000
```


The difference in available space on disk between the 2 scenarios is only 1GB. 

As the next step I will disable all of the regression scripts and leave only the new one to see if there will be any difference.

atodorov:
> As the next step I will disable all of the regression scripts and leave only the new one to see if there will be any difference.

Still failed, see https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2191825874. I don't think the problem is with disk space.

jrusz:
> > As the next step I will disable all of the regression scripts and leave only the new one to see if there will be any difference.
> 
> Still failed, see https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2191825874. I don't think the problem is with disk space.

Yeah I guess not, thanks for checking. I once saw regression tests having out of space problems in gating so I was wondering what's the situation now. 

I took a look in the journal and it seems to me as if osbuild was trying to copy something to a tmpfs and it failed because it was too small https://paste.centos.org/view/c5d12187 ... I wonder if it's related to the runner or the image used, I'll try to run it on Openstack and latest RHEL-9.0 nightly to see if it fails the same.

jrusz:
It seems to fail in the same way, I uploaded the manifest @gicmo https://paste.centos.org/view/fea6389c

atodorov:
The failures aren't related to the newly added regression test.

jabia99 (dismissed):
LGTM. Just out of curiosity, what solved the error you were getting last time?

test/cases/regression-no-explicit-rootfs-definition.sh
```diff
@@ -0,0 +1,79 @@
...
+# https://bugzilla.redhat.com/show_bug.cgi?id=2049500
+
+# Get OS data.
+source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+
+# Provision the software under test.
+/usr/libexec/osbuild-composer-test/provision.sh
+
+case "${ID}-${VERSION_ID}" in
+    "rhel-8.6" | "rhel-9.0" | "centos-9")
```
jrusz, line 11: Why not run it on `centos-8` too? And we want to only run them on those specific RHEL versions?

jrusz (changes requested):
I have just one questions otherwise looks good.

test/cases/regression-no-explicit-rootfs-definition.sh
atodorov: I think this was a copy & paste error. I've removed the entire case block.

jrusz (approved):
LGTM

jrusz (approved):
Just a rebase.

jrusz:
Regression tests on fedora are failing and the logs are longer than Gitlab is able to store so we don't know what exactly is wrong. I suggest we merge https://github.com/osbuild/osbuild-composer/pull/2498 first and the adjust this PR accordingly.

jrusz (changes requested):
Took another look and I suppose that since this bug was reported on RHEL-8.5 and was fixed in RHEL-8.6 and not backported it's expected to fail there, but I'm still not sure about fedora. @gicmo Do you know if this is supposed to be supported on Fedora? The test says it's not.

atodorov:
> Took another look and I suppose that since this bug was reported on RHEL-8.5 and was fixed in RHEL-8.6 and not backported it's expected to fail there, 

Hmm. It can be expected the test to fail on 8.5 if you are running against the nightly builds but we do compile the newest RPMs from source on the various platforms. I don't think this test should fail for `main`.

jrusz:
> > Took another look and I suppose that since this bug was reported on RHEL-8.5 and was fixed in RHEL-8.6 and not backported it's expected to fail there,
> 
> Hmm. It can be expected the test to fail on 8.5 if you are running against the nightly builds but we do compile the newest RPMs from source on the various platforms. I don't think this test should fail for `main`.

Aaah, right, well then idk what's the problem :/

gicmo:
8.5 does not support `/data`: https://github.com/osbuild/osbuild-composer/blob/e30fa53c5042b1c8f35d6d4d32c7f9dd5cf06881/internal/distro/rhel85/distro.go#L45

achilleas-k:
What's the status here?  Commit message says it's disabled on Fedora for now, but it's not and it's failing.
Are we blocking this until the Fedora rewrite is merged?

jrusz (changes requested):
Fedora not skipped properly. Suggested a fix

gicmo:
For 8.5, @jrusz extracted the manifest (thanks) and the root partition is way too small:

```
      {
        "type": "org.osbuild.mkfs.xfs",
        "options": {
          "uuid": "3d710454-a834-4458-b23c-946f23310f5a",
          "label": "root"
        },
        "devices": {
          "device": {
            "type": "org.osbuild.loopback",
            "options": {
              "filename": "disk.img",
              "start": 4403200,
              "size": 1948,
              "lock": true
            }
          }
        }
      },
```

This is because the fix was applied to 8.6/9.0 but not 8.5. I will back port the fix.

atodorov:
I've cherry-picked https://github.com/osbuild/osbuild-composer/pull/2540 for rhel 8.5.

jrusz (approved):
LGTM

jrusz:
I don't know what's wrong with the `generic_s3.sh` test, it says something about wrong credentials https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2318096799 .. @ygalblum do you perhaps know what's wrong?

I'm also moving them to aws runners because they don't need kvm and openstack resources are limited https://github.com/osbuild/osbuild-composer/pull/2542

PR #2418, opened 2022-03-09
`rhel{86,90}`: set default grub boot entry to `saved`

gicmo:
When we set the `saved_entry` to a kernel, we also need to set `GRUB_DEFAULT=saved` so that the saved entry mechanism survives grub2 config regeneration by the user. Do so for by changing the logic in the stage options constructor and reset the corresponding configuration for frozen distros that use `NewGrub2StageOptions`, i.e. 8.5 and 9.0 beta.

TODO:
 - [x] manifest regeneration
 - [x] exception


thozza:
rebased on top of `main`, will regenerate the image tests...

ondrejbudai (approved):

PR #2421, opened 2022-03-10
Add shfmt to automatically format our shell scripts

ondrejbudai:
I'm tired of our randomly formatted bash scripts. Happy to start a flame war about tabs vs. spaces.

This pull request includes:


thozza:
So yeah, I would prefer to keep using spaces instead of tabs 😄

ondrejbudai:
I was thinking that Go uses tabs so it would be nice just use tabs as much as possible in this project. Also, it's the default. :sweat_smile:

achilleas-k:
:fire: tabs in shell scripts feels wrong
I can get used to it if I have to.  What does everyone else think?

jrusz:
My `vim` config taught me to use 4 spaces so I vote for 4 spaces :smile: 
```
set tabstop=4                                                                                                                                                                                                                                  
set softtabstop=0                                                                                                                                                                                                                              
set shiftwidth=4                                                                                                                                                                                                                               
set expandtab  
```

Gundersanne:
I like spaces :] :fuelpump: :arrow_right: :fire:

bcl:
Consistent formatting is good, I like go fmt :) But I also prefer spaces (4 please, not 2, not 3...) to tabs unless forced to do it by the language.

And I think that whatever is decided it needs to be automated and enforced on the server, otherwise you end up with commits like this that touch a bunch of lines and make it harder to track the history of changes.

achilleas-k (approved):
Hooray.  Spaces!

PR #2423, opened 2022-03-10
packer: Build Fedora 35 x86_64 and aarch64 images

ondrejbudai:
This PR adds building of Fedora 35 x86_64 and aarch64 images from our Packer configuration. I had to do some shuffling of code in the packer config and I think that the result is reasonably understandable.

Regarding who builds what:

RHEL 8 images from `main` are built on Jenkins.
Other RHEL 8 images are built on Schutzbot.
All Fedora images are built on Schutzbot.

The last commit adds a very useful feature: The images built by packer are not saved on PRs which should save us a ton of money in the long run.

This pull request includes:


teg:
Any chance we could use jenkins on fedora too?

For PRs: are we actually building/testing the worker AMIs?

ondrejbudai:
> Any chance we could use jenkins on fedora too?
Yes but me and Sanne decided to not interfere with the onboarding process in any way, therefore we do it on our infra for now.

> For PRs: are we actually building/testing the worker AMIs?
Building: yes. But this PR changes the behavior a bit: now they won't be saved after `packer` is done because this ate a lot of money and it just didn't make much sense. We can always re-enable.

Testing:
No. The biggest issue is that we need composer for that. I can imagine two strategies:
1) Run an ephemeral composer in some openshift cluster and connect workers spun up from the AMIs to it and push one build per a worker.
2) Connect the workers to staging aoc composer. This would require some additions to our multi-tenancy implementation though.

Both has its pros and cons.

Gundersanne:
Regarding point 2; while older workers against a newer composer is tested to some extent in the regression test; that's not the case for a newer worker against an older composer. 1 would be the right way imo, but requires more infra, we might able to use the frontdoor project in centosci however.

Gundersanne (approved):

PR #2427, opened 2022-03-11
Pass repo name to `dnf-json` and use it in `dnf-json`

thozza:
- dnf-json: do not use reponame as repoid
- rpmmd: pass repo name to dnf-json
- dnf-json: use repository name from the request if provided
- rpmmd: rename toDNFRepoConfig() argument i -> repoID
- osbuild-pipeline: use repo name from the request if provided

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -18,6 +18,7 @@ import (
 )
 
 type repository struct {
+	Name       string `json:"name,omitempty"`
 	BaseURL    string `json:"baseurl,omitempty"`
 	Metalink   string `json:"metalink,omitempty"`
 	MirrorList string `json:"mirrorlist,omitempty"`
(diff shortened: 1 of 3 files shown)


teg (approved):
This looks great to me!

Just to double check that I understood the overall intention: we now include the proper repo name for UI/logging purposes. However, internally we still use `repoid`, so the name is just an opaque value that does not need to be unique or on any particular form etc?

thozza:
> This looks great to me!
> 
> Just to double check that I understood the overall intention: we now include the proper repo name for UI/logging purposes. However, internally we still use `repoid`, so the name is just an opaque value that does not need to be unique or on any particular form etc?

Correct.

The `repoid` is still always a unique number and it is what gets passed to `dnf-json` for any repo. The reply from `dnf-json` contains a list of packages with information from which repo they come from, specified by `repoid`. In the past we were not passing `reponame` to `dnf-json`, only the `repoid`. If the `reponame` is not set, it defaults to `repoid` in DNF.

PR #2428, opened 2022-03-11
dnf-json: use the default connection timeout

teg:
By default `timeout` is 30 seconds, but we had it set to 5. Drop
the override and use the default.

This has two effects: it increases the time before we give up on
connecting (as it says on the tin), and it also increases the time
download has to be slow for before we give up.

Internally, we were seing failures in downlaoding metadata from ODCS
and similar issues have occurred in CI too.

The potential downside to this is in case of having several mirrors
this means it takes longer before giving up on a bad one and trying
a better one. But slow is better than broken, so for now rever to
the default behavior.

This pull request includes:

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -118,9 +118,6 @@ class Solver():
         # building images, I don't care if we download even more.
         self.base.conf.zchunk = False
 
-        # Try another mirror if it takes longer than 5 seconds to connect.
-        self.base.conf.timeout = 5
-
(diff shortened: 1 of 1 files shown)


Gundersanne (approved):

teg:
See also https://github.com/osbuild/osbuild/pull/990

ondrejbudai (approved):

PR #2432, opened 2022-03-14
osbuild2: honor GPG key setting for rpm inputs

thozza:
We should honour `pkg.CheckGPG` when creating the file inputs for the
rpm stage. This was lost in the transition from v1 to v2 manifests.

Regenerate image test manifests.

Co-authored-by: Tomas Hozza <thozza@redhat.com>


achilleas-k (dismissed):
LGTM

ondrejbudai:
The last force-push was to unblock GitLab, it didn't actually make any changes to the code, I basically just wanted a new sha.

thozza:
The RPM signature verification fails on 8.5 Cloud API tests, which uses CDN repos. I reproduced this, but don't have any idea yet why it behaves this way. The "funny" thing is that the integration (Weldr API) test case for 8.5 passes just fine and in principle it should do the same thing... 🤔

ondrejbudai (approved):
Awesome, thanks a lot for this.

PR #2433, opened 2022-03-14
Client creds grant worker client

Gundersanne:


internal/worker/client.go
```diff
@@ -123,7 +119,10 @@ func NewClientUnix(conf ClientConfig) *Client {
 }
 
 // Note: Only call this function with Client.tokenMu locked!
-func (c *Client) refreshBearerToken() error {
+func (c *Client) refreshAccessToken() error {
+	c.tokenMu.Lock()
```
ondrejbudai, line 128: I think you can't get a write lock when you already have a read lock, this is basically a deadlock, see https://pkg.go.dev/sync#RWMutex

I believe that this is a reason why the multi-tenancy test fails.

Gundersanne: token() immediately releases the reader lock though

Gundersanne: Ah, no client_id in the config:

```
func (c *Client) refreshAccessToken() error {
        ...
	if c.clientId == "" {
		return fmt.Errorf("No client id specified")
	}
```

Should I make it fail on configure when there's no client_id rather than in the refresh?

ondrejbudai: I think we should, it makes sense to catch this ASAP.

ondrejbudai (approved):

PR #2437, opened 2022-03-14
osbuild2/rpm: deduplicate `rpmStageOptions()` and do not import GPG key if repo `gpgcheck` is disabled

thozza:
- Move `rpmStageOptions()` from `stage_options.go` to `osbuild2`
- osbuild2/rpm: do not import GPG keys if repo gpgcheck is disabled

### DIFF
--- a/internal/distro/rhel85/pipelines.go
+++ b/internal/distro/rhel85/pipelines.go
@@ -183,7 +183,7 @@ func ec2BaseTreePipeline(repos []rpmmd.RepoConfig, packages []rpmmd.PackageSpec,
 	p.Name = "os"
 	p.Build = "name:build"
 	packages = append(packages, bpPackages...)
-	p.AddStage(osbuild.NewRPMStage(rpmStageOptions(repos), osbuild.NewRpmStageSourceFilesInputs(packages)))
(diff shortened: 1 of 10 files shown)


teg (approved):

thozza:
I should have marked this as a draft. I still need to regenerate image test cases, because we are not consistent in setting the `gpg_check` in the repos that we use to generate them. As a result, some GPG keys will be deleted from the image test manifests.

In addition, the second commit changes the behavior only for V2 based distro definitions. Which means that Fedora, RHEL-84 and RHEL-83 keep the behavior of always importing the GPG key. I'm in doubt whether these should be changed as well. And I'm also I'm not sure if the change is actually worth it, since one could as well not include the GPG key in the repo definition with the same effect. I initially thought that it makes sense to be consistent with DNF behavior, but I'm now not sure if it is work the behavior change. What's your opinion @gicmo, @teg ?

Note that I still think that deduplicating `rpmStageOptions()` makes sense...

thozza:
I decided to not change the default behavior related to importing GPG keys provided with the repository.

In situation when it is needed to not import the GPG key, one can also not include it in the repo definition, which has the same effect as if the GPG signature check was disabled (the GPG key is not imported).

I'm happy to revisit this in the future, but for now it is IMO not worth it. I submitted the de-duplication change as #2503

PR #2443, opened 2022-03-15
test: add image types test

jabia99:
test: related to bz 2018419. We need a test to make sure every OS+arch
variant has the expected image types.

### DIFF
--- a/cmd/osbuild-composer-image-types-tests/main_test.go
+++ b/cmd/osbuild-composer-image-types-tests/main_test.go
@@ -1,0 +1,21 @@
+package main
+
+import (
+    "fmt"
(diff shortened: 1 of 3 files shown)


jabia99:
> @jabia99 please check #2350. IMO that would be the preferred way, compared to calling `composer-cli` from Go binary. We can actually verify this on a single system of any architecture for all distributions and architectures. Feel free to adapt the my PR. I put it on hold because of other work, but think that we need to test this.

We completely missed this in QE team. Thanks for letting me know!! I don't think it's that urgent right now. If you don't have time to work on it, please ping me and I'll try to finish it.

PR #2445, opened 2022-03-15
prometheus: measure pending rather than queued time

teg:
Measure the time a job has been pending on the queue (ready to be dequeued), rather than
how long it has been on the queue altogether. This will give us a measure for when we need
more workers.

Also record job types as `osbuild` rather than `osbuild:x86_64`.

### DIFF
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -171,7 +171,7 @@ func (s *Server) OSBuildJobStatus(id uuid.UUID, result *OSBuildJobResult) (*JobS
 		return nil, nil, err
 	}
 
-	if !strings.HasPrefix(jobType, "osbuild:") { // Build jobs get automatic arch suffix: Check prefix
+	if jobType != "osbuild" { // Build jobs get automatic arch suffix
 		return nil, nil, fmt.Errorf("expected osbuild:*, found %q job instead", jobType)
 	}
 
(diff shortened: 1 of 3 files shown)


internal/worker/server.go
```diff
@@ -442,8 +442,12 @@ func (s *Server) requestJob(ctx context.Context, arch string, jobTypes []string,
 	if err != nil {
 		logrus.Errorf("error retrieving job status: %v", err)
 		return
-	} else {
-		prometheus.DequeueJobMetrics(status.Queued, status.Started, jobType)
+	}
+
+	if jobType == "osbuild:"+arch {
+		jobType = "osbuild"
```
kingsleyzissou, line 448: This makes sense, we should probably do the same in other places for consistency otherwise we get this:

image

teg: oops. will do

teg: Done!

Gundersanne (approved):
Just as q ; would we ever add architecture as a separate label then?

teg:
> Just as q ; would we ever add architecture as a separate label then?

I think we should :)

ondrejbudai (dismissed):

Gundersanne:
Something was up with CI last night btw :/ I think the easiest way to restart the hting is just amend/no-edit & repush.

Gundersanne (approved):

kingsleyzissou (approved):

kingsleyzissou (dismissed):

Gundersanne (changes requested):
All prometheus and grafana queries look for osbuild:.* jobs; so this would break most of our dashboards and alerts because of the `:`.

Either we need to update (and release) all the queries first or stock with the +:arch template.

Gundersanne:
I split out the queue time fix in https://github.com/osbuild/osbuild-composer/pull/2649, I do think splitting out the jobtype architecture to a label in the metrics  is a good idea, but for that the queries need to change first.

teg:
I'm unlikely to have time to finish this up, feel free to take it over / consider it an RFE, if anyone is interested to finish up the remaining piece. I'll close this for now though, as it has mostly landed.

PR #2447, opened 2022-03-16
container: graceful shutdown

teg:
Support graceful shutdown of composer, making sure that any pending request are handled before the container is
terminated.

In the past we risked OpenShift thinking the container was dead (because we stopped answering to liveness probes) and
killing it before pending requests (in particular long-running depsolves) had been processed.

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
@@ -10,11 +11,12 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"path"
+	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-
 	logrus "github.com/sirupsen/logrus"
 
 	"github.com/osbuild/osbuild-composer/internal/auth"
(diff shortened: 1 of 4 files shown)


cmd/osbuild-composer/composer.go
```diff
@@ -307,14 +311,49 @@ func (c *Composer) Start() error {
...
 
-	// wait indefinitely
-	select {}
+	sigint := make(chan os.Signal, 1)
+
+	signal.Notify(sigint, syscall.SIGTERM)
+	signal.Notify(sigint, syscall.SIGINT)
+
+	// block until interrupted
+	<-sigint
```
ondrejbudai, line 326: Would like to see a small log message after the signal is received.

cmd/osbuild-composer/composer.go
```diff
@@ -214,91 +216,93 @@ func (c *Composer) Start() error {
...
+	}
+
+	composerAPI := &http.Server{
+		ErrorLog: c.logger,
+		Handler:  composerHandler,
+	}
+
 	if c.localWorkerListener != nil {
 		go func() {
-			s := &http.Server{
```
ondrejbudai, line 219: I think I preferred to have the listener-specific bits in the conditions. It might be slightly cleaner to define all `http.Server`s outside the conditions, initialize them in these conditions and call `Serve` in goroutines. WDYT?

cmd/osbuild-composer/composer.go
```diff
@@ -307,14 +311,49 @@ func (c *Composer) Start() error {
...
+			panic(err)
+		}
+	}
+
+	if c.weldrListener != nil {
+		err = c.weldr.Shutdown(context.Background())
+		if err != nil {
+			panic(err)
+		}
+	}
```
ondrejbudai, line 356: It would be nice if we can also `Shutdown()` the `jobqueue` here. (only `dbjobqueue` implements `Shutdown` so we might want to extend this also for `fsjobqueue` as a dummy function).

ondrejbudai (approved):
Loving this!

cmd/osbuild-composer/composer.go
teg: Hmm, what do you have in mind? We should shut down our external APIs, but not internal things, no?

cmd/osbuild-composer/composer.go
teg: This was in the wrong commit. Fixed.

cmd/osbuild-composer/composer.go
teg: Agreed. This should also make it easier to review. Fixed.

cmd/osbuild-composer/composer.go
teg: I did not change this.

ondrejbudai: While I agree that this is an improvement, I think it would be nicer to shutdown all goroutines in a defined way before we exit the whole process. We can surely do it in another PR thouugh.

ondrejbudai (commented):
Nice, I think we need https://github.com/osbuild/osbuild-composer/pull/2464 firstly though. :/

cmd/osbuild-composer/composer.go
teg: Ah, yes. We should, but need to be careful about deadlocks for that (can't do it at the same time as the others). Let's do that in a follow up

ondrejbudai: Makes sense

ondrejbudai (approved):

Issue #938, opened 2020-08-22
worker: remove hack that returns vmdk's as streamoptimized


larskarlitski:
Due to time constraints, we introduced a hack which converts vmdk images to streamoptimized with https://github.com/osbuild/osbuild-composer/pull/934.

osbuild should be able to set this option in the `qemu` assembler, which is more efficient and the correct place to do such things. Once it does, remove this hack.

ondrejbudai:
Hehe, we did this probably years ago. :)

PR #2459, opened 2022-03-18
Add `cloud-init` to VMDK image and test it in VSphere

thozza:
- Add `cloud-init` to the default package set of VMDK image on RHEL-85/86/90
- Extend the `api.sh` test case to test VMDK image uploaded to S3 in VSphere and run it in the CI
- Cloud API: upload stream-optimized VMDK to S3

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -315,7 +315,35 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			if impl.AWSBucket != "" {
 				bucket = impl.AWSBucket
 			}
-			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), bucket, key)
+
+			imagePath := path.Join(outputDirectory, exportPath, options.Filename)
+
+			// *** SPECIAL VMDK HANDLING START ***
+			// Upload the VMDK image as stream-optimized.
+			// The VMDK conversion is applied only when the job was submitted by Weldr API,
+			// therefore we need to do the conversion here explicitly if it was not done.
+			if args.StreamOptimized {
+				// If the streamOptimizedPath is empty, the conversion was not done
+				if streamOptimizedPath == "" {
+					var f *os.File
+					f, err = vmware.OpenAsStreamOptimizedVmdk(imagePath)
+					if err != nil {
+						osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidConfig, err.Error())
+						return nil
+					}
+					streamOptimizedPath = f.Name()
+					f.Close()
+				}
+				// Replace the original file by the stream-optimized one
+				err = os.Rename(streamOptimizedPath, imagePath)
+				if err != nil {
+					osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidConfig, err.Error())
+					return nil
+				}
+			}
+			// *** SPECIAL VMDK HANDLING END ***
+
+			_, err = a.Upload(imagePath, bucket, key)
 			if err != nil {
 				osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorUploadingImage, err.Error())
 				return nil
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -209,6 +209,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		if err != nil {
 			return err
 		}
+		f.Close()
 	}
 
 	if len(args.Targets) == 0 {
--- a/internal/distro/rhel85/package_sets.go
+++ b/internal/distro/rhel85/package_sets.go
@@ -304,11 +304,17 @@ func vhdCommonPackageSet(t *imageType) rpmmd.PackageSet {
 func vmdkCommonPackageSet(t *imageType) rpmmd.PackageSet {
 	return rpmmd.PackageSet{
 		Include: []string{
-			"@core", "chrony", "firewalld", "langpacks-en", "open-vm-tools",
+			"@core",
+			"chrony",
+			"cloud-init",
+			"firewalld",
+			"langpacks-en",
+			"open-vm-tools",
 			"selinux-policy-targeted",
 		},
 		Exclude: []string{
-			"dracut-config-rescue", "rng-tools",
+			"dracut-config-rescue",
+			"rng-tools",
 		},
 	}.Append(bootPackageSet(t))
 
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -54,7 +54,8 @@ else
 fi
 
 # Start the db
-sudo ${CONTAINER_RUNTIME} run -d --name osbuild-composer-db \
+DB_CONTAINER_NAME="osbuild-composer-db"
+sudo ${CONTAINER_RUNTIME} run -d --name "${DB_CONTAINER_NAME}" \
     --health-cmd "pg_isready -U postgres -d osbuildcomposer" --health-interval 2s \
     --health-timeout 2s --health-retries 10 \
     -e POSTGRES_USER=postgres \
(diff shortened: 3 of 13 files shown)


internal/distro/rhel85/package_sets.go
```diff
@@ -304,11 +304,17 @@ func vhdCommonPackageSet(t *imageType) rpmmd.PackageSet {
 func vmdkCommonPackageSet(t *imageType) rpmmd.PackageSet {
 	return rpmmd.PackageSet{
 		Include: []string{
-			"@core", "chrony", "firewalld", "langpacks-en", "open-vm-tools",
+			"@core",
+			"chrony",
+			"cloud-init",
```
ondrejbudai, line 309: Quoting @teg:

> The guarantee we give is that existing behavior does not change for a major release, so all new features are behind new config options. We guarantee this by verifying in CI that manifest generation is stable. By default all RHEL minor versions behave the same, but we have the possibility of special casing when needed. We need a process for exception to the above rule.

I think that this change violates the guarantee therefore I think we shall summon the Image Builder Exception High Council for the first time?

thozza: Sure 😉 

As a side node, this change was explicitly requested by the one and only @teg 😄

gicmo: These are all guidelines where exceptions can be made in special circumstances (-:

One possible other option to solve this that would _not_ violate our guarantee is to have the service add `cloud-init` via the customisation. On-prem you could add a user and in the service where this is not possible you would get `cloud-init`. Downside of course: different behaviour between our products.
Now I wonder, if we asked users of the VMDK, how are they using it, right now?

thozza: > Now I wonder, if we asked users of the VMDK, how are they using it, right now?

I did not, not even sure whom to ask 🤔 

To be honest, I found the VMDK image to be quite strange. It can not be imported to VSphere as is, it has to be converted to `StreamOptimized` using `qemu-img`, before it can be imported. Composer does that before uploading, but if you download the image using `composer-cli` or from the S3, you have to do it yourself. I don't know what are the reasons to not produce the image as `StreamOptimized` as the definition predates me.

test/cases/api.sh
```diff
@@ -1064,6 +1231,109 @@ function verifyDisk() {
...
+        -folder="${GOVMOMI_FOLDER}" \
+        -net="${GOVMOMI_NETWORK}" \
+        -net.adapter=vmxnet3 \
+        -m=4096 -c=2 -g=rhel8_64Guest -on=true -firmware=bios \
+        -disk="${VSPHERE_VM_NAME}/${VSPHERE_IMAGE_NAME}" \
+        -disk.controller=ide \
+        -on=false \
+        "${VSPHERE_VM_NAME}"
+
+    # uplod ISO, create CDROM device and insert the ISO in it
```
achilleas-k, line 1291: Found a typo :grin:

achilleas-k (approved):
LGTM!

internal/distro/rhel85/package_sets.go
achilleas-k: > I don't know what are the reasons to not produce the image as `StreamOptimized` as the definition predates me.

You have the opposite problem with local deployments if it's `StreamOptimized`.  It fails to boot in the local VMWare hypervisors.
Maybe we could have separate image types for local vs cloud vmdks?  Not sure I love the idea because of image-type-number inflation, but I guess it's similar to the EC2 case.  You're getting the same (or similar) image in a different format for different environments.

thozza: > You have the opposite problem with local deployments if it's `StreamOptimized`. It fails to boot in the local VMWare hypervisors.

🤦  That's exactly what I expected... Why am I not surprised? (don't answer)

test/cases/api.sh
thozza: fixed

thozza:
Fixed a typo and rebased on top of `main` to incorporate changes done to `api.sh`.

internal/distro/rhel85/package_sets.go
achilleas-k: For the on-prem situation, I suppose the current behaviour makes sense: if you're getting the vmdk out of composer directly, you get the _plain_ version and you can do whatever you need with it.  If we know it's being pushed to a VMWare server, we convert it for the user (though it seems [the current behaviour was a quick workaround](https://github.com/osbuild/osbuild-composer/issues/938) and we've been meaning to [move this conversion to osbuild](https://github.com/osbuild/osbuild/issues/528)).

For the service, I chose to keep it as not-optimised since we're not explicitly pushing it to a server, but maybe we need some consistency here, or at least a way to communicate the format to users.

achilleas-k:
> Fixed a typo and rebased on top of `main` to incorporate changes done to `api.sh`.

https://github.com/osbuild/osbuild-composer/pull/2459/commits/6a815d33b0b198dda3e26b947d87cda7779132bb
Forgot to squash? :)

thozza:
> [6a815d3](https://github.com/osbuild/osbuild-composer/commit/6a815d33b0b198dda3e26b947d87cda7779132bb) Forgot to squash? :)

Hopefully got it finally right 😄

internal/distro/rhel85/package_sets.go
thozza: I think that the current bahvior makes sense. We don't know how users will use the image uploaded to S3. But I agree that it would be nice for the user to have a way to specify which variant of the image should be produced.

WRT the compressed EC2 images, I have been thinking in the past, that we could have an option to basically compress any image type, if it is not directly imported to a cloud, but to e.g. S3 or just built on premise. We could have a format options specific to image-types, e.g. for the VMDK.

If you push this idea too far, we could have e.g. an `ostree-commit` image type and then just an option to have it "packaged" as a tarball, or in a container, or an installer... 😉  We could do all kinds of crazy stuff...

thozza:
After a discussion with @teg, we agreed that the VMDK uploaded to AWS S3 via Cloud API should be stream-optimized so that it is ready for import to VSphere.

Unfortunately, the current `qemu` osbuild stage implementation does not support passing image format-specific options. Therefore I decided to use the same approach used when importing VMDK to VSphere via Weldr API also for Cloud API. Eventually, we should always produce stream-optimized VMDK and either convert it to not-stream optimized when being downloaded via Weldr API, or not at all.

test/cases/api.sh
```diff
@@ -1039,6 +1206,109 @@ function verifyDisk() {
...
+    local _ci_iso_path
+    _ci_iso_path="$(_createCIUserdataISO "${_ci_userdata_path}" "${_ci_metadata_path}")"
+
+    VSPHERE_IMAGE_NAME="${VSPHERE_VM_NAME}.vmdk"
+
+    # import the built VMDK image to VSphere
+    # import.vmdk seems to be creating the provided directory and
+    # if one with this name exists, it appends "_<number>" to the name
+    greenprint "🚧 Converting the downloaded VMDK image to be streamOptimized"
+    qemu-img convert -O vmdk -o subformat=streamOptimized "${_filename}" "${WORKDIR}/${VSPHERE_IMAGE_NAME}"
```
teg: With your last comment, is this still needed? Or did you not rebase yet?

thozza: I added one new commit which implemented this change https://github.com/osbuild/osbuild-composer/pull/2459/commits/5c46e0c9fe7e49407eb07b14c7f7c5a3e7193586. Mainly because it seemed cleaner to me.

thozza:
I didn't get the last change in cloud api right. I'll have to fix it first so marking as Draft in the meantime...

test/cases/api.sh
teg: Ok! I'll re-review the PR as a whole once not in draft :)

teg (approved):
Thanks for this Tomas!

internal/distro/rhel85/package_sets.go
teg: This was discussed a lot more in side channels and this was the conclusion:
- the platform we are targeting is vSphere, other VMWare products are out of scope (we should rename the image type in the cloud api in a follow up)
- enabling cloud-init appears to be what's expected on this platform and doesn't seem to have any drawbacks (apart from potential subtle changes in behavior)
- this PR does not change the behavior of the weldr API with respect to stream optimization
- images delivered through the cloud api are intended for vSphere and as such should be stream optimized, as this PR ensures

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -314,7 +315,35 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			if impl.AWSBucket != "" {
 				bucket = impl.AWSBucket
 			}
-			_, err = a.Upload(path.Join(outputDirectory, exportPath, options.Filename), bucket, key)
+
+			imagePath := path.Join(outputDirectory, exportPath, options.Filename)
+
+			// *** SPECIAL VMDK HANDLING START ***
+			// Upload the VMDK image as stream-optimized.
+			// The VMDK conversion is applied only when the job was submitted by Weldr API,
```
ondrejbudai, line 323: Note that this comment isn't correct, `args.StreamOptimized` is now also passed by the cloud API.

thozza: It is, but the image is converted only if the `args.ImageName` is not empty and only Weldr API is setting it... So the comment is actually correct, but the reason why it is so is not obvious... 😕 
https://github.com/osbuild/osbuild-composer/blob/c9a08da29fdb06f1bdabf4582d9f08ed6398cc97/cmd/osbuild-worker/jobimpl-osbuild.go#L193-L201

ondrejbudai: Ah, I misread the comment, sorry! Of course, it's correct. :)

PR #2460, opened 2022-03-19
job queue: reschedule interrupted jobs

teg:
We are currently losing jobs if workers are restarted while building. Let's restart each job up to two times before giving up. We need to have a limit in place in case a bug in the worker causes a certain job to always fail.

In the future we could teach the workers to notify us on graceful shutdown and handle that case better, so we don't have to rely on heartbeats. In this case, the limit on the number of retries could be higher, as we know it is not a bug on our side, but just normal operation of the infrastructure.

### DIFF
--- a/internal/jobqueue/jobqueuetest/jobqueuetest.go
+++ b/internal/jobqueue/jobqueuetest/jobqueuetest.go
@@ -33,6 +33,8 @@ func TestJobQueue(t *testing.T, makeJobQueue MakeJobQueue) {
 	t.Run("errors", wrap(testErrors))
 	t.Run("args", wrap(testArgs))
 	t.Run("cancel", wrap(testCancel))
+	t.Run("requeue", wrap(testRequeue))
+	t.Run("requeue-limit", wrap(testRequeueLimit))
 	t.Run("job-types", wrap(testJobTypes))
 	t.Run("dependencies", wrap(testDependencies))
 	t.Run("multiple-workers", wrap(testMultipleWorkers))
@@ -61,7 +63,7 @@ func finishNextTestJob(t *testing.T, q jobqueue.JobQueue, jobType string, result
 	require.Equal(t, jobType, typ)
 	require.NotNil(t, args)
 
-	err = q.FinishJob(id, result)
+	err = q.RequeueOrFinishJob(id, 0, result)
 	require.NoError(t, err)
 
 	return id
@@ -88,7 +90,7 @@ func testErrors(t *testing.T, q jobqueue.JobQueue) {
 	require.NoError(t, err)
 	require.Equal(t, id, idFromT)
 
-	err = q.FinishJob(id, nil)
+	err = q.RequeueOrFinishJob(id, 0, nil)
 	require.NoError(t, err)
 
 	// Make sure the token gets removed
@@ -363,7 +365,7 @@ func testCancel(t *testing.T, q jobqueue.JobQueue) {
 	require.Equal(t, jobType, "clownfish")
 	require.True(t, canceled)
 	require.Nil(t, result)
-	err = q.FinishJob(id, &testResult{})
+	err = q.RequeueOrFinishJob(id, 0, &testResult{})
 	require.Error(t, err)
 
 	// Cancel a running job, which should not dequeue the canceled job from above
@@ -383,7 +385,7 @@ func testCancel(t *testing.T, q jobqueue.JobQueue) {
 	require.Equal(t, jobType, "clownfish")
 	require.True(t, canceled)
 	require.Nil(t, result)
-	err = q.FinishJob(id, &testResult{})
+	err = q.RequeueOrFinishJob(id, 0, &testResult{})
 	require.Error(t, err)
 
 	// Cancel a finished job, which is a no-op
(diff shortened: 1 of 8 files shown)


internal/jobqueue/jobqueuetest/jobqueuetest.go
```diff
@@ -404,6 +405,48 @@ func testCancel(t *testing.T, q jobqueue.JobQueue) {
 	require.NoError(t, err)
 }
 
+func testRequeue(t *testing.T, q jobqueue.JobQueue) {
+	// Cancel a non-existing job
```
ondrejbudai, line 409: I think you forgot to change this comment.

ondrejbudai (commented):
I like it. Do you want to implement the missing bits on the worker side also in this PR?

teg:
> Do you want to implement the missing bits on the worker side also in this PR?

I think that would be better as a follow up as I think we need some rework there (and don't know when I'll have time).

internal/jobqueue/jobqueuetest/jobqueuetest.go
teg: Fixed! Thanks!

internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -56,6 +56,11 @@ const (
 		)
 		RETURNING token, type, args, queued_at, started_at`
 
+	sqlRequeue = `
+		UPDATE jobs
+		UNSET (started_at, token)
```
ondrejbudai, line 61: Hmm, the `dbjobqueue` tests fail on this statement and I wasn't able to find anything about `UPDATE UNSET`. Are you sure that this is valid?

ondrejbudai:
I gave this another thought and I think we need a reschedule limit. If a certain job always crashes the worker (which is scary but can happen), we don't want to retry it forever (it would be very hard to get rid of such a job).

teg:
> I gave this another thought and I think we need a reschedule limit. If a certain job always crashes the worker (which is scary but can happen), we don't want to retry it forever (it would be very hard to get rid of such a job).

Good point... Though should we distinguish between graceful shutdown (no limit?) and heartbeat failure (potential crash so try at most three times or something like that?).

teg:
Btw, I wanted to figure out how to unit test this better before resubmitting.

teg:
@Gundersanne I think this might help with the current lack of instance protection. it would make things slow when we retry, but at least it would not fail. Haven't gotten around to finish it up though :-/

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: `UPDATE SET` with token = uuid.Nil

teg:
I would do restart on timeout (to a limit) in this PR, then a follow-up where workers restart the build on termination (maybe without limit or with a higher one?) and independently instance protection. If someone could has the time to pick it up I think that would be great as I'm unsure when I could get to it

teg:
I'm unlikely to have time to finish this up, feel free to take it over / consider it an RFE, if anyone is interested.

internal/jobqueue/dbjobqueue/dbjobqueue.go
teg: Updated :)

internal/worker/server.go
```diff
@@ -122,9 +121,9 @@ func (s *Server) WatchHeartbeats() {
 				logrus.Panicf("Cannot marshal the heartbeat error: %v", err)
 			}
 
-			err = s.FinishJob(token, resJson)
+			err = s.jobs.RequeueOrFinishJob(id, 2, resJson)
```
croissanne, line 124: s.FinishJob makes a call to prometheus.FinishJobMetrics ; by calling the queue directly here this will now be missed by heartbeats.

teg: My bad! d'oh! Will fix.

pkg/jobqueue/jobqueue.go
```diff
@@ -95,6 +98,7 @@ var (
 	ErrNotExist       = errors.New("job does not exist")
 	ErrNotPending     = errors.New("job is not pending")
 	ErrNotRunning     = errors.New("job is not running")
-	ErrCanceled       = errors.New("job ws canceled")
+	ErrCanceled       = errors.New("job was canceled")
+	ErrMaxRetries     = errors.New("job reached max retries")
```
croissanne, line 102: ErrMaxRetries isn't used anywhere

croissanne (changes requested):
Regarding the failing api tests, we actually test the heartbeat thing by starting a job, killing the worker and wait until that job has failed. Of course now there's multiple retries we won't be able to test it in the same way. Either we need to keep restarting/stopping the worker, or instead of checking via the status, we wait on the jobqueue directly until that job has 1 retry.

https://github.com/osbuild/osbuild-composer/blob/a001da966abd29dce6efff24f94782540cc39f8f/test/cases/api.sh#L484-L489

instead of waitForStateFailure; run a query against the db container?

pkg/jobqueue/jobqueue.go
teg: Oops, left over from refactor.

ondrejbudai (approved):
This looks perfect to me.

croissanne:
I [missed a little mistake](https://github.com/osbuild/osbuild-composer/compare/2db051ceae14b36cadf2b9ca8654772bd0397a30..c085e23328314dc31bbe426747cb18d051cc933e#diff-8c9fc1f11806f990402d945e7a6a3bdcf8be429120a4b8c2ec18e9ff5251f02fR504-R506) when adding the test.

ondrejbudai (approved):

croissanne:
Merging this PR as the fedora failures are unrelated, and this is still going on to a reduced extent :(

PR #2462, opened 2022-03-20
📦 [WIP] upload: initial draft for container upload

gicmo:
📦  Nothing to see here yet but incomplete experiments. 🤫


cmd/osbuild-upload-container/main.go
```diff
@@ -0,0 +1,58 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/osbuild/osbuild-composer/internal/upload/container"
+)
+
+func main() {
+	var filename string
```
ygalblum, line 14: What about username and password?

internal/target/container_target.go
```diff
@@ -0,0 +1,17 @@
...
+
+	Username string `json:"username"`
+	Password string `json:"password"`
+
+	TlsVerify *bool `json:"tls_verify,omitempty"`
+}
+
+func (ContainerTargetOptions) isTargetOptions() {}
+
+func NewContainerTarget(options *AzureTargetOptions) *Target {
```
ygalblum, line 15: Is this `AzureTargetOptions` a copy paste mistake?

internal/upload/container/container.go
```diff
@@ -0,0 +1,183 @@
...
+	MaxRetries        int
+
+	UserAgent string
+	TlsVerify bool
+
+	policy *signature.Policy
+	sysCtx *types.SystemContext
+}
+
+func NewClient(destination string) (*Client, error) {
```
ygalblum, line 51: From the usage, it seems that the `username` and `password` are mandatory. If so, I think the `NewClient` method should receive them instead of having the caller set them separately

cmd/osbuild-upload-container/main.go
gicmo: Good point!

internal/target/container_target.go
gicmo: Most definitely is ;)

internal/upload/container/container.go
gicmo: The reason it is not is because I plan to add this hybrid support so that in case they are not explicitly set we can read them from a config file, i.e. in case of brew/koji.

ygalblum: If so, then shouldn't the `username` and `password` on the `ContainerTargetOptions` be optional? Or am I missing something?

gicmo: With optional you mean they should be pointer-to-strings?

ygalblum: In its essence I mean that the caller does not have to pass them in the API (with the hybrid support). I think that it does translate to being a pointer-to-string in the Go struct and define it as `omitempty` (same as you do for `TlsVerify`)

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 00cdb643b54917dffe3615839ef1634fef7bfea0 with the main merge-base fa382465758fff8d22bb377300472b028e69976c).  Please review the changes.

achilleas-k:
> warning This PR introduces changes in at least one manifest (when comparing PR HEAD [00cdb64](https://github.com/osbuild/osbuild-composer/commit/00cdb643b54917dffe3615839ef1634fef7bfea0) with the main merge-base [fa38246](https://github.com/osbuild/osbuild-composer/commit/fa382465758fff8d22bb377300472b028e69976c)). Please review the changes.

Thank you @schutzbot 

The changes are the new `container` image type in F36.

teg (approved):
Really nice! Thanks :)

PR #2466, opened 2022-03-21
koji: fix retries when uploading chunks

diaasami:
Use properly initialized retryablehttp client

These are fixes after investigating the 3 failures from earlier last week and the log message "Koji upload failed after 0 retries" which did not make sense.

This pull request includes:

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -341,9 +341,6 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
 
 	respData, err := client.Post(u.String(), "application/octet-stream", bytes.NewBuffer(chunk))
 
-	// don't count the first call since it's not a retry
-	retries--
-
 	if err != nil {
 		return retries, err
 	}
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -331,13 +331,13 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
 		return shouldRetry, retErr
 	}
 
-	client := rh.Client{
-		HTTPClient: &http.Client{
-			Transport: k.transport,
-		},
-		CheckRetry: countingCheckRetry,
-		Logger:     rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()}),
+	client := rh.NewClient()
+
+	client.HTTPClient = &http.Client{
+		Transport: k.transport,
 	}
+	client.CheckRetry = countingCheckRetry
+	client.Logger = rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()})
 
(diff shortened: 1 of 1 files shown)


teg:
Could you explain a bit more what the problem is and how this solves it? In particular how the counting changes.

internal/upload/koji/koji.go
```diff
@@ -323,27 +322,29 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
...
+
+	client.HTTPClient = &http.Client{
+		Transport: k.transport,
 	}
+	client.CheckRetry = countingCheckRetry
+	client.Logger = rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()})
 
 	respData, err := client.Post(u.String(), "application/octet-stream", bytes.NewBuffer(chunk))
 
-	// don't count the first call since it's not a retry
```
diaasami, line 344: The function is not called for the original HTTP call, only for retries, so no need to decrement.

internal/upload/koji/koji.go
```diff
@@ -323,27 +322,29 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
...
 				if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
 					// retry if it's a timeout
-					return strings.Contains(strings.ToLower(v.Error()), "timeout"), v
+					shouldRetry = strings.Contains(strings.ToLower(v.Error()), "timeout")
+					retErr = v
 				}
 			}
 		}
 
+		if shouldRetry {
```
diaasami, line 331: Only count if we decide to retry, not on every call of the function.

internal/upload/koji/koji.go
```diff
@@ -323,27 +322,29 @@ func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint6
...
 		return shouldRetry, retErr
 	}
 
-	client := rh.Client{
-		HTTPClient: &http.Client{
-			Transport: k.transport,
-		},
-		CheckRetry: countingCheckRetry,
-		Logger:     rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()}),
+	client := rh.NewClient()
```
diaasami, line 338: Properly initialized client includes defaultMaxRetries and defaultBackoff properties which are essential.

diaasami:
> Could you explain a bit more what the problem is and how this solves it? In particular how the counting changes.

I added comments and also updated the description

teg (approved):

internal/upload/koji/koji.go
ondrejbudai: May I ask why?

ondrejbudai (approved):
Thanks a lot, this is wondeful. :)

ondrejbudai:
Hmm, CI didn't run so I used the rebase feature of Github but it still didn't help, the maintenance is apparently still happening: https://www.githubstatus.com/

ondrejbudai (approved):

Issue #2413, opened 2022-03-09
Flaky ostree simplified installer test stuck on CS8


henrywang:
To continue test, same VM has to be destroyed.

```
[centos@centos-stream-8 ~]$ sudo virsh list --all
 Id   Name                                             State
----------------------------------------------------------------
 3    edge-7da686c7-cff3-46cb-98d6-c0814a26ad54-uefi   running

[centos@centos-stream-8 ~]$ sudo virsh destroy edge-7da686c7-cff3-46cb-98d6-c0814a26ad54-uefi
Domain 'edge-7da686c7-cff3-46cb-98d6-c0814a26ad54-uefi' destroyed

[centos@centos-stream-8 ~]$ sudo virsh list --all
 Id   Name                                             State
----------------------------------------------------------------
 4    edge-7da686c7-cff3-46cb-98d6-c0814a26ad54-uefi   running
```

jrusz:
This now also happened on CS-9 and RHEL-9 as well... 
RHEL-9: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2249057799 on compose RHEL-9.0.0-20220324.2
and CS-9: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2246004901 (it has been manually unstuck here the same way like cs-8)

ondrejbudai:
Happened a lot during last week: https://www.osbuild.org/ci-flakiness/2022-03-28T08:57:39Z.html

jrusz:
Yeah it's now happening also on RHEL-9.0 and 8.6 nightly and I missed one new `virt-install` in the test, adding workaround here: https://github.com/osbuild/osbuild-composer/pull/2487

henrywang:
It's fixed by PR #2487

PR #2468, opened 2022-03-22
Minor CI changes

jrusz:
I just want to put in a few things related to the CI. 

1. I want to run `ci-details.sh` before and after the job execution to have exact information about packages installed and updated on the system. 
2. I added `yum repolist -v` to see what repos exactly are enabled on the runner
3. centos-stream repos were never overwritten by the pin in Schutzfile because of the way how the override works. It's easy to fix on centos-9 as it has one repofile `centos.repo` but on centos-8 there are is one repofile per repo so that needs an update to how the override works.

This pull request includes:


jrusz:
I've included two more changes:

1.  Re-enabling `Installers.sh` on RHEL-9 as the related BZ has been fixed (CS-9 is waiting for a newer anaconda release: https://kojihub.stream.centos.org/koji/buildinfo?buildID=17690 related bugs have to be preverified first)
2. "Fix" for `ostree-simplified-installer.sh` getting stuck. It destroys the VM after 15 minutes which should be enough time for the installation to finish. Related: https://github.com/osbuild/osbuild-composer/issues/2413

test/cases/ostree-simplified-installer.sh
```diff
@@ -495,9 +495,15 @@ sudo virt-install --name="${IMAGE_KEY}-http"\
...
-                  --wait=-1 \
+                  --wait=15 \
                   --noreboot
 
+# Installation can get stuck, destroying VM helps
+# See https://github.com/osbuild/osbuild-composer/issues/2413
+if [[ $(sudo virsh domstate "${IMAGE_KEY}-http") == "running" ]]; then
+    sudo virsh destroy "${IMAGE_KEY}-http"
+fi
+
```
thozza, line 506: This may be a stupid question, but should not this conditional and the `virt-install` command from above be in some kind of a for loop? I expect that the VM still needs to be installed for the test to pass, so just destroying it may unstuck the test, but it also means that the VM was not installed successfully, right?

test/cases/ostree-simplified-installer.sh
```diff
@@ -621,9 +627,15 @@ sudo virt-install  --name="${IMAGE_KEY}-fdosshkey"\
...
-                   --wait=-1 \
+                   --wait=15 \
                    --noreboot
 
+# Installation can get stuck, destroying VM helps
+# See https://github.com/osbuild/osbuild-composer/issues/2413
+if [[ $(sudo virsh domstate "${IMAGE_KEY}-fdosshkey") == "running" ]]; then
+    sudo virsh destroy "${IMAGE_KEY}-fdosshkey"
+fi
+
```
thozza, line 638: ditto the for loop question...

thozza (commented):
Thanks for these changes!

I'll need some help to understand the `ostree-simplified-installer.sh` workaround... 🤔

test/cases/ostree-simplified-installer.sh
jrusz: I don't think there is a loop happening, IMO it's just finished and waiting for something and AFIAK nobody knows what, am I right @henrywang ? 

So what this does is that if after 15 minutes of running `virt-install` it's expected that it's finished and the condition is there only to prevent failure in case the `virt-install` was successful and the VM was turned off already. Of course this doesn't fix anything but it's better than disabling the test entirely. And in case it was destroyed prematurely the following ansible check will fail or the VM won't boot at all so it won't introduce any false positives.

jrusz: Putting it into a loop would be probably necessary in case the stuck installation meant failed installation which apparently is not the case here.

thozza: OK, thanks for the explanation. It the VM is actually stuck after successful installation, then it makes sense to do what you did in the PR. It makes sense now 🙂

thozza (approved):
LGTM

PR #2474, opened 2022-03-23
ci/tests: Change the way artifacts are collected

jrusz:
We want to be able to safely gather any artifacts without worrying about
any possible secrets leaking. Every artifacts that we want to upload
will now have to be placed in /tmp/artifacts which will then be uploaded
to S3 by the executor and link to the artifacts will be provided in the
logs. Only people with access to our AWS account will be able to see the artifacts.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,21 +8,18 @@ stages:
 
 .base:
   before_script:
-    - schutzbot/ci_details.sh > ci-details-before-run
+    - schutzbot/ci_details.sh > ci-details-before-run.txt
     - cat schutzbot/team_ssh_keys.txt | tee -a ~/.ssh/authorized_keys > /dev/null
   after_script:
-    - schutzbot/ci_details.sh > ci-details-after-run || true
+    - schutzbot/ci_details.sh > ci-details-after-run.txt || true
     - schutzbot/update_github_status.sh update || true
     - schutzbot/save_journal.sh || true
+    - schutzbot/upload_artifacts.sh
   interruptible: true
   tags:
     - terraform
   artifacts:
     paths:
-      - journal-log.gpg
-      - ci-artifacts
-      - ci-details-before-run
-      - ci-details-after-run
       - "*.repo"
       - COMPOSER_NVR
     when: always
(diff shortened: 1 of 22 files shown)


.gitlab-ci.yml
```diff
@@ -8,21 +8,19 @@ stages:
 
 .base:
   before_script:
-    - schutzbot/ci_details.sh > ci-details-before-run
+    - schutzbot/ci_details.sh > ci-details-before-run.txt
     - cat schutzbot/team_ssh_keys.txt | tee -a ~/.ssh/authorized_keys > /dev/null
+    - mkdir /tmp/artifacts
```
atodorov, line 13: Why not create the directory first and then have the .txt, log files, etc created inside of the directory and make the artifact collection script less aware of what artifacts it needs to collect?

jrusz: I guess it was just easier to implement. Currently the CI puts pretty much everything to the current directory which is then uploaded to Gitlab CI.. So all I had to do was take any files with the relevant suffix or extension and put them in the artifacts folder.

Otherwise I think I would have to go through the tests again and make them put the logs and any important files in the artifacts directory... But I guess that makes more sense so I'll update it a bit.

jrusz:
I've made it a bit better. The script now uploads anything in the `$ARTIFACTS` folder to S3 and you can set your own `$ARTIFACTS` env var if needed otherwise it will default to `/tmp/artifacts`.

The `$ARTIFACTS` directory is created in `provision.sh` which is sourced in every test script AFAIK so it should be fine, CI should discover any issues with it.

I've also dropped the encryption of journal and some db dump in api tests because these artifacts are only accessible after logging it to the image-builder AWS account so it should be safe there and easy for us to read if needed.

jrusz:
> But what's the advantage of having it in s3 vs an encrypted tarball in gitlab?

IMO it's much easier to access because you can just open it in your browser and look at it instead of downloading it, decrypting and then looking at it. 
And if it's all in S3 where only our team has access we don't have to worry about leaking any secrets.

> Also should we be putting it in /tmp? Unsure how big some of the artifacts get. This is probably not a concern.

Yeah I didn't think of that, I hope there won't be any issues with it, we can always change it later if it becomes a problem.

atodorov (approved):

PR #2477, opened 2022-03-23
Dependency joberrors

kingsleyzissou:
This pull request includes:
Reporting that there has been an issue with a dependency job does not provide enough information about the dependency
failure. It makes more sense to bubble up the original error since this will tell us the information we need. It also has the added benefit of ensuring consistency of the status codes for metrics.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild-koji.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild-koji.go
@@ -124,12 +124,12 @@ func (impl *OSBuildKojiJobImpl) Run(job worker.Job) error {
 			args.Manifest = manifestJR.Manifest
 			if len(args.Manifest) == 0 {
 				err := fmt.Errorf("Received empty manifest")
-				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorManifestDependency, err.Error())
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorEmptyManifest, err.Error())
 				return err
 			}
 		} else {
 			err := fmt.Errorf("Job has no manifest")
-			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorManifestDependency, err.Error())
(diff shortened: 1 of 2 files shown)


Gundersanne (approved):

ondrejbudai:
I'm not sure if this approach is correct. Generally speaking, a job can have multiple dependencies and each of them can fail. With this approach, we would have to have an array of `JobError`s which sounds complicated. Also, we already have a job (`koji-finalize`) that does some stuff (and thus can fail) even though its dependency failed - so we have to communicate its status back somehow.

Since the api servers are the ones who are responsible for queueing jobs in the correct order, I think they should be ultimately responsible for inspecting the whole job chain and cherry-picking errors from it.

Gundersanne:
> I'm not sure if this approach is correct. Generally speaking, a job can have multiple dependencies and each of them can fail. With this approach, we would have to have an array of `JobError`s which sounds complicated. Also, we already have a job (`koji-finalize`) that does some stuff (and thus can fail) even though its dependency failed - so we have to communicate its status back somehow.
> 
> Since the api servers are the ones who are responsible for queueing jobs in the correct order, I think they should be ultimately responsible for inspecting the whole job chain and cherry-picking errors from it.

Didn't consider the koji case in this. In that case is it fine to consider all dependency failure errors as 4xx in metrics? That way it'll only the dependency shows up (in case of 5xxs).

kingsleyzissou:
Okay, will have to re-think this a bit then, I also hadn't considered the koji side of things properly.

> Didn't consider the koji case in this. In that case is it fine to consider all dependency failure errors as 4xx in metrics? That way it'll only the dependency shows up (in case of 5xxs).

I guess technically if a dependency failed, it's bad input for the next job (my two-cents)

kingsleyzissou:
@Gundersanne @ondrejbudai  I dropped the commit for propagating the dependency errors. I'll do that in a follow up.
For the meantime I changed the status codes for the metrics when we encounter a dependency error (assuming we're happy to go ahead with that change).

ondrejbudai (dismissed):

Gundersanne (approved):

PR #2479, opened 2022-03-24
cloudapi: improve gpgkey handling & enable edge-commit on Fedora

ondrejbudai:
We have to do a small hack to enable edge-commit on Fedora because its name
is different. We can also change this in the image definition but I want to
iterate quickly on the Fedora Integration MVP and don't want to run in
any conflicts with
https://github.com/osbuild/osbuild-composer/pull/2461

This commit also enables a test for Fedora IoT built through the API.

This pull request includes:


teg:
I would love to just call it edge commit in our API long term, but I guess that's up to @nullr0ute . For now this is definitely fine though.

ondrejbudai:
Agreed, more unification between Fedora and RHEL is always nice.

(Force-pushed a fix for missing Fedora bits in the test.)

Gundersanne (dismissed):

nullr0ute:
> I would love to just call it edge commit in our API long term, but I guess that's up to @nullr0ute . For now this is definitely fine though.

Happy to consider options, probably needs more context and OOB discussion though.

Patch looks fine to me

ondrejbudai:
Oh my, so we were actually passing gpg keys in the wrong field in our tests so that's why this PR was failing: cloudapi uses a field named `gpg_key` whereas all our other APIs (and DNF repo files) use `gpgkey`.

I decided to change that in cloudapi: both `gpgkey` and `gpg_key` can now be used. `gpg_key` is marked as deprecated but kept for backward compatibility. Also, we now enforce that if `check_gpg` is present, a GPG key must be specified. This would have caught this issue.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -669,8 +669,13 @@ components:
...
+          description: |
+            GPG key used to sign packages in this repository.
+            
+            Setting both gpg_key and gpgkey is not supported, use just the gpgkey field.
```
Gundersanne, line 678: Description needs tweak.

ondrejbudai: True, thanks!

ondrejbudai:
So we in the end decided to change the API instead of introducing cumbersome hacks. This is a breaking change but since no one is using the API, it does little to no harm.

Gundersanne (approved):

PR #2486, opened 2022-03-28
Add Xiaofeng to notifications for nightly pipeline

atodorov:
b/c we need to be able to quickly react on ostree failures when they happen.


jrusz (approved):
Good idea, LGTM

PR #2497, opened 2022-03-30
Fix excessive logging and monitoring

diaasami:
This pull request includes:

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -162,7 +162,7 @@ func NewFromPlain(server, user, password string, transport http.RoundTripper) (*
 	// The API doesn't require sessionID, sessionKey and callnum yet,
 	// so there's no need to use the custom Koji RoundTripper,
 	// let's just use the one that the called passed in.
-	rhTransport := &rh.RoundTripper{Client: rh.NewClient()}
+	rhTransport := CreateRetryableTransport()
 	loginClient, err := xmlrpc.NewClient(server, rhTransport)
 	if err != nil {
 		return nil, err
@@ -443,7 +443,7 @@ func GSSAPICredentialsFromEnv() (*GSSAPICredentials, error) {
 func CreateKojiTransport(relaxTimeout uint) http.RoundTripper {
 	// Koji for some reason needs TLS renegotiation enabled.
 	// Clone the default http rt and enable renegotiation.
-	rt := &rh.RoundTripper{Client: rh.NewClient()}
+	rt := CreateRetryableTransport()
 
 	transport := rt.Client.HTTPClient.Transport.(*http.Transport)
 
@@ -486,4 +486,10 @@ func createCustomRetryableClient() *rh.Client {
 	client.Logger = rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()})
 	client.CheckRetry = customCheckRetry
 	return client
+}
+
+func CreateRetryableTransport() *rh.RoundTripper {
+	rt := rh.RoundTripper{}
+	rt.Client = createCustomRetryableClient()
+	return &rt
 }
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -295,14 +295,14 @@ func (k *Koji) CGImport(build ImageBuild, buildRoots []BuildRoot, images []Image
 }
 
 // uploadChunk uploads a byte slice to a given filepath/filname at a given offset
-func (k *Koji) uploadChunk(chunk []byte, filepath, filename string, offset uint64) (uint, error) {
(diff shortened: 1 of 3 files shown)


ondrejbudai (approved):
Awesome, looks very nice. :)

internal/upload/koji/koji.go
```diff
@@ -492,3 +465,33 @@ func CreateKojiTransport(relaxTimeout uint) http.RoundTripper {
...
+				// retry if it's a timeout
+				return strings.Contains(strings.ToLower(v.Error()), "timeout"), v
+			}
+		}
+	}
+
+	return shouldRetry, retErr
+}
+
+func CreateCustomRetryableClient() *rh.Client {
```
teg, line 486: Should this be public?

internal/upload/koji/koji.go
```diff
@@ -492,3 +465,33 @@ func CreateKojiTransport(relaxTimeout uint) http.RoundTripper {
 
 	return rt
 }
+
+func CustomCheckRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
```
teg, line 469: Should this be public?

internal/upload/koji/koji.go
```diff
@@ -492,3 +465,33 @@ func CreateKojiTransport(relaxTimeout uint) http.RoundTripper {
...
+}
+
+func CreateCustomRetryableClient() *rh.Client {
+	client := rh.NewClient()
+	client.Logger = rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()})
+	client.CheckRetry = CustomCheckRetry
+	return client
+}
+
+func CreateRetryableTransport() *rh.RoundTripper {
```
teg, line 493: I think this and the previous function belongs in the next patch, no?

Also, the commit message seems to describe the next patch. This patch just drops log messages, right? Presumably because the previous one adds the important ones as `Info()` so these are not needed? I think a bit more of an explanation of the change in behaviour of the previous and this patch combined would be helpful.

internal/upload/koji/koji.go
```diff
@@ -162,7 +162,7 @@ func NewFromPlain(server, user, password string, transport http.RoundTripper) (*
 	// The API doesn't require sessionID, sessionKey and callnum yet,
 	// so there's no need to use the custom Koji RoundTripper,
 	// let's just use the one that the called passed in.
```
teg, line 164: Is this comment still up-to-date?

internal/upload/koji/koji.go
```diff
@@ -445,7 +445,7 @@ func GSSAPICredentialsFromEnv() (*GSSAPICredentials, error) {
 func CreateKojiTransport(relaxTimeout uint) http.RoundTripper {
 	// Koji for some reason needs TLS renegotiation enabled.
 	// Clone the default http rt and enable renegotiation.
-	rt := &rh.RoundTripper{Client: rh.NewClient()}
```
teg, line 475: Is this not also used from `osbuild-koji`? You mentioned only `init` and `finalise` in the commit message, but it applies to everything, right?
Could you explain in the commit message exactly what the behaviour difference is? Is it only about logging, or does it improve the retry logic in some additional cases too?

internal/upload/koji/koji.go
```diff
@@ -162,7 +162,7 @@ func NewFromPlain(server, user, password string, transport http.RoundTripper) (*
 	// The API doesn't require sessionID, sessionKey and callnum yet,
 	// so there's no need to use the custom Koji RoundTripper,
 	// let's just use the one that the called passed in.
-	rhTransport := &rh.RoundTripper{Client: rh.NewClient()}
```
teg, line 165: I guess this is only relevant in tests? I got a bit tripped up trying to figure out how it related to koji init / finalise.

teg (changes requested):
Looks good, but would be great if you could clarify a bit in the commit messages to make it easier for others who didn't work on this code to follow :)

internal/upload/koji/koji.go
diaasami: That's correct, I think it's important to have the code path tested to avoid deploying untested code.

teg: I agree, code looks good, just needs clarification in the commit message, I think.

internal/upload/koji/koji.go
teg: I got this now, thanks!

internal/upload/koji/koji.go
teg: I think this is clear now from the context.

teg (approved):
This is really nice! Thanks a lot!

internal/upload/koji/koji.go
diaasami: I believe it is, I will double check it.

PR #2508, opened 2022-04-01
Pin fedora repositories in Schutzfile + override mock templates with rpmrepo snapshots

jrusz:
This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,7 @@ stages:
 .base:
   before_script:
     - schutzbot/ci_details.sh > ci-details-before-run
+    - cat schutzbot/team_ssh_keys.txt | tee -a ~/.ssh/authorized_keys > /dev/null
   after_script:
     - schutzbot/ci_details.sh > ci-details-after-run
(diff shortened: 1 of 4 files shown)


ondrejbudai (commented):
What about just doing `cat /etc/yum.repos.d/* | sudo tee -a`? Might save us a few lines.

Also, it would be great to unify the cases a bit, it's almost the same code, just the file names differ.

Btw: what about Fedora?

atodorov (dismissed):
LGTM.

jrusz:
@ondrejbudai I've updated it, I think it looks a bit better now, can you take another look please? 

Also I'll need to apply some `sed` magic to https://github.com/osbuild/rpmrepo/blob/main/.github/workflows/scheduled-snapshots.yml#L111 because we don't want to override the main fedora repos because they don't change. Suggestions more than welcome :sweat_smile:

jrusz:
I forgot to remove one commit.

jrusz:
I forgot to remove one commit.

thozza (approved):
LGTM

PR #2517, opened 2022-04-05
test/ostree-ng: remove user customization from upgrade BP

jrusz:
A user does not need to be specified in upgrade blueprint in order for
it to be present after the upgrade. Related rhbz#2060575

This pull request includes:

### DIFF
--- a/test/cases/ostree-ng.sh
+++ b/test/cases/ostree-ng.sh
@@ -618,13 +618,6 @@ version = "*"
 
 [customizations.kernel]
 name = "kernel-rt"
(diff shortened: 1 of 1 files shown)


jrusz:
@henrywang @achilleas-k so do you think this change makes sense now to test the mentioned BZ? The admin user is used for checking the ssh connection so if it got removed by the fact that it's missing in the upgrade blueprint it will be caught, right?

achilleas-k (approved):
Yes, this is good.

PR #2417, opened 2022-03-09
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing enhancement.

We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

This pull request includes:

- [x] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #2518, opened 2022-04-05
cloudapi: prevent dangling goroutines after the server is terminated

ondrejbudai:
By merging https://github.com/osbuild/osbuild-composer/pull/2417 I introduced flakey tests, oh no.

`T.TempDir` basically checks whether something writes into the directory *after* the test finishes, see https://github.com/golang/go/issues/43547.

This could happen in `cloudapi` because we have the manifest generation goroutines: When a test finishes, there might be still one running and it might write something into the directory created by `T.TempDir` and the test will explode.

The underlying issue is that we don't wait for all manifest generation goroutines to finish when shutting down the cloudapi server. This is potentially dangerous because the goroutine might be interrupted in a weird, non-defined state.

This PR refactors `cloudapi` a bit and adds a synchronization mechanism that waits for all manifest generation goroutines to finish.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -330,7 +330,9 @@ func (c *Composer) Start() error {
 	logrus.Info("Shutting down.")
 
 	if c.apiListener != nil {
+		// First, close all listeners and then wait for all goroutines to finish.
 		err := composerAPI.Shutdown(context.Background())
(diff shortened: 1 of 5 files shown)


thozza (approved):
LGTM, nice work 👍

Gundersanne (approved):
lgtm!

PR #2520, opened 2022-04-06
Empty manifest check for osbuild jobs

kingsleyzissou:
This pull request includes:
- empty manifest error case for osbuild jobs
- empty packagespec error case for depsolve jobs with `4xx`

### DIFF
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -243,23 +243,6 @@ func (s *Server) DepsolveJobStatus(id uuid.UUID, result *DepsolveJobResult) (*Jo
 	return status, deps, nil
 }
 
-func (s *Server) ManifestByIdJobStatus(id uuid.UUID, result *ManifestJobByIDResult) (*JobStatus, []uuid.UUID, error) {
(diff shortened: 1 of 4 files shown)


internal/worker/server.go
```diff
@@ -249,8 +249,8 @@ func (s *Server) ManifestByIdJobStatus(id uuid.UUID, result *ManifestJobByIDResu
 		return nil, nil, err
 	}
 
-	if jobType != "manifest-by-id" {
-		return nil, nil, fmt.Errorf("expected \"koji-init\", found %q job instead", jobType)
+	if jobType != "manifest-id-only" {
```
Gundersanne, line 252: ah good catch, I think this needs 'manifest-id-only' everywhere no? Also on line 269.

Gundersanne (changes requested):

internal/worker/server.go
kingsleyzissou: Yeah good point! I actually think this function isn't used at all and only the other one has any references

Gundersanne (approved):
lgtm, thanks!

PR #2522, opened 2022-04-06
distro/rhel86: enable auto-registration for azure

gicmo:
We want to ensure that cloud images connect to Red Hat[1] independently of how the content was acquired (PAYG, BYOS, or marketplace).
This auto-registration feature is already enabled for AWS and this patch enables it for Azure with the same recommended settings:

Services:

    rhsmcertd: Enabled (already done, so not changed in the patch)

`/etc/rhsm/rhsm.conf`:

    auto_registration: enabled
    auto_registration_interval: 60 (the default, so not explicitly set)
    manage_repos: false

The latter value `manage_repos` is left enabled (the default) in case the user explicitly requested to have the system subscribed, i.e. the `RHSMConfigWithSubscription` code path.

Regenerate the relevant test manifests and image information.

[1] https://cloud.redhat.com
[2] https://docs.google.com/document/d/1VeZFJxNUlyZMQJh6s3NA3RLvadqATsGxVet6uuP87_4

### DIFF
--- a/internal/distro/rhel86/distro.go
+++ b/internal/distro/rhel86/distro.go
@@ -977,6 +977,23 @@ func newDistro(distroName string) distro.Distro {
 							Enabled: false,
 						},
 					},
+					SubMan: &osbuild.RHSMStageOptionsSubMan{
+						Rhsmcertd: &osbuild.SubManConfigRHSMCERTDSection{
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):
LGTM

PR #2538, opened 2022-04-10
distro: Improve usage of interfaces in distros

jordigilh:
This pull request includes:

- [X ] adequate testing for the new functionality or fixed issue

Refactored the code in all distributions to leverage on the interface functions for the `architecture` and `distribution` structs instead of accessing directly the field in those local structures. The purpose of this change is to be able to improve reusability and reduce duplicated code, such as in the pipelines, that is now tied to the local structures in each distribution.

@thozza can you take a look? Feel free to close the PR at your discretion.


jordigilh:
> 

Thanks for the feedback :smile: . Forgive my ignorance, I was not sure how to proceed with this and decided to tackle it without consultation, knowing full well that it could be a total disaster :wink:.

I'll follow your advice and focus only on the 8.6 and 9.0 and revert the changes for the other distributions, which I understand they will eventually be phased out and removed, so no need to pollute the interface :smile: .

jordigilh:
@thozza reverted my changes and updated with these 2 changes as you suggested:
* Changed `t.arch.name` to `t.Arch().Name()`
* Changed string references to CPU architectures to use already defined constants.


I decided not to explore any further changes to the interfaces, since it seems it would require further discussions and that was not part of my intent in this PR, except to do some housekeeping behind the scenes :smile: .

Let me know if this works better for you :smile:

thozza (approved):
LGTM, thanks!

PR #2540, opened 2022-04-11
rhel85: specify a minimum root file system size

gicmo:
Specify a size for the root filesystem in the partition table, which basically equates to a minimum size. In reality all image types specify a larger image size and thus we enlarge the root file system to more than the specified size for plain layouts.

Does not change any existing manifests.

This also prepares the enablement of auto-LVM conversion, since in that case we need to have a size for the root file system specified.

Should fix the failing 8.5 test in #2412

### DIFF
--- a/internal/distro/rhel85/partition_tables.go
+++ b/internal/distro/rhel85/partition_tables.go
@@ -41,6 +41,7 @@ var defaultBasePartitionTables = distro.BasePartitionTableMap{
 				},
 			},
 			{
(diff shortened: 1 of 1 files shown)


kingsleyzissou (approved):

jrusz:
Thanks for the fix @gicmo , it's merged as part of https://github.com/osbuild/osbuild-composer/pull/2412 so you can close this PR :)

PR #2543, opened 2022-04-11
Generic S3 test - retry creating the alias in case the service is not yet up

ygalblum:
@jrusz This addresses the issue you saw in #2412. It seems that sometimes the service takes a but longer to start so the client needs to wait and retry


jrusz (changes requested):
Cool, thanks! Can you please make Shellcheck happy? He doesn't like the unused `i` variable. Maybe change it for `_` ?

jrusz (approved):
LGTM, thanks! :)

PR #2550, opened 2022-04-13
Refuse manual start of `osbuild-composer.service`

thozza:
The `osbuild-composer.service` unit should be started only by one of the
Socket units. Starting it manually by the user results in unexpected
behavior and in general non-functional osbuild-composer service.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=2072488

This pull request includes:

### DIFF
--- a/distribution/osbuild-composer.service
+++ b/distribution/osbuild-composer.service
@@ -8,6 +8,9 @@ After=network.target
 # to disable it.
 Wants=osbuild-worker@1.service
 
+# This unit should be started only via a socket unit
+RefuseManualStart=true
+
 [Service]
 Type=simple
 ExecStart=/usr/libexec/osbuild-composer/osbuild-composer


ondrejbudai (approved):

thozza:
The end result is pretty reasonable

```
# systemctl start osbuild-composer.service
Failed to start osbuild-composer.service: Operation refused, unit osbuild-composer.service may be requested by dependency only (it is configured to refuse manual start/stop).
See system logs and 'systemctl status osbuild-composer.service' for details.
```

```
# systemctl status osbuild-composer.service
○ osbuild-composer.service - OSBuild Composer
     Loaded: loaded (/usr/lib/systemd/system/osbuild-composer.service; disabled; vendor preset: disabled)
     Active: inactive (dead)
TriggeredBy: ● osbuild-composer.socket
             ○ osbuild-local-worker.socket
```

gicmo (approved):
❤️

ondrejbudai:
I think this needs a bigger discussion because it also disallows `systemctl restart osbuild-composer.service` which is definitely a useful command.

I would actually be in favor of adding `Requires: osbuild-composer.socket` into `osbuild-composer.service`. The reason why we currently don't have it is that we don't want weldr to be started on composer.osbuild.org. This was implemented back in 71428d3131f64516c9ddc46e42cfbd22ef964678.

Since composer.osbuild.org is going away soon and after that, all of our composer deployments will be in OpenShift, I think we can safely revert the mentioned commit.

As a follow-up, I would drop `osbuild-composer-api.socket` and support running the API only in a container. As a second follow-up, I want to propose dropping even `osbuild-remote-worker.socket` that effectively means that weldr API is only able to run with a local worker, which is actually fine.

PR #2551, opened 2022-04-13
Service maintenance test

Gundersanne:
We have 2 choices really, either remove the records entirely or just zero out the manifest. This PR removes them entirely. Happy to drop this in favour of just removing the manifest however.

### DIFF
--- a/internal/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/internal/jobqueue/dbjobqueue/dbjobqueue.go
@@ -104,21 +104,18 @@ const (
 	sqlQueryJobsUptoByType = `
                 SELECT array_agg(id), type
                 FROM jobs
-                WHERE type = ANY($1) AND finished_at < $2
+                WHERE type = ANY($1) AND queued_at < $2
                 GROUP BY type`
-	sqlQueryDepedenciesRecursively = `
+	sqlQueryDepedendantsRecursively = `
                 WITH RECURSIVE dependencies(d) AS (
-                                SELECT dependency_id
+                                SELECT job_id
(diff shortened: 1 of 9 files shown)


internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -493,6 +493,9 @@ func (q *DBJobQueue) JobStatus(id uuid.UUID) (jobType string, result json.RawMes
 	var sp, fp *time.Time
 	var rp pgtype.JSON
 	err = conn.QueryRow(context.Background(), sqlQueryJobStatus, id).Scan(&jobType, &rp, &queued, &sp, &fp, &canceled)
+	if err == pgx.ErrNoRows {
```
ondrejbudai, line 496: This should be in a separate commit I think.

ondrejbudai (commented):
> If multiple osbuild jobs per depsolve are ever implemented, this
will not break.

Won't it break the data retention promise, though? If depsolve was done more than 14 days ago, the latest osbuild job based on it might be much newer/.

Gundersanne:
> > If multiple osbuild jobs per depsolve are ever implemented, this
> > will not break.
> 
> Won't it break the data retention promise, though? If depsolve was done more than 14 days ago, the latest osbuild job based on it might be much newer/.

The assumption in this script is that the build happened within 24 hours of the depsolve. I we are working towards this, then we can't do this.

Gundersanne:
We might need the osbuild result  around for a long time for registering purposes. For this reason I'll just open a PR which zeros out the manifest (and depsolve?) job results.

PR #2552, opened 2022-04-13
rhel85: automatically convert to LVM on fs customizations

gicmo:
Whenever we create a new mountpoint due to a user customization, ensure the layout uses LVM, i.e. convert plain layouts to it, if needed. This does not apply to rpm-ostree based systems.
Add "lvm2" to the build pipeline and thus generate new manifests and image infos.
Adjust the existing tests that assumed we can not create more than 4 partitions on mbr layouts, since that is now not true anymore.

This is a port from rhel86, commit 63aa1556fcff4b3038f504f6dc5e239d2c4d97f1

TODO:
 - [x] manifest regeneration


teg (dismissed):
Would love to have this for GA!

achilleas-k (approved):

teg (dismissed):

achilleas-k:
I'm generating the new image info for this

achilleas-k:
Pushed the new manifests with image-info (squashed).
Some uids and gids changed again.  I'd say it's expected given the packages changed, changing the install order.

teg (dismissed):

achilleas-k:
Manifests and image info should be good now.

achilleas-k:
Found a couple of bugs in `ensureLVM()` which make LVMification fail when the base partition table has no `/boot`

1. https://github.com/osbuild/osbuild-composer/blob/1e520719cb6aeab54a58efd0ebdd7345d0ad4295/internal/disk/partition_table.go#L478-L481
We unconditionally return from here without checking err.
2. https://github.com/osbuild/osbuild-composer/blob/1e520719cb6aeab54a58efd0ebdd7345d0ad4295/internal/disk/partition_table.go#L471-L488
The `CreateMountpoint()` method ends with
```
	pt.Partitions = append(pt.Partitions, partition)
```
which _might_ change the address of all the items in `pt.Partitions`, but we hold on to `rootPath` and use the objects we got there to create the root VolumeGroup.

This might be the cause of the Azure boot failure, which I was reproducing when I found these issues.

gicmo:
We need to fix the `org.osbuild.fix-bls` stage in the `osPipeline` pipeline to change the prefix iff there is a `/boot` like we do in the ec2 pipeline.

achilleas-k:
New commit should fix the failing Azure test with the addition of the correct prefix in the fix-bls stage.  See commit message for details.

achilleas-k:
fixing this up and rebasing it on #2580

PR #2556, opened 2022-04-14
distro/rhel90beta: alias to rhel-90

gicmo:
Remove the rhel 90 beta code, which has fallen behind in terms of features and bug fixes and make it an alias to 9.0, i.e. it can still be requested but it will end up using the rhel 9.0 image definitions. Also remove the manifests and the entries in the manifest generator.


teg (approved):

achilleas-k (approved):
I love this.

ondrejbudai:
This now has a ton of conflicts unfortunately. :(

gicmo:
;(

thozza:
This looks all good and the conflicts are easily resolvable by just deleting those manifests again while rebasing.

However, I (#2350) and QE has been attempting to create a test case, which would ensure that each supported distro x arch x image type combination is covered by an image test case. By deleting all rhel90beta, we fix the current issue with manifests changing due to aliasing to rhel90, but this will strike back once we will test that each image type has a corresponding test case. Therefore I propose that we at least generate image test cases without the `image-info` part for the aliased rhel90beta distro.

thozza:
@gicmo the `tools/test-case-generators/generate-all-test-cases` now has `manifests` command, which can be used to generate image test cases with only manifests and it does it locally for all arches. So it should be fairly easy to generate just manifests for the aliased 90 Beta 😉

PR #2562, opened 2022-04-19
templates/packer: Rely on instance metadata to set region

Gundersanne:
Vector 0.21 needs region set otherwise the healthcheck will
fail. Instead of explicitly setting the aws cloudwatch endpoint, just
set the region.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
+++ b/templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
@@ -4,6 +4,8 @@ source /tmp/cloud_init_vars
 
 echo "Writing vector config."
 
+REGION=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .region)
+
 sudo mkdir -p /etc/vector
 sudo tee /etc/vector/vector.toml > /dev/null << EOF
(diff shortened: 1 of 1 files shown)


Gundersanne:
The alternative is setting both region and endpoint; or disabling the healthcheck.

templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
```diff
@@ -13,7 +15,7 @@ exclude_units = ["vector.service"]
 [sinks.out]
 type = "aws_cloudwatch_logs"
 inputs = [ "journald" ]
-endpoint = "${CLOUDWATCH_LOGS_ENDPOINT_URL}"
+region = "${REGION}"
```
ondrejbudai, line 18: Can you bring back the `endpoint`? It will be useful when we will move internal workers to AWS.

It's weird that vector requires `region` because `endpoint` arguably encodes everything that they need but I guess they had reasons for making it required. :shrug:

Gundersanne: It's just for the healthcheck, i think it's a bug. I filed it upstream in any case.

ondrejbudai (approved):

PR #2567, opened 2022-04-20
(WIP) worker: protect the instance from upgrading

lavocatt:
Before the instance was vulnerable to an OTA update while processing a
request. Because there is no way of retriggering a job in Composer, it
is better to avoid this situation.
The way we are doing it is by setting the `protected` flag onto the
instance when a job is being processed. This way the AWS scheduler
does hopefully not shutdown the machine at the wrong time.

Main caveats of this solution:
* Starvation: If a worker keeps accepting new jobs, then it might not be
  updated.
* Inconsistency: There exist a window between the job acceptation and the
  protection where the worker can be shutdown without having the time to
  protect itself.

This pull request includes:


cmd/osbuild-worker/main.go
```diff
@@ -84,6 +88,68 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+	awsSession, err := session.NewSession()
+	if err != nil {
+		logrus.Error("Error getting AWS session")
+		return err
+	}
+
+	// get the identity for the instanceID
+	identity, err := ec2metadata.New(awsSession).GetInstanceIdentityDocument()
+	if err != nil {
+		logrus.Error("Error getting identity")
```
Gundersanne, line 118: Shouldn't be an error imo, the instance identity document is one of the main ways we can identify if we're running on aws.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +88,68 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+	}
+	return false
+}
+
+// protect an AWS instance from scaling and/or terminating.
+func setProtection(protected bool) error {
+	// create a new session
+	awsSession, err := session.NewSession()
+	if err != nil {
+		logrus.Error("Error getting AWS session")
```
Gundersanne, line 111: Shouldn't be an error. I'd use this + getting the instance identity document to see if we're running on aws.

Gundersanne (commented):
Looking good, thank you!

One major concern however, is that the workers run 2 RequestAndRunJob routines indefinitely, one for depsolving and one for osbuild jobs. And this would create a race condition. So protection needs to be synced across all RequestAndRunJob routines.

And as a second note we should be very careful about what exactly we log where, right balance between noise and signal:

- not running on aws -> don't log anything (or only .Debug logs)
- running on aws in an ASG and instance protection succeeded -> a single logrus.Info log
- running on aws, but describing ASGs was empty/we're not running in an ASG -> a single logrus.Info
- running on aws, but describing ASGs failed due to permissions -> a single logrus.Warn
- running on aws and setting the instance protection failed -> a single logrus.Warn log with the error

cmd/osbuild-worker/main.go
```diff
@@ -104,6 +159,14 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 		return err
 	}
 
+	if isRunningOnAWS() {
+		logrus.Info("Protecting AWS instance")
+		err = setProtection(true)
+		if err != nil {
+			logrus.Warn("Can't set protection for instance")
```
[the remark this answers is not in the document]
lavocatt: I'll leave the logging inside the setProtection function.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +88,57 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+
+	// make the request to protect (or unprotect) the instance
+	input := &autoscaling.SetInstanceProtectionInput{
+		AutoScalingGroupName: asInstanceOutput.AutoScalingInstances[0].AutoScalingGroupName,
+		InstanceIds: []*string{
+			aws.String(identity.InstanceID),
+		},
+		ProtectedFromScaleIn: aws.Bool(protected),
+	}
+	result, err := svc.SetInstanceProtection(input)
```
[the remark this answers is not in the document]
lavocatt: I'll check if I can have a specific warning for the lack of permissions.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +88,68 @@ func WatchJob(ctx context.Context, job worker.Job) {
 	}
 }
 
+// checks if the instance is running on EC2, it has some potential for false positives.
+func isRunningOnAWS() bool {
```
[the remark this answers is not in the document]
lavocatt: I remove it. Good for me.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +88,57 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+
+	svc := autoscaling.New(awsSession)
+
+	// get the autoscaling group info for the auto scaling group name
+	asInstanceInput := &autoscaling.DescribeAutoScalingInstancesInput{
+		InstanceIds: []*string{
+			aws.String(identity.InstanceID),
+		},
+	}
+	asInstanceOutput, err := svc.DescribeAutoScalingInstances(asInstanceInput)
```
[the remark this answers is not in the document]
Gundersanne: Yup, the instance profile is define in app-interface.

cmd/osbuild-worker/main.go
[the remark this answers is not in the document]
Gundersanne: Then a warning is fine

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,76 @@ func WatchJob(ctx context.Context, job worker.Job) {
 	}
 }
 
+// protect an AWS instance from scaling and/or terminating.
+func setProtection(protected bool) {
+	// create a new session
+	awsSession, err := session.NewSession()
+	if err != nil {
+		logrus.Debug("impossible to get an AWS session")
```
Gundersanne, line 98: Nitpick, but instead of "impossible" how about "Error (getting...)". Just to align more with the rest of the codebase.

Gundersanne (commented):
Nice, this is looking good imo. Small nitpick but other than that lgtm.

I'd like to wait until May 4th to merge this though, just in case we need any last-minute fixes to production before the freeze.

cmd/osbuild-worker/main.go
lavocatt: I kept the Debug level so that it does not pollute the log output.

lavocatt: fixed now.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,80 @@ func WatchJob(ctx context.Context, job worker.Job) {
 	}
 }
 
+// protect an AWS instance from scaling and/or terminating.
```
ondrejbudai, line 93: This needs a simple test to check whether this doesn't crash if not running on AWS. Just for the sake of sanity, we run a lot of workers outside of AWS.

ondrejbudai (commented):
I like this! My only concern is that we might want to make this opt-in (opt-out is also an option but imho a worse one).

As a person wanting to deploy his own workers, I'm not sure if I would be happy with a worker service just "randomly" flipping settings of my AWS instances. It feels like a violation of separation of concerns. It feels like the worker should have an explicit approval to do so. Thoughts?

cmd/osbuild-worker/main.go
lavocatt: Thanks for the review, in order for it to work an instance needs to be not only on AWS but inside an autoscaling group and to have the access rights to describe its autoscaling group. If these conditions are not fully met, then the process of protecting will silently fail. We keep it silent for the reason that the behavior shall not change at all on non AWS workers. The fact that it passes CI proves that it works outside the required configuration to lock an instance.

And I realize that this bit of text should be in the PR and commit description :sweat_smile:

Gundersanne: We'd see this in CI already no? Like workers would crash immediately.

lavocatt: Yes the fact that this passes CI pretty much shows that it's safe.

teg:
This is nice! Two comments:
- I wouldn't worry too much about being shut down between picking up a job and protecting the instance, we should reschedule lost jobs instead.
- there is still a race between protect and unprotect: if we finish the last job (so unprotecting) just before a new one is started (so protecting) the order could be flipped because we don't lock around the (un)protect calls. We could simplify this by only protect build jobs and relying on (future) retry to handle interrupted depsolve. Then we don't need to sync

ondrejbudai:
I ran the code locally (without aws credentials so it is no-op) and the following line

```
identity, err := ec2metadata.New(awsSession).GetInstanceIdentityDocument()
```

runs for 6 seconds. As this is also used for protecting depsolve jobs, it makes them much slower. One depsolve takes currently about 1 second. Adding two `setProtection` calls adds 12 seconds which is suboptimal.

So, in the end, I agree with Tom's idea not to protect depsolves since the protection may take more time than the depsolve itself, which can actually raise the chance for the worker to be terminated.

This is actually nice because we shouldn't need any locking mechanism if we just protect build jobs.

cmd/osbuild-worker/main.go
ondrejbudai: Fair enough, I still think that a small unit test would be beneficial as it's much simpler to run one and it feels safer but I will leave it up to you.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,80 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+		InstanceIds: []*string{
+			aws.String(identity.InstanceID),
+		},
+	}
+	asInstanceOutput, err := svc.DescribeAutoScalingInstances(asInstanceInput)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			logrus.Warningf("Error getting the Autoscaling instances: %s %s", aerr.Code(), aerr.Error())
+		} else {
+			logrus.Error("Error getting the Autoscaling instances: unknown")
```
ondrejbudai, line 122: We should also print `err` here, otherwise we have no clue what actually failed.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,80 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+			aws.String(identity.InstanceID),
+		},
+		ProtectedFromScaleIn: aws.Bool(protected),
+	}
+	_, err = svc.SetInstanceProtection(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			logrus.Warningf("Error protecting instance: %s %s", aerr.Code(), aerr.Error())
+		} else {
+			logrus.Error("Error protecting instance: unknown")
```
ondrejbudai, line 140: Same as above.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,80 @@ func WatchJob(ctx context.Context, job worker.Job) {
 	}
 }
 
+// protect an AWS instance from scaling and/or terminating.
+func setProtection(protected bool) {
+	// create a new session
+	awsSession, err := session.NewSession()
+	if err != nil {
+		logrus.Debug("Error getting an AWS session")
```
ondrejbudai, line 98: I would prefer to include `err` in the logs.

cmd/osbuild-worker/main.go
```diff
@@ -84,6 +90,80 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+		return
+	}
+
+	// get the identity for the instanceID
+	identity, err := ec2metadata.New(awsSession).GetInstanceIdentityDocument()
+	if err != nil {
+		logrus.Debug("Error getting the identity document")
```
ondrejbudai, line 105: Same here.

cmd/osbuild-worker/main.go
ondrejbudai: (I'm actually not sure if we actually need to distinguish between the two error types, I had a brief look and it seems like `aerr.Error()` always includes `aerr.Code()`. Up to you, no hard feelings.

Gundersanne:
> This is nice! Two comments:
> 
>     * I wouldn't worry too much about being shut down between picking up a job and protecting the instance, we should reschedule lost jobs instead.
> 
>     * there is still a race between protect and unprotect: if we finish the last job (so unprotecting) just before a new one is started (so protecting) the order could be flipped because we don't lock around the (un)protect calls. We could simplify this by only protect build jobs and relying on (future) retry to handle interrupted depsolve. Then we don't need to sync

Ah I missed the potential reordering. That said if we lock both depsolve and osbuild jobs we can never make it truly racefree I think as the sdk just makes http calls to a rest api and I'm not sure order is guaranteed there in any way.

Didn't realize the instance identity call outside of aws took so long (which makes sense I guess since it's an http call which times out, and all calls are retried by default). We still need the instance identity to lock the instance, so this means we probably need to make it explicit.

lavocatt:
Thanks for the review!
>     * there is still a race between protect and unprotect: if we finish the last job (so unprotecting) just before a new one is started (so protecting) the order could be flipped because we don't lock around the (un)protect calls. We could simplify this by only protect build jobs and relying on (future) retry to handle interrupted depsolve. Then we don't need to sync

Ohh right, I haven't seen this one, a lock would probably be better

ondrejbudai (approved):

ondrejbudai:
@Gundersanne, can you have your :eye: on this?

Gundersanne (approved):
LGTM! Thank you

PR #2568, opened 2022-04-20
📦🔗📦  Introduce chain dependency solving

thozza:
- Introduce a new `chain-depsolve` command to `dnf-json` for depsolving multiple package sets as chained transactions building on the previous ones.
- Support the new command internally via `rpmmdImpl.chainDepsolve()` method.
- Switch the internal implementation of `rpmmdImpl.Depsolve()` to use `rpmmdImpl.chainDepsolve()` internally. This means that the original `depsolve` `dnf-json` command is no longer used by any of the composer code.
- Extend `ImageType` with new method `PackageSetsChains()` which returns the package set chains for the package sets returned by the image type, as they should be depsolved together and in a specific order.
- Introduce `DepsolvePackageSets()` method to the `rpmmd.RPMMD` interface. It takes advantage of the new `chain-depsolve` `dnf-json` command and the package set chains defined by each image type. It returns a set of depsolved package specs, as they should be passed to the `ImageType.Manifest()` method.
- Switch to using `DepsolvePackageSets()` instead of the original `Depsolve()` method in all of the relevant places. Only some specific parts of Weldr are still using `Depsolve()` method. The reason is that Weldr does all kinds of depsolving for the purpose of presenting the result to the user. Using the new `DepsolvePackageSets()` method in these places would be too verbose and would add too little benefit. However the important part is that for any compose requests, Weldr uses the `DepsolvePackageSets()` method.
- Do not pass the `blueprint` package specs set separately to the pipeline methods in RHEL 85, 86 and 90. The package set is now depsolved together with the `osPkgSet`. `blueprint` package set was not depsolved separately in any other distro definitions.

Note: the depsolving approach in RHEL 83, 84 and Fedora was not touched. Blueprint packages are merged into the `osPkgSet` packages and depsolved together in a single transaction. I left out this from this PR intentionally to limit the scope and get something working for the new distro definitions in fast. Nevertheless the older distros should be eventually fixed as well.

This pull request includes:


thozza:
So I consider the PR to be good enough for preliminary review. I kept it as Draft, because I expect that I'll have to regenerate some image test cases for 85/86/90/90beta as a result of changed depsolving, but need CI results to figure out what changed.

internal/rpmmd/repository.go
```diff
@@ -652,6 +660,55 @@ func (r *rpmmdImpl) chainDepsolve(chains []chainPackageSet, repos []RepoConfig,
...
+// in a chain or alone, as defined based on the provided arguments.
+func (r *rpmmdImpl) DepsolvePackageSets(
+	packageSetsChains map[string][]string,
+	packageSets map[string]PackageSet,
+	repos []RepoConfig,
+	packageSetsRepos map[string][]RepoConfig,
+	modulePlatformID, arch, releasever string) (map[string][]PackageSpec, error) {
+
+	packageSpecsSets := make(map[string][]PackageSpec, len(packageSets))
+	// slice to hold package set names which were processed as chains
```
achilleas-k, line 675: `s/slice/map/` :)

internal/weldr/api.go
```diff
@@ -3123,35 +3131,42 @@ func (api *API) fetchPackageList(distroName string) (rpmmd.PackageList, error) {
 	return packages, err
 }
 
-// Returns all configured repositories (base, depending on the image type + sources) as rpmmd.RepoConfig
+// Returns only user-defined repositories, which should be used only for
+// payload package sets.
+func (api *API) payloadRepositories(distroName string) []rpmmd.RepoConfig {
+	payloadRepos := make([]rpmmd.RepoConfig, 0, len(api.store.GetAllDistroSources(distroName)))
+	for id, source := range api.store.GetAllDistroSources(distroName) {
```
achilleas-k, line 3138: A bit wasteful to be loading sources from the store just to get their length and then loading them again to iterate through.

achilleas-k (commented):
Looks good overall.  I have some minor comments in-line and some non-actionable thoughts that I felt like sharing for future considerations (some of which we discussed offline).

internal/weldr/api.go
thozza: 🙄  /me goes fixing it...

thozza: fixed.

internal/rpmmd/repository.go
thozza: fixed

internal/distro/rhel85/distro_test.go
```diff
@@ -812,3 +812,28 @@ func TestDistro_CustomUsrPartitionNotLargeEnough(t *testing.T) {
 		}
 	}
 }
+
+// Ensure that all package sets defined in the package set chains are defined for the image type
+func TestImageType_PackageSetsChains(t *testing.T) {
```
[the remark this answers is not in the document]
thozza: de-duplicated 😉

internal/rpmmd/repository.go
```diff
@@ -534,6 +543,155 @@ func (r *rpmmdImpl) Depsolve(packageSet PackageSet, repos []RepoConfig, modulePl
...
+// `packageSetsRepos` are used only for package sets with the respective name.
+//
+// The function returns a `ChainPackageSet` slice, which members are referencing
+// repositories from the returned `RepoConfig` by their index. The returned
+// `RepoConfig` slice is specific to the requested `packageSetsChain`.
+//
+// NOTE: Due to implementation limitations of DNF and dnf-json, each package set
+// in the chain must use all of the repositories used by its predecessor.
+// An error is returned if this requirement is not met.
+func chainPackageSets(packageSetsChain []string, packageSets map[string]PackageSet, repos []RepoConfig, packageSetsRepos map[string][]RepoConfig) ([]chainPackageSet, []RepoConfig, error) {
```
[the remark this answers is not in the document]
thozza: I agree with everything that you've wrote 😅

internal/rpmmd/repository.go
```diff
@@ -534,6 +543,155 @@ func (r *rpmmdImpl) Depsolve(packageSet PackageSet, repos []RepoConfig, modulePl
...
+
+		// Add any package-set-specific repos to the list of transaction repos
+		if pkgSetRepos, ok := packageSetsRepos[pkgSetName]; ok {
+			for _, pkgSetRepo := range pkgSetRepos {
+				// Check if the repo has been already used by a transaction
+				// and if yes, just use its ID. Skip the "base" repos.
+				pkgSetRepoID := -1
+				for idx := len(repos); idx < len(transactionsRepos); idx++ {
+					transactionRepo := transactionsRepos[idx]
+					if reflect.DeepEqual(pkgSetRepo, transactionRepo) {
```
ondrejbudai, line 558: I always have mixed feelings about `reflect` but it's probably fine here.

internal/rpmmd/repository.go
```diff
@@ -534,6 +543,155 @@ func (r *rpmmdImpl) Depsolve(packageSet PackageSet, repos []RepoConfig, modulePl
...
+
+		// If more than one transaction, ensure that the transaction uses
+		// all of the repos from its predecessor
+		if transactionIdx > 0 {
+			previousTransRepos := transactions[transactionIdx-1].Repos
+			if len(transaction.Repos) < len(previousTransRepos) {
+				return nil, nil, fmt.Errorf("chained packageSet %q does not use all of the repos used by its predecessor", pkgSetName)
+			}
+
+			for idx, repoID := range previousTransRepos {
```
ondrejbudai, line 584: This is basically an equality test for two subslices, might be cleaner to write it in that way (another `reflect.Deep`)

internal/rpmmd/repository.go
```diff
@@ -534,6 +543,155 @@ func (r *rpmmdImpl) Depsolve(packageSet PackageSet, repos []RepoConfig, modulePl
...
+		dnfRepo, err := repo.toDNFRepoConfig(r, i, arch, releasever)
+		if err != nil {
+			return nil, nil, err
+		}
+		dnfRepoConfigs = append(dnfRepoConfigs, dnfRepo)
+	}
+
+	type dnfTransaction struct {
+		PackageSpecs []string `json:"package-specs"`
+		ExcludSpecs  []string `json:"exclude-specs"`
```
ondrejbudai, line 612: Typo I think

dnf-json
```diff
@@ -247,6 +247,59 @@ class Solver():
...
+            if idx > 0:
+                self.base.conf.install_weak_deps=False
+
+            # set the packages from the last transaction as installed
+            for installed_pkg in last_transaction:
+                self.base.package_install(installed_pkg, strict=True)
+
+            # depsolve the current transaction
+            self.base.install_specs(
+                transaction.get("package-specs"),
```
ondrejbudai, line 267: I think that `get` isn't actually needed on these 3 rows because Go will always marshal these fields (there's no omitempty) so by using `get` we might actually hide a malformed json schema.

thozza:
CI found a bug in the `DepsolvePackageSets()` implementation (https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2389312163). The change was

thozza:
CI found a bug in the DepsolvePackageSets() implementation (https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2389312163). The change was:
```diff
diff --git a/internal/rpmmd/repository.go b/internal/rpmmd/repository.go
index 0da4e529..9ef18bca 100644
--- a/internal/rpmmd/repository.go
+++ b/internal/rpmmd/repository.go
@@ -695,7 +695,7 @@ func (r *rpmmdImpl) DepsolvePackageSets(
                if _, ok := chainPkgSets[name]; ok {
                        continue
                }
-               chainPkgSets, chainRepos, err := chainPackageSets([]string{name}, map[string]PackageSet{name: packageSet}, repos, nil)
+               chainPkgSets, chainRepos, err := chainPackageSets([]string{name}, map[string]PackageSet{name: packageSet}, repos, packageSetsRepos)
                if err != nil {
                        return nil, err
                }
```

The issue was that Fedora still combines BP package set with the OS package set before depsolving. The previous `DepsolvePackageSets()` implementation was not applying package-set-specific repositories for package sets not depsolved in a chain (my original idea was that this will be always the case as the user-defined repos should be applied only for Blueprint packages). However this is not the case for Fedora, RHEL-83/84. And this assumption was probably wrong also in general, since image-builder can potentially define package-set specific repositories also for non-payload sets.

internal/rpmmd/repository.go
```diff
@@ -212,7 +212,15 @@ type RPMMD interface {
 	// Depsolve takes a list of required content (specs), explicitly unwanted content (excludeSpecs), list
 	// or repositories, and platform ID for modularity. It returns a list of all packages (with solved
 	// dependencies) that will be installed into the system.
+	//
+	// NOTE: Kept for legacy purposes (Weldr). New code should default to using DepsolvePackageSets().
```
ondrejbudai, line 216: I think that we should use `Deprecated:` instead of `NOTE:`, see https://github.com/golang/go/wiki/Deprecated

This will probably lead to the linter blowing up in Weldr so we will need to add some ignore lines but if will very effectively prevent people from using the old function.

internal/rpmmd/repository.go
```diff
@@ -652,6 +660,55 @@ func (r *rpmmdImpl) chainDepsolve(chains []chainPackageSet, repos []RepoConfig,
 	return dependencies, reply.Checksums, err
 }
 
+// DepsolvePackageSets takes a map of package sets chains, which should be depsolved as separate transactions,
+// a map of package sets (included and excluded), a list of common and package-set-specific repositories,
+// platform ID for modularity, architecture and release version. It returns a map of Package Specs, depsolved
+// in a chain or alone, as defined based on the provided arguments.
+func (r *rpmmdImpl) DepsolvePackageSets(
+	packageSetsChains map[string][]string,
```
ondrejbudai, line 668: I think that this kind of methods taking 7 weird arguments should just take 1 struct but that's certainly something for a followup.

internal/rpmmd/repository.go
thozza: The other option would be to implement a logic which would do the deep equality check. Given the documentation for `reflect.DeepEqual()` I decided that it basically does what I needed in this case and that it should be safe to use it to compare two `RepoConfig` instances.

But I'm happy to rework this if you have any suggestions 😉

internal/rpmmd/repository.go
thozza: Yes, it is an equality check of a slice (previous transaction) and subslice (current transaction capped at the length of the previous transaction).

Happy to change it if you feel like using more of `reflect` 😄 .

I didn't use it in this case as implementing this check was considerably simpler than deep equality of two structures.

internal/rpmmd/repository.go
thozza: I was not 100% if we really want to deprecate this method, but your suggestion makes sense.

internal/rpmmd/repository.go
ondrejbudai: Yeah, agreed, it's actually very sane (if you have sane structures).

internal/rpmmd/repository.go
ondrejbudai: It just might be simpler to read but no hard feelings. :)

ondrejbudai (commented):
Very nice! :)

internal/rpmmd/repository.go
thozza: I understand your feeling, but I opted in for 7 weird arguments instead of 1 weird structure. But I agree that we should follow-up and polish the whole `RPMMD` interface.

achilleas-k: I think these kinds of changes can be folded into #2537

ondrejbudai (approved):

PR #2575, opened 2022-04-21
Don't run unsupported tests on 8.6 and 9.0

atodorov:
https://coreos.slack.com/archives/C0235DZB0DT/p1650536467098679?thread_ts=1650512166.141439&cid=C0235DZB0DT

This pull request includes:

### DIFF
--- a/test/cases/aws_s3.sh
+++ b/test/cases/aws_s3.sh
@@ -1,7 +1,19 @@
 #!/bin/bash
-set -euo pipefail
 
 source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+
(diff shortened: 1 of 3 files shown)


thozza:
@atodorov so the tricky part is that these tests are not supported only for the versions that are in RHEL repositories. But the `main` composer can run them successfully on 8.6 and 9.0. So we should just not run them in the nightly runs which do not use `osbuild-composer` from the git.

thozza (approved):
I wonder whether it would be possible to have the condition in `.gitlab-ci.yml` instead of the test case. But if not, I'm OK with this approach. 👍

PR #2576, opened 2022-04-21
worker: add proxy support to worker

ondrejbudai:
In the internal deployment, we want to talk with composer over a http/https proxy. This proxy adds new composer.proxy field to the worker config that causes the worker to connect to composer and the oauth server using a specified proxy.

NB: The proxy is not supported when connection to composer via unix sockets.

This PR also includes random improvements to the worker code but nothing major, just some things that I stumbled upon.

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -123,7 +123,7 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 
 func main() {
 	var config struct {
-		KojiServers map[string]struct {
+		Koji map[string]struct {
 			Kerberos *struct {
 				Principal string `toml:"principal"`
 				KeyTab    string `toml:"keytab"`
@@ -193,7 +193,7 @@ func main() {
 	_ = os.Mkdir(output, os.ModeDir)
 
 	kojiServers := make(map[string]koji.GSSAPICredentials)
-	for server, creds := range config.KojiServers {
+	for server, creds := range config.Koji {
 		if creds.Kerberos == nil {
 			// For now we only support Kerberos authentication.
 			continue
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -123,6 +123,9 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 
 func main() {
 	var config struct {
+		Composer *struct {
+			Proxy string `toml:"proxy"`
+		}
 		Koji map[string]struct {
 			Kerberos *struct {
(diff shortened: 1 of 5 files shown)


teg:
> NB: The proxy is only supported when OAuth authentication is enabled. If mTLS, or unix sockets are used, the proxy is currently ignored.

Could we make this an error instead? I think that would be less surprising.

teg (dismissed):
Very nice to see!

ondrejbudai:
Makes sense to make it less surprising, the proxy config option is now available also for the mTLS authentication. Does it work for you?

ondrejbudai:
@diaasami @teg This should be now ready, would love to get this in before Wednesday's release. :)

cmd/osbuild-worker/main.go
```diff
@@ -123,6 +123,9 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 
 func main() {
 	var config struct {
+		Composer *struct {
+			Proxy string `toml:"proxy"`
```
teg, line 127: Feels like we should configure the url here too (separately / in a follow-up). I'm not sure configuring that on the command line makes any sense, and having the proxy in the config file draws attention to it, IMHO.

cmd/osbuild-worker/main.go
```diff
@@ -266,10 +275,16 @@ func main() {
 			}
 		}
 
+		proxy := ""
+		if config.Composer != nil && config.Composer.Proxy != "" {
+			proxy = config.Composer.Proxy
```
teg, line 280: This could be done outside the conditional, and we could parse the proxy into an URL here rather than in the client constructor. That way catching syntax errors in `cmd/`, which is where I think it belongs.

teg (approved):
Minor style comments, but otherwise looking great!

cmd/osbuild-worker/main.go
ondrejbudai: Indeed, in the beginning, we just used the composer's URL from the systemd template unit with hardcoded path to the certificates but as we expanded the functionality of the worker, at one point we added a proper config. So I guess the correct way forward is to switch to a config completely, something for a follow-up I think.

cmd/osbuild-worker/main.go
ondrejbudai: Agreed, the worker code would really benefit from some refactoring, let's thinking about it in the future?

teg: Let's!

cmd/osbuild-worker/main.go
teg: Agreed.

PR #2578, opened 2022-04-21
Use array of objects to maintain order for RPM stage inputs

achilleas-k:
The composer counterpart to https://github.com/osbuild/osbuild/pull/1003
- [x] Manifest and image-info update

The osbuild inputs (and the osbuild2 package in general) could use some work and there are more _proper_ ways to define these inputs to match the osbuild schemata, but this fixes a critical issue so I'm simply replacing the existing supported format for rpm stage inputs.  Future work could add support for the other formats if necessary.


achilleas-k:
I'm generating the manifests and image info now and with a rebase on #2552, so we should merge that one first.

achilleas-k:
Manifests done and based on #2552

internal/osbuild2/rpm_stage.go
```diff
@@ -70,10 +80,38 @@ type RPMStageReference struct {
 //    }
 // }
 // ...
-type RPMStageReferences map[string]*RPMStageReference
+type RPMStageReferences map[string]*RPMStageSourceOptions
```
thozza, line 83: nitpick: IMO it would make sense to rename `RPMStageReferences` to something like `RPMStageSourceRefs` or even `RPMStageSourceMapRefs`, similar to `RPMStageSourceArrayRefs`.

internal/osbuild2/rpm_stage.go
```diff
@@ -33,15 +33,23 @@ type RPMPackage struct {
...
 func (RPMStageInputs) isStageInputs() {}
 
+// RPMStageInput defines a single input source.
 type RPMStageInput struct {
 	inputCommon
-	References RPMStageReferences `json:"references"`
+
+	// Collection of references. Each reference defines a package to be
+	// installed, with optional metadata.
+	References RPMStageSourceArrayRefs `json:"references"`
```
thozza, line 52: Nitpick: You could probably use the `References` interface as a type, but it may actually complicate code a bit. This would allow the structure to be agnostic of the specific Reference implementation.

test/data/manifests/rhel_84-ppc64le-qcow2_customize-boot.json
```diff
@@ -11350,7 +11350,6 @@
         "/proc": ".M.......",
         "/run/cockpit": ".M.......",
         "/sys": ".M.......",
-        "/var/lib/powerpc-utils/smt.state": "..5....T.",
```
thozza, line 11353: hmm, this change looks a bit weird 🤔

test/data/manifests/rhel_84-ppc64le-qcow2-boot.json
```diff
@@ -11231,7 +11231,6 @@
         "/proc": ".M.......",
         "/run/cockpit": ".M.......",
         "/sys": ".M.......",
-        "/var/lib/powerpc-utils/smt.state": "..5....T.",
```
thozza, line 11234: hmm, this change looks a bit weird 🤔

thozza (approved):
❤️  Great work, thanks a lot!

Added a few nitpicks, nothing serious...

internal/osbuild2/rpm_stage.go
achilleas-k: Similar to my other comment, yes it would be good to rename a lot of these to make more sense, and even define them elsewhere.  I considered removing this type altogether because it's not used at all now.  I kept it around so that we can clean it up in a followup redesign.

test/data/manifests/rhel_84-ppc64le-qcow2-boot.json
achilleas-k: Yeah, no idea what's going on there.  Is it better that it's not modified?  Or is this one of the files that's expected to be modified?  Is it possible the install order affects this?  I have no idea.

gicmo: the `smt.state` is something host specific so it is definitely better if it is not in the image; the question is, why was it there before? I guess we leak build host state into the image.

gicmo: i.e.:
```
/usr/sbin/smtstate: script is used to persist the current SMT state and to restore it at a convenient time. The idea behind it is to use the script together with a systemd service to save SMT state at any shutdown event and to restore it at an early stage of the boot process. Such implementation creates a way to persist the SMT state across reboots, without requiring any other configuration from system administrators.
```

achilleas-k: Seems that the file failing `rpm --verify` was a bug
https://bugzilla.redhat.com/show_bug.cgi?id=2014916
Though it's weird that it changed here since the package didn't change in the image.  I think you might be right @gicmo that it could be leaking from the host.

gicmo: I think we should consider removing that file from the image if it was created, since the state of the build host's really should not be in the image and maybe even have an image-info test to report if it is in the image or not.

achilleas-k:
Some 8.5 manifest tests failed.  I think it might have been a mixup from the revert of the rebase onto #2552.
Fixing them up.

test/data/manifests/rhel_84-ppc64le-qcow2-boot.json
thozza: > Seems that the file failing `rpm --verify` was a bug https://bugzilla.redhat.com/show_bug.cgi?id=2014916 Though it's weird that it changed here since the package didn't change in the image. I think you might be right @gicmo that it could be leaking from the host.

The reason for it disappearing is most probably the fact that we are running `rpm` from the host when generating image-info report.

internal/osbuild2/rpm_stage.go
thozza: Fair enough... 😉

internal/osbuild2/rpm_stage.go
thozza: Fair enough as well... 😉

achilleas-k:
All done.  Should be good now.  I hope.

achilleas-k:
Well the manifests test is green so we got that going for us.

PR #2584, opened 2022-04-25
Gitlab CI: define nightly report dependencies

jrusz:
We don't need to download all the artifacts, just the Composer version
and compose ID.

This is related to https://github.com/osbuild/osbuild-composer/pull/2569 ... My initial change broke the reporting, this should now get only the relevant files in the finish stage.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -559,6 +559,9 @@ Upgrade:
 
 NIGHTLY_FAIL:
   stage: finish
+  dependencies:
+    - COMPOSE_ID
+    - COMPOSER_NVR
   tags:
     - shell
   rules:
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Not sure if this will work, from [docs](https://docs.gitlab.com/ee/ci/yaml/#dependencies):

> Use the dependencies keyword to define a list of jobs to fetch [artifacts](https://docs.gitlab.com/ee/ci/yaml/#artifacts) from. You can also set a job to download no artifacts at all.

jrusz:
Yeah, I would just have to put in specific job names from which to download the artifacts which could change and would just give more room for error, I don't like that. Let's just keep it as is, it's running once per day and there are not as many jobs as on PRs so it's fine I guess....

PR #2586, opened 2022-04-25
cloudapi/v2: Generate valid GCP image name

Gundersanne:


internal/cloudapi/v2/handler.go
```diff
@@ -376,21 +376,21 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 					share = *gcpUploadOptions.ShareWithAccounts
 				}
 
-				// the uploaded object must have a valid extension
-				object := fmt.Sprintf("composer-api-%s.tar.gz", uuid.New().String())
+				objectName := fmt.Sprintf("composer-api-%s", uuid.New().String())
```
thozza, line 379: Lets call this `imageName` instead, since the string is not used as is for object name

test/cases/api.sh
```diff
@@ -1386,7 +1376,6 @@ function verifyInGCP() {
 
   # create the instance
   # resource ID can have max 62 characters, the $GCP_TEST_ID_HASH contains 56 characters
-  GCP_INSTANCE_NAME="vm-$GCP_TEST_ID_HASH"
```
thozza, line 1389: This should no have been deleted, since the value is used below for creating a VM. The name is intentionally predictable to allow for cleanup in case of pass or fail.

test/cases/api.sh
```diff
@@ -683,10 +683,6 @@ function createReqFileGCP() {
   # - max 62 characters
   # - must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}'
   #
-  # use sha224sum to get predictable 56 characters long testID without invalid characters
-  GCP_TEST_ID_HASH="$(echo -n "$TEST_ID" | sha224sum - | sed -E 's/([a-z0-9])\s+-/\1/')"
```
thozza, line 687: These lines should not have been deleted, since the the value is used for other resource names, besides the image name.

thozza (changes requested):
In principle this looks good, but there are some lines deleted, which look like a bug to me.

In addition, not using the predictable image name may have impact on the `cloud-cleaner`'s ability to delete leftover resources in case of failed test case, because it constructs the resource names using the same algorithm as the test case.

thozza (dismissed):
LGTM, Thanks for fixing my bugs 😅

ondrejbudai:
Just curious: What was wrong before and how did the tests pass?

thozza:
> Just curious: What was wrong before and how did the tests pass?

image-builder is not providing the image name in the compose request, but we do provide it in `api.sh` to ensure that a predictable image name is used. @Gundersanne found a bug that if the image name is not provided in the compose request, the code logic would use the object name as the image name. The object name must end with `.tar.gz` to be successfully imported as an image, however such name is not valid for an image name. Thus in this scenario, the image import would fail.

Gundersanne:
Should probably put that in the commit huh :D

thozza (approved):

ochosi:
@Gundersanne Would be great if you could improve the title of this PR - after all that's what directly ends up in our release notes/changelog.

Gundersanne:
Admin merging, someone is logged into the shell runner and the pipeline succeeded.

PR #2590, opened 2022-04-26
Schutzfile: Indicate whether a snapshot is updateable

jrusz:
Some repositories like the fedora and fedora-modular don't change after
release and so the snapshots don't change either. Having an updateable
flag to indicate this simplifies automated updating of other snapshots.

This pull request includes:

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -21,7 +21,8 @@
             "name": "fedora",
             "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f34/f34-aarch64-fedora-20210512"
           }
-        ]
+        ],
+        "updateable": false
       },
       {
         "file": "/etc/yum.repos.d/fedora-modular.repo",
@@ -38,7 +39,8 @@
             "name": "fedora-modular",
             "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f34/f34-aarch64-fedora-modular-20210512"
           }
-        ]
+        ],
+        "updateable": false
       },
       {
         "file": "/etc/yum.repos.d/fedora-updates.repo",
@@ -55,7 +57,8 @@
             "name": "updates",
             "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f34/f34-aarch64-updates-released-20220330"
           }
-        ]
+        ],
+        "updateable": true
       },
       {
         "file": "/etc/yum.repos.d/fedora-updates-modular.repo",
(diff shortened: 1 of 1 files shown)


jrusz:
See https://github.com/osbuild/rpmrepo/pull/43 for context

Gundersanne (dismissed):
i'm ok with this approach if @ondrejbudai is. It makes the schutzfile a little bigger but it's the most explicit way to do this.

ondrejbudai:
I just wonder if this is a property that should live here. rpmrepo's repo definitions already have the information whether they are singleton, or not, so we can just query it when sending the schutzfile updating PRs after a snapshot is done. There are two benefits with this approach:

- we don't have to define "updatability" in each repository with Schutzfile
- we have just one single source of truth (singleton flag in rpmrepo)

Gundersanne:
> I just wonder if this is a property that should live here. rpmrepo's repo definitions already have the information whether they are singleton, or not, so we can just query it while sending the schutzfile updating PRs after a snapshot is done. There are two benefits with this approach:
> 
>     * we don't have to define "updatability" in each repository with Schutzfile
> 
>     * we have just one single source of truth (singleton flag in rpmrepo)

Yea per-distro is probably never needed. However being able to choose an osbuild version on a PR could be useful to testing PRs in osbuild against composer (once the test move out of composer this will become much less useful of course).

Gundersanne:
> I just wonder if this is a property that should live here. rpmrepo's repo definitions already have the information whether they are singleton, or not, so we can just query it while sending the schutzfile updating PRs after a snapshot is done. There are two benefits with this approach:
> 
>     * we don't have to define "updatability" in each repository with Schutzfile
> 
>     * we have just one single source of truth (singleton flag in rpmrepo)

Much more elegant solution ; +1 on that

jrusz:
> I just wonder if this is a property that should live here. rpmrepo's repo definitions already have the information whether they are singleton, or not, so we can just query it while sending the schutzfile updating PRs after a snapshot is done. There are two benefits with this approach:
> 
> * we don't have to define "updatability" in each repository with Schutzfile
> * we have just one single source of truth (singleton flag in rpmrepo)

Oook. So the way I can image this to work would be once snapshots are done, take the Schutzfile, iterate over each repository we have defined there and look for the corresponding repository definition in the `rpmrepo` repository and if it has a `singleton` defined, skip it and don't update it... This will make it a bit more complicated but it does have those two benefits you mention.

ondrejbudai:
Sure but a small Python script should be able to do this without many issues. I think that we won't need any changes in this repository, can we close this PR?

PR #2597, opened 2022-04-28
Schutzfile: Pin osbuild version to use minimal required caps

Gundersanne:


ondrejbudai (approved):

ondrejbudai:
We don't have all RPMs built because of failed mockbuilds jobs in osbuild, I restarted them, then we should be able to restart jobs in this PR.

ondrejbudai:
RPMS built, restarting.

teg:
Didn't get any spot instances, retried, seems to work now.

PR #2598, opened 2022-04-28
composer: add support for logrus syslog hook

diaasami:
Which will be used on crc in the log forwarding setup
NeededBy: COMPOSER-1285

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/config.go
+++ b/cmd/osbuild-composer/config.go
@@ -11,11 +11,12 @@ import (
 )
 
 type ComposerConfigFile struct {
-	Koji      KojiAPIConfig   `toml:"koji"`
-	Worker    WorkerAPIConfig `toml:"worker"`
-	WeldrAPI  WeldrAPIConfig  `toml:"weldr_api"`
-	LogLevel  string          `toml:"log_level"`
-	LogFormat string          `toml:"log_format"`
+	Koji         KojiAPIConfig   `toml:"koji"`
+	Worker       WorkerAPIConfig `toml:"worker"`
+	WeldrAPI     WeldrAPIConfig  `toml:"weldr_api"`
+	SyslogServer string          `toml:"syslog_server" env:"SYSLOG_SERVER"`
+	LogLevel     string          `toml:"log_level"`
+	LogFormat    string          `toml:"log_format"`
 }
 
 type KojiAPIConfig struct {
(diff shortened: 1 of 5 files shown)


cmd/osbuild-composer/config.go
```diff
@@ -11,11 +11,12 @@ import (
...
 type ComposerConfigFile struct {
-	Koji      KojiAPIConfig   `toml:"koji"`
-	Worker    WorkerAPIConfig `toml:"worker"`
-	WeldrAPI  WeldrAPIConfig  `toml:"weldr_api"`
-	LogLevel  string          `toml:"log_level"`
-	LogFormat string          `toml:"log_format"`
+	Koji         KojiAPIConfig   `toml:"koji"`
+	Worker       WorkerAPIConfig `toml:"worker"`
+	WeldrAPI     WeldrAPIConfig  `toml:"weldr_api"`
+	SyslogServer string          `toml:"syslog_server" env:"SYSLOGSRVR"`
```
ondrejbudai, line 17: I would probably go with `SYSLOG_SERVER`, I don't think we need to save bytes,

cmd/osbuild-composer/main.go
```diff
@@ -69,6 +70,15 @@ func main() {
 		logrus.Fatal("CACHE_DIRECTORY is not set. Is the service file missing CacheDirectory=?")
 	}
 
+	if len(config.SyslogServer) > 0 {
+		hook, err := syslog.NewSyslogHook("tcp", config.SyslogServer, 0, "osbuild-composer")
+		if err != nil {
+			panic(err)
```
ondrejbudai, line 76: `logrus.Fatal` for the sake of consistency?

ondrejbudai (commented):
Can you please rebase this to get rid of the merge commit? Otherwise, this looks nice, thank you. Btw, I was thinking of writing a small test for this but I don't see an obvious way unfortunately.

cmd/osbuild-composer/config.go
diaasami: I agree `SYSLOG_SERVER` is better, I made it this way for consistency with other env vars like `PGUSER` and `PGPASSWORD`

Gundersanne: PG* etc. are standard postgres variables https://www.postgresql.org/docs/current/libpq-envars.html, so I just used the colloquial ones there.

ondrejbudai (approved):

Gundersanne (approved):

PR #2605, opened 2022-05-01
entrypoint - add parameters for socket bind address and port

ygalblum:
entrypoint.py
-------------
Allow the user to set the port number also for the remote worker
Allow the user to restrict the composer api or remote worker api to
localhost

Dockerfile-ubi
--------------
Allow the user to override the entrypoint parameters
Set the environment variable default value to keep current behavior

This pull request includes:

### DIFF
--- a/containers/osbuild-composer/entrypoint.py
+++ b/containers/osbuild-composer/entrypoint.py
@@ -50,9 +50,16 @@ class Cli(contextlib.AbstractContextManager):
         self._parser.add_argument(
             "--composer-api-port",
             type=int,
-            default=443,
+            default=8080,
             dest="composer_api_port",
             help="Port which the composer-API listens on",
+        )
+        self._parser.add_argument(
+            "--composer-api-bind-address",
+            type=str,
+            default="::",
+            dest="composer_api_bind_address",
+            help="Bind the composer API to the specified address",
         )
 
         # --[no-]local-worker-api
@@ -81,6 +88,20 @@ class Cli(contextlib.AbstractContextManager):
             action="store_false",
             dest="remote_worker_api",
             help="Disable the remote-worker-API",
+        )
+        self._parser.add_argument(
+            "--remote-worker-api-port",
+            type=int,
+            default=8700,
+            dest="remote_worker_api_port",
+            help="Port which the remote-worker API listens on",
+        )
+        self._parser.add_argument(
+            "--remote-worker-api-bind-address",
+            type=str,
+            default="::",
+            dest="remote_worker_api_bind_address",
+            help="Bind the remote worker API to the specified address",
         )
 
         # --[no-]weldr-api
@@ -171,7 +192,7 @@ class Cli(contextlib.AbstractContextManager):
             self._exitstack.enter_context(contextlib.closing(sock))
             sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
             sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
-            sock.bind(("::", self.args.composer_api_port))
+            sock.bind((self.args.composer_api_bind_address, self.args.composer_api_port))
             sock.listen()
             sockets.append(sock)
             names.append("osbuild-composer-api.socket")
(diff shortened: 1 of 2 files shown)


distribution/Dockerfile-ubi
```diff
@@ -23,4 +23,6 @@ COPY ./internal/jobqueue/dbjobqueue/schemas /opt/migrate/schemas
 COPY --from=builder2 /opt/app-root/src/go/bin/tern /opt/migrate/
 
 EXPOSE 8080 8700
-ENTRYPOINT ["python3", "/opt/entrypoint.py", "--remote-worker-api", "--composer-api", "--composer-api-port", "8080"]
+ENV ENTRYPOINT_ADDITIONAL_ARGS "--composer-api-port 8080"
```
ondrejbudai, line 26: We can even make `8080` the default value of `--composer-api-port`, I think that's a cleaner way forward as we want to drop TLS from composer.

containers/osbuild-composer/entrypoint.py
```diff
@@ -82,6 +88,19 @@ def _parse_args(self):
...
             help="Disable the remote-worker-API",
         )
+        self._parser.add_argument(
+            "--remote-worker-api-port",
+            type=int,
+            default=8700,
+            dest="remote_worker_api_port",
+            help="Port which the remote-worker API listens on",
+        )
+        self._parser.add_argument(
```
ondrejbudai, line 98: What about making this more generic? Basically accepting `address` as an argument, set to `::` by default.

ondrejbudai (commented):
Thanks a lot for this, looks good, two comments inline worth discussion I think. :)

distribution/Dockerfile-ubi
teg: Shouldn't the default port be 80?

teg: (it is possible to use one container for both composer and worker API, but we expect these to be run separately, in which case it doesn't matter if their default ports are the same)

teg: (it is possible to use one container for both composer and worker API, but we expect these to be run separately, in which case it doesn't matter if their default ports are the same)

ygalblum: The entrypoint defined in this Dockerfile starts both services, so it must use separate ports. 
Furthermore, since we allow the user to start both services together, I think it's better to have different default ports so that the user is not forced to define the port numbers.

ygalblum: I've implemented it according to @ondrejbudai 's suggestion. The only reason I didn't do it to begin with was not to break previous behavior

containers/osbuild-composer/entrypoint.py
ygalblum: Done

distribution/Dockerfile-ubi
ondrejbudai: > Shouldn't the default port be 80?

OpenShift (our primary target) doesn't like running stuff on privileged ports so I think that 8080 is fine.

> (it is possible to use one container for both composer and worker API, but we expect these to be run separately, in which case it doesn't matter if their default ports are the same)

I think that we wanted to consider this in the future but it was never implemented so I wouldn't say that running them separately is expected just yet. Composer in AOC is still running as one container.

distribution/Dockerfile-ubi
```diff
@@ -23,4 +23,6 @@ COPY ./internal/jobqueue/dbjobqueue/schemas /opt/migrate/schemas
 COPY --from=builder2 /opt/app-root/src/go/bin/tern /opt/migrate/
 
 EXPOSE 8080 8700
-ENTRYPOINT ["python3", "/opt/entrypoint.py", "--remote-worker-api", "--composer-api", "--composer-api-port", "8080"]
+ENV ENTRYPOINT_ADDITIONAL_ARGS ""
```
ondrejbudai, line 26: I'm thinking of just removing this altogether and relying on the ability of k8s to override the entrypoint. https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#entrypoint

Ideas? Feels cleaner to me.

containers/osbuild-composer/entrypoint.py
```diff
@@ -194,12 +215,12 @@ def _prepare_sockets(self):
 
         # osbuild-remote-worker.socket
         if self.args.remote_worker_api:
-            print("Create remote-worker-api socket", file=sys.stderr)
+            print("Create remote-worker-api socket on address {}:{}".format(self.args.remote_worker_api_bind_address, self.args.remote_worker_api_port), file=sys.stderr)
```
ondrejbudai, line 218: We prefer f-strings in osbuild project, can you please adjust it?

ondrejbudai: Also, can you put the address in brackers since it's an IPv6 one? https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers

distribution/Dockerfile-ubi
ygalblum: You mean removing `ENTRYPOINT_ADDITIONAL_ARGS` right? Meaning keep the default entrypoint of 
`ENTRYPOINT ["python3", "/opt/entrypoint.py", "--remote-worker-api", "--composer-api"]`

containers/osbuild-composer/entrypoint.py
ygalblum: Done and Done

ondrejbudai (approved):

ondrejbudai:
Thanks!

PR #2609, opened 2022-05-03
Pin the `staticfiles` Go dep in devcontainer.

supakeen:
The staticfiles dependency updated to support Go 1.18 generics in the
2022.1 release. This prevents the devcontainer (which has Go 1.16) from
being built. This pins to the latest release of staticfiles which builds
on 1.16.

Another option would be to bump Go to 1.18 in the devcontainer which would
allow for the dep to be installed `@latest`.

As this is a developer container I'm unsure how the checklist below
applies.

This pull request includes:

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:35
+FROM fedora:36
 
 # Install:
 #  - a few packages for convenient usage


ondrejbudai:
Idea: Fedora 36 ships with Go 1.18, we can switch to that (the GA is very near :crossed_fingers:, I'm fine with switching to it now).

Note that `osbuild-composer` still needs to be buildable by Go 1.16 until Fedora 35 is EOL (Tue 2022-11-15) but for development, I don't mind using the latest and greatest.

supakeen:
@ondrejbudai That seems like a good option, I've force-pushed the change to fedora:36 instead, the containers luckily still work afterwards. Are these included in tests?

ondrejbudai:
Nope, they aren't. FTR, we check if the code works on Fedora 35 (and thus Go 1.16) in multiple tests, so it's very hard to sneak incompatible code in.

ondrejbudai (approved):
Thanks for this! It's always lovely when dev-oriented tools just works. :-)

(admin-merging this, `.devcontainer` isn't covered)

PR #2618, opened 2022-05-05
prometheus: add tenant label

chloenayon:
This pull request includes:

### DIFF
--- a/internal/prometheus/job_metrics.go
+++ b/internal/prometheus/job_metrics.go
@@ -16,7 +16,7 @@ var (
 		Namespace: namespace,
 		Subsystem: workerSubsystem,
 		Help:      "Total jobs",
-	}, []string{"type", "status"})
+	}, []string{"type", "status", "tenant"})
 )
 
 var (
@@ -25,7 +25,7 @@ var (
 		Namespace: namespace,
 		Subsystem: workerSubsystem,
 		Help:      "Currently pending jobs",
-	}, []string{"type"})
+	}, []string{"type", "tenant"})
 )
 
 var (
@@ -34,7 +34,7 @@ var (
 		Namespace: namespace,
 		Subsystem: workerSubsystem,
 		Help:      "Currently running jobs",
-	}, []string{"type"})
+	}, []string{"type", "tenant"})
 )
 
 var (
@@ -44,7 +44,7 @@ var (
 		Subsystem: workerSubsystem,
 		Help:      "Duration spent by workers on a job.",
 		Buckets:   []float64{.1, .2, .5, 1, 2, 4, 8, 16, 32, 40, 48, 64, 96, 128, 160, 192, 224, 256, 320, 382, 448, 512, 640, 768, 896, 1024, 1280, 1536, 1792, 2049},
-	}, []string{"type", "status"})
+	}, []string{"type", "status", "tenant"})
 )
 
 var (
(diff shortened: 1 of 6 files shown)


internal/prometheus/job_metrics.go
```diff
@@ -54,35 +54,37 @@ var (
...
-		PendingJobs.WithLabelValues(jobType).Dec()
-		RunningJobs.WithLabelValues(jobType).Inc()
+		JobWaitDuration.WithLabelValues(jobType, tenant).Observe(diff)
+		PendingJobs.WithLabelValues(jobType, tenant).Dec()
+		RunningJobs.WithLabelValues(jobType, tenant).Inc()
 	}
 }
 
-func CancelJobMetrics(started time.Time, jobType string) {
+func CancelJobMetrics(started time.Time, jobType string, channel string) {
```
kingsleyzissou, line 74: This is a bit of a nitpick. It was a little confusing at first because in some cases we are passing in the `channel` and then getting the sub slice (like this case) and then in other cases we are passing the `tenant`.

Maybe, for consistency, we could take the tenant as an argument in all these functions?

internal/worker/server.go
```diff
@@ -133,7 +133,7 @@ func (s *Server) EnqueueManifestJobByID(job *ManifestJobByID, parent uuid.UUID,
 }
 
 func (s *Server) enqueue(jobType string, job interface{}, dependencies []uuid.UUID, channel string) (uuid.UUID, error) {
-	prometheus.EnqueueJobMetrics(jobType)
+	prometheus.EnqueueJobMetrics(jobType, channel)
```
kingsleyzissou, line 136: Here we could call the function like this:
`prometheus.EnqueueJobMetrics(jobType, channel[4:])`

internal/worker/server.go
```diff
@@ -350,11 +350,11 @@ func (s *Server) JobType(id uuid.UUID) (string, error) {
 }
 
 func (s *Server) Cancel(id uuid.UUID) error {
-	jobType, status, _, err := s.jobStatus(id, nil)
+	jobType, channel, status, _, err := s.jobStatus(id, nil)
 	if err != nil {
 		logrus.Errorf("error getting job status: %v", err)
 	} else {
-		prometheus.CancelJobMetrics(status.Started, jobType)
+		prometheus.CancelJobMetrics(status.Started, jobType, channel)
```
kingsleyzissou, line 357: Same here.

internal/worker/server.go
```diff
@@ -449,16 +449,16 @@ func (s *Server) requestJob(ctx context.Context, arch string, jobTypes []string,
 		return
 	}
 
-	jobType, status, _, err := s.jobStatus(jobId, nil)
+	jobType, channel, status, _, err := s.jobStatus(jobId, nil)
 	if err != nil {
 		logrus.Errorf("error retrieving job status: %v", err)
 	} else {
-		prometheus.DequeueJobMetrics(status.Queued, status.Started, jobType)
+		prometheus.DequeueJobMetrics(status.Queued, status.Started, jobType, channel)
```
kingsleyzissou, line 456: I think this function in the metrics is accepting the `tenant` as the argument and so the metrics here are showing
`org-<tenant>` in the metrics. So we need to either handle this in the `DequeueJobMetrics` function or pass in `channel[4:]`

kingsleyzissou:
To fix the linting issue I think you need to fix the `JobQueueTest` file. There are a few errors there. And then you'll need to add the same functionality to the `dbjobqueue`. 

Also could you please add a detailed commit message :)

I added a couple of inline comments (it's a bit of a nitpick). But otherwise looks good.

kingsleyzissou (dismissed):
Awesome thanks for this. LGTM

kingsleyzissou (approved):

internal/worker/server.go
```diff
@@ -133,7 +133,11 @@ func (s *Server) EnqueueManifestJobByID(job *ManifestJobByID, parent uuid.UUID,
 }
 
 func (s *Server) enqueue(jobType string, job interface{}, dependencies []uuid.UUID, channel string) (uuid.UUID, error) {
-	prometheus.EnqueueJobMetrics(jobType)
+	tenant := channel
+	if len(channel) > 4 {
+		tenant = channel[4:]
```
teg, line 138: How come we chop off the prefix?

teg: Interesting. I'm not sure why we have the prefix, so this might be a stupid question, but could we just pass the whole thing as the tenant ID and decode it (if necessary) when we consume it. In case we ever have other channels with other prefixes, I mean? If we do want to chop off the prefix, we should probably check for the string "org-" explicit, and if that is not set, error out  (or ignore it, whatever is appropriate). That way changes in the future won't take us by surprise.

cc @ondrejbudai

thozza:
This PR introduced an issue in CI -> https://github.com/osbuild/osbuild-composer/runs/6776353479?check_suite_focus=true

PR #2619, opened 2022-05-05
templates/composer: add fluentd sidecar

diaasami:
The sidecar receives logs from the service and forwards them to Splunk
HEC

This pull request includes:

### DIFF
--- a/templates/composer.yml
+++ b/templates/composer.yml
@@ -94,6 +94,8 @@ objects:
             value: "${PGSSLMODE}"
           - name: PGMAXCONNS
             value: "${PGMAXCONNS}"
+          - name: SYSLOG_SERVER
+            value: "localhost:5140"
           ports:
           - name: composer-api
             protocol: TCP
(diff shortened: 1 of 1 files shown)


templates/composer.yml
```diff
@@ -109,6 +111,32 @@ objects:
...
             mountPath: "/var/cache/osbuild-composer"
+        - image: "fluentd-hec:1.2.13"
+          name: fluentd-sidecar
+          resources:
+                requests:
+                  cpu: "${CPU_REQUEST}"
+                  memory: "${MEMORY_REQUEST}"
+                limits:
+                  cpu: "${CPU_LIMIT}"
+                  memory: "${MEMORY_LIMIT}"
```
Gundersanne, line 122: Probably a theoretical concern, but can we limit the cpu to 200m (same as the request)? Composer's limit it 1, and the namespace limit is 3 in total. I just never want to run into a situation where fluentd would slow down composer by taking up the cpu quota.

templates/composer.yml
```diff
@@ -109,6 +111,32 @@ objects:
...
+          env:
+            - name: SPLUNK_HEC_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: splunk-hec
```
Gundersanne, line 127: The secret needs to be mapped to the namespace first (in app-intf), otherwise deployment will fail.

templates/composer.yml
```diff
@@ -109,6 +111,32 @@ objects:
...
+                  key: token
+                  optional: false
+            - name: SPLUNK_HEC_URL
+              valueFrom:
+                secretKeyRef:
+                  name: splunk-hec
```
Gundersanne, line 133: Same as above.

Gundersanne (commented):
LGTM, still need to add the secret first though. Currently this would not deploy.

templates/composer.yml
diaasami: I have updated the secret name to `splunk` as it is in app-intf and the PR is merged.

Gundersanne (approved):

PR #2629, opened 2022-05-10
composer: fail to start if not listening on any API socket.

thozza:
osbuild-composer can now start and keep running without listening on any
API socket. This effectively means that there is no way to communicate
with such instance of osbuild-composer.

Change osbuild-composer to exit with error in case no API socket has been
activated. The log message is intended for the on-premise use case, since
this is the situation when users tend to start osbuild-composer.service
manually and are observing a weird behavior.

Related to https://github.com/osbuild/osbuild-composer/pull/2550

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -79,6 +79,8 @@ func main() {
 		logrus.Fatalf("Could not get listening sockets: " + err.Error())
 	}
 
+	hasAPISocket := false
+
 	if l, exists := listeners["osbuild-composer.socket"]; exists {
 		if len(l) != 1 {
 			logrus.Fatal("The osbuild-composer.socket unit is misconfigured. It should contain only one socket.")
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Nice, I just wonder how this plays out with https://github.com/osbuild/osbuild-composer/blob/60e6d936f032fe43c6cb8dad4ccc4b8990c034ab/cmd/osbuild-composer/composer.go#L216. Should we merge these checks into one place?

thozza:
> Nice, I just wonder how this plays out with
> 
> https://github.com/osbuild/osbuild-composer/blob/60e6d936f032fe43c6cb8dad4ccc4b8990c034ab/cmd/osbuild-composer/composer.go#L216
> 
> . Should we merge these checks into one place?

Thanks for pointing this out. I'll think about it...

thozza:
OK, this PR is useless, I missed the code in `composer.go`

PR #2633, opened 2022-05-11
Add 'core' group back to qcow2 customize in format-request-map

achilleas-k:
Removed in 2beb707def18e3d2b5c3f0ac83e8958af1be9442, possibly
accidentally.
The affected manifests were not regenerated based on this change so
they all already contain the core group.

### DIFF
--- a/tools/test-case-generators/format-request-map.json
+++ b/tools/test-case-generators/format-request-map.json
@@ -593,7 +593,11 @@
           }
         ],
         "modules": [],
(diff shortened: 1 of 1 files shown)


thozza:
IIRC, the `@core` group is already part of the qcow2 image base package set and that's why I didn't mind the change...

thozza (approved):
Lets add it back... Thanks

PR #2636, opened 2022-05-11
[WIP] Unify osbuild jobs

thozza:
This pull request includes:


cmd/osbuild-worker/jobimpl-koji-finalize.go
```diff
@@ -214,8 +314,55 @@ func extractDynamicArgs(job worker.Job) (*worker.KojiInitJobResult, []worker.OSB
...
+	}
+
+	return &kojiInitResult, osbuildResults, nil
+}
+
+// Tests the first osbuild job result for specific values and decides
+// if it is a `worker.OSBuildKojiJobResult` or `worker.OSBuildJobResult`.
+// In case of `worker.OSBuildKojiJobResult` returns `true`.
+// In case of 'worker.OSBuildJobResult` returns `false`.
+func isOldKojiComposeReq(job worker.Job) (bool, error) {
```
thozza, line 342: If anyone has a better idea how to handle / determine this, I'm open to suggestions 🙄

internal/cloudapi/v2/handler.go
```diff
@@ -929,43 +955,65 @@ func (h *apiHandlers) GetComposeManifests(ctx echo.Context, id string) error {
 		return HTTPError(ErrorComposeNotFound)
 	}
 
-	// TODO: support non-koji builds
```
ondrejbudai, line 922: One note here: manifests can contain secrets (most importantly activation keys) and this route isn't currently multi-tenant aware, which can lead to cross-tenant manifest (and thus secret) stealing. So before we can merge this, we need to make this route tenant-aware.

internal/worker/clienterrors/errors.go
```diff
@@ -27,6 +27,8 @@ const (
 	ErrorRPMMDError        ClientErrorCode = 23
 	ErrorEmptyPackageSpecs ClientErrorCode = 24
 	ErrorDNFRepoError      ClientErrorCode = 25
+
+	ErrorJobDependency ClientErrorCode = 26
```
ondrejbudai, line 31: Wrong formatting

internal/worker/server.go
```diff
@@ -174,6 +174,94 @@ func (s *Server) CheckBuildDependencies(dep uuid.UUID, jobErr *clienterrors.Erro
...
+	if err != nil {
+		return nil, err
+	}
+
+	var jobResult *JobResult
+	var jobDeps []uuid.UUID
+	switch jobType {
+	case WorkerJobTypeOSBuild:
+		var osbuildJR OSBuildJobResult
+		_, jobDeps, err = s.OSBuildJobStatus(id, &osbuildJR)
```
ondrejbudai, line 162: I wonder if we can make a generic `JobResult()` method that would just return deps and a job result. It should be technically possible, right?

internal/worker/server.go
```diff
@@ -174,6 +174,94 @@ func (s *Server) CheckBuildDependencies(dep uuid.UUID, jobErr *clienterrors.Erro
 	return nil
 }
 
+// DependencyChainErrors recursively gathers all errors from job's dependencies,
```
ondrejbudai, line 149: I always have mixed feelings about recursion but it's probably simpler than implementing DFS manually.

internal/cloudapi/v2/handler.go
```diff
@@ -689,6 +689,10 @@ func imageStatusFromKojiJobStatus(js *worker.JobStatus, initResult *worker.KojiI
 		return ImageStatusValueFailure
 	}
 
+	if buildResult.OSBuildOutput != nil && !buildResult.OSBuildOutput.Success {
```
ondrejbudai, line 692: Isn't this covered by `if buildResult.JobError`?

internal/cloudapi/v2/handler.go
```diff
@@ -725,6 +729,10 @@ func composeStatusFromKojiJobStatus(js *worker.JobStatus, initResult *worker.Koj
 		if buildResult.JobError != nil {
 			return ComposeStatusValueFailure
 		}
+
+		if buildResult.OSBuildOutput != nil && !buildResult.OSBuildOutput.Success {
```
ondrejbudai, line 733: Same

internal/cloudapi/v2/v2_koji_test.go
```diff
@@ -392,8 +428,9 @@ func TestKojiCompose(t *testing.T) {
 			}`,
 		},
 	}
-	for _, c := range cases[2:3] {
-		test.TestRoute(t, handler, false, "POST", "/api/image-builder-composer/v2/compose", fmt.Sprintf(`
+	for idx, c := range cases {
+		t.Logf("Test case #%d", idx)
```
ondrejbudai, line 402: `t.Run` is more idiomatic. (will make the review harder though so it should be done in a separate commit)

internal/cloudapi/v2/v2_koji_test.go
```diff
@@ -444,8 +488,24 @@ func TestKojiCompose(t *testing.T) {
 		test.TestRoute(t, workerHandler, false, "PATCH", fmt.Sprintf("/api/worker/v1/jobs/%v", token), string(initJobResult), http.StatusOK,
 			fmt.Sprintf(`{"href":"/api/worker/v1/jobs/%v","id":"%v","kind":"UpdateJobResponse"}`, token, token))
 
-		// handle osbuild-koji #1
-		_, token, jobType, rawJob, _, err = workerServer.RequestJob(context.Background(), test_distro.TestArch3Name, []string{worker.WorkerJobTypeOSBuildKoji}, []string{""})
+		// Finishing of the goroutine handling the manifest job is not deterministic and as a result, we may get
```
ondrejbudai, line 461: I kinda dislike using `workerServer.RequestJobById` because a real worker wouldn't use it. I think that dequeueing jobs and checking their IDs is a better way of differentiating.

ondrejbudai (commented):
Looks very nice, the most important issue is the multi-tenancy one, see inline.

internal/worker/server.go
```diff
@@ -28,6 +28,15 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
 )
 
+const (
+	WorkerJobTypeOSBuild        string = "osbuild"
+	WorkerJobTypeOSBuildKoji    string = "osbuild-koji"
```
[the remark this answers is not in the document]
thozza: Yes, we could. I didn't define a new type, because it didn't seem to add too much value. The usual downside of doing so is that you ten can not use the new type easily with functions which take `string` as an argument. However I'm currently not aware of us doing this with the JobType anywhere so it should be OK to do...

internal/worker/clienterrors/errors.go
thozza: Well, kind of.

I found the formatting weird to start with. The error codes seemed to be assigned without leaving any room for new codes, while being split into two blocks. The first block seems to be related to job errors in general, while the second is related to depsolving and DNF. The new error fits more to the first block, but it felt weird to add it to it while breaking the actual error code ordering.

I'll probably squash all errors into one block without any newline separating them into multiple blocks.

achilleas-k (commented):
Good stuff.  Some comments inline.

internal/worker/server.go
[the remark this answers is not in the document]
achilleas-k: The idea crossed my mind too, but probably doesn't add much value.

internal/worker/server.go
thozza: I actually had the same idea, but did it this way in the end. Yes, it should be possible to implement...

internal/worker/server.go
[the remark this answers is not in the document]
thozza: So after looking into it more, I decided to not do it. We do call `strings.HasPrefix()` on the job type, which would require casting the value to `string`. Also Prometheus metrics rely on the value being `string`. While it's all doable, I feel like it would just complicate things without bringing too much value.

internal/worker/clienterrors/errors.go
Gundersanne: One block would be fine. The actual value of the codes doesn't matter as long as they're not changed.

internal/target/koji_target.go
```diff
@@ -18,3 +18,16 @@ func (KojiTargetOptions) isTargetOptions() {}
 func NewKojiTarget(options *KojiTargetOptions) *Target {
 	return newTarget("org.osbuild.koji", options)
 }
+
+type KojiTargetResultOptions struct {
+	HostOS    string `json:"host_os"`
```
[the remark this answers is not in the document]
teg: I agree, it all seems useful.

internal/cloudapi/v2/v2_koji_test.go
```diff
@@ -510,7 +511,7 @@ func TestKojiCompose(t *testing.T) {
 		test.TestRoute(t, handler, false, "GET", fmt.Sprintf("/api/image-builder-composer/v2/composes/%v", finalizeID), ``, http.StatusOK, c.composeStatus, `href`, `id`)
 
 		// get the manifests
-		test.TestRoute(t, handler, false, "GET", fmt.Sprintf("/api/image-builder-composer/v2/composes/%v/manifests", finalizeID), ``, http.StatusOK, `{"manifests":[null,null],"kind":"ComposeManifests"}`, `href`, `id`)
+		test.TestRoute(t, handler, false, "GET", fmt.Sprintf("/api/image-builder-composer/v2/composes/%v/manifests", finalizeID), ``, http.StatusOK, `{"manifests":[{"pipeline":{},"sources":{}},{"pipeline":{},"sources":{}}],"kind":"ComposeManifests"}`, `href`, `id`)
```
teg, line 553: D'oh! Mea culpa!

teg:
Thanks for picking this up! This is really exciting :)

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -645,6 +648,62 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+			defer file.Close()
+
+			logWithId.Info("[Koji] ⬆ Uploading the image")
+			imageHash, imageSize, err := kojiAPI.Upload(file, options.UploadDirectory, options.Filename)
+			if err != nil {
+				osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorKojiBuild, err.Error())
+			}
+			logWithId.Info("[Koji] 🎉 Image successfully uploaded")
+
+			hostOS, err := distro.GetRedHatRelease()
```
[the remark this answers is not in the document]
thozza: I moved those functions to the `common` package.

internal/target/koji_target.go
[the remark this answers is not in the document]
thozza: @ondrejbudai I just realized that you probably meant all 4 values from the structure, right?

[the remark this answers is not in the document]
thozza: Some values, like the image hash and size are calculated by the code which handles uploading to Koji. Having these values for all osbuild jobs, regardless of the target would require calculating them in a common place for all targets. Totally doable, I'm just pointing this out.

I think that it would be OK to move the `HostOS` and `Arch` to the `OSBuildJobResult`, but I'm inclining to not do it for the `ImageHash` and `ImageSize`. The reason is that as we discussed the possibility of supporting multiple exports and targets, these values would loose their relevancy for the osbuild job as such and would be specific to the target and export (e.g. compressed image will have different values than raw image).

cmd/osbuild-worker/jobimpl-koji-finalize.go
ondrejbudai: There's more strict unmarshalling: https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields

internal/target/koji_target.go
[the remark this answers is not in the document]
ondrejbudai: You are absolutely right. So let's move only `HostOS` and `Arch` to the results? Does it work for you? (better to do it now so we don't need to care about backward compat later)

internal/worker/server.go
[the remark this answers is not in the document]
ondrejbudai: It would help IDEs even more. But yeah, let's not make this PR more complex, we can revisit this idea in the future if needed.

internal/worker/clienterrors/errors.go
ondrejbudai: Ah, I missed the new newline between 25 and 26. Nevemind, do whatever feels most reasonable to you. ;)

internal/worker/server.go
ondrejbudai: I would propose at least exploring this idea. The big `switch` will need an update whenever a new job is added, whereas job results are meant to be fairly stable, right?

internal/cloudapi/v2/handler.go
ondrejbudai: Better be safe than sorry? I guess we can remove it when we will get to cleaning up job error handling (this will surely happen at some point).

internal/worker/server.go
```diff
@@ -28,6 +28,15 @@ import (
...
+	WorkerJobTypeKojiInit       string = "koji-init"
+	WorkerJobTypeKojiFinalize   string = "koji-finalize"
+	WorkerJobTypeDepsolve       string = "depsolve"
+	WorkerJobTypeManifestIDOnly string = "manifest-id-only"
```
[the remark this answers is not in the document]
ondrejbudai: CLOSED DUPLICATE of https://github.com/osbuild/osbuild-composer/pull/2636#discussion_r886641458

(sorry, I just want to keep any discussion in one place :laughing: )

cmd/osbuild-worker/jobimpl-koji-finalize.go
thozza: oh, nice! Thanks.

internal/target/koji_target.go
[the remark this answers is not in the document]
thozza: > You are absolutely right. So let's move only `HostOS` and `Arch` to the results? Does it work for you? (better to do it now so we don't need to care about backward compat later)

I agree!

internal/cloudapi/v2/handler.go
```diff
@@ -917,6 +917,32 @@ func (h *apiHandlers) GetComposeLogs(ctx echo.Context, id string) error {
...
+		if err != nil {
+			return nil, err
+		}
+		if depType == worker.WorkerJobTypeManifestIDOnly {
+			foundManifestsJob = true
+			_, _, err = w.ManifestJobStatus(deps[i], &manifestResult)
+			if err != nil {
+				return nil, err
+			}
+			break
```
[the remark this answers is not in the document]
thozza: I agree. Thanks!

internal/worker/server.go
thozza: I agree and plan to rework this part... 😉

internal/cloudapi/v2/v2_koji_test.go
thozza: Well, at least it was a lot of fun to discover this bug 😄

cmd/osbuild-worker/jobimpl-koji-finalize.go
achilleas-k: Yes, `DisallowUnknownFields` was what I was thinking of.  I should be more clear about that in the future :)

internal/worker/server.go
thozza: I looked at the code and remembered why I implemented the function this way...

Each job type specific status method checks the job result in a slightly different way and sets the `result.JobError` to a specific value.

E.g.
https://github.com/osbuild/osbuild-composer/blob/335c597452ab47b42dee9fbca56e2c1c01a54153/internal/worker/server.go#L178-L191
vs.
https://github.com/osbuild/osbuild-composer/blob/335c597452ab47b42dee9fbca56e2c1c01a54153/internal/worker/server.go#L263-L269

For this reason I incline to keeping the implementation as is. I'll add a note to the commit message explaining why I took this approach.

cmd/osbuild-worker/jobimpl-koji-finalize.go
thozza: Because we have custom `UnmarshalJSON()` methods defined for these structures, using `DisallowUnknownFields` turned out to be impossible. After a discussion, we agreed that the current implementation is acceptable. In addition, it is just temporary, because once we deprecate `OSBuildKoji` jobs, it should be removed.

internal/target/koji_target.go
```diff
@@ -18,3 +18,16 @@ func (KojiTargetOptions) isTargetOptions() {}
...
+	Arch      string `json:"arch"`
+	ImageHash string `json:"image_hash"`
```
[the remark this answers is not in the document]
thozza: Renamed...

internal/worker/server.go
```diff
@@ -28,6 +28,15 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
 )
 
+const (
+	WorkerJobTypeOSBuild        string = "osbuild"
```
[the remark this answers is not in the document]
thozza: I renamed all constants and removed the `Worker` prefix.

internal/cloudapi/v2/handler.go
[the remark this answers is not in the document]
thozza: Reworked based on your feedback.

internal/worker/server.go
thozza: I incline to keeping this as is. I think that recursion should be acceptable in this case, since we most probably won't have too deep job dependency chains  to make it an issue.

internal/cloudapi/v2/v2_koji_test.go
thozza: Reworked based on your feedback.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -692,6 +694,54 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
+			if err != nil {
+				osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorKojiBuild, fmt.Sprintf("failed to open the image for reading: %v", err))
+				return nil
+			}
+			defer file.Close()
+
+			logWithId.Info("[Koji] ⬆ Uploading the image")
+			imageHash, imageSize, err := kojiAPI.Upload(file, options.UploadDirectory, options.Filename)
+			if err != nil {
+				osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorKojiBuild, err.Error())
```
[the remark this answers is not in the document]
thozza: I agree. I didn't change it, because I didn't want to unexpectedly skew any statistics that we may be tracking for Koji builds, since this is what `osbuild-koji` is using
https://github.com/osbuild/osbuild-composer/blob/9bff4a4f0fc41224361fd8fbb687be47a0dc8be2/cmd/osbuild-worker/jobimpl-osbuild-koji.go#L165

But if it won't cause any issues, lets change it.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -236,8 +236,16 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			Success: false,
 		},
 		UploadStatus: "failure",
+		Arch:         common.CurrentArch(),
 	}
 
+	hostOS, err := common.GetRedHatRelease()
+	if err != nil {
+		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorKojiBuild, err.Error())
```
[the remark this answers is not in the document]
thozza: Yes, thanks for catching this. It is a leftover from the recent changes in the PR based on the reviews.

internal/cloudapi/v2/handler.go
thozza: I added appropriate middleware and multi-tenancy is now ensured for all `compose/<id>` endpoints...

internal/cloudapi/v2/v2_koji_test.go
thozza: I reverted this change and jobs are now requested in the "usual" way.

cmd/osbuild-worker/jobimpl-osbuild.go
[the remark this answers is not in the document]
thozza: I ended up using `clienterrors.ErrorBuildJob` because it looked as the most appropriate error from the list and I didn't want to add a new error type.

thozza:
I addressed all almost all review comments. For those that were not addressed, I added a comment describing why it is so and expect further discussion.... 🙂

internal/worker/server.go
ondrejbudai: Was the commit message updated?

internal/cloudapi/v2/middleware.go
```diff
@@ -29,6 +32,32 @@ func (s *Server) TenantChannelMiddleware(next echo.HandlerFunc) echo.HandlerFunc
...
+
+		jobChannel, err := s.workers.JobChannel(jobId)
+		if err != nil {
+			fmt.Println(err)
+			return HTTPErrorWithInternal(ErrorComposeNotFound, err)
+		}
+
+		ctxChannel := c.Get(TenantChannelKey).(string)
+		if jobChannel != ctxChannel {
+			return HTTPErrorWithInternal(ErrorResourceNotFound,
```
ondrejbudai, line 53: This should be also `ErrorResourceNotFound`, I think we should make sure not to leak that the job exists if there is a tenant mismatch. (So let's even maybe return the exact same `HTTPErrorWithInternal` call?)

internal/worker/server.go
```diff
@@ -410,6 +410,11 @@ func (s *Server) OSBuildKojiJob(id uuid.UUID, job *OSBuildKojiJob) error {
 	return nil
 }
 
+func (s *Server) JobChannel(id uuid.UUID) (string, error) {
```
ondrejbudai, line 413: Just a thought: I wonder how many DB calls are we doing during handling all routes. We should probably start caching/rework our internal structures to lower the amount of queries. Definitely something for the future. Also, let's not optimize prematurely.

internal/cloudapi/v2/handler.go
```diff
@@ -489,141 +489,143 @@ func imageTypeFromApiImageType(it ImageTypes, arch distro.Arch) string {
...
-
-	jobType, err := h.server.workers.JobType(jobId)
-	if err != nil {
-		return HTTPError(ErrorComposeNotFound)
-	}
-
-	if jobType == worker.JobTypeOSBuild {
-		var result worker.OSBuildJobResult
-		status, _, err := h.server.workers.OSBuildJobStatus(jobId, &result)
+	return h.server.EnsureJobChannel(func(ctx echo.Context, id string) error {
```
ondrejbudai, line 492: I just wonder if a simple condition at the beginning for the handler methods wouldn't be cleaner. Using too many closures/lambdas feels a bit weird in Go. What do you think?

internal/worker/server.go
thozza: Obviously not, sorry... 🙄

internal/worker/server.go
thozza: I fully agree. I was thinking about the same thing and thought at least about not adding `JobChannel()`, but modifying `JobType()` to something like `JobInfo()` and returning more data about the job at once. But given that I ended up wrapping handlers in a middleware method which checks only the channel, it would not save us many calls.

internal/cloudapi/v2/handler.go
thozza: I agree that the readability is a bit questionable.

I like the approach of using middleware method, because it easily allows us to stack multiple methods on top of each other.

I was considering moving the actual handler implementation into a private method, and then just calling it from the exported handler method as the last one. I think that this could improve the readability, while keeping the option to stack multiple methods on top of each other.

internal/cloudapi/v2/middleware.go
thozza: > This should be also `ErrorResourceNotFound`, I think we should make sure not to leak that the job exists if there is a tenant mismatch.

Did you mean `ErrorComposeNotFound`, instead of `ErrorResourceNotFound`?

Nevertheless I agree and my intention was explicitly to not leak any information, but I apparently misunderstood how  `HTTPErrorWithInternal` works. After reading more about it, I think that returning `HTTPError(ErrorComposeNotFound)` should be more appropriate.

> (So let's even maybe return the exact same `HTTPErrorWithInternal` call?)

I'm not sure what do you mean by this suggestion. `HTTPErrorWithInternal` expects a second argument and if we do not provide additional information, there is nothing to pass 🤔

internal/cloudapi/v2/handler.go
thozza: I reworked the code a bit in the latest push, please have a look.

internal/cloudapi/v2/middleware.go
thozza: I reworked the code to return `HTTPError(ErrorComposeNotFound)` in the latest push.

internal/worker/server.go
thozza: I amended the commit message.

internal/cloudapi/v2/middleware.go
ondrejbudai: `HTTPError(ErrorComposeNotFound)` is indeed the best option imho, thanks! :)

internal/cloudapi/v2/handler.go
```diff
@@ -596,11 +596,11 @@ func (h *apiHandlers) GetComposeStatus(ctx echo.Context, id string) error {
 		if err != nil {
 			return HTTPError(ErrorMalformedOSBuildJobResult)
 		}
-		var buildJobResults []worker.OSBuildKojiJobResult
+		var buildJobResults []worker.OSBuildJobResult
```
teg, line 597: Could we get in trouble here if we call `GetComposeStatus` on an old job? Do we need some sort of translation?

thozza: Yes, but we do not care. This is covered in the "Migration notes" part of the PR description.

achilleas-k (approved):
LGTM.

I think there might be a leftover fixup note in this commit message though: https://github.com/osbuild/osbuild-composer/pull/2636/commits/13db116b1425138ef6acdd84977f3d47dfcb00ff

thozza:
> LGTM.
> 
> I think there might be a leftover fixup note in this commit message though: [13db116](https://github.com/osbuild/osbuild-composer/commit/13db116b1425138ef6acdd84977f3d47dfcb00ff)

Thanks for the review. I fixed the commit message.

achilleas-k (approved):
Re-approving

internal/cloudapi/v2/handler.go
ondrejbudai: Will have a look in a bit, sorry, a busy week.

ondrejbudai (approved):

internal/cloudapi/v2/handler.go
ondrejbudai: I like it, thanks! :)

PR #2649, opened 2022-05-14
prometheus: measure pending rather than queued time

Gundersanne:


teg (approved):
Thanks for picking this up!

PR #2652, opened 2022-05-16
Improve GCP test cases to pick regions with available quota

thozza:
We currently use a single GCP Compute region when spinning up VMs using
the imported GCE image. As a result, we are often hitting the
'IN_USE_ADDRESSES' quota limit when there are multiple CI jobs running.
Google does not allow us to increase the quota limit any more.

Change the GCP test cases to use the CI `GCP_REGION` variable to list
all GCE regions with available quota and pick a random one from the
list. The `GCP_REGION` value is used as the region name prefix when
filtering available regions. This means that if you specify an exact GCE
region, such as `us-west1`, you'll always get the same region, but if a
GCP multi-region is used, such as `us`, then a random region prefixed
with 'us' will be used.

Update `cloud-cleaner` accordingly.

Once this PR is merged, we should update the `GCP_REGION` CI variable to `us`.

This pull request includes:

### DIFF
--- a/cmd/cloud-cleaner/main.go
+++ b/cmd/cloud-cleaner/main.go
@@ -22,7 +22,7 @@ func cleanupGCP(testID string, wg *sync.WaitGroup) {
 
 	log.Println("[GCP] Running clean up")
 
-	GCPRegion, ok := os.LookupEnv("GCP_REGION")
+	GCPRegionName, ok := os.LookupEnv("GCP_REGION")
 	if !ok {
 		log.Println("[GCP] Error: 'GCP_REGION' is not set in the environment.")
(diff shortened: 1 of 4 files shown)


thozza:
I tested that `cloud-cleaner` still iterates over regions and zones via https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2464143104#L7244 but I fixed the log message in the last rebase to print only the zone name, not the whole object content.

The PR is ready for review.

thozza:
I had to fix one typo in the last push... 🙄

achilleas-k (approved):
LGTM

PR #2654, opened 2022-05-16
Old worker - New composer test: Use Cloud API

achilleas-k:
Use the cloud API for the `old-worker-new-composer` test.

All of the changes are bits and pieces copied from the `api.sh` script.  Most conditions are removed since we only run one test here: a guest image (`qcow2`) build uploaded to S3.  The test only really checks that the job succeeds and the image has the expected contents.  It doesn't check any fail states or image booting like the API test does.

Reasons for this change:
- Mixed versions of composer and worker aren't a realistic use-case for the weldr API (on prem) but we do run mixed versions in hosted IB, so this test is closer to real world scenarios.
- The cloud API runs depsolve jobs in the worker, whereas the weldr API runs them in composer.  By testing the cloud API we also test the backwards compatibility of the depsolve job.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -251,6 +251,11 @@ regression-include-excluded-packages:
     SCRIPT: regression-include-excluded-packages.sh
 
 regression-old-worker-new-composer:
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/rhel-8.5-ga-x86_64
+        INTERNAL_NETWORK: ["true"]
   extends: .regression
   variables:
     SCRIPT: regression-old-worker-new-composer.sh
--- a/test/cases/regression-old-worker-new-composer.sh
+++ b/test/cases/regression-old-worker-new-composer.sh
@@ -16,7 +16,7 @@ if [ "$ARCH" != "x86_64" ] || [ "$ID" != rhel ] || ! sudo subscription-manager s
     exit 0
 fi
 
-# Provision the software under test.                                                                                                                                                                              
+# Provision the software under test.
 /usr/libexec/osbuild-composer-test/provision.sh
 
 function get_build_info() {
--- a/test/cases/regression-old-worker-new-composer.sh
+++ b/test/cases/regression-old-worker-new-composer.sh
@@ -1,16 +1,23 @@
 #!/bin/bash
 
-# Verify that an older worker (v33) is still compatible with this composer
+# Verify that an older worker (v51) is still compatible with this composer
 # version.
 #
 # Any tweaks to the worker api need to be backwards compatible.
 
 set -exuo pipefail
 
+# Colorful timestamped output.
+function greenprint {
+    echo -e "\033[1;32m[$(date -Isecond)] ${1}\033[0m"
+}
+
+ARTIFACTS=ci-artifacts
+mkdir -p "${ARTIFACTS}"
+
 source /usr/libexec/osbuild-composer-test/set-env-variables.sh
 
-# Only run this on x86 and rhel8 GA; since the container is based on the ubi
-# container, and we use the weldr api
+# Only run this on x86 and rhel8 GA
 if [ "$ARCH" != "x86_64" ] || [ "$ID" != rhel ] || ! sudo subscription-manager status; then
     echo "Test only supported on GA RHEL."
     exit 0
@@ -19,24 +26,22 @@ fi
 # Provision the software under test.
 /usr/libexec/osbuild-composer-test/provision.sh
 
-function get_build_info() {
-    key="$1"
-    fname="$2"
-    if rpm -q --quiet weldr-client; then
-        key=".body${key}"
-    fi
-    jq -r "${key}" "${fname}"
-}
+WORKER_VERSION=67727d1e5cb3f1f86eafd890541381834d001743
+WORKER_RPM=osbuild-composer-worker-51-1.20220504git67727d1.el8.x86_64
 
-WORKER_VERSION=8f21f0b873420a38a261d78a7df130f28b8e2867
-WORKER_RPM=osbuild-composer-worker-33-1.20210830git8f21f0b.el8.x86_64
+# Container image used for cloud provider CLI tools
+CONTAINER_IMAGE_CLOUD_TOOLS="quay.io/osbuild/cloud-tools:latest"
 
-# grab the repos from the test rpms
+greenprint "Copying repository configs from test rpms"
 REPOS=$(mktemp -d)
 sudo dnf -y install osbuild-composer-tests
 sudo cp -a /usr/share/tests/osbuild-composer/repositories "$REPOS/repositories"
 
-# Remove the "new" worker
+greenprint "Stop and disable all services and sockets"
+sudo systemctl stop osbuild-composer.service osbuild-composer.socket osbuild-worker@1.service osbuild-dnf-json.service osbuild-dnf-json.socket
+sudo systemctl disable osbuild-composer.service osbuild-composer.socket osbuild-worker@1.service osbuild-dnf-json.service osbuild-dnf-json.socket
+
+greenprint "Removing latest worker"
 sudo dnf remove -y osbuild-composer osbuild-composer-worker osbuild-composer-tests
 
 function setup_repo {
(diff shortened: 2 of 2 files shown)


thozza (commented):
Nice work 👍  Submitting partial review, will finish after meetings...

achilleas-k:
Thanks for all this.  Too much copy-pasting mess.  You're a hero!

test/cases/regression-old-worker-new-composer.sh
```diff
@@ -97,80 +122,322 @@ while ! sudo systemctl --quiet is-active osbuild-remote-worker@localhost:8700.se
...
+    ],
+    "packages": [
+      "postgresql",
+      "dummy"
+    ],
+    "users":[
+      {
+        "name": "user1",
+        "groups": ["wheel"],
+        "key": "$(cat "${WORKDIR}/usertest.pub")"
```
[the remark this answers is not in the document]
achilleas-k: c&p leftover of course.  Removing the keys since we don't really need to test everything.  Just user creation should be enough.

achilleas-k:
Fixed the stuff you pointed out, thanks again.  I think there's an issue with the setup though and the worker isn't pulling jobs from composer.  I'll jump on the test instance and poke around.

thozza (approved):
LGTM, thanks for all the fixes.

Nevertheless I have been thinking that there are basically two situations:

1. New composer running with old workers.
2. Old composer running with new workers.

For some changes, the migration path may be one of those two, but potentially not both. Although we do try to keep the backward compatibility, sometimes it may be tricky. Also it may be good to rethink the purpose of this test case.

I think that the reason to use Weldr API in this test case was to ensure backward compatibility for the on-premise use case in RHEL (also the reason why we run it only on the latest GA RHEL-8).

With the change to using Cloud API, we are probably changing the purpose of the test a bit. The IB service may not be running the latest RHEL-8 GA composer, so testing the compatibility against it may not make sense. But I'm not sure what should be tested against what 🤷 

So my questions are:

- Would not make sense to test both cases in this test case from above? 
- Should we also test Koji builds via Cloud API, since they use different job types and job result structures.

And I'm sorry for complicating stuff further 😄

test/cases/regression-old-worker-new-composer.sh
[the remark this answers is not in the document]
thozza: ACK

achilleas-k:
> LGTM, thanks for all the fixes.
> 
> Nevertheless I have been thinking that there are basically two situations:
> 
> 1. New composer running with old workers.
> 2. Old composer running with new workers.

Is the second scenario a realistic one?  My thinking was that in the service, we do run new composer with old workers, but I think never the other way around?  I don't know (@Gundersanne ?).

> For some changes, the migration path may be one of those two, but potentially not both. Although we do try to keep the backward compatibility, sometimes it may be tricky. Also it may be good to rethink the purpose of this test case.
> 
> I think that the reason to use Weldr API in this test case was to ensure backward compatibility for the on-premise use case in RHEL (also the reason why we run it only on the latest GA RHEL-8).

What's the scenario where on-prem has mixed version?  Is it only for cases where an upgrade happens while there are jobs in the queue?  Or are we covering cases where users upgrade one subpackage and not the other?

> With the change to using Cloud API, we are probably changing the purpose of the test a bit. The IB service may not be running the latest RHEL-8 GA composer, so testing the compatibility against it may not make sense. But I'm not sure what should be tested against what shrug
> 
> So my questions are:
> 
> * Would not make sense to test both cases in this test case from above?

Perhaps.  We could certainly run both and we don't need to run them for every host distro, so we don't have to add multiple test runners for each scenario.

> * Should we also test Koji builds via Cloud API, since they use different job types and job result structures.

Probably.
It'd be nice if we had a way to test the communication between composer and worker without needing to trigger a whole build.
Ideally, we'd be able to put a request in composer version X, have worker version Y pick it up, and produce manifest M without building it.
Then we would be able to test a bunch of combinations of X vs Y and check if M ever changes.

> And I'm sorry for complicating stuff further smile

It's what we do :)

thozza:
> > LGTM, thanks for all the fixes.
> > Nevertheless I have been thinking that there are basically two situations:
> > 
> > 1. New composer running with old workers.
> > 2. Old composer running with new workers.
> 
> Is the second scenario a realistic one? My thinking was that, in the service, we do run new composer with old workers, but I think never the other way around? I don't know (@Gundersanne ?).

So even if we do not do it now, while trying to unify the `osbuild` and `osbuild-koji` jobs, I think that there is no way to make the new composer work with old workers. Old workers with their job implementations won't be able to handle the differences in dynamic job arguments, etc. The only way to have a migration path is to update workers first, which will be able to handle old and new requests and only then update composer.

But maybe this is needed for now, since we sill use Koji requests via Cloud API only in the internal PSI deployment.

> > For some changes, the migration path may be one of those two, but potentially not both. Although we do try to keep the backward compatibility, sometimes it may be tricky. Also it may be good to rethink the purpose of this test case.
> > I think that the reason to use Weldr API in this test case was to ensure backward compatibility for the on-premise use case in RHEL (also the reason why we run it only on the latest GA RHEL-8).
> 
> What's the scenario where on-prem has mixed versions? Is it only for cases where an upgrade happens while there are jobs in the queue? Or are we covering cases where users upgrade one subpackage and not the other?

I'm not sure TBH, but we do not prohibit anyone from using a remote worker with Weldr... AFAICT one must update both composer and worker together when using them on the same system due to RPM dependencies.

> > With the change to using Cloud API, we are probably changing the purpose of the test a bit. The IB service may not be running the latest RHEL-8 GA composer, so testing the compatibility against it may not make sense. But I'm not sure what should be tested against what shrug
> > So my questions are:
> > 
> > * Would not make sense to test both cases in this test case from above?
> 
> Perhaps. We could certainly run both and we don't need to run them for every host distro, so we don't have to add multiple test runners for each scenario.
> 
> > * Should we also test Koji builds via Cloud API, since they use different job types and job result structures.
> 
> Probably. It'd be nice if we had a way to test the communication between composer and worker without needing to trigger a whole build. Ideally, we'd be able to put a request in composer version X, have worker version Y pick it up, and produce manifest M without building it. Then we would be able to test a bunch of combinations of X vs Y and check if M ever changes.

Well, we probably need to build the image to also test the reporting of results from jobs. Especially with Koji when there are multiple inter-dependent jobs.

achilleas-k:
> Well, we probably need to build the image to also test the reporting of results from jobs. Especially with Koji when there are multiple inter-dependent jobs.

Yeah this is a big topic that I have a lot of ideas for, but it's a whole other thing for another time.

achilleas-k:
> So even if we do not do it now, while trying to unify the `osbuild` and `osbuild-koji` jobs, I think that there is no way to make the new composer work with old workers. Old workers with their job implementations won't be able to handle the differences in dynamic job arguments, etc. The only way to have a migration path is to update workers first, which will be able to handle old and new requests and only then update composer.
> 
> But maybe this is needed for now, since we sill use Koji requests via Cloud API only in the internal PSI deployment.
> 

Do we really **need** to do it this way?  If we do, then we should do it properly.  We would need to formalise the range of compatible versions between composer and worker if we're going to be doing this in both directions (new composer - old workers in cloud; old composer - new workers in Koji).

But the more I think about it, the more I'm thinking this is more work than it's worth.  New composer - old workers is how we run now and there's almost never a need for the other way around.  In the cloud we generally update composer constantly and the workers less frequently.  Afaik, we don't really fret much about compatibility issues unless we know something big changed.  This test change is meant as a guarantee for that.

> > > For some changes, the migration path may be one of those two, but potentially not both. Although we do try to keep the backward compatibility, sometimes it may be tricky. Also it may be good to rethink the purpose of this test case.
> > > I think that the reason to use Weldr API in this test case was to ensure backward compatibility for the on-premise use case in RHEL (also the reason why we run it only on the latest GA RHEL-8).
> > 
> > 
> > What's the scenario where on-prem has mixed versions? Is it only for cases where an upgrade happens while there are jobs in the queue? Or are we covering cases where users upgrade one subpackage and not the other?
> 
> I'm not sure TBH, but we do not prohibit anyone from using a remote worker with Weldr... AFAICT one must update both composer and worker together when using them on the same system due to RPM dependencies.
> 

Yeah that's exactly my thinking behind this change.  On-prem mixed versions isn't a very realistic scenario, but that's what we were testing.  The dnf-json PR is proof that we need to test the real scenario of new composer - old worker using the cloud API, because in the current state of the PR, this test should fail (it did not).

> > > With the change to using Cloud API, we are probably changing the purpose of the test a bit. The IB service may not be running the latest RHEL-8 GA composer, so testing the compatibility against it may not make sense. But I'm not sure what should be tested against what shrug
> > > So my questions are:
> > > 
> > > * Would not make sense to test both cases in this test case from above?
> > 
> > 
> > Perhaps. We could certainly run both and we don't need to run them for every host distro, so we don't have to add multiple test runners for each scenario.
> > > * Should we also test Koji builds via Cloud API, since they use different job types and job result structures.
> > 
> > 
> > Probably. It'd be nice if we had a way to test the communication between composer and worker without needing to trigger a whole build. Ideally, we'd be able to put a request in composer version X, have worker version Y pick it up, and produce manifest M without building it. Then we would be able to test a bunch of combinations of X vs Y and check if M ever changes.
> 
> Well, we probably need to build the image to also test the reporting of results from jobs. Especially with Koji when there are multiple inter-dependent jobs.

Yes of course.  Discussion for another time though.  I have thoughts :)

achilleas-k:
At first I included the dummy package installation from a custom repository (copied from the api.sh test), but this doesn't work with cross-version setups right now.  I'll open an issue to include a test for it once it's fixed.

achilleas-k:
More incompatibilities:

Even running composer v51 against worker v50 doesn't work, because in the pipelines for v50 we install two package sets: `os` and `blueprint`.  In v51 we only install `os` because they're combined.  This is fine if the depsolve happens in composer (v51), which happens in the original version of this script (in current `main`) because then the depsolves are combined and we only need `os`.
But if we depsolve in v50, they're not combined, and we get back `os` and `blueprint` from a v50 depsolve and then we only install `os` so the user's file selection is basically ignored (though it appears in the sources in the manifest).

achilleas-k:
I've changed the script to use worker v51 against the `HEAD` version of composer (`main`, PR, etc).
As of now v51 is the latest version (though v52 should be coming today?), but since the worker will be pinned to v51, it will stay as a compatibility test so that we catch incompatibilities in the future.
This test script should keep testing v51 workers until we decide to break compatibility, which we can always do when necessary but it'll always be reflected in the tests.

.gitlab-ci.yml
```diff
@@ -251,6 +251,11 @@ regression-include-excluded-packages:
     SCRIPT: regression-include-excluded-packages.sh
 
 regression-old-worker-new-composer:
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/rhel-8.5-ga-x86_64
+        INTERNAL_NETWORK: ["true"]
   extends: .regression
```
Gundersanne, line 259: +1 this test only needs to run on rhel ga

test/cases/regression-old-worker-new-composer.sh
```diff
@@ -16,7 +16,7 @@ if [ "$ARCH" != "x86_64" ] || [ "$ID" != rhel ] || ! sudo subscription-manager s
     exit 0
 fi
 
-# Provision the software under test.                                                                                                                                                                              
+# Provision the software under test.
```
Gundersanne, line 26: the best change

test/cases/regression-old-worker-new-composer.sh
```diff
@@ -54,123 +59,362 @@ priority=${priority}
...
+    "users": [
+      {
+        "name": "user1",
+        "groups": ["wheel"]
+      },
+      {
+        "name": "user2"
+      }
+    ]
+  },
```
Gundersanne, line 270: So user customization is something we technically don't do in the service (yet). But it can't hurt to have it here.

test/cases/regression-old-worker-new-composer.sh
```diff
@@ -54,123 +59,362 @@ priority=${priority}
...
+sudo ${CONTAINER_RUNTIME} pull --creds "${V2_QUAY_USERNAME}":"${V2_QUAY_PASSWORD}" \
      "quay.io/osbuild/osbuild-composer-ubi-pr:${CI_COMMIT_SHA}"
 
+cat <<EOF | sudo tee "/etc/osbuild-composer/osbuild-composer.toml"
+log_level = "debug"
+[koji]
+allowed_domains = [ "localhost", "client.osbuild.org" ]
+ca = "/etc/osbuild-composer/ca-crt.pem"
+[koji.aws_config]
+bucket = "${AWS_BUCKET}"
```
Gundersanne, line 93: This makes me realize that another reason to update the worker in the regression test here is configs. So the bucket in the hosted service is configured in the worker itself, I'll be opening a PR soon to drop the bucket config from the cloudapi. So it's fine here because the worker has supported this for a while now, but in future we might need to update the worker here to benefit from updated configs (multi-region support is one of the things coming up).

test/cases/regression-old-worker-new-composer.sh
```diff
@@ -54,123 +59,362 @@ priority=${priority}
...
+sudo systemctl enable --now osbuild-dnf-json.socket
+while ! sudo systemctl --quiet is-active osbuild-dnf-json.socket; do
+    sudo systemctl status osbuild-dnf-json.socket
+    sleep 1
+    sudo systemctl enable --now osbuild-dnf-json.socket
 done
 set -e
 
-function log_on_exit() {
-    sudo podman logs composer
```
Gundersanne, line 101: should we save the container log as an artifact?

Gundersanne (approved):
I'll merge this once shutzy is green. Thanks for improving this test by a lot. Just a few thoughts.

test/cases/regression-old-worker-new-composer.sh
achilleas-k: ah yeah good point.  Copied from the old test.  Like you said, can't hurt to cover it, unless there's any chance it interferes with the real test.

test/cases/regression-old-worker-new-composer.sh
achilleas-k: Good to know.  Should definitely update these.  It's not rare that I'll look at tests (instead of docs :upside_down_face:) to figure out how to configure something, so the tests should do things properly.

test/cases/regression-old-worker-new-composer.sh
achilleas-k: Indeed that would be very useful.

PR #2655, opened 2022-05-16
Upload to HTTPS S3 - Support self signed certificate

ygalblum:
API
---
Allow the user to pass the CA public certification or skip the verification

AWSCloud
--------
Restore the old version of newAwsFromCreds for access to AWS
Create a new method newAwsFromCredsWithEndpoint for Generic S3 which sets the endpoint and optionally overrides the CA Bundle or skips the SSL certificate verification

jobimpl-osbuild
---------------
Update with the new parameters

osbuild-upload-generic-s3
-------------------------
Add ca-bunlde and skip-ssl-verification flags

tests
-----
Split the tests into http, https with certificate and https skip certificate check
Create a new base test for S3 over HTTPS for secure and insecure
Move the generic S3 test to tools to reuse for secure and insecure connections
All S3 tests now use the aws cli tool
Update the libvirt test to be able to download over HTTPS
Update the RPM spec

This pull request includes:

### DIFF
--- a/internal/cloud/awscloud/awscloud.go
+++ b/internal/cloud/awscloud/awscloud.go
@@ -1,7 +1,9 @@
 package awscloud
 
 import (
+	"crypto/tls"
 	"fmt"
+	"net/http"
 	"os"
 	"time"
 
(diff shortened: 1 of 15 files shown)


internal/cloud/awscloud/awscloud.go
```diff
@@ -57,18 +56,60 @@ func New(region string, accessKeyID string, accessKey string, sessionToken strin
...
+		if err != nil {
+			return nil, err
+		}
+		defer caBundleReader.Close()
+		sessionOptions.CustomCABundle = caBundleReader
+	}
+
+	if skipSSLVerification { // #nosec G402
+		sessionOptions.Config.HTTPClient = &http.Client{
+			Transport: &http.Transport{
```
ondrejbudai, line 92: Creating `http.Transport` out of thin air is generally discouraged because `Transport: nil` is different (e.g. timeouts) from `Transport: &http.Transport{}`. Can you please clone the default transport instead and modify that one?

https://github.com/osbuild/osbuild-composer/blob/a6e2755fadb8d3da6f52fdc95c49c5c642ca5588/internal/worker/client.go#L94

ygalblum: Thanks for the best practice, done.

ondrejbudai (approved):

PR #2665, opened 2022-05-18
distro/rhel90: add support for azure marketplace

gicmo:
Add support for building images for the Azure marketplace: add a new image type "azure-rhui" that can be used to build images tailored to the Azure marketplace.
This code is based on the corresponding image type in 8.6.

NB: does not have systemd-resovled (following RHEL 9 defaults)


gicmo:
I had to fix the `rhel-91` tests manifests in `repos.json` in order to generate RHEL 9.1 manifests for Azure.

thozza:
#2660 needs to land to make PackIt pass on Fedora...

PR #2666, opened 2022-05-18
cloudapi: get specific error from openapi

chloenayon:
When validation fails, get the specific error from openapi and return it.

This pull request includes:


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -392,13 +392,16 @@ components:
...
+          - details
         properties:
           code:
             type: string
           reason:
             type: string
           operation_id:
             type: string
+          details:
+            type: string
```
Gundersanne, line 404: Should be type: {}; similar to `ComposeStatusError` 's error.

internal/cloudapi/v2/server.go
```diff
@@ -113,7 +113,12 @@ func (s *Server) ValidateRequest(next echo.HandlerFunc) echo.HandlerFunc {
 
 		ctx := request.Context()
 		if err := openapi3filter.ValidateRequest(ctx, input); err != nil {
-			return HTTPErrorWithInternal(ErrorInvalidRequest, err)
+			details := ""
+			re, ok := err.(*openapi3filter.RequestError)
+			if ok {
+				details = re.Error()
```
Gundersanne, line 119: What does this error look like? If it's not a good/readable/presentable error string, we'd need to look at the other fields in the error.
```
type RequestError struct {
	Input       *RequestValidationInput
	Parameter   *openapi3.Parameter
	RequestBody *openapi3.RequestBody
	Reason      string
	Err         error
}
```

Gundersanne, line 119: What does this error look like? If it's not a good/readable/presentable error string, we'd need to look at the other fields in the error.
```
type RequestError struct {
	Input       *RequestValidationInput
	Parameter   *openapi3.Parameter
	RequestBody *openapi3.RequestBody
	Reason      string
	Err         error
}
```

internal/cloudapi/v2/errors.go
```diff
@@ -72,69 +72,71 @@ const (
 )
 
 type ServiceErrorCode int
+type ErrorMessage string
 
 type serviceError struct {
 	code       ServiceErrorCode
 	httpStatus int
 	reason     string
+	details    string
```
Gundersanne, line 81: Can we make this interface{} ; to map better to the eventual type of details in the api?

Gundersanne, line 81: Can we make this interface{} ; to map better to the eventual type of details in the api?

internal/cloudapi/v2/errors.go
```diff
@@ -144,24 +146,43 @@ func find(code ServiceErrorCode) *serviceError {
...
 
 // Make an echo compatible error out of a service error
 func HTTPError(code ServiceErrorCode) error {
-	return HTTPErrorWithInternal(code, nil)
+	return HTTPErrorWithInternal(code, nil, "")
 }
 
 // echo.HTTPError has a message interface{} field, which can be used to include the ServiceErrorCode
-func HTTPErrorWithInternal(code ServiceErrorCode, internalErr error) error {
+func HTTPErrorWithInternal(code ServiceErrorCode, internalErr error, details string) error {
```
Gundersanne, line 158: To reduce the diff a bit can we just have `HTTPErrorWithDetails` ? Rename this function to httpError (with all fields, not calitapized), and then keep the HTTPErrorWithInternal signature the same?

Gundersanne, line 158: To reduce the diff a bit can we just have `HTTPErrorWithDetails` ? Rename this function to httpError (with all fields, not calitapized), and then keep the HTTPErrorWithInternal signature the same?

internal/cloudapi/v2/v2_test.go
```diff
@@ -85,6 +85,7 @@ func TestUnknownRoute(t *testing.T) {
 		"id": "21",
 		"kind": "Error",
 		"code": "IMAGE-BUILDER-COMPOSER-21",
+		"details": "",
```
Gundersanne, line 88: Just add it to ignored fields like 'operation_id' below. 'details' shouldn't have any schema attached to it.

Gundersanne, line 88: Just add it to ignored fields like 'operation_id' below. 'details' shouldn't have any schema attached to it.

internal/cloudapi/v2/errors.go
```diff
@@ -72,69 +72,71 @@ const (
 )
 
 type ServiceErrorCode int
+type ErrorMessage string
```
Gundersanne, line 75: This isn't used anywhere I think?

Gundersanne, line 75: This isn't used anywhere I think?

Gundersanne (commented):
Thank you, some comments.

internal/cloudapi/v2/errors.go
```diff
@@ -144,24 +146,43 @@ func find(code ServiceErrorCode) *serviceError {
...
 func HTTPError(code ServiceErrorCode) error {
-	return HTTPErrorWithInternal(code, nil)
+	return HTTPErrorWithInternal(code, nil, "")
 }
 
 // echo.HTTPError has a message interface{} field, which can be used to include the ServiceErrorCode
-func HTTPErrorWithInternal(code ServiceErrorCode, internalErr error) error {
+func HTTPErrorWithInternal(code ServiceErrorCode, internalErr error, details string) error {
 	se := find(code)
+	se.details = details
```
Gundersanne, line 160: This won't actually assign the details to the error, as the service error is thrown away after constructing the echo error below. It needs to make it all the way to the error handler.

internal/cloudapi/v2/errors_test.go
```diff
@@ -25,7 +25,7 @@ func TestAPIError(t *testing.T) {
 	for _, se := range getServiceErrors() {
 		ctx := e.NewContext(nil, nil)
 		ctx.Set("operationID", "test-operation-id")
-		apiError := APIError(se.code, nil, ctx)
+		apiError := APIError(se.code, nil, ctx, "")
```
Gundersanne, line 28: same here, I'd default to nil.

Gundersanne (approved):
lgtm thank you!

PR #2674, opened 2022-05-22
[rhel9] Fixes for grub2 config (ImageConfig) and azure-rhui

gicmo:
- We need to actually set the grub2 configuration if there is one. Doh.
- Install the "NetworkManager-cloud-setup" on Azure Marketplace images.


thozza (approved):
👍

PR #2677, opened 2022-05-23
Allow the user to set the location of the certificates

bardielle:
Allow the user to set the location of the certificates for the composer and the worker in the composer-config file.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -10,9 +10,7 @@ import (
 )
 
 const (
-	configFile     = "/etc/osbuild-composer/osbuild-composer.toml"
-	ServerKeyFile  = "/etc/osbuild-composer/composer-key.pem"
-	ServerCertFile = "/etc/osbuild-composer/composer-crt.pem"
+	configFile = "/etc/osbuild-composer/osbuild-composer.toml"
 )
 
 var repositoryConfigs = []string{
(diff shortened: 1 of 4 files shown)


cmd/osbuild-composer/main.go
```diff
@@ -107,13 +107,22 @@ func main() {
...
 	}
-
+	
 	if l, exists := listeners["osbuild-composer-api.socket"]; exists {
 		if len(l) != 1 {
 			logrus.Fatal("The osbuild-composer-api.socket unit is misconfigured. It should contain only one socket.")
 		}
 
-		err = composer.InitAPI(ServerCertFile, ServerKeyFile, config.Koji.EnableTLS, config.Koji.EnableMTLS, config.Koji.EnableJWT, l[0])
+		serverCertFile := config.Koji.ServerCertFile
```
ygalblum, line 116: The default values are loaded into the config structure in this function: https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-composer/config.go#L86. The returned structure gets overridden by the values in the configuration file. So, instead of checking and overriding here, you should add the default values to the default configuration

cmd/osbuild-composer/testdata/test.toml
```diff
@@ -1,13 +1,17 @@
 [koji]
 allowed_domains = [ "osbuild.org" ]
 ca = "/etc/osbuild-composer/ca-crt.pem"
+server_key_file = "/etc/osbuild-composer/certs/composer-key.pem"
```
ygalblum, line 4: Does the empty value case (that takes the defaults) get tested somewhere?

ondrejbudai:
@bardielle @ygalblum is this still needed? :)

PR #2678, opened 2022-05-23
Switch to latest RHEL versions for testing

atodorov:
This pull request includes:


.gitlab-ci.yml
```diff
@@ -166,33 +166,35 @@ Base:
...
+  rules:
+    - !reference [.upstream_rules, rules]
+    - !reference [.nightly_rules, rules]
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/manifest_tests.sh
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/rhel-9.1-nightly-x86_64
```
thozza, line 196: Lets use centos stream here to prevent us disabling it the next time, because the test does not care about the actual runner OS.

.gitlab-ci.yml
```diff
@@ -291,9 +295,10 @@ OSTree:
     matrix:
       - RUNNER:
           - rhos-01/fedora-35-x86_64
-          - rhos-01/rhel-8.5-ga-x86_64
-          #- rhos-01/rhel-8.6-nightly-x86_64-large
-          #- rhos-01/rhel-9.0-nightly-x86_64
+          - rhos-01/rhel-8.6-ga-x86_64
+          - rhos-01/rhel-9.0-ga-x86_64
+          - rhos-01/rhel-8.7-nightly-x86_64-large
+          - rhos-01/rhel-9.1-nightly-x86_64
```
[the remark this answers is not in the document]
atodorov: No. I guess it hasn't been an issue and I've just changed the RHEL version without changing the VM type. Are you requesting this to change or not ?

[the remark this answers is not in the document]
thozza: I'm OK with keeping the runner as is, I just noticed this difference and was wondering why it is so....

.gitlab-ci.yml
```diff
@@ -166,33 +166,35 @@ Base:
...
-#  rules:
-#    - !reference [.upstream_rules, rules]
-#    - !reference [.nightly_rules, rules]
-#  script:
-#    - schutzbot/deploy.sh
-#    - /usr/libexec/tests/osbuild-composer/manifest_tests.sh
-#  parallel:
-#    matrix:
-#      - RUNNER:
-#          - aws/rhel-9.0-nightly-x86_64
```
ondrejbudai, line 194: Why is this running on CS9 instead of RHEL 9?

thozza: The test does not care about the OS. So I suggested to use C9S for it to prevent us from disabling it in the future when doing the same exercise of updating runners after RHEL-9 GA... We could even run it on Fedora, but since C9S is a new and rolling release, it felt like a better choice.

jrusz:
@ondrejbudai I've updated `Schutzfile` to use https://github.com/osbuild/osbuild/commit/d602b5ff7f40dd3abb956d29aa0c7573e6e0ed29 which as a bonus fixes an issue we found last week on RHEL-9.1

ondrejbudai (approved):

PR #2684, opened 2022-05-24
tests/gcp: pick machine type from those available in the zone

thozza:
Do not rely on the default machine type when creating a GCE instance,
but rather list the available machine types in the zone and pick from
them. Test cases will pick the smallest machine type which name matches
the `^n\d-standard-\d$` regular expression.

This should prevent CI failures like
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2497043942#L2930

This pull request includes:


ondrejbudai (approved):
Thanks! GCP is such a pain, isn't it? :D

PR #2686, opened 2022-05-24
Add support for uploading to generic S3 service using the Composer API

ygalblum:
Use case
--------
If Endpoint is not set and Region is - upload to AWS S3
If both the Endpoint and Region are set - upload the Generic S3 via Weldr API
If neither the Endpoint and Region are set - upload the Generic S3 via Composer API (use configuration)

jobimpl-osbuild
---------------
Add configuration fields for Generic S3 upload
Support S3 upload requests coming from Weldr or Composer API to either AWS or Generic S3
Weldr API for Generic S3 requires that all connection parameters but the credentials be passed in the API call
Composer API for Generic S3 requires that all conneciton parameters are taken from the configuration
Adjust to the consolidation in Target and UploadOptions

Target and UploadOptions
------------------------
Add the fields that were specific to the Generic S3 structures to the AWS S3 one
Remove the structures for Generic S3 and always use the AWS S3 ones

Worker Main
-----------
Add Endpoint, Region, Bucket, CABundle and SkipSSLVerification to the configuration structure
Pass the values to the Server

Weldr API
---------
Keep the generic.s3 provider name to maintain the API, but unmarshel into awsS3UploadSettings

tests - api.sh
--------------
Allow the caller to specifiy either AWS or Generic S3 upload targets for specific image types
Implement the pieces required for testing upload to a Generic S3 service
In some cases generalize the AWS S3 functions for reuse

GitLab CI
---------
Add test case for api.sh tests with edge-commit and generic S3

This pull request includes:


test/cases/api.sh
```diff
@@ -12,16 +12,69 @@
...
+IMAGE_TYPE_GUEST="guest-image"
+IMAGE_TYPE_VSPHERE="vsphere"
+
+if (( $# > 2 )); then
+    echo "$0 does not support more than two arguments"
+    exit 1
+fi
+
+if (( $# == 0 )); then
+    echo "$0 requires that you set the image type to build"
```
Gundersanne, line 44: This diff feels a bit off, is there a reason to move the IMAGE_TYPE switch block up?

test/cases/api.sh
```diff
@@ -12,16 +12,69 @@
...
+        if [ "${CLOUD_PROVIDER}" != "${CLOUD_PROVIDER_AWS_S3}" ] && [ "${CLOUD_PROVIDER}" != "${CLOUD_PROVIDER_GENERIC_S3}" ]; then
+            echo "${IMAGE_TYPE} can only be uploaded to either ${CLOUD_PROVIDER_AWS_S3} or ${CLOUD_PROVIDER_GENERIC_S3}"
+            exit 1
+        fi
+        ;;
+    *)
+        echo "Unknown image type: ${IMAGE_TYPE}"
+        exit 1
+esac
+
```
Gundersanne, line 77: ditto

Gundersanne (commented):
Thank you, just some qs about the test diff. Looks good overall.

test/cases/api.sh
ygalblum: As explained, before my change, this code only set the CLOUD_PROVIDER, but now it also validates the input. As such, I think it should be done at the start of the script

internal/weldr/upload.go
```diff
@@ -137,7 +131,7 @@ func (u *uploadRequest) UnmarshalJSON(data []byte) error {
 	case "oci":
 		settings = new(ociUploadSettings)
 	case "generic.s3":
-		settings = new(genericS3UploadSettings)
+		settings = new(awsS3UploadSettings)
```
ondrejbudai, line 134: I would like a comment in the code here explaining what's going on.

test/cases/api.sh
```diff
@@ -1315,7 +1437,8 @@ function verifyInVSphere() {
 }
 
 # Verify s3 blobs
-function verifyInAWSS3() {
+function verifyInS3() {
+    BUCKET_NAME=${1}
```
ondrejbudai, line 1441: `local` would be nicer

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -30,14 +30,19 @@ import (
...
-	AWSBucket      string
-	GenericS3Creds string
+	Store                        string
+	Output                       string
+	KojiServers                  map[string]koji.GSSAPICredentials
+	GCPCreds                     string
+	AzureCreds                   *azure.Credentials
+	AWSCreds                     string
+	AWSBucket                    string
+	GenericS3Creds               string
```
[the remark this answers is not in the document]
ygalblum: I did not want to unify the details of the AWS and Generic S3 because the AWS details are used not only when uploading to S3 but also when building an AWS AMI (the image is uploaded to the same S3 bucket and then imported as an AMI).

As for the struct, should I create a struct just for the Generic S3 case (with `Creds`, `Endpoint`, `Region`, `Bucket`, `CABundle` and `SSLVerification`)? Or also for the AWS details (same struct, but with only the required details)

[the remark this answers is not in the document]
Gundersanne: IMO long term the aws details should become ec2 specific config (not in the PR), and generic s3 config should be just for s3. But to make sure we don't break it currently, we need to fill the bucket with the AWSBucket (ec2 config) details.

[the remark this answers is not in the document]
ondrejbudai: That's a valid point, @Gundersanne. Anyway, a struct for (generic) S3 seems like a good idea to me.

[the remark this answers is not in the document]
Gundersanne: Yea we can still have the separate struct and then fill it for now until we have all the configs changed everywhere.

internal/weldr/upload.go
ygalblum: I added a comment, please let me know if it's enough, or if you have any other suggestions

cmd/osbuild-worker/jobimpl-osbuild.go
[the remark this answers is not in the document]
ygalblum: I've put all S3 configuration values into a new struct inside the `OSBuildJobImpl` struct

ondrejbudai (commented):
Nice, can you squash, please? Then, this is ready to go in.

ygalblum:
> Nice, can you squash, please? Then, this is ready to go in.

@ondrejbudai Done

ondrejbudai (approved):

PR #2693, opened 2022-05-25
Compress Azure RHUI artefacts

gicmo:
Those images are forced to be 64GiB in size but mostly consist of zeros. This makes them hard to handle, e.g. uploading to brew takes a forever. Therefore compress the RHUI artefacts.

### DIFF
--- a/internal/distro/rhel86/pipelines.go
+++ b/internal/distro/rhel86/pipelines.go
@@ -52,32 +52,47 @@ func prependKernelCmdlineStage(pipeline *osbuild.Pipeline, t *imageType, pt *dis
 	return pipeline
 }
 
-func vhdPipelines(t *imageType, customizations *blueprint.Customizations, options distro.ImageOptions, repos []rpmmd.RepoConfig, packageSetSpecs map[string][]rpmmd.PackageSpec, rng *rand.Rand) ([]osbuild.Pipeline, error) {
-	pipelines := make([]osbuild.Pipeline, 0)
-	pipelines = append(pipelines, *buildPipeline(repos, packageSetSpecs[buildPkgsKey], t.arch.distro.runner))
+func vhdPipelines(compress bool) pipelinesFunc {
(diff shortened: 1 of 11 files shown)


ondrejbudai (commented):
LGTM but I want at least one more approval from someone who does more image definitions than me. :)

internal/distro/rhel86/pipelines.go
```diff
@@ -52,32 +52,47 @@ func prependKernelCmdlineStage(pipeline *osbuild.Pipeline, t *imageType, pt *dis
 	return pipeline
 }
 
-func vhdPipelines(t *imageType, customizations *blueprint.Customizations, options distro.ImageOptions, repos []rpmmd.RepoConfig, packageSetSpecs map[string][]rpmmd.PackageSpec, rng *rand.Rand) ([]osbuild.Pipeline, error) {
-	pipelines := make([]osbuild.Pipeline, 0)
-	pipelines = append(pipelines, *buildPipeline(repos, packageSetSpecs[buildPkgsKey], t.arch.distro.runner))
+func vhdPipelines(compress bool) pipelinesFunc {
```
achilleas-k, line 55: This is cool.  We should do more of this.

achilleas-k (approved):
Neat!

internal/distro/rhel86/pipelines.go
thozza: I agree that this is clever, but not sure if we do more of this 🙄  The only reason from my PoV is that I'm not sure how it will play out once we at some point start pulling definitions to more declarative form. I feel like these kind of things will make it more complicated in the future.

We do a very similar thing, but in a different way for the `ec2` and `ami` images.

thozza:
@gicmo I forgot to ask how do you envision this working when building the images through Koji and uploading them directly to Koji as well as to Azure? I expect that we won't be able to import the image as compressed file to Azure and would have to decompress it first. Wouldn't it make more sense to have multiple exports in the image type definition (compressed and not-compressed) from different pipelines and have each of them uploaded to a different location?

However I'm aware that this would be probably a huge thing to implement.

internal/distro/rhel86/pipelines.go
gicmo: More descriptive pipelines would be very nice but I think it will need quite a bit of (re-)design work on various levels.

PR #2695, opened 2022-05-26
oscap: implement OpenSCAP build remediation

kingsleyzissou:
This pull request includes:
- blueprint customizations for openscap integration
- osbuild2 stage for oscap remediatoin
- validation of customization options
- add oscap stage to fedora, rhel87 and rhel91 non-rpmostree builds

Depends on:
https://github.com/osbuild/osbuild/pull/1033

### DIFF
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -22,6 +22,7 @@ type Customizations struct {
 	Filesystem         []FilesystemCustomization `json:"filesystem,omitempty" toml:"filesystem,omitempty"`
 	InstallationDevice string                    `json:"installation_device,omitempty" toml:"installation_device,omitempty"`
 	FDO                *FDOCustomization         `json:"fdo,omitempty" toml:"fdo,omitempty"`
+	OpenSCAP           *OpenSCAPCustomization    `json:"openscap,omitempty" toml:"openscap,omitempty"`
 }
 
 type FDOCustomization struct {
@@ -88,6 +89,11 @@ type ServicesCustomization struct {
 type FilesystemCustomization struct {
 	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
 	MinSize    uint64 `json:"minsize,omitempty" toml:"size,omitempty"`
+}
+
+type OpenSCAPCustomization struct {
+	DataStream string `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	ProfileID  string `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
 }
 
 func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
@@ -366,4 +372,11 @@ func (c *Customizations) GetFDO() *FDOCustomization {
 		return nil
 	}
 	return c.FDO
+}
+
+func (c *Customizations) GetOpenSCAP() *OpenSCAPCustomization {
+	if c == nil {
+		return nil
+	}
+	return c.OpenSCAP
 }
--- a/internal/distro/fedora/distro.go
+++ b/internal/distro/fedora/distro.go
@@ -14,6 +14,7 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/environment"
 	"github.com/osbuild/osbuild-composer/internal/image"
 	"github.com/osbuild/osbuild-composer/internal/manifest"
+	"github.com/osbuild/osbuild-composer/internal/oscap"
 	"github.com/osbuild/osbuild-composer/internal/platform"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/runner"
@@ -47,6 +48,12 @@ const (
 var (
 	mountpointAllowList = []string{
 		"/", "/var", "/opt", "/srv", "/usr", "/app", "/data", "/home", "/tmp",
+	}
+
+	oscapProfileAllowList = []oscap.Profile{
+		oscap.Ospp,
+		oscap.PciDss,
+		oscap.Standard,
 	}
 
 	// Services
@@ -702,6 +709,22 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 
 	if len(invalidMountpoints) > 0 {
 		return fmt.Errorf("The following custom mountpoints are not supported %+q", invalidMountpoints)
+	}
+
+	if osc := customizations.GetOpenSCAP(); osc != nil {
+		supported := oscap.IsProfileAllowed(osc.ProfileID, oscapProfileAllowList)
+		if !supported {
+			return fmt.Errorf(fmt.Sprintf("OpenSCAP unsupported profile: %s", osc.ProfileID))
+		}
+		if t.rpmOstree {
+			return fmt.Errorf("OpenSCAP customizations are not supported for ostree types")
+		}
+		if osc.DataStream == "" {
+			return fmt.Errorf("OpenSCAP datastream cannot be empty")
+		}
+		if osc.ProfileID == "" {
+			return fmt.Errorf("OpenSCAP profile cannot be empty")
+		}
 	}
 
 	return nil
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -186,6 +186,9 @@ func (p *OS) getBuildPackages() []string {
 		packages = append(packages, "policycoreutils")
 		packages = append(packages, fmt.Sprintf("selinux-policy-%s", p.SElinux))
 	}
+	if p.OpenSCAPConfig != nil {
+		packages = append(packages, "openscap-scanner", "scap-security-guide")
+	}
 	return packages
 }
 
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -70,23 +70,24 @@ type OSCustomizations struct {
 	Users    []blueprint.UserCustomization
 	Firewall *blueprint.FirewallCustomization
 	// TODO: drop osbuild types from the API
-	Grub2Config   *osbuild.GRUB2Config
-	Sysconfig     []*osbuild.SysconfigStageOptions
-	SystemdLogind []*osbuild.SystemdLogindStageOptions
-	CloudInit     []*osbuild.CloudInitStageOptions
-	Modprobe      []*osbuild.ModprobeStageOptions
-	DracutConf    []*osbuild.DracutConfStageOptions
-	SystemdUnit   []*osbuild.SystemdUnitStageOptions
-	Authselect    *osbuild.AuthselectStageOptions
-	SELinuxConfig *osbuild.SELinuxConfigStageOptions
-	Tuned         *osbuild.TunedStageOptions
-	Tmpfilesd     []*osbuild.TmpfilesdStageOptions
-	PamLimitsConf []*osbuild.PamLimitsConfStageOptions
-	Sysctld       []*osbuild.SysctldStageOptions
-	DNFConfig     []*osbuild.DNFConfigStageOptions
-	SshdConfig    *osbuild.SshdConfigStageOptions
-	AuthConfig    *osbuild.AuthconfigStageOptions
-	PwQuality     *osbuild.PwqualityConfStageOptions
+	Grub2Config    *osbuild.GRUB2Config
+	Sysconfig      []*osbuild.SysconfigStageOptions
+	SystemdLogind  []*osbuild.SystemdLogindStageOptions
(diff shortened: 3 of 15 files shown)


kingsleyzissou:
Yeah that's a good point. I was wondering if it would be better to do as you suggested or just have a predefined set of options once you specified the security benchmark. I will iterate over this and try something like that out

internal/distro/fedora/distro.go
```diff
@@ -631,6 +641,17 @@ func (t *imageType) initializeManifest(bp *blueprint.Blueprint,
...
 	}
+
+	// if oscap customizations are enabled the
+	// image requires additional packages
+	if bp.Customizations.GetOpenSCAP() != nil {
+		oscapPackages := []string{
+			"openscap-scanner",
+			"scap-security-guide",
+		}
+		w.Packages = append(w.Packages, oscapPackages...)
```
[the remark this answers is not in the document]
kingsleyzissou: Nvm found where the `manifest.getPackageSetChain()` function

internal/osbuild2/osap_remediation_stage.go
```diff
@@ -0,0 +1,46 @@
...
+			HtmlReport:   options.HtmlReport,
+			VerboseLog:   options.VerboseLog,
+			VerboseLevel: options.VerboseLevel,
+		},
+	}
+}
+
+type OscapConfig struct {
+	Datastream   string `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	ProfileID    string `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
```
achilleas-k, line 36: These are required by the stage in osbuild so they shouldn't be marked `omitempty`.

internal/osbuild2/osap_remediation_stage.go
```diff
@@ -0,0 +1,46 @@
...
+	DatastreamID string `json:"datastream_id,omitempty" toml:"datastream_id,omitempty"`
+	XCCDFID      string `json:"xccdf_id,omitempty" toml:"xccdf_id,omitempty"`
+	BenchmarkID  string `json:"benchmark_id,omitempty" toml:"benchmark_id,omitempty"`
+	Tailoring    string `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
+	TailoringID  string `json:"tailoring_id,omitempty" toml:"tailoring_id,omitempty"`
+	ArfResult    string `json:"arf_result,omitempty" toml:"arf_result,omitempty"`
+	HtmlReport   string `json:"html_report,omitempty" toml:"html_report,omitempty"`
+	VerboseLog   string `json:"verbose_log,omitempty" toml:"verbose_log,omitempty"`
+	VerboseLevel string `json:"verbose_level,omitempty" toml:"verbose_level,omitempty"`
```
achilleas-k, line 45: We could validate the values for this since it's an enum.
Have a look at the tar stage.
Validation like this can be hard (and annoying) to write and read, and we're not consistent about it, but I think we should try to add it for new stages.

internal/oscap/main.go
```diff
@@ -0,0 +1,45 @@
+package oscap
```
achilleas-k, line 1: The file shouldn't be called `main.go` unless it has a `main()` function.  Typically the main (or only) file in a package is named after the package, so `oscap.go`.

internal/oscap/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+import "strings"
+
+type Profile string
+
+func (p Profile) toString() string {
```
achilleas-k, line 7: If you rename this method to `String()`, it gets called automatically when you format it with `fmt.Print()` and its variants (it implements the [Stringer](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) interface), which could be useful.

test/cases/oscap.sh
```diff
@@ -0,0 +1,453 @@
...
+        sudo virsh destroy "${VIRSH_DOMAIN}"
+    fi
+    sudo virsh undefine "${VIRSH_DOMAIN}" --nvram
+
+    # Remove qcow2 file.
+    sudo virsh vol-delete --pool images "${LIBVIRT_IMAGE_PATH}"
+}
+
+# Test result checking
+check_result () {
```
achilleas-k, line 234: Since this function is only called at the end, can we just move it tot he end of the script, outside the function so that it can be read in order?

achilleas-k (changes requested):
All good in general.  I have some suggestions below, but nothing major.

achilleas-k:
Oh, one more thing.
Could you rebase on `main` now?  There aren't any conflicts, but since #2805 was merged, we might get some fun side effects with 8.4 in rhel8.

internal/oscap/main.go
kingsleyzissou: Ah cool thanks! I was wondering about that `main.go` didn't seem right, good to know about the convention.

kingsleyzissou:
Thanks for the reveiw @achilleas-k, I'll rebase (I intentionally left `8.4` alone for that reason) and address the comments.

achilleas-k (approved):
LGTM

achilleas-k:
Did something change or was this just a rebase on `main`?

kingsleyzissou:
There was a merge conflict. So I rebased. I also updated the fedora `aarch` repos because I suspect those were breaking some of the base tests.

achilleas-k (approved):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 6d32cc672476c018f054ea3875ff4efd2c0e48e6 with the main merge-base 2eb9833370b6d24128ac4d511f91104e900315da).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

kingsleyzissou:
Following some feedback from a call with the OpenSCAP team I made some slight changes to the PR:
- field for the `datastream` in the blueprints (since there can be multiple per distro)
- Limited the tests to `RHEL 8` & `RHEL-9`

test/cases/oscap.sh
```diff
@@ -0,0 +1,434 @@
...
+# Check for ssh ready to go.
+greenprint "🛃 Checking for SSH is ready to go"
+for LOOP_COUNTER in $(seq 0 45); do
+    RESULTS="$(wait_for_ssh_up ${HTTP_GUEST_ADDRESS})"
+    if [[ ${RESULTS} == 1 ]]; then
+        echo "SSH is ready now! 🥳"
+        break
+    fi
+    sleep 20
+done
```
atodorov, line 308: What happens if SSH never becomes ready? I don't see handling that error condition or am I missing something?

test/cases/oscap.sh
```diff
@@ -0,0 +1,434 @@
...
+# shellcheck disable=SC2034  # Unused variables left for readability
+for LOOP_COUNTER in $(seq 0 45); do
+    RESULTS="$(wait_for_ssh_up ${HTTP_GUEST_ADDRESS})"
+    if [[ ${RESULTS} == 1 ]]; then
+        echo "SSH is ready now! 🥳"
+        break
+    fi
+    # higher timeout for first-boot remediation to complete
+    sleep 20 
+done
```
atodorov, line 409: Similar comment for SSH never becoming ready. 

Also this snippet could probably become a function.

atodorov (changes requested):
Mostly questions and minor changes requested.

internal/blueprint/customizations.go
```diff
@@ -90,6 +91,12 @@ type FilesystemCustomization struct {
 	MinSize    uint64 `json:"minsize,omitempty" toml:"size,omitempty"`
 }
 
+type OpenSCAPCustomization struct {
+	DataDir    string `json:"data_dir,omitempty" toml:"data_dir,omitempty"`
```
teg, line 95: Why is this configurable? Could we hardcode a value?

teg (changes requested):

internal/blueprint/customizations.go
kingsleyzissou: Ah yeah we could hardcode the value here. Do you have any suggestions? Maybe /var/log/openscap_data?

teg: Maybe use our own namespace? /var/log/osbuild/openscap? @gicmo thoughts?

gicmo: Yeah, that sounds good!

internal/manifest/os.go
```diff
@@ -151,6 +151,10 @@ func (p *OS) getPackageSetChain() []rpmmd.PackageSet {
 		packages = append(packages, "chrony")
 	}
 
+	if p.OpenSCAPConfig != nil {
+		packages = append(packages, "openscap-scanner", "scap-security-guide")
```
teg, line 155: Do these have to be installed on the image, is it not sufficient to have them in the build root?

internal/distro/fedora/distro.go
```diff
@@ -43,6 +43,9 @@ const (
 
 	//Kernel options for ami, qcow2, openstack, vhd and vmdk types
 	defaultKernelOptions = "ro no_timer_check console=ttyS0,115200n8 biosdevname=0 net.ifnames=0"
+
+	// Directory for saving openscap results
+	oscapDataDir = "/var/log/osbuild/openscap_data"
```
teg, line 48: Are we sure we want to save this to the image?

kingsleyzissou: Yeah sorry, didn't get the chance to push the changes earlier.

internal/manifest/os.go
kingsleyzissou: So the `openscap` team said that customers would want to run scans again on the image after it has been built.
But to answer the question, no it's not necessary and we could add them in the build root.

internal/distro/fedora/distro.go
teg: Ah, sorry, misread the timestamps.

internal/manifest/os.go
teg: I think we should make it optional in the image, and let people install it if they want to. I think, but might be wrong, that you can run scans through satellite and insights without anything installed. Might be worth checking.

teg (commented):
This looks great to me, but I'll leave the final determination up to @gicmo

test/cases/oscap.sh
```diff
@@ -0,0 +1,497 @@
...
+
+# compare floating point numbers
+if python3 -c "exit(${HARDENED_SCORE} > ${BASELINE_SCORE})"; then
+  greenprint "❌ Failed"
+  echo "Hardened image did not improve baseline score"
+  exit 1
+fi
+
+# one grub rule fails (expected)
+# check if any other rules have failed
```
gicmo, line 490: Is there a bug open about this? would be great to note it.

test/cases/oscap.sh
```diff
@@ -0,0 +1,497 @@
...
+
+# Remomve tmp dir.
+sudo rm -rf "${TEMPDIR}"
+
+greenprint "🎏 Checking for test result"
+echo "Baseline score: ${BASELINE_SCORE}%"
+echo "Hardened score: ${HARDENED_SCORE}%"
+
+# compare floating point numbers
+if python3 -c "exit(${HARDENED_SCORE} > ${BASELINE_SCORE})"; then
```
gicmo, line 483: So I guess a tiny bit better is still ok?

gicmo (approved):
👍 

All great. I have two comments about the testing but nothing to block on!

test/cases/oscap.sh
kingsleyzissou: Yeah I guess this can be updated in a follow up - I figured with checking the rest results that had high severity this was an added bonus. We could maybe improve this by checking that it's at least x% better or maybe remove the baseline and check that the score is greater than like 80/90%

test/cases/oscap.sh
kingsleyzissou: I don't think it's a bug:
```
warning  To prevent hard-coded passwords, automatic remediation of this control is not available.  
Remediation must be  automated as a component of machine provisioning, or followed manually  
as outlined above. Also, do NOT manually add the superuser account and password to the grub.cfg  
file as the grub2-mkconfig command overwrites this file.
```

gicmo: Ah, ok. I thought it was an actual failure.

thozza:
This PR broke CI (https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2815462708) due to not being rebased on top of https://github.com/osbuild/osbuild-composer/pull/2853 I'll submit a fix.

thozza:
I added the fix into https://github.com/osbuild/osbuild-composer/pull/2880

PR #2697, opened 2022-05-27
cloudapi: Drop bucket from composer config

Gundersanne:
This value is set in the worker config. In future it might also be
passed through the api to upload into target accounts, but it should
never be set in composer.


ondrejbudai (dismissed):
Yes, please!

thozza (approved):
LGTM 👍  Thanks!

PR #2705, opened 2022-05-30
[rhel7] add initial support (qcow2, vhd, azure-rhui)

gicmo:
Based on the RHEL 8.6 pipelines, needs a special buildroot with two
extra packages: python3-iniparse and python3-PyYAML.
Only x86_64 support for now.

This is based on to of #2537

### DIFF
--- a/internal/distro/rhel7/distro.go
+++ b/internal/distro/rhel7/distro.go
@@ -493,6 +493,8 @@ func newDistro(distroName string) distro.Distro {
 
 	x86_64.addImageTypes(
 		qcow2ImgType,
+		vhdImgType,
+		azureRhuiImgType,
 	)
 
 	rd.addArches(
--- a/internal/distro/rhel7/distro.go
+++ b/internal/distro/rhel7/distro.go
@@ -1,0 +1,503 @@
+package rhel7
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand"
+	"path"
+	"sort"
+	"strings"
+
+	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/disk"
+	"github.com/osbuild/osbuild-composer/internal/distro"
+	osbuild "github.com/osbuild/osbuild-composer/internal/osbuild2"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+)
+
+const (
+	// package set names
+
+	// build package set name
+	buildPkgsKey = "build"
+
+	// main/common os image package set name
+	osPkgsKey = "packages"
+
+	// blueprint package set name
+	blueprintPkgsKey = "blueprint"
+)
+
+var mountpointAllowList = []string{
+	"/", "/var", "/opt", "/srv", "/usr", "/app", "/data", "/home", "/tmp",
+}
+
+// RHEL-based OS image configuration defaults
+var defaultDistroImageConfig = &distro.ImageConfig{
+	Timezone: "America/New_York",
+	Locale:   "en_US.UTF-8",
+	GPGKeyFiles: []string{
+		"/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release",
+	},
+	Sysconfig: []*osbuild.SysconfigStageOptions{
+		{
+			Kernel: &osbuild.SysconfigKernelOptions{
+				UpdateDefault: true,
+				DefaultKernel: "kernel",
+			},
+			Network: &osbuild.SysconfigNetworkOptions{
+				Networking: true,
+				NoZeroConf: true,
+			},
+		},
+	},
+}
+
+// distribution objects without the arches > image types
+var distroMap = map[string]distribution{
+	"rhel-7": {
+		name:               "rhel-7",
+		product:            "Red Hat Enterprise Linux",
+		osVersion:          "7.9",
+		nick:               "Maipo",
+		releaseVersion:     "7",
+		modulePlatformID:   "platform:el7",
+		vendor:             "redhat",
+		runner:             "org.osbuild.rhel7",
+		defaultImageConfig: defaultDistroImageConfig,
+	},
+}
+
+// --- Distribution ---
+type distribution struct {
+	name               string
+	product            string
+	nick               string
+	osVersion          string
+	releaseVersion     string
+	modulePlatformID   string
+	vendor             string
+	runner             string
+	arches             map[string]distro.Arch
+	defaultImageConfig *distro.ImageConfig
+}
+
+func (d *distribution) Name() string {
+	return d.name
+}
+
+func (d *distribution) Releasever() string {
+	return d.releaseVersion
+}
+
+func (d *distribution) ModulePlatformID() string {
+	return d.modulePlatformID
+}
+
+func (d *distribution) OSTreeRef() string {
+	return "" // not supported
+}
+
+func (d *distribution) ListArches() []string {
+	archNames := make([]string, 0, len(d.arches))
+	for name := range d.arches {
+		archNames = append(archNames, name)
+	}
+	sort.Strings(archNames)
+	return archNames
+}
+
+func (d *distribution) GetArch(name string) (distro.Arch, error) {
+	arch, exists := d.arches[name]
+	if !exists {
+		return nil, errors.New("invalid architecture: " + name)
+	}
+	return arch, nil
+}
+
+func (d *distribution) addArches(arches ...architecture) {
+	if d.arches == nil {
+		d.arches = map[string]distro.Arch{}
+	}
+
+	// Do not make copies of architectures, as opposed to image types,
+	// because architecture definitions are not used by more than a single
+	// distro definition.
+	for idx := range arches {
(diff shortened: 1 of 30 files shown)


internal/distro/rhel7/distro_test.go
```diff
@@ -0,0 +1,445 @@
...
+	for _, archName := range r7distro.ListArches() {
+		arch, _ := r7distro.GetArch(archName)
+		for _, imgTypeName := range arch.ListImageTypes() {
+			imgType, _ := arch.GetImageType(imgTypeName)
+			testPackageSpecSets := distro_test_common.GetTestingImagePackageSpecSets("kernel", imgType)
+			_, err := imgType.Manifest(bp.Customizations, distro.ImageOptions{}, nil, testPackageSpecSets, 0)
+			assert.NoError(t, err)
+		}
+	}
+}
```
thozza, line 465: The logic of all these test cases is exactly the same. I propose to merge them into a single unit test with array of test case blueprints and execute the same testing logic in a `for` cycle on them. We use the same approach in various unit tests in composer.

internal/distro/rhel7/pipelines.go
```diff
@@ -0,0 +1,335 @@
...
+	p := new(osbuild.Pipeline)
+
+	p.Name = "os"
+	p.Build = "name:build"
+
+	rpmOptions := osbuild.NewRPMStageOptions(repos)
+	rpmOptions.GPGKeysFromTree = imageConfig.GPGKeyFiles
+	p.AddStage(osbuild.NewRPMStage(rpmOptions, osbuild.NewRpmStageSourceFilesInputs(packages)))
+
+	// Difference to RHEL8, 9 pipelines: no BLS stage
```
thozza, line 51: hmm... I wonder how could we keep the `osPipeline` implementations identical across distro definitions (to allow deduplicating it eventually), but still being able to not apply the BLS stage on RHEL-7 🤔 

What about adding `bool` variable to the `ImageConfig` to distinguish whether BLS should be used or not and making it part of the default distro `ImageConfig`?

internal/osbuild2/stage.go
```diff
@@ -128,6 +128,8 @@ func (stage *Stage) UnmarshalJSON(data []byte) error {
 		options = new(TunedStageOptions)
 	case "org.osbuild.sfdisk":
 		options = new(SfdiskStageOptions)
+	case "org.osbuild.sgdisk":
+		options = new(SgdiskStageOptions)
```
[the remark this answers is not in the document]
achilleas-k: Why do we have this?  I'm trying to remember  the purpose of unmarshalling stage options.
It's a bigger conversation, for another time, but I think it was because osbuild would return the options of a stage with the results.  I think it doesn't do that anymore (not for v2 manifests) but also, I don't think we use them anywhere.

Maybe it's time we got rid of this?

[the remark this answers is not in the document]
thozza: Yeah, I didn't check it in detail, but I also think that we do not use this anywhere. +1 for getting rid of the whole UnmarshalJSON method completely 😉

internal/osbuild2/grub2_legacy_stage.go
```diff
@@ -0,0 +1,227 @@
...
+		have = append(have, "`label`")
+	}
+
+	if d.UUID != nil {
+		have = append(have, "`uuid`")
+	}
+
+	count := len(have)
+	if count == 0 {
+		return fmt.Errorf("need `device`, `label` or `uuid`")
```
achilleas-k, line 36: Please add a comma after `label` ;)

internal/distro/rhel7/pipelines.go
achilleas-k: I think we can keep it right now as is, especially since it has this difference, and we can sort it out in the eventual deduplication, rather than adding dead code.

achilleas-k (commented):
Does RHEL 7 need a RHEL 8 build root for qemu?  I seem to recall this was a requirement but I don't see it anywhere.

internal/osbuild2/grub2_legacy_stage.go
gicmo: > Does RHEL 7 need a RHEL 8 build root for qemu? I seem to recall this was a requirement but I don't see it anywhere.

Yeah, I had the same though (also because I made a not w.r.t to that in the corresponding jira ticket). Will double check.

gicmo: Just checked: we have `qemu-img version 1.5.3`, it can do `subformat=fixed` but not `force_size`, which we also use for `vpc`.

```
[gicmo@ckellner-rhel7 ~]$ qemu-img convert -O vpc -o subformat=fixed,force_size disk.qcow2 disk.vpc
Unknown option 'force_size'
qemu-img: Invalid options for file format 'vpc'.
```

gicmo: I think we might need the `force_size` option, see https://repo.or.cz/qemu/kevin.git/blobdiff/798609bbe29df4edf4bbec46927fc12862b02378..fb9245c2610932d33ce148b58714fcc7b3c6eb5f:/block/vpc.c

internal/osbuild2/stage.go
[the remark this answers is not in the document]
gicmo: Done in https://github.com/osbuild/osbuild-composer/pull/2741

internal/osbuild2/grub2_legacy_stage.go
```diff
@@ -0,0 +1,227 @@
...
+	Platform string `json:"platform,"`
+}
+
+type GRUB2LegacyConfig struct {
+	GRUB2Config
+	CmdLine     string `json:"cmdline,omitempty"`
+	Distributor string `json:"distributor,omitempty"`
+}
+
+type GRUB2LegacyStageOptions struct {
```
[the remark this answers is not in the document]
gicmo: Obsolete via https://github.com/osbuild/osbuild-composer/pull/2741

internal/distro/rhel7/distro.go
```diff
@@ -0,0 +1,542 @@
...
+		for i := 0; !haveNewMountpoint && i < len(fs); i++ {
+			haveNewMountpoint = !pt.ContainsMountpoint(fs[i].Mountpoint)
+		}
+	}
+
+	if haveNewMountpoint {
+		bpPackages = append(bpPackages, "lvm2")
+	}
+
+	// depsolve bp packages separately
```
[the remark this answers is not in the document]
gicmo: Happy to remove it; it is also present in other distros though, e.g. RHEL9: https://github.com/osbuild/osbuild-composer/blob/460a85f245f4be07f508679039a1b6d007aad1fe/internal/distro/rhel90/distro.go#L353

internal/distro/rhel7/distro.go
```diff
@@ -0,0 +1,542 @@
...
+	// main/common os image package set name
+	osPkgsKey = "packages"
+
+	// blueprint package set name
+	blueprintPkgsKey = "blueprint"
+)
+
+var mountpointAllowList = []string{
+	"/", "/var", "/opt", "/srv", "/usr", "/app", "/data", "/home", "/tmp",
+}
```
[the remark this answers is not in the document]
gicmo: Yeah, I think it makes sense to have this be the same across all distros, so that you can re-use the same blueprint across all of them, and since we use LVM it should be the same everywhere. Just wondering if I should not do this in a follow up instead.

internal/distro/rhel7/distro_test.go
gicmo: This is a copy & paste from all the other distros. Probably should be moved to a common place with all the other copies once we move the mount point handling to a common place as well.

internal/distro/rhel7/distro.go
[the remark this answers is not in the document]
thozza: Yes, I noticed that. I thought that it would be nice to remove it if we are adding new code, but didn't want to ask you to fix all distros... It is a nitpick so leaving it as is is also acceptable 😉

internal/distro/rhel7/distro.go
```diff
@@ -0,0 +1,542 @@
...
+		ils := osbuild.NewInlineSource()
+		for _, data := range inlineData {
+			ils.AddItem(data)
+		}
+
+		sources["org.osbuild.inline"] = ils
+	}
+
+	return sources
+}
```
thozza, line 472: This should be removed in favor of `osbuild2.GenSources()`

thozza (commented):
the `sources()` function in the `rhel-7` distro definition should be removed, the rest looks good 👍

internal/osbuild2/grub2_legacy_stage.go
gicmo: For future reference. Old `qemu-img` does not need the `force_size` option, so we made that optional in osbuild.

internal/distro/rhel7/distro.go
gicmo: Ah, cool, yes! Done!

achilleas-k (approved):
LGTM

PR #2707, opened 2022-05-31
Service maintenance: Delete results from manifest and depsolve jobs

Gundersanne:


internal/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -657,59 +645,20 @@ func (q *DBJobQueue) JobsUptoByType(jobTypes []string, upto time.Time) (result m
 	return
 }
 
-// Deletes single job and dependencies (recursively)
-func (q *DBJobQueue) DeleteJobIncludingDependencies(jobId uuid.UUID) error {
+func (q *DBJobQueue) DeleteJobResult(jobId uuid.UUID) error {
```
ondrejbudai, line 648: What about changing this method to accept a slice of job ids? Nullifying all of them in one query might be faster.

cmd/osbuild-service-maintenance/db.go
```diff
@@ -10,37 +10,34 @@ import (
...
 		return err
 	}
 
-	var jobTypes []string
-	for _, a := range archs {
-		jobTypes = append(jobTypes, fmt.Sprintf("%s:%s", jobType, a))
-	}
-
-	jobsByType, err := jobs.JobsUptoByType(jobTypes, cutoff)
+	// The results of these jobs take up the most space and can contain sensitive data. Delete
```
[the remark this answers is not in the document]
Gundersanne: True, but the result data might be useful long term and shouldn't contain sensitive info like activation keys. There's also business metrics things coming up where we might need to know if a job succeeded or not.

To conserve space on stage we should just delete them (after #2712) imo, but we can also do that via scheduled queries in app-intf rather than the maintenance, could also add it here of course.

internal/jobqueue/dbjobqueue/dbjobqueue.go
Gundersanne: True, I just wondered how big the slice could be :|

Gundersanne: ok I looked it up and it's about 1GB of text, that's a lot of uuids. I'll change this :D

cmd/osbuild-service-maintenance/db.go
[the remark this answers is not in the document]
ondrejbudai: Sounds good, thanks for explaining.

ondrejbudai (approved):

PR #2708, opened 2022-05-31
distro/rhel90: remove skx_edac, intel_cstate from denylist again

gicmo:
In commit 5c1530e we disabled `skx_edac` and `intel_cstate` but after further consultation with Prarit Bhargava it was agreed that  for RHEL 9 we should indeed allow them.


teg (approved):

major (approved):
This change should be super helpful for bare metal cloud instances and shouldn't impact virtualized instances in any meaningful ways.

PR #2721, opened 2022-06-03
dbjobqueue: use background context when closing listener

lzap:
Hello,

when `Close()` is called in all cases it ends up with "Error unlistening for jobs in dequeue" because in the defer block then driver performs `Exec` call with the caller's context, it is already done.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer-dbjobqueue-tests/main_test.go
+++ b/cmd/osbuild-composer-dbjobqueue-tests/main_test.go
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
 
 package main
--- a/pkg/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/pkg/jobqueue/dbjobqueue/dbjobqueue.go
@@ -182,7 +182,8 @@ func (q *DBJobQueue) listen(ctx context.Context, ready chan<- struct{}) {
 		panic(fmt.Errorf("error connecting to database: %v", err))
 	}
 	defer func() {
-		_, err := conn.Exec(ctx, sqlUnlisten)
+		// use the empty context as the listening context is already cancelled at this point
+		_, err := conn.Exec(context.Background(), sqlUnlisten)
 		if err != nil && !errors.Is(err, context.DeadlineExceeded) {
 			logrus.Error("Error unlistening for jobs in dequeue: ", err)
 		}


cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -64,6 +64,7 @@ func TestJobQueueInterface(t *testing.T) {
 
 	t.Run("maintenance-query-jobs-before", wrap(testJobsUptoByType))
 	t.Run("maintenance-delete-job-results", wrap(testDeleteJobResult))
+	t.Run("close", wrap(testClose))
```
lzap, line 67: This needs to be the last test in the suite, I am lazy yeah. Let me know how to implement this in a nicer way, perhaps to reopen the connection afterwards? :)

ondrejbudai: Two ideas:

- put a comment before this subtest to indicate that it must be run last (lazier, gets the job done)
- make a completely separate test-case that spins up its own queue and then immediately close it (cleaner)

I'm fine with both. :)

ondrejbudai (changes requested):

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
lzap: I was thinking exactly the same. Going for the cleanest solution.

I cannot figure out how to actually run those tests (I am still new to Go), so force-pushed blindly with fingers-crossed. If it fails, teach me please how to run them (this appear to be some kind of test suite, how do I trigger it)?

lzap:
Workflow still awaiting approval, some RPM builds failed as well.

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -144,3 +148,8 @@ func testDeleteJobResult(t *testing.T, q *dbjobqueue.DBJobQueue) {
 	require.NoError(t, err)
 	require.Nil(t, r)
 }
+
+func testClose(t *testing.T, q *dbjobqueue.DBJobQueue) {
+	err := q.Close()
```
ondrejbudai, line 153: `q.Close()` doesn't return any value, does it?

lzap: Right, yeah. Hmmm, to test that I am going to check if a subsequent call returns an error. That's the only idea I have other than making something exported to check if the connection was closed.

ondrejbudai:
Hi @lzap, do you have any plans to get this finished? I'm currently in the process of cleaning up stale PRs. :) Thanks!

lzap:
Apologies, so I rebased this. I can't figure out how to test this, the error happens in the worker goroutine. On top of that, `Close` function is not part of the common interface. So I changed the patch so it only fixes the error, it's a cosmetic thing.

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
```
lzap, line 2: My IDE insists on this, I did not fight it - I see few places where you still have this.

ondrejbudai: Sure, `+build` will get removed when we switch to Go 1.18 completely (we cannot yet due to Fedora 35).

ondrejbudai (approved):

ondrejbudai:
Just an unrelated OSTree failure, merging.

PR #2723, opened 2022-06-06

ygalblum:
# RPM Spec
- Remove all Go dependecies and replace with bundling indication
- Add '-k' to goprep to preserve the vendor directory
    
# tools
- Add script to update the RPM spec file to generate the indication lines
    
# Makefile
- Add target for running the script
    
# GitHub Actions
- Add CI check to make sure the bundling lines are up to date
- Add a workflow to update the RPM spec when the go.mod changes

### DIFF
--- a/.github/workflows/rpm-spec-bundle.yml
+++ b/.github/workflows/rpm-spec-bundle.yml
@@ -1,0 +1,38 @@
+name: "Update Provides bundle list in the RPM spec"
+
+on:
+  pull_request:
+    branches:
+      - "*"
+    paths:
+      - "go.mod"
+
+jobs:
+  rpm-spec-bundle:
+    name: "Update the RPM Spec with the new provides bundle list"
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Checkout PR
+        uses: dawidd6/action-checkout-pr@v1
+        with:
+          pr: ${{ github.event.pull_request.number }}
+
+      - name: Update the the packages list
+        run: make rpm-spec-bundle
+
+      - name: Push changes if needed
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ -n "$(git status --porcelain osbuild-composer.spec)" ]; then
+            git config user.name github-actions
+            git config user.email github-actions@github.com
+            git add osbuild-composer.spec
+            git commit -m"Update RPM deploy spec"
+            git push
+          else
+            echo ""
+          fi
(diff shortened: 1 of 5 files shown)


ondrejbudai:
In order to get this merged, we have to:

- Remove all go deps from the spec file
- Add bundled provides (see https://src.fedoraproject.org/rpms/podman/blob/8d5f758a1d3b695815945f9daa0907c9c2977aa4/f/podman.spec#_82 )
  - This must be enforced by upstream CI otherwise we will go mad.
- change %goprep to %goprep -k, see https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/#_prep_goprep_removing_vendoring

(copied from https://issues.redhat.com/browse/COMPOSER-1608, RH-internal only)

ygalblum:
@ondrejbudai How do you envision item 2? If I understand correctly, this means that every time dependabot updates a package, the RPM spec file needs to change as well. Do you agree? 
Furthermore, Is this a must? I mean, and correct me if I am wrong here, does vendoring really bring in the package to the system? Does it not only compile it into the executable?

ondrejbudai:
> @ondrejbudai How do you envision item 2? If I understand correctly, this means that every time dependabot updates a package, the RPM spec file needs to change as well. Do you agree?

Yes, this sucks. I wonder if we can do this when pushing the spec file downstream. Although, this would break the rule that we ship the exactly same specfile to both up- and downstream.

> Furthermore, Is this a must? I mean, and correct me if I am wrong here, does vendoring really bring in the package to the system? Does it not only compile it into the executable?

If I understand it correctly, this is enforced by the packaging guidelines: https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling The important sentence is this one:

> All packages whose upstreams have no mechanism to build against system libraries MAY opt to carry bundled libraries, but if they do, they MUST include an indication of what they bundle.

cc @gicmo

ygalblum:
Thanks for the links. 
Reading the packaging guidelines, I think that their use of the word "Provides" here is misleading because this RPM does not provide any of these libraries to the system on which it is being installed. They are actually being statically compiled into the executable. Instead, they want to get the list of these libraries in order to track possible issues. So, I guess, they wanted to reuse a keyword.
In any case, yes, this means that all dependabot PRs will have to have an additional manual step before being merged. In addition, we will have to make sure the list is up to date in the CI.
So, I think this PR will require the following scripts:
1. Generate the lines listing the packages from the go.mod file into the RPM
2. CI - Verify that the list of packages in go.mod are all listed in the RPM spec file

ondrejbudai:
We should also probably think of licensing in the spec file. Since we are building everything from the SRPM, I think we have to change the license field to include licenses of all deps. WDYT?

ygalblum:
> We should also probably think of licensing in the spec file. Since we are building everything from the SRPM, I think we have to change the license field to include licenses of all deps. WDYT?

Looking at the Podman example, I don't think so. It will turn into quite a rabbit hole

gicmo:
Will defer to @ondrejbudai but it looks good to me!

PR #2727, opened 2022-06-07
osbuild-service-maintenance:  vacuum  analyze after update

Gundersanne:
A few notes:

1. nulling results doesn't clear disk space without vacuum full, in fact it increases disk usage of the table
2. Vacuum analyze does free up space within the table, meaning there's room for additional inserts without disk space increasing.

---

exapmle stats: https://bpa.st/raw/X6CA


ondrejbudai (approved):
Let's go!

PR #2728, opened 2022-06-07
Migrate scheduled cloud cleaner to separate repo

jabia99:
This pull request includes:


jrusz (approved):
LGTM. 

Just a thought, it might be easier to just `curl` the raw url of the `cloud-cleaner.sh` in this case since it's just one file and the URL won't change. The way it is now it will always clone the whole repository with it's history so it would be noticeable slower if the target repo was bigger :) But don't worry about it in this case, just something I wanted to share :)

Issue #2446, opened 2022-03-15
RFE: make dnf-json cache garbage collection size rather than time based


teg:
Currently `dnf-json` cleans up caches that are older than 24 hours. This is a good approximation, but we risk deleting useful caches too soon as well as not deleting large caches soon enough, depending on the usage patterns.

Suggestion:
After each depsolve, delete the least recently used cache entries until the total disk usage of the caches fall below a certain threshold.

Implementation suggestion:
 - Maintain a least recently used list containing the total space usage of each cache entry and its location.
 - In `update_used()`, `touch()` the directory entry on disk and update the LRU (time and size).
 - In `clean_unused()` delete the least recently used entries until the total usage falls below a configured threshold.
 - In `CacheState.load()` instantiate the LRU by iterating the cache entries in the cache dir (if it exists), read the `mtime` from the dirent and sum up the file sizes.
 - Drop `CacheState.store()`.

teg:
@lavocatt @achilleas-k @gicmo @Gundersanne @ondrejbudai (and I'm sure others), any thoughts on this? I was having a look at `dnf-json` to better understand its performance and thinking ahead for supporting both Fedora and 3rd party repos where the behaviour might be quite different from what we saw so far.

achilleas-k:
I think this makes sense, especially for on-prem situations where a user might not be building things that frequently.

PR #2733, opened 2022-06-09
Size-based cleanup for dnf-json cache directories

achilleas-k:
This PR implements size-based garbage collection for the dnf-json cache directory, based on the outline in https://github.com/osbuild/osbuild-composer/issues/2446.

The workflow is as follows:
- Update the `mtime` for a repository's cache on-disk using `touchRepo()` whenever it's used in a transaction (`Depsolve()` or `FetchMetadata()`).
- Update the internal cache information when needed by calling `updateInfo()` (see note below about the frequency of this call).
- Shrink the cache to below the configured `maxSize` by calling `shrink()`.

The most important work happens in `updateInfo()`.  It collects all the information it needs from the on-disk cache directories and organises it in a way that makes it convenient for the `shrink()` function to run efficiently.  It stores three important pieces of information:
1. `repoElements`: a map that links a repository ID with all the information about a repository's cache:
    - the top-level elements (files and directories) for the cache
    - size of the repository cache (total of all elements)
    - most recent `mtime` from all the elements which, if the `touchRepo()` call is consistently used, should reflect the most recent time the repository was used
2. `repoRecency`: a list of repository IDs sorted by `mtime` (oldest first)
3. size: the total size of the cache (total of all repository caches)

This way, when `shrink()` is called, the paths associated with the least-recently-used repositories can be easily deleted by iterating on `repoRecency`, obtaining the repository info from the map, deleting every path in the `repoElements` array, and subtracting the repository's `size` from the total.  The `shrink()` function stops when the new size is below the `maxSize` (or when all repositories have been deleted).

## Locks

Each cache instance has a RWMutex lock.  A global map of cache locks is maintained, keyed by the absolute path to the cache directory, so multiple cache instances can coexist and share locks if they use the same cache root.

The RWMutex lock allows for multiple concurrent read locks to be held or one exclusive write lock.  A read lock is acquired for every transaction and a write lock is acquired when deleting caches.
The _read lock_ is a bit of a misnomer since a transaction can (and will) write files to an individual repository's metadata cache.  Since `dnf` takes care of locking individual repository caches when in use, we don't need an exclusive lock for these transactions.
The _write lock_ prevents any transactions from running, regardless of repository, when a cleanup (`shrink()`) is being run.

### No interprocess locking

I considered (and [implemented](https://github.com/achilleas-k/osbuild-composer/tree/dnfjson/size-based-cleanup-file-locks)) file-based locking of the cache with the same logic as the RWMutex.  This worked and provided locks across processes, but required more work to release locks when osbuild-composer was interrupted or killed.  I think it's fine to just _require_ (as a policy) that multiple potential instances of osbuild-composer, osbuild-worker, or perhaps a future dnf-json service don't share a cache.

## Note on `updateInfo()`

The function is called after every transaction, which might seem like a lot of work to do every time `Depsolve()` is called, especially when running multiple transactions successively for a single manifest.  I considered at some point keeping an internal timestamp (`lastUpdated`) of the last time the info was updated (the `mtime` and size).  The idea was that if the `lastUpdated` was before `info.mtime`, then the size data might be outdated and it should be reread.
I reconsidered because this felt like excessive book-keeping.  In reality, updating the size info consists of a handful of stat calls per repo and, on top of that, the system caches the stat info anyway, so we can let the system keep it efficient for us.

### DIFF
--- a/internal/dnfjson/cache.go
+++ b/internal/dnfjson/cache.go
@@ -136,7 +136,13 @@ func (r *rpmCache) shrink() error {
 	nDeleted := 0
 	for idx := 0; idx < len(r.repoRecency) && r.size >= r.maxSize; idx++ {
 		repoID := r.repoRecency[idx]
-		repo := r.repoElements[repoID]
+		nDeleted++
+		repo, ok := r.repoElements[repoID]
+		if !ok {
+			// cache inconsistency?
+			// ignore and let the ID be removed from the recency list
+			continue
+		}
 		for _, gPath := range repo.paths {
 			if err := os.RemoveAll(gPath); err != nil {
 				return err
@@ -144,7 +150,6 @@ func (r *rpmCache) shrink() error {
 		}
 		r.size -= repo.size
 		delete(r.repoElements, repoID)
-		nDeleted++
 	}
 
 	// update recency list
--- a/internal/dnfjson/cache.go
+++ b/internal/dnfjson/cache.go
@@ -6,10 +6,14 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/gobwas/glob"
 )
+
+// global cache locker
+var cacheLocks sync.Map
 
 // A collection of directory paths, their total size, and their most recent
 // modification time.
@@ -34,14 +38,28 @@ type rpmCache struct {
 
 	// max cache size
 	maxSize uint64
+
+	// locker for this cache directory
+	locker *sync.RWMutex
 }
 
 func newRPMCache(path string, maxSize uint64) *rpmCache {
+	absPath, err := filepath.Abs(path) // convert to abs if it's not already
+	if err != nil {
+		panic(err) // can only happen if the CWD does not exist and the path isn't already absolute
+	}
+	path = absPath
+	locker := new(sync.RWMutex)
+	if l, loaded := cacheLocks.LoadOrStore(path, locker); loaded {
+		// value existed and was loaded
+		locker = l.(*sync.RWMutex)
+	}
 	r := &rpmCache{
 		root:         path,
 		repoElements: make(map[string]pathInfo),
 		size:         0,
 		maxSize:      maxSize,
+		locker:       locker,
 	}
 	// collect existing cache paths and timestamps
 	r.updateInfo()
@@ -121,6 +139,9 @@ func (r *rpmCache) updateInfo() {
 }
 
 func (r *rpmCache) shrink() error {
+	r.locker.Lock()
+	defer r.locker.Unlock()
+
 	// start deleting until we drop below r.maxSize
 	nDeleted := 0
 	for idx := 0; idx < len(r.repoRecency) && r.size >= r.maxSize; idx++ {
--- a/internal/dnfjson/cache.go
+++ b/internal/dnfjson/cache.go
@@ -6,7 +6,22 @@ import (
 	"path/filepath"
 )
 
-func dirSize(path string) (uint64, error) {
+type rpmCache struct {
+	// root path for the cache
+	root string
+
+	// max cache size
+	maxSize uint64
+}
+
+func newRPMCache(path string, maxSize uint64) *rpmCache {
+	return &rpmCache{
+		root:    path,
+		maxSize: maxSize,
+	}
+}
+
+func (r *rpmCache) size() (uint64, error) {
 	var size uint64
 	sizer := func(path string, info fs.FileInfo, err error) error {
 		if err != nil {
@@ -15,17 +30,17 @@ func dirSize(path string) (uint64, error) {
 		size += uint64(info.Size())
 		return nil
 	}
-	err := filepath.Walk(path, sizer)
+	err := filepath.Walk(r.root, sizer)
 	return size, err
 }
 
-func (bs *BaseSolver) CleanCache() error {
-	curSize, err := dirSize(bs.cacheDir)
+func (r *rpmCache) clean() error {
+	curSize, err := r.size()
 	if err != nil {
 		return err
 	}
-	if curSize > bs.maxCacheSize {
-		return os.RemoveAll(bs.cacheDir)
+	if curSize > r.maxSize {
+		return os.RemoveAll(r.root)
 	}
 	return nil
 }
--- a/internal/dnfjson/cache.go
+++ b/internal/dnfjson/cache.go
@@ -1,27 +1,186 @@
 package dnfjson
 
 import (
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
+	"time"
+
+	"github.com/gobwas/glob"
 )
+
+// A collection of directory paths, their total size, and their most recent
+// modification time.
+type pathInfo struct {
+	paths []string
+	size  uint64
+	mtime time.Time
+}
 
 type rpmCache struct {
 	// root path for the cache
 	root string
+
+	// individual repository cache data
+	repoElements map[string]pathInfo
+
+	// list of known repository IDs, sorted by mtime
+	repoRecency []string
+
+	// total cache size
+	size uint64
 
 	// max cache size
 	maxSize uint64
 }
 
 func newRPMCache(path string, maxSize uint64) *rpmCache {
-	return &rpmCache{
-		root:    path,
-		maxSize: maxSize,
+	r := &rpmCache{
+		root:         path,
+		repoElements: make(map[string]pathInfo),
+		size:         0,
+		maxSize:      maxSize,
 	}
+	// collect existing cache paths and timestamps
+	r.updateInfo()
+	return r
 }
 
-func (r *rpmCache) size() (uint64, error) {
+func (r *rpmCache) clean() error {
+	curSize, err := dirSize(r.root)
+	if err != nil {
+		return err
+	}
+	if curSize > r.maxSize {
+		return os.RemoveAll(r.root)
+	}
+	return nil
+}
+
+// updateInfo updates the repoPaths and repoRecency fields of the rpmCache.
+func (r *rpmCache) updateInfo() {
+	cacheEntries, _ := os.ReadDir(r.root)
+
+	// each repository has multiple cache entries (3 on average), so using the
+	// number of cacheEntries to allocate the map and ID slice is a high upper
+	// bound, but guarantees we wont need to grow and reallocate either.
+	repos := make(map[string]pathInfo, len(cacheEntries))
+	repoIDs := make([]string, 0, len(cacheEntries))
+
+	var totalSize uint64
+
+	// Collect the paths grouped by their repo ID
+	// We assume the first 64 characters of a file or directory name are the
+	// repository ID because we use a sha256 sum of the repository config to
+	// create the ID (64 hex chars)
+	for _, entry := range cacheEntries {
+		eInfo, err := entry.Info()
+		if err != nil {
+			// skip it
+			continue
+		}
+
+		fname := entry.Name()
+		if len(fname) < 64 {
+			// unknown file in cache; ignore
+			continue
+		}
+		repoID := fname[:64]
+		repo, ok := repos[repoID]
+		if !ok {
+			// new repo ID
+			repoIDs = append(repoIDs, repoID)
+		}
+		mtime := eInfo.ModTime()
+		ePath := filepath.Join(r.root, entry.Name())
+
+		// calculate and add entry size
+		size, err := dirSize(ePath)
+		if err != nil {
+			// skip it
+			continue
+		}
+		repo.size += size
+		totalSize += size
+
+		// add path
+		repo.paths = append(repo.paths, ePath)
+
+		// if for some reason the mtimes of the various entries of a single
+		// repository are out of sync, use the most recent one
+		if repo.mtime.Before(mtime) {
+			repo.mtime = mtime
+		}
+
+		// update the collection
+		repos[repoID] = repo
+	}
+	sortFunc := func(idx, jdx int) bool {
+		ir := repos[repoIDs[idx]]
+		jr := repos[repoIDs[jdx]]
+		return ir.mtime.Before(jr.mtime)
+	}
+
+	// sort IDs by mtime (oldest first)
+	sort.Slice(repoIDs, sortFunc)
+
+	r.size = totalSize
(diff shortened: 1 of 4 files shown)


internal/dnfjson/cache.go
```diff
@@ -1,46 +1,187 @@
...
+	if curSize > r.maxSize {
+		return os.RemoveAll(r.root)
+	}
+	return nil
+}
+
+// updateInfo updates the repoPaths and repoRecency fields of the rpmCache.
+func (r *rpmCache) updateInfo() {
+	repos := make(map[string]pathInfo)
+	repoIDs := make([]string, 0)
```
thozza, line 64: It would be more efficient to first get `cacheEntries` and use its `len()` when calling `make()` to create `repos` and `repoIDs`. Currently, there is unnecessary resizing happening when adding elements to the slice and map.

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,187 @@
...
+	cacheEntries, _ := os.ReadDir(r.root)
+
+	var totalSize uint64
+
+	// collect the paths grouped by their repo ID (first 64 characters of a file or directory name)
```
thozza, line 69: Could you explain why first 64? ideally in the comment.

thozza (commented):
Very nice work!

I think that the memory allocation could be done more efficiently in `updateInfo()`, but other than that I had only nitpicks.

Happy to approve once we sort out the allocation.

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,187 @@
...
 	return nil
 }
+
+// A collection of directory paths, their total size, and their most recent
+// modification time.
+type pathInfo struct {
+	paths []string
+	size  uint64
+	mtime time.Time
+}
```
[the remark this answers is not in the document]
achilleas-k: yeah good point.  I struggle with ordering these things sometimes.

internal/dnfjson/cache.go
achilleas-k: For repo IDs we use a sha256 of the repository config, which is 64 (hex) chars.  So we can just get the ID based on that assumption.  Another option would be to split the name on `-` and get the first element, in case we ever change the ID, but that might not work for everything.  For example the `.solv` file for a repo is `<ID>.solv`.

Here's an example cache with two repositories:
```
6e791d3334530bbed72986752494e81cc21ede63c4fee9f2b39ef25b879b1b3c-0d0eb10b0c459614/
6e791d3334530bbed72986752494e81cc21ede63c4fee9f2b39ef25b879b1b3c-filenames.solvx
6e791d3334530bbed72986752494e81cc21ede63c4fee9f2b39ef25b879b1b3c.solv
7e7c26ea9923d3a82e7a6049ac567c246e4ee0dcedb9183e01bf8c86e16e66d9-e58e3c9673d413e5/
7e7c26ea9923d3a82e7a6049ac567c246e4ee0dcedb9183e01bf8c86e16e66d9-filenames.solvx
7e7c26ea9923d3a82e7a6049ac567c246e4ee0dcedb9183e01bf8c86e16e66d9.solv
fastestmirror.cache
```

I'm open to suggestions for more robust ways of identifying repo IDs.

internal/dnfjson/cache.go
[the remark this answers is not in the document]
achilleas-k: Done

internal/dnfjson/cache.go
achilleas-k: Considered doing something "smart" like allocating `len(cacheEntries)/3` (which is the average number of entries per cache), but I just went with a plain `len(cacheEntries)` instead.

internal/dnfjson/cache.go
achilleas-k: Added comment.

internal/dnfjson/cache.go
thozza: Yeah, I meant the capacity. I must admit that I missed the fact that there would be more than one cache entries per repo (it was not that obvious from the code at the first sight), but after your comment and connecting the dots with the unit test, I see it now.

thozza (approved):
LGTM 👍

gicmo:
I have only glanced at the code and I think it looks fine!

I just wanted to highlight the behavioural change:  _Currently_ we fill the cache if it "overflows" we clear all of it and let it re-fill. The size-based clearing obviously changes this (intentionally). Let's assume we have N different cache items (say from different RHEL composes) that are all roughly the same size K. N*K is smaller than the limit but (N+1)*K is over it. We would clean up only one cache item and the next new item would kick us over the limit again. That is, we are operating at the cache limit most of the time (more possible cache hits) but also have to do the cleanup more often (depsolve jobs more often will have a small extra latency instead of one bigger one less frequently). I wonder if we could collect metrics on this to better guide us about this, since we could also have an extra parameter that would specify how much below the limit to shrink, i.e. how much we "bounce".

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+	"sort"
+	"time"
+
+	"github.com/gobwas/glob"
 )
 
+// A collection of directory paths, their total size, and their most recent
+// modification time.
+type pathInfo struct {
+	paths []string
```
teg, line 21: For our purposes, could would it make sense to drop this and instead do glob `<repoid>*` at runtime? Not sure it makes a big difference, just always think computing things is better than storing them.

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
 	"path/filepath"
+	"sort"
+	"time"
+
+	"github.com/gobwas/glob"
 )
 
+// A collection of directory paths, their total size, and their most recent
+// modification time.
+type pathInfo struct {
```
teg, line 20: Would it be more intuitive to name this something like `repoCacheEntry` or something like that? The fact that it is (a collection of) path(s) is not really the point is it?

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+}
+
+// updateInfo updates the repoPaths and repoRecency fields of the rpmCache.
+func (r *rpmCache) updateInfo() {
+	cacheEntries, _ := os.ReadDir(r.root)
+
+	// each repository has multiple cache entries (3 on average), so using the
+	// number of cacheEntries to allocate the map and ID slice is a high upper
+	// bound, but guarantees we wont need to grow and reallocate either.
+	repos := make(map[string]pathInfo, len(cacheEntries))
```
teg, line 76: I saw the discussion about his, FTR, I'm pretty sure this is a premature optimisation, as I'm pretty sure map resizing is a noop in most cases.

achilleas-k:
Yeah you make a good point.  A bounce would be a good solution.
I suspect that this might be less of an issue in reality (though we'd have to investigate).  It depends on how fast the repo turnover is.  It's possible that with a big enough `maxSize` (big enough to hold all frequently used repos and some old unused ones), that we'd be cleaning up a repo very infrequently (only when a new set comes in).  But also, what's the situation in prod?  We have the CDNs, which means the same repos get reused for the lifetime of a distro.  If we can keep around all repo caches for say 5 distributions (RHEL 9.0, RHEL 8.6, RHEL 8.5, CS9, CS8), then as soon as RHEL 8.7 gets added, we'd drop 8.5 and nothing changes again for a long time.

This is oversimplified of course because the sizes aren't uniform and I don't know anything about the RHEL vs CS usage patterns.  It might be more likely that CS8 and CS9 get cleaned because they're rarely used maybe?  User defined repositories will probably always be the first to go and then the cache stays stable for 6 months until the next distro arrives.

@Gundersanne can you comment on this?

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+	}
+	if curSize > r.maxSize {
+		return os.RemoveAll(r.root)
+	}
+	return nil
+}
+
+// updateInfo updates the repoPaths and repoRecency fields of the rpmCache.
+func (r *rpmCache) updateInfo() {
+	cacheEntries, _ := os.ReadDir(r.root)
```
teg, line 71: Are any of the files that constitute a cache mandatory? Say the `.solv` ones? Could we simply do a glob over `*.solv` and then we would turn that into a list of repoid's by chopping off the `.solv` postfix? That way we would not have to care about the format/length of the repoid. We could then do the `updateInfo()` call per ID, where we would again glob over `<repoid>*`. Really a nitpick, this is fine.

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+			continue
+		}
+		repo.size += size
+		totalSize += size
+
+		// add path
+		repo.paths = append(repo.paths, ePath)
+
+		// if for some reason the mtimes of the various entries of a single
+		// repository are out of sync, use the most recent one
```
teg, line 119: Same question as above: if any of the entries are mandatory, I think it would be more robust to only touch that one and only read the mtime from the canonical entry (but of course the size from all).

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+	// update recency list
+	r.repoRecency = r.repoRecency[nDeleted:]
+	return nil
+}
+
+// Update file atime and mtime on the filesystem to time t for all files in the
+// root of the cache that match the repo ID.  This should be called whenever a
+// repository is used.
+// This function does not update the internal cache info.  A call to
+// updateInfo() should be made after touching one or more repositories.
```
teg, line 174: Why not update the internal cache info here too? That way `updateInfo()` would only have to be called at instantiation time.

internal/dnfjson/cache.go
```diff
@@ -1,46 +1,194 @@
...
+func (r *rpmCache) touchRepo(repoID string, t time.Time) error {
+	repoGlob, err := glob.Compile(fmt.Sprintf("%s*", repoID))
 	if err != nil {
 		return err
 	}
-	if curSize > r.maxSize {
-		return os.RemoveAll(r.root)
+
+	// we only touch the top-level directories and files of the cache
```
teg, line 181: As mentioned above, why not only touch one of them? (as long as we know one will always exist).

internal/dnfjson/cache.go
```diff
@@ -136,15 +136,20 @@ func (r *rpmCache) shrink() error {
 	nDeleted := 0
 	for idx := 0; idx < len(r.repoRecency) && r.size >= r.maxSize; idx++ {
 		repoID := r.repoRecency[idx]
-		repo := r.repoElements[repoID]
+		nDeleted++
+		repo, ok := r.repoElements[repoID]
+		if !ok {
+			// cache inconsistency?
```
teg, line 152: Should we not panic here? How could this happen?

internal/dnfjson/dnfjson.go
```diff
@@ -115,6 +119,14 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, erro
 	if err != nil {
 		return nil, err
 	}
+	// touch repos to now
+	now := time.Now().Local()
+	for _, r := range repoMap {
+		// ignore errors
+		_ = s.cache.touchRepo(r.Hash(), now)
+	}
+	s.cache.updateInfo()
```
teg, line 132: As suggested in previous patch, it would be nice if this was not needed. However, if we keep this, it would be nice if we could somehow log about the time spent so we would know in case it was a bottleneck.

internal/dnfjson/cache.go
```diff
@@ -6,11 +6,15 @@ import (
...
 	"path/filepath"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/gobwas/glob"
 )
 
+// global cache locker
+var cacheLocks sync.Map
```
teg, line 16: This sounds like it would make things overly complicated (and indeed would make my suggestions about the update logic not work). Why don't we just require there only to be one cache for a given root at any given time?

teg: (considering this would anyway only work if there were several in the same process, in that case we could just share the object, no?)

internal/dnfjson/dnfjson.go
```diff
@@ -115,6 +115,10 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, erro
 		return nil, err
 	}
 
+	// get non-exclusive read lock
+	s.cache.locker.RLock()
```
teg, line 119: I would squash this into the previous commit as it would make the consistency more obvious.

teg: (The whole commit I mean).

teg (commented):
Left some comments, I would be in favour of merging as is and discuss/address in a follow-up. Nothing critical at all.

internal/dnfjson/cache.go
achilleas-k: Yeah that's a nice idea.  Less bookkeeping.

internal/dnfjson/cache.go
achilleas-k: Yes!

internal/dnfjson/cache.go
achilleas-k: Oh yeah this is a good point.  I should read some manuals (or talk to the dnf devs).

internal/dnfjson/cache.go
achilleas-k: I wanted to limit `stat` calls at first thinking it might slow things down, but that's probably not a big deal.
One thing about this function is that it's called in a loop, once for each repo, for a list of repos used in a transaction, and I didn't want to force each call to start `stat`-ing files and summing sizes.
Probably an unnecessary "optimisation" though.

internal/dnfjson/cache.go
achilleas-k: > How could this happen?

You're right.  This should only happen as a result of a bug in the code that uses these things and they're all private, so a panic would catch a programming error.

internal/dnfjson/cache.go
achilleas-k: Sharing the entire object instead of just the lock does sound like a better idea.  We'd still need to maintain the same kind of map, only it would give you the whole `rpmCache` instead of just the lock.

internal/dnfjson/dnfjson.go
achilleas-k: Bit late for that now! :laughing:

PR #2734, opened 2022-06-09
dnfjson: add repository name and URL to repo-related error messages

achilleas-k:
If dnf-json returns an error that is related to a repository, it uses the ID to identify the repository that caused the error.  Since IDs can't easily be mapped back to a configuration, appending the URL and name (if any) to the error message makes it easier to identify which repository failed.

Keeping the ID in the message is also useful for finding the cache directory of the repository if needed.

For reference, this is what an error message looks like when it has both a name and URL:
```
DNF error occurred: RepoError: There was a problem reading a repository: Failed to download metadata for repo '7a71592ff8a03486179a467c6416bf29e6a7ce46ffee9bc8aa703ff07c3b16ad' [baseos: https://0.0.0.0/baseos/repo]: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

### DIFF
--- a/internal/dnfjson/dnfjson.go
+++ b/internal/dnfjson/dnfjson.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/osbuild/osbuild-composer/internal/rhsm"
@@ -407,8 +408,10 @@ func (err Error) Error() string {
 	return fmt.Sprintf("DNF error occurred: %s: %s", err.Kind, err.Reason)
 }
 
-// parseError parses the response from dnf-json into the Error type.
-func parseError(data []byte) Error {
+// parseError parses the response from dnf-json into the Error type and appends
+// the name and URL of a repository to all detected repository IDs in the
+// message.
+func parseError(data []byte, repos []repoConfig) Error {
 	var e Error
 	if err := json.Unmarshal(data, &e); err != nil {
 		// dumping the error into the Reason can get noisy, but it's good for troubleshooting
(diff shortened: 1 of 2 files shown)


Gundersanne (dismissed):
Perfect, thank you! :)

achilleas-k:

ondrejbudai (dismissed):

thozza (dismissed):
Looks good.

However, DNF uses repo `id` only if the repo `name` is not set. We could actually omit the logic of modifying the error added in this PR and instead ensure that we always set the repo `name` to a reasonable value. Basically setting the repo `name` to the same human-readable value `[name: URL]` here https://github.com/osbuild/osbuild-composer/blob/09f57b6c2f8199a0a8a9365dfb56b3e52e7a60e4/internal/dnfjson/dnfjson.go#L186

achilleas-k:
> Looks good.
> 
> However, DNF uses repo `id` only if the repo `name` is not set. We could actually omit the logic of modifying the error added in this PR and instead ensure that we always set the repo `name` to a reasonable value. Basically setting the repo `name` to the same human-readable value `[name: URL]` here
> 
> https://github.com/osbuild/osbuild-composer/blob/09f57b6c2f8199a0a8a9365dfb56b3e52e7a60e4/internal/dnfjson/dnfjson.go#L186

I tried this.  The error always gets reported with the repo ID.

achilleas-k:
A minimal example:
```
$ cat dnfname.py
import dnf


base = dnf.Base()
base.conf.persistdir = "/tmp/rpmpersist"
base.conf.cachedir = "/tmp/rpmcache"


repo = dnf.repo.Repo("repoid", base.conf)
repo.name = "reponame"
repo.baseurl = "https://0.0.0.0/repo"

base.repos.add(repo)
base.fill_sack(load_system_repo=False)
base.install_specs(["whatever"], reponame="repoid")
base.resolve()

$ python3 ./dnfname.py
Errors during downloading metadata for repository 'repoid':
  - Curl error (7): Couldn't connect to server for https://0.0.0.0/repo/repodata/repomd.xml [Failed to connect to 0.0.0.0 port 443: Connection refused]
  File "/usr/lib/python3.9/site-packages/dnf/repo.py", line 573, in load
    ret = self._repo.load()
  File "/usr/lib64/python3.9/site-packages/libdnf/repo.py", line 331, in load
    return _repo.Repo_load(self)
libdnf._error.Error: Failed to download metadata for repo 'repoid': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried

During handling of the above exception, another exception occurred:

  File "/home/achilleas/./dnfname.py", line 14, in <module>
    base.fill_sack(load_system_repo=False)
  File "/usr/lib/python3.9/site-packages/dnf/base.py", line 399, in fill_sack
    self._add_repo_to_sack(r)
  File "/usr/lib/python3.9/site-packages/dnf/base.py", line 139, in _add_repo_to_sack
    repo.load()
  File "/usr/lib/python3.9/site-packages/dnf/repo.py", line 580, in load
    raise dnf.exceptions.RepoError(str(e))
dnf.exceptions.RepoError: Failed to download metadata for repo 'repoid': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

The URL appears in the first exception here because it's coming from `Curl` failing to connect to the server.  In most of our cases there's no curl error so it doesn't appear.
The `repo.name` (`reponame`) doesn't show up in the error even though I set it.  This even happens if I call `install_specs()` with `reponame` instead of `repoid`.

thozza (dismissed):

achilleas-k:
And now fedora builds are failing because it doesn't use the vendored version of `testify` so it doesn't have the (I guess new) `assert.ErrorContains()` function.

gicmo (approved):
:ok:

PR #2736, opened 2022-06-10
jobqueue: Move jobqueue out of internal packages

Gundersanne:
TODO:
- [x] Move maintenance related queries out of dbjoqueue


teg:
Should we use `/pkg`?

Ref https://github.com/golang-standards/project-layout#pkg

bcl:
> Should we use `/pkg`?

I think `/jobqueue` is just fine, AFAIK that page of suggestions is kind of a catch-all of the way people have done things throughout the history of go and not all of it applies anymore. It seems more natural to me to import from the top level of the module instead of having to add `/pkg/` to it.

teg:
I'm fine with whatever is the current convention

Gundersanne:
> Should we use `/pkg`?
> 
> Ref https://github.com/golang-standards/project-layout#pkg

I thought the same, but after reading issue 10 on that project I wasn't entirely sure anymore. Looking at some of the projects we depend on, some have it, others don't.

ondrejbudai:
> It's ok not to use it if your app project is really small and where an extra level of nesting doesn't add much value (unless you really want to :-)). Think about it when it's getting big enough and your root directory gets pretty busy (especially if you have a lot of non-Go app components).
>
> *Taken from: https://github.com/golang-standards/project-layout#pkg*

This seems like a good argument for using `/pkg` as we have plenty of non-go code in the repository. Would be great if we can restrict go code to just `cmd`, `internal` and `pkg`. WDYT?

Gundersanne:
> > It's ok not to use it if your app project is really small and where an extra level of nesting doesn't add much value (unless you really want to :-)). Think about it when it's getting big enough and your root directory gets pretty busy (especially if you have a lot of non-Go app components).
> > _Taken from: https://github.com/golang-standards/project-layout#pkg_
> 
> This seems like a good argument for using `/pkg` as we have plenty of non-go code in the repository. Would be great if we can restrict go code to just `cmd`, `internal` and `pkg`. WDYT?

I'm happy with either. We already have a lot of stuff in our top level directory so it makes sense.

teg:
I would, as an extra bonus, be interested to understand why we made different choices and if we should unify? We should be having very similar requirements.

teg:
Thanks for the detailed answer, that's very interesting!

cc @diaasami

lzap:
I will file PRs for pool/log abstraction as soon as this gets merged. Looks like ready for review 🥽

ondrejbudai:
> Hm, I think we need the string formatting for on-prem, json messages in the journal wouldn't be ideal imo, plain text is still our default too.

Definitely but if we can use `fmt.Sprintf`, it works for me.

> Also it's not packaged in fedora (but that might not be a huge deal anymore soon as we might switch to vendoring).


I think it is: https://src.fedoraproject.org/rpms/golang-github-logr

lzap:
> json messages in the journal wouldn't be ideal imo

Oh no, please don't say that! :) Syslog is indeed text based, journald is actually meant for structured logging. That's what I was advocating for a very long time in Satellite which now has a fully structured logging (request id, session id, login name, account info...). It is very useful.

But I get the point, your app does have text-based logging approach and that's fine.

> I think it is: https://src.fedoraproject.org/rpms/golang-github-logr

Indeed, however, composer will also need https://github.com/bombsimon/logrusr and that one does not exist, so I think that is the blocker.

Allright, that leaves us with:

```
$ ys golang | grep logging
Poslední kontrola metadat: před 0:44:54, St 22. června 2022, 12:38:39.
golang-github-grpc-ecosystem-middleware-devel.noarch : Golang gRPC Middlewares: interceptor chaining, auth, logging, retries and more
golang-github-hashicorp-logutils-devel.noarch : Utilities for slightly better logging in go (golang)
golang-github-op-logging-devel.noarch : Golang logging library
golang-github-schollz-logger-devel.noarch : Simplistic, opinionated logging for Golang
compat-golang-github-sirupsen-logrus-devel.noarch : Structured, pluggable logging for Go
compat-golang-gopkg-inconshreveable-log15-2-devel.noarch : Structured, composable logging for go
golang-github-apex-log-devel.noarch : Structured logging package for Go
golang-github-apex-log-handlers-kinesis-devel.noarch : Structured logging package for Go
golang-github-coreos-log-devel.noarch : Go logging library with systemd journal support
golang-github-devigned-tab-devel.noarch : OpenTracing and OpenCensus abstraction for tracing and logging
golang-github-facebookincubator-flog-devel.noarch : A Go library for logging
golang-github-gobuffalo-logger-devel.noarch : Common logging interface for the Buffalo ecosystem
golang-github-googlecloudplatform-guest-logging-devel.noarch : Common logger used by GCP guest environment
golang-github-haproxytech-logger-devel.noarch : Go package that provides interface for logging
golang-github-hashicorp-hclog.x86_64 : Common logging package for hashicorp tools
golang-github-hashicorp-hclog-devel.noarch : Common logging package for hashicorp tools
golang-github-inconshreveable-log15-devel.noarch : Structured, composable logging for go
golang-github-ipfs-log-devel.noarch : A logging library used by go-ipfs
golang-github-istio-glog-devel.noarch : A replacement for the glog package that directs logging to zap
golang-github-jpillora-requestlog-devel.noarch : Simple request logging
golang-github-logr-devel.noarch : Simple logging interface for Go
golang-github-projectdiscovery-gologger-devel.noarch : A simple layer for leveled logging in go
golang-github-quay-zlog-devel.noarch : Contextual logging for go
golang-github-revel-log15-devel.noarch : Structured, composable logging for Go
golang-github-sirupsen-logrus-devel.noarch : Structured, pluggable logging for Go
golang-github-spacemonkeygo-spacelog-devel.noarch : Hierarchical, leveled, and structured logging library for go
golang-github-spf13-jwalterweatherman-devel.noarch : Seamless printing to the terminal and logging to a file
golang-uber-zap-devel.noarch : Blazing fast, structured, leveled logging in Go
```

In that case, I will create just a simple interface custom-built for your use case around logrus API. It looks like the code only uses Error/Info/Debug so let's start with that. We can discuss the details in the upcoming PR.

Gundersanne:
> > Also it's not packaged in fedora (but that might not be a huge deal anymore soon as we might switch to vendoring).
> 
> I think it is: https://src.fedoraproject.org/rpms/golang-github-logr

Ugh I missed that :facepalm: 

Anyway we also have an adapter for echo https://github.com/osbuild/osbuild-composer/blob/main/internal/common/echo_logrus.go ; something similar would work.

lzap:
Yeah that's exactly what I have on my mind, it even supports Infof and Infoj (j stands for JSON) for structured logging which I think can be really useful for jobs (job uuid searchable via CloudWatch can be quite useful. I will write a simple adapter similar to this one.

achilleas-k (approved):
Everything's sorted and we're good to go on this then.

PR #2741, opened 2022-06-13
Remove UnmarshalJSON for Stage and StageOptions in osbuild packages

achilleas-k:
OSBuild used to return the stage options as part of the result object for v1 manifests.  We didn't use this information anywhere and in v2 they're not returned at all.  Currently we convert v1 results to the v2 format while parsing the results of jobs from old manifests (old distro definitions), but the StageOptions are ignored and we only care about the StageMetadata.

This PR removes the special parsers (`UnmarshalJSON()`) for Stage types that were needed to take care of all the `StageOptions` implementations.


gicmo:
Corresponding osbuild PR https://github.com/osbuild/osbuild/pull/1044

PR #2749, opened 2022-06-14
CI: new test for checking if a manifest changed in a given PR

achilleas-k:
The new script runs the `gen-manifests` command first on the PR `HEAD` and then on the merge-base with `main` and checks for any differences.  It creates a review comment on the PR on GitHub accordingly.

The message is posted is a simple COMMENT type review to inform the author and reviewers that changes exist.

The script doesn't fail if there's a diff.  CI shouldn't fail if changes are detected since they can be intentional.  The job fails if something goes wrong with the script execution (manifest generation, comment posting, etc).

The script exits immediately if run on main (or an ancestor of main).


ondrejbudai:
Can we run this in github actions? Should be able to run in a Fedora container, doesn't it?

achilleas-k:
The PoC I made ran in GH actions, but we need internal repo access to depsolve RHEL.

ondrejbudai:
ofc :facepalm: Sorry!

achilleas-k:
So this worked but it didn't.  Good progress.

achilleas-k (commented):
test test test

schutzbot (commented):
test test test

schutzbot (commented):
test test test

schutzbot (commented):
✅ This PR does **not** introduce any manifest changes

test/cases/diff-manifests.sh
```diff
@@ -0,0 +1,79 @@
...
+function redprint {
+    echo -e "\033[1;31m[$(date -Isecond)] ${1}\033[0m"
+}
+
+greenprint "Fetching origin/main"
+git fetch origin main
+
+greenprint "Getting revision IDs for HEAD and merge-base"
+head=$(git rev-parse HEAD)
+mergebase=$(git merge-base HEAD origin/main)
```
ondrejbudai, line 18: This will be weird when we make PR to rhel-* branches. Not 100% sure if it's worth solving and how to solve it.

achilleas-k: Oh yeah, didn't think of that.  Can we pass the target branch name from GitHub to GitLab CI and use that instead?

achilleas-k: Alternatively, we can probably query GH for the PR info and get the target branch name from there.

achilleas-k: Updated.  Gets the base branch from GH.  Also, it skips the run if the `$CI_COMMIT_BRANCH` doesn't match `PR-*`.

gicmo (approved):
Love it! :heart:

PR #2751, opened 2022-06-15
dnfjson: Load subscriptions when creating a new solver

Gundersanne:
The BaseSolver is an object which gets constructed when the worker
starts, and the subscriptions attached to it expire after about 3
days. By refreshing the subscriptions each time a new Solver is created,
valid subscriptions are used.


thozza (approved):
LGTM 👍

PR #2752, opened 2022-06-15
Support the insecure option in curl sources

achilleas-k:
_Counterpart to https://github.com/osbuild/osbuild/pull/1047_

This PR fixes a bug where a repository could be specified with the `ignore_ssl` option, but the rule only applied when depsolving.
If a repository was configured with `ignore_ssl` and had an invalid or private certificate, the depsolve would succeed, but the download of the packages from the repository would fail because osbuild's `org.osbuild.curl` source did not support setting the `--insecure` flag.

This PR:
- Adds the `Insecure` option to the `org.osbuild.curl` source.
- Adds the `ignore_ssl` (`IgnoreSSL`) field to the `repository` object that is used to read the global repo configs.  This option was already supported for repositories added through Weldr sources and the Cloud API request but could not be specified through the configs in  `/usr/share/` and `/etc/`.
- Adds the `IgnoreSSL` field to the `PackageSpec`, which specifies that any SSL checks should be ignored when downloading a given package from a secure server.  This field inherits the value from the appropriate repository configuration.
- Deduplicates the code used in each distro to generates the `sources` section of the manifest.  This is only done for distros that use v2 manifests (so RHEL 8 and the original part of RHEL 8.4 are not changed).


achilleas-k:
I didn't update the osbulid dependency version in the Schutzfile since this shouldn't affect any tests.  Do we want to add a test for this?

teg (approved):

achilleas-k:
Cert generation needed a fix.  Was prompting for password.

thozza (approved):
Everything looks good 👍

PR #2754, opened 2022-06-15

ygalblum:
Remove all Go BuildRequires dependecies
Add '-k' to goprep to preserve the vendor directory
Generate 'Provides: bundled' lines by processing vendor/modules.txt in lua

Signed-off-by: Ygal Blum <ygal.blum@gmail.com>


ygalblum:
The PackIt process parses the spec file at least 3 times. I was able to overcome the first two, but I could not make the third case. The RPM build command is executed without setting any parameters that will allow the lua script to find the file to process. Instead see #2762

PR #2758, opened 2022-06-16
Consolidate Koji target options values meaning

thozza:
When the Koji target support was added to the osbuild job, based on the osbuild-koji job, the meaning of target option values got messed up.

This PR is ensuring, that the Koji target interprets values set in target options in the same way as any other target.

In addition, ensure that the worker server returns "Bad Request" if the worker tries to upload an artifact to it, but the server is configured to not accept any artifacts from workers.

Please not that this change is backward incompatible. This was done intentionally after a discussion with @ondrejbudai, because the Koji composes via Cloud API are currently not used anywhere in production.

This pull request includes:

### DIFF
--- a/internal/target/koji_target.go
+++ b/internal/target/koji_target.go
@@ -1,6 +1,7 @@
 package target
 
 type KojiTargetOptions struct {
+	// Filename of the image as produced by osbuild for a given export
 	Filename        string `json:"filename"`
 	UploadDirectory string `json:"upload_directory"`
 	Server          string `json:"server"`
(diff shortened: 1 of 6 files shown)


internal/target/koji_target.go
```diff
@@ -1,7 +1,7 @@
 package target
 
 type KojiTargetOptions struct {
-	Filename        string `json:"filename"`
+	Filename        string `json:"filename"` // Filename of the image as produced by osbuild for a given export
```
achilleas-k, line 4: Nitpick but moving the comment to the line above the field name would make it the docstring by convention and I think looks nicer.

Not an issue though :)
I think it gets picked up as the docstring like this as well.

achilleas-k (approved):
Good stuff.  Extremely minor nitpick.  Feel free to ignore :)

internal/target/koji_target.go
thozza: I'll change it, because in addition to that I noticed that the change in `v2_koji_test.go` from the second commit should have been in the first one.

thozza: I fixed the nitpick and moved the test change to the first commit.... The PR should be perfect now 😇

achilleas-k (approved):

PR #2762, opened 2022-06-19

ygalblum:
## RPM Spec
Remove all Go dependecies
Add Start and End marker comments for bundling information
Add '-k' to goprep to preserve the vendor directory
    
## tools
Add script to update the RPM spec file to generate the indication lines based on vendor/modules.txt
    
## Packit
Run the new script as a post-upstream-clone hook
    
## Makefile
Run the new script on the generated spec file before generating the RPM

This pull request includes:

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,6 +16,7 @@ upstream_tag_template: v{version}
 srpm_build_deps: []
 actions:
   get-current-version: bash -c "git describe --tags --abbrev=0 | sed 's|v||'"
+  post-upstream-clone: "./tools/rpm_spec_add_provides_bundle.sh"
 
 jobs:
 - job: koji_build
--- a/Makefile
+++ b/Makefile
@@ -214,6 +214,7 @@ RPM_TARBALL=rpmbuild/SOURCES/osbuild-composer-$(COMMIT).tar.gz
 $(RPM_SPECFILE):
 	mkdir -p $(CURDIR)/rpmbuild/SPECS
 	git show HEAD:osbuild-composer.spec > $(RPM_SPECFILE)
+	./tools/rpm_spec_add_provides_bundle.sh $(RPM_SPECFILE)
 
 $(RPM_TARBALL):
 	mkdir -p $(CURDIR)/rpmbuild/SOURCES
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -42,34 +42,9 @@ BuildRequires:  make
 %if 0%{?fedora}
 BuildRequires:  systemd-rpm-macros
 BuildRequires:  git
-BuildRequires:  golang(github.com/aws/aws-sdk-go)
-BuildRequires:  golang(github.com/Azure/azure-sdk-for-go)
-BuildRequires:  golang(github.com/Azure/azure-storage-blob-go/azblob)
-BuildRequires:  golang(github.com/BurntSushi/toml)
-BuildRequires:  golang(github.com/coreos/go-semver/semver)
-BuildRequires:  golang(github.com/coreos/go-systemd/activation)
-BuildRequires:  golang(github.com/deepmap/oapi-codegen/pkg/codegen)
-BuildRequires:  golang(github.com/go-chi/chi)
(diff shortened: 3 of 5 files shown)


ondrejbudai (commented):
We agreed this is good to go. I would just want one change: Can you run the script also in `mockbuild.sh` which is used in our CI to build the RPMs? For the sake of consistency. It should run somewhere before this line: https://github.com/osbuild/osbuild-composer/blob/main/schutzbot/mockbuild.sh#L108

ygalblum:
> We agreed this is good to go. I would just want one change: Can you run the script also in `mockbuild.sh` which is used in our CI to build the RPMs? For the sake of consistency. It should run somewhere before this line: https://github.com/osbuild/osbuild-composer/blob/main/schutzbot/mockbuild.sh#L108

@ondrejbudai Done

ondrejbudai (approved):

gotmax23:
As a member of the Fedora Go SIG, I'm a bit sad to see this. Can you elaborate on what led you to stop using the RPM packaged dependencies?

gotmax23:
Hi @ondrejbudai,

I appreciate your detailed response. I agree with some of your points, but not all of them. I guess I'll respond to those for informational purposes. It seems you all have already made up your mind, so I'm not trying to make you change it :).

> We recently implemented container upload support using podman go libraries not packaged in Fedora (podman bundles as well). I walked through the dep tree and discovered that we would need to package 17 new Fedora SRPMs.

Yeah, I can understand this. Unfortunately, go projects tend to have huge swathes of dependencies. When there's more than 10 or even just 5 new dependencies that need to be packaged, it often justifies bundling, despite the disadvantages of doing so.

> Traditionally, it was a great idea to split libraries into separate libraries so packagers can fix a bug/CVE in one package and the fix would propagate to all packages that depend on it

> Go (Rust) is different, though, it compiles everything statically. So even though we fix a bug/CVE in a library, we need to go through all packages that depend on it, bump the release, rebuild and push a bodhi update.

I am already doing this. If you follow the Fedora devel mailing list, you've probably seen emails from me about rebuilds for various golang, prometheus, golang-x-text, go-restful, and other CVEs. (I'm not sure if all of these rebuilds were explicitly announced. Some of them were batched together.) It is not too difficult to run my rebuild script to rebuild every affected package in a side tag and then create the Bodhi update. Using unbundled dependencies often means that vulnerabilities are fixed *faster* and *more* easily than if we had to wait for every upstream to update their go.mod / vendor directory and create a release. Some upstreams are on top of this, but many upstreams aren't very active or otherwise don't quickly fix these issues.

>  Currently, Go packages in Fedora are built against different libraries than defined in their upstream go.sum - therefore all upstream testing is basically invalid for Fedora because we build it differently.

I wouldn't say upstreams' CI pipelines are completely invalid, and we run the tests downstream if it's at all practical, anyways.

> So with explicit downstream patches

The go packages that I've seen to have very few or zero downstream patches. I'm not sure if you're saying that using slightly different dependency versions amounts to downstream patching. I would say it doesn't.

> CS bundles all deps which meant that we had to be careful about our deps - it was certainly possible that CS would get different dep versions than Fedora which is yet another potential source of bugs.

I'm a bit conflicted about how much I agree or disagree with this point, but I will say that, if you really need dependency versions that are newer than those packaged in Fedora, you can always submit PRs and/or request to join the Go SIG which has `commit` access to most of the go library packages. We are working with FESCO (Fedora Engineering Steering Committee) to require all packagers to add the Go SIG to their go packages. See https://pagure.io/fesco/issue/2837 more information about the justification for this. That reminds me, you still haven't added `go-sig` to osbuild-composer. Please go to https://src.fedoraproject.org/rpms/osbuild-composer/addgroup and give `go-sig` the `commit` ACL.

> the benefits of having a separate RPM per a go library are basically minimal in our eyes

I won't deny the disadvantages, but there are some other advantages besides the unified CVE fixes:

1. Packages recieve bug fixes and security fixes from newer library versions, when they may not have been updated upstream.
2. Package *and libraries* are tested against new golang versions, which we track closely.
3. The dependencies that packages and libraries build/test against are always the same, even if the `go.mod`s of the two projects pin to different versions.

PR #2768, opened 2022-06-22
Remove vhd image type from RHEL 7

achilleas-k:
1. Broken: specifies packages that aren't available in the repositories (langpacks-en and dhcp-client).
2. Not needed.


ondrejbudai (approved):

ondrejbudai:
Manifest-diff tests are failing in this PR because there's a broken image definition on `main`. This PR should fix it so let's merge.

PR #2262, opened 2022-02-04
RHEL-90: Remove deprecated `crashkernel=auto` option

thozza:
Setting of the `crashkernel` option to the appropriate value is now done
by the `kexec-tools` package when installed and when any new kernel is
installed.

Regenerate relevant image test cases.

Fix #1819
Fix [rhbz#2006692](https://bugzilla.redhat.com/show_bug.cgi?id=2006692)

Signed-off-by: Tomas Hozza <thozza@redhat.com>

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #2771, opened 2022-06-23
[WIP] Enable Image Builder to build GCP-compatible RHEL 9.0 images

FKolwa:
**This pull request includes:**
internal/distro: 
- Added GCE RHUI image configuration and image type for RHEL90.
- Added GCE RHUI PackageSet

tools: Updated google-compute-engine and google-cloud-sdk repository versions.

**What needs to be done:**
- Regenerate test cases
- Build proof of work GCE  image for Stratosphere

- [] adequate testing for the new functionality or fixed issue
- [] adequate documentation informing people about the change such as
- [] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

Disclaimer: Sorry I've been searching for the Contributor Guidelines but I've found them after writing my first commit message.


internal/distro/rhel90/distro.go
```diff
@@ -1648,7 +1648,55 @@ func newDistro(distroName string) distro.Distro {
 			osPkgsKey: {osPkgsKey, blueprintPkgsKey},
 		},
 		defaultImageConfig:  defaultGceImageConfig,
-		kernelOptions:       "net.ifnames=0 biosdevname=0 scsi_mod.use_blk_mq=Y console=ttyS0,38400n8d",
+		kernelOptions:       "net.ifnames=0 biosdevname=0 scsi_mod.use_blk_mq=Y crashkernel=auto console=ttyS0,38400n8d",
```
thozza, line 1651: As we discussed, the `crashkernel=auto` option should not be used on RHEL-9 any more. More info in #2262

internal/distro/rhel90/distro.go
```diff
@@ -1648,7 +1648,55 @@ func newDistro(distroName string) distro.Distro {
...
+		mimeType: "application/gzip",
+		packageSets: map[string]packageSetFunc{
+			buildPkgsKey: distroBuildPackageSet,
+			osPkgsKey:    gceRhuiPackageSet,
+		},
+		packageSetChains: map[string][]string{
+			osPkgsKey: {osPkgsKey, blueprintPkgsKey},
+		},
+		defaultImageConfig:  defaultGceRhuiImageConfig,
+		kernelOptions:       "net.ifnames=0 biosdevname=0 scsi_mod.use_blk_mq=Y crashkernel=auto console=ttyS0,38400n8d",
```
thozza, line 1699: As we discussed, the `crashkernel=auto` option should not be used on RHEL-9 any more. More info in #2262

thozza (commented):
Added a few comments, but otherwise this looks great 👍

tools/test-case-generators/repos.json
```diff
@@ -999,8 +999,8 @@
...
+        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/el9/el9-x86_64-google-compute-engine-20220617",
+        "gpgkey": "",
```
thozza, line 1003: nitpick: you don't have to specify `gpgkey` at all if it is empty.

tools/test-case-generators/repos.json
```diff
@@ -1009,8 +1009,8 @@
...
+        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/el9/el9-x86_64-google-cloud-sdk-20220617",
+        "gpgkey": "",
```
thozza, line 1013: ditto - not specifying `gpgkey` at all.

test/data/manifests/rhel_91-x86_64-gce-boot.json
```diff
@@ -18,17 +18,19 @@
...
+        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/el9/el9-x86_64-google-compute-engine-20220617",
+        "gpgkey": "",
```
thozza, line 22: This test case manifest should be regenerated, because the GPG key from the `rhel-91` google repos definition was not removed in the end, however this manifest does not contain it. The reason is most probably that you generated it after you removed the GPG key from repo definitions, but then didn't do it again once you reverted the change.

thozza (commented):
I reviewed the change and in general it looks good. 👍 

However I do think that you should clean up changes to the `tools/test-case-generators/repos.json` related to the Google repos. By that I mean squashing some changed together into a single commit and regenerating all image test cases with the final state of `tools/test-case-generators/repos.json`. Currently:

- 444f6a55bd95c981960328ed88beddf27e55470d updates version of `rhel-90` google repositories, but does not regenerate relevant image test cases.
- dc2cb0221d9758c3d2a7fa995315ba645ad899b5 removed the GPG key from the `rhel-91` google repositories definition.
- e677fedb019770c130b3b0f54a5198d39625f11e adds back the GPG key to the `rhel-91` google repositories definition and regenerates the image test cases (but not with the final state of `tools/test-case-generators/repos.json`.

In addition, I have a few comments related to commit message style and split of changes into commits. This is to ensure a common style in the project.

- You should use imperative style in commit message subject. So instead of e.g. `Adding repos ...` you should use `Add repos ...`.
- The commit message subject should be ideally at maximum 50 characters long, but definitely not longer than 72 characters. Many of your commits have longer subject. This should be fixed, so that they don't overflow to the next line.
- In general, the project should be able to compile after each commit from your PR, because we don't squash them when we merge. Not doing so will make it harder to potentially bisect the code in the future if needed.
  - E.g. in the commit f3b752c7fe2a0c67d3ea3e97d4e78e917f7dbd88 you add the `gce-rhui` image type to RHEL-91 x86_64 architecture, but you are adding the image type to the unit test that checks the available image types in an architecture 2 commits later in e80dca614f1d873543e4709507a281ee5f2fd2c0. This means that the unit test would be failing for two commits.
  - Another issue is that in the commit f3b752c7fe2a0c67d3ea3e97d4e78e917f7dbd88 you are adding the `crashkernel=auto` option to the kernel command line and later you are removing it in dc2cb0221d9758c3d2a7fa995315ba645ad899b5. That change should not be present at all in the PR and you should have squashed dc2cb0221d9758c3d2a7fa995315ba645ad899b5 into f3b752c7fe2a0c67d3ea3e97d4e78e917f7dbd88.

achilleas-k (approved):
Thanks for this.  Looks all good to me and all cleaned up as well.

achilleas-k (approved):

PR #2786, opened 2022-06-29
tag azure resources with gitlab-ci-test

ghost:
cloud cleaner removes old enough resources with 'gitlab-ci-test' tag.

This pull request includes:

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -1547,16 +1547,17 @@ function verifyInAzure() {
   $AZURE_CMD login --service-principal --username "${V2_AZURE_CLIENT_ID}" --password "${V2_AZURE_CLIENT_SECRET}" --tenant "${AZURE_TENANT_ID}"
   set -x
 
-  # verify that the image exists
+  # verify that the image exists and tag it
   $AZURE_CMD image show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_IMAGE_NAME}"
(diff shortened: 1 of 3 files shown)


jabia99:
A new cleaning method for azure has been added ([commit](https://github.com/osbuild/cloud-cleaner/commit/843160f45e6189b39297dac4d947b9c621085063)), we need to tag every resource with `gitlab-ci-test`

jrusz (commented):
What is the `deployment-template.json` used for? I don't really think our tests use it? 

Also did you try to let some tests fail intentionally and then clean up with the cloud-cleaner to see if it works?

It looks good otherwise would just be nice to see it in action to be sure :)

jabia99:
> What is the `deployment-template.json` used for?

it is used as a template, declared in this [go file](https://github.com/osbuild/osbuild-composer/blob/fc1d754999cc58d4e3e4d947a24a4170274b5a0f/cmd/osbuild-image-tests/constants/constants.go#L42) (under `osbuild-image-tests`). It was an exact copy of the `main.tf` file, so I thought it made sense to add the tag there too

jabia99:
> Also did you try to let some tests fail intentionally and then clean up with the cloud-cleaner to see if it works?

Yes, I tried cancelling one job mid run and it works great. The only issue is that it takes some time for the resource to be available through the API (30-60m), but it's less than `HOURS_BACK`, so i don't think it'll be a problem

jrusz (approved):
Ok thanks, it's all good then I guess, approved :)

PR #2787, opened 2022-06-29
Merge RHEL 8.5 distro definition into the distro/rhel86 package

achilleas-k:
RHEL 8.5 is now defined in the RHEL 8.6 `distro/rhel86` package.

All manifest changes are explained in the relevant commit message.

This is the first step towards merging all minor RHEL versions into a single package.  RHEL 8.5 is the simplest since it has the smallest number of differences compared to RHEL 8.6.  The ultimate goal is to have a single `distro/rhel8` (or `distro/centos8`) package that defines all RHEL 8 versions.  Older RHEL versions (8.3 and 8.4) have v1 manifests which will need to be examined for changes more carefully.

The first commit updates the RHEL 8.5 manifests to pull in new package versions from the RHUI repository, so that the final manifest regeneration shows clearly what changed from the merge.


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 5b28c0e06575fb91bdb6056ba61d1651ffd5d92c with the main merge-base 3fcb737243b742fa9143cd9df068044eeff54eb1).  Please review the changes.

teg (commented):
Is it ok to drop simplified installer from 8.5? I assume so, but a comment why would be great.

teg (approved):
Amazing! Thank you so much!

gicmo:
> Is it ok to drop simplified installer from 8.5? I assume so, but a comment why would be great.

The `edge-simplified-installer` is not even exposed in the CloudAPI, so this only affects on-prem:

https://github.com/osbuild/osbuild-composer/blob/3fcb737243b742fa9143cd9df068044eeff54eb1/internal/cloudapi/v2/openapi.v2.yml#L647-L660

gicmo (approved):
💚  MERGE IT!

internal/distro/rhel86/distro.go
```diff
@@ -1757,24 +1808,66 @@ func newDistro(distroName string) distro.Distro {
...
 		// add ec2 image types to RHEL distro only
-		x86_64.addImageTypes(ec2ImgTypeX86_64, ec2HaImgTypeX86_64, ec2SapImgTypeX86_64)
+		x86_64.addImageTypes(ec2ImgTypeX86_64, ec2HaImgTypeX86_64)
+		if !versionLessThan(rd.osVersion, "8.6") {
+			x86_64.addImageTypes(ec2SapImgTypeX86_64)
+
+			// edge simplified installer is only available on 8.6 and later on RHEL
+			x86_64.addImageTypes(edgeSimplifiedInstallerImgType, edgeRawImgType)
+			aarch64.addImageTypes(edgeSimplifiedInstallerImgType, edgeRawImgType)
+		}
```
thozza, line 1859: What is the risk of not making this distinction for older RHEL releases? It is not like we would be introducing these image types to the RHEL-8.5 which is already released. Also one can not build 8.5 in the service, because AFAIK it supports only the latest RHEL.

So if these images can be built with RHEL-8.5 repositories, I would be for supporting them there just to make things simpler.

achilleas-k: The images can't be built on 8.5 anymore.  The FDO packages we added in 8.6 aren't available in 8.5.  It's simple enough to conditionally add packages, but there's LUKS and FDO things all over the pipelines that we'd have to conditionally toggle.  Supporting them in 8.5 would add a lot of mess to the pipelines for no benefit since, afaict, no one would want to build these for 8.5.

PR #2341, opened 2022-02-22
test/api: split into smaller files

ondrejbudai:
Each cloud now has its own file that's sourced on-demand by the main api.sh
script. The main goal of this commit is to reduce the amount of clutter in
api.sh. I, personally, find 1300 lines of bash overwhelming and I think that
this is a reasonable beginning to start cleaning things up.

Signed-off-by: Ondřej Budai <ondrej@budai.cz>

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #2789, opened 2022-06-30
Api test split into multiple files

Gundersanne:
Original PR: https://github.com/osbuild/osbuild-composer/pull/2341

---

The api test has kinda grown to be 1 huge shell script which isn't readable/maintainable. This PR rebases #2341, and fills in the remaining gaps. With the multiple images per compose for the cloudapi each cloud will get tests more specific to it's own image/upload type, so the 1 single test workflow for all targets is less suitable.


teg:
Looks good! With a PR description it would be perfect!

achilleas-k (approved):
Great stuff.  Thank you!

PR #2791, opened 2022-06-30
distro: don't enable fdo-client service on RHEL 8.5

achilleas-k:
RHEL 8.5 doesn't have FDO and it's not installed, but we were enabling the service for it in edge.


gicmo (approved):
🩹

achilleas-k:
Accidentally disabled on CS8 as well.  Updated.

gicmo (approved):

PR #2797, opened 2022-07-01
containers/osbuild-composer: Sleep before shutdown to mitigate connections being reset/terminated on shutdown

Gundersanne:


ondrejbudai:
Shall we immediately enable it?

teg:
This sounds very odd / surprising. If open shift is the one doing the routing and the one sending the SIGTERM, why isn't this synced on their side? I obviously know nothing, but would be interested in learning more about the problem. Is this a known issue? Were there any discussions/reports about it?

ondrejbudai:
> This sounds very odd / surprising. If open shift is the one doing the routing and the one sending the SIGTERM, why isn't this synced on their side? I obviously know nothing, but would be interested in learning more about the problem. Is this a known issue? Were there any discussions/reports about it?

Good questions! Back story: I was running 22 parallel `osbuildImage` jobs running against staging AOC composer. FTR, `koji-osbuild` queries the job status every 2 seconds. If it fails to query composer, if fails the job. Everything went smoothly until stage got redeployed. During the next 15 seconds, all of these jobs failed because of `GenericError: Failed to get the compose status: upstream connect error or disconnect/reset before headers. reset reason: connection failure`.

That basically points at the fact that our redeployments aren't zero-downtime. I googled and found the following articles:

- https://deepsource.io/blog/zero-downtime-deployment/
- https://engineering.rakuten.today/post/graceful-k8s-delpoyments/

Quoting the first article:

> Once a new pod is alive and ready, Kubernetes will take an old one out of service and thus update the pod’s status to Terminating, remove it from the endpoints object, and send a SIGTERM. The SIGTERM causes the container to shutdown, in a (hopefully) graceful manner, and to not accept any new connections. After the pod has been evicted from the endpoints object, the load balancers will route the traffic to the remaining (new) ones. This is what caused our availability gap in our deployment; the pod was being deactivated by the termination signal, before, or rather while the load balancer notices the change and can update its configuration. This re-configuration happens asynchronously, thus makes no guarantees of correct ordering, and can and will result in few unlucky requests being routed to the terminating pod.

Both articles suggest adding a bit of sleep before stopping accepting new connections, so k8s/oc has time to realize that this pod is going down and remove it from its routing tables. I even found a project that had to do the same thing: https://github.com/open-policy-agent/opa/pull/2834

Happy to find a better solution, but my (limited) research showed that this is what people are apparently doing. :shrug:

BTW, we should also make `koji-osbuild`'s behavior better. It should probably tolerate a bit of a downtime when querying the status? Or maybe implement yet another retrying mechanism? Curious to hear your thoughts.

teg:
Thanks for the details! Sounds like the best call, though might be worth investigating further to see if there really is no other way, or if this should be in the AppSRE guidelines?

How about we don't consider a missing status as failure in koji? Just ignore it?

teg (approved):

ochosi:
@Gundersanne Would you mind making the PR title more expressive? That would help make the release notes more useful - thanks <3

PR #2799, opened 2022-07-04
manifest/os: minor refactoring to make more arguments optional

teg:
All arguments that has reasonable defaults and where querying the package set chain as well as serialisation works without
the defaults being overridden, should be made public fields rather than arguments to the constructor.

It is the caller's responsibility not to change the parameters between querying the package set chains and serialising the
manifest, but otherwise they are free to reuse the pipeline objects and modify them freely.

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -9,6 +9,15 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 )
 
+type Arch uint64
+
+const (
+	ARCH_X86_64 Arch = iota
+	ARCH_AARCH64
+	ARCH_S390X
+	ARCH_PPC64LE
+)
+
 type osTreeCommit struct {
 	checksum string
 	url      string
--- a/cmd/osbuild-playground/playground.go
+++ b/cmd/osbuild-playground/playground.go
@@ -31,7 +31,7 @@ func MyManifest(m *manifest.Manifest, options *MyOptions, repos []rpmmd.RepoConf
 	build := manifest.NewBuildPipeline(m, runner, repos)
 
 	// create a non-bootable OS tree containing the `core` comps group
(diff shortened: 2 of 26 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4b43c1b4c97f8edee16e88dbea5fe4a1556478ff with the main merge-base ad847a6fcab4d470e15fc17aeb5ad3aa296ff6dd).  Please review the changes.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 83b53228a08783c24f1455b1cd919355a173ceb7 with the main merge-base ad847a6fcab4d470e15fc17aeb5ad3aa296ff6dd).  Please review the changes.

achilleas-k (approved):
I like this PR :+1:

internal/manifest/manifest.go
```diff
@@ -9,6 +9,15 @@ import (
...
 )
 
+type Arch uint64
+
+const (
+	ARCH_X86_64 Arch = iota
+	ARCH_AARCH64
+	ARCH_S390X
+	ARCH_PPC64LE
+)
```
thozza, line 19: We already have something like this defined in https://github.com/osbuild/osbuild-composer/blob/2c2d4812d4b381275e8c52d86ec81b7949b42cfe/internal/distro/distro.go#L15-L22

teg: so far i have the distro package depend on all the new things, but not the other way around. let's see if we should change that once it all settles (or if we should drop things from the distro package)

thozza: Sure, one way or the other sounds good. I just wanted to point out that we already have something similar defined, so that we don't end up with both definitions in the end.

PR #2800, opened 2022-07-05
test/cases: add new mountpoints in filesystem.sh

jabia99:
There are a list of mountpoints in the [documentation](https://www.osbuild.org/guides/blueprint-reference/blueprint-reference.html#filesystem-support), but only
some of this mountpoints are tested in the test filesystem.sh. Add the
ones missing.

### DIFF
--- a/test/cases/filesystem.sh
+++ b/test/cases/filesystem.sh
@@ -169,6 +169,26 @@ size = 131072000
 [[customizations.filesystem]]
 mountpoint = "/var/tmp"
 size = 131072000
+
+[[customizations.filesystem]]
+mountpoint = "/home"
+size = 131072000
+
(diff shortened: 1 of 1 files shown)


test/cases/filesystem.sh
```diff
@@ -169,6 +169,26 @@ size = 131072000
...
+size = 131072000
+
+[[customizations.filesystem]]
+mountpoint = "/app"
+size = 131072000
+
+[[customizations.filesystem]]
+mountpoint = "/data"
+size = 131072000
 EOF
```
atodorov, line 192: You have to explicitly assert on these mount points below.

atodorov (changes requested):

atodorov (approved):

PR #2804, opened 2022-07-05
distro/fedora: no longer use `MakePackageSetChains()`

teg:
This completes the move to have the package set chains computed in the manifest, rather than from the outside.

There is a lot of potential work left to make this nicer, but at least we are now rid of the package set chain logic.

See individual commits for more details.


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d13ad573ba9e43a4160645ea4ba4d513d9327a1e with the main merge-base c2c43594b109174263eb9df3fb12052a6e48d90a).  Please review the changes.

internal/distro/fedora/distro.go
```diff
@@ -544,44 +544,40 @@ func (t *imageType) Size(size uint64) uint64 {
...
+
+	// amend with repository information
+	for name := range packageSets {
+		packageSet := packageSets[name]
+		for _, repo := range repos {
+			if repo.AppliesToPackageSet(name) {
+				packageSet.Repositories = append(packageSet.Repositories, repo)
+			}
+		}
+		packageSets[name] = packageSet
```
achilleas-k, line 569: The triple loop here (which includes the inner `AppliestoPackageSet()` loop) seems a bit hard to follow and maybe unnecessarily loopy?

I think this would be equivalent and doesn't need the repo helper:
```go
	for _, repo := range repos {
		for _, psName := range repo.PackageSets {
			ps := packageSets[psName]
			ps.Repositories = append(ps.Repositories, repo)
			packageSets[psName] = ps
		}
	}
```

achilleas-k (commented):
Looks good.  One small comment for a simplification.  I'm pretty sure it's equivalent.

internal/distro/fedora/distro.go
teg: D'oh, you are so right. I'll fix this (it also fixes a bug in a corner case).

teg: Fixed.

teg:
@achilleas-k fixed your comment, and added a bunch more patches ;)

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD dfa31a0d65ac62b3860ff23bff036d1f1b085e53 with the main merge-base 4e2cea61bbe018cf2dcbea2061384c9370e4a4c9).  Please review the changes.

gicmo:
I *love* the direction this is going and how we separate the concerns here. Just one thing: I wondering if we get away with this "pull" model, i.e. properties defined at the different workloads that need to be integrated into the pipeline options. If you look at SAP and Azure, they do set quite a few options (udev rules, tempfiles, kernel command line, password rules).
I wonder if a "operator" model would be better, where the workloads are given the pipelines and then modify them.

teg:
> I *love* the direction this is going and how we separate the concerns here. Just one thing: I wondering if we get away with this "pull" model, i.e. properties defined at the different workloads that need to be integrated into the pipeline options. If you look at SAP and Azure, they do set quite a few options (udev rules, tempfiles, kernel command line, password rules).
> I wonder if a "operator" model would be better, where the workloads are given the pipelines and then modify them.

We discussed this by phone and agreed that if we get away with the pull model it would be nice, and if we don't we can refactor after.

PR #2809, opened 2022-07-07
Create multiple aws images from a single compose

Gundersanne:
TODO:
- [x] transfer sharing attributes to the copied image
- [x] transfer tags needed? - decided to just tag similarly to the registration in the osbuild job, but with the new Name => no transfer of original tags
- [x] don't copy an image to a new region if it was already copied there - with the caveat that if 2 jobs are queued at the same time with the same target region the copy will happen twice :/ Could also look at job inputs and chain the share job on the existing copy job that way.

questions:
- [ ] allow to add these on top of a running osbuild job? => use dynargs rather than static ones?
- [ ] split across 2 jobs rn, should be just the 1? Only downside of using 1 job is that you lose the ability to optimize this further, if a copy job has already been queued (but not completed), you could attach multiple share jobs to it.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-awsec2.go
+++ b/cmd/osbuild-worker/jobimpl-awsec2.go
@@ -1,0 +1,144 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/sirupsen/logrus"
+
+	"github.com/osbuild/osbuild-composer/internal/cloud/awscloud"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+func getAWS(awsCreds, region string) (*awscloud.AWS, error) {
+	if awsCreds != "" {
+		return awscloud.NewFromFile(awsCreds, region)
+	}
+	return awscloud.NewDefault(region)
+}
+
+type AWSEC2CopyJobImpl struct {
+	AWSCreds string
+}
+
+func (impl *AWSEC2CopyJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id())
+	result := worker.AWSEC2CopyJobResult{}
+
+	defer func() {
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	var args worker.AWSEC2CopyJob
+	err := job.Args(&args)
+	if err != nil {
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorParsingJobArgs, fmt.Sprintf("Error parsing arguments: %v", err), nil)
+		return err
+	}
+
+	aws, err := getAWS(impl.AWSCreds, args.TargetRegion)
+	if err != nil {
+		logWithId.Errorf("Error creating aws client: %v", err)
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidConfig, "Invalid worker config", nil)
+		return err
+	}
+
+	ami, err := aws.CopyImage(args.TargetName, args.Ami, args.SourceRegion)
+	if err != nil {
+		logWithId.Errorf("Error copying ami: %v", err)
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Error copying ami %s", args.Ami), nil)
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "InvalidRegion":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Invalid source region '%s'", args.SourceRegion), nil)
+			case "InvalidAMIID.Malformed":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Malformed source ami id '%s'", args.Ami), nil)
+			case "InvalidAMIID.NotFound":
+				fallthrough // CopyImage returns InvalidRequest instead of InvalidAMIID.NotFound
+			case "InvalidRequest":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Source ami '%s' not found", args.Ami), nil)
+			}
+		}
+		return err
+	}
+
+	result.Ami = ami
+	result.Region = args.TargetRegion
+	return nil
+}
+
+type AWSEC2ShareJobImpl struct {
+	AWSCreds string
+}
+
+func (impl *AWSEC2ShareJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id())
+	result := worker.AWSEC2ShareJobResult{}
+
+	defer func() {
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	var args worker.AWSEC2ShareJob
+	err := job.Args(&args)
+	if err != nil {
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorParsingJobArgs, fmt.Sprintf("Error parsing arguments: %v", err), nil)
+		return err
+	}
+
+	if args.Ami == "" || args.Region == "" {
+		if job.NDynamicArgs() != 1 {
+			logWithId.Error("No arguments given and dynamic args empty")
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorNoDynamicArgs, "An ec2 share job should have args or depend on an ec2 copy job", nil)
+			return nil
+		}
+		var cjResult worker.AWSEC2CopyJobResult
+		err = job.DynamicArgs(0, &cjResult)
+		if err != nil {
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorParsingDynamicArgs, "Error parsing dynamic args as ec2 copy job", nil)
+			return err
+		}
+		if cjResult.JobError != nil {
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorJobDependency, "AWSEC2CopyJob dependency failed", nil)
+			return nil
+		}
+
+		args.Ami = cjResult.Ami
+		args.Region = cjResult.Region
+	}
+
+	aws, err := getAWS(impl.AWSCreds, args.Region)
+	if err != nil {
+		logWithId.Errorf("Error creating aws client: %v", err)
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidConfig, "Invalid worker config", nil)
+		return err
+	}
+
+	err = aws.ShareImage(args.Ami, args.ShareWithAccounts)
+	if err != nil {
+		logWithId.Errorf("Error sharing image: %v", err)
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Error sharing image with target %v", args.ShareWithAccounts), nil)
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "InvalidAMIID.Malformed":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Malformed ami id '%s'", args.Ami), nil)
+			case "InvalidAMIID.NotFound":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Ami '%s' not found in region '%s'", args.Ami, args.Region), nil)
+			case "InvalidAMIAttributeItemValue":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Invalid user id to share ami with: %v", args.ShareWithAccounts), nil)
+			}
+		}
+		return err
+	}
+
+	result.Ami = args.Ami
+	result.Region = args.Region
+	return nil
+}
--- a/internal/cloud/awscloud/awscloud.go
+++ b/internal/cloud/awscloud/awscloud.go
@@ -266,29 +266,6 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	snapshotID := importOutput.ImportSnapshotTasks[0].SnapshotTaskDetail.SnapshotId
 
-	if len(shareWith) > 0 {
-		logrus.Info("[AWS] 🎥 Sharing ec2 snapshot")
-		var userIds []*string
-		for _, v := range shareWith {
-			// Implicit memory alasing doesn't couse any bug in this case
-			/* #nosec G601 */
-			userIds = append(userIds, &v)
-		}
-		_, err := a.ec2.ModifySnapshotAttribute(
-			&ec2.ModifySnapshotAttributeInput{
-				Attribute:     aws.String("createVolumePermission"),
-				OperationType: aws.String("add"),
-				SnapshotId:    snapshotID,
-				UserIds:       userIds,
-			},
-		)
-		if err != nil {
-			logrus.Warnf("[AWS] 📨 Error sharing ec2 snapshot: %v", err)
-			return nil, err
-		}
-		logrus.Info("[AWS] 📨 Shared ec2 snapshot")
-	}
-
 	// Tag the snapshot with the image name.
 	req, _ := a.ec2.CreateTagsRequest(
 		&ec2.CreateTagsInput{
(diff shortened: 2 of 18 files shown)


cmd/osbuild-worker/jobimpl-awsec2.go
```diff
@@ -0,0 +1,177 @@
...
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	var args worker.AWSEC2CopyJob
+	err := job.Args(&args)
+	if err != nil {
+		return err
```
ondrejbudai, line 39: This codepath doesn't set `result.JobError`.

internal/cloud/awscloud/awscloud.go
```diff
@@ -348,31 +325,176 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
...
+	)
+	if err != nil {
+		return err
+	}
+	if len(imgs.Images) == 0 {
+		return fmt.Errorf("Unable to find image with id: %v", ami)
+	}
+
+	for _, bdm := range imgs.Images[0].BlockDeviceMappings {
+		err = a.shareImage(bdm.Ebs.SnapshotId, aws.String(ami), userIds)
```
ondrejbudai, line 445: This means that we will call `ModifyImageAttribute` with the same params multiple times, right? Might be better to have two methods: `shareAMI` that would get called once and `shareSnapshot` that would get multiple times depending on how many snapshots are contained in an AMI.

cmd/osbuild-worker/jobimpl-awsec2.go
```diff
@@ -0,0 +1,177 @@
...
+			case "InvalidAMIID.NotFound":
+				fallthrough // CopyImage returns InvalidRequest instead of InvalidAMIID.NotFound
+			case "InvalidRequest":
+				result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorSharingTarget, fmt.Sprintf("Source ami '%s' not found", args.Ami))
+			}
+		}
+		return nil
+	}
+
+	// Copy all permissions over from source ami
```
ondrejbudai, line 68: I just wonder - should this be optional? Depending on a `preservePermissions` boolean in the jobargs?

cmd/osbuild-worker/jobimpl-awsec2.go
```diff
@@ -0,0 +1,177 @@
...
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	var args worker.AWSEC2ShareJob
+	err := job.Args(&args)
+	if err != nil {
+		return err
```
ondrejbudai, line 131: no `JobError` again

ondrejbudai (commented):
Looks good, some nitpicks and ideas inline.

internal/cloud/awscloud/awscloud.go
Gundersanne: makes sense to split them, in practice I don't think we'll ever have an image with multiple snapshots, but splitting them is much better. Will do

internal/cloudapi/v2/handler.go
```diff
@@ -1178,3 +1186,183 @@ func genRepoConfig(repo Repository) (*rpmmd.RepoConfig, error) {
...
+		shareAmi := options.Ami
+		shareRegion := img.Region
+		if img.Region != options.Region {
+			// Let the share job use dynArgs
+			shareAmi = ""
+			shareRegion = ""
+
+			// Check deps if we need to do a copyjob
+			foundDep := false
+			for _, d := range deps {
```
thozza, line 1250: `deps` are dependencies of the osbuild job from above, not jobs which have the osbuild job as a dependency (I mean unless I'm heavily confused).

We never enqueue osbuild job with the ec2-copy job as its dependency and we also don't modify dependencies of the original osbuild job.

So my question is whether it makes any sense to check the osbuild job dependencies? 🤔 

Personally I think that the ec2-copy dependency can not be ever found and the code path will always follow into 
```go
if !foundDep {
   ...
}
```

internal/cloudapi/v2/handler.go
```diff
@@ -1178,3 +1186,183 @@ func genRepoConfig(repo Repository) (*rpmmd.RepoConfig, error) {
...
+		},
+		Id: finalJob.String(),
+	})
+}
+
+func (h *apiHandlers) GetImageFromComposeStatus(ctx echo.Context, id string) error {
+	return h.server.EnsureJobChannel(h.getImageFromComposeStatus)(ctx, id)
+}
+
+func (h *apiHandlers) getImageFromComposeStatus(ctx echo.Context, id string) error {
```
thozza, line 1313: Is there a reason why we can't or don't want to treat the ID returned by the `compose/{id}/image` endpoint as another compose ID?

If we did, like we do for Koji composes, we could get rid of the new status endpoint and just extend the `/compose/{id}` endpoint to handle new job type, in addition to the osbuild job and koji-finalize job.

thozza (commented):
Nice work! I added some comments inline...

internal/worker/server.go
```diff
@@ -362,6 +372,32 @@ func (s *Server) ManifestJobStatus(id uuid.UUID, result *ManifestJobByIDResult)
...
+		return nil, nil, err
+	}
+
+	if jobType != JobTypeAWSEC2Share {
+		return nil, nil, fmt.Errorf("expected %q, found %q job instead", JobTypeAWSEC2Share, jobType)
+	}
+
+	return status, deps, nil
+}
+
```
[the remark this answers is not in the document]
Gundersanne: Well we only do that for specific errors right? Like for depsolve errors it was for backwards compatibility with older results (which we could by now remove).

In general it's better if the job itself sets it imo.

internal/cloudapi/v2/handler.go
Gundersanne: Ugh you're right, this is wrong. I'll need to add something which queries dependents rather.

internal/cloudapi/v2/handler.go
Gundersanne: Makes sense, no  need to duplicate this everywhere.

internal/worker/server.go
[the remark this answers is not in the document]
thozza: > Well we only do that for specific errors right? Like for depsolve errors it was for backwards compatibility with older results (which we could by now remove).

Ideally yes, but I'm not 100% sure. I think that some job implementations still need to be enhanced to always report fail back to the worker server, but in general we should be getting there eventually.

> In general it's better if the job itself sets it imo.

I fully agree that the job should always set the proper error. I think that we set job errors in the `xxxJobStatus()` functions either due to backward compatibility or as a fail-safe. I was wondering whether we want to add it here mostly as a fail-safe.

I'm OK with not doing so... 😉

cmd/osbuild-worker/jobimpl-awsec2.go
Gundersanne: I did just consider it to be a copy which preserves attributes. Happy to remove the preserving and just always have a share job which achieves the same.

I think what the sharing does might still change as well. Could be that if you add a new account you want it on all images from the same compose, rather than just a specific 1 which is how it is now.

thozza: > I did just consider it to be a copy which preserves attributes. Happy to remove the preserving and just always have a share job which achieves the same.

I think that it would make sense. This way we would have more flexibility (we can share the image with any account we want, including the original one, using the share job) and we would not need to consider extending the job options whether to preserve the permissions or not (since we would be able to do any of those with the share job).

internal/cloudapi/v2/handler.go
Gundersanne: hm, going through it, it does return a different kind of object, just the upload status part really.

cmd/osbuild-worker/jobimpl-awsec2.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+func getAWS(awsCreds, region string) (*awscloud.AWS, error) {
+	if awsCreds != "" {
+		return awscloud.NewFromFile(awsCreds, region)
+	}
+	return awscloud.NewDefault(region)
+}
```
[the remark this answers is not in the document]
Gundersanne: Trying this out, I'm not a huge fan, unless we want to move this to internal/cloud itself. Plus they all need slightly different things. The way we instantiate  an aws session for s3 is different than for ec2 in osbuild jobs, and then the share/copy jobs would never accept an access key/secret from the options as these jobs don't really make sense for the weldr api.

cmd/osbuild-worker/jobimpl-awsec2.go
Gundersanne: Changed this now to use the copy job, the implementation is a bit quirky though, as it now reads both the input and the output of the parent osbuild job.

internal/cloudapi/v2/handler.go
ondrejbudai: `compose/{id}/image` makes sense to me, as an image always belongs to an existing compose. Also, the returned data are different, so I would rather not merge them.

cmd/osbuild-worker/jobimpl-awsec2.go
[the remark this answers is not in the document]
ondrejbudai: No hard feelings tbh, the duplicated code feels manageable to me.

cmd/osbuild-worker/jobimpl-awsec2.go
Gundersanne: Should we make the shared accounts part of the osbuild target result?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -211,6 +211,101 @@ paths:
               schema:
                 type: string
 
+  /composes/{id}/image:
```
ondrejbudai, line 214: I've realized that I'm not sure if I like that we are calling this new entity an "image". My reasoning is that if you are a new person wanting to use this API, you will look for the "image" keyword which leads you to this new route. But often, creating a compose is just fine (e.g. for the koji usecase), you don't even need to copy the image in order to use it. I don't have a better name though... :thinking:

Gundersanne: Yea I agree the naming might be a bit confusing. First we had 'promotions', but the idea is that this is creating another image with the same content, just in a new location. And the only way to really guarantee that it's the same content is to copy/move around the blob (rebuilding an old manifest can have too many side effects, especially with user defined repos). 

The key is that you need a compose before you can create an image from it, and in the CRC api I think we should work towards an /images endpoint which parses both the compose result, and any images coming from that compose.

internal/cloudapi/v2/handler.go
```diff
@@ -1180,3 +1180,227 @@ func genRepoConfig(repo Repository) (*rpmmd.RepoConfig, error) {
...
+	}
+
+	if len(osbuildJob.Targets) != 1 {
+		return HTTPError(ErrorSeveralUploadTargets)
+	}
+
+	// the id of the last job in the dependency chain which users should wait on
+	finalJob := jobId
+	// look at the upload status of the osbuild dependency to decide what to do
+	switch us.Type {
```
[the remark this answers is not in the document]
Gundersanne: Oki, I was just half anticipating more of these, but I guess let's do that when we're there.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -601,3 +610,27 @@ func (q *DBJobQueue) jobDependencies(ctx context.Context, conn *pgxpool.Conn, id
 
 	return dependencies, nil
 }
+
+func (q *DBJobQueue) jobDependents(ctx context.Context, conn *pgxpool.Conn, id uuid.UUID) ([]uuid.UUID, error) {
+	rows, err := conn.Query(ctx, sqlQueryDependents, id)
```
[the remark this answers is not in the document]
Gundersanne: Hm wait, do you mean put the entire JobStatus (which are 3 separate queries), in 1 tx?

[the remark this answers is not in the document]
Gundersanne: So I think it doesn't make sense to query the dependencies in the tx because you can't change those after the fact.

For the dependents however, which are also queried in jobstatus, I'm not sure. Adding a dependency to a job doesn't change the job in any way, it just updates the dependencies table. So I don't fully understand where the inconsistency would arise.

My understanding is that transactions for multiple reads are useful in case subsequent queries are based on the results of previous queries, in which case you'd really want the data consistency. As the job's uuid is consistent throughout here, and it's the only piece of data required for the dependencies/dependents queries, I don't think inconsistency is possible.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 0f5303923f82f61aefe1919a31ae7ddc1f37fea2 with the main merge-base 53d4d7f2849a8f3064f5db8468990bfe3e681e76).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2888744839/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2888744839/artifacts/browse

internal/cloudapi/v2/openapi.v2.yml
Gundersanne: Another option would be to just have a POST request  on /composes/{id}; and expand the GET request to include any copied and/or reshared images.

thozza: What about calling it simply `copy`?

I agree that accepting `/composes/{id}` to accept POST and extending the GET seems a bit obscure.

Gundersanne: Well for s3 images this might become the endpoint to regenerate the s3 url on an object. Because of the image storage limitations of gcp, we might need to store those as a blob and this would become the endpoint to upload/register the image into gcp etc... So it's a bit more than just copy

Gundersanne: `/composes/{id}/share` might be more correct ?

thozza: So, given your comments, the purpose of the endpoint is that exactly the same image build may get copied to a different region, shared with different accounts and there may be more going on in the background. But that should not be something to bother the user with (at least IMHO as it is an implementation detail if we e.g. reimport a blob from S3 to GCP). But the core point seems to be that the image won't be rebuilt, but used as it was built in the past. 

Funnily enough, the `/composes/{id}/share` endpoint seems restrictive to me, because more than just "sharing` could be happening - e.g. the image will be available in a new region.

Given all of that, I'll suggest what I didn't last time.  That is `/composes/{id}/clone`. This could be expressing that the original result of the compose is used as is and the ability to share it with different accounts or to copy to a different region does not seem too weird.

Gundersanne: I guess we'll end up with `/clones/{id}` then for a status endpoint?

@teg wdyt?

internal/cloudapi/v2/handler.go
```diff
@@ -1227,3 +1227,226 @@ func genRepoConfig(repo Repository) (*rpmmd.RepoConfig, error) {
...
+
+	if js.Started.IsZero() {
+		return UploadStatusValuePending
+	}
+
+	if js.Finished.IsZero() {
+		return UploadStatusValueRunning
+	}
+	return UploadStatusValueSuccess
+}
```
thozza, line 1452: It would be great if this could be used also in https://github.com/osbuild/osbuild-composer/blob/af65baa6fe38364f28c1b6630ea1303c221091c6/internal/cloudapi/v2/handler.go#L644-L645 and https://github.com/osbuild/osbuild-composer/blob/af65baa6fe38364f28c1b6630ea1303c221091c6/internal/cloudapi/v2/handler.go#L701-L702
so that we are determining the upload status consistently in the implementation.

thozza (commented):
Except for comments and the endpoint name discussion still ongoing, this all looks great! Nice job 👍

internal/cloudapi/v2/openapi.v2.yml
thozza: Didn't see the last comment before I finished my review from yesterday, but I like it!

cmd/osbuild-worker/jobimpl-awsec2.go
```diff
@@ -0,0 +1,144 @@
...
+			return nil
+		}
+		var cjResult worker.AWSEC2CopyJobResult
+		err = job.DynamicArgs(0, &cjResult)
+		if err != nil {
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorParsingDynamicArgs, "Error parsing dynamic args as ec2 copy job", nil)
+			return err
+		}
+		if cjResult.JobError != nil {
+			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorJobDependency, "AWSEC2CopyJob dependency failed", nil)
```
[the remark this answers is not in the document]
Gundersanne: This will be done by the worker server when querying the status of this job no?

internal/cloudapi/v2/openapi.v2.yml
Gundersanne: Also @ondrejbudai wdyt?

pkg/jobqueue/dbjobqueue/dbjobqueue.go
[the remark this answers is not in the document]
ondrejbudai: I was mostly concerned about a write in-between two reads but we don't actually care about this in this case since both possible outcomes are valid.

internal/cloudapi/v2/openapi.v2.yml
ondrejbudai: Works for me. :)

test/cases/api.sh
```diff
@@ -418,6 +419,64 @@ function waitForState() {
     export UPLOAD_OPTIONS
 }
 
+function sendImgFromCompose() {
```
ondrejbudai, line 422: This should probably be something like `cloneCompose`

test/cases/api.sh
```diff
@@ -418,6 +419,64 @@ function waitForState() {
...
+                 --data @"$1" \
+                 --write-out '%{http_code}' \
+                 --output "$OUTPUT" \
+                 https://localhost/api/image-builder-composer/v2/composes/"$COMPOSE_ID"/clone)
+
+    test "$HTTPSTATUS" = "201"
+    IMG_ID=$(jq -r '.id' "$OUTPUT")
+}
+
+function waitForImgState() {
```
ondrejbudai, line 441: ditto

cmd/osbuild-worker/jobimpl-awsec2.go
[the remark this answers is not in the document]
thozza: You are right.

thozza (approved):
I have nothing to add on top of latests comment from Ondrej. Great work!

ondrejbudai:

PR #2811, opened 2022-07-08
Introduce logging adapter for jobqueue

lzap:
Thanks for extracting jobqueue into pkg/. Now, in order to use it in our project, we have two changes to make:

* We rely on database/sql connection pool instead of pgx.
* We do not use logrus.

This patch aims to solve the latter by creating a very simple facade via interface called `SimpleLogger` (naming things is hard bring your own suggestions). It seems the jobqueue only uses logging messages at info level and then error level, so the interface only supports these to keep things simple. It makes sense, as a user of any library you typically want to either enable or disable logging, verbosity mapping has always been problematic thing to solve.

The patch introduces a logrus implementation and a noop implementation. It will require a change in all callers of the New function, I tried my best to find all uses, but since I do not have a full development setup yet, I could not test this properly.
 
This pull request includes:

- [x] adequate documentation informing people about the change (not relevant I guess)

I am going to make a draft first to gather feedback. I would appreciate any help in regard to how to test this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ __pycache__
 
 .vscode
 .DS_Store
+/.idea
 
 /containers/config/
 /bin
--- a/internal/common/slogger/logrus.go
+++ b/internal/common/slogger/logrus.go
@@ -1,0 +1,38 @@
+package slogger
+
+import (
+	"github.com/osbuild/osbuild-composer/pkg/jobqueue"
+	"github.com/sirupsen/logrus"
+)
+
+type simpleLogrus struct {
+	logger *logrus.Logger
+}
+
+func NewLogrusLogger(logger *logrus.Logger) jobqueue.SimpleLogger {
+	return &simpleLogrus{logger: logger}
+}
+
+func (s *simpleLogrus) log(level logrus.Level, err error, msg string, args ...string) {
+	if len(args)%2 != 0 {
+		panic("log arguments must be even (key value pairs)")
+	}
+	var fields = make(logrus.Fields, len(args)/2+1)
+	for i := 0; i < len(args); i += 2 {
+		k := args[i]
+		v := args[i+1]
+		fields[k] = v
+	}
+	if err != nil {
+		fields["error"] = err.Error()
+	}
+	s.logger.WithFields(fields).Log(level, msg)
+}
+
+func (s *simpleLogrus) Info(msg string, args ...string) {
+	s.log(logrus.InfoLevel, nil, msg, args...)
+}
+
+func (s *simpleLogrus) Error(err error, msg string, args ...string) {
+	s.log(logrus.ErrorLevel, err, msg, args...)
+}
--- a/pkg/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/pkg/jobqueue/dbjobqueue/dbjobqueue.go
@@ -18,10 +18,9 @@ import (
 	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
-
+	"github.com/osbuild/osbuild-composer/internal/common/slogger"
 	"github.com/osbuild/osbuild-composer/pkg/jobqueue"
-
-	logrus "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -106,6 +105,7 @@ const (
 )
 
 type DBJobQueue struct {
+	logger       jobqueue.SimpleLogger
 	pool         *pgxpool.Pool
 	dequeuers    *dequeuers
 	stopListener func()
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@ __pycache__
 /rpmbuild
 
 /tools/appsre-ansible/inventory
+/.idea
```
gicmo, line 11: Sure, but maybe a different commit, thanks! ;)

lzap: Absolutely. I am hoping for some opinions on the actual logging work tho :-)

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
```
lzap, line 2: I think my JetBrains IDE did this, can I leave this? I am assuming this is some kind of new syntax?

lzap:
Rebased, split the IDEA git ignore hunk into a separate commit.

ondrejbudai (commented):
Looks alright to me, it doesn't add too much complexity and if this works for you, I'm happy to merge this. Some comments are inline that I would like to be resolved first though.

The code is currently not buildable, though.

ondrejbudai:
Hi @lzap, do you have any plans to get this finished? I'm currently in the process of cleaning up stale PRs. :) Thanks!

internal/common/slogger/logrus.go
```diff
@@ -0,0 +1,52 @@
...
+	var fields = make(logrus.Fields, len(args)/2+1)
+	var sb strings.Builder
+	sb.WriteString(msg)
+	if len(args) > 0 || err != nil {
+		sb.WriteString(":")
+	}
+	for i := 0; i < len(args); i += 2 {
+		k := args[i]
+		v := args[i+1]
+		fields[k] = v
```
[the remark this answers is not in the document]
lzap: I want to be extra careful not to drop any bit of information in case you run some configurations without structured logging (e.g. text-based). Your surprise reveals, however, that this is unlikely the case, removing :-)

lzap:
Apologies, I rebased.

I feel like I should remain the old behavior when job queue is created via `New` function, that is logging into the standard logrus logger by default.

But I created a new function `NewWithConfig` and introduced a new `Config` type which allows overriding this behavior. The no operation logger is currently unused, but I think it's a good idea to have it in the codebase - it has literally zero code so there can't be bugs (I think) :-)

Should be building now, fixed some linter complaints.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -234,8 +251,8 @@ func (q *DBJobQueue) Enqueue(jobType string, args interface{}, dependencies []uu
 	}
 	defer func() {
 		err := tx.Rollback(context.Background())
-		if err != nil && !errors.Is(err, pgx.ErrTxClosed) {
-			logrus.Error("error rolling back enqueue transaction: ", err)
+		if err != nil && !errors.As(err, &pgx.ErrTxClosed) {
```
ondrejbudai, line 254: Bad merge (breaks build by Go 1.19).

ondrejbudai (changes requested):
Looks good but there's one wrongly merge line. @diaasami can you also have a look?

pkg/jobqueue/dbjobqueue/dbjobqueue.go
lzap: Damn, I have many conflicts and I hoped to catch them all (TM).

Good eyes, rebased!

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -202,13 +219,13 @@ func (q *DBJobQueue) listen(ctx context.Context, ready chan<- struct{}) {
...
 			// shutdown the listener if the context is canceled
 			if errors.Is(err, context.Canceled) {
-				logrus.Info("Shutting down the listener")
+				q.logger.Info("Shutting down the listener")
 				return
 			}
 
 			// otherwise, just log the error and continue, there might just
 			// be a temporary networking issue
-			logrus.Debugf("error waiting for notification on jobs channel: %v", err)
```
ondrejbudai, line 211: Hmm, I wonder if we wanted to introduce a `Debug` level to `slogger`? Would it be valuable for your needs?

lzap: I can turn this into info or introduce debug level. To my ears, temporary network issues are network issues. These should be considered as errors from the operational perspective ;-)

My concern is once we start adding levels, these can be a challenge to map onto levels from underlying implementations, therefore I was aiming for as minimum interface as possible (info/error only).

ondrejbudai: Let's just keep them as errors right now, if this becomes too spammy, we can either modify our filters, lower the severity, or drop it. You are right, it's an error, let's treat it as such.

ondrejbudai (dismissed):
All good!

lzap:
Rebased! Thanks.

ondrejbudai (approved):

lzap:
New OSTree pipeline failed which I do not fully understand what is going on there.

PR #2812, opened 2022-07-08
manifest+playground: improve developer experience

teg:
This simplifies and refines the pipeline definitions and makes it
easier to work with the playground.

The build pipelines are trimmed down a bit by only include required
packages.

### DIFF
--- a/internal/disk/partition_table.go
+++ b/internal/disk/partition_table.go
@@ -571,3 +571,54 @@ func (pt *PartitionTable) ensureLVM() error {
 
 	return nil
 }
+
+func (pt *PartitionTable) GetBuildPackages() []string {
+	packages := []string{}
+
+	hasLVM := false
+	hasBtrfs := false
+	hasXFS := false
+	hasFAT := false
+	hasEXT4 := false
+
+	introspectPT := func(e Entity, path []Entity) error {
+		switch ent := e.(type) {
+		case *LVMLogicalVolume:
+			hasLVM = true
+		case *Btrfs:
+			hasBtrfs = true
+		case *Filesystem:
+			switch ent.GetFSType() {
+			case "vfat":
+				hasFAT = true
+			case "btrfs":
+				hasBtrfs = true
+			case "xfs":
+				hasXFS = true
+			case "ext4":
+				hasEXT4 = true
+			}
+		}
+		return nil
+	}
+	_ = pt.ForEachEntity(introspectPT)
+
+	// TODO: LUKS
+	if hasLVM {
+		packages = append(packages, "lvm2")
+	}
+	if hasBtrfs {
+		packages = append(packages, "btrfs-progs")
+	}
+	if hasXFS {
+		packages = append(packages, "xfsprogs")
+	}
+	if hasFAT {
+		packages = append(packages, "dosfstools")
+	}
+	if hasEXT4 {
+		packages = append(packages, "e2fsprogs")
+	}
+
+	return packages
+}
--- a/internal/manifest/build.go
+++ b/internal/manifest/build.go
@@ -3,6 +3,7 @@ package manifest
 import (
 	"github.com/osbuild/osbuild-composer/internal/osbuild2"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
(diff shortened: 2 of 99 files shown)


achilleas-k (dismissed):
I like this very much.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 5d1100d41f3b9c150c6d543c1c90b18819b2f1da with the main merge-base 2c2d4812d4b381275e8c52d86ec81b7949b42cfe).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 1122a65ad457a48e3db9f84e366030da9911a322 with the main merge-base 2c2d4812d4b381275e8c52d86ec81b7949b42cfe).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

internal/manifest/build.go
```diff
@@ -15,16 +16,18 @@ import (
 type BuildPipeline struct {
 	BasePipeline
 
+	runner       runner.Runner
```
gicmo, line 19: This is interesting since it is like the kernel version a transient property, i.e. it depends on the specific package set instead of some configuration.

internal/manifest/commit_server_tree.go
```diff
@@ -20,6 +21,7 @@ type OSTreeCommitServer struct {
 	// TODO: should this be configurable?
 	Language string
 
+	platform        platform.Platform
```
gicmo, line 24: Do we anticipate that we need the platform outside of `serialise`? It seems like it is another one of these transient properties that should ideally only be relevant during the serialisation process.

gicmo (dismissed):
I continue to be very excited where this is headed! There are a few nits (again, very long commit lines) and the last commit ideally would be before `regenerate to drop build packages` so we keep a green history, but it is not super important.

More on the general architecture side of things I am curious about the "transient" properties and how we handle them in the future. Some pipeline do really depend "more" on the architecture / platform than others so it might indeed be good to encode that explicitly; otoh it also seems like something that could also be only effective / relevant during the serialisation process so could be a properly that process.

internal/disk/partition_table.go
```diff
@@ -571,3 +571,54 @@ func (pt *PartitionTable) ensureLVM() error {
...
 }
+
+func (pt *PartitionTable) GetBuildPackages() []string {
+	packages := []string{}
+
+	hasLVM := false
+	hasBtrfs := false
+	hasXFS := false
+	hasFAT := false
+	hasEXT4 := false
```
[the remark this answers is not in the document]
teg: Yeah, though I'll leave that for a potential follow-up (not really sure about the pros/cons or if it makes a difference).

teg:
> There are a few nits (again, very long commit lines) and the last commit ideally would be before regenerate to drop build packages so we keep a green history, but it is not super important.

Fixed! (and sorry...)

> More on the general architecture side of things I am curious about the "transient" properties and how we handle them in the future. Some pipeline do really depend "more" on the architecture / platform than others so it might indeed be good to encode that explicitly; otoh it also seems like something that could also be only effective / relevant during the serialisation process so could be a properly that process.

Yeah, this is something to take stock of and reconsider once we see a bit more how the code develops after porting a few things over I think. So far my thinking has been that the way this would be used in practice is:

1. declare all the configuration you want for an image in some struct
2. call some function that generates a manifest based on that config
3. get packages from the manifest
4. depsolve
5. serialize

If you wanted to serialize something with a different platform (say) but everything else the same, then just change the config struct in step 1 and start over from 2.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 5440ae3ec9ef5e8f64550e03932fc7c7f1e1affe with the main merge-base 2c2d4812d4b381275e8c52d86ec81b7949b42cfe).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

gicmo (approved):
❤️

PR #2817, opened 2022-07-11
fedora: add image-installer image type

ondrejbudai:
image

(the title says it all, will add a description later)

This pull request includes:


internal/distro/fedora/package_sets.go
```diff
@@ -524,3 +525,19 @@ func containerPackageSet(t *imageType) rpmmd.PackageSet {
...
+
+func bareMetalPackageSet(t *imageType) rpmmd.PackageSet {
+	return rpmmd.PackageSet{
+		Include: []string{
+			"policycoreutils",
+			"selinux-policy-targeted",
+			"kernel",
+			"firewalld",
+			"chrony",
+			"langpacks-en",
```
teg, line 537: I believe many of these packages are implicitly included, so this could be simplified.

internal/distro/fedora/manifests.go
```diff
@@ -460,3 +460,48 @@ func bootISOPipeline(m *manifest.Manifest,
 	p.Filename = filename
 	return p
 }
+
+func tarBootISOTreePipeline(m *manifest.Manifest,
+	buildPipeline *manifest.Build,
+	anacondaPipeline *manifest.Anaconda,
+	treePipelineName,
```
teg, line 467: We should reference pipelines by reference, not by name, that way we get type safety.

internal/manifest/iso_tree.go
```diff
@@ -148,6 +135,17 @@ func bootISOMonoStageOptions(kernelVer, arch, vendor, product, osVersion, isolab
 	}
 }
 
+type ISOTreePayload interface {
```
teg, line 138: Not sure I love this abstraction. Really what we want is a proper union, but golang... How about we just make both the `OSTree` and the `Payload` arguments (which should be of type `pipelien.OS`) optional, assert on exactly one being set, and then just generate the right stages in the pipeline function?

internal/manifest/iso_tree_ostree_payload.go
```diff
@@ -0,0 +1,64 @@
...
+			url:      o.osTreeURL,
+		},
+	}
+}
+
+func (o osTreeISOTreePayload) getOSName() string {
+	return o.osName
+}
+
+func (o osTreeISOTreePayload) getPayloadStages() []*osbuild2.Stage {
```
teg, line 56: So far all stages were generated only in the pipeline types, not in helpers, if we go this route (though I suggested above an alternative), would it make sense to just return the `osTreeCommit` and `osTreeRef` properties from this function, and instead do the stage stuff in the pipeline generation function? That way we could reuse this even if we change how the stages are put together.

teg (commented):
I love that you are picking this up! I left a smattering of comments, don't let any of those hold you back though, important thing is to get the tests to work ;)

ondrejbudai:
superseded by #2878

Issue #2767, opened 2022-06-22
Improvements for gen-manifests tool and Manifest-diff test


achilleas-k:
The `gen-manifests` tool produces a lot of progress output meant for running in a terminal but it's also used in CI in the `Manifest-diff` test.  In the test environment, the frequent progress output makes it hard to see what's going on, so it's silenced with a `> /dev/null` redirect.  This means that collected errors that's printed at the end isn't visible.

- [ ] Add command line flags to control output frequency or _noisiness_ (e.g., `--quiet` or `--output-frequency` in seconds).

---

For local development and testing, it might be useful to limit the generation to specific distros, architectures, or image types.  This is possible with the `generate-test-cases` and `generate-all-test-cases` scripts.

- [x] Add command line flags for limiting manifest generation

---

Jobs often produce output on stderr from dnf that we don't capture and makes the output noisy.

- [ ] (Try to) Capture stderr and prefix it with the job name or combine it with the error output at the end.

---

Sometimes manifest generation might fail.  Currently we fail the job, but it might be useful to provide information (through a comment on the PR) when this happens and not fail the job when the `merge-base` generation fails.  This happened recently with manifests failing on `main`, which required merging the fix PR on red.  If PR `HEAD` succeeds and `merge-base main` fails, the PR should be valid.

- [x] If the PR `HEAD` manifest generation fails, fail the job (manifest generation in a PR should always work).
- [x] If the `merge-base` manifest generation fails, do **not** fail the job, but write an appropriate comment on the PR.

---

Most PRs run the CI pipeline multiple times while they're being developed in draft mode or reviewed after they are first opened.  Schutzbot will make multiple comments, one for each run, if there are changes in the manifests.  Also, if the initial version of a PR changes the manifests when it shouldn't have and the PR is revised to fix this issue, the original comment will remain.  We should have Schutzbot clean up old comments when the state changes.

- [ ] Delete outdated comments when the pipeline is re-rerun.  Comments can be identified by author (Schutzbot) but it might be useful to have a fixed UUID attached to the comment to protect against other comments we might add to the bot in the future that could be deleted by this process.

---

The comment simply mentions the diffs and asks the people to review the changes.  We should provide some useful links or info.

Initial, simple improvement:
- [ ] Link to the job log
- [x] Put the diffs in the artifacts of the job

Future improvement:
- [ ] Add the diffs to the comment, folded
- [ ] Separate the diffs by file: one fold per file

[...]

PR #2821, opened 2022-07-11
Improvements for gen-manifests tool and Manifest-diff test

achilleas-k:
Partially addresses issue #2767.

Changes for `gen-manifests`:
- Prints the final error summary on stderr.
- Only print progress update after receiving a message on the message queue.  This makes the command less noisy in CI where the carriage return `\r` produces a new line and constant progress updates were producing new lines.  The logic relies on the fact that each job reports when its finished with a line on the message queue, which could change, but for now is adequate.
- Support `-arches`, `-distros`, and `-images` flags.  Each supports is a comma-separated list of strings for manifest generation filtering.
- Appends the text `[failed]` to the `Finished` message for failed jobs.

Changes for `diff-manifests` test:
- Print all the output from `gen-manifests`.
- The test now fails if manifest generation fails on the PR `HEAD`.
- The test does not fail if manifest generation fails on the `merge-base`, but a note is added to the review comment.
- The comment links directly to the artifacts directory for the CI job if there are differences in the manifests.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -122,8 +122,14 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 			Parent: cr.OSTree.Parent,
 		}
 	}
-	job := func(msgq chan string) error {
-		defer func() { msgq <- fmt.Sprintf("Finished job %s", filename) }()
+	job := func(msgq chan string) (err error) {
+		defer func() {
+			msg := fmt.Sprintf("Finished job %s", filename)
+			if err != nil {
+				msg += " [failed]"
+			}
+			msgq <- msg
+		}()
 		msgq <- fmt.Sprintf("Starting job %s", filename)
 		repos := convertRepos(cr.Repositories)
 		var bp blueprint.Blueprint
@@ -139,10 +145,12 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 
 		packageSpecs, err := depsolve(cacheDir, imgType, bp, options, repos, distribution, archName)
 		if err != nil {
-			return fmt.Errorf("[%s] depsolve failed: %s", filename, err.Error())
+			err = fmt.Errorf("[%s] depsolve failed: %s", filename, err.Error())
+			return
 		}
 		if packageSpecs == nil {
-			return fmt.Errorf("[%s] nil package specs", filename)
+			err = fmt.Errorf("[%s] nil package specs", filename)
+			return
 		}
 		if options.OSTree.Ref == "" {
 			// use default OSTreeRef for image type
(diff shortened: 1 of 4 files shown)


cmd/gen-manifests/main.go
```diff
@@ -305,9 +317,21 @@ func mergeOverrides(base, overrides composeRequest) composeRequest {
...
 
 func main() {
+	// common args
 	outputDirFlag := flag.String("output", "test/data/manifests.plain/", "manifest store directory")
 	nWorkersFlag := flag.Int("workers", 16, "number of workers to run concurrently")
 	cacheRootFlag := flag.String("cache", "/tmp/rpmmd", "rpm metadata cache directory")
+
+	// manifest selection args
+	archesFlag := new(multiValue)
+	flag.Var(archesFlag, "arches", "architectures")
```
ondrejbudai, line 338: Please, add a comment about commas.

Also, if you change `multiValue` a bit, you can drop the comma thing and just allow the user to use `-arches` multiple times. https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-upload-azure/main.go#L52 does something similar.

cmd/gen-manifests/main.go
```diff
@@ -121,7 +121,14 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 		}
 	}
 	job := func(msgq chan string) error {
```
ondrejbudai, line 123: You can switch to

```
func(msgq chan string) (err error)
```

and then drop `success` and just rely on `err != nil`. No hard feelings about this though.

test/cases/diff-manifests.sh
```diff
@@ -50,12 +50,20 @@ sudo dnf install -y go
 manifestdir=$(mktemp -d)
 
 greenprint "Generating all manifests for HEAD (PR #${prnum})"
+err=0
 go run ./cmd/gen-manifests --output "${manifestdir}/PR" --workers 50
```
ondrejbudai, line 54: I think you forgot `|| err=$?`

cmd/gen-manifests/main.go
achilleas-k: I like having the option to put a bunch on a list without needing to type out the argument multiple times, but allowing multiples is also useful at times.  Would it be weird if it supported both?  In other words, if
```
... -arches x86_64,aarch64 -arches s390x
```
was a valid way to specify 3 architectures?

test/cases/diff-manifests.sh
achilleas-k: I did.  Thanks!

achilleas-k:
All cleaned up.  Kept the comma-separated list only.

cmd/gen-manifests/main.go
```diff
@@ -325,17 +345,33 @@ func main() {
...
+			distroArches = distribution.ListArches()
+		}
+		for _, archName := range distroArches {
 			arch, err := distribution.GetArch(archName)
 			if err != nil {
-				panic(fmt.Sprintf("distro %q lists arch %q but GetArch() failed: %s", distroName, archName, err.Error()))
+				panic(fmt.Sprintf("invalid arch name %q for distro %q: %s", archName, distroName, err.Error()))
+			}
+
+			daImgTypes := imgTypes
```
ondrejbudai, line 377: is this really needed? :O

ondrejbudai (commented):
Looks awesome, one minor nitpick left!

cmd/gen-manifests/main.go
ondrejbudai: Riiiiight!

ondrejbudai (dismissed):

achilleas-k:
Added new commit that fixes the artifact collection and rebased on main.

gicmo (approved):
❤️

PR #2822, opened 2022-07-12
Remove koji API and the osbuild-koji job

ondrejbudai:
It's no longer used anywhere, let's just drop it. I strongly believe that we don't need to keep anything from the API nor the job to keep backward compatibility. When migrating the internal service, we dropped all past jobs. Fedora deployment is still experimental and not used in production. Therefore, I believe that full-drop is 100% possible.

This pull request includes:


internal/cloudapi/v2/v2_koji_test.go
```diff
@@ -553,7 +552,7 @@ func TestKojiRequest(t *testing.T) {
 	handler.ServeHTTP(rec, req)
 	resp := rec.Result()
 
-	var status api.Status
+	var status interface{}
```
thozza, line 555: nitpick: Shouldn't this be `ComposeStatus` instead of `interface{}`? Nevertheless I can not think of any situation when using `interface{}` in this unit test would cause an issue or cover a potential bug.

internal/cloudapi/v2/v2_koji_test.go
```diff
@@ -635,7 +635,7 @@ func TestKojiJobTypeValidation(t *testing.T) {
 
 	t.Logf("%q job ID: %s", worker.JobTypeKojiInit, initID)
 	t.Logf("%q job ID: %s", worker.JobTypeKojiFinalize, finalizeID)
-	t.Logf("%q job IDs: %v", worker.JobTypeOSBuildKoji, buildJobIDs)
+	t.Logf("%q job IDs: %v", worker.JobTypeOSBuild, buildJobIDs)
```
thozza, line 638: ups, thanks for fixing 🙄

thozza (dismissed):
It is great to see `kojiapi` going away, together with all the backward compatibility code... 😍 All looks good from my PoV 👍

teg (dismissed):
Yay!

gicmo:
This seems to need a rebase :(

internal/cloudapi/v2/v2_koji_test.go
ondrejbudai: I prepended a commit that removes this test while making sure that everything is still covered (in even better way).

ondrejbudai:
Rebased!

thozza (approved):
Still looking good 🚀

thozza (approved):

ondrejbudai:
Fedora 35 AWS test is failing but me and @thozza agreed that this is not connected with the Koji API in anyway and it's safe to merge this.

PR #2805, opened 2022-07-06
Merge RHEL 8.4 distro definition into distro/rhel8 package

achilleas-k:
RHEL 8.4 is now defined in the RHEL 8 `distro/rhel8` package.

Manifest changes are hard to track because (most) RHEL 8.4 manifests were in the old (v1) format.  I converted the manifests to a v2 format before updating them to make comparing easier, but it doesn't help much.
I regenerated the image info for 8.4 image types after the merge for more clarity on what changed.

RHEL 8.4 now supports all cloud image types supported by RHEL 8.6.  Repositories were added to support the ec2 variants (HA, SAP, SAPHANA).
The only image types not supported by RHEL 8.4 are the ones that require FIDO: `edge-raw` and `edge-simplified-installer`.

I enabled testing for RHEL 8.4 since the definitions of the images for this distro are no longer "frozen".

PR #2823, opened 2022-07-12
Remove osbuild1 package

achilleas-k:
Since #2805, the `osbuild1` package is no longer required.  We only keep around some code for parsing results from jobs that used v1 manifests.  The `osbuild2` package has also been renamed to plain `osbuild`.


thozza (approved):
LGTM ❤️

teg (approved):
Thank you!

gicmo (commented):
🚀 🎸

PR #2825, opened 2022-07-12
Regenerate fedora-35 manifests + switch RHOS-01 to non ssd

jrusz:
I need the new manifests to hopefully finish https://github.com/osbuild/osbuild/pull/1051

I'm also switching to use non-ssd instances in RHOS-01 because there is apparently not enough hypervisors. Support ticket is opened, will revert once it's resolved. The terraform commit is from a separate branch in the gitlab-ci-terraform repo, I hope this situation won't last long so I don't think it needs to be merged into main.

This pull request includes:


atodorov (approved):

jrusz:
I just got a reply on my ticket, that more hypervisors were added and it does seem to be that way, didn't expect such a quick fix so I've removed the commit switching the instance flavor.

PR #2828, opened 2022-07-13
packit: Enable Bodhi update feature

ochosi:
While this feature is 'not mature yet' according to Packit developers, we can enable it because there's no harm done. If Packit fails to publish the Bodhi update then fedora-bot will take care of it.

For reference see https://packit.dev/docs/configuration/#bodhi_update


thozza (approved):
👍

PR #2833, opened 2022-07-14
use fdo-aio in edge simplified installer test

yih-redhat:
This pull request includes:


test/data/ansible/check_ostree.yaml
```diff
@@ -45,28 +45,30 @@
     - set_fact:
         checking_stage: "{{ result_stage.stdout }}"
 
+    # Remove this check because sometimes onboarding happens too fast and /boot/device-credentials will be gone,
```
henrywang, line 48: We have to keep this test because this case is the only case we check FDO onboarding. We have to find a way to make it stable.

henrywang (changes requested):
Thanks @yih-redhat for doing this. I have my comment.

osbuild-composer.spec
```diff
@@ -402,6 +402,7 @@ Requires:   dnf-plugins-core
 Requires:   skopeo
 Requires:   make
 Requires:   python3-pip
+Requires:   fdo-admin-cli
```
henrywang, line 405: There's no this package on Fedora yet. That might caused osbuild-composer-test installation error on Fedora.
`nothing provides fdo-admin-cli needed by osbuild-composer-tests-58-1.20220719gitba5e4ab.fc35.x86_64`
This package should not be added here, but install this package in ostree-simplified-installer.sh. And more important is Fedora should keep use fdo container, and update to fdo package on RHEL and CentOS Stream only.

henrywang (changes requested):

test/data/ansible/check_ostree.yaml
```diff
@@ -48,6 +48,8 @@
 
     # case: check /boot/device-credentials exists
     # simplified installer installed Edge system ONLY
+    # This case is not stable because sometimes onboarding happens too fast and /boot/device-credentials will be deleted very soon,
```
henrywang, line 51: Could you please find a better solution to make it stable? Thanks.

henrywang (changes requested):

henrywang:
@yih-redhat Could you please squash three commits into one, because the last two commits do not make any sense and they're confused when merged into `main`. Thanks.

osbuild-composer.spec
yih-redhat: removed, thanks.

test/data/ansible/check_ostree.yaml
yih-redhat: Now I will wait and check /etc/device-credentials, this file only exists after onboarding completed.

yih-redhat:
> 

This pull was created weeks ago, and after rebase there are too many commits. It's hard to find the original commits in git command line and squash them. Is there an easier way to do it? Or we can leave it now just for this pr?

henrywang:
Your two commits `Merge branch 'osbuild:main' into add-fdo-aio` and `Merge branch 'main' into add-fdo-aio` do not make any sense. You can close this PR and send a new one if you do not know how to squash them. Thanks.

yih-redhat:
The only failed job is on CS9, I rerun this job several times and it always fail at http boot step. Then I tried it in my local cs9 machine, it passed without problem.

yih-redhat:
Closing this pull request.

PR #2834, opened 2022-07-14
blueprint: Hash all user passwords

bcl:
This commit changes blueprint behavior to always store the hash of the
password for the 'customizations.user' accounts. Note that missing or
blank passwords are not hashed and should be dealt with at a lower
layer.

Resolves: rhbz#2107358

This pull request includes:

### DIFF
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -144,9 +144,16 @@ func (b *Blueprint) CryptPasswords() error {
 	// Any passwords for users?
 	for i := range b.Customizations.User {
 		// Missing or empty password
-		if b.Customizations.User[i].Password == nil || len(*b.Customizations.User[i].Password) == 0 {
+		if b.Customizations.User[i].Password == nil {
 			continue
 		}
+
+		// Prevent empty password from being hashed
+		if len(*b.Customizations.User[i].Password) == 0 {
+			b.Customizations.User[i].Password = nil
+			continue
+		}
+
 		if !crypt.PasswordIsCrypted(*b.Customizations.User[i].Password) {
 			pw, err := crypt.CryptSHA512(*b.Customizations.User[i].Password)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


internal/blueprint/blueprint.go
```diff
@@ -144,9 +144,16 @@ func (b *Blueprint) CryptPasswords() error {
 	// Any passwords for users?
 	for i := range b.Customizations.User {
 		// Missing or empty password
-		if b.Customizations.User[i].Password == nil || len(*b.Customizations.User[i].Password) == 0 {
+		if b.Customizations.User[i].Password == nil {
 			continue
 		}
+
+		// Prevent empty password from being hashed
```
teg, line 151: Are empty passwords special in some way? Maybe add a line to explain why we don't hash them?

internal/blueprint/blueprint.go
```diff
@@ -74,6 +76,12 @@ func (b *Blueprint) Initialize() error {
 	if err != nil {
 		return fmt.Errorf("Invalid 'version', must use Semantic Versioning: %s", err.Error())
 	}
+
+	err = b.CryptPasswords()
```
teg, line 80: This works because `Initialise()` is always called after serialisation into the struct?

internal/blueprint/blueprint.go
teg: Ah, sorry, I read the commit without the commit message. Do I understand correctly that an empty/unset password should mean the account is locked, but without this logic it would mean that it can be logged into with the empty password?

bcl: Right, in my experiment I couldn't log in from the console or ssh, but as another normal user I could run 'su - user' to switch to the account with the hashed empty password.

internal/blueprint/blueprint.go
bcl: Correct.

teg (approved):
Thanks for doing this so quickly @bcl!

PR #2836, opened 2022-07-15
composer: let expiry be passed through APIs

supakeen:
The Koji API always sets the expiry to 1 day, for the Weldr and Cloud APIs the value can be provided in the request. If the value is not passed then the default is used: 14 days.

I'm not entirely too sure on using '0' as the placeholder for 'no expiry', I don't like in-value things like this but I'm unsure if it'd be sane to pass around *int64.

This PR still needs tests and (likely) some documentation but I'd like people to tell me if I'm doing crazy stuff as this is my-first-Go.

This pull request includes:


teg:
What's the behavior of the weldr api if the backing job is gone? Is this a behavioral change? If so: do we want it? Is it necessary?

I wonder if the default for weldr should be to not expire, but I could be wrong?

cc @ondrejbudai @bcl

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -235,7 +235,19 @@ func (q *DBJobQueue) Enqueue(jobType string, args interface{}, dependencies []uu
...
+	var deltaDays int64
+
+	if expiry != 0 {
+		deltaDays = expiry
+	} else {
+		// A '0' being passed means 'no expiry' (aka: very very long time)
+		deltaDays = 9999
+	}
+
+	expires := time.Now().UTC().Add(time.Hour * time.Duration(24*deltaDays))
```
ondrejbudai, line 248: I think that the `time.Duration` cast doesn't need to be there. Correct me if I'm wrong, I find it slightly confusing.

ondrejbudai:
> What's the behavior of the weldr api if the backing job is gone? Is this a behavioral change? If so: do we want it? Is it necessary?

I would prefer no expiration for weldr jobs, seems too smart for a local service (imagine buildah expiring images after 14 days).

PR #2837, opened 2022-07-16
github: fix the coverity scan action

ondrejbudai:
It was failing on:

vendor/github.com/proglottis/gpgme/data.go:4:11: fatal error: gpgme.h: No such file or directory

Let's install this package before running the check, I verified that this
fix works locally.

@atodorov This check was failing for 16 days so I presume that nobody checks it. Is it even useful if nobody checks its result?

This pull request includes:


gicmo (approved):
Thanks a lot, this is partly on me because I overlooked that. I wonder if we should have a file were we keep track of the dependencies and use that everywhere instead of having a lot of difference places.

PR #2840, opened 2022-07-17
dnf-json: explicitly set priority

teg:
Prioritise the repositories in the order they are passed. Currently their priorities are all
the same, which gives (to me) unexpected results.

The logic is now that appending a repo to the list of repositories, means that it overrides
the previously passed ones. This is relevant in case the same packages exist in the
base repositories as in the payload repositories. In this case we should prefer the
payload ones.

This pull request includes:

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -52,12 +52,14 @@ class Solver():
         self.base.conf.substitutions['arch'] = arch
         self.base.conf.substitutions['basearch'] = dnf.rpm.basearch(arch)
 
-        for repo in repos:
-            self.base.repos.add(self._dnfrepo(repo, self.base.conf))
+        # The order repositories are passed indicate their priority, with
+        # each repository overriding the previous ones in the list.
+        for priority, repo in enumerate(list.reverse(repos), start=1):
+            self.base.repos.add(self._dnfrepo(repo, self.base.conf, priority))
         self.base.fill_sack(load_system_repo=False)
 
     @staticmethod
-    def _dnfrepo(desc, parent_conf=None):
+    def _dnfrepo(desc, parent_conf=None, priority=99):
         """Makes a dnf.repo.Repo out of a JSON repository description"""
 
         repo = dnf.repo.Repo(desc["id"], parent_conf)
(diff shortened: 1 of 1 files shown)


teg:
I ran into this when trying to use image builder to build our worker images from the RPMs in COPR. I ended up with RPMs from AppStream, which is not what I wanted.

I think it stands to reason that for the packages in our base package set (the first depsolve we do in the chain) we get the packages from the base repos. However, in the subsequent depsolve, when we depsolve the packages given in the blueprint we should prefer the packages in the payload repository, which is what this PR attempts to do (untested...).

teg:
The alternative approach is to set the priority by "group of repositories", and then keep it the same for the whole group, there could be that BaseOS and AppStream having the same priority is intentional and important (for instance). The important part is that payload repos override base repos (and if ever we get "payload override repos" they should fit in-between), how the repos are prioritised internally in the group might be less important, and maybe we should not make it explicit.

achilleas-k:
Good way to test this would be with the `dummy` package repo we create in a few tests: we can create two of them and control which one gets selected based on order.

teg:
I agree to both comments. I might not get the chance to finish this up, so please consider it a glorified RFE, and feel free to take it over :)

thozza:
I'm not 100% sure how will DNF behave if we would try to depsolve 2nd+ transaction with payload repos, which would contain newer versions of packages in the base image package set. IIRC, we do not want to allow user to replace packages from the base package set with packages in repos provided by them. If this is still the case, we should cover this case with a test case before using priorities. Otherwise I think that it makes sense to put higher priority on user-provided repositories.

Unless users need to have different priority set for the repositories provided by them, it makes sense to start with setting the priority per "group of repositories". Also it IMO does not make much sense for the "base" RHEL repositories to not have the same priority.

I also like Ondrej's suggestion to not make `dnf-json` smarter, and instead make composer to pass the priorities to it.

teg:
Regarding replacing base packages. I think we have the following scenarios (and I think @ondrejbudai might be looking into clarifying this):
 - user wants to add software to a RHEL system, which does not exist in RHEL (the current payload support)
 - user wants to add software to a RHEL system, which exists in RHEL, but is not installed in the base system (what I'm trying to do, which I think could be made to work by setting the priority right for the payload repositories, though needs verification - it really should work as if you added high-prio repos and did a manual dnf install on top of your base system)
 - user wants to replace packages from the base system with their own (say from COPR if you are doing rhel/centos/fedora development), this would be needed by anaconda, and is not currently supported (would need some sort of `BaseReposOverride`), this should work as if the given repositories were used (at a higher priority) to create the base system.

PR #2842, opened 2022-07-19
worker/osbuild: use `os-release` to determine host OS

achilleas-k:
When running an osbuild job, we read `/etc/redhat-release` to get the host OS name to attach as metadata to the job result.

Only Fedora and RHEL ship this file, which makes the osbuild job always fail on other distributions.

The main reason to report host OS back to the worker server is due to Koji composes and the koji-finalize job, which pushes it to Koji. The motivation is to have enough information to potentially re-instantiate / identify the original builder host OS. There are no specific requirements on the string.

Modify the code to use `/etc/os-release` to determine the host OS. Fall back to using `linux` as the host OS, in case reading `os-release` fails, log the error and continue with the job. The `linux` fallback is suggested by the `os-release` spec [1]

[1] https://www.freedesktop.org/software/systemd/man/os-release.html#ID=

Had issues running on Arch and wasn't sure what was happening. :)

Note that with the current behaviour (on `main`), the job doesn't fail immediately because after the error is caught we `return nil`.  The job stops and then times out (essentially failing) but the error message is lost (or hard to find).

## Miscellaneous changes

-  Drop special handling for RHEL < 8.4 in `GetHostDistroName()` and always include the minor version in the host distro string.
- Simplify trimming of quotes in `readOSRelease()`

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -246,10 +246,10 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		Arch:         common.CurrentArch(),
 	}
 
-	hostOS, err := common.GetRedHatRelease()
+	hostOS, _, _, err := common.GetHostDistroName()
 	if err != nil {
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, err.Error(), nil)
-		return nil
+		logWithId.Warnf("Failed to get host distro name: %v", err)
+		hostOS = "linux"
 	}
 	osbuildJobResult.HostOS = hostOS
 
--- a/internal/common/distro.go
+++ b/internal/common/distro.go
@@ -3,9 +3,7 @@ package common
 import (
 	"bufio"
 	"errors"
-	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"strings"
 )
@@ -29,20 +27,6 @@ func GetHostDistroName() (string, bool, bool, error) {
 	// TODO: We should probably index these things by the full CPE
 	beta := strings.Contains(osrelease["CPE_NAME"], "beta")
 	return name, beta, isStream, nil
-}
-
-// GetRedHatRelease returns the content of /etc/redhat-release
-// without the trailing new-line.
-func GetRedHatRelease() (string, error) {
-	raw, err := ioutil.ReadFile("/etc/redhat-release")
-	if err != nil {
-		return "", fmt.Errorf("cannot read /etc/redhat-release: %v", err)
-	}
-
-	//Remove the trailing new-line.
-	redHatRelease := strings.TrimSpace(string(raw))
-
-	return redHatRelease, nil
 }
 
 func readOSRelease(r io.Reader) (map[string]string, error) {
--- a/internal/common/distro.go
+++ b/internal/common/distro.go
@@ -60,14 +60,8 @@ func readOSRelease(r io.Reader) (map[string]string, error) {
 		}
 
 		key := strings.TrimSpace(parts[0])
-		value := strings.TrimSpace(parts[1])
-		if value[0] == '"' {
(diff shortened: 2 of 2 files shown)


teg:
Should we be reading `/etc/os-release`?

thozza:
> Note that with the current behaviour (on `main`), the job doesn't fail immediately because after the error is caught we `return nil`. The job stops and then times out (essentially failing) but the error message is lost (or hard to find).

Yeah, good catch. The code should have been moved under the block that defers the function which updates the worker server...

achilleas-k:
> Should we be reading `/etc/os-release`?

We could as a fallback.  Maybe use the `PRETTY_NAME`?

thozza:
> Should we be reading `/etc/os-release`?

We probably could assemble a very similar string using `/etc/os-release`. This information was added originally because of Koji composes reporting it back to Koji.

thozza:
> > This information was added originally because of Koji composes reporting it back to Koji.
> 
> Does Koji have any requirements for what this should look like, or is it just human-readable metadata?

I'll have to summon @gicmo and @ondrejbudai because I don't know. The information is reported back in the `koji-finalize` job in the buildroot metadata. If there are no specific requirements, I would be all for using `os-release` all the time and if reading it fails, then fail the job. If ` unknown` would be acceptable for Koji, we don't have to fail...

ondrejbudai:
IIRC we were told that they use `redhat-release` in their stuff so we used the same thing. We should investigate koji's behaviour if we decide to change that to (more sane) `os-release`.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -211,8 +211,8 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	hostOS, err := common.GetRedHatRelease()
 	if err != nil {
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, err.Error())
-		return nil
+		logrus.Warn(err.Error())
+		hostOS = "(unknown)"
```
thozza, line 215: nitpick: I don't think that it is necessary to put the "unknown" into brackets, but not harmful either.

thozza (commented):
I reached out to Tomas Kopecek asking about using `/etc/os-release`. Until then, I think it makes sense to merge this PR.

The main reason IMO is that currently, the fail will happen before we defer the function, which ensures that a status is reported back to worker server. Which is quite bad behavior...

It just needs rebasing.

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: Just trying to make sure there's less of a chance it gets interpreted as saying we're running on a distro called _Unknown_ :laughing:

thozza:
So the answer I got from Tomas is that `basically any info which is sufficient to reinstatiate/identify original OS. There is no codified standard as mixing it with non-linux OS is already a mess.`

Answer to:
So using e.g. the `ID="rhel"` combined with `VERSION_ID="9.1"` as `rhel-9.1` or `rhel-91` could be good enough?

was **yes**

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Fair point, one can never know 🙄

teg: Is a fallback defined in the os-release spec? I seem to remember it being `linux`.

thozza: `If not set, a default of "ID=linux" may be used.` from https://www.freedesktop.org/software/systemd/man/os-release.html#ID=

thozza:
I agreed with @achilleas-k to pick this PR. I won't review or approve, since I'm the (co)author and I'm obviously very pleased by these changes 😅

achilleas-k (commented):
Not reviewing a PR I started but chiming in to say I approve the changes @thozza made.
Should get a 3rd party to approve before merging.

PR #2845, opened 2022-07-20
metrics: add `arch` label to prometheus metrics

kingsleyzissou:
This pull request includes:
- updating the prometheus metrics to accept the `arch` label

### DIFF
--- a/internal/cloudapi/v2/handler.go
+++ b/internal/cloudapi/v2/handler.go
@@ -576,7 +576,7 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 
 	if jobType == worker.JobTypeOSBuild {
 		var result worker.OSBuildJobResult
-		status, _, err := h.server.workers.OSBuildJobStatus(jobId, &result)
+		jobInfo, err := h.server.workers.OSBuildJobStatus(jobId, &result)
 		if err != nil {
 			return HTTPError(ErrorMalformedOSBuildJobResult)
 		}
@@ -607,36 +607,36 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 				Id:   jobId.String(),
 				Kind: "ComposeStatus",
 			},
-			Status: composeStatusFromOSBuildJobStatus(status, &result),
+			Status: composeStatusFromOSBuildJobStatus(jobInfo.JobStatus, &result),
 			ImageStatus: ImageStatus{
-				Status:       imageStatusFromOSBuildJobStatus(status, &result),
+				Status:       imageStatusFromOSBuildJobStatus(jobInfo.JobStatus, &result),
 				Error:        composeStatusErrorFromJobError(jobError),
 				UploadStatus: us,
 			},
 		})
 	} else if jobType == worker.JobTypeKojiFinalize {
 		var result worker.KojiFinalizeJobResult
-		finalizeStatus, deps, err := h.server.workers.KojiFinalizeJobStatus(jobId, &result)
+		finalizeInfo, err := h.server.workers.KojiFinalizeJobStatus(jobId, &result)
 		if err != nil {
 			return HTTPError(ErrorMalformedOSBuildJobResult)
 		}
-		if len(deps) < 2 {
+		if len(finalizeInfo.Deps) < 2 {
 			return HTTPError(ErrorUnexpectedNumberOfImageBuilds)
 		}
 		var initResult worker.KojiInitJobResult
-		_, _, err = h.server.workers.KojiInitJobStatus(deps[0], &initResult)
+		_, err = h.server.workers.KojiInitJobStatus(finalizeInfo.Deps[0], &initResult)
 		if err != nil {
 			return HTTPError(ErrorMalformedOSBuildJobResult)
 		}
 		var buildJobResults []worker.OSBuildJobResult
 		var buildJobStatuses []ImageStatus
-		for i := 1; i < len(deps); i++ {
+		for i := 1; i < len(finalizeInfo.Deps); i++ {
 			var buildJobResult worker.OSBuildJobResult
-			buildJobStatus, _, err := h.server.workers.OSBuildJobStatus(deps[i], &buildJobResult)
+			buildInfo, err := h.server.workers.OSBuildJobStatus(finalizeInfo.Deps[i], &buildJobResult)
 			if err != nil {
 				return HTTPError(ErrorMalformedOSBuildJobResult)
 			}
-			buildJobError, err := h.server.workers.JobDependencyChainErrors(deps[i])
+			buildJobError, err := h.server.workers.JobDependencyChainErrors(finalizeInfo.Deps[i])
 			if err != nil {
 				return HTTPError(ErrorGettingBuildDependencyStatus)
 			}
@@ -661,7 +661,7 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 
 			buildJobResults = append(buildJobResults, buildJobResult)
 			buildJobStatuses = append(buildJobStatuses, ImageStatus{
-				Status:       imageStatusFromKojiJobStatus(buildJobStatus, &initResult, &buildJobResult),
+				Status:       imageStatusFromKojiJobStatus(buildInfo.JobStatus, &initResult, &buildJobResult),
 				Error:        composeStatusErrorFromJobError(buildJobError),
 				UploadStatus: us,
 			})
@@ -672,7 +672,7 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 				Id:   jobId.String(),
 				Kind: "ComposeStatus",
 			},
-			Status:        composeStatusFromKojiJobStatus(finalizeStatus, &initResult, buildJobResults, &result),
+			Status:        composeStatusFromKojiJobStatus(finalizeInfo.JobStatus, &initResult, buildJobResults, &result),
 			ImageStatus:   buildJobStatuses[0], // backwards compatibility
 			ImageStatuses: &buildJobStatuses,
 			KojiStatus:    &KojiStatus{},
@@ -816,7 +816,7 @@ func (h *apiHandlers) getComposeMetadataImpl(ctx echo.Context, id string) error 
 	}
 
 	var result worker.OSBuildJobResult
-	status, _, err := h.server.workers.OSBuildJobStatus(jobId, &result)
+	buildInfo, err := h.server.workers.OSBuildJobStatus(jobId, &result)
 	if err != nil {
 		return HTTPErrorWithInternal(ErrorComposeNotFound, err)
 	}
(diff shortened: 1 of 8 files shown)


internal/worker/api/openapi.yml
```diff
@@ -295,8 +295,11 @@ components:
     UpdateJobRequest:
       type: object
       required:
+        - arch
         - result
       properties:
+        arch:
+          type: string
```
Gundersanne, line 302: I think we should avoid changing the worker api if at all possible, this would be a breaking change. Can't we parse the arch from the jobtype? We actually already have globs in the prometheus queries I think to account for the `:$arch` at the end of osbuild jobs.

kingsleyzissou: Yeah didn't seem like a great option. The only thing is that only the `osbuild` jobtype has the arch in it, the others don't.
We could either pass in an empty string for the other types (also not great). 

Or maybe we can store the `arch` somewhere else, any suggestions on where though?

teg: makes sense to me to only track the arch for arch specific jobs (build)

kingsleyzissou: Doh so I didn't realise that prometheus would hide a field if you pass it an empty string.
Updating the PR, it's actually a very simple fix in the end.

kingsleyzissou:
I've changed the jobType from `osbuild:$arch` to just `osbuild`. So I will sort out the alerts and dashboard tomorrow.

internal/worker/json.go
```diff
@@ -40,6 +40,9 @@ func (j OSBuildJob) OsbuildExports() []string {
 }
 
 type JobResult struct {
+	// build job results have the `arch` field
+	// use this for the job metrics arch
+	Arch     string              `json:"arch,omitempty"`
```
ondrejbudai, line 45: Where is this set? Am I blind? :sweat_smile:

kingsleyzissou: @thozza added it here:
https://github.com/osbuild/osbuild-composer/blob/a5a16113355ea167376ddfb020ab50fac7e3bf8d/internal/worker/json.go#L55

So only the build jobs will have the `arch` field the rest won't

kingsleyzissou: I could just use an if/else statement and check the `jobType` instead of this though if that's more readable

ondrejbudai: Aren't these fields stored separately, though? See https://go.dev/play/p/cz0tGv4iCar

thozza: I vote for checking the job type and casting its results explicitly. I find explicit better than implicit and this code is a bit magic and works correctly only in very specific situations.

kingsleyzissou: Yeah makes sense, it was being unmarshalled here:
https://github.com/osbuild/osbuild-composer/blob/a5a16113355ea167376ddfb020ab50fac7e3bf8d/internal/worker/server.go#L603

But I agree that it's a bit magic and not very explicit. Busy refactoring it now.

internal/worker/server.go
```diff
@@ -562,6 +560,27 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
...
+
+func getBuildJobArch(jobType string, jobResult json.RawMessage) string {
+	if jobType != JobTypeOSBuild {
+		return ""
+	}
+	var osbuildResult OSBuildJobResult
+	err := json.Unmarshal(jobResult, &osbuildResult)
+	if err != nil {
+		logrus.Errorf("error unmarshalling build result: %v", err)
+		// don't block on this, it's only needed form metrics
```
thozza, line 578: typo: `form` -> `for`

internal/worker/server.go
```diff
@@ -546,7 +542,9 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
 		logrus.Errorf("error finding job status: %v", err)
 	} else {
 		statusCode := clienterrors.GetStatusCode(jobResult.JobError)
```
thozza, line 544: I just realized that by using `jobStatus()` in `FinishJob()` we are potentially missing various errors set in `jobResult.JobError` by the job-type-specific status methods, such as `OSBuildJobStatus`.

What would probably make sense is to have a `switch` block here, based on the job type determined using `JobType()` (which will eliminate the need for `getMetricsJobType()`). And subsequently call the appropriate job-type-specific status method (which will eliminate the need for `getBuildJobArch`) to get the job status and result.

It will be one more call to the DB compared to the current state, but that may be acceptable.

internal/worker/server.go
```diff
@@ -562,6 +560,27 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
 	return nil
 }
 
+func getMetricsJobType(jobType string) string {
+	if strings.HasPrefix(jobType, JobTypeOSBuild+":") {
+		return JobTypeOSBuild
+	}
+	return jobType
+}
```
thozza, line 568: From what I checked, we should be able to change the `Server.jobStatus()` to strip the osbuild job type in the same way as it is done by `Server.JobType()`. Especially since you won't need it for Prometheus any more.

IMO it will also make the code more sane and consistent, because the returned job type will be always the same, regardless of the used method.

The only place where this won't help is the method that handles enqueueing of jobs, where you'll still need to strip the architecture from job type. But there you could use `strings.Split(jobType, ":")[0]` instead as used by `JobType()`.

thozza (commented):
Thanks for the changes!

I added some comments that I would like to discuss before I approve. Nevertheless all looks functional as is from my PoV.

internal/worker/server.go
kingsleyzissou: Oh yeah this is a good point, I'm going to have to rework each of the `JobStatus` functions to return the channel since we need that, but that's not a big deal.

internal/cloudapi/v2/handler.go
```diff
@@ -659,9 +659,8 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 
 			buildJobResults = append(buildJobResults, buildJobResult)
 			buildJobStatuses = append(buildJobStatuses, ImageStatus{
-				Status:       imageStatusFromKojiJobStatus(buildJobStatus, &initResult, &buildJobResult),
-				Error:        composeStatusErrorFromJobError(buildJobError),
-				UploadStatus: us,
```
kingsleyzissou, line 664: somehow deleted this. Busy fixing

internal/cloudapi/v2/handler.go
```diff
@@ -607,36 +607,36 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
...
 		})
 	} else if jobType == worker.JobTypeKojiFinalize {
 		var result worker.KojiFinalizeJobResult
-		finalizeStatus, deps, err := h.server.workers.KojiFinalizeJobStatus(jobId, &result)
+		finalizeInfo, err := h.server.workers.KojiFinalizeJobStatus(jobId, &result)
 		if err != nil {
 			return HTTPError(ErrorMalformedOSBuildJobResult)
 		}
-		if len(deps) < 2 {
+		if len(finalizeInfo.Deps[0]) < 2 {
```
kingsleyzissou, line 623: Also fixing

teg (approved):

internal/worker/server.go
```diff
@@ -551,16 +546,63 @@ func (s *Server) FinishJob(token uuid.UUID, result json.RawMessage) error {
...
 	if err != nil {
-		logrus.Errorf("error finding job status: %v", err)
-	} else {
-		statusCode := clienterrors.GetStatusCode(jobResult.JobError)
-		arch := getBuildJobArch(jobInfo.JobType, result)
-		prometheus.FinishJobMetrics(jobInfo.JobStatus.Started, jobInfo.JobStatus.Finished, jobInfo.JobStatus.Canceled, jobInfo.JobType, jobInfo.Channel, arch, statusCode)
+		return err
+	}
+
+	var arch = ""
```
thozza, line 554: nitpick: you should have defined the variable by specifying type `var arch string`, instead of assigning a value, similar to the variable on the line below this one. The variable would be initialized to the default value for `string`, which is an empty string `""`.

thozza (dismissed):
Thanks for the changes, it all looks good 👍 

I added one nitpick. Feel free to ignore it.

FWIW, I find the naming of the `JobInfo` structure a bit unfortunate, but alternatives would require changing a lot of lines. Don't take this personally, it is a general comment. The thing that I find unfortunate is that functions named `xxxJobStatus()` are returning `*JobInfo`, which contains `JobStatus` inside itself. The issue is that you could not use `JobStatus`, because the structure already exists and renaming all `xxxJobStatus()` functions to something like `xxxJobInfo()` would result in a gazillion of changed lines.

kingsleyzissou:
> FWIW, I find the naming of the JobInfo structure a bit unfortunate, but alternatives would require changing a lot of lines. Don't take this personally, it is a general comment. The thing that I find unfortunate is that functions named xxxJobStatus() are returning *JobInfo, which contains JobStatus inside itself. The issue is that you could not use JobStatus, because the structure already exists and renaming all xxxJobStatus() functions to something like xxxJobInfo() would result in a gazillion of changed lines.
Thanks for this feedback @thozza - I 100% agree with this and it felt a bit dirty to return `JobInfo` from the `JobStatus` function and the `JobInfo` contains the actual `JobStatus` struct. But like you say, it is unfortunate and would require a very large amount of changes - that's why I initially decided against it. But open to changing that if everyone feels strongly about it

internal/worker/server.go
kingsleyzissou: Looks like there is something wrong with the RPMS so if I have to rebase the PR then I will fix this nitpick

thozza:
> Thanks for this feedback @thozza - I 100% agree with this and it felt a bit dirty to return `JobInfo` from the `JobStatus` function and the `JobInfo` contains the actual `JobStatus` struct. But like you say, it is unfortunate and would require a very large amount of changes - that's why I initially decided against it. But open to changing that if everyone feels strongly about it

I don't feel strongly about it. I'm OK to keep it as is and I would be also OK to rename all the `xxxJobStatus()` functions.

internal/worker/server.go
```diff
@@ -161,51 +168,51 @@ func (s *Server) JobDependencyChainErrors(id uuid.UUID) (*clienterrors.Error, er
...
+		jobInfo, err = s.KojiFinalizeJobInfo(id, &kojiFinalizeJR)
 		if err != nil {
 			return nil, err
 		}
 		jobResult = &kojiFinalizeJR.JobResult
 
 	case JobTypeContainerResolve:
 		var containerResolveJR ContainerResolveJobResult
-		_, jobDeps, err = s.ContainerResolveJobStatus(id, &containerResolveJR)
+		jobInfo, err = s.ContainerResolveJobStatus(id, &containerResolveJR)
```
thozza, line 215: You missed `ContainerResolveJobStatus` -> `ContainerResolveJobInfo`

thozza (commented):
All looks great, but unfortunately you missed one function rename 😕

thozza (approved):
👍

PR #2849, opened 2022-07-21
Container embedding: support accessing protected resources

gicmo:
The osbuild worker gained a new configuration section `containers` with a new key `auth_file_path` to be able to specify a `containers-auth.conf(5)` file to be used for request. All container resolution is moved to a new worker job type so that there is only one central place that does container resolution. The containers auth file, if specified, is also forwarded to `osbuild` via the `REGISTRY_AUTH_FILE` file that is then used by `skopeo` in the `org.osbuild.skopeo` source.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-container-resolve.go
+++ b/cmd/osbuild-worker/jobimpl-container-resolve.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/osbuild/osbuild-composer/internal/container"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+type ContainerResolveJobImpl struct {
+	AuthFilePath string
+}
+
+func (impl *ContainerResolveJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id())
+	var args worker.ContainerResolveJob
+	err := job.Args(&args)
+	if err != nil {
+		return err
+	}
+
+	result := worker.ContainerResolveJobResult{
+		Specs: make([]worker.ContainerSpec, len(args.Specs)),
+	}
+
+	logWithId.Infof("Resolving containers (%d)", len(args.Specs))
+
+	resolver := container.NewResolver(args.Arch)
+
+	for _, s := range args.Specs {
+		resolver.Add(s.Source, s.Name, s.TLSVerify)
+	}
+
+	specs, err := resolver.Finish()
+
+	if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/osbuild-worker/main.go
```diff
@@ -430,13 +435,17 @@ func main() {
...
 		worker.JobTypeKojiInit: &KojiInitJobImpl{
 			KojiServers: kojiServers,
 		},
 		worker.JobTypeKojiFinalize: &KojiFinalizeJobImpl{
 			KojiServers: kojiServers,
 		},
+		worker.JobTypeContainerResolve: &ContainerResolveJobImpl{
+			AuthFilePath: containersAuthFilePath,
+		},
 	}
```
teg, line 449: For future reference: I think we should not have this be depsolve v non-depsolve jobs, but rather build v non-build jobs. I.e., put all the short-running ones together as their latency will be much shorter.

teg (dismissed):
Looks great! And considering all the complexities we discussed: very simple!

achilleas-k (dismissed):

cmd/osbuild-worker/jobimpl-container-resolve.go
```diff
@@ -0,0 +1,60 @@
...
+	resolver.AuthFilePath = impl.AuthFilePath
+
+	for _, s := range args.Specs {
+		resolver.Add(s.Source, s.Name, s.TLSVerify)
+	}
+
+	specs, err := resolver.Finish()
+
+	if err != nil {
+		result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorDNFOtherError, err.Error())
```
gicmo, line 41: Wrong error type ;)

internal/weldr/api.go
```diff
@@ -2165,13 +2165,66 @@ func (api *API) depsolveBlueprintForImageType(bp blueprint.Blueprint, options di
...
+
+	for {
+		status, _, err := api.workers.ContainerResolveJobStatus(jobId, &result)
+
+		if err != nil {
+			return specs, err
+		}
+
+		if result.JobError != nil {
+			return specs, fmt.Errorf("Failed to resolve containers: %s", result.JobError.Reason)
```
gicmo, line 2205: This will lead to a double "Failed to resolve containers":

`ERROR: ContainerResolveError: Failed to resolve containers: failed to resolve container: 'localhost:8080/osbuild/demo@sha256:23b808ef648cee4db38ca39026f1a077507898916e741ceab39563c2dc033bc8': error getting manifest: reading manifest sha256:23b808ef648cee4db38ca39026f1a077507898916e741ceab39563c2dc033bc8 in localhost:8080/osbuild/demo: unauthorized: access to the requested resource is not authorized`

cmd/osbuild-worker/main.go
gicmo: Makes sense

achilleas-k (approved):
Good to go!

PR #2853, opened 2022-07-23
Use JWT for Koji tests

thozza:
### Changes to `tools/provision.sh`
The script used to provision SUT always in the same way for both, the Service scenario and the on-premise scenario. While this was not causing any issues, it didn't not realistically represent how we expect osbuild-composer and worker to be used in these scenarios.

The script was modified to take an optional argument representing the desired authentication method to be used with composer and workers. The use of authentication methods is implicitly tied to a specific use scenario (Service vs. on-premise) The script currently supports the following authentication options:

- `none`
  - Intended for the on-premise scenario with Weldr API.
  - NO certificates are generated.
  - NO osbuild-composer configuration file is created.
  - NO osbuild-worker configuration file is created. This means that no cloud provider credentials are configured directly in the worker.
  - Only the local worker is started and used.
  - Only the Weldr API socker is started.
  - Appropriate repository definitions are copied to `/etc/osbuild-composer/repositories/`.
- `jwt`
  - Intended for the Service scenario with Cloud API.
  - Should be the only method supported in the Service scenario in the future.
  - Certificates are generated and copied to `/etc/osbuild-composer`.
  - osbuild-composer configuration file is created and configured for JWT authentication.
  - osbuild-worker configuration file is created, configured for JWT authentication and with appropriate cloud provider credentials.
  - Local worker unit is masked. Only the remote worker is used (the socket is started and one remote-worker instance is created).
  - Only the Cloud API socket is started (Weldr API socket is stopped).
  - NO repository definitions are copied to `/etc/osbuild-composer/repositories/`.
- `tls`
  - **This is the default if no argument is provided to the script.**
  - Intended for the Service scenario with Cloud API.
  - Should eventually go away.
  - Certificates are generated and copied to `/etc/osbuild-composer`.
  - osbuild-composer configuration file is created and configured for TLS client cert authentication.
  - osbuild-worker configuration file is created, configured for TLS authentication and with appropriate cloud provider credentials.
  - Local worker unit is masked. Only the remote worker is used (the socket is started and one remote-worker instance is created).
  - Only the Cloud API socket is started (Weldr API socket is stopped).
  - NO repository definitions are copied to
    `/etc/osbuild-composer/repositories/`.

### Other notable changes

- Some common code related to testing composes with JWT has been moved from `test/cases/multi-tenancy.sh` to `test/cases/api/common/common.sh`.
- `koji.sh` has been modified to provision SUT with `jwt` and Koji compose scenarios are now tested only using JWT. The change included rework of `tools/koji-compose.py`.
- `multi-tenancy.sh` has been modified to provision SUT with `jwt`.
- All other test cases except for `api.sh` and `regression-old-worker-new-composer.sh` have been modified to provision SUT with `none`.
- `tools/koji-compose.py` is now tested using pylint in GitHub Action.
- `tools/provision.sh` does not do any basic verification of SUT. This was just duplicating testing done by specific test cases and it was using Weldr API, which is no longer enabled in all scenarios.

This pull request includes:


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ee00ae6efc15f8c96b60dd00ff88c2e2fbfb5ef2 with the main merge-base 5f3f477d4722073f51544cd32bf49d3027132a94).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

thozza:
> ⚠️ This PR introduces changes in at least one manifest

Actually this PR does not touch any image definitions 🤔

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 719bf7f1ea36aff30bfe512e2c211e8bf8d3756c with the main merge-base 0f5846326c8b51437b8c8610a0b2bc44e8a6e7a0).  Please review the changes.  The changes can be found in the job artifacts of the `Manifest-diff` job as `manifests.diff`

ondrejbudai:
> > warning This PR introduces changes in at least one manifest
> 
> Actually this PR does not touch any image definitions thinking

@achilleas-k Interesting... :thinking:

achilleas-k:
> > > warning This PR introduces changes in at least one manifest
> > 
> > 
> > Actually this PR does not touch any image definitions thinking
> 
> @achilleas-k Interesting... thinking

Very strange.  I see [three runs for this PR](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines?page=1&scope=all) and the first two report changes, but don't list any (nor do the jobs have any artifacts).
The third one is clear though.

test/cases/koji.sh
```diff
@@ -139,14 +153,17 @@ elif [[ "$DISTRO_CODE" == rhel-9* ]]; then
     DISTRO_CODE=rhel-91
 fi
 
+ORG_ID=42
+set_worker_org "${ORG_ID}"
```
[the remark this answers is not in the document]
thozza: Let me think about this more. I was trying to just reuse as most of the existing pieces as possible, but I can rework this more... My thinking was that it should be OK for a test case to define the used ORG once and then use it throughout the test case. I think that it is OK to hide that from all the test cases, but the multi-tenancy one...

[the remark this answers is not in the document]
thozza: I reworked the code a bit. I modified the original functions to not take any `ORG_ID` as an argument and instead just read it from `/etc/osbuild-worker/token`. The original implementation of the relevant functions is now available under `<orig_name>_with_org_id` name and they are used only in the `multi-tenancy.sh`. WDYT?

test/cases/koji.sh
```diff
@@ -122,9 +135,12 @@ sudo cp \
     /etc/pki/ca-trust/source/anchors/osbuild-composer-tests-ca-crt.pem
 sudo update-ca-trust
 
-greenprint "Restarting composer to pick up new config"
+start_jwt_auth_server
+
+greenprint "Restarting composer, stopping a local worker and starting a remote worker"
 sudo systemctl restart osbuild-composer
```
ondrejbudai, line 141: If I read this correctly, this script doesn't modify composer's configuration, can we move this line to a file that actually changes the configuration?

test/cases/koji.sh
```diff
@@ -122,9 +135,12 @@ sudo cp \
...
-sudo systemctl restart osbuild-worker\@1
+sudo systemctl stop osbuild-worker@1
```
ondrejbudai, line 142: Local worker doesn't work at all with JWT, right? So maybe let's call this in `provision.sh` when `jwt` is requested?

ondrejbudai (changes requested):
Nice work! Is this PR ready btw? It's marked as `[WIP]` but not set as a draft. I have some comments inline to (maybe) improve the role of `provision.sh` in our test suite, thus marking as "request changes" for now.

test/cases/koji.sh
```diff
@@ -122,9 +135,12 @@ sudo cp \
     /etc/pki/ca-trust/source/anchors/osbuild-composer-tests-ca-crt.pem
 sudo update-ca-trust
 
-greenprint "Restarting composer to pick up new config"
+start_jwt_auth_server
```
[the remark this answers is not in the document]
thozza: I thought about this and had it so initially, but then we would have to source it, instead of executing it. We need to keep the PIDs somewhere, so that we can kill them later.

test/cases/koji.sh
thozza: It does not, but my reasoning was that it needs to pick up the token from the mock server, once they are started.

test/cases/koji.sh
thozza: Yes, that sounds reasonable. The provision.sh does weird thing TBH, it starts the remote worker `.socket` unit, but it does not mask the local worker. So in reality, they both may be running  as part of some test cases.

thozza:
> Nice work! Is this PR ready btw? It's marked as `[WIP]` but not set as a draft.

It is still WIP, because for some reason, the `koji-compose.py` fails to get the compose status, after it submitted it successfully 🤷 . I need to reproduce the issue locally and debug some more... I didn't mark the PR as Draft, because I was not able to get the Gitlab CI running when it was marked as Draft. 😕 Even with the proper label, nothing was happening. This is a result of the flakiness that we are recently seeing WRT triggering Gitlab CI for PRs...

thozza:
@ondrejbudai I think that all your previous comments should be addressed. Please review once you have some time.

thozza:
There was one more hiccup with the `test/cases/regression-old-worker-new-composer.sh`, which I had to fix.

ondrejbudai (approved):
I love this! One naming nitpick: I think that in the future, `provision.sh` should take a target environment as an argument rather than an auth method.

Basically: `oauth` => `service`, `none` => `on-prem`

I think it would be better to express the big idea in the argument rather than its implementation.

thozza:
> I love this! One naming nitpick: I think that in the future, `provision.sh` should take a target environment as an argument rather than an auth method.
> 
> Basically: `oauth` => `service`, `none` => `on-prem`
> 
> I think it would be better to express the big idea in the argument rather than its implementation.

I fully agree. I would say that this should happen as part of getting rid of TLS authentication. Right now, we have two auth methods for the service scenario, which made it weird to use it as the script argument...

In addition, it would be great if the DB was configured and started as part of provisioning for the "service" scenario.

PR #2855, opened 2022-07-25
Remove image info from all test manifests

achilleas-k:
We never generated it for the newest distro versions and we don't test
it on any of the old ones.

It's better this way.

I used `cmd/gen-manifests` to generate these.  The order of elements in the compose request for some manifests has changed.  The order is now fixed, based on the struct defined (and serialised) in the `gen-manifests` tool, rather than the order they appear in the `format-request-map.json`, which `generate-test-cases` retains.


teg:
Are these now tested elsewhere?

gicmo:
> Are these now tested elsewhere?

It tested via the diff-manifest check that will post the warning message on PRs that change manifests, like this: https://github.com/osbuild/osbuild-composer/pull/2812#pullrequestreview-1033356663

ondrejbudai:
Note that this actually doesn't fix the failing tests in rawhide.

achilleas-k:
> > Note that this actually doesn't fix the failing tests in rawhide.
> 
> We need the remaining two commits from [main...achilleas-k:osbuild-composer:sort-fstab](https://github.com/osbuild/osbuild-composer/compare/main...achilleas-k:osbuild-composer:sort-fstab)

I didn't want to mix the two in the same PR in case a discussion starts about a better way to fix the sorting issue.

Feel free to add those commits here if you think it's alright or make a new PR out of them while I'm away.

thozza (approved):

gicmo (approved):
👍

PR #2858, opened 2022-07-26
Cloud API - support uploading to container registries

ygalblum:
Worker
------
Add configuration for the default container registry
Use the default container registry if not provided as part of the image name
When using the default registry use the configured values
Return the image url as part of the result

Composer Worker API
-------------------
Add ContainerTargetResultOptions to return the image url

Composer API
------------
Add UploadOptions to allow setting of the image name and tag
Add UploadStatus to return the url of the uploaded image

This pull request includes:


cmd/osbuild-worker/config.go
```diff
@@ -52,6 +52,10 @@ type authenticationConfig struct {
 
 type containersConfig struct {
 	AuthFilePath string `toml:"auth_file_path"`
+	Domain       string `toml:"domain"`
+	Account      string `toml:"account"`
```
gicmo, line 56: Maybe we calls this `Path` to have the same nomenclature as the containers library? Maybe also a `Default` prefix to indicate that this is indeed only used if not otherwise specified in the `Source` argument.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -199,6 +207,40 @@ func uploadToS3(a *awscloud.AWS, outputDirectory, exportPath, bucket, key, filen
...
+	if i == -1 || (!strings.ContainsAny(destination[:i], ".:") && destination[:i] != "localhost") {
+		if impl.ContainersConfig.Domain != "" {
+			base := impl.ContainersConfig.Domain
+			if impl.ContainersConfig.Account != "" {
+				base = fmt.Sprintf("%s/%s", base, impl.ContainersConfig.Account)
+			}
+			destination = fmt.Sprintf("%s/%s", base, destination)
+			useImpl = true
+		}
+	}
```
gicmo, line 222: Extract this as a helper into the `container` package.

gicmo (commented):
Overall looks good to me! Like the approach. I wonder, we might want to do the same for the container embedding, i.e. use `Domain` and `Path`/`Account` if it is not given, so we are consistent? If so, we might as well extend the `Client` constructor to take the `Domain`/`Path` fallbacks? But that could also be done in a follow up.

Would be nice to add a test to `api.sh` for this as well.

cmd/osbuild-worker/config.go
ygalblum: Can you please clarify what you think should be changed. Just the `Account` to `Path`, or consolidate the two into `Path`?
I took the name `Domain` from here: https://github.com/containers/image/blob/main/docker/reference/normalize.go#L93. This function also uses `officialRepoName`, but I though `Account` is better that `RepoName` (since I am never sure how to call this part of the url).
As for the `Default` prefix, sure, I'll make that change once we clarify the above

ygalblum:
> Would be nice to add a test to api.sh for this as well.
Working on it.

test/cases/api/container.registry.sh
```diff
@@ -0,0 +1,106 @@
...
+
+  GET_METADATA_CURL_REQUEST="curl --silent \
+      --show-error \
+      --cacert /etc/osbuild-composer/ca-crt.pem \
+      --key /etc/osbuild-composer/client-key.pem \
+      --cert /etc/osbuild-composer/client-crt.pem \
+      https://localhost/api/image-builder-composer/v2/composes/${COMPOSE_ID}/metadata"
+
+  BUILD_OSTREE_COMMIT=$(${GET_METADATA_CURL_REQUEST} | jq '.ostree_commit')
+  SERVICED_OSTREE_COMMIT=$(curl http://localhost:8080//repo/refs/head/rhel/8/x86_64/edge)
```
gicmo, line 103: `OSTREE_REF=test/rhel/8/edge` vs `rhel/8/x86_64/edge`

test/cases/api/container.registry.sh
```diff
@@ -0,0 +1,106 @@
...
+
+function checkEnv() {
+    printenv CI_REGISTRY_USER > /dev/null
+    printenv CI_JOB_TOKEN > /dev/null
+    printenv CI_REGISTRY > /dev/null
+    printenv CI_PROJECT_PATH > /dev/null
+}
+
+# Global var for ostree ref
+export OSTREE_REF="test/rhel/8/edge"
```
gicmo, line 13: Get the distro version from the env?

Gundersanne:
Supplying the registry url in the request and then having the credentials in the worker doesn't make sense to me, having configuration split across the user input and the worker is something to avoid. Either:

- the worker has it's own organization and can create registries at will
- the worker configuration has access to 1 single registry but pushes images which aren't related to one another to it (not really how registries are supposed to be used)
- the request contains both the url and the credentials to push to that specific registry

Furthermore I think the workers should be able to push images with an expire time (quay has support for this), unless you can set that on an entire registry when creating it and we go with option 2 (unsure if that's possible).

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -199,6 +207,33 @@ func uploadToS3(a *awscloud.AWS, outputDirectory, exportPath, bucket, key, filen
...
+	if err != nil {
+		return nil, err
+	}
+
+	if impl.ContainersConfig.AuthFilePath != "" {
+		client.SetAuthFilePath(impl.ContainersConfig.AuthFilePath)
+	}
+
+	if appliedDefaults {
+
```
Gundersanne, line 222: stray newline

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -566,7 +568,22 @@ components:
         build_id:
           type: integer
           example: 42
-
+    ContainerUploadStatus:
+      type: object
+      additionalProperties: false
+      required:
+        - url
```
Gundersanne, line 575: The url should not be required if there's a default option.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -566,7 +568,22 @@ components:
...
+        - digest
```
Gundersanne, line 576: Shouldn't this be generated by composer and returned?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -824,6 +842,23 @@ components:
...
+        name:
+          type: string
+          example: 'osbuild'
+          description: |
+            Name for the created container image
+        tag:
+          type: string
+          example: 'latest'
+          description: |
+            Tag for the created container image
```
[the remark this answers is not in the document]
ygalblum: The name of the image is part what one will refer to as the registry name: <Service URL>/<Base Path>/<Image Name> and it's the tags that change.
So, while I see your point about the tag being generated by the composer, shouldn't we let the user set the <Image Name>?

ygalblum:
> Furthermore I think the workers should be able to push images with an expire time (quay has support for this), unless you can set that on an entire registry when creating it and we go with option 2 (unsure if that's possible).

Adding expiration tags to an image is a quay specific feature. Some repositories don't support any automatic expiration (e.g. GCP) while others support cleanup policies (e.g. AWS ECR supports Lifecycle policies). Since there is no standard way, I don't think a generic solution can be applied here

cmd/osbuild-worker/jobimpl-osbuild.go
ygalblum: This was actually intentional (I like to let my code breath). But, np, I'll fix it. Just don't want to push such a minor change to kickoff all the CI for it. I'll do it as part of any other change that we might do

internal/cloudapi/v2/openapi.v2.yml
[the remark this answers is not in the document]
Gundersanne: Anything that can generate conflicts due to user input at least have the ability to be generated by/chosen by composer or be configured in the workers imo. I know some of the other upload options allow you to set the image name (but not as a required property), but in practice these options aren't used.

I can imagine something like 1 image name per distro or 1 image name per image if chosen by composer. Or just 1 single image name configured in the workers.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -40,16 +40,24 @@ type S3Configuration struct {
...
-	AzureCreds        *azure.Credentials
-	AWSCreds          string
-	AWSBucket         string
-	S3Config          S3Configuration
+type ContainersConfiguration struct {
 	ContainerAuthFile string
+	Domain            string
+	Account           string
+	CertPath          string
+	TlsVerify         *bool
```
ondrejbudai: `TLSVerify` is more idiomatic, see e.g. https://pkg.go.dev/net/http#Server

internal/cloudapi/v2/handler.go
```diff
@@ -365,6 +363,22 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 				t.ImageName = key
 				t.OsbuildArtifact.ExportFilename = imageType.Filename()
 
+				irTarget = t
+			case ImageTypesEdgeContainer:
```
ondrejbudai, line 367: @Gundersanne Are we fine with breaking the backward compatibility here?

ondrejbudai (commented):
Looks good!

internal/cloudapi/v2/handler.go
ygalblum: I was also concerned about this. But, we had a discussion about it a few weeks ago and I don't remember who said that edge-containers are not really used currently, so breaking it should be OK. But, yes, this change makes old Workers not be able to build edge-container configured by a newer composer service

cmd/osbuild-worker/jobimpl-osbuild.go
ygalblum: Done

internal/cloudapi/v2/openapi.v2.yml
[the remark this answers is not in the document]
ygalblum: OK. I've changes it so that the name and tag are optional. Tag is used only if the name was provided.

cmd/osbuild-worker/jobimpl-osbuild.go
ygalblum: Done

internal/cloudapi/v2/handler.go
Gundersanne: It's true that the image type isn't used atm. This does break the image type, as configuration won't be in place when this PR lands, and I do not want to block releasing on an unused image type.

internal/cloudapi/v2/openapi.v2.yml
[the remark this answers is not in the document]
Gundersanne: Having talked to @gicmo about this, it's ok if it's required, we can always drop the requirement later as that's not a breaking change. We should pull the container image type from image-builder in this case however.

Sorry for the flip-flop.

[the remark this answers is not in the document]
ygalblum: I've already made it optional. As you wrote, shouldn't make much of a difference at this point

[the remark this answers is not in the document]
Gundersanne: https://github.com/osbuild/image-builder/pull/396 opened

internal/cloudapi/v2/handler.go
teg: Approved PR, I agree, this is fine as-is.

teg (approved):

PR #2854, opened 2022-07-25
Fix UEFI HTTP boot for RHEL 9 ISOs

achilleas-k:
The RHEL 9 and CS9 ISOs would fail to boot over UEFI HTTP.  Added the necessary dracut and kernel modules to support this.

The second commit adds all kernel modules that are in the official RHEL 9 ISO's initramfs and were not on ours.

See rhbz#2030730

Manifests regenerated for each commit.  Image info dropped for all updated manifests that had any.

PR #2860, opened 2022-07-27
Fix Go 1.19 issues (stable `fstab` ordering and wrong `errors.As` usages)

gicmo:
Two fixes for Go 1.19

##  osbuild: use path as secondary sort key for fstab 

Most filesystems entries in `fstab` don't have a `PassNo`, which makes the order of those entries dependent on the sorting algorithm.  Changes in the algorithm can introduce changes in the sort order, which we don't like. This has happened with `go 1.19`, breaking our copr builds.

Add a secondary sorting key, the Path, which is guaranteed unique, to guarantee stable ordering.

## dbjobqueue: fix bad errors.As usages

`errors.As` is meant to check whether err (or other error in its chain) can be assigned to the value that target is pointing at.

Let's consider this example:

```
errors.As(err, &pgx.ErrNoRows)
```

`pgx.ErrNoRows` (and `pgx.ErrTxClosed`) is typed as `error`, thus in all `errors.As` calls, the target is typed as `*error`. `err` is always an `error`. So this call is basically asking whether `error` can be assigned to `error`. If `err != nil`, this is always `true`, thus this check doesn't make any sense over a plain `err != nil`.

Go 1.19 now checks this issue and if it's found, it refuses to compile the code, see:

https://go-review.googlesource.com/c/tools/+/339889

This commit changes usages of `errors.As()` to `errors.Is()`. The `Is()` method doesn't check assignability but equality (the only different between `Is()` and a plain old `==` operator is that `Is()` also inspects the whole error chain).

This fixes the check because now, we are basically checking if `err` (or any other error in its chain) `== pgx.ErrTxClosed` which is exactly what we want.

### DIFF
--- a/internal/osbuild/fstab_stage.go
+++ b/internal/osbuild/fstab_stage.go
@@ -1,6 +1,7 @@
 package osbuild
 
 import (
+	"fmt"
 	"sort"
 
 	"github.com/osbuild/osbuild-composer/internal/disk"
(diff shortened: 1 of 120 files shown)


thozza (dismissed):
👍

thozza (approved):
Fabulous 😍

achilleas-k:
The manifest diff error handling is a bit **_wrong_**.
The last commit in https://github.com/osbuild/osbuild-composer/pull/2854 fixes it.

gicmo:
This is required for rawhide build fixes and the failed manifest diff currently broken anyway. I am merging this.

PR #2862, opened 2022-07-27
Exclude dracut-config-rescue in ec2 and qemu-guest-agent in ec2 and gce images

jrusz:
This pull request includes:

### DIFF
--- a/internal/distro/rhel8/package_sets.go
+++ b/internal/distro/rhel8/package_sets.go
@@ -538,7 +538,6 @@ func ec2CommonPackageSet(t *imageType) rpmmd.PackageSet {
 			"langpacks-en",
 			"NetworkManager",
 			"NetworkManager-cloud-setup",
-			"qemu-guest-agent",
 			"redhat-release",
 			"redhat-release-eula",
 			"rsync",
@@ -573,6 +572,8 @@ func ec2CommonPackageSet(t *imageType) rpmmd.PackageSet {
 			"libertas-sd8787-firmware",
 			"libertas-usb8388-firmware",
 			"plymouth",
+			// RHBZ#2075815
+			"qemu-guest-agent",
 		},
 	}.Append(bootPackageSet(t)).Append(distroSpecificPackageSet(t))
 }
@@ -681,8 +682,6 @@ func gceCommonPackageSet(t *imageType) rpmmd.PackageSet {
 			// for time synchronization
 			"chrony",
 			"timedatex",
-			// Detected Platform requirements by Anaconda
-			"qemu-guest-agent",
 			// EFI
 			"grub2-tools-efi",
 		},
@@ -722,6 +721,8 @@ func gceCommonPackageSet(t *imageType) rpmmd.PackageSet {
 			"rt73usb-firmware",
 			"xorg-x11-drv-ati-firmware",
 			"zd1211-firmware",
+			// RHBZ#2075815
+			"qemu-guest-agent",
 		},
 	}.Append(bootPackageSet(t)).Append(distroSpecificPackageSet(t))
 }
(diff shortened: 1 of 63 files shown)


jrusz:
@thozza @gicmo Hi, could you please check if this is any good ? This is supposed to cover two RHBZs which I've mentioned in the commit messages. My initial intention was to cover them with tests but since this would be best covered by image-tests and it seemed "easy" to remove the packages I went ahead and did that. No idea if that's enough though or if the way I did it is acceptable.

thozza (commented):
Changes to the definitions look good to me.

Since we dropped image-info from image tests in composer, I'm not sure how useful is to run them in the upstream. Changes in manifests are already covered by other tests.

jrusz:
> Changes to the definitions look good to me.
> 
> Since we dropped image-info from image tests in composer, I'm not sure how useful is to run them in the upstream. Changes in manifests are already covered by other tests.

Thanks!. I though image tests would verify that the packages are really removed, I only ran them here to check that, I don't intend to enable them. Do we have other tests that would help verify this? Are `diff-manifests.sh` enough for this?

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD bf812ea6480e71dbac02a0531b00dfe9cf448245 with the main merge-base e3de1d03c05250775b007cb802fd02b60b500325).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2787778521/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2787778521/artifacts/browse

thozza:
> Thanks!. I though image tests would verify that the packages are really removed, I only ran them here to check that, I don't intend to enable them, I expect it's enough to have them run on nightlies. Do we have other tests that would help verify this? Are `diff-manifests.sh` enough for this?

So `diff-manifests.sh` will catch cases when a manifest is changed as part of the PR. We are still checking the generated manifests against those in the repository as part of unit tests.

atodorov:
> > Thanks!. I though image tests would verify that the packages are really removed, I only ran them here to check that, I don't intend to enable them, I expect it's enough to have them run on nightlies. Do we have other tests that would help verify this? Are `diff-manifests.sh` enough for this?
> 
> So `diff-manifests.sh` will catch cases when a manifest is changed as part of the PR. We are still checking the generated manifests against those in the repository as part of unit tests.

@thozza , @jrusz a few things here: 

1) RHBZ needs devel_ack+ b/c from what I can tell this PR also contains the fix for the bug
2) I am not following about the unit tests here. Which test is checking the generated manifests against the ones in the repository? IOW which test is making sure that packages which have been excluded don't actually end-up on the image ?

thozza:
> > > > Thanks!. I though image tests would verify that the packages are really removed, I only ran them here to check that, I don't intend to enable them, I expect it's enough to have them run on nightlies. Do we have other tests that would help verify this? Are `diff-manifests.sh` enough for this?
> > 
> > 
> > So `diff-manifests.sh` will catch cases when a manifest is changed as part of the PR. We are still checking the generated manifests against those in the repository as part of unit tests.
> 
> @thozza , @jrusz a few things here:
> 
> 1. RHBZ needs devel_ack+ b/c from what I can tell this PR also contains the fix for the bug

Done.

> 2. I am not following about the unit tests here. Which test is checking the generated manifests against the ones in the repository? IOW which test is making sure that packages which have been excluded don't actually end-up on the image ?

Unit tests, specifically https://github.com/osbuild/osbuild-composer/blob/fa18206918bb3cbd434fb22816e17b23ee4e5d82/internal/distro/distro_test.go#L16-L27, which are run as part of any RPM package build. If the test fails, then the RPM build fails as well.

The test checks that the manifest didn't change. No package change in the resulting image can happen without a change in the respective manifest (due to nature of how osbuild works and how we build images).

In addition, the same thing, but with depsolving the package set again, is tested as part of https://github.com/osbuild/osbuild-composer/blob/fa18206918bb3cbd434fb22816e17b23ee4e5d82/.gitlab-ci.yml#L189-L202

thozza (approved):
LGTM, Thanks!

PR #2870, opened 2022-08-01
test: Remove BIOS installation test because edge-installer supports UEFI only

henrywang:
Due to Edge device security reason, `edge-simplified-installer` and `edge-installer` support UEFI only.


jrusz (approved):
LGTM

PR #2875, opened 2022-08-02
tests: Workaround for mkksiso options coming from newer lorax RPM

atodorov:
This pull request includes:

### DIFF
--- a/test/cases/ostree-ng.sh
+++ b/test/cases/ostree-ng.sh
@@ -195,7 +195,7 @@ ostree remote add --no-gpg-verify --no-sign-verify ${OSTREE_OSNAME} ${PROD_REPO_
 EOFKS
 
     echo "Writing new ISO"
-    if nvrGreaterOrEqual "lorax" "34.9.18"; then
+    if [ "${ID}" != "fedora" && nvrGreaterOrEqual "lorax" "34.9.18" ]; then
         sudo mkksiso -c "console=ttyS0,115200" --ks "${newksfile}" "${iso}" "${newiso}"
     else
         sudo mkksiso -c "console=ttyS0,115200" "${newksfile}" "${iso}" "${newiso}"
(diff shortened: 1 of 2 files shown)


jrusz (approved):
I've ran a debug nightly pipeline and the test passed there https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2801598013 so LGTM

ondrejbudai (approved):

jrusz:
Ugh, there is a newer lorax in Fedora and it doesn't have the new `--ks` option... I guess the patch was only released to el9 branch? https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2801244119#L4095

I guess add something like `&& "$ID" == "rhel"` to the if condition and come back to if it causes problems in the future?

test/cases/ostree-ng.sh
```diff
@@ -194,7 +195,11 @@ ostree remote add --no-gpg-verify --no-sign-verify ${OSTREE_OSNAME} ${PROD_REPO_
 EOFKS
 
     echo "Writing new ISO"
-    sudo mkksiso -c "console=ttyS0,115200" "${newksfile}" "${iso}" "${newiso}"
+    if [ "${ID}" != "fedora" && nvrGreaterOrEqual "lorax" "34.9.18" ]; then
```
jrusz, line 198:
```suggestion
    if [ "${ID}" != "fedora" ] && nvrGreaterOrEqual "lorax" "34.9.18"; then
```
Shellcheck complains, maybe this will work?

teg:
@bcl is this still necessary? Will a fixed version of mkksiso land in RHEL9? (By fixed I mean backwards compatible).

bcl:
> @bcl is this still necessary? Will a fixed version of mkksiso land in RHEL9? (By fixed I mean backwards compatible).

Yes, it is in lorax-34.9.19, see https://bugzilla.redhat.com/show_bug.cgi?id=2037015

jrusz (approved):
LGTM. We don't know when the fix is going to land and we can always revert if this causes problems later.

PR #2876, opened 2022-08-02
clienterrors: Remove ellipsis operator

Gundersanne:
The alternative is keeping the ellipsis operator, but always only adding the 1st element as details to the job error.


kingsleyzissou (approved):

thozza (approved):
👍

PR #2879, opened 2022-08-03
worker: fix crash if no autoscale instance is defined

ygalblum:
The worker assumes that when running on AWS an autoscale is defined. If not defined, the worker crashes.

This crash occurred when triggering a build via the Weldr API while running a local installation of osbuild-composer and osbuild-worker on an EC2 instance 

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/main.go
+++ b/cmd/osbuild-worker/main.go
@@ -128,6 +128,10 @@ func setProtection(protected bool) {
 		}
 		return
 	}
+	if len(asInstanceOutput.AutoScalingInstances) == 0 {
+		logrus.Info("No Autoscaling instace is defined")
+		return
+	}
 
 	// make the request to protect (or unprotect) the instance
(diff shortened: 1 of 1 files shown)


cmd/osbuild-worker/main.go
```diff
@@ -128,6 +128,9 @@ func setProtection(protected bool) {
 		}
 		return
 	}
+	if len(asInstanceOutput.AutoScalingInstances) == 0 {
+		logrus.Info("No Autoscaling instace is defined ")
```
ondrejbudai, line 132:
```suggestion
		logrus.Info("No Autoscaling instace is defined")
```

More importantly, there's no `return`.

ygalblum: D'oh! Thanks for that 🤦🏼‍♂️. Fixed

ondrejbudai (approved):
Thanks!

lavocatt (approved):
Thanks for fixing this!

PR #2880, opened 2022-08-04
osbuild-mock-openid-provider: support `client_credentials` grant type

thozza:
Extend the implementation of mock openid server to take the `grant_type`
into consideration for the `/token` endpoint.

In addition to the previously supported `refresh_topen`, the
implementation now supports also `client_credentials`.

This is necessary to make it possible to use the mock server in
the `koji-osbuild` CI, because the builder plugin uses
`client_credentials` to get access token.

The implementation behaves in the following way:
 - For `refresh_token` grant type, it takes the `refresh_token` value
   from the request and adds it to the `rh-org-id` field in the custom
   claim, which is part of the returned token.
 - For `client_credentials` grant type, it takes the `client_secret`
   value from the request and adds it to the `rh-org-id` field in the
   custom claim, which is part of the returned token.

Requests without the supported `grant_type` set are rejected.

Modify affected test cases to specify `grant_type` when fetching a new
access token.

Related to https://github.com/osbuild/koji-osbuild/pull/104

This pull request includes:

### DIFF
--- a/cmd/osbuild-mock-openid-provider/main.go
+++ b/cmd/osbuild-mock-openid-provider/main.go
@@ -90,13 +90,29 @@ func main() {
 			panic(err)
 		}
 
-		cc := customClaims{
-			Type:      "Bearer",
-			ExpiresAt: 0,
-			IssuedAt:  time.Now().Unix(),
-			// Use refresh_token as rh-org-id
-			RHOrgID: r.Form.Get("refresh_token"),
+		var cc customClaims
+		switch r.Form.Get("grant_type") {
+		case "refresh_token":
+			cc = customClaims{
+				Type:      "Bearer",
+				ExpiresAt: 0,
+				IssuedAt:  time.Now().Unix(),
+				// Use refresh_token as rh-org-id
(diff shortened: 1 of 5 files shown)


ondrejbudai (approved):

ondrejbudai (dismissed):
Thanks, this is very nice and clean. :)

thozza:
OK, I forgot to modify `tools/koji-compose.py` 🤦  fixing now....

ondrejbudai (approved):
I double-checked and I think we should have covered all calls of the mock OAuth provider. Thanks!

ondrejbudai:
Looks like `api.sh` is broken

thozza:
> Looks like `api.sh` is broken

Yes, I just pushed a fix. The issue was that the script was waiting for the mock openid server to come up with the `curl --fail`. The mock openid server now returns `400` for requests which don't have any `grant_type` set. Which in turn resulted in non-zero exit code from curl (specifically `22`, which happens if `--fail` is used and the returned status code is `400` or above).

ondrejbudai (approved):
Thanks for the quick fix!

PR #2881, opened 2022-08-05
tests: Remove useless JSON file overrides

atodorov:
This pull request includes:

### DIFF
--- a/test/cases/regression-composer-works-behind-satellite-fallback.sh
+++ b/test/cases/regression-composer-works-behind-satellite-fallback.sh
@@ -61,24 +61,11 @@ case "${ID}" in
     "rhel")
         echo "Running on RHEL"
         case "${VERSION_ID%.*}" in
-            "8" )
+            "8" | "9")
                 echo "Running on RHEL ${VERSION_ID}"
                 # starting in 8.5 the override file contains minor version number as well
                 VERSION_SUFFIX=$(echo "${VERSION_ID}" | tr -d ".")
-                DISTRO_NAME="rhel-8"
-                REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
-                REPO1_NAME="baseos"
-                REPO2_NAME="appstream"
-                if [ -n "${NIGHTLY:-}" ]; then
-                    REPO1_NAME="baseos-${ARCH}"
-                    REPO2_NAME="appstream-${ARCH}"
-                fi
-                ;;
-            "9" )
-                echo "Running on RHEL ${VERSION_ID}"
-                # in 9.0 the override file contains minor version number as well
-                VERSION_SUFFIX=$(echo "${VERSION_ID}" | tr -d ".")
-                DISTRO_NAME="rhel-90"
+                DISTRO_NAME="rhel-${VERSION_SUFFIX}"
                 REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
                 REPO1_NAME="baseos"
                 REPO2_NAME="appstream"
--- a/test/cases/regression-composer-works-behind-satellite-fallback.sh
+++ b/test/cases/regression-composer-works-behind-satellite-fallback.sh
@@ -65,14 +65,8 @@ case "${ID}" in
                 echo "Running on RHEL ${VERSION_ID}"
                 # starting in 8.5 the override file contains minor version number as well
                 VERSION_SUFFIX=$(echo "${VERSION_ID}" | tr -d ".")
-                if grep beta /etc/os-release;
-                then
-                    DISTRO_NAME="rhel-8-beta"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}-beta.json"
-                else
-                    DISTRO_NAME="rhel-8"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
-                fi
+                DISTRO_NAME="rhel-8"
+                REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
                 REPO1_NAME="baseos"
                 REPO2_NAME="appstream"
                 if [ -n "${NIGHTLY:-}" ]; then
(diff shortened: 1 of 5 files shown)


tools/provision.sh
```diff
@@ -143,17 +143,17 @@ else # AUTH_METHOD_NONE
 
     # RHEL nightly repos need to be overridden
     case "${ID}-${VERSION_ID}" in
-        "rhel-8.6")
-            # Override old rhel-8.json and rhel-8-beta.json because RHEL 8.6 test needs nightly repos
-            sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-86.json "$REPODIR/rhel-8.json"
+        "rhel-8.7")
+            # Override old rhel-8.json and rhel-8-beta.json because RHEL 8.7 test needs nightly repos
+            sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-87.json "$REPODIR/rhel-8.json"
```
jrusz, line 148: Does this still need to be `rhel-8.json`? I think we want to use `rhel-87.json` now, but I'd have to check to be sure...

jrusz (changes requested):
Uh, does this mean we were testing with older sources until now? Thanks for fixing this, totally forgot this was happening and needed update, got just one comment.

jrusz:
> Uh, does this mean we were testing with older sources until now? Thanks for fixing this, totally forgot this was happening and needed update, got just one comment.

Hm, actually no, it seems that we were using the correct sources during nightly testing https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2811520975#L3105 so idk what this override does tbh...

tools/provision.sh
atodorov: I'm not sure. Probably not. @ondrejbudai can you help us here?

atodorov:
> Hm, actually no, it seems that we were using the correct sources during nightly testing https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2811520975#L3105 so idk what this override does tbh...

Could that be because the repos are overriden one more time a few lines below the current snippet ? So effectively for nightly you are not seeing the overrides coming from this `case` statement.

tools/provision.sh
ondrejbudai: I think it should be `rhel-87.json`. @thozza?

thozza: Yes, this should be `rhel-87.json`.

Unfortunately, we are still matching the full distro name as defined in the distro definition against the file name. So `rhel-8.json` would be ever used only for RHEL-8.3.

tools/provision.sh
```diff
@@ -143,17 +143,17 @@ else # AUTH_METHOD_NONE
...
-            sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-90.json "$REPODIR/rhel-90.json"
+        "rhel-9.1")
+            # Override old rhel-91.json and rhel-91-beta.json because RHEL 9.1 test needs nightly repos
+            sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-91.json "$REPODIR/rhel-91.json"
             # If multiple tests are run and call provision.sh the symlink will need to be overridden with -f
-            sudo ln -sf /etc/osbuild-composer/repositories/rhel-90.json "$REPODIR/rhel-90-beta.json"
+            sudo ln -sf /etc/osbuild-composer/repositories/rhel-91.json "$REPODIR/rhel-91-beta.json"
             ;;
         *) ;;
     esac
```
thozza, line 159: So I checked the script in more detail.

This whole block can be IMO removed, because it is useless. The same files are already copied over to the `$REPODIR` above this block, so copying them again does not add any value.

In addition, `-beta.json` repos are nor used in any case, even for Beta.

atodorov: @thozza what about https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression-composer-works-behind-satellite-fallback.sh#L68 ? 

If CI jobs don't fail I will remove those references to -beta.json as well.

thozza: @atodorov that should be the same case. So yes, IMO it can be removed as well. The current composer implementation will never pick the `-beta.json` repo definitions, even if running on Beta release.

In short:

1. When composer starts, it constructs the list of supported distros and determines on which distro it is running:
https://github.com/osbuild/osbuild-composer/blob/13e287733a07873f26b311fc9a86692b6bad7958/internal/distroregistry/distroregistry.go#L69

2. It then calls `New*HostDistro()` on the appropriate distro implementation, while passing in the information if the distro is Beta: https://github.com/osbuild/osbuild-composer/blob/13e287733a07873f26b311fc9a86692b6bad7958/internal/distroregistry/distroregistry.go#L75

3. However, all `New*HostDistro()` implementations are ignoring the passed values and always return the "base" distro: https://github.com/osbuild/osbuild-composer/blob/13e287733a07873f26b311fc9a86692b6bad7958/internal/distro/rhel8/distro.go#L741-L743

4. Weldr API then uses `RepoRegistry` to load all repo definitions and uses the filename as the distro name:https://github.com/osbuild/osbuild-composer/blob/13e287733a07873f26b311fc9a86692b6bad7958/internal/rpmmd/repository.go#L293

5. And the `RepoRegistry` implementation will always look for the repos using the value returned by distro implementation's `Name()` method, which will never contain the `-beta` suffix: https://github.com/osbuild/osbuild-composer/blob/13e287733a07873f26b311fc9a86692b6bad7958/internal/reporegistry/reporegistry.go#L44

test/cases/regression-composer-works-behind-satellite-fallback.sh
```diff
@@ -65,14 +65,8 @@ case "${ID}" in
...
                 VERSION_SUFFIX=$(echo "${VERSION_ID}" | tr -d ".")
-                if grep beta /etc/os-release;
-                then
-                    DISTRO_NAME="rhel-8-beta"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}-beta.json"
-                else
-                    DISTRO_NAME="rhel-8"
-                    REPOSITORY_OVERRIDE="/etc/osbuild-composer/repositories/rhel-${VERSION_SUFFIX}.json"
-                fi
+                DISTRO_NAME="rhel-8"
```
thozza, line 68: I'm not sure what is the intention here, but this will make the test case to use RHEL-8.3 repositories be copied as `$REPOSITORY_OVERRIDE`.

thozza (approved):
LGTM 👍  thanks for doing the cleanup

PR #2885, opened 2022-08-09
Disable skipped tests

jabia99:
This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -237,26 +237,46 @@ Manifests:
 
 regression-bigiso:
   extends: .regression
+  rules: 
+    # WHITELIST: Run only on x86_64 and rhel like systems
+    - if: $RUNNER =~ "/^.*(rhel-8.6.*x86_64|rhel-9.0.*x86_64|centos-stream-8.*x86_64|centos-stream-9.*x86_64).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-bigiso.sh
 
 regression-composer-works-behind-satellite-fallback:
   extends: .regression
+  rules: 
+    # BLACKLIST: Skipped on subscribed RHEL machines
+    - if: $RUNNER !~ "/^.*(rhel-.*-ga|centos).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-composer-works-behind-satellite-fallback.sh
 
 regression-composer-works-behind-satellite:
   extends: .regression
+  rules: 
+    # BLACKLIST: Skipped on subscribed RHEL machines
+    - if: $RUNNER !~ "/^.*(rhel-.*-ga|centos).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-composer-works-behind-satellite.sh
 
 regression-excluded-dependency:
   extends: .regression
+  rules: 
+    # WHITELIST
+    - if: $RUNNER =~ "/^.*(rhel-8.6|rhel-9.0|centos-stream-8|centos-stream-9).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-excluded-dependency.sh
 
 regression-include-excluded-packages:
   extends: .regression
+  rules: 
+    # BLACKLIST: Skipped on fedora systems
+    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-include-excluded-packages.sh
 
@@ -284,6 +304,10 @@ regression-insecure-repo:
 
 regression-no-explicit-rootfs-definition:
   extends: .regression
+  rules: 
+    # BLACKLIST: Skipped on fedora systems
+    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
   variables:
     SCRIPT: regression-no-explicit-rootfs-definition.sh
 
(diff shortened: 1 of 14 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD aca21814a03f56366dd62727f2c8b15def361006 with the main merge-base e9a1e66cbe8231b8ec84acab0d0f26539b868f33).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2915662474/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2915662474/artifacts/browse

jrusz (changes requested):
My original intention with this was to remove the conditions from the test cases and have one place where all conditions would be set which could probably only be done in `.gitlab-ci.yml` file.

I'm also missing some other tests which have these conditions set like `vmware.sh` , `regression-composer-works-behind-satellite.sh`, `regression-composer-works-behind-satellite-fallback.sh`

It would also be nice to add a comment on top of the rules to explain it a bit, some of them look like whitelists, some like blacklists.. I would like it to be clear so that others can easily spot what they're supposed to do if they want to add a runner to execute the test on or remove it...

jrusz (changes requested):
Also in the case of `regression-bigiso.sh` there is a condition that exits with 0 on `aarch64` runners which is also something we don't want.

I've also missed `regression-include-excluded-packages.sh` there is an `exit 0` for non rhel distros

jabia99:
> My original intention with this was to remove the conditions from the test cases and have one place where all conditions would be set which could probably only be done in `.gitlab-ci.yml` file.

I plan to remove the condition in the tests scripts, but I wanted a fast way to see if everything was properly skipped. I also couldn't think of an easy way of skipping the tests that didn't rely on adding new `rules`, with the inconvenience that this has.

test/cases/aws_s3.sh
```diff
@@ -5,8 +5,8 @@ source /usr/libexec/osbuild-composer-test/set-env-variables.sh
 if [ "${NIGHTLY:=false}" == "true" ]; then
     case "${ID}-${VERSION_ID}" in
         "rhel-8.7" | "rhel-9.1" )
-            echo "$0 is not enabled for ${ID}-${VERSION_ID} skipping..."
-            exit 0
+            echo "$0 should be skipped on ${ID}-${VERSION_ID} check gitlab-ci.yml"
+            exit 1
             ;;
         *)
             ;;
```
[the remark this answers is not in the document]
jabia99: In gitlab-ci.yml I already have the `nightly` check. I plan to remove this block of code

.gitlab-ci.yml
```diff
@@ -237,6 +237,9 @@ Manifests:
 
 regression-bigiso:
   extends: .regression
+  rules: 
+    - if: $RUNNER =~ "/^.*(rhel-8.6|rhel-9.0|centos-stream-8|centos-stream-9).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - !reference [.nightly_rules, rules]
```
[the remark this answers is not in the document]
jabia99: I'll add a comment like:
`WHITELIST`
or
`BLACKLIST`

On top of each rule. Thanks!!

jabia99:
> I'm also missing some other tests which have these conditions set like `vmware.sh` , `regression-composer-works-behind-satellite.sh`, `regression-composer-works-behind-satellite-fallback.sh`

for the regression tests, they are skipped if running on subscribed rhel. Do you think there's a way to check this from gitlab-ci.yml @jrusz ?

jrusz:
> for the regression tests, they are skipped if running on subscribed rhel. Do you think there's a way to check this from gitlab-ci.yml @jrusz ?

Yes, all GA runners are subscribed so any RHEL that has `-ga` in the name.

.gitlab-ci.yml
```diff
@@ -237,26 +237,46 @@ Manifests:
 
 regression-bigiso:
   extends: .regression
+  rules: 
+    # WHITELIST: Run only on x86_64 arch
```
jrusz, line 241: If it's only `x86_64` why also specify rhel a centos? I guess you mean `x86_64 and rhel like` systems?

jrusz (changes requested):
I looked over the tests again and in `regression-composer-works-behind-satellite-(fallback).sh` there is a case structure right after the subscription part you removed and there are `exit 0` in both of them in case the distro is unsupported. Could you please change it to `exit 1`? That way it will fail if it runs where it's not supposed to and will consistent with the changes you made already :) 

 Also could you please expand the commit message a bit? It would be nice if it had more info about the BLACKLISTS and WHITELISTS that you're introducing so that if someone stumbles upon it it's more clear what's the intention here.

And one minor nitpick in the ci file, otherwise looks good :)

.gitlab-ci.yml
```diff
@@ -624,7 +663,8 @@ OpenSCAP:
   stage: test
   extends: .terraform
   rules:
-    - !reference [.upstream_rules, rules]
+    # BLACKLIST
+    - if: $RUNNER =~ "/^.*(rhel-8.7|rhel-9.1).*$/" && $CI_PIPELINE_SOURCE != "schedule"
```
jrusz, line 667: Is this not a Whiteliest? Also since this job has separate RUNNER definition it can follow just upstream rules like before because it won't run on anything else that's not specified there.

jrusz (changes requested):
Sorry if I wasn't clear in the previous comment, but I'm still seeing `exit 0` in case it's running on unsupported distro in these places:
`regression-composer-works-behind-satellite.sh` on line 98 and 103
`regression-composer-works-behind-satellite-fallback.sh` on line 69
`regression-old-worker-new-composer.sh` line 22 (that one is new, sorry)

I've also just noticed that `koji.sh` is ran and skipped on nightly pipelines, could you just remove the nightly rules there on line 481 of the `.gitla-ci.yml` file and change the `exit 0` to `exit 1` on line 16 in `test/cases/koji.sh` please?

Also put one comment inline. Sorry for bothering you with these details but this should be the last changes I'm requesting :)

.gitlab-ci.yml
jabia99: > satellite-fallback.sh` on line 69 `regression-old-worker-new-composer.sh` line 22 (that one is new, sorry)
In this case the test also has a separate RUNNER definition, so additional rules are not needed.

jabia99:
>satellite-fallback.shon line 69regression-old-worker-new-composer.sh` line 22 (that one is new, sorry)

In this case the test also has a separate RUNNER definition, so additional rules are not needed.

jabia99:
> Sorry for bothering you with these details but this should be the last changes I'm requesting :)

It's alright! Thanks for noticing all this little things and taking the time to review ❤️

PR #2894, opened 2022-08-15
Modify repositories/rhel-xy.json file before testing nightly compose

atodorov:
- uses a wildcard match for sed in order to replace only lines which
  match and keep the rest intact

This pull request includes:


tools/provision.sh
```diff
@@ -174,3 +168,10 @@ case "${AUTH_METHOD}" in
         sudo systemctl start osbuild-composer.socket
         ;;
 esac
+
+# Print debugging info about content sources
+sudo composer-cli status show
```
ondrejbudai, line 173: This should be moved under the `AUTH_METHOD_NONE` section, tests meant for the service don't have Weldr enabled at all.

atodorov:
Nightly test job is here:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/613845213

failures are unrelated.

jrusz (changes requested):
> * the schutzbot/prepare-rhel-internal-json.sh script will be executed
>   before mockbuild.sh

How and why? Looking at the `.gitlab-ci.yml` file I can see it's running the `mockbuild.sh` first to create the test rpms and the it's running the `prepare-rhel-internal` stage...

Also this means that currently the `prepare-rhel-internal` is just making sure the correct system repos are set and optinally a custom `-tests` package repo url is used, right?

Otherwise it looks good I like the simple `sed` to replace the source repos during provisining of the SUT.

jrusz:
@atodorov the issue is that this https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression-composer-works-behind-satellite.sh#L127 doesn't return anything

Latest run on main: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2887431645#L3037
Run with this PR:  https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2890016059#L3183

And the reason seems to be this one:https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression-composer-works-behind-satellite.sh#L101 
It's expecting the repo name to be all lowercase in nightly run

atodorov:
> > * the schutzbot/prepare-rhel-internal-json.sh script will be executed
> >   before mockbuild.sh
> 
> How and why? Looking at the `.gitlab-ci.yml` file I can see it's running the `mockbuild.sh` first to create the test rpms and the it's running the `prepare-rhel-internal` stage...
> 

^^^ no longerr valid, that was an initial approach. Commit log message updated to match.


> Also this means that currently the `prepare-rhel-internal` is just making sure the correct system repos are set and optinally a custom `-tests` package repo url is used, right?
> 

Yes.

> Otherwise it looks good I like the simple `sed` to replace the source repos during provisining of the SUT.

atodorov:
> And the reason seems to be this one:https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression-composer-works-behind-satellite.sh#L101

Good catch, thanks. I didn't scroll all the way down apparently.  Fixed in the latest commit.

atodorov:
9.1 test job:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/615634410

8.7 test job:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/615634450

jrusz (approved):
Alright, nightly pipelines looks good, so LGTM

PR #2896, opened 2022-08-15
Add support for Fedora 37 and 38

ondrejbudai:
For the [Build Fedora IoT Artifacts with osbuild](https://fedoraproject.org/wiki/Changes/IoTArtifactsWithOSBuild) initiative, we need to support Fedora 37 and 38 (aka rawhide). This PR does that in two commits:

1) The first commit deduplicates the version logic under the Fedora distro slightly.
2) The second commit adds support for Fedora 37 and 38.

Neither of these commits enables integration testing for these versions, and I would like to hear your thoughts here: Do we want them? Of course, without enabling them, these Fedora versions are completely uncovered. On the other hand, I don't want us to become an operating system CI (aka OSCI :hand_over_mouth:) of unstable Fedora versions. What do you think?

This pull request includes:


bcl:
> What do you think?

I think it becomes more useful once a Fedora released has branched (as F37 just did). At the least I think having integration testing results available, but not required for PRs, can be helpful in catching problems early.

teg:
I would vote in favor of CI. I'm assuming:
- we use our snapshots
- the tests succeeded against some snapshot at some point (how else do we know the code is correct?)
- we only update snapshots if they don't break CI

So even for rawhide this should not add to our instability, but it will protect against regressions in our code.

Obviously if rawhide regresses things will still be broken if we don't do anything about the snapshot update failing, but that's better than us breaking ourselves.

gicmo:
> I would vote in favor of CI. [...]

I 100% agree with everything here; just would like to add: we should have a procedure in place that if the auto-update for snapshot fails we do investigate and see what is broken, i.e. the compose or something in Fedora changed and broke us. This would have presented the image installer debacle we had before 8.6.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 33058789ff269fd4457f55a1b85f56b50995b3cb with the main merge-base 53d4d7f2849a8f3064f5db8468990bfe3e681e76).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2896002845/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/2896002845/artifacts/browse

ondrejbudai:
Superseded by #2914

PR #2897, opened 2022-08-15
s3 upload: add an option to upload images publicly

ondrejbudai:
This PR adds support for uploading S3 artifacts via the cloud API publicly.

This is slightly weird because our existing S3 uploads are already kinda public, the difference is that:

The existing method:
- the URL is long and obfuscated
- the expiration is in 12 hours because of the limitation of the S3 API

The new public option:
- the URL is shorter and straightforward
- the expiration works in the same way as for other artifacts. For our external service, that's currently 14 days.

I'm not sure if the difference is big enough but 12 hours feels kinda short for our Packit usecase so I would be in favor of adding this public option.

This pull request includes:

- [x] adequate documentation informing people about the change (this change is documented in the OpenAPI spec itself)


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -749,6 +749,15 @@ components:
         region:
           type: string
           example: 'eu-west-1'
+        public:
+          type: boolean
+          default: false
+          description: |
+            By default, a long, obfuscated URL is returned. Its expiration
```
gicmo, line 756: I was a bit confused by this, maybe `If false, the default, a long ..."

Gundersanne: it would make it a little clearer. Otherwise lgtm

Gundersanne:
I think the expiration limit is 6 hours btw :/

internal/cloudapi/v2/openapi.v2.yml
ondrejbudai: I reworded the description slightly.

Gundersanne (dismissed):
lgtm, thank you!

ondrejbudai:
The test was broken, I think it should be working now.

Gundersanne (dismissed):
LGTM!

ondrejbudai:
force-pushed to resolve a conflict in the generated openAPI code

teg (dismissed):

ondrejbudai:
One small fix was needed in the AWS S3 test, so I did that and also rebased the PR.

thozza:
Rebased on top of `main`, to ensure that nothing broke due to #2988

internal/cloud/awscloud/awscloud.go
```diff
@@ -581,3 +581,18 @@ func (a *AWS) S3ObjectPresignedURL(bucket, objectKey string) (string, error) {
 	logrus.Info("[AWS] 🎉 S3 Presigned URL ready")
 	return url, nil
 }
+
+func (a *AWS) MarkS3ObjectAsPublic(bucket, objectKey string) error {
+	logrus.Infof("[AWS] 👐 Making S3 object public %s/%s", bucket, objectKey)
```
thozza, line 586: I know that other methods in this file log messages as well, but I find it weird in a "library" implementation. The caller of this method could to it id needed (log before calling it and log message based on the fact if an `error` is returned. This is probably not something worth changing at this point, but I wanted to call this out to see if I'm the only one 😄

test/cases/api/aws.s3.sh
```diff
@@ -101,6 +109,7 @@ function verify() {
             ;;
 
         "${IMAGE_TYPE_VSPHERE}")
+
```
thozza, line 112: added whitespace 😱

thozza (approved):
LGTM, nice work 👍

internal/cloud/awscloud/awscloud.go
ondrejbudai: Agreed, alternatively, we can pass logger inside the AWS object.

test/cases/api/aws.s3.sh
ondrejbudai: I'm sorry! :/

PR #2898, opened 2022-08-15
distro: add oscap packages to image

kingsleyzissou:
Since the oscap remediation stage in osbuild runs the oscap pacakge in `chroot`, it is not possible to have
the packages in the build root. The packages are needed in the image itself.

This pull request includes:

### DIFF
--- a/internal/distro/rhel8/distro.go
+++ b/internal/distro/rhel8/distro.go
@@ -450,13 +450,11 @@ func (t *imageType) PackageSets(bp blueprint.Blueprint, options distro.ImageOpti
 		mergedSets[buildPkgsKey] = mergedSets[buildPkgsKey].Append(extraPkgs)
 	}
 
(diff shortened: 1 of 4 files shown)


teg:
If that's the only way

kingsleyzissou:
@teg thinking about this a little more. I think the oscap package is necessary, but the user could provide the package for the datastream content - because they might want to use a different package to the scrap-security-guide

teg:
I don't think we should rely on the user to add arbitrary packages for things to work. So either keep the security guides default to included, or deduce it from an option in the ospac section of the blueprint (might be tricky / not worth it). I would say for now stick with keeping the package included by default, and if we want something more clever in the future, we could do something similar to how the kernel package works (if you don't specify it you get `kernel`, but if you specify it in the blueprint it replaces the default).

thozza (approved):
LGTM and the change makes sense.

PR #2908, opened 2022-08-23
Add search command to dnf-json and use it for package searches

bcl:
This pull request includes:


dnf-json
```diff
@@ -116,6 +116,52 @@ class Solver():
...
+
+        packages = []
+
+        # NOTE: Build query one piece at a time, don't pass all to filterm at the same
+        # time.
+        available = self.base.sack.query().available()
+        for name in pkg_globs:
+            # If the package name glob has * in it, use glob.
+            # If it has *name* use substr
+            # If it has neither use exact match
```
teg, line 137: What was the reason for this logic? Is it the same as dnf uses on the command line?

internal/weldr/api.go
```diff
@@ -1207,23 +1193,12 @@ func (api *API) modulesInfoHandler(writer http.ResponseWriter, request *http.Req
...
-		errors := responseError{
-			ID:  "ModulesError",
-			Msg: fmt.Sprintf("msg: %s", err.Error()),
-		}
-		statusResponseError(writer, http.StatusBadRequest, errors)
-		return
-	}
-
-	foundPackages, err := availablePackages.Search(names...)
+	foundPackages, err := api.fetchPackageList(distroName, names)
```
teg, line 1196: Does this change the behaviour in any way compared to before?

teg (commented):
This looks great!

Did you try checking the latency difference between the old and the new API? I'd be curious to know how long a search query took before and how long it takes now (with hot caches).

I imagine it being significantly faster, but verifying it and putting a ballpark figure in the commit message would be nice I think.

dnf-json
bcl: No, this matches the behavior of the API queries (which IIRC were first written to use dnf via lorax-composer). I split it between globbing all the time and glob/substr hoping that there would be some speed difference.

internal/weldr/api.go
bcl: No, the results are actually the same, just faster.

bcl:
On my VM (4G ram, 2 cores) and a hot metadata cache the searches drop from around 10s to 2-3s which makes a noticeable difference.

This does not speed up initial loading with a cold cache of course, and does not speed up listing all packages (I am looking into that next, to see if something can be done there).

teg (approved):
Love this! Would be great to include in the release, I think.

PR #2902, opened 2022-08-17
Update snapshots to 20220817

schutzbot:
Job(s) succeeded: 
Job(s) failed: 0
Workflow run: https://github.com/osbuild/rpmrepo/actions/runs/2874370114

PR #2909, opened 2022-08-23
Add the `rhsm.facts` stage.

supakeen:
Work has been ongoing to store RHSM facts into built images. This PR is part of the `osbuild-composer` side of things but has some problems to solve and/or discuss.

In this initial pull request only the `ApiType` used to request the compose is stored in the `rhsm.facts` file eventually created by `osbuild`.

This pull request includes:

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -2,7 +2,7 @@
   "fedora-35": {
     "dependencies": {
       "osbuild": {
-        "commit": "376cbffd136bc4ba86fc7c63697fa5b88fe3acef"
+        "commit": "ba218f781d4c7455cd995eea55be09e902370905"
       }
     },
     "repos": [
@@ -79,7 +79,7 @@
   "fedora-36": {
     "dependencies": {
       "osbuild": {
-        "commit": "376cbffd136bc4ba86fc7c63697fa5b88fe3acef"
+        "commit": "ba218f781d4c7455cd995eea55be09e902370905"
       }
     },
     "repos": [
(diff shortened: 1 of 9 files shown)


internal/cloudapi/v2/errors.go
```diff
@@ -44,6 +44,7 @@ const (
 	ErrorTenantNotFound               ServiceErrorCode = 28
 	ErrorNoGPGKey                     ServiceErrorCode = 29
 	ErrorValidationFailed             ServiceErrorCode = 30
+	ErrorInvalidFactsParams           ServiceErrorCode = 31
```
Gundersanne, line 47: The error can be dropped I think, no api changes etc...

supakeen: woops, leftover that's now removed. thank you!

thozza:
@supakeen you need to update the `Requires` in the SPEC file to osbuild v63, because there is currently v62, which does not include the `rhsm.facts` stage.
https://github.com/osbuild/osbuild-composer/blob/de6b8e8f5b783113a8d4cfd50fddc6c25c1568cd/osbuild-composer.spec#L308-L316

thozza (commented):
All looks great 👍 , just the SPEC file needs to be updated with a dependency on osbuild v63.

teg:
Apart from the comment from @thozza , this looks great!

supakeen:
Thank you @thozza, I've force pushed the dependency update.

thozza (dismissed):
👍

Gundersanne:
The osbuild commit in Schutzfile needs to be udpated to f3e2c8d5e9fcfdb8d73d2f111f7ed178c52310eb as well, otherwise it won't find v63

Gundersanne (approved):
thx!

thozza (approved):

thozza:
> The osbuild commit in the Schutzfile needs to be updated to f3e2c8d5e9fcfdb8d73d2f111f7ed178c52310eb as well, as that's where CI constructs osbuild's repo url from.

So the v63 was supposed to be already in the stable repos, but https://github.com/osbuild/osbuild-composer/pull/2902 was not merged yet so we came to the same conclusion with Simon... Ideally the override should not be needed at all.

Gundersanne:
> > The osbuild commit in the Schutzfile needs to be updated to f3e2c8d5e9fcfdb8d73d2f111f7ed178c52310eb as well, as that's where CI constructs osbuild's repo url from.
> 
> So the v63 was supposed to be already in the stable repos, but #2902 was not merged yet so we came to the same conclusion with Simon... Ideally the override should not be needed at all.

Ahh makes sense. Thanks for the info.

supakeen:
CI breaking on a depsolve job (invalid package set requested) and OSTree related things, force pushed the change for the second bit (moving the stage before prepping OSTree).

teg (approved):

PR #2910, opened 2022-08-24
CI: run libvirt tests on GCP runners

jrusz:
This pull request includes:


jrusz:
Ok this should finally be ready. I've decided to switch `libvirt.sh` test because that's ran on many different distros so we should get a better idea about how well it's going to work.

schutzbot/ci_details.sh
```diff
@@ -42,19 +42,22 @@ echo "List of installed packages:"
 rpm -qa | sort
 echo "------------------------------------------------------------------------------"
 
-if rpm --quiet -q python36; then
-    echo -e "\n FAIL: python36 is installed, see #794 ..."
-    exit 1
-else
-    echo -e "\n PASS: python36 not insalled"
+# gcp runners don't use cloud-init and some of the images have python36 installed
```
atodorov, line 45: - Which of the GCP images have python36 installed? 
- Do they have another installation of python, e.g. platform-python ? What version is it?

schutzbot/ci_details.sh
```diff
@@ -42,19 +42,22 @@ echo "List of installed packages:"
...
-
-# Ensure cloud-init has completely finished on the instance. This ensures that
-# the instance is fully ready to go.
-while true; do
-  if [[ -f /var/lib/cloud/instance/boot-finished ]]; then
-    break
-  fi
-  echo -e "\n🤔 Waiting for cloud-init to finish running..."
-  sleep 5
-done
```
atodorov, line 60: The commit log for dd0fc32b02c47402f284646d1a2fab355635eead says

>    We need to wait for cloud-init to be completely finished since it is
>    registering RHEL instances with RHN and preparing a swap file.

I'm not sure if this is an issue with GCP but worth investigating a bit more. 

- How do we register GCP systems to RHN if there is no cloud-init? 
- How do we know the VM has booted up completely and ready for testing ?

atodorov (commented):
Tests are passing. Added a few comments which need some more investigation.

schutzbot/ci_details.sh
jrusz: I believe it's all el-8. So centos-8 and any rhel-8... They have both `platform-python` and `python36` ... As an example you can see this job output where I've let it list all the packages: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3776825895#L1028 

there is the same runner on aws in that pipeline and it's just missing the `python36` package from what I can tell. They bot have `platform-python-3.6.8` though. I don't even know what the original issue with this was as the Github issue has link to Jenkins which no longer exists. It doesn't seem to cause problems from what I can tel...

schutzbot/ci_details.sh
jrusz: These images have some google guest packages which are described a bit here https://github.com/osbuild/osbuild-composer/blob/main/image-types/rhel8/google-gce.md#implementation-choices-osbuild-composer-gce-rhel-guest-image 

We register them with subscription-manager during the `prepare` step https://github.com/osbuild/gitlab-ci-terraform-executor/blob/main/prepare#L75 and it waits for the registration to finish I believe.

And how do we know VM is ready I'm not sure. Maybe terraform has a way to tell? It does wait until the provisioning is finished and then it waits for the ssh connection to be available, that sounds like enough to me... But I can try taking a look at how the google guest stuff works and if there is something we can check similar to `cloud-init`

jrusz: Since this comes from the age of Jenkins I feel like it might have been some issue with the way runners were provisioned then...

atodorov (approved):

PR #2913, opened 2022-08-24
gen-manifests: don't fail on invalid distro-arch-image combos

achilleas-k:
When the user specifies any of the distro, arch, or image type values to filter generation, invalid combinations would cause a panic, which made it hard to filter requests based just on an image type.
Instead of failing, print an error message to inform the user, but continue with the rest of the jobs.
This way, a user is informed that a certain combination is invalid if they make a mistake, but can also filter on a single image type and only get valid manifests out of the run.

This change enables running (for example):
```
go run ./cmd/gen-manifests -images edge-installer,installer
```
without worrying about distros or architectures that don't support those image types.

The error message is printed to avoid situations where typos create only invalid requests that are silently skipped.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -379,7 +379,8 @@ func main() {
 	for _, distroName := range distros {
 		distribution := distroReg.GetDistro(distroName)
 		if distribution == nil {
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):

ondrejbudai:
Just a development tool improvement, no need to wait for the CI.

PR #2917, opened 2022-08-25
test/gcp: Run cleanup function at the end

jrusz:
All EXIT traps are cleared on line 280 so the cleanup trap is never run
and VMs are waiting for 4 hours to get cleaned by
scheduled-cloud-cleaner. Run the cleanup at the end and rely on
scheduled-cloud-cleaner only in case of failures before that.

This pull request includes:

### DIFF
--- a/test/cases/gcp.sh
+++ b/test/cases/gcp.sh
@@ -278,6 +278,8 @@ get_compose_metadata "$COMPOSE_ID"
 # Kill the journal monitor immediately and remove the trap
 sudo pkill -P ${WORKER_JOURNAL_PID}
 trap - EXIT
+# trap cleanup again
+trap cleanup EXIT
 
 # Did the compose finish with success?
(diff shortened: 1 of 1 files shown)


test/cases/gcp.sh
```diff
@@ -287,6 +287,9 @@ fi
 
 verifyInGCP
 
+# Run cleanup function
+cleanup
```
ondrejbudai, line 291: If `verifyInGCP` fails, the `cleanup` still won't run, right?

jrusz: Yeah, then it's up to the scheduled cloud cleaner... Maybe I could trap the cleanup again after the line where the traps are removed instead?

ondrejbudai: Yep, that makes sense.

thozza (approved):
👍  thanks!

ondrejbudai (approved):

ondrejbudai:

Issue #2887, opened 2022-08-10
Add search and sort capabilities to dnf-json


bcl:
In https://github.com/osbuild/cockpit-composer/issues/1462 @teg mentioned that it would be a good idea to implement searching and pagination in dnf-json instead of the current behavior which has it dumping all packages into json and processing that in osbuild-composer. This is to help speed up the responsiveness of the cockpit-composer ui, but composer-cli cmdline use will also see the same benefit.

My current 'plan' is to add a search command to dnf-json which ties into the dnf query parameters (https://dnf.readthedocs.io/en/latest/api_queries.html) and then add the ability to use them to the weldr api. I'll work with @jkozol to figure out exactly what the UI needs from the API.

bcl:
In addition to adding the search command to dnf-json I've been experimenting with ways to speed up the listing of all packages. See pr #2918 for what I've come up with as the 'best' solution so far.

Currently, on my 4G/2core VM running Fedora 35, it takes around 5-6s to list all packages with `/api/v0/modules/list/**`. I experimented with changes to dnf-json and was able to shave off maybe 2s from that by limiting it to the latest NEVRA and by filtering on x86_64 + noarch packages. The timing there is fuzzy though, dnf isn't very consistent in how long it takes to return results. The other problem with this is that it changes the package list so that it doesn't include other builds or i686 versions of the packages.

So I added a 60s cache to the dnf-json results in osbuild-composer. This works really well in my testing, with results being returned in under 1s at the API level. The UI takes longer, either due to my browser or some other factors that I cannot measure. The result is that the UI feels much more responsive while you are using it. And yes, you will get a slow response every 60s when it refreshes the results, but every other page transition is fast so I think the improvement is worth it.

teg:
@jkozol do we still list all packages in the new UI, or is it all search based now?

If we do list them, the additional speedup possibility I had in mind was pagination all the way from dnf-json. I think we would need to change the return structs a bit to collect all versions of a package together in one object and paginate at that level.

Something like: give me package 10 to 20, each version of each package.

Not sure how much is gained by this given the other optimisations, but my thinking was what we would never pass tens of thousands of entries from dnf-json to composer.

teg:
Another idea, I don't know why we never did this:

Could we add a call to just download metadata for a repo? On start (or repo add), we could download all metadata in the background so that when someone starts using composer the metadata has most likely been downloaded. We still need to deal with expiration, but it would become the exception rather than the rule.

[...]

PR #2918, opened 2022-08-26
Cache the results of dnf-json dump and search commands

bcl:
This is a proposed change, so I'm making it a Draft for now. Discussion will be in #2887

This caches the results of dnf-json for 60s, the cache is based on the Request made, so different distros, repos, search terms are considered different. It returns the results for 60s and then makes a new request to dnf-json, and it cleans up the map when CleanCache() is called so that memory use doesn't go crazy.

The result of this is that when using cockpit-composer to page through lists of packages, or explore package dependencies, it speeds things up 'a whole bunch'. The only possible drawback I see is that every 60s it takes a bit longer to get results.

This pull request includes:

### DIFF
--- a/internal/dnfjson/cache.go
+++ b/internal/dnfjson/cache.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+
 	"github.com/gobwas/glob"
 )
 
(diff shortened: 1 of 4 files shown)


teg:
Nice speedup!

I wonder if we should limit the size (maybe in number of entries?) just in case? Given the short time out, I can't see anyone creating too big caches, but just in case someone does something imaginative...

I wonder if a further improvement is possible by somehow passing in an argument to dnf-json that says "if the metadata didn't change, do nothing, we will keep using the cache even after 60s".

bcl:
I'm not sure if we need to do more to keep the cache size down. I'll see if I can add some diagnostics to it and then hit it with a pile of requests to get a better handle on the memory usage. Most of it should be the full listing result. And second will be dependency lists for packages being viewed in the UI. And they are all shared among all users of the instance. Anything older than 60s and not repeated will then be deleted when CleanCache() is called after a Depsolve so I think (without good evidence :)) that it should be ok.

Most of the slowdown is in dnf-json so I'd try to avoid calling it at all. I can try some kind of 'tell me metadata changed, I have the old version' query though, just to see if it makes any significant difference.

bcl:
Ok, so I did a bunch of poking around at memory usage using pprof and a debug api route to dump details and I'm pretty sure this is ready for review.

I don't think there is any need to try and limit the memory usage other than with the timeout. It looks like it is 'naturally' limiting itself with something in the dnf metadata access causing the dnf-json calls to be serial, not parallel. If you hit the API with a bunch of depsolve requests and you can see them start and wait their turn. So it can only add new data to the cache as fast as it can run these requests. I have seen as many as 80 results cached in my tests, but never more than that.

For memory usage, I saw a max of around 90M when looking at it with pprof (you can't see the memory usage for the map itself, but the functions adding to the map seem to reflect the size) and with my debug output it hit about 100M. The largest part of the usage is 55M when you do a listing of all the packages. This is on Fedora 35 so the package list is really big, 63987 of them so on CentOS and RHEL this will be quite a bit less. With the giant listing timed out it was less than 1M for most requests.

internal/dnfjson/cache.go
```diff
@@ -207,3 +209,60 @@ func dirSize(path string) (uint64, error) {
...
+func NewDNFCache(timeout time.Duration) *dnfCache {
+	return &dnfCache{
+		results: make(map[string]dnfResults),
+		timeout: timeout,
+		RWMutex: new(sync.RWMutex),
+	}
+}
+
+// CleanCache deletes unused cache entries
+// This prevents the cache from growning for longer than the timeout interval
```
ondrejbudai, line 237: growning => growing

internal/dnfjson/dnfjson.go
```diff
@@ -422,6 +454,21 @@ type Request struct {
 	Arguments arguments `json:"arguments"`
 }
 
+// Hash returns a hash of the unique aspects of the Request
+func (r *Request) Hash() string {
+	h := sha256.New()
+	h.Write([]byte(r.Command))
```
ondrejbudai, line 460:
```suggestion
	_ = h.Write([]byte(r.Command))
```

feels slightly ugly but it's probably the easiest fix.

ondrejbudai (approved):
I think that cache cleaning via timeout is reasonable enough. So we just need to fix the linter and this is imho ready to go.

internal/dnfjson/cache.go
bcl: Whoops.

internal/dnfjson/dnfjson.go
bcl: Yeah, Write actually cannot return an error but it also returns the length of the input (which it also doesn't change) so it's cleaner to just use `//nolint:errcheck` on the whole block.

ondrejbudai (approved):
I took the liberty of removing the merge commit and rebasing this PR. Looks good, let's get it merged. Thanks, @bcl!

PR #2919, opened 2022-08-26
Facts koji n centos

Gundersanne:


teg:
CentOS Stream can still be subscribed, can't it?

teg:
Special casing the koji calls makes it feel to me that something is off. Maybe we don't add any facts for the cloud api, but require the caller to set them?

supakeen:
I would:
- Remove the facts for Koji-builds
- Remove the facts for CentOS

So exactly as this PR does.

We can then later decide on how to approach these, personally I feel there should be a split between RHEL and non-RHEL in how we store our facts (same format, don't create the symlink by default). We can do this based on subscription status when the image is built, or differ on distro.

As for the Koji build, remove it for now as it affects official builds going through Koji then decide if it's OK to store these at a later time.

supakeen (approved):

Gundersanne:
Reworded the first commit. after https://github.com/osbuild/osbuild-composer/pull/2920 got merged it didn't make much sense anymore.

ondrejbudai (approved):

ondrejbudai:

PR #2927, opened 2022-08-29
WIP: Add support for AWS RHEL 9.0 EUS images to Image Builder

FKolwa:
**WIP PR for easier reviewing**

- Added EC2 EUS image config
- Added EC2 EUS image type
- Added package set with dummy package name for not yet released rhui client package.

Open questions:
- What am I missing to enable image builder to build RHEL 9.0 EUS images for AWS?


internal/distro/rhel9/package_sets.go
```diff
@@ -730,6 +730,17 @@ func rhelEc2PackageSet(t *imageType) rpmmd.PackageSet {
 	return ec2PackageSet
 }
 
+// rhel-eus-ec2 image package set
+func rhelEc2EusPackageSet(t *imageType) rpmmd.PackageSet {
+	ec2PackageSet := ec2CommonPackageSet(t)
+	ec2PackageSet = ec2PackageSet.Append(rpmmd.PackageSet{
+		Include: []string{
+			"rh-amazon-rhui-client-sap-bundle-e4s", //TODO: Replace with the package build in RHUI-289
```
achilleas-k, line 738: Adding a change request here to make it more visible and block the PR.

achilleas-k (changes requested):

achilleas-k:
Quickly read through and it all looks good.  I don't think there would be anything left for the image definition itself.  We might need to figure out which repository sources to use for building the image, especially for EUS versions that aren't the current GA (e.g., when building 9.0 EUS after 9.1 is released).

FKolwa:
Closed as the implementation was deemed unnecessary at this point. I will keep the changes in my branch for now just in case the issue comes up again in the near future.

PR #2932, opened 2022-08-29
test/aws.sh: use bigger instance for `cloud-image-val-test`

jabia99:
For some tests, using a micro instance type is not enough. Changing to t3.medium will solve these issues.

The PR also includes conversation around CIV release process.

This pull request includes:

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -229,7 +229,7 @@ tee "${TEMPDIR}/resource-file.json" <<EOF
         {
             "ami": "$AMI_IMAGE_ID",
             "region": "us-east-1",
-            "instance_type": "t3a.micro",
+            "instance_type": "t3.medium",
             "username": "$SSH_USER",
             "name": "civ-testing-image"
         }


test/cases/aws.sh
```diff
@@ -216,7 +216,7 @@ fi
 
 greenprint "Pulling cloud-image-val container"
 
-CONTAINER_CLOUD_IMAGE_VAL="quay.io/cloudexperience/cloud-image-val-test:latest"
```
thozza, line 219: Alternatively, you could be pushing the latest container built from main under the `latest` tag to the registry, no? I don't have anything against this change, but IMO it makes sense for the `latest` tag to point to the newest build of the container, which perhaps should be the one built from `main`...

thozza (dismissed):

test/cases/aws.sh
jabia99: This is confusing I know :S. The reason we push to main and not latest is to avoid code duplication in CIV CI. We are not going to use the :latest tag anymore.

jabia99: I just realized it won't be that hard to fix this. I just have to keep an eye out after a few systems that depend on latest. Thanks!

thozza: OK, fair enough.

thozza (approved):

jrusz (changes requested):
Can you walk us through the process of the container release please? Because when I look at the CI in CIV repo, it will first build and push the container and then run the tests, meaning if the tests fail the container will be pushed and tagged regardless, or am I missing something?

jabia99:
> Can you walk us through the process of the container release please? Because when I look at the CI in CIV repo, it will first build and push the container and then run the tests, meaning if the tests fail the container will be pushed and tagged regardless, or am I missing something?

Sure. :)  Each time we run the CI, a new image will be tagged and pushed. But it'll be tagged with the number of the PR i.e. `PR-42`. If all the tests pass and we merge the PR, when the new commits get into main, the pipeline will get triggered again but this time with the `:latest` tag. So if an image has the latest tag, it means it's with code from main.

(PR to change from `:main` to `:latest` tag)

thozza:
So I considered the process as described by @jabia99, however I admit that it would be better if the container would get pushed to `latest` tag only once all tests pass on the `main` branch. Because it could happen that the container would be pushed to the `latest` tag even if the CI pipeline failed on `main`. So doing the push as the last thin e.g. in the `finish` job would make more sense.

jrusz:
> > Can you walk us through the process of the container release please? Because when I look at the CI in CIV repo, it will first build and push the container and then run the tests, meaning if the tests fail the container will be pushed and tagged regardless, or am I missing something?
> 
> Sure. :) Each time we run the CI, a new image will be tagged and pushed. But it'll be tagged with the number of the PR i.e. `PR-42`. If all the tests pass and we merge the PR, when the new commits get into main, the pipeline will get triggered again but this time with the `:latest` tag. So if an image has the latest tag, it means it's with code from main.
> 
> (PR to change from `:main` to `:latest` tag: [osbuild/cloud-image-val#116](https://github.com/osbuild/cloud-image-val/pull/116))

Is there no risk, that the tests would pass on PR and not on main? I can imagine that if you don't work with latest rebase in your branch a change could be introduced that after merge on main would break stuff... That happens here in composer from time to time as well...

Wouldn't it be safer to run the tests first and then release the container with the correct tag in the next stage? Like keep what you have but don't use `latest` tag on main before tests, use it afterwards in a stage that will run only if tests pass...

jabia99:
> So I considered the process as described by @jabia99, however I admit that it would be better if the container would get pushed to `latest` tag only once all tests pass on the `main` branch. Because it could happen that the container would be pushed to the `latest` tag even if the CI pipeline failed on `main`. So doing the push as the last thin e.g. in the `finish` job would make more sense.

I though about that. But is it not duplicating tests? The tests that would get executed in main are exactly the same that the tests executed in the PR. I can add a new stage to only push to :latest once the tests pass in main no problem though.

I can only think of one scenario, that is if someone pushes directly to main instead of going the PR route.

jabia99:
> Wouldn't it be safer to run the tests first and then release the container with the correct tag in the next stage? Like keep what you have but don't use `latest` tag on main before tests, use it afterwards in a stage that will run only if tests pass...

Ohh, I see. Okay I'll add that new stage. Thanks both of you :heart:

ondrejbudai (changes requested):
Can you sync the PR description with the PR title?

ondrejbudai (approved):

jrusz (approved):

PR #2941, opened 2022-08-30
weldr: Preload metadata at startup

bcl:
For each of the supported distros start a goroutine to depsolve
'filesystem' which will preload the metadata making subsequent responses
faster.

This is safe to do without limits because we only supposed a limited
number of distros, and without additional locking because this is the
the same as hitting the API with multiple depsolve requests at the same
time.

This pull request includes:

I didn't add testing for this because I'm not sure it can be tested. Ideas are welcome :)

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -123,6 +123,8 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener,
 	}
 	c.weldrListener = weldrListener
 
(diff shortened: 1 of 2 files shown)


teg:
I think this makes sense! @jkozol and @ondrejbudai, what do you think?

ondrejbudai (approved):
Looks very nice! :)

achilleas-k (approved):
Looks nice indeed.  Gonna be great for that first compose request that always takes ages :)

PR #2953, opened 2022-09-02
distro/`ImageConfig`: use pointers to simple types and reflection in `InheritFrom()`

thozza:
These changes are required in order to enable unifying the `vhd` and `azure-rhui` image definitions. Detailed reasoning for each change is described in each commit message.

This pull request includes:

### DIFF
--- a/internal/distro/image_config.go
+++ b/internal/distro/image_config.go
@@ -1,6 +1,11 @@
 package distro
 
-import "github.com/osbuild/osbuild-composer/internal/osbuild"
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/osbuild/osbuild-composer/internal/osbuild"
+)
 
 type RHSMSubscriptionStatus string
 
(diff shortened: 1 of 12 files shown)


ondrejbudai (changes requested):
Fedora 35 has go 1.16, either we drop support for F35 and we stay on 1.16.

thozza:
I removed the update to newer Golang version.

internal/distro/image_config.go
```diff
@@ -62,98 +67,31 @@ type ImageConfig struct {
...
+			case reflect.Slice:
+				if field.IsNil() {
+					setValue = true
+				}
+			case reflect.Map:
+				if field.IsNil() {
+					setValue = true
+				}
+			default:
+				panic(fmt.Sprintf("unsupported field type: %s", field.Kind()))
```
achilleas-k, line 89: To avoid concrete types being added in the future, maybe we should add a comment or expand the error message to add something like "use pointers instead of concrete types".

internal/distro/image_config.go
```diff
@@ -62,98 +67,31 @@ type ImageConfig struct {
...
+			}
+
+			if setValue {
```
achilleas-k, line 92: We could skip using `setValue` and just use `field.IsNil()` directly here, making the case bodies no-ops, right?
Minor detail though.  Not an issue.

achilleas-k (commented):
It gets a bit annoying working with pointers to basic types when there are so many literals in the code, but I get the need for it.

Using reflection in `InheritFrom()` is great though.  Any addition that requires "registering" it in a different location is always a high risk for bugs.

internal/distro/image_config.go
thozza: Yes, it makes sense. Good idea 👍

achilleas-k (approved):
All good.

PR #2958, opened 2022-09-06
test: change CIV tag

jabia99:
We were using `latest` as tag, this can be dangerous as it's the default
tag, an anyone can accidentally update it. Using `prod` is safer.

Also use dev container image if the test script is running in CIV CI.

This pull request includes:

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -216,7 +216,15 @@ fi
 
 greenprint "Pulling cloud-image-val container"
 
-CONTAINER_CLOUD_IMAGE_VAL="quay.io/cloudexperience/cloud-image-val-test:latest"
+if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
+  # If running on CIV, get dev container
+  TAG=${CI_COMMIT_REF_SLUG}
+else
+  # If not, get prod container
+  TAG="prod"
+fi
+
+CONTAINER_CLOUD_IMAGE_VAL="quay.io/cloudexperience/cloud-image-val-test:$TAG"
 
 sudo ${CONTAINER_RUNTIME} pull ${CONTAINER_CLOUD_IMAGE_VAL}
 


Gundersanne:
I didn't quite get the conclusion from https://github.com/osbuild/osbuild-composer/pull/2932; I thought the latest branch was now only updated after main was green? What exactly is the prod tag?

jabia99:
> I didn't quite get the conclusion from #2932; I thought the latest branch was now only updated after main was green? What exactly is the prod tag?

It's exactly the same. Both `latest` and `prod` get updated only if code on main passes the CI. But as `latest` is the default tag, it can happen that someone accidentally updates it. I doubt something like that would happen as Nico and I are the only ones working on CIV, but it's safer nonetheless

Gundersanne:
> > I didn't quite get the conclusion from #2932; I thought the latest branch was now only updated after main was green? What exactly is the prod tag?
> 
> They are the same. Both `latest` and `prod` get updated only if code on main passes the CI. But as `latest` is the default tag, it can happen that someone accidentally updates it manually. I doubt something like that would happen as Nico and I are the only ones working on CIV, but it's safer nonetheless

Ah ok, thanks for the explanation!

test/cases/aws.sh
```diff
@@ -216,7 +216,16 @@ fi
 
 greenprint "Pulling cloud-image-val container"
 
-CONTAINER_CLOUD_IMAGE_VAL="quay.io/cloudexperience/cloud-image-val-test:latest"
+REPO="${REPO:-os-build-composer}"
```
jrusz, line 219: Hm, I don't think this is correct, shouldn't it be `osbuild-composer`? Also if you plan on using the same script in CIV repo I guess it would be better to use some CI variable? `CI_PROJECT_NAME` might work. Or just set it in CI env vars or parse it from some project url...

jrusz (changes requested):
Sounds good, just have one question/suggestion.

test/cases/aws.sh
jabia99: yes, The REPO variable is set in CIV env vars. I didn't saw any variable in `osbuild-composer` CI that could identify where the script is running so I just created that one. `CI_PROJECT_NAME` is a much better name though

jrusz: `CI_PROJECT_NAME` comes from https://docs.gitlab.com/ee/ci/variables/predefined_variables.html idk if the value is the actual name of the repo though, it sounds more like the project all the repos are in, I'd double check first.

jabia99: Ohh, I missunderstood you. Thanks, I'll double check

Gundersanne (approved):

PR #2963, opened 2022-09-07
pkg/dbjobqueue: fix dequeue constraint error

kingsleyzissou:
Trying to reproduce and debug an issue with the db jobqueue tests.

### DIFF
--- a/pkg/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/pkg/jobqueue/dbjobqueue/dbjobqueue.go
@@ -28,10 +28,10 @@ const (
 	sqlListen   = `LISTEN jobs`
 	sqlUnlisten = `UNLISTEN jobs`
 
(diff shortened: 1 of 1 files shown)


kingsleyzissou:
Thanks to @ondrejbudai, we established that this is more or less the scenario causing the postgres constraint violation:
- the dequeueing transaction starts (earlier)
- the enqueueing transaction starts
- the enqueueing command does its thing
- the dequeueing command does its thing

The dequeue & enqueue queries make use of the `now()` postgres timestamp function which returns the start time of the current transaction. To resolve this we can use the `statement_timestamp()` postgres function to return the start time of the current statement instead.

Gundersanne (approved):
Amazing, thank you!

ondrejbudai (approved):
Thanks for digging into this!

I hope it helps, you can never be sure when fixing race conditions.

PR #2973, opened 2022-09-08
workflow: Update to golangci-lint v1.49.0

### LINKED ISSUE #2972: golangci-lint should be updated to 1.49.0
The version of golangci-lint used in the workflow is old, v1.30.0, and needs to be updated and the new linting error fixed.

bcl:
Update golangci-lint version used in the workflow. Draft for now to see what it complains about. This will fix #2972

### DIFF
--- a/internal/upload/azure/azurestorage.go
+++ b/internal/upload/azure/azurestorage.go
@@ -186,7 +186,7 @@ func (c StorageClient) CreateStorageContainerIfNotExist(ctx context.Context, sto
 
 	_, err := containerURL.Create(ctx, azblob.Metadata{}, azblob.PublicAccessNone)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Great! Thank you for picking this up.

ondrejbudai:
Btw, it might be a good idea to switch to their official github action](https://golangci-lint.run/usage/install/#github-actions) since it's finally available. The benefit is that it would get autoupdated by dependabot.

bcl:
Ah, ha! So the linter complaints are only summarized in the logs, there are no line numbers. But GitHub adds the annotations to the file tab even for files not touched by the PR so that's where you look when it fails, not at the logs.

ondrejbudai (approved):
Looks good, thanks!

ondrejbudai (approved):
Thanks! :heart_eyes:

PR #2974, opened 2022-09-09
[rhel 8.7/9.1] explicitly add containernetworking-plugins to edge

achilleas-k:
Cherry-picked #2951 for backported release into RHEL 9.1.0.

----

In podman v4.0.0 the default network backend was switched from cni to netavark.  However, podman will choose cni if there are already containers, images, or cni networks preset on a system [1].

Starting with podman v4.2.0, containernetworking-plugins is no longer a hard requirement for podman.  So when an edge commit is built with an embedded container, podman v4.2.0+ will choose the cni network and fail with an error because the plugin isn't installed.

Adding the package explicitly alongside podman to avoid this issue with future RHEL 9.1 edge builds when they include containers.

This change does not affect test manifests.  The package is already included in manifests as a dependency of podman < v4.2.0.

See rhbz#2123210

[1] https://github.com/containers/common/blob/a083f790abf21048518f2d66b7e01370cc055790/pkg/config/containers.conf#L275-L278


ondrejbudai (dismissed):

achilleas-k:
The `diff-manifests` script needed a fix for PRs made against any branch other than `main`.  Branches aren't synced to GitLab so we need to fetch the base branch from GitHub.

achilleas-k:
Added commits required to fix some CI jobs.

achilleas-k:
Failures in new ostree test are being fixed in https://github.com/osbuild/osbuild-composer/pull/2979

henrywang (approved):
LGTM. Thanks @achilleas-k!

ondrejbudai (approved):

PR #2978, opened 2022-09-12
diff-manifest: fetch base branch from GitHub

achilleas-k:
Distro version branches aren't synced to GitLab, so we will need to fetch them from GitHub directly.
This is required for any PR made against any branch other than main.


achilleas-k:
The only failures here are the ones being fixed in https://github.com/osbuild/osbuild-composer/pull/2979

supakeen (approved):
Sure, but I'd be much happier with actually pushing these in the GitLab action instead. This works though.

PR #2981, opened 2022-09-12
Specify additional package in blueprint for tests

atodorov:
Related RHBZ#2065734

This pull request includes:

### DIFF
--- a/test/cases/vmware.sh
+++ b/test/cases/vmware.sh
@@ -124,6 +124,11 @@ version = "0.0.1"
 [[packages]]
 name = "bash"
 
+# Related RHBZ#2065734
+[[packages]]
+name = "ipa-client"
+version = "*"
+
 [customizations.services]
 enabled = ["sshd"]
 
--- a/tools/libvirt_test.sh
+++ b/tools/libvirt_test.sh
@@ -144,6 +144,11 @@ if [ -z "${LIBVIRT_IMAGE_URL}" ]; then
 name = "bp"
 description = "A base system"
 version = "0.0.1"
+
+# Related RHBZ#2065734
+[[packages]]
+name = "ipa-client"
+version = "*"
 EOF
 
     # Prepare the blueprint for the compose.


jrusz (dismissed):
Why not add the package for the libvirt test which builds the qcow2 image? Here https://github.com/osbuild/osbuild-composer/blob/main/tools/libvirt_test.sh#L143

I guess it doesn't really matter, just curious, lgtm otherwise.

test/cases/vmware.sh
```diff
@@ -124,6 +124,10 @@ version = "0.0.1"
 [[packages]]
 name = "bash"
 
+[[packages]]
```
jrusz, line 127: Actually, I think it would be good to have a comment here with the rhbz# as well not just in the commit. Might be easier to find in the future.

test/cases/vmware.sh
```diff
@@ -124,6 +124,10 @@ version = "0.0.1"
...
+name = "ipa-client"
+version = "*"
```
jabia99, line 129: IIRC the version is not needed when we don't specify one

jabia99 (approved):
LGTM! Just a curious small comment

test/cases/vmware.sh
atodorov: B/c I searched in `test/cases` and didn't find the file. Added.

test/cases/vmware.sh
atodorov: I find it more explicit that way. Let me know if you have strong objections.

jrusz (dismissed):
I'm not sure we want it in two places, but can't hurt, LGTM

test/cases/vmware.sh
jabia99: Makes sense, no objection on my side

jrusz:
Needs rebase to get https://github.com/osbuild/osbuild-composer/commit/05503c58c2589105431b8985dd4b72590d521ed4

jrusz:
What happened here? Workflows are not triggering...

jrusz (approved):

PR #2991, opened 2022-09-15
tests/upgrade: copy shared_lib.sh to the test vm

jrusz:
shared_lib.sh now contains some more helper functions and the verification script uses them. Copy it over so that we can source it. Also install all other dependencies before the actual upgrade.

This pull request includes:

### DIFF
--- a/test/data/upgrade8to9/upgrade_verify.sh
+++ b/test/data/upgrade8to9/upgrade_verify.sh
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-source /usr/libexec/tests/osbuild-composer/shared_lib.sh
+source shared_lib.sh
 
 WORKSPACE=$(mktemp -d)
 function cleanup() {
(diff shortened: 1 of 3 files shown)


jrusz:
Link to Upgrade test job https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3034252818

test/data/upgrade8to9/upgrade_verify.sh
```diff
@@ -2,7 +2,7 @@
 
 set -euxo pipefail
 
-source /usr/libexec/tests/osbuild-composer/shared_lib.sh
```
thozza, line 5: I'm probably missing some more context, but isn't the osbuild-composer-tests RPM installed on the upgraded system? If not, then this makes sense, if it is, then I'm not sure why is this change needed. Could you please elaborate? Thanks.

jrusz: Just the `osbuild-composer` package from the latest nightly is installed on the test VM. See https://github.com/osbuild/osbuild-composer/blob/main/test/data/upgrade8to9/upgrade_prepare.sh#L11

So I copy the script over to the test VM so it can be sourced. The change is visible in this commit.

PR #3006, opened 2022-09-21
cloudapi: add option to upload public AMIs

ondrejbudai:
Similar thing to #2897, but for AMIs. In the upload options for AWS, there's a new field public. If set to true, the image will be made public.

This means that `share_with_accounts` is no longer required because sharing a public image doesn't make sense. Instead, I added a check that at least one of `share_with_accounts` or `public` is set. Tests were added and adjusted so everything is sanely tested. See the individual commits for details

This pull request includes:

  - Documented in the API.


internal/cloudapi/v2/v2_test.go
```diff
@@ -558,6 +559,82 @@ func TestCompose(t *testing.T) {
...
+		"image_request":{
+			"architecture": "%s",
+			"image_type": "aws",
+			"repositories": [{
+				"baseurl": "somerepo.org",
+				"rhsm": false
+			}],
+			"upload_options": {
+				"region": "eu-central-1",
+				"shared_with_accounts": 0
```
thozza, line 627: I would expect this to violate the API specification and schema. `shared_with_accounts` is expected to be an array of strings consisting of 12 digits.. You should IMHO use `[]` instead, no? 🤔

internal/cloudapi/v2/errors.go
```diff
@@ -123,6 +124,7 @@ func getServiceErrors() serviceErrors {
 		serviceError{ErrorUnsupportedImage, http.StatusBadRequest, "This compose doesn't support the creation of multiple images"},
 		serviceError{ErrorInvalidImageFromComposeId, http.StatusBadRequest, "Invalid format for image id"},
 		serviceError{ErrorImageNotFound, http.StatusBadRequest, "Image with given id not found"},
+		serviceError{ErrorUnreachableImage, http.StatusBadRequest, "You need to use either share_with_accounts, or public"},
```
thozza, line 127: The message should be probably `"You need to specify at least one of share_with_accounts, or public"`. IOW the expectation is not to specify one or the other exclusively, but at least one of them...

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -826,7 +826,10 @@ components:
       type: object
       required:
         - region
-        - share_with_accounts
```
thozza, line 829: I wonder if it would not be better to have `oneOf` block here and specify the two variants of the required fields. It is in both cases the `region`, but in one it would be `share_with_accounts` and on the second the `public` field.

thozza (commented):
Nice work. Added some comments and there are some failing unit tests...

Gundersanne (dismissed):
lgtm, ty!

internal/cloudapi/v2/v2_test.go
ondrejbudai: Ugh, I had some issues with rebases and they probably didn't work as expected, lemme fix.

PR #3010, opened 2022-09-21
templates/packer: Allow token url to be set by cloud-init vars

Gundersanne:
Hardcoding the token url renders the image useless if it ever needs to be changed.


diaasami (dismissed):

ondrejbudai (changes requested):
Can you document the new variable in `templates/package/README.md`? This is one of the rare cases when we have docs. :laughing: :cry:

ondrejbudai (approved):
Perfect!

PR #3017, opened 2022-09-23
packit: Enable Bodhi updates for unstable Fedoras

ochosi:
As we don't only want to get Bodhi updates for the stable releases, but also the ones still in development, we need to use 'fedora-branched'.

See https://packit.dev/docs/configuration/#aliases


thozza (approved):
LGTM, Thanks ❤️

ondrejbudai:
Just a config change, merging.

PR #3018, opened 2022-09-23
cloudapi: make ImageRequests tied to their UploadOptions

ondrejbudai:
Previously, one could pass GCPUploadOptions to an aws image request because the spec technically allows that. I haven't found any validation in our code, so I presume that the result would be pretty undefined.

A real-life example: If someone wanted an aws image but didn't specify `share_with_accounts`, this is valid according to AWSS3UploadOptions. Composer doesn't perform any additional checks, so it would just proceed to building such an image, and uploading it to our account where no one can use access it.

This commit makes ImageRequest be one of AWSEC2ImageRequest, AzureImageRequest etc. XYZImageRequest always requires its `upload_options` to be XYZUploadOptions. As all of these ImageRequests contain some common fields, I created AbstractImageRequest to deduplicate them.

One thing that I don't like is that image types called AWSEC2ImageRequestImageTypeAws are long, and in most cases we need to cast them to `string`, which loses a bit of type safety.

What I like is that the API spec is now more self-explanatory and helpful for composer's users.

I also had to adjust several tests that build aws images without setting `share_with_accounts`.

This  breaks backward compatibility but the only broken requests are non-sense ones.

---
This is actually a prequel to https://github.com/osbuild/osbuild-composer/pull/3006. A suggestion from @thozza was to make AWSEC2UploadOptions specified like this:

```
oneOf:
  - required:
    - share_with_accounts
  - required:
    - public
```

The issue is that, the spec validator will happily validate `AWSEC2UploadOptions` without neither of these fields set using the `AWSS3UploadOptions` spec. I did not like that, so I began thinking how to make the schema more error-proof and came out with this.

--- 

I was thinking if this is fine for future extensions: One extension that I can think of is to add the ability to upload any image to an S3-like storage. I think this is fine in the spec because we just make all image requests uploadable to both their target and to S3. A bit of code-duplication can happen but with `oneOf` and `allOf`, we should be just fine.

This pull request includes:

  - the API spec is now more self-explanatory

### DIFF
--- a/internal/cloudapi/v2/openapi.v2.yml
+++ b/internal/cloudapi/v2/openapi.v2.yml
@@ -735,45 +735,111 @@ components:
           $ref: '#/components/schemas/Customizations'
         koji:
           $ref: '#/components/schemas/Koji'
+
     ImageRequest:
+      oneOf:
+        - $ref: '#/components/schemas/AWSEC2ImageRequest'
+        - $ref: '#/components/schemas/AWSS3ImageRequest'
+        - $ref: '#/components/schemas/GCPImageRequest'
+        - $ref: '#/components/schemas/AzureImageRequest'
+        - $ref: '#/components/schemas/ContainerImageRequest'
+
+    AbstractImageRequest:
       required:
         - architecture
-        - image_type
         - repositories
       properties:
         architecture:
           type: string
           example: 'x86_64'
-        image_type:
-          $ref: '#/components/schemas/ImageTypes'
         repositories:
           type: array
           items:
             $ref: '#/components/schemas/Repository'
         ostree:
           $ref: '#/components/schemas/OSTree'
-        upload_options:
-          $ref: '#/components/schemas/UploadOptions'
-    ImageTypes:
-      type: string
-      enum:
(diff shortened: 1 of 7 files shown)


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -735,45 +735,111 @@ components:
...
+        - $ref: '#/components/schemas/AbstractImageRequest'
+        - type: object
+          required:
+            - image_type
+          properties:
+            image_type:
+              type: string
+              enum:
+                - edge-container
+                - iot-container
```
thozza, line 840: Are you sure that nobody relies on uploading these also to S3? 🤔  Just checking...

thozza (approved):
Overall, this IMO makes sense and looks good 👍 .

I don't like the many occurrences of casting to `string`, but I don't have any better suggestion since the code is generated from the API spec. I also think that the code may get uglier if we ever decide to allow uploading images to S3, in addition to their native upload target, but that is something for our future selfs... 😅

Gundersanne (approved):
lgtm!

PR #3023, opened 2022-09-27
Weldr/Cloud API: simplify GCP upload options

thozza:
In both, the Weldr and Cloud APIs, users had to previously provide values, which could have been generated automatically by osbuild-composer, or were in reality static and could be configured on the worker.

Specifically:

- `Bucket` - bucket name usually does not change for Cloud API calls, since it is specific to the GCP credentials provisioned on the worker and as a result, to the respective GCP project. Therefore it makes more sense to allow setting the GCP bucket in the worker configuration when composer is running in a SaaS deployment.
- `Object` - object name used for the image when uploading it to the GCP Storage had to be previously provided by the user of the Weldr API. For Cloud API, the value has been generated by the implementation. In reality, the object name is not very important, as long as it is unique. The reason is that the object gets deleted after the image importing finishes.

Notable changes:

- Allow setting of the GCP Bucket in the worker configuration.
- Make the `Bucket` optional for GCP upload options in the **Cloud API**. Adjust the Cloud API test case to set the Bucket in the worker configuration, instead of passing it in the compose request.
- Extend Weldr API to generate a unique GCP Object name if it was not provided in the job.
- Make the GCP Object name optional for Weldr API. Adjust the Weldr API test case to not provide the Object name when passing the GCP upload options.
- Flip the logic in the worker when uploading image to AWS EC2/S3 WRT the Bucket name preference. The Bucket value set in the worker configuration is now used only if the Bucket was not provided in the job.
- Extend the packer template ansible role to configure the GCP Bucket in worker configuration.

This pull request includes:


internal/cloudapi/v2/server.go
```diff
@@ -1,3 +1,4 @@
+//go:generate go run -mod=mod github.com/deepmap/oapi-codegen/cmd/oapi-codegen --package=v2 --generate types,server,spec -o openapi.v2.gen.go openapi.v2.yml
```
achilleas-k, line 1: This is already specified in `handler.go`.  Does adding it here provide anything extra?

achilleas-k (approved):
LGTM.  One tiny question.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -472,7 +472,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			}
 
 			bucket := targetOptions.Bucket
-			if impl.AWSBucket != "" {
+			if bucket == "" {
```
ondrejbudai, line 476: @Gundersanne Is this fine for the service?

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -472,7 +472,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 			}
 
 			bucket := targetOptions.Bucket
```
ondrejbudai, line 475: I think that a nice addition would be erroring out if the bucket name isn't set in either the config, nor target options.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -564,24 +565,29 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				break
 			}
 
-			logWithId.Infof("[GCP] 🚀 Uploading image to: %s/%s", targetOptions.Bucket, targetOptions.Object)
+			bucket := targetOptions.Bucket
```
ondrejbudai, line 573: I think that a nice addition would be erroring out if the bucket name isn't set in either the config, nor target options.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -565,29 +565,34 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				break
 			}
 
+			object := targetOptions.Object
+			if object == "" {
```
ondrejbudai, line 569: Not a fan tbh. It means that both the Cloud API and the worker can generate names. I suggest doing it on just one layer.

internal/weldr/upload.go
```diff
@@ -291,9 +291,10 @@ func uploadRequestToTarget(u uploadRequest, imageType distro.ImageType) *target.
 			}
 		}
 
-		// The uploaded image object name must have 'tar.gz' suffix to be imported
+		// Providing the Object name is optional. If it is provided, we must
+		// ensure that it has a '.tar.gz' suffix to be successfully imported.
 		objectName := options.Object
-		if !strings.HasSuffix(objectName, ".tar.gz") {
+		if objectName != "" && !strings.HasSuffix(objectName, ".tar.gz") {
```
ondrejbudai, line 297: I would suggest generating the object name here for the sake of consistency, see my other comment in the worker.

ondrejbudai (changes requested):
I have concerns about generating the GCP object names on multiple levels, I think that we should make it more consistent and do it in the API.

Also, I would really love to make the worker more dumb and predictive.

internal/cloudapi/v2/server.go
thozza: hmm... I missed this and for some reason, the generation of the API code didn't work for me without this... I'll try it again 🤔

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Just FYI, we discussed this on Slack. The current logic is IMO weird, but I can live with it. I changed it to make it consistent with more appealing logic that I used for GCP. If we agree that we want to stick to the current logic used with AWS, I'll change it also for GCP, although I don't like it.

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Sounds reasonable....

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: "doing it on just one layer" means in this case doing it either in two places (Weldr and Cloud API) or in worker.

internal/weldr/upload.go
thozza: Fine with me. Although if you insist, we should change it for AWS as well, because it generates the `key` already in the worker's code https://github.com/osbuild/osbuild-composer/blob/d9cfbea399426d4f52236c96a0cea769146919e4/cmd/osbuild-worker/jobimpl-osbuild.go#L469-L471

cmd/osbuild-worker/jobimpl-osbuild.go
Gundersanne: Should be fine, the bucket is only specified in the worker.

Gundersanne:
Should we update https://github.com/osbuild/osbuild-composer/blob/main/templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_gcp_creds.sh in the same PR?

internal/cloudapi/v2/server.go
thozza: It all works now even without my addition. I'm not sure what I was doing wrong 🤷‍♂️  I'll drop this change.

thozza:
> Should we update
>  ...
> in this PR?

Looking at it, probably yes. Everything should keep working even without this change as the option to pass the bucket via the API call is still present. Nevertheless changing the script will be needed anyway in order to stop passing the bucket over the API, so I may as well change it now.

Thanks for pointing this out!

thozza:
I addressed all review comments and added a change for the packer template as suggested by Sanne. Please have a look.

ondrejbudai:
Oh no, it now needs to be rebased. :/ Otherwise, this is amazing!

thozza:
Rebased on top of `main`... 🦾

ondrejbudai (dismissed):
Thanks a lot, a great cleanup!

thozza:
Rebase on `main` without doing any other changes.

achilleas-k (approved):

PR #3056, opened 2022-10-11
enable persistent logging on ostree images

runcom:
building a r4e image results in no persistent logging on the system (which is highly surprising). This is because journal isn't configured to do persistent logging nor /var/log/journal is in the ostree (which is a trigger for journald to start logging in persistent mode, yeah, this is also kind of surprising).
This patch makes ostree deploy create that directory too and I can verify logs is present on a booted r4e image installed via simplified installer.

coreos* images ship a postprocess script to enable the config option instead https://pagure.io/workstation-ostree-config/blob/main/f/fedora-common-ostree.yaml#_100

cc @cgwalters

Signed-off-by: Antonio Murdaca <runcom@linux.com>

This pull request includes:

### DIFF
--- a/test/data/ansible/check_ostree.yaml
+++ b/test/data/ansible/check_ostree.yaml
@@ -697,6 +697,28 @@
             failed_counter: "{{ failed_counter | int + 1 }}"
       when: skip_rollback_test == "false"
 
+    - name: check journald has persistent logging
+      block:
+        - name: lsit boots
+          shell: journalctl --list-boots
+          register: result_list_boots
+
+        - assert:
+            that:
+              - result_list_boots.stdout_lines | length > 1
+            fail_msg: "journald hasn't persistent logging"
+            success_msg: "journald has persistent logging"
+      always:
+        - set_fact:
+            total_counter: "{{ total_counter | int + 1 }}"
+      rescue:
+        - name: failed count + 1
+          set_fact:
+            failed_counter: "{{ failed_counter | int + 1 }}"
+      when:
+        - skip_rollback_test == "false"
+        - result_rollback is succeeded
+
     # case: check ostree commit after rollback
     - name: check ostree commit after rollback
       block:


runcom:
@achilleas-k can we check what's happening in the fedora images too (following up from our previous conversation) - Christian also mentioned something around systemd tempfiles too

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 22fd6334c6789143c86a63a9d80db9307bc29bcc with the main merge-base 3933c37f466672821971c0f7b67957eaba778a44).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3155209625/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3155209625/artifacts/browse

achilleas-k:
> @achilleas-k can we check what's happening in the fedora images too (following up from our previous conversation) - Christian also mentioned something around systemd tempfiles too

Yeah I'll have a look

cgwalters:
From my PoV the systemd default of "enable journal if /var/log/journal exists" was a compatibility hack from the timeframe of introducing systemd.  But nowadays, I think we should switch systemd to enabling it by default, and telling anyone who wants it off to explicitly set `Storage=volatile|none` in `/etc/systemd/journald.conf` or so.

Anyways for now, I would say that it's better to have systems that want this to use a journald drop-in instead of making `/var/log/journal` by default.

I know Image Builder only wants to accept RPMs today - but is there really not an extant RPM that this could go into?  

Hmmm...I guess we could create a subpackage of ostree like `ostree-default-system-units` that includes tweaks like this...

gicmo:
> From my PoV the systemd default of "enable journal if /var/log/journal exists" was a compatibility hack from the timeframe of introducing systemd. But nowadays, I think we should switch systemd to enabling it by default, and telling anyone who wants it off to explicitly set `Storage=volatile|none` in `/etc/systemd/journald.conf` or so.

I am glad you are saying that because I was leaning towards that too. For the time being I think we should explicitly configure it via `[Journal]\nStorage=persistent` in a drop-in. We can totally do that. Just need a new osbuild stage.

runcom:
moved this to a draft while the team handles a new osbuild stage for this :)

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 472d103352a7a4b2b09e5465398dae401b9f5dc3 with the main merge-base 3933c37f466672821971c0f7b67957eaba778a44).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3160688660/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3160688660/artifacts/browse

runcom:
added a failing test to validate a fix when we'll have one meanwhile

runcom:
cherry-picked in #3069

PR #3072, opened 2022-10-17
Ostree resolve job

croissanne:
This pull request includes:

### DIFF
--- a/internal/ostree/ostree.go
+++ b/internal/ostree/ostree.go
@@ -80,22 +80,15 @@ func ResolveRef(location, ref string) (string, error) {
 // without URL results in a ParameterComboError. Failure to resolve the
 // checksum results in a ResolveRefError.
 //
-// If Ref is not specified in the RequestParams, the defaultRef is used.
-// If Parent is not specified in the RequestParams, the value of Ref is used
-// (which will be defaultRef if neither is specified).
+// If Parent is not specified in the RequestParams, the value of Ref is used.
 //
 // If any ref (Ref or Parent) is malformed, the function returns with a RefError.
-func ResolveParams(params RequestParams, defaultRef string) (ref, checksum string, err error) {
-	// Determine value of ref
+func ResolveParams(params RequestParams) (ref, checksum string, err error) {
 	ref = params.Ref
-	if ref != "" {
-		// verify format of provided ref
-		if !VerifyRef(params.Ref) {
-			return "", "", NewRefError("Invalid ostree ref %q", params.Ref)
-		}
-	} else {
-		// if ref is not provided, use distro default
-		ref = defaultRef
+
+	// Determine value of ref
+	if !VerifyRef(params.Ref) {
+		return "", "", NewRefError("Invalid ostree ref %q", params.Ref)
 	}
 
 	// Determine value of parentRef
@@ -111,7 +104,7 @@ func ResolveParams(params RequestParams, defaultRef string) (ref, checksum strin
 		}
 	} else {
 		// if parent is not provided, use ref
-		parentRef = ref
+		parentRef = params.Ref
 	}
 
 	// Resolve parent checksum
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -2357,8 +2357,13 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
 	if testMode == "1" || testMode == "2" {
 		// Fake a parent commit for test requests
 		cr.OSTree.Parent = "02604b2da6e954bd34b8b82a835e5a77d2b60ffa"
-	} else {
-		ref, checksum, err := ostree.ResolveParams(cr.OSTree, imageType.OSTreeRef())
+	} else if imageType.OSTreeRef() != "" {
+		// If the image type has a default ostree ref, assume this is an OSTree image
+		reqParams := cr.OSTree
+		if reqParams.Ref == "" {
+			reqParams.Ref = imageType.OSTreeRef()
+		}
+		ref, checksum, err := ostree.ResolveParams(reqParams)
 		if err != nil {
 			errors := responseError{
 				ID:  "OSTreeOptionsError",
--- a/internal/worker/json.go
+++ b/internal/worker/json.go
@@ -256,6 +256,28 @@ type ContainerResolveJobResult struct {
 	JobResult
 }
 
+type OSTreeResolveSpec struct {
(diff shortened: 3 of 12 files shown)


internal/worker/json.go
```diff
@@ -256,6 +256,29 @@ type ContainerResolveJobResult struct {
...
+
+// TODO better name
+type OSTreeResolvedSpec struct {
+	URL      string `json:"url"`
+	Ref      string `json:"ref"`
+	Checksum string `json:"checksum"`
+}
+
+type OSTreeResolveJob struct {
+	Specs []OSTreeResolveSpec `json:"ostree_resolve_specs"`
```
croissanne, line 273: @achilleas-k I noticed in `internal/distro` there's some indication of supporting multiple commits. But the `ImageOptions` don't account for this at all. 

Should I bother with a slice here to make any switch to multiple commits easier in a worker/job api?

achilleas-k: That's probably a good idea.  I also saw that @teg added multiple ostree commit support to the Pipeline object, but I'm not sure what the use case for this is exactly.  If it's for pipelines that can **build** multiple commits, that's one thing.  We would only need to resolve multiple ostree specs if we have multiple parents and I'm not sure we have a use case for that.

All that said, having a slice in the job args and only ever using one element can't hurt, right?


https://github.com/osbuild/osbuild-composer/blob/a23b1c4b3e414d80600d19c5fc81659b4f1a24c1/internal/manifest/pipeline.go#L29

croissanne:
I will split out  switching over the cloudapi (the last commit) into a separate PR; that way we can easily transition this on production.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c4cf3ebdc1922a1823f1989a45266f0a7bb34c50 with the main merge-base 545ce3412e85c838247adc0d2210dd632528c322).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3188097578/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3188097578/artifacts/browse

internal/weldr/api.go
```diff
@@ -2357,8 +2357,13 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
 	if testMode == "1" || testMode == "2" {
 		// Fake a parent commit for test requests
 		cr.OSTree.Parent = "02604b2da6e954bd34b8b82a835e5a77d2b60ffa"
-	} else {
-		ref, checksum, err := ostree.ResolveParams(cr.OSTree, imageType.OSTreeRef())
+	} else if imageType.OSTreeRef() != "" {
+		// If the image type has a default ostree ref, assume this is an OSTree image
```
croissanne, line 2361: @achilleas-k is this a reasonable assumption to make? I found out this bit got executed every time, and the way the default ref worked is that ResolveParams actually allowed an empty ref to be returned (if it isn't an ostree image).

achilleas-k: Yes, all implementations of that method look like this:

```go
func (t *imageType) OSTreeRef() string {
	d := t.arch.distro
	if t.rpmOstree {
		return fmt.Sprintf(d.ostreeRefTmpl, t.Arch().Name())
	}
	return ""
}
```

and that's generally how we identify ostree-based images when we need to.

croissanne: perfect, I'll use this as much as possible then to be consistent

internal/ostree/ostree.go
```diff
@@ -80,22 +80,15 @@ func ResolveRef(location, ref string) (string, error) {
...
 //
-// If Ref is not specified in the RequestParams, the defaultRef is used.
-// If Parent is not specified in the RequestParams, the value of Ref is used
-// (which will be defaultRef if neither is specified).
+// If Parent is not specified in the RequestParams, the value of Ref is used.
 //
 // If any ref (Ref or Parent) is malformed, the function returns with a RefError.
-func ResolveParams(params RequestParams, defaultRef string) (ref, checksum string, err error) {
-	// Determine value of ref
+func ResolveParams(params RequestParams) (ref, checksum string, err error) {
```
achilleas-k, line 86: Reading this now I notice that returning the `ref` is essentially useless (like when we used to return the URL).  Maybe we can drop `ref` from the returns and only return `checksum` and `error`, and since the caller is responsible for using the default ref if one isn't specified by the user, then it's also responsible for using it directly instead of getting it straight back from this function.

achilleas-k: We could also rename the function maybe?  `ResolveRef()` is the other function in the package that does the actual resolving.  This function is now essentially a wrapper around that which also runs refs through `VerifyRef()`.  We could call it `VerifyAndResolve()` but I hate that.

achilleas-k: Never mind, don't rename the function.  Also we can drop the `ref` from the return in a follow-up.  It's nbd.

test/cases/api.sh
```diff
@@ -193,9 +193,6 @@ function dump_db() {
   # Disable -x for these commands to avoid printing the whole result and manifest into the log
   set +x
 
-  # Make sure we get 3 job entries in the db per compose (depsolve + manifest + build)
-  sudo ${CONTAINER_RUNTIME} exec "${DB_CONTAINER_NAME}" psql -U postgres -d osbuildcomposer -c "SELECT * FROM jobs;" | grep "9 rows" > /dev/null
-
```
achilleas-k, line 198: Can this be replaced with the expected number or is it not fixed anymore because sometimes there's an ostree resolve job?  Could we make it conditional?

achilleas-k (approved):
Beautiful work.  I added some comments along the way but I don't think we need to do anything about them now.  Can address in a follow-up.

test/cases/api.sh
croissanne: So I wonder how useful this check was. Do we really care about how many jobs it takes to get to the end result?

Also for ostree jobs it's now 4 ; and for non-ostree it's 3.

croissanne:
I will merge this after https://github.com/osbuild/osbuild-composer/pull/3081 ; as deploying this PR will require a staggered approach (bump to commit no2 of this PR until all workers are redeployed -> bump to no3).

PR #3073, opened 2022-10-17
schutzbot/mockbuild: stop running mock as root

croissanne:
Running mock as root is broken, this commit can be reverted once this is fixed.

rhbz#2135203


jrusz (approved):
LGTM, I guess we don't even have to revert it if this works or is there any disadvantage here?

croissanne:
> LGTM, I guess we don't even have to revert it if this works or is there any disadvantage here?

We don't need to revert, it's purely preference afaik. In our case it really doesn't matter whether or not we run it as root.

PR #3004, opened 2022-09-20
Add `image_installer` type for Fedora.

supakeen:
This pull request is a takeover of @teg's old PR as I couldn't quite figure out how to properly do this workflow wise. I've redone those commits but have included the `!fixup`s by @ondrejbudai and other comments that were on the PR.

Tracking here now, the old PR is here: https://github.com/osbuild/osbuild-composer/pull/2878

PR #3077, opened 2022-10-18
image: create image-installer image type for fedora

supakeen:
This adds the `image-installer` image type which is similar to the RHEL image installer. If the version of Fedora is >= 38 then the anaconda-webui package is added to the package sets and the appropriate kernel options are set to boot into the webui.

This is the final form of: #3004 

Screenshot_2022-10-18_16-20-06
Screenshot_2022-10-18_16-20-26

## Questions

- Is checking the version in `manifest/iso_tree.go` appropriate to select kernel options?
- Should (or could) this image type be scoped to only Fedora >= 38?

## Installer

- Currently when the installer opens it is blinking, this seems to be caused by the installer getting restarted as it can't find a default locale (some related messages in logs), however using the same image type on Fedora 37 (and the 'old' anaconda-ui) does work correctly.

## Followup?

- `liveimg --url="file://live.tar.xz"` into `/usr/share/anaconda/interactive-defaults.ks` instead of the kickstart stage. Followup or in this PR?


teg:
I'm in favour of scoping to >=f38, so that we only ever support the new installer, no need to put effort into making the old one work if it is anyway going away (or to fix bugs in it etc).

teg:
I'd do a follow-up PR, so we are not blocked on osbuild release :)

supakeen:
Currently working on adjusting the kickstart stage to write to `/usr/share/anaconda/interactive-defaults.ks` and comparing the resulting ISO with the one I've gotten from the installer team to establish where the unknown locale shenanigans comes from.

supakeen:
This has now been moved to using `interactive-defaults.ks` to do so we need to write the file in `manifest.Anaconda`, there is now a new `InteractiveDefaults` property on the `Anaconda` struct to allow passing options to write to that kickstart file.

This did not resolve the constant reloading of the installer image, next steps will be to take a deeper look at anaconda-webui and why/how it is selecting the locale to use. It initially sets the locale to `en_US.UTF-8` but then switches it back to an empty locale.

supakeen:
> I'd do a follow-up PR, so we are not blocked on osbuild release :)

The installer team requested to use interactive-defaults.ks as automated install mode is untested in anaconda-webui, hence it has landed in this PR though through the general kickstart stage which removes the blocker :)

supakeen:
@ondrejbudai @achilleas-k this PR is now ready for review. As soon as RPMs are built I'll run our external CI to provide a testing ISO for the Anaconda team.

internal/image/image_installer.go
```diff
@@ -0,0 +1,152 @@
...
+	buildPipeline := manifest.NewBuild(m, runner, repos)
+	buildPipeline.Checkpoint()
+
+	version, err := strconv.Atoi(img.OSVersion)
+
+	if err != nil {
+		panic("cannot convert version to int: " + err.Error())
+	}
+
+	useWebUi := img.OSName == "fedora" && version >= 38
```
achilleas-k, line 60: This is the kind of thing we generally want to avoid outside of the actual distro code.  Basically anything that's in the `image` or `manifest` packages should be distro agnostic.  The distro code itself (in `distro/fedora/`) should understand distribution and version requirements and set options on the pipeline generation code accordingly.

In this case, the way I would do it would be to add a public field to the `ImageInstaller` struct called `AdditionalAnacondaModules` (or just `AnacondaModules`) so that any caller of `NewImageInstaller()` can attach extra modules based on the caller's knowledge of distro version requirements.

There's a similar situation with RHSM stages and options in the code I'm working on for RHEL: https://github.com/osbuild/osbuild-composer/pull/3120/commits/0fa1bca92e8cff0d1e263b37957ac625c8bf90e5

Before, we had RHSM options on the image config for the image types in `distro/rhel9/`, but those options would only be applied (i.e., the rhsm stage would only be added to the pipelines) if the distro was RHEL (i.e., not CentOS).  With the new separation of logic we have now though, the decision moves up into the distro code, so that the RHSM options are conditionally added to the image config and then the pipeline generation code that lives outside `distro/` adds RHSM stages if it's in the image config without needing to check any other conditions.

internal/image/image_installer.go
```diff
@@ -0,0 +1,152 @@
...
+	rootfsImagePipeline.Size = 4 * common.GibiByte
+
+	bootTreePipeline := manifest.NewEFIBootTree(m, buildPipeline, anacondaPipeline)
+	bootTreePipeline.Platform = img.Platform
+	bootTreePipeline.UEFIVendor = img.Platform.GetUEFIVendor()
+	bootTreePipeline.ISOLabel = isoLabel
+
+	if useWebUi {
+		bootTreePipeline.KernelOpts = []string{"inst.webui", "inst.webui.remote"}
+	}
```
achilleas-k, line 120: Same as above.  `KernelOpts` should be a field on `ImageInstaller` that is added externally by the caller.

achilleas-k (changes requested):
Thanks.  This looks good on first read but I'd like to read through again tomorrow because I'm (also) still not 100% used to the new framework.

My comment in code below is a bit important though.  It's not an issue functionally, but these conventions are a big part of the reason for the redesign.  It's also something I want to make more explicit in code going forward so that the separation of concerns is self-evident.

internal/image/image_installer.go
supakeen: I've rewritten this to be placed into the `distro/fedora/images.go` now by adding `AdditionalKernelOpts` and `AdditionalAnacondaModules` on the `image_installer.go` struct(s).

supakeen:
So I've moved customizations to image installer one level up so they now live in `distro/fedora/images.go` instead of inside `image/image_installer.go`. However, I still don't like having any conditionals like these.

@ondrejbudai What are your thoughts on where these conditionals should live? We could introduce the `image-installer-preview` image type which would only be attached to Fedora 38 in `newDistro` instead.

Or is it perhaps not that weird (feels weird to me) to have image types do entirely different things based on the distro version? :)

achilleas-k:
> @ondrejbudai What are your thoughts on where these conditionals should live? We could introduce the `image-installer-preview` image type which would only be attached to Fedora 38 in `newDistro` instead.

FTR I wouldn't be against a separate image type.  It could share most of the code and config values of the `image-installer` so that they don't accidentally diverge in places they don't need to.

ondrejbudai:
Yeah, two image types don't sound awful for this kind of situation. I'm fine with that.

internal/manifest/iso_tree.go
```diff
@@ -159,21 +192,29 @@ func (p *ISOTree) serialize() osbuild.Pipeline {
...
-	if err != nil {
-		panic("password encryption failed")
+	if p.OSTree != nil {
+		pipeline.AddStage(osbuild.NewOSTreeInitStage(&osbuild.OSTreeInitStageOptions{Path: ostreeRepoPath}))
+		pipeline.AddStage(osbuild.NewOSTreePullStage(
+			&osbuild.OSTreePullStageOptions{Repo: ostreeRepoPath},
+			osbuild.NewOstreePullStageInputs("org.osbuild.source", p.OSTree.Checksum, p.OSTree.Ref),
+		))
+
+		kickstartOptions, err := osbuild.NewKickstartStageOptions(p.KSPath, "", p.Users, p.Groups, p.OSTree.Ref, makeISORootPath(ostreeRepoPath), p.OSName)
```
achilleas-k, line 202: The ostree ref and repo path are flipped here.

test/data/manifests/fedora_38-x86_64-iot_installer-boot.json
```diff
@@ -11080,8 +11072,8 @@
               "path": "/ostree.ks",
               "ostree": {
                 "osname": "fedora",
-                "url": "file:///run/install/repo/ostree/repo",
-                "ref": "test/iot",
+                "url": "test/iot",
+                "ref": "file:///run/install/repo/ostree/repo",
```
achilleas-k, line 11076: The flipped values

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4df301ae9bcbf4f06ef1a4872efff420397ac822 with the main merge-base 587e0436681b74ecb359945066ae63afeba9085a).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3325354576/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3325354576/artifacts/browse

achilleas-k:
Rebased for Packer fix.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 2e39f336ddf3a6e55adade061c8aa2f746cdf1e6 with the main merge-base 28319f50d8e3411996f72455117643b28f112157).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3340198989/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3340198989/artifacts/browse

achilleas-k:
```
    --- FAIL: TestManifests/fedora_37-aarch64-image_installer-boot.json (21.54s)
panic: Could not depsolve: DNF error occurred: MarkingErrors: Error occurred when marking packages for installation: Problems in request:
missing packages: sil-scheherazade-new-fonts [recovered]
	panic: Could not depsolve: DNF error occurred: MarkingErrors: Error occurred when marking packages for installation: Problems in request:
missing packages: sil-scheherazade-new-fonts
```
???

achilleas-k:
Did a rebase again to fix the font package issue and changed the commits a bit so that the argument flipping doesn't happen.  The second commit doesn't have any manifest changes now and the `manifests.plain/` directory is removed so all manifests are in `test/data/manifests/`

achilleas-k (dismissed):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD e86a39c7893d9430b95cf915c417c9ad98cc9c90 with the main merge-base a1a3e5ba2d8eac64ba2d6121bb48b463f8b9068f).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3347621852/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3347621852/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 3dbcf971ef4ea45fc7b34dcae15bf6a835bd876d with the main merge-base a1a3e5ba2d8eac64ba2d6121bb48b463f8b9068f).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350700912/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350700912/artifacts/browse

achilleas-k (approved):

ondrejbudai (approved):
Amazing!

PR #3082, opened 2022-10-19
schutzbot/update_github_status: fix release fast-forwarding

croissanne:
The checkout in the working directory is a shallow clone with gitlab as a remote. As a result fast-forwarding failed due to not recognizing the remote release branch as an ancestor of $CI_COMMIT_SHA.

This clones (with `--bare`) the github remote in a separate directory. It should not be necessary to remove the release-ff-clone directory each time, simply fetch if it already exists.


ondrejbudai (approved):
Looks good.

PR #3084, opened 2022-10-19
osbuild: add masked services to systemd stage

achilleas-k:
The feature was included in the osbuild stage and was never added to composer.


gicmo (approved):
❤️

PR #3090, opened 2022-10-20
Fix loading cross distro compose results

bcl:
Fixes a problem with the store load not being able to load composes for other distros when the image type isn't supported on the host distro.

This pull request includes:


achilleas-k (commented):
This is great.  Thought it would require more work but completely missed that we can just read the distro key from the blueprint stored with the compose :)

The rpm build are failing on a test that _seems_ relevant, but all the gitlab CI tests are fine, which is weird.

achilleas-k:
Did an auto-rebase on main.

bcl:
ppc64le and s390x on Fedora don't have the same image types so the update test was failing. Forcing the host arch to x86_64 like it did previously should fix that for now, but that test could use some work at some point in the future.

ondrejbudai (commented):
Love it!

achilleas-k (dismissed):

PR #3093, opened 2022-10-24
Cloud API: make `location` optional for Azure Upload Options

thozza:
Previously, the Cloud API implementation required that the client provides the Location of the Resource Group to be used for importing the image to Azure. This forced the client to provide a redundant information, that it does not necessarily have. In addition, the Resource Group Location can be easily determined using Azure API.

Let's change the implementation and make the `location` property optional when uploading images to Azure. If the `location` is not provided, osbuild-composer will deduct the appropriate location from the provided Resource Group.

Change the Cloud API test case for Azure to not set the `location` property when calling Cloud API.

Related to https://issues.redhat.com/browse/HMSIB-133

This pull request includes:


thozza:
The PR should be ready for review.

- The c9s build issue is caused by `vendor/github.com/containers/libtrust/ec_key_openssl.go:23:15: undefined: ecdsa.HashSign` and should be resolved when a [new version of the `golang` package](https://kojihub.stream.centos.org/koji/buildinfo?buildID=25450) gets to c9s repos.
- F37 failures in CI are caused by our CI using old repo snapshots and installing old `podman` but new `containers-common` which results in `crun` not being installed as a dependency.

croissanne (dismissed):
Perfect, thank you!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -917,7 +917,6 @@ components:
         - tenant_id
         - subscription_id
         - resource_group
-        - location
```
ondrejbudai, line 920: Can you document on the line 943 what happens when the location isn't defined? (The location is determined from the RG.)

ondrejbudai (changes requested):
Amazing! Just one small request for the API spec.

internal/cloudapi/v2/openapi.v2.yml
thozza: sure

thozza: done

ondrejbudai (approved):

PR #3111, opened 2022-11-02
tools/provision.sh: copy RHEL repo overrides using wildcard

thozza:
Don't list specific RHEL repos which should be copied over to the repo overrides directory, but instead copy all of them, similar to how it's done for Fedora and CentOS Stream. This is less error prone when support for a new RHEL release is added to composer.

This was forgotten as part of enabling RHEL 8.8 and 9.2.

This pull request includes:

### DIFF
--- a/tools/provision.sh
+++ b/tools/provision.sh
@@ -140,11 +140,7 @@ else # AUTH_METHOD_NONE
     # Copy all fedora repo overrides
     sudo cp -a /usr/share/tests/osbuild-composer/repositories/{fedora,centos}-*.json "$REPODIR"
     # Copy RHEL point release repos
-    sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-85.json "$REPODIR"
-    sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-86.json "$REPODIR"
-    sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-87.json "$REPODIR"
-    sudo cp /usr/share/tests/osbuild-composer/repositories/rhel-90.json "$REPODIR"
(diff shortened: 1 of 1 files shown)


bcl:
@atodorov should have a look at this. He did a bunch of cleanup related to this, and I think that this change would add the non-point releases. eg. rhel-8.json and rhel-8-beta.json 

There's a commit message for commit 860b122f2c6d1bfa6e5205f5492f882343a58e31 pointing to https://github.com/osbuild/osbuild-composer/pull/2881/files#r940065803 as the explanation but the explanation seems to be missing.

thozza:
> @atodorov should have a look at this. He did a bunch of cleanup related to this, and I think that this change would add the non-point releases. eg. rhel-8.json and rhel-8-beta.json
> 
> There's a commit message for commit [860b122](https://github.com/osbuild/osbuild-composer/commit/860b122f2c6d1bfa6e5205f5492f882343a58e31) pointing to https://github.com/osbuild/osbuild-composer/pull/2881/files#r940065803 as the explanation but the explanation seems to be missing.

I think that it was https://github.com/osbuild/osbuild-composer/pull/2881#discussion_r940065803, which disappeared from the diff after rebasing.

There is already no `*-beta.json` file in `test/data/repositories`.  The code will currently default to picking `rhel-XY.json` for the RHEL X.Y that it is running on (for 8.4 and above). So the impact of copying the non-point release repo to overrides is IMO none.

bcl:
> There is already no `*-beta.json` file in `test/data/repositories`.

Ah, ha! I was looking in `./repositories/` not `test/data/repositories/`

achilleas-k (approved):

atodorov (approved):

atodorov (approved):

PR #3116, opened 2022-11-03
New conn for each listen

croissanne:


kingsleyzissou (dismissed):

kingsleyzissou (approved):

croissanne:
> Out of curiosity, how did this manifest in practice? Were we seeing failures in prod due to this? Were they captured by out SLOs?

Not really, there were generally enough pods available to keep it going.

teg:
> Not really, there were generally enough pods available to keep it going. Just a lot of logging statements.

Thanks for your attention to detail and looking at all the things!

Issue #2954, opened 2022-09-02
`SELinux` member in the `manifest.OSCustomizations` is redundant (to `SELinuxConfig`)


thozza:
**Describe the bug**
The `SELinux` member in `manifest.OSCustomizations` is redundant (to `SELinuxConfig`):
https://github.com/osbuild/osbuild-composer/blob/b03a131f1322401fda8c9e5de490ce7dda0f5668/internal/manifest/os.go#L61-L63

The value is intended to hold the Selinux policy name, but that is already part of another member variable of the structure. Specifically the `SELinuxConfig  *osbuild.SELinuxConfigStageOptions`, which type contains the same information under `Type`:
https://github.com/osbuild/osbuild-composer/blob/b03a131f1322401fda8c9e5de490ce7dda0f5668/internal/osbuild/selinux_config_stage.go#L5-L8 

As a result, some files and the image filesystem may be relabeled using a different SELinux policy, than the one specified in the `ImageConfig`. The `SELinux` value defaults to `targeted` if not set, but the `SELinuxConfig.Type` can be set to `minimum` or `mls` at the same time.

**Expected behavior**
The `SELinux` member should be IMO removed from the `manifest.OSCustomizations` and the following code blocks should use the `SELinuxConfig.Type` value instead:

https://github.com/osbuild/osbuild-composer/blob/b03a131f1322401fda8c9e5de490ce7dda0f5668/internal/manifest/os.go#L154-L156
https://github.com/osbuild/osbuild-composer/blob/b03a131f1322401fda8c9e5de490ce7dda0f5668/internal/manifest/os.go#L189-L192
https://github.com/osbuild/osbuild-composer/blob/b03a131f1322401fda8c9e5de490ce7dda0f5668/internal/manifest/os.go#L430-L434

CC @teg

PR #3120, opened 2022-11-07
Rewrite RHEL 9 and CS9 image definitions using the new framework

achilleas-k:
This is a rewrite of the `distro/rhel9/` package to use the new image definition framework that was so far used only for Fedora.
Some changes are not final but will probably be addressed in a follow-up rather than here.

A rewrite of `distro/rhel8/` is also planned, but will be done separately.  That should require less work after everything that went into this one.

I've described and provided context for all changes in individual commit messages.

Remaining tasks:
- [x] image installer
- [x] tar
- [x] change edge tests to check for `sysroot=ro` on RHEL 9 and CS9
- [ ] ~~edge simplified installer~~ (will be handled separately)

### Updates
1. The edge simplified installer will be done in a follow-up.
https://github.com/osbuild/osbuild-composer/pull/3120#issuecomment-1323593242

2. Changes to installer definitions.
https://github.com/osbuild/osbuild-composer/pull/3120#issuecomment-1324116599

3. Status of sysroot readonly for RHEL 9 and CS9 edge images.
https://github.com/osbuild/osbuild-composer/pull/3120#issuecomment-1324407557


achilleas-k:
@lavocatt Is there a way to get manifest-db to run image info on all of these and check for changes?  This is one of the cases where that would be perfect since all the manifests have changed slightly but the image info should be identical.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4ef0386714113781cdd0b4c66955e9512e4fb5bf with the main merge-base 63f1b8d9fbdb5961f2fe3a953b012abb229703c1).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3286299953/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3286299953/artifacts/browse

lavocatt:
> @lavocatt Is there a way to get manifest-db to run image info on all of these and check for changes? This is one of the cases where that would be perfect since all the manifests have changed slightly but the image info should be identical.

Not yet, because it fetches the last manifests on top of composer's main. I could hack something for you to specify a branch name for composer in a variable on the trigger workflow ?

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD bf08e2e3ea2fbe353ee532e5dfa4decef8ba3f52 with the main merge-base d73171fdde5be7d131313e599cb1d74f6d128403).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3297215219/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3297215219/artifacts/browse

internal/runner/rhel.go
```diff
@@ -13,7 +13,8 @@ func (r *RHEL) String() string {
 
 func (p *RHEL) GetBuildPackages() []string {
 	packages := []string{
-		"glibc", // ldconfig
+		"glibc",           // ldconfig
+		"platform-python", // osbuild
```
thozza, line 17: I'm really sure this will work due to https://bugzilla.redhat.com/show_bug.cgi?id=2121775

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 974636418207c4b987e1bad3c10b174724e00afb with the main merge-base 4e7b021514ee416dc435f9617326f4f15bce4c27).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3311120109/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3311120109/artifacts/browse

internal/distro/rhel9/gce.go
```diff
@@ -0,0 +1,290 @@
...
+		bootType:            distro.UEFIBootType,
+		defaultSize:         20 * common.GibiByte,
+		pipelines:           gcePipelines,
+		buildPipelines:      []string{"build"},
+		payloadPipelines:    []string{"os", "image", "archive"},
+		exports:             []string{"archive"},
+		basePartitionTables: defaultBasePartitionTables,
+	}
+
+	// default AMI (EC2 BYOS) images config
```
thozza: The comment is wrong "GCE" vs. "AMI"

internal/distro/rhel9/imagetype.go
```diff
@@ -327,6 +390,65 @@ func (t *imageType) Manifest(customizations *blueprint.Customizations,
...
+			globalRepos = append(globalRepos, repo)
+		}
+	}
+
+	// create a manifest object and instantiate it with the computed packageSetChains
+	manifest, err := t.initializeManifest(&bp, options, globalRepos, packageSets, nil, 0)
+	if err != nil {
+		// TODO: handle manifest initialization errors more gracefully, we
+		// refuse to initialize manifests with invalid config.
+		return nil
```
thozza, line 460: you should at least print the error here (as done in #3132)

internal/distro/rhel9/imagetype.go
```diff
@@ -273,13 +287,62 @@ func (t *imageType) PartitionType() string {
...
+	}
+
+	// TODO: let image types specify valid workloads, rather than
+	// always assume Custom.
+	w := &workload.Custom{
+		BaseWorkload: workload.BaseWorkload{
+			Repos: packageSets[blueprintPkgsKey].Repositories,
+		},
+		Packages: bp.GetPackagesEx(false),
+	}
```
thozza, line 311: Note: We will actually have to implement a way to merge workloads. The reason is that if we define SAP as a workload, instead of an image type, then we would still have to have a way for user to customize image with SAP workload with potentially additional packages, enabled services, etc... Also the complication is that we now depsolve packages in chain where the workload packages are always depsolved as a 2nd transaction, which may not be desirable for "standard" workloads like SAP, but really only for packages provided by the user in BP. Probably not that relevant for your rework, but this will have to be figured out before we use Workloads for anything else than BP customizations.

thozza (commented):
Just submitting comments from my partial review... This is huge so more to come 😆 (note that I was adding comments as I was going through separate commits)

internal/manifest/os.go
```diff
@@ -176,14 +193,30 @@ func (p *OS) getPackageSetChain() []rpmmd.PackageSet {
...
+		packages = append(packages, p.PartitionTable.GetBuildPackages()...)
+	}
 	packages = append(packages, "rpm")
 	if p.OSTreeRef != "" {
 		packages = append(packages, "rpm-ostree")
 	}
 	if p.SElinux != "" {
-		packages = append(packages, "policycoreutils")
-		packages = append(packages, fmt.Sprintf("selinux-policy-%s", p.SElinux))
+		packages = append(packages, "policycoreutils", fmt.Sprintf("selinux-policy-%s", p.SElinux))
```
thozza, line 204: Note that this is actually wrong (https://github.com/osbuild/osbuild-composer/issues/2954). We should take the policy specified in the image configuration into account. I think that those two values should be eventually merged and that's actually the plan (as to move away from osbuild-specific data structures). Just pointing out that this currently uses hardcoded value.

tools/test-case-generators/format-request-map.json
```diff
@@ -717,6 +734,92 @@
       }
     },
     "overrides": {
+      "rhel-91": {
```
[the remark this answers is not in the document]
achilleas-k: They will, I was just too lazy to copy-paste the overrides for all since they use the same pipelines, it should be fine.  When RHEL 8 gets the feature, we can remove the override and make it part of the standard `with-customizations` variant.

internal/runner/rhel.go
achilleas-k: I realised this was missing because I tried to build all the manifests that changed and they were failing immediately.  Adding this was enough for RHEL 9.  I guess if we find that we need `python3` for RHEL 8 we can add that too.

internal/manifest/os.go
achilleas-k: ah good to know, thanks.  I was just merging the two lines here, but good that you noticed so we can fix it now if it's not too big a change.

tools/test-case-generators/format-request-map.json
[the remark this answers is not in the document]
thozza: I see. It would be probably much easier if we could use wildcards in the distro name in `format-request-map.json`...

internal/runner/rhel.go
thozza: Interesting. I though (based on the bug report) that osbuild stages are actually not able to use `platform-python` due to hardcoding the binary path in the shebang. So installing it should not help much 🤔 I didn't investigate this more at all so I may be all wrong...

internal/manifest/os.go
thozza: I think that it is fine to use this as is for the first rework, but we should make sure to not introduce any regressions in RHEL images before the development window for 9.2 and 8.8 closes...

internal/runner/rhel.go
achilleas-k: The [RHEL 8.x runners](https://github.com/osbuild/osbuild/blob/f34bee944bdfc590cf25159521ee66c515d0cf10/runners/org.osbuild.rhel82#L1) use `platform-python`, so that's all we need in the build root.

I think the bug report is a case of removing the dependency from the host and not being able to bootstrap the build root.

thozza: Thanks for the additional info. This is very helpful!

However I'm afraid that it is not completely true, since `python36` gets pulled into the buildroot pipeline and it is installed there. So while you are right that the bug is caused by the situation that you've described, AFAICT osbuild stages still use the `python3` installed in the buildroot, because they are executed in a separate process without an explicit interpreter path provided in the [`subprocess.run()` call](https://github.com/osbuild/osbuild/blob/f34bee944bdfc590cf25159521ee66c515d0cf10/runners/org.osbuild.rhel82#L76-L78).

gicmo: Our current RHEL 8 (and thus 9) runner also has an implicit dependency on `python3.6`:

https://github.com/osbuild/osbuild/blob/f34bee944bdfc590cf25159521ee66c515d0cf10/runners/org.osbuild.rhel82#L54-L63

thozza: That's exactly the point that I'm trying to make... Therefore the buildroot packages should explicitly list `python36` and not rely on it being pulled in as a dependency of something else that is installed in the buildroot...

gicmo: See also https://bugzilla.redhat.com/show_bug.cgi?id=2121775

achilleas-k: So do we add `python36` to the runner deps for RHEL to be explicit?

thozza: So it looks like it is pulled into the DNF transaction by some other package, so specifying it in the build packages will have basically no real effect.

Personally I suggest to add it there to make the dependency explicit and prevent any unexpected issues due to `python36` not being installed. It should be there until the rhel8 runner is "fixed" to depend only on `platform-python`.

Including a comment about the whole situation explaining two pythons being listed would be probably also good, but I'm pretty sure you would add something like this anyway 😉

achilleas-k: Done!

internal/distro/rhel9/gce.go
achilleas-k: Oh, it was deleted in a followup commit after the copy/paste to multiple files.  I've cleared it from that commit as well.

internal/distro/rhel9/imagetype.go
achilleas-k: Done!

internal/distro/rhel9/imagetype.go
teg: My thinking with workloads was that we would not expose the same options in SAP as Custom, but only those options that make sense in that setting. I imagine that for an SAP image we would not support adding extra packages, systems services etc, maybe we would not support any options at all, or just a few SAP specific ones. The benefit of the named workloads would be that we know the use case so can be more restrictive.

thozza: hmm, I see. It is definitely an option. If that would be the case, then things could stay as they are. However it would also mean regressing in our current functionality, because we support customization of any image type, even those that are focused on specific workload.

So as a result, I expected that we want to allow building custom images of any type. I can imagine that the user may need to install some additional packages and services also into predefined-workload-specific images for monitoring, inspection of the system, etc. E.g. also we do forward logs to Splunk, the customer's security team may require a special software to be installed on the OS for CVE and threat assessment.

With the SAP image as an example, it is preconfigured for SAP, but does not come with SAP software. Maybe someone would want to build custom SAP images with the actual SAP software installed from custom repos 🤔 I'm not actually sure, just describing situations in which it would make sense to allow (package and service) customizations. The workload-specific customizations are definitely an interesting idea and IMO would make sense.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 37588b7baeb85b6fdd660f40b8a51942d6deb319 with the main merge-base 587e0436681b74ecb359945066ae63afeba9085a).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3325096739/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3325096739/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 5ce61f1c188f69a4c67e891282cab8a85eea8cd9 with the main merge-base 587e0436681b74ecb359945066ae63afeba9085a).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3327785754/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3327785754/artifacts/browse

internal/image/live.go
```diff
@@ -79,6 +79,9 @@ func (img *LiveImage) InstantiateManifest(m *manifest.Manifest,
 		artifactPipeline = vmdkPipeline
 		artifact = vmdkPipeline.Export()
 	case platform.FORMAT_GCE:
+		// NOTE(akoutsou): temporary workaround; filename required for GCP
+		// TODO: define internal raw filename on image type
+		imagePipeline.Filename = "disk.raw"
```
thozza, line 84: Yeah, the filename really matters. Requirements are defined in https://cloud.google.com/compute/docs/import/import-existing-image#requirements_for_the_image_file. You probably plan it anyway, but I suggest to merge this fix and the regenerated GCE manifests with the previous commit that changes the pipeline used for GCE images and the commit which regenerated GCE manifests (It contained a commit message that the filename does not matter, which would be confusing (because it does).

achilleas-k: yeah thanks.  I knew GCP had strict requirements for the archive but forgot that the internal filename was one of them until I saw the test failing.

Rebased and squashed.

internal/distro/rhel9/imagetype.go
teg: I agree, I also don't know for certain. It could be that we'll need a lot (all?) the customisations for some specific workloads, but I think we should have the discussion on a case-by-case basis on what makes sense, to limit our test exposure etc.

Do we already expose SAP images with customisations anywhere? I thought they were only built internally?

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 3da969933ec02ac3b9b36f7af50134b781e7f61b with the main merge-base 422281698a4586f4fc6d27e159a4d522732779ab).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3336132963/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3336132963/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 40062ba4a370ca6236b699ad6ba26e523ca0dbfa with the main merge-base 28319f50d8e3411996f72455117643b28f112157).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3340554461/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3340554461/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 35fbbffe27aa3626cc15213ab13552dadf7e2bbb with the main merge-base 28319f50d8e3411996f72455117643b28f112157).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3341034131/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3341034131/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD e8cd40de856b2812f259f8be85b192219af72647 with the main merge-base 28319f50d8e3411996f72455117643b28f112157).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3342826800/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3342826800/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c34b08b59f2691828d94f992ab278757f41a2653 with the main merge-base 9ea58d1486d3bed0be7658f21c49abb2019093c7).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3347790966/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3347790966/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 2860cf6667035f781a5efbf2d3d25a1a22b6e2a5 with the main merge-base 9ea58d1486d3bed0be7658f21c49abb2019093c7).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350701412/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350701412/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 9e37d982fde0107f32ab1cd96e8f1a2a928e7029 with the main merge-base 9ea58d1486d3bed0be7658f21c49abb2019093c7).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350947740/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3350947740/artifacts/browse

internal/distro/rhel9/imagetype.go
thozza: > Do we already expose SAP images with customisations anywhere? I thought they were only built internally?

The image type is filtered out by default via Weldr API. With a configuration change and adding a few new repo definitions for SAP-specific packages would definitely make it possible to build it on-premise. Nevertheless this is not really documented anywhere and would require knowing what to do. So we can IMO safely consider the SAP image in general not being buildable on-premise.

For Clous API (including the internal service), any customizations can be already passed when building the SAP image (as well as any other image). AFAIK it is not really used by anyone nor by any pipeline. My point was that it is possible, not that it is used.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 6471d9dd8de54690117f43953a2694d88e18d709 with the main merge-base 44f4225c02d1a82f177dcd5e69bec0b12545b8dc).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3358719337/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3358719337/artifacts/browse

achilleas-k:
> Remaining tasks:
> 
> * [x]  image installer
> * [ ]  edge simplified installer
> * [x]  tar
> * [x]  change edge tests to check for `sysroot=ro` on RHEL 9 and CS9

The edge simplified installer will be done in a follow-up.

achilleas-k:
### Ready for review!

#### Some notes on the most recent changes.

I've made some changes to the new installer definitions that were added in #3077.

I added options so that each distro can configure things a little differently when it comes to the kickstart file and the squashfs compression. 

For the kickstart file: Fedora uses the default kickstart location (interactive-defaults) but in RHEL so far we put the kickstart file in the root of the ISO at `/osbuld.ks`.  For now, I decided we should maintain the old behaviour for RHEL 9 and CS9 and make changes later.  This would require changing tests and possibly docs, so the change should be more isolated and purposeful.

For the squashfs compression: I couldn't get the RHEL 9.2 ISO to boot with lz4 compression, which is used in Fedora.  So far we used xz compression in RHEL and CS, so again I'm keeping the old behaviour.  If we need to change it, we can figure out what the issue is and make the change.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 453f4e9f79148f3d2bacbf3c0246fbc4db0473c5 with the main merge-base dc3c0d9725acea76e53c306a4ffd06f17e02fa46).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3364576798/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3364576798/artifacts/browse

achilleas-k:
> * [x]  change edge tests to check for `sysroot=ro` on RHEL 9 and CS9

This was reverted.  Since anaconda doesn't set `sysroot` to read only in RHEL 9, our raw image would be inconsistent with installer deployments.  We should make the change in RHEL when anaconda does the same.

internal/distro/rhel9/imagetype.go
teg: Thanks for the context!

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 51b91e5d8bc3c6ca6043062b4c0c5787c7148123 with the main merge-base ccb3a67d94e28561ba064aef1f4f35fc63362046).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3373529785/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3373529785/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 2a43d662f8657d8913cdb40d29c3715decb253f4 with the main merge-base a295ed7c90cad7129512c2b18fbe8b51b22e2d80).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3373743157/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3373743157/artifacts/browse

internal/manifest/tar.go
```diff
@@ -10,24 +10,27 @@ type Tar struct {
...
+	inputPipeline Pipeline,
 	pipelinename string) *Tar {
 	p := &Tar{
 		Base:          NewBase(m, pipelinename, buildPipeline),
 		inputPipeline: inputPipeline,
 		Filename:      "image.tar",
 	}
-	if inputPipeline.manifest != m {
-		panic("tree pipeline from different manifest")
-	}
```
teg, line 30: Should this be replaced by anything? Again, not worth a rebase.

internal/distro/rhel9/images.go
```diff
@@ -346,6 +346,9 @@ func imageInstallerImage(workload workload.Workload,
 
 	img.SquashfsCompression = "xz"
 
+	// put the kickstart file in the root of the iso
+	img.ISORootKickstart = true
```
teg, line 351: What's the ideal? I remember there was a discussion with the installer team, but forgot the outcome. If not for compat, where should we be putting the .ks?

tools/test-case-generators/format-request-map.json
```diff
@@ -490,6 +490,12 @@
       "filename": "installer.iso",
       "blueprint": {
         "name": "image-installer-users",
+        "packages": [
+          {
+            "name": "podman",
```
teg, line 495: Why?

teg (dismissed):
Love all of this!
I threw in a bunch of comments/questions, but regardless of the answers, please address anything in follow-ups.

internal/distro/rhel9/distro.go
```diff
@@ -640,41 +620,61 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
...
+func NewRHEL92() distro.Distro {
+	return newDistro("rhel", 9, 2)
+}
+
+func NewRHEL92HostDistro(name, modulePlatformID, ostreeRef string) distro.Distro {
+	// NOTE: args are ignored - host distro constructors are deprecated
+	return NewRHEL92()
+}
+
+func newDistro(name string, major, minor int) distro.Distro {
```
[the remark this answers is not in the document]
achilleas-k: Yes.  Good point!

internal/distro/rhel9/distro.go
```diff
@@ -237,6 +237,7 @@ func newDistro(name string, major, minor int) distro.Distro {
 		bootType: distro.LegacyBootType,
 	}
 
+	qcow2ImgType := mkQcow2ImgType(rd)
```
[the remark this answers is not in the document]
achilleas-k: No, it's right.  The problem is that the local `qcow2ImgType` here shadows the package global one which doesn't have an image config.  Very error prone.

internal/distro/rhel9/distro.go
```diff
@@ -327,10 +326,20 @@ func newDistro(name string, major, minor int) distro.Distro {
 			BasePlatform: platform.BasePlatform{
 				ImageFormat: platform.FORMAT_RAW,
 			},
-			BIOS:       false,
+			BIOS:       true,
```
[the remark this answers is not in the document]
achilleas-k: Diff artefact.  The actual change is splitting one platform into two:

From
```go
	x86_64.addImageTypes(
		&platform.X86{
			BasePlatform: platform.BasePlatform{
				ImageFormat: platform.FORMAT_RAW,
			},
			BIOS:       false,
			UEFIVendor: rd.vendor,
		},
		edgeRawImgType,
		edgeSimplifiedInstallerImgType,
	)

```

To
```go
	x86_64.addImageTypes(
		&platform.X86{
			BasePlatform: platform.BasePlatform{
				ImageFormat: platform.FORMAT_RAW,
			},
			BIOS:       true,
			UEFIVendor: rd.vendor,
		},
		edgeRawImgType,
	)

	x86_64.addImageTypes(
		&platform.X86{
			BasePlatform: platform.BasePlatform{
				ImageFormat: platform.FORMAT_RAW,
			},
			BIOS:       false,
			UEFIVendor: rd.vendor,
		},
		edgeSimplifiedInstallerImgType,
	)
```

The raw image got moved to a new platform that supprts BIOS.

internal/manifest/tar.go
achilleas-k: Yes it's unfortunate that we lost the sanity checks but we could figure out a new way to do it.

internal/distro/rhel9/images.go
achilleas-k: For this PR, I'm making sure to not make any changes to image definitions, but we should think about this and decide what we want.

internal/platform/ppc64le.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (p *PPC64LE) GetArch() Arch {
+	return ARCH_PPC64LE
+}
+
+func (p *PPC64LE) GetBIOSPlatform() string {
+	if p.BIOS {
+		return "powerpc-ieee1275"
+	}
+	return ""
```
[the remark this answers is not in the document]
nullr0ute: BIOS is an x86 term, the POWER firmware is OpenFirmware based, so depends on the context.

tools/test-case-generators/format-request-map.json
achilleas-k: 1. Why not?
2. Good to have a package in there to see that it goes in the right place (payload packages), because the package set handling for the installer is a little different than other image types.  The Fedora installer in main right now ignores user/blueprint packages.

internal/platform/ppc64le.go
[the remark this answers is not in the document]
achilleas-k: We generally (and sloppily) use BIOS to mean "non-UEFI".

internal/distro/rhel9/distro.go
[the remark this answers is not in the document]
gicmo: Tripped over this too, it is fine! ;)

internal/distro/rhel9/images.go
gicmo: https://hackmd.io/hwMbx1CXQRKvHXGW8yQTMg

gicmo (dismissed):
❤️ 
Went through it commit by commit. Found one minor thing, but I am not telling because it is not a big deal. ;)

internal/distro/rhel9/images.go
```diff
@@ -111,6 +111,9 @@ func osCustomizations(
 			},
 		)
 	}
+	if imageConfig.Keyboard != nil && imageConfig.Keyboard.X11Keymap != nil {
+		osc.X11Layouts = imageConfig.Keyboard.X11Keymap.Layouts
+	}
```
thozza, line 130: This should be moved to the block above:
```go
if keyboard != nil {
	osc.Keyboard = keyboard
} else if imageConfig.Keyboard != nil {
	osc.Keyboard = &imageConfig.Keyboard.Keymap
}
```
The whole config (keymap and x11layouts) are currently applied as a whole - https://github.com/osbuild/osbuild-composer/blob/3443484502b448fd00c5596ab546c190be781518/internal/distro/rhel9/pipelines.go#L420-L424

Lastly, the `X11Keymap.Layouts` tend to be related to the value set in `Keyboard`.

achilleas-k:
Rebased and fixed merge conflicts

thozza (commented):
Ended up in a half, submitting what I found so far...

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 91aa9469eb2dab8b710db504bb52e79d71e0a58d with the main merge-base 3443484502b448fd00c5596ab546c190be781518).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3376499017/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3376499017/artifacts/browse

internal/distro/rhel9/qcow2.go
```diff
@@ -158,3 +143,31 @@ func openstackCommonPackageSet(t *imageType) rpmmd.PackageSet {
 
 	return ps
 }
+
+func qcowImageConfig(d distribution) *distro.ImageConfig {
+	ic := &distro.ImageConfig{
+		DefaultTarget: common.StringToPtr("multi-user.target"),
+	}
+	if d.isRHEL() {
```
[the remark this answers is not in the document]
teg: We would have to make it conditional on subscription options being enabled.

internal/distro/rhel9/azure.go
```diff
@@ -30,7 +24,7 @@ var (
 		kernelOptions:       defaultAzureKernelOptions,
 		bootable:            true,
 		defaultSize:         4 * common.GibiByte,
-		pipelines:           vhdPipelines(false),
+		image:               liveImage,
```
gicmo, line 27: I think this commit also switches the azure image type over to the new style, but the commit message doesn't say so. Just in case you are changing something in the PR anyway.

internal/runner/centos.go
```diff
@@ -12,12 +12,24 @@ func (c *CentOS) String() string {
...
 		)
 	}
+	if c.Version < 9 {
+		packages = append(packages,
+			// The RHEL 8 runner (which is also used for CS8) in osbuild runs
+			// with platform-python but explicitly symlinks python 3.6 to
+			// /etc/alternatives (which in turn is the target for
+			// /usr/bin/python3) for the stages.
+			// https://github.com/osbuild/osbuild/blob/ea8261cad6c5c606c00c0f2824c3f483c01a0cc9/runners/org.osbuild.rhel82#L61
+			// Install python36 explicitly for CS8.
```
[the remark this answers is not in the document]
achilleas-k: Yeah it seems we need it for both.

internal/runner/rhel.go
```diff
@@ -13,12 +13,23 @@ func (r *RHEL) String() string {
...
 			"systemd", // systemd-tmpfiles and systemd-sysusers
 		)
 	}
+	if p.Major < 9 {
+		packages = append(packages,
+			// The RHEL 8 runner in osbuild runs with platform-python but
+			// explicitly symlinks python 3.6 to /etc/alternatives (which in turn
+			// is the target for /usr/bin/python3) for the stages.
+			// https://github.com/osbuild/osbuild/blob/ea8261cad6c5c606c00c0f2824c3f483c01a0cc9/runners/org.osbuild.rhel82#L61
+			// Install python36 explicitly for RHEL 8.
```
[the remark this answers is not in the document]
achilleas-k: Yeah it seems we need it for both.

internal/distro/rhel9/qcow2.go
```diff
@@ -158,3 +143,31 @@ func openstackCommonPackageSet(t *imageType) rpmmd.PackageSet {
...
+				},
+			},
+		}
+
+	}
+	return ic
+}
+
+func mkQcow2ImgType(d distribution) imageType {
+	it := qcow2ImgType
```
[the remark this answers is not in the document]
achilleas-k: Yes, this came up here: https://github.com/osbuild/osbuild-composer/pull/3120#discussion_r1031490462
I wasn't going to change it if no other changes were necessary, but I'll do it now since it can lead to problems.

internal/distro/rhel9/ami.go
```diff
@@ -491,38 +403,101 @@ func rhelEc2SapPackageSet(t *imageType) rpmmd.PackageSet {
...
+	it := amiImgTypeX86_64
+	ic := defaultAMIImageConfigX86_64(osVersion, rhsm)
+	if rhsm && common.VersionLessThan(osVersion, "9.1") {
+		// Don't disable RHSM redhat.repo management on the AMI
+		// image, which is BYOS and does not use RHUI for content.
+		// Otherwise subscribing the system manually after booting
+		// it would result in empty redhat.repo. Without RHUI, such
+		// system would have no way to get Red Hat content, but
+		// enable the repo management manually, which would be very
+		// confusing.
```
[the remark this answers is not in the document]
achilleas-k: Right.  Done.

internal/distro/rhel9/azure.go
achilleas-k: oooh that's weird.  I might try to separate them actually.

achilleas-k: Separated.

internal/manifest/os.go
```diff
@@ -45,6 +45,7 @@ type OSCustomizations struct {
 	GPGKeyFiles      []string
 	Language         string
 	Keyboard         *string
+	X11Layouts       []string
```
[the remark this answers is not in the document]
achilleas-k: Done.

internal/distro/rhel9/images.go
achilleas-k: The way it's done now (see block below) allows for a customization to change the base `Keymap` but still maintain the `X11Keymap.Layouts` if the image config defines them.
Indeed, they are related, but since we don't support customizing the ``X11Keymap` (yet?), I wasn't sure whether an external customization for the base `Keymap` should override it or not.

```go
	if keyboard != nil {
		osc.Keyboard = keyboard
	} else if imageConfig.Keyboard != nil {
		osc.Keyboard = &imageConfig.Keyboard.Keymap
	}
	if imageConfig.Keyboard != nil && imageConfig.Keyboard.X11Keymap != nil {
		osc.X11KeymapLayouts = imageConfig.Keyboard.X11Keymap.Layouts
	}
```

achilleas-k: I guess keeping the image config can lead to weird issues where you have an X11KeymapLayout that's a whole different thing from the base Keymap, so it makes sense to override completely.  I'll change it.

achilleas-k:
Addressed comments.  Force push diff: https://github.com/osbuild/osbuild-composer/compare/91aa9469eb2dab8b710db504bb52e79d71e0a58d..0b1fca9dcebdf04ed5e22e4997cd1a2307746e69

thozza (commented):
Overall great work 👏 

I found the RHSM configuration rework in EC2 images to be messy, but functionally it works AFAICT. I don't consider any of my comments to be blockers, I'm fine with a follow-up given the size of this PR...

internal/distro/rhel9/images.go
thozza: Yeah, my point was that the image config values (all of them) related to keyboard settings should be thrown away if the user specified anything in the Blueprint. At least that's how it works now and IMO it makes sense. The reason is that the X11Keymap layout tends to be related to the Keymap value, so overriding just one of them with user-provided value and applying the other as it was set in image config does not IMO make sense.

internal/image/live.go
```diff
@@ -47,8 +48,15 @@ func (img *LiveImage) InstantiateManifest(m *manifest.Manifest,
...
-		imagePipeline.Filename = img.Filename
-		artifact = imagePipeline.Export()
+		switch img.Compression {
+		case "xz":
+			xzPipeline := manifest.NewXZ(m, buildPipeline, imagePipeline)
+			xzPipeline.Filename = img.Filename
+			artifact = xzPipeline.Export()
+		default:
+			imagePipeline.Filename = img.Filename
+			artifact = imagePipeline.Export()
```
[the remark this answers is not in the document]
achilleas-k: This appears outdated but the same applies to the latest version.  Good idea, I'll panic.

test/data/manifests/rhel_90-x86_64-ec2-boot.json
```diff
@@ -2360,7 +2140,7 @@
                 "origin": "org.osbuild.pipeline",
                 "references": {
                   "name:image": {
-                    "file": "image.raw"
```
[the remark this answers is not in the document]
achilleas-k: Yes, seeing how everything works now, it might be worth having a way to specify the base filename separately from the final artifact filename in cases where an image is compressed or tarred.

internal/distro/rhel9/ami.go
```diff
@@ -491,38 +403,101 @@ func rhelEc2SapPackageSet(t *imageType) rpmmd.PackageSet {
...
-	it.defaultImageConfig = defaultAMIImageConfigX86_64(osVersion)
+func mkEc2ImgTypeX86_64(osVersion string, rhsm bool) imageType {
+	it := ec2ImgTypeX86_64
+	ic := defaultEc2ImageConfigX86_64(osVersion, rhsm)
+	if rhsm && common.VersionLessThan(osVersion, "9.1") {
+		// Disable RHSM redhat.repo management
+		rhsmConf := ic.RHSMConfig[distro.RHSMConfigNoSubscription]
+		rhsmConf.SubMan.Rhsm = &osbuild.SubManConfigRHSMSection{ManageRepos: common.BoolToPtr(false)}
+		ic.RHSMConfig[distro.RHSMConfigNoSubscription] = rhsmConf
+	}
```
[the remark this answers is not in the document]
achilleas-k: Good catch.  Removed!

internal/distro/rhel9/ami.go
```diff
@@ -491,38 +403,101 @@ func rhelEc2SapPackageSet(t *imageType) rpmmd.PackageSet {
...
+					// do not disable the redhat.repo management if the user
+					// explicitly request the system to be subscribed
+				},
+			},
+		},
+	}
+	return rhsm.InheritFrom(ic)
+}
+
+func appendEC2Dracut(ic *distro.ImageConfig) *distro.ImageConfig {
```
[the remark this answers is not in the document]
achilleas-k: Done!

thozza (dismissed):
LGTM (to the extent that Github was able to actually show, but I went commit by commit before...)

internal/distro/rhel9/ami.go
```diff
@@ -491,38 +403,95 @@ func rhelEc2SapPackageSet(t *imageType) rpmmd.PackageSet {
...
 }
 
-func mkEC2SapImgTypeX86_64(osVersion string) imageType {
-	it := ec2SapImgTypeX86_64
-	it.defaultImageConfig = sapImageConfig(osVersion).InheritFrom(defaultEc2ImageConfigX86_64(osVersion))
+func mkAMIImgTypeX86_64(osVersion string, rhsm bool) imageType {
+	it := amiImgTypeX86_64
+	ic := defaultAMIImageConfigX86_64(osVersion, rhsm)
+	if rhsm && common.VersionLessThan(osVersion, "9.1") {
+		ic = appendRHSM(ic)
```
thozza, line 419: This should have been removed as well, since it is already done in `defaultAMIImageConfig()` called from `defaultAMIImageConfigX86_64()`. Also the configuration is applied unconditionally to AMIs regardless of the minor version so the condition in `defaultAMIImageConfig()` is actually the correct one. 

Nevertheless this "bug" has no effect on resulting functionality.

achilleas-k: I'll clean this up in a followup.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 26c94833c5ca9d2a7b6916b75e8413eb7872861c with the main merge-base 401a59b6e40adf9bdae9d074c6d3dd9b0b1a4a7c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3381037603/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3381037603/artifacts/browse

achilleas-k:
Found a bug in the installers.  The UEFI kernel command line arguments weren't being set properly for the kickstart file.
Fixed.

gicmo (approved):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 015c7e8183b3b6223440c73804f0427a70425306 with the main merge-base 401a59b6e40adf9bdae9d074c6d3dd9b0b1a4a7c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3382479539/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3382479539/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 09b4cd3481800039902ac34f30f1e4cc9791ccf4 with the main merge-base 7ae79e7786711a367038e5452ef7ca3c73e4339c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3388347888/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3388347888/artifacts/browse

gicmo (approved):
Reviewed only the recent changes, looking good.

PR #3121, opened 2022-11-08
Add /blueprints/change/NAME/COMMIT route and save blueprint changes in the store

bcl:
This pull request includes:

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -236,6 +236,7 @@ func setupRouter(api *API) *API {
 	api.router.GET("/api/v:version/blueprints/depsolve/*blueprints", api.blueprintsDepsolveHandler)
 	api.router.GET("/api/v:version/blueprints/freeze/*blueprints", api.blueprintsFreezeHandler)
 	api.router.GET("/api/v:version/blueprints/diff/:blueprint/:from/:to", api.blueprintsDiffHandler)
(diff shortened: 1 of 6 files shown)


internal/weldr/api.go
```diff
@@ -1805,6 +1806,70 @@ func (api *API) blueprintsDiffHandler(writer http.ResponseWriter, request *http.
...
+	if !verifyRequestVersion(writer, params, 1) {
+		return
+	}
+	name := params.ByName("blueprint")
+	if !verifyStringsWithRegex(writer, []string{name}, ValidBlueprintName) {
+		return
+	}
+
+	commit := params.ByName("commit")
+	if !verifyStringsWithRegex(writer, []string{commit}, ValidBlueprintName) {
```
ondrejbudai, line 1820: Isn't `commit` supposed to have a different regex than `ValidBlueprintName`?

ondrejbudai (commented):
I've found just one minor issue, otherwise, this looks very useful. :)

internal/weldr/api.go
bcl: It could, but in practice it doesn't matter since the commit is just a string, and if it doesn't match it will fail. We're using SHA1 for commit strings right now, but could use anything including uuid's including a `-`.

ondrejbudai (approved):

PR #3122, opened 2022-11-08
Use go install instead of deprecated go get

bcl:
In go 1.18 'go install' needs to be used to install an executable. See https://go.dev/doc/go-get-install-deprecation and
https://go.dev/doc/manage-install

This pull request includes:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,8 +74,8 @@ jobs:
         WORKDIR=$(readlink -f pkg/jobqueue/dbjobqueue/schemas)
         pushd $(mktemp -d)
         go mod init temp
-        go get -u github.com/jackc/tern
-        go run github.com/jackc/tern migrate -m "$WORKDIR"
+        go install github.com/jackc/tern@latest
+        $(go env GOPATH)/bin/tern migrate -m "$WORKDIR"
         popd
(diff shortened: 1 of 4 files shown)


bcl:
Ah, running it needs the @latest tag, like: `go run github.com/jackc/tern@latest migrate -m "$WORKDIR" `

supakeen:
Seems OK to my unused-to-Go eyes, would likely unbreak current CI failures.

gicmo (approved):
Thanks for this! Looks good to me.

.github/workflows/tests.yml
```diff
@@ -74,8 +74,8 @@ jobs:
         WORKDIR=$(readlink -f pkg/jobqueue/dbjobqueue/schemas)
         pushd $(mktemp -d)
         go mod init temp
-        go get -u github.com/jackc/tern
-        go run github.com/jackc/tern migrate -m "$WORKDIR"
+        go install github.com/jackc/tern@latest
```
gicmo, line 77: Wonder if we shouldn't pin a well known version, but not in this PR.

ondrejbudai: How is this done in prod, @croissanne ? Would be great to match these.

PR #3137, opened 2022-11-15
disk: align LVM2 volumes to the extent size 

gicmo:
When the size of a logical volume is not aligned to the extent size of the volume group, LVM2 will automatically align it by rounding up (see [metadata.c](https://github.com/lvmteam/lvm2/blob/868665766491d9d42b8acedaf07cafc7118d165c/lib/metadata/metadata.c#L1072)):

	Rounding up size to full physical extent 29.80 GiB
	Rounding up size to full physical extent <3.82 GiB

Since we don't take that into account when we create a new volume or set the size of an existing one, the size for the whole volume group will be short by that amount and thus the creation of the last volume
will fail:

  	Volume group <uuid> has insufficient free space (975 extents): 977 required.

To fix this a new `AlignUp` method is added to the `MountpointCreator` creator interface. It will align a given size to the requirements of the implementing container, like e.g. `LVMVolumeGroup`. It is then used by a new `alignEntityBranch` which takes a size and walks the entity path, calling `AlignUp` for all entities that implement said `MountpointCreator` interface; thus the resulting size should fullfil the alignment requirement for all elements in the path. NB: `PartitionTable` already had an `AlignUp` method.

Add a corresponding test.

### DIFF
--- a/internal/disk/btrfs.go
+++ b/internal/disk/btrfs.go
@@ -67,6 +67,10 @@ func (b *Btrfs) CreateMountpoint(mountpoint string, size uint64) (Entity, error)
 	return &b.Subvolumes[len(b.Subvolumes)-1], nil
 }
 
+func (b *Btrfs) AlignUp(size uint64) uint64 {
+	return size // No extra alignment necessary for subvolumes
+}
+
 func (b *Btrfs) GenUUID(rng *rand.Rand) {
 	if b.UUID == "" {
 		b.UUID = uuid.Must(newRandomUUIDFromReader(rng)).String()
--- a/internal/disk/lvm.go
+++ b/internal/disk/lvm.go
@@ -3,7 +3,13 @@ package disk
 import (
 	"fmt"
 	"strings"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
 )
+
+// Default physical extent size in bytes: logical volumes
+// created inside the VG will be aligned to this.
+const LVMDefaultExtentSize = 4 * common.MebiByte
 
 type LVMVolumeGroup struct {
 	Name        string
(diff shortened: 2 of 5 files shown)


internal/disk/lvm.go
```diff
@@ -3,8 +3,14 @@ package disk
 import (
 	"fmt"
 	"strings"
+
+	"github.com/osbuild/osbuild-composer/internal/common"
 )
 
+// Default physical extent size in bytes: logical volumes
+// created inside the VG will be aligned to this.
+const LVMDefaultExtentSize = 4 * common.MebiByte
```
thozza, line 12: Although this is the default, it could change and the manifest could end up having values rounded to a different extent size. It would probably make sense to set the value explicitly when calling lvm tooling in the osbuild stage (ask you mentioned as well when I raised the bug in out team channel). Do you plan a followup?

thozza (commented):
I added a few minor comments, otherwise this looks great!

Thanks for a quick fix.

internal/disk/btrfs.go
```diff
@@ -67,6 +67,10 @@ func (b *Btrfs) CreateMountpoint(mountpoint string, size uint64) (Entity, error)
 	return &b.Subvolumes[len(b.Subvolumes)-1], nil
 }
 
+func (b *Btrfs) AlignUp(size uint64) uint64 {
+	return size // None necessary for subvolumes
```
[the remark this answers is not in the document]
gicmo: I thought `None` as in `No alignment`, but I agree `Not` reads nicer.

internal/disk/lvm.go
gicmo: Yeah, indeed. Was the plan for a follow up because for this we also need osbuild changes. 👍

internal/disk/btrfs.go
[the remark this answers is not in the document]
thozza: Ah... It reads OK now 😅

internal/disk/lvm.go
```diff
@@ -99,15 +111,25 @@ func (vg *LVMVolumeGroup) CreateMountpoint(mountpoint string, size uint64) (Enti
...
 	vg.LogicalVolumes = append(vg.LogicalVolumes, lv)
 
 	return &vg.LogicalVolumes[len(vg.LogicalVolumes)-1], nil
 }
 
+func (vg *LVMVolumeGroup) AlignUp(size uint64) uint64 {
+
+	if size%LVMDefaultExtentSize != 0 {
+		size += LVMDefaultExtentSize - size%LVMDefaultExtentSize
+		fmt.Printf("LVM AlignUp %d\n", size)
```
bcl, line 127: Leftover debugging print?

gicmo: Doh, yes! ;) Thanks!

achilleas-k (approved):
LGTM

PR #3139, opened 2022-11-16
ci: fix the packer job

ondrejbudai:
Podman doesn't work when installed from 9.1 repositories on 9.0 image.

I found that upgrading the whole system helps. Sadly, that requires removing
python-unversioned-command for some reason.

This pull request includes:

### DIFF
--- a/tools/appsre-build-worker-packer.sh
+++ b/tools/appsre-build-worker-packer.sh
@@ -22,6 +22,12 @@ elif [ -n "$GIT_BRANCH" ]; then
 fi
 
 if [ "$ON_JENKINS" = false ]; then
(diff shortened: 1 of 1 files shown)


ondrejbudai:
podman bug: https://bugzilla.redhat.com/show_bug.cgi?id=2143282

thozza (dismissed):

ondrejbudai:
OK, RHEL 9.1 doesn't work because we have a ton of hacks somewhere to replace the repositories to the internal ones, which is something we don't want for GA.

croissanne (approved):
thank you for figuring this out

PR #3144, opened 2022-11-18
rhsm: remove CA from consumer secrets

croissanne:
This CA is only valid for the entitlement certificates, not the consumer certificates.

As a result resolving the ostree ref should use the system's CA cert pool.


achilleas-k (approved):
LGTM

Just to make sure I understand the issue and fix:
Before we assumed the cert for pulling commits was self-signed (but it isn't) so we would read the cert from the subscription info.
Now we're instead using the system cert.

PR #3146, opened 2022-11-18
osbuild-worker: handle interruptions gracefully

teg:
When a machine is being shut down systemd will send each service a SIGTERM signal, essentially
asking them to please shut down. If they do not do so within a certain timeout, systemd sends
SIGKILL instead, which forcibly shuts down the service immediately. This is what currently happens
if a worker machine is being shut down, for instance because AWS decides they want their VMs
back.

When a worker is shut down mid-job we currently lose track of the job, and wait for the
heartbeat to expire, before rescheduling the job. This is both slow and means we can't
distinguish between jobs that failed because of a bug (i.e., unexpectedly) and jobs that
failed because they were interrupted for external reasons (to be expected).

This PR improves the situation by listening for SIGTERM, and when we receive it immediately
cancelling all in-flight work. Moreover, if a job is cancelled in this way, composer is notified
that the job was interrupted, and as a result composer immediately reschedules the job.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -8,6 +8,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -103,7 +104,7 @@ func loadFormatRequestMap() formatRequestMap {
 	return frm
 }
 
-type manifestJob func(chan string) error
+type manifestJob func(context.Context, chan string) error
 
 func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, distribution distro.Distro, archName string, seedArg int64, path string, cacheRoot string) manifestJob {
 	distroName := distribution.Name()
@@ -121,7 +122,7 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 			FetchChecksum: cr.OSTree.Parent,
 		}
 	}
-	job := func(msgq chan string) (err error) {
+	job := func(ctx context.Context, msgq chan string) (err error) {
 		defer func() {
 			msg := fmt.Sprintf("Finished job %s", filename)
 			if err != nil {
@@ -146,7 +147,7 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 			options.OSTree.ImageRef = imgType.OSTreeRef()
 		}
 
-		packageSpecs, err := depsolve(cacheDir, imgType, bp, options, repos, distribution, archName)
+		packageSpecs, err := depsolve(ctx, cacheDir, imgType, bp, options, repos, distribution, archName)
 		if err != nil {
 			err = fmt.Errorf("[%s] depsolve failed: %s", filename, err.Error())
 			return
@@ -226,13 +227,13 @@ func resolveContainers(containers []blueprint.Container, archName string) ([]con
 	return resolver.Finish()
 }
 
-func depsolve(cacheDir string, imageType distro.ImageType, bp blueprint.Blueprint, options distro.ImageOptions, repos []rpmmd.RepoConfig, d distro.Distro, arch string) (map[string][]rpmmd.PackageSpec, error) {
+func depsolve(ctx context.Context, cacheDir string, imageType distro.ImageType, bp blueprint.Blueprint, options distro.ImageOptions, repos []rpmmd.RepoConfig, d distro.Distro, arch string) (map[string][]rpmmd.PackageSpec, error) {
 	solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, cacheDir)
 	solver.SetDNFJSONPath("./dnf-json")
 	packageSets := imageType.PackageSets(bp, options, repos)
 	depsolvedSets := make(map[string][]rpmmd.PackageSpec)
 	for name, pkgSet := range packageSets {
-		res, err := solver.Depsolve(pkgSet)
+		res, err := solver.Depsolve(ctx, pkgSet)
 		if err != nil {
 			return nil, err
 		}
@@ -438,7 +439,7 @@ func main() {
 	nJobs := len(jobs)
 	fmt.Printf("Collected %d jobs\n", nJobs)
 	wq := newWorkerQueue(uint32(nWorkers), uint32(nJobs))
-	wq.start()
+	wq.start(context.Background())
 	fmt.Printf("Initialised %d workers\n", nWorkers)
 	fmt.Printf("Submitting %d jobs... ", nJobs)
 	for _, j := range jobs {
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -254,7 +254,7 @@ func (impl *OSBuildJobImpl) getContainerClient(destination string, targetOptions
 	return client, nil
 }
 
-func (impl *OSBuildJobImpl) Run(ctx context.Context, job worker.Job) (interface{}, error) {
+func (impl *OSBuildJobImpl) Run(ctx context.Context, job *worker.Job) (interface{}, error) {
 	logWithId := logrus.WithField("jobId", job.Id().String())
 	// Initialize variable needed for reporting back to osbuild-composer.
 	var osbuildJobResult *worker.OSBuildJobResult = &worker.OSBuildJobResult{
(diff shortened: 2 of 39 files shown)


internal/cloud/instanceprotector/instance-protector.go
```diff
@@ -0,0 +1,166 @@
...
+			case <-ctx.Done():
+				return
+			}
+			select {
+			case <-ip.protect:
+			case <-time.After(time.Minute):
+				setProtection(false)
+				break unprotect
+			case <-ctx.Done():
+				return
```
thozza, line 100: I wonder if we should not call `setProtection(false)` here as well. If I'm reading the code correctly, it could happen that we protected the VM as part of starting a job, then finished the job, but if the `ctx` is canceled before the minute passes, we would keep the VM protected and return. 🤔  Given the comments, it does not sound like the cloud provider would care anyway.

thozza (commented):
Great PR 👍 . I like how you hooked up all the contexts in the worker jobs and internal cloud provider APIs.

I have one request: I noticed that there are many cases in worker job implementations `Run()` method which is returning `nil` even when an error is set in the job-specific result structure. Since this is then logged by the worker, it would be IMO good to be consistent and always return the `error` from `Run()`, if there was an error which caused the job to fail. This should improve our ability to construct better metrics using these messages.

internal/cloud/instanceprotector/instance-protector.go
teg: D'oh! Very good point! I think you are right.

teg:
> Great PR 👍 . I like how you hooked up all the contexts in the worker jobs and internal cloud provider APIs.

Thanks!

> I have one request: I noticed that there are many cases in worker job implementations `Run()` method which is returning `nil` even when an error is set in the job-specific result structure. Since this is then logged by the worker, it would be IMO good to be consistent and always return the `error` from `Run()`, if there was an error which caused the job to fail. This should improve our ability to construct better metrics using these messages.

I hadn't reviewed the error handling, but that sounds like it makes a lot of sense, I'll update the PR to make this more consistent.

internal/cloud/instanceprotector/instance-protector.go
```diff
@@ -0,0 +1,187 @@
...
+				default:
+					// All protect and unprotect calls have been matched up before
+					// the timeout expired and no protect requests are pending. Discard
+					// the request to protect the instance and start over.
+					continue protect
+				}
+			}
+		}
+		setProtection(true)
+		defer setProtection(false)
```
ondrejbudai, line 100: This shouldn't be here. `defer` is scoped to a function, not to a block. `setProtection` should be therefore called before `break unprotect` after line 118.

internal/cloud/instanceprotector/instance-protector.go
```diff
@@ -0,0 +1,187 @@
...
+	unprotect:
+		for {
+			// Pop off alternating unprotect / protect tokens until we receive
+			// an unprotect token without a matching protect one for one minute.
+			// At that point, onprotect the instance and break from this loop and
```
ondrejbudai, line 105: s/onprotect/unprotect

ondrejbudai (changes requested):
Looks cool, thanks! Found some issues, see inline.

cmd/osbuild-worker/worker.go
```diff
@@ -241,6 +249,25 @@ func NewWorker(config *workerConfig, unix bool, address, cacheDir string) (*Work
...
+	w.jobImplKinds = append(w.jobImplKinds, map[string]JobImplementation{
+		worker.JobTypeAWSEC2Copy: &AWSEC2CopyJobImpl{
+			AWSCreds: awsCredentials,
+		},
+		worker.JobTypeAWSEC2Share: &AWSEC2ShareJobImpl{
+			AWSCreds: awsCredentials,
+		},
+	})
+
+	// the remaining job types are neither CPU nor IO intensive, and are
```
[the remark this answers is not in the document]
teg: Ah, thanks!

cmd/gen-manifests/main.go
```diff
@@ -437,7 +438,7 @@ func main() {
 	nJobs := len(jobs)
 	fmt.Printf("Collected %d jobs\n", nJobs)
 	wq := newWorkerQueue(uint32(nWorkers), uint32(nJobs))
-	wq.start()
+	wq.start(context.Background())
```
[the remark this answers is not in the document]
teg: Created issue.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -279,11 +279,6 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	defer func() {
 		validateResult(osbuildJobResult, job.Id().String())
```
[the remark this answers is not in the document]
teg: I think that's similar to the comment from @thozza. I started thinking of how to clean this up, but may possibly delay it to a follow-up is it looks quite fragile/hairy. Or do you think this PR is making it worse?

internal/cloud/instanceprotector/instance-protector.go
teg: Thanks! I think I might add a helper function to get the right scope. I was trying to make it obvious what's going on, still not quite there ;) I'll also add a test for this function.

cmd/osbuild-worker/jobimpl-osbuild.go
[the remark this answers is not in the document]
ondrejbudai: Slightly, because `osbuildJobResult` is now returned. Previously, everything happened in this function. Anyway, I think it's fine to leave it as is. Getting it clean feels out of scope of this PR.

internal/cloud/instanceprotector/instance-protector.go
teg: Fixed.

internal/cloud/instanceprotector/instance-protector.go
teg: Fixed.

internal/cloud/instanceprotector/instance-protector.go
teg: Fixed.

cmd/osbuild-worker/worker.go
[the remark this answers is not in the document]
teg: Fixed.

internal/cloud/instanceprotector/instance-protector.go
```diff
@@ -0,0 +1,154 @@
...
+			// for at least ten seconds
+			timeout = time.After(ip.protectTimeout)
+		case <-ctx.Done():
+			return
+		}
+		for {
+			// Drain the channels of already scheduled protect / unprotect pairs
+			select {
+			case <-ip.unprotect:
+			case <-timeout:
```
ondrejbudai, line 97: `case <-ctx.Done: return` wouldn't hurt here. It might be an issue if `ip.protectTimeout` is set to a long time.

internal/worker/server.go
```diff
@@ -40,6 +40,9 @@ const (
 	JobTypeAWSEC2Share      string = "aws-ec2-share"
 )
 
+const maxHeartbeatRetries = 2
```
ondrejbudai, line 43: Why don't just use the same retry count for both failure cases? This seems a bit counter-intuitive with regard to the current implementation. E.g. After 2 interrupt retries, no heartbeat retry will actually happen as we have only one retry counter.

teg: Fair, let's not try to be clever. If we need this to be more fine-grained we can update that in the future. For now I upped the retry count a bit and used the same in both cases.

internal/cloud/instanceprotector/instance-protector.go
teg: Fixed.

teg:
> I noticed that there are many cases in worker job implementations `Run()` method which is returning `nil` even when an error is set in the job-specific result structure. Since this is then logged by the worker, it would be IMO good to be consistent and always return the `error` from `Run()`, if there was an error which caused the job to fail. This should improve our ability to construct better metrics using these messages.

I looked at this a bit, but ultimately decided to defer that to a future PR. I still think you are probably right, but it all looked a bit too fragile to change as part of this (already quite large) PR.

The logic is that `err` being non-null means that there was an internal error in the job implementation, but `err` being null and the osbuild not succeeding meaning that everything was ok from the worker's point of view and there was (possibly an expected) error in the osbuild run itself. I think this can be made more consistent, but as it is right now there is a slight difference in behaviour, so I thought I'd figuring out the finer points for a follow-up.

PR #3147, opened 2022-11-21
Measure 5xx errors on all requests for image-builder-composer/worker

kingsleyzissou:
This pull request includes:

Creates a custom middleware function to record all 5xx errors for composer & worker as we
are currently only recording internal errors for the `/compose` endpoint.

The existing compose metrics can be removed in a follow up PR once the relevant alerts and dashboards
have been implemented.

### DIFF
--- a/internal/prometheus/middleware.go
+++ b/internal/prometheus/middleware.go
@@ -1,6 +1,8 @@
 package prometheus
 
 import (
+	"errors"
+	"strconv"
 	"strings"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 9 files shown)


internal/prometheus/middleware.go
```diff
@@ -18,3 +20,35 @@ func MetricsMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
...
+			err := next(ctx)
+
+			path := pathLabel(ctx.Path())
+			method := ctx.Request().Method
+			status := ctx.Response().Status
+
+			httpErr := new(echo.HTTPError)
+			if errors.As(err, &httpErr) {
+				status = httpErr.Code
+			}
```
croissanne, line 42: Does the code in the echo error sometimes differ from the status one?

Also any reason for using As over Is?

kingsleyzissou: So the status code before that check is always `200`. If I used Is over As, it doesn't pick it up and the error code stays `200`.

kingsleyzissou:
I think I found a bug with this. Marking as draft while I look into it

thozza (approved):
All looks good, nice work.

PR #3148, opened 2022-11-21
ocm metrics: measure 5xx errors for all requests

kingsleyzissou:
This pull request includes:

Uses the `ocm-sdk-go/metrics` middleware to register and measure all requests for composer & worker
(currently we only measure 5xx errors for composer requests).


kingsleyzissou:
I've created another PR for a custom middleware function here:
https://github.com/osbuild/osbuild-composer/pull/3147

PR #3149, opened 2022-11-22
SPEC: run the %preun commands in worker package only on removal

thozza:
PR#553 [1] introduced a change to the worker subpackage and replaced the systemd %preun RPM macro with a literal calls to `systemctl` due to missing functionality in systemd. However, the change omitted the fact, that the RPM macro executes commands only on package removal, not upgrade.

As a result a local or remote worker which is running on the system while the osbuild-composer-worker package is updated, gets stopped.

Fix the scriptlet and run the commands only on package removal.

[1] https://github.com/osbuild/osbuild-composer/pull/553

Signed-off-by: Tomáš Hozza <thozza@redhat.com>

This pull request includes:


gicmo (approved):
🙈 ❤️

PR #3158, opened 2022-11-28
CloudAPI: add description for `Repository` definition

thozza:
Specifically describe that at least one of `baseurl`, `mirrorlist` or `metalink` properties must be specified. Although this is not enforced by the schema, it is the reality. Modifying the schema to enforce this results in the generated data structure for `Repository` to be simply an `interface{}`, which would just make the code more ugly, Therefore this is covered by the API implementation logic.

Signed-off-by: Tomáš Hozza <thozza@redhat.com>

This pull request includes:


croissanne (approved):

croissanne:
Thank you!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -778,6 +778,11 @@ components:
         - vsphere
     Repository:
       type: object
+      description: |
+        Repository configuration.
+        At least one of the 'baseurl', 'mirrorlist', 'metalink' properties must
+        be specified. If more of them are specified, the order of precedence is
```
ondrejbudai, line 784: I wonder if we should internally change the logic to "oneOf", i.e. erroring out on requests with multiple "repo sources" defined. This is just confusing.

thozza: I had the same thought, but DNF does not enforce this AFAICT...

thozza: Nevertheless I agree that it is confusing 😉

PR #3160, opened 2022-11-29
Bugfix: Add RHSM Fact (APIType) to RHEL 9 image definitions

achilleas-k:
- Added an API type to the RHSM Fact option when generating manifests with the `gen-manifests` tool.  The value is `test-manifest` which should never appear in our data but can be used to check if the stage is properly applied in the test manifests.
- Fixed bug in RHEL 9 where the RHSM facts were missing.

### DIFF
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -95,6 +95,7 @@ type OSCustomizations struct {
 	WAAgentConfig       *osbuild.WAAgentConfStageOptions
 	UdevRules           *osbuild.UdevRulesStageOptions
 	LeapSecTZ           *string
+	FactAPIType         string
 
 	Subscription *distro.SubscriptionImageOptions
 	RHSMConfig   map[distro.RHSMSubscriptionStatus]*osbuild.RHSMStageOptions
@@ -519,6 +520,14 @@ func (p *OS) serialize() osbuild.Pipeline {
 
 	if p.OpenSCAPConfig != nil {
 		pipeline.AddStage(osbuild.NewOscapRemediationStage(p.OpenSCAPConfig))
+	}
+
+	if p.FactAPIType != "" {
+		pipeline.AddStage(osbuild.NewRHSMFactsStage(&osbuild.RHSMFactsStageOptions{
+			Facts: osbuild.RHSMFacts{
+				ApiType: p.FactAPIType,
+			},
+		}))
 	}
 
 	if p.SElinux != "" {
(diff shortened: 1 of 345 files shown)


test/data/manifests/rhel_7-x86_64-azure_rhui-boot.json
```diff
@@ -22,10 +22,7 @@
       },
       {
         "name": "rhel-eng",
-        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el7/el7-x86_64-eng-rel-20220530",
-        "package-sets": [
-          "build"
-        ]
```
thozza, line 28: This does not look right 🤔

internal/manifest/os.go
```diff
@@ -95,6 +95,7 @@ type OSCustomizations struct {
 	WAAgentConfig       *osbuild.WAAgentConfStageOptions
 	UdevRules           *osbuild.UdevRulesStageOptions
 	LeapSecTZ           *string
+	FactAPIType         *string
```
thozza, line 98: nitpick: IMO this could be a plain `string`, because an empty string is probably not a valid value and we can use it to check if it is set. This would make it a bit "easier" to use and not to have to do the dance with pointers...

thozza (commented):
Overall looking great. Just one change in the el7 manifest is probably a bug. Happy to approve afterwards...

internal/manifest/os.go
achilleas-k: Thought about it and thought it might be valid to have an empty string and just the key, but maybe not for our purposes.  Will change.

test/data/manifests/rhel_7-x86_64-azure_rhui-boot.json
achilleas-k: hm, yeah, `gen-manifests` drops the package-sets field from repo configs.  Fixing.

achilleas-k: Fixed.

internal/manifest/os.go
achilleas-k: Fixed.

thozza (dismissed):
😍

achilleas-k:
Oh.  Unit tests are going to fail.  We need to set the option there as well.

thozza (approved):

PR #3130, opened 2022-11-11
Support ignition in simplified-installer and raw-image

runcom:
This patchset is the basic work needed to support ignition in r4e - it doesn't include any blueprint or file injection change as I wanted to maintain it easy to review and since this won't break anything, I think it'll be beneficial to merge this before other work to plumb ignition files/config in the blueprints. This PR also includes a commit that enables kargs injection from the installer and raw-image blueprint to effectively enable passing the artifact an `ignition.config.url` and test it out. Lastly, to make the work bound to the sysroot.ro=true patch, this also includes https://github.com/osbuild/osbuild-composer/pull/3053

Things missing:

- the `ignition-edge` RPM isn't yet packaged anywhere (working with the coreos team to ship it), the source lives https://github.com/fedora-iot/ignition-edge
- polish commits

PR #3161, opened 2022-11-29
Ignition blueprint config support on rhel9

7flying:
This allows a user to configure the system via `edge-simplified-installer`
using an ignition configuration specified in the blueprint.

This ignition config can be embedded in the ISO as a Base64
encoded file (`ignition.embedded.data`) or as a file
containing the URL where the ignition config file is served
(`ignition.embedded.url`).
This embedded config (either data or url) ends up in the root
of the ISO.

The user can also instead specify an URL serving an ignition
config file that will passed as a karg and be fetched at first
boot (`ignition.firstboot.url`).

<hr>
depends on osbuild/osbuild-composer#3130 (that PR sets the things needed to run ignition, this PR allows the user to configure the things to pass to ignition, so this PR doesn't break anything)
<hr>

This pull request includes:

- [x] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/ --> see osbuild/guides#96

### DIFF
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -23,6 +23,21 @@ type Customizations struct {
 	InstallationDevice string                    `json:"installation_device,omitempty" toml:"installation_device,omitempty"`
 	FDO                *FDOCustomization         `json:"fdo,omitempty" toml:"fdo,omitempty"`
 	OpenSCAP           *OpenSCAPCustomization    `json:"openscap,omitempty" toml:"openscap,omitempty"`
+	Ignition           *IgnitionCustomization    `json:"ignition,omitempty" toml:"ignition,omitempty"`
+}
+
+type IgnitionCustomization struct {
+	Embedded  *EmbeddedIgnitionCustomization  `json:"embedded,omitempty" toml:"embedded,omitempty"`
+	FirstBoot *FirstBootIgnitionCustomization `json:"firstboot,omitempty" toml:"firstboot,omitempty"`
+}
+
+type EmbeddedIgnitionCustomization struct {
+	ProvisioningURL string `json:"url,omitempty" toml:"url,omitempty"`
+	Data64          string `json:"data,omitempty" toml:"data,omitempty"`
+}
+
+type FirstBootIgnitionCustomization struct {
+	ProvisioningURL string `json:"url,omitempty" toml:"url"`
 }
 
 type FDOCustomization struct {
@@ -389,4 +404,26 @@ func (c *Customizations) GetOpenSCAP() *OpenSCAPCustomization {
 
 func (f *FDOCustomization) HasFDO() bool {
 	return f != nil
+}
+
+func (c *Customizations) GetIgnition() *IgnitionCustomization {
+	if c == nil {
+		return nil
+	}
+	return c.Ignition
+}
+
+func (c *IgnitionCustomization) HasIgnition() bool {
+	return c != nil
+}
+
+func (c *EmbeddedIgnitionCustomization) CheckEmbeddedIgnition() error {
+	if c == nil {
+		return nil
+	}
+	if c.Data64 != "" && c.ProvisioningURL != "" {
+		t := reflect.TypeOf(*c)
+		return &CustomizationError{fmt.Sprintf("'%s' and '%s' are not allowed at the same time", t.Field(0).Name, t.Field(1).Name)}
+	}
+	return nil
 }
--- a/internal/manifest/coi_iso_tree.go
+++ b/internal/manifest/coi_iso_tree.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"crypto/sha256"
 	"fmt"
 
 	"github.com/osbuild/osbuild-composer/internal/disk"
@@ -65,6 +66,30 @@ func (p *CoreOSISOTree) serialize() osbuild.Pipeline {
 		},
 		osbuild.NewFilesInputs(osbuild.NewFilesInputReferencesPipeline(p.payloadPipeline.Name(), p.payloadPipeline.Filename)),
 	))
+
+	if p.coiPipeline.Ignition != nil {
+		filename := ""
+		copyInput := ""
+		// These specific filenames in the root of the ISO are expected by
+		// coreos-installer-dracut during installation
+		if p.coiPipeline.Ignition.Data64 != "" {
+			filename = "ignition_config"
+			copyInput = p.coiPipeline.Ignition.Data64
+		} else {
+			filename = "ignition_url"
+			copyInput = p.coiPipeline.Ignition.ProvisioningURL
+		}
+		pipeline.AddStage(osbuild.NewCopyStageSimple(
+			&osbuild.CopyStageOptions{
+				Paths: []osbuild.CopyStagePath{
+					{
+						From: fmt.Sprintf("input://inlinefile/sha256:%x", sha256.Sum256([]byte(copyInput))),
+						To:   fmt.Sprintf("tree:///%s", filename),
+					},
+				},
+			},
+			osbuild.NewIgnitionInlineInput(copyInput)))
+	}
 
 	pipeline.AddStage(osbuild.NewMkdirStage(&osbuild.MkdirStageOptions{
 		Paths: []osbuild.Path{
--- a/test/cases/ostree-simplified-installer.sh
+++ b/test/cases/ostree-simplified-installer.sh
@@ -885,6 +885,60 @@ EOF
 sudo ansible-playbook -v -i "${TEMPDIR}"/inventory -e image_type=redhat -e ostree_commit="${INSTALL_HASH}" -e edge_type=edge-simplified-installer -e fdo_credential="false" /usr/share/tests/osbuild-composer/ansible/check_ostree.yaml || RESULTS=0
 check_result
 
+########################################################################
+##
+## Build edge-simplified-installer with ignition embedded 
+## (only on rhel9+)
+##
+########################################################################
+
+if [[ "${ID}-${VERSION_ID}" = "rhel-9.1" || "${ID}-${VERSION_ID}" = "centos-9" ]]; then
+   tee "$BLUEPRINT_FILE" > /dev/null <<EOF
+name = "simplified_iso_with_ignition_embedded_url"
+description = "A rhel-edge simplified-installer image with an embedded ignition config URL"
+version = "0.0.1"
+
+[customizations]
+installation_device = "/dev/vda"
+
+[customizations.ignition.embedded]
+url = "http://some-server/config.ig"
+EOF
+
+   greenprint "📄 simplified_iso_with_ignition_embedded_url blueprint "
+   cat "$BLUEPRINT_FILE"
+
+   # Prepare the blueprint for the compose.
+   greenprint "📋 Preparing installer blueprint"
+   sudo composer-cli blueprints push "$BLUEPRINT_FILE"
+   sudo composer-cli blueprints depsolve simplified_iso_with_ignition_embedded_url
+
+   # Build simplified installer iso image.
+   build_image simplified_iso_with_ignition_embedded_url "${INSTALLER_TYPE}" "${PROD_REPO_URL}/"
+
+   # Download the image
+   greenprint "📥 Downloading the simplified_iso_with_ignition_embedded_url image"
+   sudo composer-cli compose image "${COMPOSE_ID}" > /dev/null
+   ISO_FILENAME="${COMPOSE_ID}-${INSTALLER_FILENAME}"
+
+   # Mount the image
+   sudo mkdir /mnt/installer
+   sudo mount -o loop "${ISO_FILENAME}" /mnt/installer
+
+   # Check that the image contains the ignition_url file
+   if [[ -f "/mnt/installer/ignition_url" ]]; then
+       RESULTS=1
+   fi
+
+   # Check the resulting image
+   check_result
+
+   sudo umount /mnt/installer
+else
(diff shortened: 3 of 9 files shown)


7flying:
@runcom , regarding the kernel arguments, I know that in #3130  via `support kargs in blueprint` commit we can already do this, but from the user point of view it might be more straightforward to add it via the specific blueprint option so that you have less room to mess thins up. What do you think?

Also, the PRs are going to collide in the `igniton_stage.go` file, that is intentional to handle all the ignition stuff in one place.

bcl:
Pleas also remember to update the documentation in the guide - https://github.com/osbuild/guides/blob/main/osbuild-composer/src/blueprint-reference/blueprint-reference.md

internal/blueprint/customizations.go
```diff
@@ -384,3 +399,22 @@ func (c *Customizations) GetOpenSCAP() *OpenSCAPCustomization {
...
+	}
+	return c.Ignition
+}
+
+func (c *IgnitionCustomization) HasIgnition() bool {
+	return c != nil
+}
+
+func (c *EmbeddedIgnitionCustomization) CheckEmbeddedIgnition() error {
+	if c.Data64 != "" && c.ProvisioningURL != "" {
```
bcl, line 415: This should check for `c == nil` first

teg:
> regarding the kernel arguments, I know that in #3130 via `support kargs in blueprint` commit we can already do this, but from the user point of view it might be more straightforward to add it via the specific blueprint option so that you have less room to mess things up. What do you think?

If we can, we should try to give users high-level options that we translate to kargs (or other things), so totally agree with this principle :)

teg:
Could `ignition_data` be called just `data`? As it is already in the `ignition` section. In general, how does this compare to how CoreOS handles ignition configuration? Do you have any thoughts on this @cgwalters ?

7flying:
> Could `ignition_data` be called just `data`? As it is already in the `ignition` section. In general, how does this compare to how CoreOS handles ignition configuration? Do you have any thoughts on this @cgwalters ?

Sure thing, will change it :+1: 
We are using some of the scripts that the CoreOS team uses to handle ignition (https://github.com/fedora-iot/ignition-edge), small changes here and there but it should be very similar.

cgwalters:
> This ignition config can be embedded in the ISO as a Base64
encoded file (ignition.embedded.ignition_data)

Where is this file physically present?  It's not clear to me from the PR.  Is that referencing a kernel argument?

You may already be aware but I think it's still worth cross-referencing here: For CoreOS we¹ developed an IMO neat system where an Ignition config can be *dynamically* embedded in an ISO image: https://coreos.github.io/coreos-installer/cmd/iso/#coreos-installer-iso-ignition-embed 

The way this works is by sticking the config into a "reserved" area of the ISO.  Given that Edge is already using coreos-installer, this could make sense to also support (would I think require a bit of work to replicate bits of [cosa buildextend-live](https://github.com/coreos/coreos-assembler/blob/main/src/cmd-buildextend-live) in the ISO).  

One advantage of this is that it gives a clear model for building a "generic" ISO that can have e.g. machine-specific customization injected later.

But - that's all specific to the ISO.

To support the *generic* functionality of embedding a custom Ignition config "statically", it can be injected into `/usr/lib/dracut/modules.d/40ignition-conf` today.  Note on CoreOS derivatives we have:

```
[root@cosa-devsh ~]# cat /usr/lib/dracut/modules.d/40ignition-conf/00-core.ign 
{
  "ignition": {
    "version": "3.0.0"
  },
  "passwd": {
    "users": [
      {
        "name": "core",
        "gecos": "CoreOS Admin",
        "groups": [
          "adm",
          "sudo",
          "systemd-journal",
          "wheel"
        ]
      }
    ]
  }
}
```

(This config should likely *not* be present on Edge derivatives as the "core" user is obviously a continuation of a CoreOS default and when moving to full custom images it's likely not desired)

But, other "static" configuration can be injected in the same way.

¹ Specifically IIRC @bgilbert and @jlebon, I was only tangentially involved

7flying:
> > This ignition config can be embedded in the ISO as a Base64
> > encoded file (ignition.embedded.ignition_data)
> 
> Where is this file physically present? It's not clear to me from the PR. Is that referencing a kernel argument?
> 
We put the file in the root of the ISO:
```
[root@citest-1 si]# ll /mnt/disk/
total 802405
drwxr-xr-x. 1 root root      2048 Nov 29 06:30 EFI
-r--r--r--. 1 root root      2048 Nov 29 06:34 boot.catalog
-rw-r--r--. 1 root root 821653588 Nov 29 06:33 disk.img.xz
-rw-r--r--. 1 root root      1629 Nov 29 06:30 ignition_config
drwxr-xr-x. 1 root root      2048 Nov 29 06:33 images
```
It does not reference a kernel argument.

> You may already be aware but I think it's still worth cross-referencing here: For CoreOS we¹ developed an IMO neat system where an Ignition config can be _dynamically_ embedded in an ISO image: https://coreos.github.io/coreos-installer/cmd/iso/#coreos-installer-iso-ignition-embed
> 
I wasn't aware of that at all! CC'ing  @runcom ! From a quick look at it, if I understood it correctly, it would require producing the ISO first and then the installer should be ran later on to embed the config file. We do not run the installer until we are installing the image, so maybe that "later" (regarding the customisation injection that is made later on) wouldn't work for us since we just want to have a .iso with everything in there that can be installed and that's it.
But, again, CC'ing @runcom since I do have less context.

> The way this works is by sticking the config into a "reserved" area of the ISO. Given that Edge is already using coreos-installer, this could make sense to also support (would I think require a bit of work to replicate bits of [cosa buildextend-live](https://github.com/coreos/coreos-assembler/blob/main/src/cmd-buildextend-live) in the ISO).
> 
> One advantage of this is that it gives a clear model for building a "generic" ISO that can have e.g. machine-specific customization injected later.
> 
> But - that's all specific to the ISO.
> 
> To support the _generic_ functionality of embedding a custom Ignition config "statically", it can be injected into `/usr/lib/dracut/modules.d/40ignition-conf` today. Note on CoreOS derivatives we have:
> 
> ```
> [root@cosa-devsh ~]# cat /usr/lib/dracut/modules.d/40ignition-conf/00-core.ign 
> {
>   "ignition": {
>     "version": "3.0.0"
>   },
>   "passwd": {
>     "users": [
>       {
>         "name": "core",
>         "gecos": "CoreOS Admin",
>         "groups": [
>           "adm",
>           "sudo",
>           "systemd-journal",
>           "wheel"
>         ]
>       }
>     ]
>   }
> }
> ```
> 
> (This config should likely _not_ be present on Edge derivatives as the "core" user is obviously a continuation of a CoreOS default and when moving to full custom images it's likely not desired)
> 
> But, other "static" configuration can be injected in the same way.
> 
> ¹ Specifically IIRC @bgilbert and @jlebon, I was only tangentially involved

jlebon:
> I wasn't aware of that at all! CC'ing @runcom ! From a quick look at it, if I understood it correctly, it would require producing the ISO first and then the installer should be ran later on to embed the config file. We do not run the installer until we are installing the image, so maybe that "later" (regarding the customisation injection that is made later on) wouldn't work for us since we just want to have a .iso with everything in there that can be installed and that's it. But, again, CC'ing @runcom since I do have less context.

I'm not familiar with the Edge install flow, but I can explain the CoreOS install flow this fits in and you can see from there whether it makes sense to adapt to it as well.

Users download a generic live ISO from e.g. [getfedora.org](https://getfedora.org/en/coreos/download/) (or for RHCOS, e.g. [the mirror](https://mirror.openshift.com/pub/)). They can then customize that ISO using [`coreos-installer iso customize`](https://coreos.github.io/coreos-installer/customizing-install/). Users can choose how much to customize it. All this is done "offline", without having to actually boot the ISO. Crucially, the `--dest-device` switch will make it so that simply booting the ISO will initiate an install and reboot. All this to say that the Ignition embedding feature that Colin mentioned above provides a lot of power (that's what backs most of the `iso customize` features).

Though it's not clear to me from this PR whether the Ignition config we're talking about is for the ISO or for the installed system. Obviously a lot of this flexibility comes from the fact that CoreOS ISOs themselves run Ignition (and then we talk about "Ignition configs within Ignition configs" when embedding the target system's config within the ISO's config, but that's what the `iso customize` work is intended to simplify).

teg:
Thanks for our input on this @cgwalters and @jlebon!

I think you are right @7flying , that we want to make any customisations (and hence ignition insertion) at image build time, rather than after the fact, so static ISOs + dynamic customisations doesn't really make sense.

That said, it might be worth considering putting our ignition file in the same "reserved" location that the coreos tool does? I think this is sort of an internal detail, so could be done in a follow-up, but if it makes sense to be consistent we should be :)

I think a good mental model would be that the R4E image that comes out of image builder is structured the same way as a coreos installer ISO after the dynamic configuration has been applied. Unless there are technical quirks I'm not aware of that would make this not worth it, of course.

7flying:
Rebased due to #3166

7flying:
> Thanks for our input on this @cgwalters and @jlebon!
> 
> I think you are right @7flying , that we want to make any customisations (and hence ignition insertion) at image build time, rather than after the fact, so static ISOs + dynamic customisations doesn't really make sense.
> 
> That said, it might be worth considering putting our ignition file in the same "reserved" location that the coreos tool does? I think this is sort of an internal detail, so could be done in a follow-up, but if it makes sense to be consistent we should be :)
> 

Having an standardised  way of doing things sounds always good to me :), but I have to check with others on the team.

> I think a good mental model would be that the R4E image that comes out of image builder is structured the same way as a coreos installer ISO after the dynamic configuration has been applied. Unless there are technical quirks I'm not aware of that would make this not worth it, of course.

Right now I have no clue about how to do this in osbuild-composer, we might need to modify/extend one of osbuild's copy stages to handle that placement (size/alignment stuff I guess), but that would be it.

internal/manifest/coi_iso_tree.go
```diff
@@ -66,6 +67,24 @@ func (p *CoreOSISOTree) serialize() osbuild.Pipeline {
 		osbuild.NewFilesInputs(osbuild.NewFilesInputReferencesPipeline(p.payloadPipeline.Name(), p.payloadPipeline.Filename)),
 	))
 
+	if p.coiPipeline.Ignition != nil {
+		cso := osbuild.CopyStageOptions{}
+		filename := ""
+		copyInput := ""
+		if p.coiPipeline.Ignition.Data64 != "" {
+			filename = "ignition_config"
```
cgwalters, line 75: What reads this filename?

cgwalters:
I think it's not really important to match the same mechanism right now; i.e. I would definitely not block on it.  It's something that can be done later.

internal/manifest/coi_iso_tree.go
7flying: It is passed to a format string in the `CopyStageOptions` a couple of lines below

cgwalters: Ahh OK, I hadn't realized the dracut glue picked it up by default.  This seems very much worth adding as a comment into the code.

7flying: Will do :+1:

teg:
> I think it's not really important to match the same mechanism right now; i.e. I would definitely not block on it. It's something that can be done later.

Yup, I agree with this. Just something to keep in mind as an aspirational goal ;)

test/cases/ostree-simplified-installer.sh
```diff
@@ -885,6 +885,56 @@ EOF
 sudo ansible-playbook -v -i "${TEMPDIR}"/inventory -e image_type=redhat -e ostree_commit="${INSTALL_HASH}" -e edge_type=edge-simplified-installer -e fdo_credential="false" /usr/share/tests/osbuild-composer/ansible/check_ostree.yaml || RESULTS=0
 check_result
 
+########################################################################
+##
+## Build edge-simplified-installer with ignition embedded
```
achilleas-k, line 890: This script is run on RHEL 8 as well but the feature was only added to RHEL 9.

achilleas-k (changes requested):
Looks good but the test script needs to be conditioned to only run on RHEL 9 since the new features weren't added to RHEL 8.
Also, can we drop the second-to-last commit with the manifest generation?  The manifests haven't changed, it's just some metadata key reordering and it doesn't affect any tests, so no need to add extra noise.

internal/manifest/coi_iso_tree.go
```diff
@@ -66,6 +67,26 @@ func (p *CoreOSISOTree) serialize() osbuild.Pipeline {
...
+			copyInput = p.coiPipeline.Ignition.Data64
+		} else {
+			filename = "ignition_url"
+			copyInput = p.coiPipeline.Ignition.ProvisioningURL
+		}
+		cso.Paths = append(cso.Paths, osbuild.CopyStagePath{
+			From: fmt.Sprintf("input://inlinefile/sha256:%x", sha256.Sum256([]byte(copyInput))),
+			To:   fmt.Sprintf("tree:///%s", filename),
+		})
```
achilleas-k, line 86: TINIEST of nitpicks, but I find the initialisation of `cso` and the `append` a bit strange here.  It could just be initialised at after the conditions with
```go
cso := osbuild.CopyStageOptions{
    Paths = ...
}
```
instead of appending.

7flying: very true, I was over complicating things

test/cases/ostree-simplified-installer.sh
7flying: Thanks, will do

achilleas-k (approved):
Looks good.  Thanks.

7flying:
thanks for the review @achilleas-k !

PR #2487, opened 2022-03-28
test/ostree-simplified-installer: destroy VM fixup

jrusz:
There was another virt-install added recently and it was missed in
1d51c0764936902c681283a3489845ef660fef5e

@henrywang I didn't notice the new `virt-install` and this now also started getting stuck on RHEL-9.0 nightly. I had to manually unstuck this job: https://gitlab.com/osbuild/ci/osbuild-composer/-/jobs/2256353462

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3166, opened 2022-12-01
RHEL 9: update edge-simplified-installer to new definitions

achilleas-k:
Following up from #3120, this PR finishes the rewrite of image definitions to the new framework for RHEL 9 and CentOS 9.

### Framework changes

The `rootfs-image` and `bootiso` pipelines now take any pipeline as input (Pipeline interface instead of a concrete Pipeline implementation).  This make it possible to reuse the pipelines with any other pipeline dependency.  Of course not all pipelines are valid inputs (though they may create "valid" manifests in the strictest sense), so we might define an interface in the future for the input pipelines that are compatible with each pipeline.

The `efiboot-tree` pipeline now doesn't depend on any other pipeline, but instead takes two strings, `product` and `version` as arguments.

### Image definition changes

Added `prefixdevname` dracut modules to all installers.

This was always included in RHEL but not Fedora.  Adding it to the default dracut modules list will include it in all installers (both Anaconda and COI) for RHEL 9 and Fedora.

### Test changes

Test manifests that require an edge commit ID used to have `test/edge` as a parent reference, which was invalid.
I changes this to a valid (but still fake) parent commit ID: `f{64}`.  The manifests are still not buildable but now they are valid in the techincal sense (i.e., `osbuild --inspect` doesn't produce an error).


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4c776dc933c0a2f59037488e94817c7b65e4f7ba with the main merge-base fd72a6d16ef03f9d101e80881be7fb907ccddbd4).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3411347016/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3411347016/artifacts/browse

internal/image/ostree_simplified_installer.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/osbuild/osbuild-composer/internal/environment"
+	"github.com/osbuild/osbuild-composer/internal/fdo"
+	"github.com/osbuild/osbuild-composer/internal/manifest"
+	"github.com/osbuild/osbuild-composer/internal/platform"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
+	"github.com/osbuild/osbuild-composer/internal/runner"
+	"github.com/osbuild/osbuild-composer/internal/workload"
+)
+
+type OSTreeSimplifiedInstaller struct {
```
teg, line 19: Potentially controversial, should this be `CoreOSInstaller`? It doesn't necessarily install CoreOS, but it does use the CoreOS installer.

internal/image/image_installer.go
```diff
@@ -109,7 +109,7 @@ func (img *ImageInstaller) InstantiateManifest(m *manifest.Manifest,
 	rootfsImagePipeline := manifest.NewISORootfsImg(m, buildPipeline, anacondaPipeline)
 	rootfsImagePipeline.Size = 4 * common.GibiByte
 
-	bootTreePipeline := manifest.NewEFIBootTree(m, buildPipeline, anacondaPipeline)
+	bootTreePipeline := manifest.NewEFIBootTree(m, buildPipeline, img.Product, img.OSVersion)
```
teg, line 112: For the future: What's the semantics of `Product` and `OSVersion`? Do they describe the payload? Should the payload then be the one encoding them and expose them via some interface? If it is just a coincidence that they values are the same between the EFIBootTree and the Anaconda pipeline we should keep it as is done in this patch. Either way, no need to figure this out as part of this PR, this makes sense for now.

achilleas-k: I've been thinking about the same questions too and I don't have solid answers.  I made a lot of decisions here for technical simplicity, but I'd rather we figured out something that's more consistent semantically.

teg: Let's discuss it all when the dust settles ;)

internal/image/ostree_simplified_installer.go
gicmo: @nullr0ute wanted to rename it to "Simplified Provisioner", IIRC.

achilleas-k: Any name works for me.

teg: I think we should still make the specific image type "RHEL for Edge Simplified Provisioner" (or whatever it ends up being), but this is the internal image kind which should be generic (whether the payload is R4E or CoreOS or anything else).

gicmo: @teg but `CoreOSInstaller` is a bit ambiguous, because that exists as its own thing too; I mean it is used in a very specific way here that is different way here than in FOCS/RHCOS.

teg: Yup, it is ambiguous. I meant to refer to the name of the software, not the name of the artifact.
Think "Anaconda" rather than "Fedora Workstation Installer", in this case the terms are just a but overloaded.

achilleas-k:
Need `prefixdevname` to be installed for the installers.

internal/distro/rhel9/images.go
```diff
@@ -332,6 +333,66 @@ func edgeRawImage(workload workload.Workload,
...
+	if bpFDO := customizations.GetFDO(); bpFDO != nil {
+		img.FDO = fdo.FromBP(*bpFDO)
+	}
+
+	d := t.arch.distro
+	img.ISOLabelTempl = d.isolabelTmpl
+	img.Product = d.product
+	img.Variant = "edge"
+	img.OSName = "redhat"
+	img.OSVersion = d.osVersion
```
runcom, line 391: sysroot readonly stuff are missing from the raw image def - https://github.com/osbuild/osbuild-composer/pull/3130/commits/59291dbb8ff9c6482362aabd3ef9bd9a47c7c24b#diff-b94d65ecb56ac51cbc7b24dcf2fbe1d88e608805d3c041977624c088c9c809a3R310 (perhaps you can pull in https://github.com/osbuild/osbuild-composer/pull/3130/commits/59291dbb8ff9c6482362aabd3ef9bd9a47c7c24b which also includes testing?)

internal/image/ostree_simplified_installer.go
teg: *scratches head* you might be right... let's table this for now anyway, not crucial to the PR at all :)

internal/distro/rhel9/images.go
achilleas-k: Sure thing.  I was just fixing up the other stuff but I'll get to this right after.

internal/manifest/coreos_installer.go
```diff
@@ -0,0 +1,161 @@
...
+	// Biosdevname indicates whether or not biosdevname should be used to
+	// name network devices when booting the installer. This may affect
+	// the naming of network devices on the target system.
+	Biosdevname bool
+
+	FDO *blueprint.FDOCustomization
+}
+
+// NewCOI creates an CoreOS installer pipeline object.
+func NewCOI(m *Manifest,
```
bcl, line 42: Small nit, I think this should be the full name - NewCoreOSInstaller, since this is all in one package that makes it clearer what's being called. 'COI' means I have to figure out which acronym to match against in my head :)

internal/manifest/coreos_installer.go
```diff
@@ -0,0 +1,161 @@
...
+		version:    version,
+	}
+	buildPipeline.addDependent(p)
+	m.addPipeline(p)
+	return p
+}
+
+// TODO: refactor - what is required to boot and what to build, and
+// do they all belong in this pipeline?
+func (p *CoreOSInstaller) coiBootPackageSet() []string {
```
bcl, line 64: I guess I'm picking at naming today :) I think 'getBootPackageSer' would be clearer. coi seems redundant since it's attached to the CoreOSInstaller struct.

internal/manifest/coi_iso_tree.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"fmt"
+
+	"github.com/osbuild/osbuild-composer/internal/disk"
+	"github.com/osbuild/osbuild-composer/internal/osbuild"
+	"github.com/osbuild/osbuild-composer/internal/users"
+)
+
+type COIISOTree struct {
```
bcl, line 11: I lean towards thinking this should be CoreOSInstallerISOTree but that's a bit long. Maybe CoreOSISOTree?

internal/manifest/coreos_installer.go
achilleas-k: That's a good point.  We shouldn't obscure public functions like this.  I'll rename it.

internal/manifest/coreos_installer.go
achilleas-k: Thanks for picking the naming nits :)
These are good.

internal/manifest/coi_iso_tree.go
achilleas-k: I like `CoreOSISOTree`.  Thanks.

achilleas-k:
Renamed types and functions from @bcl's comments.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 2d96faf90a5deb63c8ce111ebb869beb7bae85f6 with the main merge-base 8c75975917ff59ab1bfa43ee7110a383f5023a5d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3423209873/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3423209873/artifacts/browse

thozza (commented):
Overall looks good, I found just minor things which I commented on...

achilleas-k:
Status on this: I'm investigating the install failure before making any changes.

achilleas-k:
While looking into this, I think I found that the post-install growfs never worked.  Our CI forcibly ends the installation (https://github.com/osbuild/osbuild-composer/pull/2487) and boots the machine after some time so we would not have caught it.

achilleas-k:
Rebased on `main`.

achilleas-k:
Updated CS9 test manifest repos and addressed @thozza's comments.

achilleas-k:
Rebased on main again and resolved conflicts.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 8428c9e9914d97a455c48b5d819ad337fcddf90f with the main merge-base ff21f2935fd57e480093dc673e34eb401352d43e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3446449053/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3446449053/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD b27c58c0ae41fcb88caa41e4f3f34e7992c5b0ae with the main merge-base ff21f2935fd57e480093dc673e34eb401352d43e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3450459074/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3450459074/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 434a8ffcf7f8812052307ccfb0c830c9ccbd48c6 with the main merge-base ff21f2935fd57e480093dc673e34eb401352d43e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3454493329/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3454493329/artifacts/browse

teg (approved):
Great stuff!

internal/distro/rhel9/images.go
teg: @achilleas said he would do this in a follow-up rather than part of this already quite big PR

PR #3099, opened 2022-10-26
Fix blueprint firewall support

bcl:
This pull request includes:

- [X] adequate testing for the new functionality or fixed issue
- [X] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3170, opened 2022-12-06
Update toml library and add test

bcl:
This pull request includes:

- [ ] 1st commit of any `rhbz#` related PR contains bug reproducer; CI reports FAIL or


.github/workflows/tests.yml
```diff
@@ -74,8 +74,8 @@ jobs:
         WORKDIR=$(readlink -f pkg/jobqueue/dbjobqueue/schemas)
         pushd $(mktemp -d)
         go mod init temp
-        go get -u github.com/jackc/tern
-        go run github.com/jackc/tern migrate -m "$WORKDIR"
+        go install github.com/jackc/tern@latest
```
teg, line 77: Could we pin a specific version to avoid incompatible changes biting us in the future?

bcl: The problem here isn't the version, it's that go now wants us to use install instead of get.

bcl:
Closing, no chance of the PR working at this point.

ondrejbudai:
@bcl I think that this bug was fixed on `main` by basically dependabot updating the dep. However, it would be nice to have the new test on `main` as well. Would it take much effort to "port" it?

ondrejbudai:
Ignore me, this was done in https://github.com/osbuild/osbuild-composer/pull/3099, I failed at using the `git log`.

Issue #3168, opened 2022-12-05
iot-installer failing on aarch64 -  No such file or directory: '/run/osbuild/inputs/data/usr/share/syslinux/isolinux.bin'


pcdubs:
**Describe the bug**
The Fedora IoT compose is currently failing when attempting to build an AArch64 IoT installer: 

https://koji.fedoraproject.org/koji/taskinfo?taskID=94935722

From https://kojipkgs.fedoraproject.org//work/tasks/5722/94935722/aarch64-iot-installer.log.json:

[...]

PR #3171, opened 2022-12-06
Enable isolinux only for x86_64

achilleas-k:
ISOLinux was unconditionally enabled for all ISOs, which is incorrect and breaks builds on aarch64.
Enable it only for x86_64.

Fixes https://github.com/osbuild/osbuild-composer/issues/3168

### DIFF
--- a/internal/image/ostree_installer.go
+++ b/internal/image/ostree_installer.go
@@ -114,7 +114,7 @@ func (img *OSTreeInstaller) InstantiateManifest(m *manifest.Manifest,
 
 	isoPipeline := manifest.NewISO(m, buildPipeline, isoTreePipeline)
 	isoPipeline.Filename = img.Filename
-	isoPipeline.ISOLinux = true
+	isoPipeline.ISOLinux = img.Platform.GetArch() == platform.ARCH_X86_64
 	artifact := isoPipeline.Export()
 
 	return artifact, nil
--- a/internal/image/image_installer.go
+++ b/internal/image/image_installer.go
@@ -146,7 +146,7 @@ func (img *ImageInstaller) InstantiateManifest(m *manifest.Manifest,
 
 	isoPipeline := manifest.NewISO(m, buildPipeline, isoTreePipeline)
 	isoPipeline.Filename = img.Filename
-	isoPipeline.ISOLinux = true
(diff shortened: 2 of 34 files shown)


teg:
Just a comment regarding the PR description:

If we cannot boot-test, we should still build-test in image definitions CI. manifest-db should protect against regressions in osbuild, not that images work as expected.

internal/image/ostree_installer.go
```diff
@@ -114,7 +114,7 @@ func (img *OSTreeInstaller) InstantiateManifest(m *manifest.Manifest,
 
 	isoPipeline := manifest.NewISO(m, buildPipeline, isoTreePipeline)
 	isoPipeline.Filename = img.Filename
-	isoPipeline.ISOLinux = true
+	isoPipeline.ISOLinux = img.Platform.GetArch() == platform.ARCH_X86_64
```
teg, line 117: Potentially for follow-up (or just to consider): should the platform interface have a `ISOLinuxSupported()` function that we would query instead of this logic here?

teg (approved):

internal/image/ostree_installer.go
bcl: I don't think that's worth the effort. isolinux is x86 only, and it is going away in favor of booting everything using grub2.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD b61ca29e73e0b3bd901c2f50fabdcedc9b6876bc with the main merge-base 51851bfd4920ffdae5f779f49c88421129a18734).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3433809734/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3433809734/artifacts/browse

internal/image/ostree_installer.go
achilleas-k: I considered making it more explicit like you mentioned Tom, but Brian's point is valid.  Maybe we could do it for now just for the _cleanliness_.

teg: Yup, agreed.

PR #3172, opened 2022-12-07
cloudapi/v2: set ostree rhsm option on image options

croissanne:
The ostree options are used during the ostree resolve job, but when generating the manifest the rhsm value comes from the image options, so it's necessary to set it on both.


achilleas-k:
It's obvious that the interaction between request options, image options, and job arguments is getting too messy and we have a lot of these issues cropping up.  We need a good way to centralise all this and make it obvious what goes where.

Thanks for the fix!

achilleas-k (approved):

PR #3180, opened 2022-12-12
RHEL 9 & Fedora: Anaconda boot arguments

achilleas-k:
The bootiso.mono stage in osbuild that we used until recently adds the inst.stage2 option unconditionally [1] whereas the current grub2.iso stage that we use now doesn't.

This addresses rhbz#2152192: the lack of `inst.stage2` broke a workflow that relied on it.

[1] https://github.com/osbuild/osbuild/blob/8511add1695c49cc7f3df2220044a4eed17ba1a5/stages/org.osbuild.bootiso.mono#L369

### DIFF
--- a/internal/image/image_installer.go
+++ b/internal/image/image_installer.go
@@ -116,11 +116,9 @@ func (img *ImageInstaller) InstantiateManifest(m *manifest.Manifest,
 	bootTreePipeline.UEFIVendor = img.Platform.GetUEFIVendor()
 	bootTreePipeline.ISOLabel = isoLabel
 
-	kernelOpts := make([]string, 0)
+	kernelOpts := []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", isoLabel)}
 	if img.ISORootKickstart {
 		kernelOpts = append(kernelOpts, fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", isoLabel, kspath))
-	} else {
-		kernelOpts = append(kernelOpts, fmt.Sprintf("inst.stage2=hd:LABEL=%s", isoLabel))
 	}
 	kernelOpts = append(kernelOpts, img.AdditionalKernelOpts...)
 	bootTreePipeline.KernelOpts = kernelOpts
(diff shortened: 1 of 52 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4b7e27698080ef5fc1d0f545b728af6ca5ecf1b8 with the main merge-base 19ec3be6bd70e737e5dc8d493b355a9f473bbe0b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3458969045/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3458969045/artifacts/browse

supakeen (commented):
I'm not a huge fan of the difference between `efi_boot_tree` and `iso_tree` as it is now. Can we by default set `inst.stage2` in `efi_boot_tree` as well so the images only have to pass `inst.ks` if applicable?

achilleas-k:
> I'm not a huge fan of the difference between `efi_boot_tree` and `iso_tree` as it is now. Can we by default set `inst.stage2` in `efi_boot_tree` as well so the images only have to pass `inst.ks` if applicable?

`EFIBootTree` is also used by the simplified (CoreOS-based) installer so we can't add it unconditionally there.  I would like consistency here too though.

I think the proper solution for this would be for the installer pipeline (either `coiPipeline` or `anacondaPipeline`) should specify the arguments it requires.  The this pipeline can be passed to the efi boot tree and the pipeline generator can read the required command line arguments for the associated payload.
There was a similar flow before, but it was specific to Anaconda.  It would be nice if we could define Pipeline interfaces for specific categories of pipelines, for example `InstallerPipeline` that implements methods that a dependant pipeline can use to read options specific to installers that not all generic Pipelines support.

teg (approved):

PR #3187, opened 2022-12-13
RHEL 9: Do not enable user module in Anaconda Edge Installer when no users are specified

achilleas-k:
In RHEL 9, for the edge installer, only enable the users module when it is required for creating a user specified in the blueprint or request.  This was the behaviour until recently.  We changed it to enable interactive user creation even when no users were defined in the request.  However this interferes with tooling for fully automated installations.

Note that the `edge_installer_with_users` manifests are unaffected.  The module is enabled in these manifests so that the user specified in the build request can be created during installation.

Also, the image installer continues to always enable the Users module.  This is required for the service, since there is no way to specify users during the request in the frontend and creating users interactively at install time is the only way to create a user (aside from injecting a kickstart file).

**TESTING UPDATE**: I built a test image and shared it with Edge Fleet Management and they confirmed it's working.

### DIFF
--- a/internal/distro/fedora/images.go
+++ b/internal/distro/fedora/images.go
@@ -211,12 +211,12 @@ func imageInstallerImage(workload workload.Workload,
 	if strings.HasPrefix(distro.Name(), "fedora") && !common.VersionLessThan(distro.Releasever(), "38") {
 		img.AdditionalAnacondaModules = []string{
 			"org.fedoraproject.Anaconda.Modules.Security",
-			"org.fedoraproject.Anaconda.Modules.Users",
 			"org.fedoraproject.Anaconda.Modules.Timezone",
 			"org.fedoraproject.Anaconda.Modules.Localization",
 		}
 		img.AdditionalKernelOpts = []string{"inst.webui", "inst.webui.remote"}
 	}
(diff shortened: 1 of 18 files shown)


croissanne (approved):
There is a slight change in behaviour now in fedora vs rhel, but if that's ok, LGTM! Thank you for getting on this so quickly.

achilleas-k:
> There is a slight change in behaviour now in fedora vs rhel, but if that's ok, LGTM! Thank you for getting on this so quickly.

Yes, this is different now in Fedora and RHEL 9.  @teg how do we feel about this?  I think it's okay for the Fedora IoT Installer to have different behaviour from the RHEL Edge Installer.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 951325c566c1e9332e83058f17a237c33606c058 with the main merge-base 07a8f3d5ea1f148bb5c8330cecf7955bc054f7b3).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3467733992/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3467733992/artifacts/browse

teg (approved):
If we want to change Fedora IoT we can do that separately.

PR #3196, opened 2022-12-18
DO NOT MERGE: regenerate mainfests on a clean checkout

runcom:
On a clean main checkout, on F37,, runinng `./tools/test-case-generators/generate-all-test-cases -i /dev/null --output test/data/manifests --image-type edge-commit --image-type edge-simplified-installer --image-type edge-container --image-type edge-raw-image manifests` I get the diff in this PR - something must be off on either my side or projects (likely mine). The workaround I did was to run:

- `go run ./cmd/gen-manifests -output test/data/manifests -workers 20 -images edge-container`
- `go run ./cmd/gen-manifests -output test/data/manifests -workers 20 -images edge-commit`

but I'd have expected the command above to just work like it did in the past

Signed-off-by: Antonio Murdaca <antoniomurdaca@gmail.com>

This pull request includes:

### DIFF
--- a/test/data/manifests/centos_8-aarch64-edge_commit-boot.json
+++ b/test/data/manifests/centos_8-aarch64-edge_commit-boot.json
@@ -12474,6 +12474,5 @@
         "check_gpg": true
       }
     ]
-  },
-  "no-image-info": true
+  }
 }
--- a/test/data/manifests/centos_8-aarch64-edge_commit_with_container-boot.json
+++ b/test/data/manifests/centos_8-aarch64-edge_commit_with_container-boot.json
@@ -13014,15 +13014,5 @@
         "check_gpg": true
       }
     ]
-  },
-  "containers": [
-    {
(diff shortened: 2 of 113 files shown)


achilleas-k:
These are differences between the older `tools/test-case-generators` scripts and `cmd/gen-manifests`.

From a quick glance, here are the major issues:
1. The existing manifests include `"no-image-info": true`.  This used to be useful to signal that a manifest didn't include image info back when we had image infos.  `cmd/gen-manifests` just adds it because it was never meant to include image info, but now that we never do, we should drop it.
2. The older scripts don't include support for adding containers to a request/blueprint.  This is a newer feature and I added support for containers in the test manifests while doing the RHEL 9 rewrite to make sure they carried over.  If the older scripts are run, the container in the request is dropped so the source and stage are removed.
3. Empties omitted or not: some parts of the request or metadata are omitted when empty in one tool and not the other.

The long term solution is to drop the manifests from the composer repo altogether and rely on manifest-db.  For now I think we should always use the gen-manifests tool for consistency.

PR #3081, opened 2022-10-19
RHEL-8.7+/9.1+: replace RHSM config on EC2 RHUI images with `redhat-cloud-client-configuration` package

thozza:
- [tools: update repos used to generate 8.7 and 9.1 image manifests](https://github.com/osbuild/osbuild-composer/commit/826a8b32cfefa4f43fab135f6c55a07155a24982)
- [distro/rhel8: adjust RHSM-related config on EC2 8.7+ images](https://github.com/osbuild/osbuild-composer/commit/77befd04d9cccd8760433b3758a327dbea89da5b)
- [distro/rhel9: adjust RHSM-related config on EC2 9.1+ images](https://github.com/osbuild/osbuild-composer/commit/d68d7f40bc3c003aead0136c3c58cf37343ad7e5)

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3203, opened 2023-01-06
distro/rhel8: enable autoregistration for all Azure images

ondrejbudai:
COMPOSER-1866 reported that Azure SAP images should also have rhsm autoregistration enabled. Since all Azure images except for the SAP one already have autoreg enabled, we can just move the config into the default image config that's inherited by all azure-related image config.

This pull request includes:

### DIFF
--- a/internal/distro/rhel8/azure.go
+++ b/internal/distro/rhel8/azure.go
@@ -450,6 +450,23 @@ var defaultAzureImageConfig = &distro.ImageConfig{
 		},
 	},
 	DefaultTarget: common.StringToPtr("multi-user.target"),
+	RHSMConfig: map[distro.RHSMSubscriptionStatus]*osbuild.RHSMStageOptions{
+		// Nowadays, we want autoregistration everywhere.
+		distro.RHSMConfigNoSubscription: {
+			SubMan: &osbuild.RHSMStageOptionsSubMan{
+				Rhsmcertd: &osbuild.SubManConfigRHSMCERTDSection{
+					AutoRegistration: common.BoolToPtr(true),
+				},
+			},
+		},
+		distro.RHSMConfigWithSubscription: {
+			SubMan: &osbuild.RHSMStageOptionsSubMan{
+				Rhsmcertd: &osbuild.SubManConfigRHSMCERTDSection{
+					AutoRegistration: common.BoolToPtr(true),
+				},
+			},
+		},
+	},
 }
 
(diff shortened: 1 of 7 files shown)


thozza:
@ondrejbudai I'm not really sure it is correct to make the change for 8.7+ images. I think that we should check with the Entitlements team, since they requested that `redhat-cloud-client-configuration` is installed on EC2 images, instead of directly configuring RHSM (see https://github.com/osbuild/osbuild-composer/pull/3081). I expect that it may be desirable to use the same approach also for Azure images...

internal/distro/rhel8/azure.go
```diff
@@ -481,31 +498,6 @@ var defaultAzureByosImageConfig = &distro.ImageConfig{
...
-		distro.RHSMConfigWithSubscription: {
-			SubMan: &osbuild.RHSMStageOptionsSubMan{
-				Rhsmcertd: &osbuild.SubManConfigRHSMCERTDSection{
-					AutoRegistration: common.BoolToPtr(true),
-				},
-				// do not disable the redhat.repo management if the user
-				// explicitly request the system to be subscribed
-			},
-		},
-	},
```
thozza, line 508: So this is the configuration that is expected to be applied also to EC2 BYOS images. The reason is that on RHUI images, this is done by installing `redhat-cloud-client-configuration` rpm instead of configuring RHSM directly.

More information is in https://github.com/osbuild/osbuild-composer/pull/3081

ondrejbudai:
Let's then wait with this PR until we have guidance from the RHSM team. ;) Thanks, Tomáš, for keeping an eye on this!

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 6ea3d9d4f1de45a399ab7216a3ef774acdf9b788 with the main merge-base 4bafe1ead91d321197cb0c5fef0d5a3be9434e5d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3564596716/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3564596716/artifacts/browse

ondrejbudai:
Implemented in #3233

PR #3209, opened 2023-01-10
internal/prometheus: remove compose fail metrics

kingsleyzissou:
We have switched how 5xx errors are being recorded internally and we are now recording all failures
for all endpoints. As a result, a dedicated metric only for compose failures is no longer required.

This pull request includes:


croissanne (approved):
thanks!

PR #3214, opened 2023-01-13
Schutzfile: bump osbuild commit for GA RHEL

croissanne:
Since composer is not updated into these distributions, it's safe to use
    a newer version of osbuild on them without the risk of releasing a
    composer version which would require an unreleased osbuild version.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -156,21 +156,21 @@
   "rhel-8.4": {
     "dependencies": {
       "osbuild": {
-        "commit": "fdf064b15a1ef42bd13abdfcb7e08f764f596a60"
+        "commit": "ae563ff8962be0ae13e5becd0a3c2c646eeacc24"
       }
     }
   },
   "rhel-8.6": {
     "dependencies": {
       "osbuild": {
-        "commit": "fdf064b15a1ef42bd13abdfcb7e08f764f596a60"
+        "commit": "ae563ff8962be0ae13e5becd0a3c2c646eeacc24"
       }
     }
   },
   "rhel-8.7": {
     "dependencies": {
       "osbuild": {
-        "commit": "fdf064b15a1ef42bd13abdfcb7e08f764f596a60"
+        "commit": "ae563ff8962be0ae13e5becd0a3c2c646eeacc24"
       }
     }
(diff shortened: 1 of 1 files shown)


teg:
How do we make sure the osbuild version in the worker vm is the one tested in ci?

croissanne:
> How do we make sure the osbuild version in the worker vm is the one tested in ci?

Sadly, this decouples it.

But after talking with @achilleas-k this might be preferable to the risk of releasing composer depending on features of an unreleased osbuild, I think we do want to be able to move quicker than release cycles in the service. If we merge this,  we should update the worker and the schutzfile with the next osbuild release of course.

teg:
I think we should distinguish our CI between
 - Testing for RHEL releases, ideally never pinned in Schutzfile, but using the dependencies from the target (nightly) repos; and
 - Testing for service release, use the pinned dependency pulled from our own repos on top of the GA version of the distro we are testing (as that's what we will use in the service).

I thought that was how it currently worked, but it does not appear so, so maybe something changed (or I misremember).

In other words: I think we can on the one hand both enforce that we never accidentally release composer with a too new dependency and make sure that we always test against what we will push into the service.

achilleas-k:
> I think we should distinguish our CI between ...

I fully agree and I think we should make this change now, in this PR.  We already test the APIs separately in CI, so why not have separate osbuild versions defined and use each where appropriate?  For example the API tests should use an osbuild version ID that corresponds to what we want to run in the service and others (like base and ostree tests) should use what we intend to release for each RHEL version.  So for RHEL 9.1 we should have two osbuild version IDs in the Schutzfile, one for on-prem and one for cloud tests.

croissanne:
So I changed this to just update to the latest release.

achilleas-k (approved):

achilleas-k:
Updates discussed in other channels:
- Updating the osbuild commit ID for RHEL versions that were already released is fine because we wont release into them again.
- This tests and updates the osbuild version that's used in the workers because we run them on GA (or supported) RHEL releases.
- We should still be mindful about not pinning osbuild versions for unreleased RHEL distros.

PR #3219, opened 2023-01-16
osbuild-service-maintenance: clean up all regions

croissanne:
Since we started cloning images to different regions, the maintenance script should clean up all of these regions.


ondrejbudai (approved):
Looks good!

PR #3227, opened 2023-01-19
Fix workflow shellcheck action deprecated set-output error

henrywang:
Use ludeeus/action-shellcheck@master to have fix included.
`master` has this fix https://github.com/ludeeus/action-shellcheck/pull/70 landed. And the readme file [suggest](https://github.com/ludeeus/action-shellcheck#example) to use `master`


.github/workflows/tests.yml
```diff
@@ -167,7 +167,7 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
     - name: Run ShellCheck
-      uses: ludeeus/action-shellcheck@1.1.0
+      uses: ludeeus/action-shellcheck@master
```
thozza, line 170: Would it be possible to refer to a specific commit instead? E.g. the last one from the default branch? This way we will prevent unexpected breakages if they change anything on that branch?

thozza: Alternatively we can use different shellcheck action, but I don't like using `master`... 🙄

henrywang: Thanks @thozza. Latest commit works.

thozza (approved):
LGTM, Thanks! 👍

ondrejbudai (approved):
Thanks!

PR #3228, opened 2023-01-19
Add custom repos to /etc/yum.repos.d

kingsleyzissou:
This pull request includes:
The ability to specify repo customizations so that custom repositories will be saved to
the `/etc/yum.repos.d` directory in the built image. 

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -571,6 +571,11 @@ filesystem.sh:
   variables:
     SCRIPT: filesystem.sh
 
+custom-repos.sh:
+  extends: .integration
+  variables:
+    SCRIPT: custom-repos.sh
+
 cross-distro.sh:
   extends: .integration
   variables:
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -23,6 +23,7 @@ type Customizations struct {
 	Ignition           *IgnitionCustomization    `json:"ignition,omitempty" toml:"ignition,omitempty"`
 	Directories        []DirectoryCustomization  `json:"directories,omitempty" toml:"directories,omitempty"`
 	Files              []FileCustomization       `json:"files,omitempty" toml:"files,omitempty"`
+	Repositories       []RepositoryCustomization `json:"repositories,omitempty" toml:"repositories,omitempty"`
 }
 
 type IgnitionCustomization struct {
@@ -332,4 +333,19 @@ func (c *Customizations) GetFiles() []FileCustomization {
 		return nil
 	}
 	return c.Files
+}
+
+func (c *Customizations) GetRepositories() ([]RepositoryCustomization, error) {
+	if c == nil {
+		return nil, nil
+	}
+
+	for idx := range c.Repositories {
+		err := validateCustomRepository(&c.Repositories[idx])
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return c.Repositories, nil
 }
--- a/internal/blueprint/repository_customizations.go
+++ b/internal/blueprint/repository_customizations.go
@@ -2,6 +2,10 @@ package blueprint
 
 import (
 	"fmt"
+	"net/url"
+
+	"github.com/osbuild/osbuild-composer/internal/fsnode"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 )
 
 type RepositoryCustomization struct {
@@ -32,4 +36,69 @@ func validateCustomRepository(repo *RepositoryCustomization) error {
 		return fmt.Errorf("Repository gpg check is set to true but no gpg keys are provided")
 	}
 	return nil
+}
+
+func (rc *RepositoryCustomization) getFilename() string {
+	if rc.Filename == "" {
+		return fmt.Sprintf("%s.repo", rc.Id)
+	}
+	return rc.Filename
+}
+
+func RepoCustomizationsToRepoConfigAndGPGKeyFiles(repos []RepositoryCustomization) (map[string][]rpmmd.RepoConfig, []*fsnode.File, error) {
+	if len(repos) == 0 {
+		return nil, nil, nil
+	}
+
+	repoMap := make(map[string][]rpmmd.RepoConfig, len(repos))
+	var gpgKeyFiles []*fsnode.File
+	for _, repo := range repos {
+		filename := repo.getFilename()
+		convertedRepo := repo.customRepoToRepoConfig()
+
+		// convert any inline gpgkeys to fsnode.File and
+		// replace the gpgkey with the file path
+		for idx, gpgkey := range repo.GPGKeys {
+			if _, ok := url.ParseRequestURI(gpgkey); ok != nil {
+				// create the file path
+				path := fmt.Sprintf("/etc/pki/rpm-gpg/RPM-GPG-KEY-%s-%d", repo.Id, idx)
+				// replace the gpgkey with the file path
+				convertedRepo.GPGKeys[idx] = fmt.Sprintf("file://%s", path)
+				// create the fsnode for the gpgkey keyFile
+				keyFile, err := fsnode.NewFile(path, nil, nil, nil, []byte(gpgkey))
+				if err != nil {
+					return nil, nil, err
+				}
+				gpgKeyFiles = append(gpgKeyFiles, keyFile)
+			}
+		}
+
+		repoMap[filename] = append(repoMap[filename], convertedRepo)
+	}
+
+	return repoMap, gpgKeyFiles, nil
+}
+
+func (repo RepositoryCustomization) customRepoToRepoConfig() rpmmd.RepoConfig {
+	urls := make([]string, len(repo.BaseURLs))
+	copy(urls, repo.BaseURLs)
+
+	keys := make([]string, len(repo.GPGKeys))
+	copy(keys, repo.GPGKeys)
+
+	repoConfig := rpmmd.RepoConfig{
+		Id:           repo.Id,
+		BaseURLs:     urls,
+		GPGKeys:      keys,
+		Name:         repo.Name,
+		Metalink:     repo.Metalink,
+		MirrorList:   repo.Mirrorlist,
+		CheckGPG:     repo.GPGCheck,
+		CheckRepoGPG: repo.RepoGPGCheck,
+		Priority:     repo.Priority,
+		Enabled:      repo.Enabled,
+		IgnoreSSL:    !repo.SSLVerify,
+	}
+
+	return repoConfig
 }
--- a/internal/blueprint/repository_customizations.go
+++ b/internal/blueprint/repository_customizations.go
@@ -1,0 +1,35 @@
+package blueprint
+
+import (
+	"fmt"
+)
+
+type RepositoryCustomization struct {
+	Id           string   `json:"id" toml:"id"`
+	BaseURLs     []string `json:"baseurls,omitempty" toml:"baseurls,omitempty"`
+	GPGKeys      []string `json:"gpgkeys,omitempty" toml:"gpgkeys,omitempty"`
+	Metalink     string   `json:"metalink,omitempty" toml:"metalink,omitempty"`
+	Mirrorlist   string   `json:"mirrorlist,omitempty" toml:"mirrorlist,omitempty"`
+	Name         string   `json:"name,omitempty" toml:"name,omitempty"`
+	Priority     *int     `json:"priority,omitempty" toml:"priority,omitempty"`
+	Enabled      *bool    `json:"enabled,omitempty" toml:"enabled,omitempty"`
+	GPGCheck     *bool    `json:"gpgcheck,omitempty" toml:"gpgcheck,omitempty"`
+	RepoGPGCheck *bool    `json:"repo_gpgcheck,omitempty" toml:"repo_gpgcheck,omitempty"`
+	SSLVerify    bool     `json:"sslverify,omitempty" toml:"sslverify,omitempty"`
+	Filename     string   `json:"filename,omitempty" toml:"filename,omitempty"`
+}
+
+func validateCustomRepository(repo *RepositoryCustomization) error {
+	if repo.Id == "" {
+		return fmt.Errorf("Repository ID is required")
+	}
+
+	if len(repo.BaseURLs) == 0 && repo.Mirrorlist == "" && repo.Metalink == "" {
+		return fmt.Errorf("Repository base URL, mirrorlist or metalink is required")
+	}
+
+	if repo.GPGCheck != nil && *repo.GPGCheck && len(repo.GPGKeys) == 0 {
+		return fmt.Errorf("Repository gpg check is set to true but no gpg keys are provided")
+	}
+	return nil
+}
--- a/internal/blueprint/repository_customizations_test.go
+++ b/internal/blueprint/repository_customizations_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/osbuild/osbuild-composer/internal/common"
 	"github.com/stretchr/testify/assert"
 )
 
--- a/internal/blueprint/repository_customizations_test.go
+++ b/internal/blueprint/repository_customizations_test.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 
 	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/fsnode"
+	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -75,6 +77,259 @@ func TestGetCustomRepositories(t *testing.T) {
 				_, err := tt.expectedCustomizations.GetRepositories()
 				assert.Equal(t, tt.wantErr, err)
 			}
+		})
+	}
+}
+
+func TestCustomRepoFilename(t *testing.T) {
+	testCases := []struct {
+		Name         string
+		Repo         RepositoryCustomization
+		WantFilename string
+	}{
+		{
+			Name: "Test default filename #1",
+			Repo: RepositoryCustomization{
+				Id:       "example-1",
+				BaseURLs: []string{"http://example-1.com"},
+			},
+			WantFilename: "example-1.repo",
+		},
+		{
+			Name: "Test default filename #2",
+			Repo: RepositoryCustomization{
+				Id:       "example-2",
+				BaseURLs: []string{"http://example-1.com"},
+			},
+			WantFilename: "example-2.repo",
+		},
+		{
+			Name: "Test custom filename",
+			Repo: RepositoryCustomization{
+				Id:       "example-1",
+				BaseURLs: []string{"http://example-1.com"},
+				Filename: "test.repo",
+			},
+			WantFilename: "test.repo",
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.Name, func(t *testing.T) {
+			got := tt.Repo.getFilename()
+			assert.Equal(t, tt.WantFilename, got)
+		})
+	}
+}
+
+func TestCustomRepoToRepoConfigAndGPGKeys(t *testing.T) {
+	ensureFileCreation := func(file *fsnode.File, err error) *fsnode.File {
+		t.Helper()
+		assert.NoError(t, err)
+		assert.NotNil(t, file)
+		return file
+	}
+	testCases := []struct {
+		Name           string
+		Repos          []RepositoryCustomization
+		WantRepoConfig map[string][]rpmmd.RepoConfig
+		WantGPGKeys    []*fsnode.File
+	}{
+		{
+			Name: "Test no gpg keys, no filenames",
+			Repos: []RepositoryCustomization{
+				{
+					Id:        "example-1",
+					BaseURLs:  []string{"http://example-1.com"},
+					SSLVerify: true,
+				},
+				{
+					Id:        "example-2",
+					BaseURLs:  []string{"http://example-2.com"},
+					SSLVerify: true,
+				},
+			},
+			WantRepoConfig: map[string][]rpmmd.RepoConfig{
+				"example-1.repo": {
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGKeys:  []string{},
+					},
+				},
+				"example-2.repo": {
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+						GPGKeys:  []string{},
+					},
+				},
+			},
+			WantGPGKeys: nil,
+		},
+		{
+			Name: "Test no gpg keys, filenames",
+			Repos: []RepositoryCustomization{
+				{
+					Id:        "example-1",
+					BaseURLs:  []string{"http://example-1.com"},
+					SSLVerify: true,
+					Filename:  "test-1.repo",
+				},
+				{
+					Id:        "example-2",
+					BaseURLs:  []string{"http://example-2.com"},
+					SSLVerify: true,
+					Filename:  "test-2.repo",
+				},
+			},
+			WantRepoConfig: map[string][]rpmmd.RepoConfig{
+				"test-1.repo": {
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGKeys:  []string{},
+					},
+				},
+				"test-2.repo": {
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+						GPGKeys:  []string{},
+					},
+				},
+			},
+			WantGPGKeys: nil,
+		},
+		{
+			Name: "Test remote gpgkeys",
+			Repos: []RepositoryCustomization{
+				{
+					Id:        "example-1",
+					BaseURLs:  []string{"http://example-1.com"},
+					GPGKeys:   []string{"http://example-1.com/gpgkey"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+				{
+					Id:        "example-2",
+					BaseURLs:  []string{"http://example-2.com"},
+					GPGKeys:   []string{"http://example-2.com/gpgkey"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+			},
+			WantRepoConfig: map[string][]rpmmd.RepoConfig{
+				"example-1.repo": {
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGKeys:  []string{"http://example-1.com/gpgkey"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+				"example-2.repo": {
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+						GPGKeys:  []string{"http://example-2.com/gpgkey"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+			},
+			WantGPGKeys: nil,
+		},
+		{
+			Name: "Test inline gpgkeys",
+			Repos: []RepositoryCustomization{
+				{
+					Id:        "example-1",
+					BaseURLs:  []string{"http://example-1.com"},
+					GPGKeys:   []string{"fake-gpg-key-1"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+				{
+					Id:        "example-2",
+					BaseURLs:  []string{"http://example-2.com"},
+					GPGKeys:   []string{"fake-gpg-key-2"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+			},
+			WantRepoConfig: map[string][]rpmmd.RepoConfig{
+				"example-1.repo": {
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGKeys:  []string{"file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-0"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+				"example-2.repo": {
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+						GPGKeys:  []string{"file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-0"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+			},
+			WantGPGKeys: []*fsnode.File{
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-0", nil, nil, nil, []byte("fake-gpg-key-1"))),
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-0", nil, nil, nil, []byte("fake-gpg-key-1"))),
+			},
+		},
+		{
+			Name: "Test multiple inline gpgkeys",
+			Repos: []RepositoryCustomization{
+				{
+					Id:        "example-1",
+					BaseURLs:  []string{"http://example-1.com"},
+					GPGKeys:   []string{"fake-gpg-key-1", "fake-gpg-key-2"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+				{
+					Id:        "example-2",
+					BaseURLs:  []string{"http://example-2.com"},
+					GPGKeys:   []string{"fake-gpg-key-1", "fake-gpg-key-2"},
+					GPGCheck:  common.ToPtr(true),
+					SSLVerify: true,
+				},
+			},
+			WantRepoConfig: map[string][]rpmmd.RepoConfig{
+				"example-1.repo": {
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGKeys:  []string{"file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-0", "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-1"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+				"example-2.repo": {
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+						GPGKeys:  []string{"file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-0", "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-1"},
+						CheckGPG: common.ToPtr(true),
+					},
+				},
+			},
+			WantGPGKeys: []*fsnode.File{
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-0", nil, nil, nil, []byte("fake-gpg-key-1"))),
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-1-1", nil, nil, nil, []byte("fake-gpg-key-2"))),
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-0", nil, nil, nil, []byte("fake-gpg-key-1"))),
+				ensureFileCreation(fsnode.NewFile("/etc/pki/rpm-gpg/RPM-GPG-KEY-example-2-1", nil, nil, nil, []byte("fake-gpg-key-2"))),
+			},
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.Name, func(t *testing.T) {
+			got, _, err := RepoCustomizationsToRepoConfigAndGPGKeyFiles(tt.Repos)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.WantRepoConfig, got)
 		})
 	}
 }
--- a/internal/blueprint/repository_customizations_test.go
+++ b/internal/blueprint/repository_customizations_test.go
@@ -1,0 +1,79 @@
+package blueprint
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetCustomRepositories(t *testing.T) {
+	testCases := []struct {
+		name                   string
+		expectedCustomizations Customizations
+		wantErr                error
+	}{
+		{
+			name: "Test no errors",
+			expectedCustomizations: Customizations{
+				Repositories: []RepositoryCustomization{
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+					},
+					{
+						Id:       "example-2",
+						BaseURLs: []string{"http://example-2.com"},
+					},
+				},
+			},
+			wantErr: nil,
+		},
+		{
+			name: "Test empty id error",
+			expectedCustomizations: Customizations{
+				Repositories: []RepositoryCustomization{
+					{},
+				},
+			},
+			wantErr: fmt.Errorf("Repository ID is required"),
+		},
+		{
+			name: "Test empty baseurl, mirrorlist or metalink error",
+			expectedCustomizations: Customizations{
+				Repositories: []RepositoryCustomization{
+					{
+						Id: "example-1",
+					},
+				},
+			},
+			wantErr: fmt.Errorf("Repository base URL, mirrorlist or metalink is required"),
+		},
+		{
+			name: "Test missing GPG keys error",
+			expectedCustomizations: Customizations{
+				Repositories: []RepositoryCustomization{
+					{
+						Id:       "example-1",
+						BaseURLs: []string{"http://example-1.com"},
+						GPGCheck: common.ToPtr(true),
+					},
+				},
+			},
+			wantErr: fmt.Errorf("Repository gpg check is set to true but no gpg keys are provided"),
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.wantErr == nil {
+				retCustomizations, err := tt.expectedCustomizations.GetRepositories()
+				assert.NoError(t, err)
+				assert.EqualValues(t, tt.expectedCustomizations.Repositories, retCustomizations)
+			} else {
+				_, err := tt.expectedCustomizations.GetRepositories()
+				assert.Equal(t, tt.wantErr, err)
+			}
+		})
+	}
+}
--- a/internal/distro/fedora/distro.go
+++ b/internal/distro/fedora/distro.go
@@ -845,6 +845,12 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 		return nil, err
 	}
 
+	// check if repository customizations are valid
+	_, err = customizations.GetRepositories()
+	if err != nil {
+		return nil, err
+	}
+
 	return nil, nil
 }
 
--- a/internal/distro/fedora/images.go
+++ b/internal/distro/fedora/images.go
@@ -143,6 +143,32 @@ func osCustomizations(
 		// should have been validated before this point.
 		panic(fmt.Sprintf("failed to convert file customizations to fs node files: %v", err))
 	}
+
+	customRepos, err := c.GetRepositories()
+	if err != nil {
+		// This shouldn't happen and since the repos
+		// should have already been validated
+		panic(fmt.Sprintf("failed to get custom repos: %v", err))
+	}
+
+	// This function returns a map of filename and corresponding yum repos
+	// and a list of fs node files for the inline gpg keys so we can save
+	// them to disk. This step also swaps the inline gpg key with the path
+	// to the file in the os file tree
+	yumRepos, gpgKeyFiles, err := blueprint.RepoCustomizationsToRepoConfigAndGPGKeyFiles(customRepos)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert inline gpgkeys to fs node files: %v", err))
+	}
+
+	// add the gpg key files to the list of files to be added to the tree
+	if len(gpgKeyFiles) > 0 {
+		osc.Files = append(osc.Files, gpgKeyFiles...)
+	}
+
+	for filename, repos := range yumRepos {
+		osc.YUMRepos = append(osc.YUMRepos, osbuild.NewYumReposStageOptions(filename, repos))
+	}
+
 	osc.ShellInit = imageConfig.ShellInit
 
 	osc.Grub2Config = imageConfig.Grub2Config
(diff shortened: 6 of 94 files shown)


thozza (commented):
Nice work. I added a few comments 😇 

I think that we can introduce this feature in steps:

1. Allow configuration of YUM repositories which will be baed-into the image, but not used for the image content. This can be workarounded on image-builder side, because it can implement a logic to duplicate the user-provided repos as customizations, in addition to providing them as payload repositories if the user chooses to have this.
2. We can later make the repository to be (optionally?) used also for image content.

In such a case, it is reasonable to let the user to provide the GPG key in the Repo definition as URL first. If we would then like to use the repo for image content, composer could fetch it once and embed it in the manifest for the purpose of the `rpm` stage. Of course only if the repository has `gpgcheck` enabled.

I'm currently working on the "custom files in /etc" feature. Once it is merged, we should have internal tool-agnostic representation of File and Directory with helper functions to generate the needed osbuild stages to create them in the image. It should be then fairly easy to use the same internal functions (that would be used for custom files in /etc) internally to save the user-repo's GPG key in the image and use it with `file:///...` URL in the YUM repo configuration.

IMO this can be introduced in steps and in backward compatible way (e.g. composer could determine if the GPG is provided in the customization "literally" or as an URL and handle it differently).

achilleas-k:
Regarding the GPG key issue: I had a little one-on-one with @kingsleyzissou on Friday to figure out some things about this together and it turns out we can import the keys into the database without needing to write files.  The `org.osbuild.rpm` stage imports keys listed in the options before installing packages.  If we have a key for a repository that will be configured / baked into the image, we can put the key in the OS pipeline's rpm stage options and it will be imported, even if the key isn't used for any of the packages that will be installed at that stage.

kingsleyzissou:
Thanks for updating this PR to add a comment @achilleas-k 🙌. Was going to get to that this morning but you beat me to it

thozza:
> Regarding the GPG key issue: I had a little one-on-one with @kingsleyzissou on Friday to figure out some things about this together and it turns out we can import the keys into the database without needing to write files. The `org.osbuild.rpm` stage imports keys listed in the options before installing packages. If we have a key for a repository that will be configured / baked into the image, we can put the key in the OS pipeline's rpm stage options and it will be imported, even if the key isn't used for any of the packages that will be installed at that stage.
> 
> If the repository is used for packages in the OS rpm stage, the key will be imported anyway.

I'm not a big fan of doing something like this TBH, at least for the following reasons:

- It is not obvious from a repo definition which GPG key it uses and where it can be inspected.
- If the repo is not really used for any packages baked into the image (which is IMO a reasonable use case), then importing the key seems counter-intuitive. Sure, the user will be asked to import it once DNF would need to use it for the first time, but that's OK, because the user gets a chance to verify the key (not that anyone does that).
- If a repository in `/etc/yum.repos.d` has `gpgcheck` enabled, then I think it is reasonable to expect that the GPG key to be used is specified in the repo definition.

internal/blueprint/customizations.go
```diff
@@ -117,6 +118,24 @@ type OpenSCAPCustomization struct {
 	ProfileID  string `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
 }
 
+type RepositoryCustomization struct {
+	Id            string   `json:"id" toml:"id"`
+	BaseURL       []string `json:"baseurl,omitempty" toml:"baseurl,omitempty"`
+	GPGKey        []string `json:"gpgkey,omitempty" toml:"gpgkey,omitempty"`
+	Cost          *int     `json:"cost,omitempty" toml:"cost,omitempty"`
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah cool, I'll update the PR with that.

internal/blueprint/customizations.go
```diff
@@ -117,6 +118,24 @@ type OpenSCAPCustomization struct {
...
+	PayloadGPGKey *string  `json:"payload_gpgkey,omitempty" toml:"payload_gpgkey,omitempty"`
```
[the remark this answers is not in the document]
kingsleyzissou: So this was regarding the issue I was having with the gpgkeys. If we want to add the package to the image at image-build we need to import the gpgkey either with the key or a file pointing to the key. The yum repo file doesn't accept a gpgkey though, so in that case we would need a url to a gpgkey or a file uri. I'm not a fan, but this was a way to provide both in the customizations as part of the WIP.

internal/osbuild/yum_repos_stage.go
```diff
@@ -23,6 +25,32 @@ type YumRepository struct {
...
+		yr.Name = *repo.Name
+	}
+
+	if repo.Mirrorlist != nil {
+		yr.Mirrorlist = *repo.Mirrorlist
+	}
+
+	if repo.Metalink != nil {
+		yr.Metalink = *repo.Metalink
+	}
```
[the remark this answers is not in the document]
kingsleyzissou: Okay cool. Makes sense

internal/osbuild/yum_repos_stage.go
```diff
@@ -23,6 +25,32 @@ type YumRepository struct {
 	RepoGPGCheck   *bool    `json:"repo_gpgcheck,omitempty"`
 }
 
+func NewYumRepository(repo blueprint.RepositoryCustomization) *YumRepository {
```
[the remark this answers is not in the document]
kingsleyzissou: This is something Achilleas and I spoke about on Friday and it totally makes sense. We would have to change the `rpmmd.Repoconfig` to have a slice of `basurl` and `gpgkey`... or stick with a string for `baseurl` and `gpgkey` in the customizations.

internal/blueprint/customizations.go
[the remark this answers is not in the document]
kingsleyzissou: Ahhh... nevermind this previous comment. I missed your comment above about doing this in stages and letting image-builder handle the payload repos. Will fix this up

achilleas-k:
We do this anyway when a user specifies a repo for a payload package.  We import the key into the rpm db but don't have a repo configured for it.  Is it regarded good practice to have a gpg key for each configured repo in `/etc/pki/...`?  If it is, I see your point.

thozza:
> We do this anyway when a user specifies a repo for a payload package. We import the key into the rpm db but don't have a repo configured for it. Is it regarded good practice to have a gpg key for each configured repo in `/etc/pki/...`? If it is, I see your point.

Fair point with the user-repos 🤔 I'm not sure how to determine if it is a good practice and not just my opinion... It just feels weird if a repo definition would specify `gpgcheck` enabled and not have any `gpgkey` specified. So if we can make it so that this does not happen, I would prefer it. I'll probably reach out to the DNF/RPM team and ask them...

achilleas-k:
> I'll probably reach out to the DNF/RPM team and ask them...

Good idea :+1:

kingsleyzissou:
Still to do: GCE YUMRepos. I can start that while we wait for feedback from the dnf team.

I'm piggybacking of the changes @bcl is making in #3230 (I'll drop those commits once the PR has landed)

internal/distro/image_config.go
```diff
@@ -34,6 +35,7 @@ type ImageConfig struct {
 	// Disable documentation
 	ExcludeDocs *bool
 
+	YUMRepos []rpmmd.RepoConfig
```
kingsleyzissou, line 38: Going to change this to a map so we can still accept a filename

internal/manifest/os.go
```diff
@@ -74,8 +79,9 @@ type OSCustomizations struct {
 	// Do not install documentation
 	ExcludeDocs bool
 
-	Groups []users.Group
-	Users  []users.User
+	Groups   []users.Group
+	Users    []users.User
+	YUMRepos []rpmmd.RepoConfig
```
kingsleyzissou, line 84: Same here. Going to change this to a map to accept filename

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ca5a042996396b9ff8fe76f3ed08e704ad11cb9a with the main merge-base 73882b728906eac805682644b75aeba7ff016995).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3651909451/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3651909451/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 0d50e2fdd94c66651d143df61cdd1a249511cb3c with the main merge-base 8624ff65f806bf65790749929679015f59d286e4).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3654463539/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3654463539/artifacts/browse

internal/manifest/os.go
thozza: Alternatively you can define it as a new type in `rpmmd`, which would be IMO also reasonable. Something like:
```go
type YUMRepoFile map[string]RepoConfig
```

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 6bca73a1268f9d32ff33f9e5483d724b85d6e25f with the main merge-base 77bd1cba0848261ad8593597bddbdc6a9ace9b41).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3657577201/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3657577201/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d481e1a77c0c67e50704d06ef2b747b706c0b877 with the main merge-base 7e13d20ca606def258ac3bb80d830b37f03b126d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4023116566/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4023116566/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD f5d6f7f94eb993e42532958efbcd6e169706e74b with the main merge-base 47bb682221a6700a3a1bb760fb59ffe3fafd3281).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4041528082/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4041528082/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 8f76097a0f7055601a00520b58eb7b46750d4c4b with the main merge-base da2e80f99f65bf708e0c7c3a31ee4844846287d6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4101268755/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4101268755/artifacts/browse

internal/rpmmd/repository.go
```diff
@@ -375,3 +378,45 @@ func (packages PackageList) ToPackageInfos() []PackageInfo {
 
 	return results
 }
+
+func (r RepoConfig) MarshalJSON() ([]byte, error) {
+	type aliasType RepoConfig
+	type compatType struct {
+		aliasType
+
+		BaseURL string `json:"Baseurl,omitempty"`
```
kingsleyzissou, line 391: Repo config doesn't actually have any json tags... so I did this for now. Any suggestions welcome.

internal/distro/fedora/distro.go
```diff
@@ -767,6 +767,8 @@ func (t *imageType) Manifest(customizations *blueprint.Customizations,
 		mergedRepos = append(mergedRepos, repo)
 	}
 
+	mergedRepos = append(mergedRepos, customRepos...)
```
kingsleyzissou, line 770: Wasn't too sure how to work around this... this seems a bit meh. But the custom repos get removed otherwise.

internal/rpmmd/repository.go
bcl: I'm fairly sure `RepoConfig` never gets serialized to/from json. It's an internal struct that is filled in manually by `loadRepositoriesFromFile` which uses `repository` as the on-disk json struct.

kingsleyzissou: Ah right, I had to add this in though because the `regression-old-worker` test kept failing with an rpmmd error because it couldn't unmarshal the fields. 

Just to check, is it not getting marshalled here though?
https://github.com/osbuild/osbuild-composer/blob/da2e80f99f65bf708e0c7c3a31ee4844846287d6/internal/worker/json.go#L224

bcl: Ah, yes it would appear so :/ The function says it is temporary, so maybe that will change. But I really have no idea, not having touched the worked code at all.

achilleas-k: It's probably time to drop the `compatJob` part here.  The change it was covering for was made almost a year ago (94c7fda779).  Since da2e80f99f we only care about and test workers from 3 versions back, so I think we're safe to remove it and it's not completely out of scope for this PR I think.

achilleas-k: Looking at it again, it seems that even in the new format we would serialise `rpmmd.RepoConfig` as part of the `rpmmd.PackageSet` in `DepsolveJob`.  That said, we've been serialising `rpmmd.RepoConfig` for ages, since the inception of the `DepsolveJob` 0f90aa9c782cace1fc88c39c7ec752bd1f12d799, so maybe it's fine?

internal/distro/fedora/distro.go
```diff
@@ -856,6 +858,12 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 		return nil, err
 	}
 
+	repos := customizations.GetRepositories()
+	err = blueprint.ValidateCustomRepositories(repos)
```
thozza, line 862: I think that it would be better to validate when we unmarshal from TOML, but this is also OK.

internal/distro/fedora/images.go
```diff
@@ -48,6 +49,7 @@ func osCustomizations(
 	osc.ExtraBaseRepos = osPackageSet.Repositories
 
 	osc.Containers = containers
+	osc.CustomRepos = customRepos
```
thozza, line 52: You should take this from passed blueprint customizations instead, not as a separate function argument.

thozza (changes requested):
I added many comments, sorry. I'll need to do another iteration, but I'm submitting what I have now...

kingsleyzissou:
So to briefly summarise our call, this is what we discussed:
- only save repos to `/etc/yum.repos.d` and not use the custom repos as payload repos (this will greatly simplify the pr)
- resolve url gpgkeys and import them, but keep the url in the final `/etc/yum.repos/d` repo file
- save inline gpgkeys to `/etc/pki` and replace them in the `/etc/yum.repos.d/` with a `file://` uri
- either use a map or specify the filename of the custom repository in the customizations rather than saving all the repos to one file.

internal/rpmmd/repository.go
kingsleyzissou: Because we're simplifying this PR, this technically could be done in a separate PR? Happy to add the json tags and create a PR for that, wdyt?

internal/rpmmd/repository.go
```diff
@@ -423,3 +424,35 @@ func (r *RepoConfig) UnmarshalJSON(data []byte) error {
...
+		Mirrorlist:   customRepo.Mirrorlist,
+		GPGCheck:     customRepo.GPGCheck,
+		RepoGPGCheck: customRepo.RepoGPGCheck,
+		Priority:     customRepo.Priority,
+		Enabled:      customRepo.Enabled,
+		IgnoreSSL:    !customRepo.SSLVerify,
+	}
+
+	return repoConfig
+}
```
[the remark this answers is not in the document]
kingsleyzissou: Okay cool, will move that. That explanation makes sense and hadn't crossed my mind

internal/distro/distro.go
```diff
@@ -121,7 +121,7 @@ type ImageType interface {
 	// specified in the given blueprint; it also returns any warnings (e.g.
 	// deprecation notices) generated by the manifest.
 	// The packageSpecSets must be labelled in the same way as the originating PackageSets.
-	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, containers []container.Spec, seed int64) (Manifest, []string, error)
+	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, containers []container.Spec, customRepos []rpmmd.RepoConfig, seed int64) (Manifest, []string, error)
```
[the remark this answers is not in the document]
kingsleyzissou: I think this was because I was trying to do some cleverness with the file resolver and passing the keys down, but I don't think there'll be any need for this with the simplifications... so I'll revert back to using the customizations.

internal/osbuild/rpm_key_input.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func (RPMKeyInlineInput) isStageInputs() {}
+
+func NewRPMInlineInput(contents []byte) Inputs {
+	input := NewFilesInput(NewFilesInputSourcePlainRef([]string{
+		fmt.Sprintf("%x", sha256.Sum256(contents)),
+	}))
+	return &RPMKeyInlineInput{InlineFile: input}
+}
```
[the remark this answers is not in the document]
kingsleyzissou: Cool, yeah that makes sense. Like I said in the call, might pick your brains on this when I get to it.

internal/rpmmd/repository.go
thozza: Whatever is easier for you...

thozza:
> So to briefly summarise our call, this is what we discussed:
> 
> * only save repos to `/etc/yum.repos.d` and not use the custom repos as payload repos (this will greatly simplify the pr)
> * resolve url gpgkeys and import them, but keep the url in the final `/etc/yum.repos/d` repo file

TBH, I don't this this ^^^ is necessary at all if we are going to not use the custom repository for installing content in the image. DNF will import the key when installing the first package from the custom repo. So we may as well skip importing it at all.

> * save inline gpgkeys to `/etc/pki` and replace them in the `/etc/yum.repos.d/` with a `file://` uri
> * either use a map or specify the filename of the custom repository in the customizations rather than saving all the repos to one file.

internal/osbuild/yum_repos_stage.go
```diff
@@ -68,13 +71,45 @@ type YumReposStageOptions struct {
...
+		Id:           repo.Id,
+		Name:         repo.Name,
+		Mirrorlist:   repo.Mirrorlist,
+		Metalink:     repo.Metalink,
+		BaseURL:      urls,
+		GPGKey:       keys,
+		GPGCheck:     &repo.GPGCheck,
+		RepoGPGCheck: &repo.RepoGPGCheck,
+		Enabled:      &repo.Enabled,
+		Priority:     &repo.Priority,
```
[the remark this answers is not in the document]
kingsleyzissou: I managed to work around this when deserialising the json/toml and only explicitly setting the fields that are there.
The code is slightly more verbose there for the unmarshaling functions, but it means we don't have to change those values to pointers in `rpmmd.RepoConfig`. Lmk what you think

internal/rpmmd/repository.go
kingsleyzissou: Did it here :)

internal/rpmmd/repository.go
```diff
@@ -375,3 +378,45 @@ func (packages PackageList) ToPackageInfos() []PackageInfo {
...
+		BaseURL string `json:"Baseurl,omitempty"`
+	}
+
+	var compatRepo compatType
+	if err := json.Unmarshal(data, &compatRepo); err != nil {
+		return err
+	}
+
+	if compatRepo.BaseURL != "" {
+		compatRepo.BaseURLs = strings.Split(compatRepo.BaseURL, ",")
```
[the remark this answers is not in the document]
kingsleyzissou: Ah  I missed this, I've left it as a list since the `yum_repo_stage` takes as a list of baseurls as an option.
Happy to change that though. But not sure if it makes a huge difference.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD db8df89c564f1242057bf8149805736357e05b4b with the main merge-base a56c70d5053b1c2e5293d68dc3665eefe694829f).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4129223418/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4129223418/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 293bff816214a55e0f602d384a81d9fa323431c4 with the main merge-base 258f45002166ec3539d2aa0f366702fb533776c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4136155346/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4136155346/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 39829b38f77c84cf507630dfecf757befab13eed with the main merge-base 258f45002166ec3539d2aa0f366702fb533776c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4137067107/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4137067107/artifacts/browse

test/data/manifests/rhel_7-x86_64-qcow2_customize-boot.json
```diff
@@ -146,6 +146,22 @@
             "user": 0,
             "group": 0
           }
+        ],
```
kingsleyzissou, line 149: Ah I think I missed a step for the rhel implementation. The yum repo stage is not being added to the generated manifest.

test/data/manifests/centos_8-aarch64-qcow2_customize-boot.json
```diff
@@ -131,6 +131,22 @@
             "user": 0,
             "group": 0
           }
+        ],
+        "repositories": [
```
kingsleyzissou, line 135: Same as above

test/data/manifests/centos_9-aarch64-qcow2_customize-boot.json
```diff
@@ -131,6 +131,22 @@
             "user": 0,
             "group": 0
           }
+        ],
+        "repositories": [
```
kingsleyzissou, line 135: Same as above

test/data/manifests/rhel_7-x86_64-qcow2_customize-boot.json
kingsleyzissou: The `imageConfig.YUMRepos` were overwriting the `osc.YUMRepos`

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD fd61e02f2a3a86117a9147a2dd59040d06a3ed60 with the main merge-base d8342862d5e86a97953cbfc29ee373aca1b5c4a1).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4140459483/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4140459483/artifacts/browse

internal/rpmmd/test/repository_test.go
```diff
@@ -228,14 +228,14 @@ func TestOldWorkerRepositoryCompatMarshal(t *testing.T) {
 		repo     rpmmd.RepoConfig
 	}{
 		{
-			repoJSON: []byte(`{"Name":"fedora","BaseURLs":["http://example.com/fedora"],"Metalink":"","Mirrorlist":"","GPGKeys":null,"GPGCheck":false,"RepoGPGCheck":false,"IgnoreSSL":false,"MetadataExpire":"","RHSM":false,"ImageTypeTags":null,"PackageSets":null,"Baseurl":"http://example.com/fedora"}`),
+			repoJSON: []byte(`{"Name":"fedora","BaseURLs":["http://example.com/fedora"],"Metalink":"","Mirrorlist":"","GPGKeys":null,"GPGCheck":null,"RepoGPGCheck":null,"IgnoreSSL":false,"MetadataExpire":"","RHSM":false,"ImageTypeTags":null,"PackageSets":null,"Baseurl":"http://example.com/fedora"}`),
```
thozza: nitpick: seeing the `null` in the JSON string, I think that it would be good to eventually use dedicated structures for marshaling or at least add JSON tags to the `RepoConfig` struct and make these `omitempty`. However not a blocker for this PR from my PoV.

internal/store/store_test.go
```diff
@@ -444,6 +446,7 @@ func (suite *storeTest) TestNewSourceConfigWithMirrorlist() {
 	myRepoConfig := rpmmd.RepoConfig{
 		Name:       "testRepo",
 		Mirrorlist: "testURL",
+		IgnoreSSL:  false,
```
thozza, line 449: I'm wondering why was this needed, since in this commit (https://github.com/osbuild/osbuild-composer/pull/3228/commits/d34fe71bdded6dd6f3a46c8111ba6da0bb375062) I don't see any change related to `IgnoreSSL` struct member 🤔

internal/blueprint/repository_customizations.go
```diff
@@ -326,3 +330,84 @@ func validateCustomRepository(repo *RepositoryCustomization) error {
...
+func RepoCustomizationsToRepoConfig(repos []RepositoryCustomization) map[string][]rpmmd.RepoConfig {
+	if len(repos) == 0 {
+		return nil
+	}
+
+	m := make(map[string][]rpmmd.RepoConfig)
+
+	for _, repo := range repos {
+		convertedRepo := repo.customRepoToRepoConfig()
+		m[repo.Filename] = append(m[repo.Filename], convertedRepo)
```
thozza, line 385: If I read the implementation correctly, `repo.Filename` may be an empty string. It could make sense to add a method for the `RepositoryCustomization` struct which will generate one if not set explicitly by user. We discussed that we could default to the `<repo_id>.repo`.

internal/blueprint/repository_customizations.go
```diff
@@ -326,3 +330,84 @@ func validateCustomRepository(repo *RepositoryCustomization) error {
...
+	}
+	return files, nil
+}
+
+func RepoCustomizationsToRepoConfig(repos []RepositoryCustomization) map[string][]rpmmd.RepoConfig {
+	if len(repos) == 0 {
+		return nil
+	}
+
+	m := make(map[string][]rpmmd.RepoConfig)
```
thozza, line 381: nitpick: you should probably pass `len(repos)` as the initial capacity of the map.

internal/blueprint/repository_customizations.go
```diff
@@ -326,3 +330,84 @@ func validateCustomRepository(repo *RepositoryCustomization) error {
 	}
 	return nil
 }
+
+// DeepCopy returns a deep copy of the given slice of RepositoryCustomization
+// Since we are replacing the gpgkeys of the customization with the file path
+// we don't want to modify the original slice.
+func DeepCopyCustomRepositories(repos []RepositoryCustomization) []RepositoryCustomization {
```
thozza, line 337: I don't think this is needed. My suggestion would be to:

Merge `GetInlineGPGKeysFromCustomRepositories()` and `RepoCustomizationsToRepoConfig()` into a single function, which returns `(map[string][]rpmmd.RepoConfig, []*fsnode.File, error)`. In a single iteration, you create a new `rpmmd.RepoConfig` which already has deep copy of all values from the repository customization structure. You would then generate a `*fsnode.File` for the repo if the GPG key was provided inline and you would modify the `rpmmd.RepoConfig` to point to a file, before adding it to the map of slices of repo configs. This should not create any issues (AFAICT).

Doing the conversion to internal representations in two steps is the only reason that you need this helper function and it also creates possibility to call them in the wrong order and not have inline GPG keys converted to file paths. Modifying the `rpmmd.RepoConfig` after conversion in a single function will eliminate any need for deep copy since you won't ever modify the customization structure.

Also I would appreciate if you could add unit tests for all of these newly added helper functions.

internal/distro/fedora/images.go
```diff
@@ -143,6 +143,31 @@ func osCustomizations(
...
+	// we need to make a copy of the repos so we don't modify the original
+	customRepos := blueprint.DeepCopyCustomRepositories(c.GetRepositories())
+
+	// get the inline gpg keys from the repos and convert them to fs node files
+	// so we can save them to disk. This step also swaps the inline gpg key
+	// with the path to the file in the os file tree
+	gpgKeyFiles, err := blueprint.GetInlineGPGKeysFromCustomRepositories(customRepos)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert inline gpgkeys to fs node files: %v", err))
+	}
```
thozza, line 157: As I commented elsewhere, I propose to get `gpgKeyFiles` and `yumRepos` in a single function call without the need to make a copy of the customizations.

thozza (commented):
Very nice iteration. Added some suggestions inline...

internal/blueprint/repository_customizations.go
kingsleyzissou: Okay, cool sounds good to me. I will try merge the functions like you suggest and yes, absolutely, I'll add unit tests... just wanted to make sure I was on the right track before doing that :)

internal/rpmmd/test/repository_test.go
kingsleyzissou: Oh I did add JSON tags to RepoConfig, but I did it in a later commit... thinking about this further, I probably should have done that first.

internal/blueprint/repository_customizations.go
kingsleyzissou: So at the moment, I'm setting the filename in the unmarshal functions if nothing is set - it might not the best place to do it and I could make that more explicit, like you suggest.

For reference:
https://github.com/osbuild/osbuild-composer/blob/fd61e02f2a3a86117a9147a2dd59040d06a3ed60/internal/blueprint/repository_customizations.go#L195

internal/rpmmd/test/repository_test.go
thozza: Sorry, I was reviewing the PR commit by commit and thinking about this now, you are right that you added tags in later commit, but I didn't get back to my comment. Adding them first would be probably better.

internal/blueprint/repository_customizations.go
thozza: Ah, I indeed missed that. I think that doing this in the custom unmarshal function may not be the best place. Specifically, because this will solve the problem of filling the not provided filename in CloudAPI where we just copy customizations from the compose request to the blueprint created on the fly https://github.com/osbuild/osbuild-composer/blob/03b093acb024675eab75b7fe266b3e6db2504dc7/internal/cloudapi/v2/handler.go#L142

We do not unmarshal there into the BP customizations so the missing filename would not be filled in in such case.

WRT the filename if it is not provided, I'll let others to comment as well, but I find a bit nicer if we simply use the repo ID as the repo filename.

In addition, the repo IDs must be unique across all configuration files, which is not currently enforced in any way and it would probably make sense to check as it would solve some corner cases. To complicate things more, the user-provided repo ids can conflict with repos specified in the image configuration (currently used only by GCE images).

internal/blueprint/repository_customizations.go
thozza: You are definitely on the right track, you got it right! My suggestion would just simplify the implementation, but the resulting behavior should stay the same 😉

internal/blueprint/repository_customizations.go
kingsleyzissou: Yeah no problem at all... I like your solution better, it's simpler and more elegant and avoids magic. I've fixed that already and will push the changes in a bit :). 

It's also less of headache if we just save them to a repo with their own id rather than grouping all the repos with the same filename together.

thozza:
@kingsleyzissou one thing I realized that's missing is the exposure of this customization through the CloudAPI. I guess that's still on your radar, but wanted to call it out.

internal/rpmmd/test/repository_test.go
kingsleyzissou: Yeah no, I figured you did them commit by commit... I've added the json tags now in the commit you've commented on.

kingsleyzissou: Yeah no, I figured you did them commit by commit... I've added the json tags now in the commit you've commented on.

kingsleyzissou:
> @kingsleyzissou one thing I realized that's missing is the exposure of this customization through the CloudAPI. I guess that's still on your radar, but wanted to call it out.

Yeah it is indeed, I kind of have a WIP that I turned into a debugging PR. I think I'll do the cloudapi stuff in a separate PR.

internal/store/store_test.go
kingsleyzissou: Good catch! I think I made `IgnoreSSL` a pointer and then undid it - but that's my only guess.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 503fee05990655e286057ca9448a8e8a34fc0626 with the main merge-base 03b093acb024675eab75b7fe266b3e6db2504dc7).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4143573579/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4143573579/artifacts/browse

internal/blueprint/repository_customizations.go
```diff
@@ -0,0 +1,326 @@
...
+	Name         string   `json:"name,omitempty" toml:"name,omitempty"`
+	Priority     *int     `json:"priority,omitempty" toml:"priority,omitempty"`
+	Enabled      *bool    `json:"enabled,omitempty" toml:"enabled,omitempty"`
+	GPGCheck     *bool    `json:"gpgcheck,omitempty" toml:"gpgcheck,omitempty"`
+	RepoGPGCheck *bool    `json:"repo_gpgcheck,omitempty" toml:"repo_gpgcheck,omitempty"`
+	SSLVerify    bool     `json:"sslverify,omitempty" toml:"sslverify,omitempty"`
+	Filename     string   `json:"filename,omitempty" toml:"filename,omitempty"`
+}
+
+func (rc *RepositoryCustomization) UnmarshalTOML(data interface{}) error {
```
achilleas-k, line 28: Why do we need the custom unmarshaller?  Is it just for checking required fields and their types?  Would it be easier to check it when reading instead of needing to define our own functions?

tools/test-case-generators/format-request-map.json
```diff
@@ -943,6 +943,21 @@
...
               "group": 0
             }
+          ],
+          "repositories": [
+            {
+              "id": "webmin-noarch",
+              "name": "Webmin - noarch",
+              "baseurls": [
+                "https://download.webmin.com/download/yum"
+              ],
```
achilleas-k, line 953: I'm not sure about adding a repository we don't control (or can rely on) as part of our tests.  I guess it's fine since they're not used during the build, but maybe we can add one of our own (maybe something like `sap-rhui-azure` from our snapshots) or even a fake one (like example.org)?

achilleas-k (commented):
Nice work!  Thanks for working on this for so long through all the changes :)
I have some comments below but they're mostly nitpicks and questions.

internal/blueprint/repository_customizations.go
```diff
@@ -324,3 +327,73 @@ func validateCustomRepository(repo *RepositoryCustomization) error {
...
+	return rc.Filename
+}
+
+func RepoCustomizationsToRepoConfigAndGPGKeyFiles(repos []RepositoryCustomization) (map[string][]rpmmd.RepoConfig, []*fsnode.File, error) {
+	if len(repos) == 0 {
+		return nil, nil, nil
+	}
+
+	repoMap := make(map[string][]rpmmd.RepoConfig, len(repos))
+	gpgKeyFiles := []*fsnode.File{}
```
thozza, line 344: IMO it would be better to declare the variable without assignment (so `var gpgKeyFiles []*fsnode.File`, so that if there are no Files for GPG keys after the conversion, the returned value will be `nil`. Checking the `len()` of such value will work fine, but you'll need to adjust unit tests.

internal/blueprint/repository_customizations_test.go
```diff
@@ -93,3 +96,340 @@ func TestCustomRepoSerialization(t *testing.T) {
...
+	for _, tt := range testCases {
+		t.Run(tt.Name, func(t *testing.T) {
+			got, _, err := RepoCustomizationsToRepoConfigAndGPGKeyFiles(tt.Repos)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.WantRepoConfig, got)
+		})
+	}
+}
+
+func TestRepoCustomizationsToGPGKeyFiles(t *testing.T) {
```
thozza, line 335: Functionally, this is all good. Logically, I think that you could test conversion of repo customizations to `RepoConfig` and `File` in a single test case. However I'm OK with keeping these test cases split.

test/cases/custom-repos.sh
```diff
@@ -0,0 +1,198 @@
...
+        exit 1
+    fi
+}
+
+greenprint "🚀 Checking custom repositories"
+
+REPO_ID="webmin-noarch"
+REPO_NAME="Webmin - noarch"
+REPO_FILENAME="custom.repo"
+REPO_BASEURL="https://download.webmin.com/download/yum"
```
thozza, line 123: As mentioned elsewhere, using `example.com` or some of our own repos would be better.

internal/rpmmd/repository.go
```diff
@@ -18,9 +18,9 @@ type repository struct {
 	Name           string   `json:"name"`
 	BaseURL        string   `json:"baseurl,omitempty"`
 	Metalink       string   `json:"metalink,omitempty"`
-	MirrorList     string   `json:"mirrorlist,omitempty"`
+	Mirrorlist     string   `json:"mirrorlist,omitempty"`
 	GPGKey         string   `json:"gpgkey,omitempty"`
-	CheckGPG       bool     `json:"check_gpg,omitempty"`
+	GPGCheck       bool     `json:"check_gpg,omitempty"`
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah I've been wondering about that. At first it was super confusing for me. But I guess we convert them in different places... so maybe it is better to drop that commit and leave things as is.

internal/rpmmd/repository.go
```diff
@@ -375,3 +378,45 @@ func (packages PackageList) ToPackageInfos() []PackageInfo {
 
 	return results
 }
+
+func (r RepoConfig) MarshalJSON() ([]byte, error) {
+	type aliasType RepoConfig
```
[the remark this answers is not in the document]
kingsleyzissou: Good point, will add that.

internal/blueprint/repository_customizations.go
kingsleyzissou: I think this was a suggestion from @thozza... and then we're doing the validation inside of this function rather than doing it when generating the manifest.

thozza: I suggested this. Doing it in e.g. `GetRepositories()` would be nice, but it does not return any error, but it would not be a problem to add one. The custom unmarshaller is a bit of an overkill, I admit.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ea869b01311f4090e9f13a32dd61f269e5272e73 with the main merge-base edf4f7e879636096fd1ebf591ec08e24e8cdc939).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4147703232/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4147703232/artifacts/browse

internal/blueprint/repository_customizations.go
```diff
@@ -0,0 +1,326 @@
...
+		break
+	default:
+		return fmt.Errorf("TOML unmarshal: name must be a string")
+	}
+
+	switch baseurls := dataMap["baseurls"].(type) {
+	case interface{}:
+		urls, ok := baseurls.([]interface{})
+		if !ok {
+			return fmt.Errorf("TOML unmarshal: baseurl must be a string")
```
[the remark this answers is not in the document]
kingsleyzissou: Think like you mention... I can simplify the json unmarshaller like you've done with the directory customization... this will get rid of that.

internal/blueprint/repository_customizations.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return err
+	}
+
+	*rc = repo
+	return nil
+}
+
+func (rc *RepositoryCustomization) UnmarshalJSON(data []byte) error {
+	var dataMap map[string]interface{}
```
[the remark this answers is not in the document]
kingsleyzissou: Cool cool, I'll update this and simplify.

internal/blueprint/repository_customizations.go
kingsleyzissou: Ah good point, I'll add it to the `GetRepositories()` function instead and return an error

.gitlab-ci.yml
```diff
@@ -571,6 +571,11 @@ filesystem.sh:
   variables:
     SCRIPT: filesystem.sh
 
+custom-repos.sh:
+  extends: .integration
+  variables:
+    SCRIPT: custom-repos.sh
```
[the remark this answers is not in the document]
thozza: Keeping the test here for now is OK from my PoV. We should probably rethink how we test customizations on premise in general...

internal/blueprint/repository_customizations.go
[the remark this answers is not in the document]
thozza: But if you meant doing it in `GetRepositories()`, then yeah, that should be possible and simplify things a lot!

internal/rpmmd/repository.go
```diff
@@ -49,15 +49,21 @@ func (r *RepoConfig) Hash() string {
 	bts := func(b bool) string {
 		return fmt.Sprintf("%T", b)
 	}
+	bpts := func(b *bool) string {
+		if b == nil {
+			return fmt.Sprintf("%T", false)
+		}
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah I think the `bpts` function isn't needed anymore then.

internal/blueprint/repository_customizations.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	if len(repo.BaseURLs) == 0 && repo.Mirrorlist == "" && repo.Metalink == "" {
+		return fmt.Errorf("Repository base URL, mirrorlist or metalink is required")
+	}
+
+	if repo.GPGCheck != nil && *repo.GPGCheck && len(repo.GPGKeys) == 0 {
+		return fmt.Errorf("Repository gpg check is set to true but no gpg keys are provided")
+	}
+	return nil
```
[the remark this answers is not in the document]
kingsleyzissou: I'll add this in the cloudapi PR :)

internal/blueprint/repository_customizations.go
kingsleyzissou: Updated this... what I don't love about it though is that I have to call this twice,
Here (to validate the repos):
https://github.com/osbuild/osbuild-composer/blob/18240dee91343c50bacf2fb1dd102fc0924b54a2/internal/distro/fedora/distro.go#L849

And here:
https://github.com/osbuild/osbuild-composer/blob/18240dee91343c50bacf2fb1dd102fc0924b54a2/internal/distro/fedora/images.go#L147-L152

It feels weird not calling the validation earlier and waiting until the second call and panicking. Would it be better to export the validation function again at that point? Or is this approach okay?

thozza: @kingsleyzissou I think that this approach is OK. As you noted in the comment, we should not reach the second call if there was an error and if we did, it is a programming error.

Going forward, it would probably make sense to add a `Validate()` method to the `blueprint.Customizations` structure and call only that from anywhere (in image definitions, APIs, etc). Such method should then validate each specific customization using its own validator.

internal/blueprint/repository_customizations.go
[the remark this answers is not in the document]
thozza: OK

thozza (approved):
Great work! 👏 

I think that everything has been sorted out and some suggestions will be implemented as a follow-up.

achilleas-k (approved):
Great work.  Thanks for all the effort you put in to this.
**LGTM!**

achilleas-k:
Wow, CI is very unhappy.  Some weird errors showing up.

kingsleyzissou:
@achilleas-k tx!
Yeah I just rebased... the previous iteration broke the api tests

It was changing the `bts` function inside the `rpmmd.Hash` function from `%T` to `%v`.
I've reverted that now (and tested) and CI _should_ go green now

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 86a9217fcfc36373fe3188bcf4710444039660a0 with the main merge-base 2c23894e2a5b3c38f3aec9340f26e15fd1b6cd3d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4155747103/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4155747103/artifacts/browse

PR #3230, opened 2023-01-20
Adding metadata signing check

bcl:
Add support for checking repomd.xml gpg signature to sources.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer-cli-tests/main_test.go
+++ b/cmd/osbuild-composer-cli-tests/main_test.go
@@ -1,4 +1,4 @@
-// +build integration
+//go:build integration
 
 package main
 
@@ -171,7 +171,7 @@ type = "yum-baseurl"
 proxy = "https://proxy-url/"
 check_ssl = true
 check_gpg = true
-gpgkey_urls = ["https://url/path/to/gpg-key"]
+gpgkeys = ["https://url/path/to/gpg-key"]
 `))
 	require.NoError(t, err)
 
--- a/dnf-json
+++ b/dnf-json
@@ -82,6 +82,29 @@ class Solver():
         if "sslclientcert" in desc:
             repo.sslclientcert = desc["sslclientcert"]
 
+        if "check_gpg" in desc:
+            repo.gpgcheck = desc["check_gpg"]
+        if "check_repogpg" in desc:
+            repo.repo_gpgcheck = desc["check_repogpg"]
+        if "gpgkey" in desc:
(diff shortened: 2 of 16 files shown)


dnf-json
```diff
@@ -98,6 +99,9 @@ class Solver():
 
             for key in desc["gpgkeys"]:
                 if key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK-----"):
+                    # Not using with because it needs to be a valid file for the duration. It
+                    # is inside the temporary persistdir so will be cleaned up on exit.
+                    # pylint: disable=consider-using-with
                     keyfile = tempfile.NamedTemporaryFile(dir=keydir, delete=False)
```
croissanne, line 105: As long as `delete=False` you can still use `with...` I think? The file won't be removed on close.

dnf-json
```diff
@@ -82,6 +82,29 @@ class Solver():
         if "sslclientcert" in desc:
             repo.sslclientcert = desc["sslclientcert"]
 
+        if "check_gpg" in desc:
+            repo.gpgcheck = desc["check_gpg"]
+        if "check_repogpg" in desc:
+            repo.repo_gpgcheck = desc["check_repogpg"]
+        if "gpgkey" in desc:
```
croissanne, line 90: Is this kept for compatibility reasons?

cmd/osbuild-composer-cli-tests/main_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
croissanne, line 1: Is the `// +build integration` below it not sufficient?

croissanne (commented):
thank you! lgtm overall

dnf-json
bcl: Possibly, but I think this is clearer to future me :)

dnf-json
bcl: correct, no need to break dnf-json's api.

cmd/osbuild-composer-cli-tests/main_test.go
bcl: Interesting. I didn't add that on purpose. I'm guessing vim-go may have done that...

bcl: Ah, I see. go:build is now preferred by go 1.18 and, so I'll remove the old form.

croissanne (approved):
Lovely!

PR #3238, opened 2023-01-24
tools/appsre-ansible: workaround missing authselect-compat

croissanne:
See PR#3237.


thozza (approved):

croissanne:
Admin merging as this playbook is only run on appsre's infra, and not in our ci.

PR #3239, opened 2023-01-24
Rewrite RHEL 7 image definitions using the new framework

achilleas-k:
Followup to #3120 and #3213.

This is a rewrite of the `distro/rhel7/` package to use the new image definition framework.  I've described and provided context for all changes in individual commit messages.

The changeset is much smaller because of the limited number (2) of image types in RHEL 7.

One notable change is the addition of the `force_autorelabel` option in the global `ImageConfig`.  The RHEL 7 images need to enable the `force_autorelabel` option for the SELinux stage in osbuild.  This option should almost never be used but it was added specifically for RHEL 7.  With the rewrite to the new definitions and the sharing of pipeline code between all distros, we need to add support to all stages of the pipeline generation to be able to enable it.

EDIT: We don't test RHEL 7 in CI.  I did a build test of all three test manifests without issue to confirm the build root isn't missing any required packages.


internal/distro/rhel7/images.go
```diff
@@ -0,0 +1,194 @@
...
+		osc.NTPServers = imageConfig.TimeSynchronization.Servers
+		osc.LeapSecTZ = imageConfig.TimeSynchronization.LeapsecTz
+	}
+
+	// Relabel the tree, unless the `NoSElinux` flag is explicitly set to `true`
+	if imageConfig.NoSElinux == nil || imageConfig.NoSElinux != nil && !*imageConfig.NoSElinux {
+		osc.SElinux = "targeted"
+	}
+
+	if oscapConfig := c.GetOpenSCAP(); oscapConfig != nil {
```
kingsleyzissou, line 131: I think we mentioned this and then forgot to come back to it. We have validation to make sure that oscap customizations don't get set for a RHEL7 image... but we could probably remove this altogether here.

thozza: I think that the eventual goal is to have these functions in `images.go` deduplicated across distros, so making is as similar and doing policy checks on the distro level probably makes more sense. So I would not remove it for el7 in this case.

achilleas-k: I'm good with keeping it in (less work :laughing:)

internal/distro/rhel7/distro.go
```diff
@@ -61,7 +62,7 @@ var distroMap = map[string]distribution{
 		releaseVersion:     "7",
 		modulePlatformID:   "platform:el7",
 		vendor:             "redhat",
-		runner:             "org.osbuild.rhel7",
+		runner:             &runner.RHEL{Major: uint64(7), Minor: uint64(9)},
```
thozza, line 67: nitpick: technically, there is no `rhel79` runner (see https://github.com/osbuild/osbuild/tree/main/runners). However this will work with osbuild v69+ (https://github.com/osbuild/osbuild/pull/996). The SPEC file specifies Requires for osbuild-70, so this should be fine.

internal/distro/rhel7/images.go
thozza: I incline towards having all `osCustomizations` in all distro definitions the same and eventually deduplicated and doing the policy check elsewhere.

internal/manifest/os.go
```diff
@@ -562,7 +593,8 @@ func (p *OS) serialize() osbuild.Pipeline {
 
 	if p.SElinux != "" {
 		pipeline.AddStage(osbuild.NewSELinuxStage(&osbuild.SELinuxStageOptions{
-			FileContexts: fmt.Sprintf("etc/selinux/%s/contexts/files/file_contexts", p.SElinux),
+			FileContexts:     fmt.Sprintf("etc/selinux/%s/contexts/files/file_contexts", p.SElinux),
+			ForceAutorelabel: common.ToPtr(true),
```
thozza: This is part of https://github.com/osbuild/osbuild-composer/pull/3239/commits/6930b567d66e0e164ed556951094d0ac96abc1f1 but I don't see any obvious reason why this change is part of that commit. Why it's needed? 🤔

internal/manifest/vpc.go
```diff
@@ -35,8 +37,13 @@ func NewVPC(m *Manifest,
 func (p *VPC) serialize() osbuild.Pipeline {
 	pipeline := p.Base.serialize()
 
+	formatOptions := osbuild.VPCOptions{
+		Type:      "vpc",
```
thozza, line 41: nitpick: this can be omitted, because `NewQEMUStageOptions` will set it if not set based on the type of the specific `formatOptions` type. But I'm OK with keeping it as is, but in that case you should use `osbuild.QEMUFormatVPC` instead of a string literal.

thozza (commented):
Great to see the work converging. I added a few comments, but noting serious 😉 

Also found a typo in https://github.com/osbuild/osbuild-composer/pull/3239/commits/4344de0fb406f1eb0bf51ba2ab1fbc5338a8ea83 commit message. Specifically `The RHEL 7 vpc subformat in qemu does support force_size so we need to be able to disable it.`  is missing NOT in `..qemu does <HERE> support...` 😎

internal/manifest/os.go
thozza: I think that the change should be completely dropped from that commit, because it is later overwritten by https://github.com/osbuild/osbuild-composer/pull/3239/commits/cf559006da6c50d36c9de0c44e581e89d2d6e7b3

internal/distro/rhel7/distro.go
achilleas-k: With the new runners specified in composer we need to specify a minor version and since we're _technically_ building a 7.9 (specified in the `distro.osVersion`), and given the fallback mechanisms in osbuild now, I think this is more correct than 7.0.

internal/manifest/os.go
achilleas-k: Bad rebase / squash after trying out things that ended up in other commits.  I'll amend.

internal/manifest/vpc.go
achilleas-k: I think omitting and letting it get set automatically is better.  I'll fix it.

achilleas-k:
> Also found a typo in [4344de0](https://github.com/osbuild/osbuild-composer/commit/4344de0fb406f1eb0bf51ba2ab1fbc5338a8ea83) commit message. Specifically `The RHEL 7 vpc subformat in qemu does support force_size so we need to be able to disable it.` is missing NOT in `..qemu does <HERE> support...` sunglasses

Good catch, thanks.

achilleas-k:
- Removed `Type: "vpd"` from `osbuild.VPCOptions{}`
- Dropped `ForceAutorelabel: common.ToPtr(true)` from irrelevant commit.  This is later set to `ForceAutorelabel: p.SELinuxForceRelabel`.
- Fixed typo in commit message.

achilleas-k:
And a quick rebase on main.

internal/distro/rhel7/distro.go
thozza: Oh, I didn't mean that we should use `rhel70` runner, but `rhel7` would be ideal 😅

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 3af99aa8e29b507fbb13cd2a9543ac20b5a76c1f with the main merge-base 2e3dd16220c8858b1767c0d3c0150e8bd24213a4).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3659107510/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3659107510/artifacts/browse

thozza (approved):
👍

PR #3240, opened 2023-01-25
cloudapi: Add subscription option for rhc 

bcl:
This pull request includes:

### DIFF
--- a/internal/cloudapi/v2/openapi.v2.yml
+++ b/internal/cloudapi/v2/openapi.v2.yml
@@ -1103,6 +1103,12 @@ components:
         insights:
           type: boolean
           example: true
+        rhc:
+          type: boolean
+          default: false
+          example: true
+          description: |
+            Optional flag to use rhc to register the system, which also always enables Insights.
     User:
       type: object
       required:
--- a/internal/cloudapi/v2/v2_test.go
+++ b/internal/cloudapi/v2/v2_test.go
@@ -947,6 +947,42 @@ func TestComposeCustomizations(t *testing.T) {
 	}`, "id")
 }
 
+func TestComposeRhcSubscription(t *testing.T) {
+	srv, _, _, cancel := newV2Server(t, t.TempDir(), []string{""}, false, false)
+	defer cancel()
+
+	test.TestRoute(t, srv.Handler("/api/image-builder-composer/v2"), false, "POST", "/api/image-builder-composer/v2/compose", fmt.Sprintf(`
+	{
+		"distribution": "%s",
+		"customizations": {
+			"subscription": {
+				"organization": "2040324",
+				"activation_key": "my-secret-key",
+				"server_url": "subscription.rhsm.redhat.com",
+				"base_url": "http://cdn.redhat.com/",
+				"insights": false,
+				"rhc": true
+			},
+			"packages": [ "pkg1", "pkg2" ]
+		},
+		"image_request":{
+			"architecture": "%s",
+			"image_type": "aws",
+			"repositories": [{
+				"baseurl": "somerepo.org",
+				"rhsm": false
+			}],
+			"upload_options": {
+				"region": "eu-central-1"
+			}
+		 }
+	}`, test_distro.TestDistroName, test_distro.TestArch3Name), http.StatusCreated, `
+	{
+		"href": "/api/image-builder-composer/v2/compose",
+		"kind": "ComposeId"
+	}`, "id")
+}
+
 func TestImageTypes(t *testing.T) {
 	srv, _, _, cancel := newV2Server(t, t.TempDir(), []string{""}, false, false)
 	defer cancel()
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -189,6 +189,18 @@ func (p *OS) getPackageSetChain() []rpmmd.PackageSet {
 		packages = append(packages, "openscap-scanner", "scap-security-guide")
 	}
 
+	// Make sure the right packages are included for subscriptions
+	// rhc always uses insights, and depends on subscription-manager
+	// non-rhc uses subscription-manager and optionally includes Insights
+	if p.Subscription != nil {
+		packages = append(packages, "subscription-manager")
+		if p.Subscription.Rhc {
+			packages = append(packages, "rhc", "insights-client")
+		} else if p.Subscription.Insights {
+			packages = append(packages, "insights-client")
+		}
+	}
+
 	chain := []rpmmd.PackageSet{
 		{
 			Include:      append(packages, p.ExtraBasePackages...),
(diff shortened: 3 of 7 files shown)


internal/manifest/os.go
```diff
@@ -465,12 +465,16 @@ func (p *OS) serialize() osbuild.Pipeline {
 	}
 
 	if p.Subscription != nil {
-		commands := []string{
-			fmt.Sprintf("/usr/sbin/subscription-manager register --org=%s --activationkey=%s --serverurl %s --baseurl %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl, p.Subscription.BaseUrl),
```
croissanne, line 469: afaik we always need subscription-manager as a base.

internal/manifest/os.go
```diff
@@ -465,12 +465,16 @@ func (p *OS) serialize() osbuild.Pipeline {
...
-		}
+		var commands []string
 		if p.Subscription.Insights {
+			// When using Insights, register using rhc instead of subscription-manager
+			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
```
croissanne, line 471: I think we might need to install rhc, as we don't install it by default. So if rhc would be specified maybe include it, wdyt @achilleas-k ? Otherwise include it by default but idk if that's ok.

croissanne (commented):
Thank you!

I'd expose it as an additional customization. Currently we can specify insights true/false, I'd add something similar for rhc. There might be cases where users want subman + insights, but not rhc, etc...

And it's kinda like a layered thing, rhc requires insights requires subman.

internal/manifest/os.go
bcl: The way I read the documentation is that it replaces it (or uses it behind the scenes). The rhc connect arguments are pretty similar, except for not having a baseurl. But without actually being able to run this I'm not 100% sure :)

internal/manifest/os.go
bcl: Probably so. It doesn't look like the code does anything to check for or add packages based on the Subscriptions being populated though. So we need to work that out so that it's not possible to end up with broken images (commands added but no supporting packages).

bcl:
> Currently we can specify insights true/false, I'd add something similar for rhc.

Sure that can be done as well, although the jira issue (HMS-198) made it sound like this was a replacement and that it should only use subscription manager if insights wasn't selected. It's also confusing that rhc remediation requires yet another thing (smart management subscription) that we're not setting up.

teg:
@croissanne should it be an enum rather than two bools? That way we can make sure we only expose combinations that make sense?

I'm not really sure why we would use insight client at all, did we get that confirmed from PM?

internal/manifest/os.go
croissanne: ah I just tried it out, it'll just do the subman + insights steps for you. So you're right.

internal/manifest/os.go
achilleas-k: We do add packages conditionally here.  The `getPackageSetChain()` function in this file does exactly that: checks what options are enabled and adds packages to the OS to support them:
https://github.com/osbuild/osbuild-composer/blob/356f261bcd9f3e36b9109e9c82b8c3c9d3875ac3/internal/manifest/os.go#L162-L211

achilleas-k: All that said, I'm not against `rhc` being installed unconditionally on RHEL.
In fact this kb article [0] implies it should be available on 8.4+.

[0] https://access.redhat.com/articles/6538061

bcl: I went with an optional rhc bool, and adding required packages as needed.

bcl:
Any suggestions for adding testing for this?

croissanne:
> Any suggestions for adding testing for this?

We could add it to the subscription block:
https://github.com/osbuild/osbuild-composer/blob/356f261bcd9f3e36b9109e9c82b8c3c9d3875ac3/test/cases/api.sh#L317-L331

And then verify/call `rhc` in instancecheck:
https://github.com/osbuild/osbuild-composer/blob/356f261bcd9f3e36b9109e9c82b8c3c9d3875ac3/test/cases/api/common/common.sh#L18-L46

bcl:
I'd rather test things sooner than later, so I figured out how to add some tests to manifest (which looks like it could use quite a bit more testing coverage). It may also be useful to add something to api.sh but it doesn't look very flexible. We would want to test all  4 subscription states and it looks like it can only handle one.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD f2cde0dba4999503899b859aca1b63486b22f2ae with the main merge-base d877b7419e5f9b6217f93fdd7c749137ff0cf10d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3707346360/artifacts/browse) as `manifests.diff`.

**NOTE:** Manifest generation on merge-base with `main` (d877b7419e5f9b6217f93fdd7c749137ff0cf10d) failed.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3707346360/artifacts/browse

achilleas-k (dismissed):
LGTM thanks.
The `os_test` is a nice addition.  We should probably do more things like this.

achilleas-k:
> > @croissanne should it be an enum rather than two bools? That way we can make sure we only expose combinations that make sense?
> > I'm not really sure why we would use insight client at all, did we get that confirmed from PM?
> 
> rhc will automatically enable insights-client for you. An enum would make more sense but unsure how to not break the api here.

I've approved but holding off on merging because this topic wasn't finalised.

internal/cloudapi/v2/v2_test.go
```diff
@@ -947,6 +947,42 @@ func TestComposeCustomizations(t *testing.T) {
...
+	{
+		"distribution": "%s",
+		"customizations": {
+			"subscription": {
+				"organization": "2040324",
+				"activation_key": "my-secret-key",
+				"server_url": "subscription.rhsm.redhat.com",
+				"base_url": "http://cdn.redhat.com/",
+				"insights": false,
+				"rhc": true
```
croissanne, line 964: I think this case should result in a 400. As rhc will automatically subscribe you to insights.

internal/manifest/os.go
```diff
@@ -480,13 +492,27 @@ func (p *OS) serialize() osbuild.Pipeline {
...
-			fmt.Sprintf("/usr/sbin/subscription-manager register --org=%s --activationkey=%s --serverurl %s --baseurl %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl, p.Subscription.BaseUrl),
-		}
-		if p.Subscription.Insights {
+		var commands []string
+		if p.Subscription.Rhc {
+			// Use rhc for registration instead of subscription manager
+			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
+
+			// Always enable Insights when using rhc
 			commands = append(commands, "/usr/bin/insights-client --register")
```
croissanne, line 506: rhc will automatically register with insights. So this isn't necessary afaik. Can't hurt though.

croissanne (commented):
Ok with merging, but I think the combo of insights: false  + rhc: true should result in a 400 error.

internal/cloudapi/v2/v2_test.go
bcl: The way I was thinking of it is that rhc overrides any insights selection. If it returns an error, and requires the front end to set it to true that implies that the frontend has control over it. I also documented it that way in the openapi yaml just so it would be clear.

I can change it if you're sure, but I think it is fine as it is.

internal/manifest/os.go
bcl: ok. Could have sworn I read that sequence somewhere, but I can't find it again.

internal/cloudapi/v2/v2_test.go
croissanne: Ah sorry, I should've read the description on the endpoint. Then it's fine.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1103,6 +1103,12 @@ components:
         insights:
           type: boolean
           example: true
+        rhc:
+          type: boolean
+          default: false
+          example: true
+          description: |
+            Optional flag to use rhc to register the system, which also always enable Insights.
```
croissanne, line 1111: s/enable/enables

croissanne (approved):
lgtm! Thank you!

Issue #3220, opened 2023-01-16
AArch64 iot-raw-image should use '0x06' for ESP partition type (Raspberry Pi compatability) 


paulwhalen:
**Describe the bug**
To be bootable on the Raspberry Pi family of SBC's, the ESP partition should be formatted as '0x06'. 

Fedora 37 disk Image (produced with ImageFactory):
```
fdisk -l Fedora-IoT-37-20220910.0.aarch64.raw
Disk Fedora-IoT-37-20220910.0.aarch64.raw: 4 GiB, 4294967296 bytes, 8388608 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x97a847b7

Device                                Boot   Start     End Sectors  Size Id Type
Fedora-IoT-37-20220910.0.aarch64.raw1 *       2048 1028095 1026048  501M  6 FAT16
Fedora-IoT-37-20220910.0.aarch64.raw2      1028096 3125247 2097152    1G 83 Linux
Fedora-IoT-37-20220910.0.aarch64.raw3      3125248 8386559 5261312  2.5G 83 Linux
```
Fedora 38 disk image (produced with Osbuild):
```
fdisk -l Fedora-IoT-38.20230112.0-20230112.0.aarch64.raw
Disk Fedora-IoT-38.20230112.0-20230112.0.aarch64.raw: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: D209C89E-EA5E-4FBD-B161-B461CCE297E0

Device                                             Start      End  Sectors  Size Type
Fedora-IoT-38.20230112.0-20230112.0.aarch64.raw1    2048  1028095  1026048  501M EFI System
Fedora-IoT-38.20230112.0-20230112.0.aarch64.raw2 1028096  3125247  2097152    1G Linux filesystem
Fedora-IoT-38.20230112.0-20230112.0.aarch64.raw3 3125248 20971486 17846239  8.5G Linux filesystem
```

**Environment**
 * Fedora Koji 

**Additional context**
Bug with more context:
[...]

PR #3246, opened 2023-01-26
iot-raw-image: partitioning changes

supakeen:
I bring to you adjustments to the partioning for the `iot-raw-image` image type as requested in #3220.

The changes here are to:

1. Change the partition table for the `iot-raw-image` to be a DOS partition table.
2. Set a non-standard ID on the bootable partition.

When reviewing this PR please confirm that the partition layout of a generated image and its size is the same as those currently being built by ImageFactory: https://getfedora.org/en/iot/download/

3. Changes were made to the partition table and disk logic, it is now possible to pass `RequiredSizes` through `ImageOptions` to override or nullify our default standard sizes for partitions. I'm not entirely happy with the approach and would love a nicer idea for this.

Some output:

```
user@desktopa day € fdisk -l iot/x86_64.raw  # downloaded from fedoraproject.org
Disk iot/x86_64.raw: 4 GiB, 4294967296 bytes, 8388608 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x3ba9402d

Device          Boot   Start     End Sectors  Size Id Type
iot/x86_64.raw1 *       2048 1028095 1026048  501M  6 FAT16
iot/x86_64.raw2      1028096 3125247 2097152    1G 83 Linux
iot/x86_64.raw3      3125248 8386559 5261312  2.5G 83 Linux
```
and
```
user@desktopa day € fdisk -l 47cb114a-959d-4ac4-a5f3-1ed83b18d8b7-image.raw  # generated and freshly baked on this PR
Disk 47cb114a-959d-4ac4-a5f3-1ed83b18d8b7-image.raw: 4 GiB, 4294967296 bytes, 8388608 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x3ba9402d

Device                                          Boot   Start     End Sectors  Size Id Type
47cb114a-959d-4ac4-a5f3-1ed83b18d8b7-image.raw1 *       2048 1028095 1026048  501M  6 FAT16
47cb114a-959d-4ac4-a5f3-1ed83b18d8b7-image.raw2      1028096 3125247 2097152    1G 83 Linux
47cb114a-959d-4ac4-a5f3-1ed83b18d8b7-image.raw3      3125248 8388607 5263360  2.5G 83 Linux
```

### DIFF
--- a/internal/disk/disk_test.go
+++ b/internal/disk/disk_test.go
@@ -74,7 +74,7 @@ func TestDisk_DynamicallyResizePartitionTable(t *testing.T) {
 	// math/rand is good enough in this case
 	/* #nosec G404 */
 	rng := rand.New(rand.NewSource(0))
-	newpt, err := NewPartitionTable(&pt, mountpoints, 1024, false, rng)
+	newpt, err := NewPartitionTable(&pt, mountpoints, 1024, false, nil, rng)
 	assert.NoError(t, err)
 	assert.GreaterOrEqual(t, newpt.Size, expectedSize)
 }
@@ -449,7 +449,7 @@ func TestCreatePartitionTable(t *testing.T) {
 	for ptName := range testPartitionTables {
 		pt := testPartitionTables[ptName]
 		for bpName, bp := range testBlueprints {
-			mpt, err := NewPartitionTable(&pt, bp, uint64(13*MiB), false, rng)
+			mpt, err := NewPartitionTable(&pt, bp, uint64(13*MiB), false, nil, rng)
 			assert.NoError(err, "Partition table generation failed: PT %q BP %q (%s)", ptName, bpName, err)
 			assert.NotNil(mpt, "Partition table generation failed: PT %q BP %q (nil partition table)", ptName, bpName)
 			assert.Greater(mpt.GetSize(), sumSizes(bp))
@@ -475,12 +475,12 @@ func TestCreatePartitionTableLVMify(t *testing.T) {
 
 			if tbp != nil && (ptName == "btrfs" || ptName == "luks") {
 				assert.Panics(func() {
-					_, _ = NewPartitionTable(&pt, tbp, uint64(13*MiB), true, rng)
+					_, _ = NewPartitionTable(&pt, tbp, uint64(13*MiB), true, nil, rng)
 				}, fmt.Sprintf("PT %q BP %q: should panic", ptName, bpName))
 				continue
 			}
 
-			mpt, err := NewPartitionTable(&pt, tbp, uint64(13*MiB), true, rng)
+			mpt, err := NewPartitionTable(&pt, tbp, uint64(13*MiB), true, nil, rng)
 			assert.NoError(err, "PT %q BP %q: Partition table generation failed: (%s)", ptName, bpName, err)
 
 			rootPath := entityPath(mpt, "/")
@@ -588,7 +588,7 @@ func TestMinimumSizes(t *testing.T) {
 
 	for idx, tc := range testCases {
 		{ // without LVM
-			mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), false, rng)
+			mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), false, nil, rng)
 			assert.NoError(err)
 			for mnt, minSize := range tc.ExpectedMinSizes {
 				path := entityPath(mpt, mnt)
@@ -602,7 +602,7 @@ func TestMinimumSizes(t *testing.T) {
 		}
 
 		{ // with LVM
-			mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), true, rng)
+			mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), true, nil, rng)
 			assert.NoError(err)
 			for mnt, minSize := range tc.ExpectedMinSizes {
 				path := entityPath(mpt, mnt)
@@ -689,7 +689,7 @@ func TestLVMExtentAlignment(t *testing.T) {
 	}
 
 	for idx, tc := range testCases {
-		mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), true, rng)
+		mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), true, nil, rng)
 		assert.NoError(err)
 		for mnt, expSize := range tc.ExpectedSizes {
 			path := entityPath(mpt, mnt)
@@ -718,7 +718,7 @@ func TestNewBootWithSizeLVMify(t *testing.T) {
 		},
 	}
 
-	mpt, err := NewPartitionTable(&pt, custom, uint64(3*GiB), true, rng)
+	mpt, err := NewPartitionTable(&pt, custom, uint64(3*GiB), true, nil, rng)
 	assert.NoError(err)
 
 	for idx, c := range custom {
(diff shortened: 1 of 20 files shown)


ondrejbudai:
My suggestion is to put the non-usual `RequiresSizes` in the base partition table of the iot raw image. This way, we wouldn't need to pass them as a parameter to the `NewPartitionTable`, which simplifies the last commit a lot. I also think that `imageOptions` are not the right place to define them since these are meant to for customizing options per-image, not per-image type. I would put them there only if we were to expose them in one of our APIs.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 7e49b75d89738c7b06681f0a5666f4323719ea7d with the main merge-base fe7860778c8ac3c604d26a1550425474a65ead8e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3664761101/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3664761101/artifacts/browse

achilleas-k:
The linked issue is only about aarch64.  Is there a reason to change the x86 partitioning as well?

supakeen:
> The linked issue is only about aarch64. Is there a reason to change the x86 partitioning as well? Is it to copy the images built by imagefactory?

Yes, this partitioning scheme is used on both current x86 and aarch64 raw images.

```
user@desktopa iot € fdisk -l x86_64.raw
Disk x86_64.raw: 4 GiB, 4294967296 bytes, 8388608 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x3ba9402d

Device      Boot   Start     End Sectors  Size Id Type
x86_64.raw1 *       2048 1028095 1026048  501M  6 FAT16
x86_64.raw2      1028096 3125247 2097152    1G 83 Linux
x86_64.raw3      3125248 8386559 5261312  2.5G 83 Linux
```

achilleas-k:
> My suggestion is to put the non-usual `RequiresSizes` in the base partition table of the iot raw image. This way, we wouldn't need to pass them as a parameter to the `NewPartitionTable`, which simplifies the last commit a lot. I also think that `imageOptions` are not the right place to define them since these are meant to for customizing options per-image, not per-image type. I would put them there only if we were to expose them in one of our APIs.

Fully agree and given the big ol' `TODO` there, the hard-coded required sizes in the function were meant to be removed.  I think making them a property of the partition table is a good idea and I'd go as far as to move all of them to the partition tables in each distro and have no requirements by default in the function.

supakeen:
@ondrejbudai, @achilleas-k I'm not going to push back on putting this in the partition table. I don't like the place it is at right now. However to me it feels like the partition tables are also the wrong place.

The partition table is shared across multiple images and I feel that different image types could have very different constraints on minimum mountpoint sizes (e.g. an ostree-based image might want a larger /var since it puts everything in there?).

ondrejbudai:
I see your point, required sizes indeed seem more connected to an image type than its partition table. In the end, a partition table is just a way to store data defined by an image type. Thus, an image type itself should define how much space does it need for its content. @thozza @achilleas-k Does this sound reasonable?

achilleas-k:
> The partition table is shared across multiple images and I feel that different image types could have very different constraints on minimum mountpoint sizes

I feel that image types that share a partition table would have the same or similar constraints and they can be overriden on the base PT if necessary.

That said
> However to me it feels like the partition tables are also the wrong place.
you might be right.  I keep flip-flopping on whether it makes sense to have it as part of the PT or not.
My reasoning for having it as a property of the base PT is that we can think of it as a base requirement of the partition table.  In other words, if you don't override the requirements explicitly, the PT will take any modifications the user asks for but maintain those constraints.

On the other hand, it does also make sense to be an image type property, much like the `defaultSize`.  That's a modification to the PT that's made at the image type level.  It's taken into account when modifying the partitions, it can be overriden externally, and it's a parameter to the PT creation function.

Given @ondrejbudai's preference between these two ideas, let's move it to the image type so that the NePT looks like
```
	return disk.NewPartitionTable(&basePartitionTable, mountpoints, imageSize, lvmify, t.RequiredSizes, rng)
```

Should we pull the default out of the function and onto all the image types?  We can do it in a followup if you want.

internal/disk/disk_test.go
```diff
@@ -971,3 +971,118 @@ func TestEnsureDirectorySizes(t *testing.T) {
...
+			ExpectedMinSizes: map[string]uint64{
+				"/opt": 10 * GiB,
+				"/":    1 * GiB,
+			},
+		},
+	}
+
+	for idx, tc := range testCases {
+		{ // without LVM
+			mpt, err := NewPartitionTable(&pt, tc.Blueprint, uint64(3*GiB), false, map[string]uint64{"/": 1 * GiB, "/usr": 3 * GiB}, rng)
```
thozza, line 1061: So this is a bit confusing. Because of the `/usr` is not specified as a filesystem customization and ends up on the same partition as the root `/`, then the root partition will be 1 GiB, but `/usr` won't be. So I'm wondering it the code should use the biggest size for a mountpoint pass in the `RequiredSizes` if it won't end up on a dedicated partition. 🤔

thozza: Alternatively, we should explain the behavior in this regard in a comment (which I hope I didn't miss).

thozza (commented):
I added one question. Otherwise this looks all good. Happy to approve afterwards.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD b9664777fb80a084072170ee269a833ded8786fe with the main merge-base 0e4a5b34b214a6496b22f808fae402f88bebe6a2).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3896300999/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3896300999/artifacts/browse

thozza:
After properly reading all the comments I would say that the base partition table is also considered a property of the image type. Although we share the base partition table among multiple image types, this is mostly our desire. I think that having the required sizes defined as part of the base partition table does make some sense. Unfortunately it would force us to define multiple base partition tables for various image types, in which the only difference would be the required sizes.

That said, I think that defining the required sizes per mount point in the image type and passing this to `NewPartitionTable()`, as implemented now, makes sense and I like it more. It will allow us to share the same base partition table among multiple image types, while being able to define different required sizes for various mount points on the image type level.

supakeen:
@thozza Thanks, I updated this PR with the final changes as requested so this is now ready to actually have this discussion; @ondrejbudai, @achilleas-k do your previous comments still stand?

achilleas-k (approved):
LGTM

internal/disk/disk_test.go
thozza: @supakeen could you please comment on this before we merge the PR?

supakeen: The current comment says: `the sum of their subdirectories plus their own sizes.` I feel that that is clear enough that having a 2 GiB /usr inside a 1 GiB / ends up with a 3 GiB total /.

thozza: @supakeen the referenced comment looks good to me, but the confusing thing is the last test case in this test. Maybe I'm not reading it correctly 😇 

```go
		{ // no separate /usr and no size for / -> / gets sum of default sizes for / and /usr
			Blueprint: []blueprint.FilesystemCustomization{
				{
					Mountpoint: "/opt",
					MinSize:    10 * GiB,
				},
			},
			ExpectedMinSizes: map[string]uint64{
				"/opt": 10 * GiB,
				"/":    1 * GiB,  // <-- here I would expect 4 GiB ( 1 GiB for "/' + 3 GiB for "/usr") 
			},
		},

```

I'm considering

```go
map[string]uint64{"/": 1 * GiB, "/usr": 3 * GiB}
```

passed to `NewPartitionTable()` as `requiredSizes` in this test...

supakeen: Right, so, it is indeed a bit confusing. `/` does end up being 4 GiB in that testcase which is larger than the ExpectedMinSizes so the test passes. I'll up the expected minsize to 4 GiB there so it actually tests what it's supposed to test (1 GiB `/` + 3 GiB `/usr`).

internal/distro/distro.go
```diff
@@ -125,10 +125,11 @@ type ImageType interface {
...
 type ImageOptions struct {
-	Size         uint64
-	OSTree       OSTreeImageOptions
-	Subscription *SubscriptionImageOptions
-	Facts        *FactsImageOptions
+	Size          uint64
+	OSTree        OSTreeImageOptions
+	Subscription  *SubscriptionImageOptions
+	Facts         *FactsImageOptions
+	RequiredSizes map[string]uint64
```
thozza, line 132: Sorry, but I noticed this just now. 😕 

We tend to use `ImakeOptions` that are specific to the build and usually provided by users (except for `Facts`). I do not expect this to be the case for `RequiredSizes`. I thought that the `RequiredSizes` is supposed to be the image-type-specific property, therefore it would IMO make sense to make it part of the `imageType` struct instead.

thozza (commented):
Thanks for the change in the test case, it now makes sense.

Unfortunately, I noticed one more thing that I commented on. I apologize for noticing it just now.

internal/distro/distro.go
thozza: And consequently define the empty `map` in here https://github.com/osbuild/osbuild-composer/blob/4023b61e1a56a26f27388bb03724028ce25be2e2/internal/distro/fedora/distro.go#L154-L171 instead of the `iotRawImage()`.

supakeen: Agreed, and done.

thozza (approved):
Great work 😍  Thanks for your patience 😇

thozza (dismissed):
I realized that you added the `requiredPartitionSizes` only to the Fedora distro, while we aim to keep those structures the same in all distros. So the same change should be reflected in all distributions.

I'm OK with a follow-up though...

internal/distro/fedora/distro.go
```diff
@@ -160,14 +160,15 @@ var (
...
-		basePartitionTables: iotBasePartitionTables,
+		defaultSize:            4 * common.GibiByte,
+		rpmOstree:              true,
+		bootable:               true,
+		image:                  iotRawImage,
+		buildPipelines:         []string{"build"},
+		payloadPipelines:       []string{"image-tree", "image", "xz"},
+		exports:                []string{"xz"},
+		basePartitionTables:    iotBasePartitionTables,
+		requiredPartitionSizes: map[string]uint64{},
```
ondrejbudai, line 171: What about a small comment above this line to highlight that the empty map is actually needed here? It might make the code less surprising for future selfs?

internal/disk/partition_table.go
```diff
@@ -17,9 +17,11 @@ type PartitionTable struct {
 
 	SectorSize   uint64 // Sector size in bytes
 	ExtraPadding uint64 // Extra space at the end of the partition table (sectors)
+
+	RequiredSizes map[string]uint64
```
ondrejbudai, line 21: Does this need to be a property of `partitionTable`. I think it doesn't provide any value: `NewPartitionTable` uses it basically as a local variable and `Clone` copies it over, but I don't think there's a code path that would actually use the variable. Please correct me if I'm missing something.

ondrejbudai (commented):
I'm very happy how this turned out, I just have two minor comments inline.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 56b5450a7f3d1c30883f6d23cb301e74dd5a677e with the main merge-base 4023b61e1a56a26f27388bb03724028ce25be2e2).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3937927352/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3937927352/artifacts/browse

internal/distro/fedora/distro.go
supakeen: Done.

internal/disk/partition_table.go
supakeen: Done.

ondrejbudai (approved):
LGTM! Thanks!

supakeen:
@ondrejbudai Sorry, had to regen manifests after the rebase since the previous f39 manifests (when the repo was added) were generated on main and they have different partition sizes here.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 80c3b4913e29946a5a3a6e53a8edb64820e1c8da with the main merge-base 4fee181fecb055a46fdbd3c8dab8b8f7b08572fa).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3953604856/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3953604856/artifacts/browse

ondrejbudai:
Fedora tests passed, merging.

internal/distro/fedora/partition_tables.go
```diff
@@ -155,27 +155,17 @@ var iotBasePartitionTables = distro.BasePartitionTableMap{
 		},
 	},
 	distro.Aarch64ArchName: disk.PartitionTable{
-		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
-		Type: "gpt",
+		UUID: "0xc1748067",
+		Type: "dos",
```
runcom, line 159: just an heads up, this makes any other arm device that I have ubootable - i.e. on a rockpro64 this doesn't boot, but it boots using the "old" method

runcom: I'll tset with https://github.com/osbuild/osbuild-composer/pull/3374 perhaps?

supakeen: @runcom Yes please, I'm doing the same.

PR #3248, opened 2023-01-27
osbuild: unify implementations of files input 🧹

thozza:
This PR is a pre-work to support custom files in `/etc`, because the copy stage implementation will need to be extended to support files input with origin source. Instead of implementing 6th version of it, I decided to rework files input implementation, which we wanted to do anyway (me and @achilleas-k).

Rework files input implementation to support all reference types supported by the input schema. Also implement helper functions to generate supported reference types. In some cases, the reference supports e.g. referencing multiple pipelines in the stage inputs, but this is currently not implemented, since no pipeline in composer uses it.

Rework the stage inputs implementation of the following stages to use the common files input:

- XZ stage
- FDO stage
- QEMU stage
- RPM stage
- Ignition stage

This rework should not have any effect on manifests produced by osbuild-composer.

This pull request includes:

### DIFF
--- a/internal/osbuild/files_input.go
+++ b/internal/osbuild/files_input.go
@@ -13,7 +13,7 @@ type FilesInputs struct {
 
 func (FilesInputs) isStageInputs() {}
 
-func NewFilesInputs(references FilesInputReferences) *FilesInputs {
+func NewFilesInputs(references FilesInputRef) *FilesInputs {
 	return &FilesInputs{
 		File: NewFilesInput(references),
 	}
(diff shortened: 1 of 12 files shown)


7flying:
(just butting in to say that this makes the files input handling more easy to understand, and it would had made my life easier handling the Ignition stuff, so thanks!!)

internal/osbuild/files_input.go
```diff
@@ -60,46 +43,233 @@ func (f *FilesInput) UnmarshalJSON(data []byte) error {
...
+//
+//	`[
+//		"sha256:<sha256sum>",
+//		...
+//	]`
+type FilesInputSourcePlainRef []string
+
+func (*FilesInputSourcePlainRef) isFilesInputRef() {}
+
+func NewFilesInputSourcePlainRef(sha256Sums []string) FilesInputRef {
```
7flying, line 196: Nit: I would add the same comment that we have below, saying that the items of the array do not need to have any prefix, aka no `sha256:`

internal/osbuild/files_input.go
```diff
@@ -60,46 +43,233 @@ func (f *FilesInput) UnmarshalJSON(data []byte) error {
...
+}
+
+type FilesInputSourceArrayRefEntry struct {
+	ID      string                   `json:"id"`
+	Options *FilesInputSourceOptions `json:"options,omitempty"`
+}
+
+// NewFilesInputSourceArrayRefEntry creates a FilesInputSourceArrayRefEntry from a sha256sum and metadata.
+// The sha256sum is the SHA256 checksum of the file as a hexadecimal string without any prefix (e.g. "sha256:").
+func NewFilesInputSourceArrayRefEntry(sha256Sum string, metadata FilesInputRefMetadata) FilesInputSourceArrayRefEntry {
```
7flying, line 236: (ref given in the above comment)

achilleas-k (dismissed):
Great stuff.  Love how the stages cleaned up now.  All is how it should be, for files inputs at least.

achilleas-k (approved):
Still great :)

PR #3254, opened 2023-01-31
File resolver job

kingsleyzissou:
This pull request includes:
- adding a file resolver to generate a checksum for the contents of a remote file.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-file-resolve.go
+++ b/cmd/osbuild-worker/jobimpl-file-resolve.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/osbuild/osbuild-composer/internal/remotefile"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+type FileResolveJobImpl struct{}
+
+func (impl *FileResolveJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id())
+
+	var err error
+	result := worker.FileResolveJobResult{
+		Success: false,
+		Results: []worker.FileResolveJobResultItem{},
+	}
+
+	defer func() {
+		logWithId := logrus.WithField("jobId", job.Id().String())
+		if result.JobError != nil {
+			logWithId.Errorf("file content resolve job failed: %s", result.JobError.Reason)
+			if result.JobError.Details != nil {
+				logWithId.Errorf("failure details: %v", result.JobError.Details)
+			}
+		}
+
+		if result.Results == nil || len(result.Results) == 0 {
+			logWithId.Infof("Resolving file contents failed: %v", err)
+			result.JobError = clienterrors.WorkerClientError(
+				clienterrors.ErrorRemoteFileResolution,
+				"Error resolving file contents",
+				"All remote file contents returned empty",
+			)
+		}
+
+		err := job.Update(result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	var args worker.FileResolveJob
+	err = job.Args(&args)
+	if err != nil {
+		return err
+	}
+
+	logWithId.Infof("Resolving file contents (%d)", len(args.URLs))
+
+	resolver := remotefile.NewResolver()
+	for _, url := range args.URLs {
+		resolver.Add(url)
+	}
+
+	resultItems := resolver.Finish()
+
+	for idx := range resultItems {
+		result.Results = append(result.Results, worker.FileResolveJobResultItem{
+			URL:             resultItems[idx].URL,
+			Content:         resultItems[idx].Content,
+			ResolutionError: resultItems[idx].ResolutionError,
+		})
+	}
+
+	resolutionErrors := result.ResolutionErrors()
+	if len(resolutionErrors) == 0 {
+		result.Success = true
+	} else {
+		result.JobError = clienterrors.WorkerClientError(
+			clienterrors.ErrorRemoteFileResolution,
+			"at least one file resolution failed",
+			resolutionErrors,
+		)
+	}
+
+	return nil
+}
--- a/internal/remotefile/resolver.go
+++ b/internal/remotefile/resolver.go
@@ -1,0 +1,66 @@
+package remotefile
+
+import (
+	"context"
+
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+type resolveResult struct {
+	url     string
+	content []byte
+	err     error
+}
+
+// TODO: could make this more generic
+// since this is shared with the container
+// resolver
+type Resolver struct {
+	jobs  int
+	queue chan resolveResult
+
+	ctx context.Context
+}
+
+func NewResolver() *Resolver {
+	return &Resolver{
+		ctx:   context.Background(),
+		queue: make(chan resolveResult, 2),
+	}
+}
+
+func (r *Resolver) Add(url string) {
+	client := NewClient()
+	r.jobs += 1
+
+	go func() {
+		content, err := client.Resolve(url)
+		r.queue <- resolveResult{url: url, content: content, err: err}
+	}()
+}
+
+func (r *Resolver) Finish() []Spec {
+
+	resultItems := make([]Spec, 0, r.jobs)
+	for r.jobs > 0 {
+		result := <-r.queue
+		r.jobs -= 1
+
+		var resultError *clienterrors.Error
+		if result.err != nil {
+			resultError = clienterrors.WorkerClientError(
+				clienterrors.ErrorRemoteFileResolution,
+				result.err.Error(),
+				result.url,
+			)
+		}
+
+		resultItems = append(resultItems, Spec{
+			URL:             result.url,
+			Content:         result.content,
+			ResolutionError: resultError,
+		})
+	}
+
+	return resultItems
+}
--- a/internal/remotefile/resolver_test.go
+++ b/internal/remotefile/resolver_test.go
@@ -1,0 +1,102 @@
+package remotefile
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSingleInputResolver(t *testing.T) {
+	server := makeTestServer()
+	url := server.URL + "/key1"
+
+	resolver := NewResolver()
+
+	expectedOutput := Spec{
+		URL:             url,
+		Content:         []byte("key1\n"),
+		ResolutionError: nil,
+	}
+
+	resolver.Add(url)
+
+	resultItems := resolver.Finish()
+	assert.Contains(t, resultItems, expectedOutput)
+
+	for _, item := range resultItems {
+		assert.Nil(t, item.ResolutionError)
+	}
+}
+
+func TestMultiInputResolver(t *testing.T) {
+	server := makeTestServer()
+
+	urlOne := server.URL + "/key1"
+	urlTwo := server.URL + "/key2"
+
+	expectedOutputOne := Spec{
+		URL:             urlOne,
+		Content:         []byte("key1\n"),
+		ResolutionError: nil,
+	}
+
+	expectedOutputTwo := Spec{
+		URL:             urlTwo,
+		Content:         []byte("key2\n"),
+		ResolutionError: nil,
+	}
+
+	resolver := NewResolver()
+
+	resolver.Add(urlOne)
+	resolver.Add(urlTwo)
+
+	resultItems := resolver.Finish()
+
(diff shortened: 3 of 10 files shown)


kingsleyzissou:
I created a separate PR for this just to make things a little more manageable (since the other one would be huge if I added this too)

internal/file/spec.go
```diff
@@ -0,0 +1,29 @@
...
+	Filename string
+	Type     FileSpecType
+}
+
+type Spec struct {
+	URL      string
+	Filename string
+	Checksum string
+	Type     FileSpecType
+}
```
achilleas-k, line 29: Can we get away with having just one spec?  Though I wonder if we even need a spec at all?  Looking at the resolve function, it might be simpler to just have the URL as an argument and the checksum as output.  The caller can group the checksum with the filename and URL after the fact (from outside) if necessary.  Though maybe it's convenient to have these all bundled in the job.

internal/file/client.go
```diff
@@ -0,0 +1,83 @@
...
+	if !input.Type.IsValid() {
+		return fmt.Errorf("File resolver: spectype is invalid")
+	}
+	return nil
+}
+
+// resolve the contents of a remote file and
+// return the checksum of the contents to
+// be used in the pipeline
+func (c *Client) Resolve(input InputSpec) (Spec, error) {
```
achilleas-k, line 60: See my comment on the specs about this.

achilleas-k (commented):
Good stuff.  Small comment/question about the input and result specs.

internal/file/spec.go
kingsleyzissou: Yeah absolutely. I updated it to a single spec... I misread the second part of your comment. We could also group the URL and checksum on the caller's side - we'd still need to add that to some kind of struct and then pass that down the chain to the manifest call though, right?

thozza: I think that getting the content may be useful and we can always calculate checksum once we have it. returning just the checksum seems a bit of a oversimplification to me, but I would be OK with it if everyone finds it as a better approach. +1 to the rest...

internal/file/client.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"crypto/sha256"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+type Client struct {
+	client *http.Client
+}
```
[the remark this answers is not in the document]
thozza: Thinking about this, it would make sense to keep it as is if the package is called something like `remote_file` or `remote_content`. 🤔

cmd/osbuild-worker/jobimpl-file-resolve.go
```diff
@@ -0,0 +1,65 @@
...
+
+		for i, s := range specs {
+			result.Specs[i] = worker.FileResolveResultSpec{
+				URL:      s.URL,
+				Checksum: s.Checksum,
+			}
+		}
+	}
+
+	err = job.Update(&result)
```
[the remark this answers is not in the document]
kingsleyzissou: Ah right. Thanks for that. I will update this and make the other suggested changes. Thanks for the feedback

internal/file/spec.go
kingsleyzissou: Ah yeah good point, we may still need the contents of the file. I will update that and we can calculate the checksum at a later stage, like you say

internal/file/spec.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+var fileSpecTypes = map[FileSpecType]struct{}{
+	SpecTypeGPGKey: {},
+}
+
+func (t FileSpecType) IsValid() bool {
+	_, ok := fileSpecTypes[t]
+	return ok
+}
```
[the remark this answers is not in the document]
kingsleyzissou: This was me getting a few steps ahead of myself. I figured down the line that we would want to differentiate the file types... but this is probably not the best place to be doing that. Will update.

internal/file/spec.go
kingsleyzissou: Oh yeah brilliant - I hadn't thought of that, but it would make it easier to import the gpgkey in general. I'm going to push the changes now, just fixing up some of the tests.

kingsleyzissou: I've left the contents in `[]byte`, not sure if we want to return a string or keep the contents in bytes. Let me know

thozza: `[]byte` is IMO preferred, because it is the most generic representation. We may end up fetching also binary files (not that I know about any requirement). It is simple to convert `[]byte` to `string`...

internal/worker/json.go
```diff
@@ -256,6 +256,16 @@ type ContainerResolveJobResult struct {
 	JobResult
 }
 
+type FileResolveJob struct {
+	URLs []string `json:"urls"`
+}
+
+type FileResolveJobResult struct {
+	Contents [][]byte `json:"contents"`
```
thozza, line 264: Wouldn't it make sense to somehow tie the returned contents to the URLs that were passed in the job? So either making this something like:

```go
type FileResolveJobResultItem struct {
	URL      string `json:"url"`
	Content  []byte `json:"content"`
}

type FileResolveJobResult struct {
	Results []FileResolveJobResultItem `json:"results"`
 	JobResult
}
```

or

```go
type FileResolveJobResult struct {
	Contents map[string][]byte `json:"contents"`
 	JobResult
}
```

I know that the index of the content in the returned result will be probably the same as that the one passed in the job itself. I just find it nicer if the result content is useful on its own, without having the access to another data structure that was passed when requesting the job.

Just something to consider...

cmd/osbuild-worker/jobimpl-file-resolve.go
```diff
@@ -0,0 +1,61 @@
...
+	result := worker.FileResolveJobResult{
+		Contents: make([][]byte, len(args.URLs)),
+	}
+
+	defer func() {
+		logWithId := logrus.WithField("jobId", job.Id().String())
+		if result.JobError != nil {
+			logWithId.Errorf("file content resolve job failed: %s", result.JobError.Reason)
+			if result.JobError.Details != nil {
+				logWithId.Errorf("failure details : %v", result.JobError.Details)
```
thozza, line 30: nitpick: an extra space before `:`

cmd/osbuild-worker/jobimpl-file-resolve.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	result := worker.FileResolveJobResult{
+		Contents: make([][]byte, len(args.URLs)),
+	}
+
+	defer func() {
+		logWithId := logrus.WithField("jobId", job.Id().String())
+		if result.JobError != nil {
```
thozza, line 27: In other job types, we have the `success` field, which we set by default to `false` at the beginning. This job type does not have that. I don't see any case when the function would return without setting an error, but I wonder whether we should set it in this deferred function if the `result.Contents` is empty? It probably does not make make sense to submit a job that resolves zero files, thus returning empty results seems like an error...

thozza (commented):
Nice work 👍 

I added a few nitpicks and things that I would like to get you PoV on. But other than that I don't see any blockers, this is a very nice PR. Happy to approve once we discuss my comments.

internal/worker/json.go
kingsleyzissou: Yeah actually I've been wondering about this. I think I had something similar in the first iteration but I removed it. I think this would be quite helpful to keep track of the url for use later down the road. I also agree that keeping track of this is nicer (and more explicit) than just relying on the indexes being the same (since we're using a goroutine, there's a chance that they might not actually be the same).

internal/remotefile/resolver.go
```diff
@@ -0,0 +1,62 @@
...
+		if result.err == nil {
+			contents = append(contents, result.contents)
+		} else {
+			errs = append(errs, result.err.Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		detail := strings.Join(errs, "; ")
+		return contents, fmt.Errorf(detail)
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah good point :) I'll update

cmd/osbuild-worker/jobimpl-file-resolve.go
kingsleyzissou: Haha whoops

internal/worker/json.go
kingsleyzissou: I went with the first one since it seems more readable to me :)

internal/remotefile/resolver.go
```diff
@@ -0,0 +1,68 @@
...
+				URL:     result.url,
+				Content: result.content,
+			})
+		}
+	}
+
+	if len(errs) > 0 {
+		detail := strings.Join(errs, "; ")
+		return nil, fmt.Errorf(detail)
+	}
```
thozza, line 65: This block got me thinking if it would make sense to connect the `FileResolveJobResultItem` with the error that could have happened. Especially the URL. My thinking is that the error may not contain the URL that failed to resolve and as a result it will be hard to determine what actually failed.

I was thinking about potentially embedding the error itself in the `FileResolveJobResultItem` structure. You could then add a variant of 
https://github.com/osbuild/osbuild-composer/blob/f8b39cd34efd175fbb59a9c0b08072e4e2a951d9/internal/worker/json.go#L59-L76
to collect errors from the slice of `FileResolveJobResultItem` and use it in the job implementation to decide if the resolution was successful.

Alternatively, you could add a new type implementing the `Error` interface, which would hold the error and then you could just return `[]error`.

Something like:

```go
type ResolutionError struct {
	URL string
	Err error
}

func (e *ResolutionError) Error() string {
	return fmt.Sprintf("%s: %v", e.URL, e.Err)
}
```

I think that we are potentially loosing some information by just concatenating all errors into single one and returning it.

Nevertheless I checked the container resolver and it has the same behavior (which I don't like)...

thozza (commented):
Great work. I like your changes and the current state. I added a few comments / questions.

In addition, this PR got me thinking if we should have some functional test for worker jobs using some dummy worker server... Because the newly added worker job is currently not being tested in any way. And it won't be until it is used by some other test case. This would be useful and relevant even for existing job implementations. Nevertheless I admit that this would be probably a bigger thing (I'm not sure).

internal/remotefile/resolver.go
kingsleyzissou: Cool, I've added a variant of this - that is if I understood this comment correctly. It may need to be fleshed out a little bit more (or possibly some additional tests might be needed)

achilleas-k (commented):
Looks good.  Some comments but nothing major.
Are we adding any kind of integration test for this to our CI?  Maybe in a followup?

internal/remotefile/resolver.go
```diff
@@ -0,0 +1,67 @@
...
+	client := NewClient()
+	r.jobs += 1
+
+	go func() {
+		content, err := client.Resolve(url)
+		r.queue <- resolveResult{url: url, content: content, err: err}
+	}()
+}
+
+func (r *Resolver) Finish() []worker.FileResolveJobResultItem {
```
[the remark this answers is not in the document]
kingsleyzissou: I can't remember now why I removed the spec in the first place. Makes sense to have that and not depend on the `worker` package. I'll fix that up

cmd/osbuild-worker/jobimpl-file-resolve.go
```diff
@@ -0,0 +1,75 @@
...
+
+	resultSuccess := true
+	for idx := range resultItems {
+		if resultItems[idx].ResolutionError != nil {
+			resultSuccess = false
+		}
+		result.Results = append(result.Results, worker.FileResolveJobResultItem{
+			URL:     resultItems[idx].URL,
+			Content: resultItems[idx].Content,
+		})
```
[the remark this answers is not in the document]
kingsleyzissou: Okay, I see. Yeah that makes a lot of sense. I think I missed adding the resolution here... so good catch. Will fix that up and take my time to make sure I get everything. And then I'll implement the resolution error check similar to the target errors.

internal/remotefile/resolver_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	resultItems := resolver.Finish()
+
+	errs := []string{}
+	for _, item := range resultItems {
+		errs = append(errs, item.ResolutionError.Reason)
+	}
+
+	assert.Contains(t, errs, expectedErrMessageOne)
+	assert.Contains(t, errs, expectedErrMessageTwo)
```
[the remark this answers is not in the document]
kingsleyzissou: Ah good point. I think adding the URL parse as extra validation would be a quick way to make sure this is working. So I'll add that here since there are a few changes I need to make.

kingsleyzissou:
> Are we adding any kind of integration test for this to our CI? Maybe in a followup?
I will add integration tests in a follow up - when I start connecting it to the other things then I can add the integration tests :)

thozza (approved):
Everything looks great 🥳  Thanks for all your work and patience!

PR #3181, opened 2022-12-12
image: add new image type "minimal-raw" for fedora

sarmahaj:
Signed-off-by: Sarita Mahajan <sarmahaj@redhat.com>

This pull request includes:
New image type named 'minimal-raw' has been added to fedora.
Reference taken from https://github.com/osbuild/osbuild-composer/pull/3077 
Image is basically a minimal rpm bootable image supports aarch64 & x86
Image is tested manually.
From a RHEL PoV this will enable us to support similar devices as part of the SystemReady "Edge on Arm" initiative where the customer may not wish to run R4E or to use for development/testing/prototyping where a ostree image may not be the most straight forward way to do things.

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3256, opened 2023-02-01
test: Add minimal-raw image test

henrywang:
The minimal-raw image is for Fedora only and feature merged by PR #3181.
Test only covers UEFI because the comment https://github.com/osbuild/osbuild-composer/pull/3181#issuecomment-1380029923
The minimal-raw is not ostree like file system, but "normal" RPM one. So no upgrade test, but have package installation and reboot test included.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -434,6 +434,17 @@ Rebase OSTree UEFI:
           - rhos-01/centos-stream-8-x86_64
           - rhos-01/centos-stream-9-x86_64
 
+Minimal raw:
+  stage: test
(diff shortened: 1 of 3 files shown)


jrusz (approved):
LGTM

achilleas-k (approved):
LGTM

Something for the future: Given how much of these tests for bare-metal type images (edge, minimal raw, iso) have in common, I think it would be useful to modularise the test scripts to have reusable parts.  There are times when changes need to be made to common code in multiple files and the number of them keeps growing.
I know it's not that easy with bash, but hopefully doable to some extent.

PR #3262, opened 2023-02-02
Introduce jobtype variable in worker dashboard

croissanne:
image

---

The tradeoff is the 2nd target_duration variable. As it's tricky to have a variable with multiple values (so we could define the target duration per jobtype), you sort of need to know which target_duration to set for which jobtype (1792 for osbuild jobs, 32 for depsolve/resolve jobs).

https://grafana.stage.devshift.net/d/image-builder-worker-sanne/image-builder-worker-sanne?orgId=1


lavocatt:
Can you explain a bit more why is there a need to specify the target time statically please ? I think I kinda understand but a bit unsure.

croissanne:
> Can you explain a bit more why is there a need to specify the target time statically please ? I think I kinda understand but a bit unsure.

So per jobtype we have different targets of what consititutes "fast". We commit to 95% of depsolve jobs taking 32s or lower, and we commit to osbuild jobs taking less than 30 minutes. I haven't really found a way to change the jobtype and the target time at the same time, so I made a 2nd variable.

The target time is needed to correctly calculate the error budget and slow request rates in the `job duration` and `job wait duration` sections.

kingsleyzissou:
> The tradeoff is the 2nd target_duration variable. As it's tricky to have a variable with multiple values (so we could define the target duration per jobtype), you sort of need to know which target_duration to set for which jobtype (1792 for osbuild jobs, 32 for depsolve/resolve jobs).

Awesome stuff. It looks a lot better and less cluttered like this. Pity about the jobtype... but I don't think it's a big deal. I mean this is mostly for us right? And I guess we'll know to switch the target duration for each job. would it make sense then to have. Having said that, would it make more sense to have the input selector be a single selector so you can only view one job at a time rather than a multi select?

croissanne:
Discussed with Tom:

- job wait duration per architecture
- future: stats per image type

croissanne:
> Discussed with Tom:
> 
>     * job wait duration per architecture

Ok so we don't actually add the arch label to the job_wait_duration metric.

> 
>     * future: stats per image type

croissanne:
Job wait duration per architecture is now available after #3289

kingsleyzissou (approved):

PR #3277, opened 2023-02-08
internal/worker: add String to clienterrors.Error struct

### LINKED ISSUE #3272: Log details when osbuild job fails with clienterrors.ErrorTargetError
If the worker fails with (target) errors,
https://github.com/osbuild/osbuild-composer/blob/b8ddbd3031f68a611005018b9e7983608e85c8d7/cmd/osbuild-worker/jobimpl-osbuild.go#L901-L903

they get logged.
https://github.com/osbuild/osbuild-composer/blob/b8ddbd3031f68a611005018b9e7983608e85c8d7/cmd/osbuild-worker/jobimpl-osbuild.go#L179-L184

The details contain relevant information in this case, so they should be logged as well. The tricky part is that `logWithId.Errorf("failure details : %v", result.JobError.Details)` will be able to log details that are a string, but not details that are a slice of other JobErrors, which is what the details contain in the case of target errors.

This results in unhelpful logs:
```
time="2023-01-13T07:12:32Z" level=error msg="failure details : [0xc00299c330]" jobId=
time="2023-01-13T07:12:32Z" level=error msg="osbuild job failed: at least one target failed" jobId=
time="2023-01-13T07:12:32Z" level=info msg="[Azure] 📦 Creating a new storage account" jobId=
```

Unsure how to solve this without reflection in the error handler. Or we should just log these details when creating the capsulating target error.

croissanne:
This should make the logging better when errors have another error struct.

Fixes #3272

### DIFF
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -1,5 +1,9 @@
 package clienterrors
 
+import (
+	"fmt"
+)
+
 const (
 	ErrorNoDynamicArgs        ClientErrorCode = 1
 	ErrorInvalidTargetConfig  ClientErrorCode = 2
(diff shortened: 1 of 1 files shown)


kingsleyzissou (dismissed):
Thanks!

kingsleyzissou (approved):

PR #3279, opened 2023-02-10
test/cases: move CIV options into a variable

jabia99:
In order to provide different options when running CIV from it's own CI, move them to the variable CIV_OPTIONS

This pull request includes:

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -199,9 +199,14 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+  if [ -z "${CIV_OPTIONS:-}" ]; then
+    echo "ERROR: Please provide the variable CIV_OPTIONS"
+    exit 1
+  fi
 else
   # If not, get prod container
   TAG="prod"
+  CIV_OPTIONS=( -r=/tmp/resource-file.json -d -o=/tmp/report.xml -m="not pub" )
 fi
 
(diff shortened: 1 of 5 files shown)


test/cases/aws.sh
```diff
@@ -199,9 +199,14 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+  if [ -z "${CIV_OPTIONS}" ]; then
```
jrusz, line 202: This will fail with `CIV_OPTIONS: unbound variable` since this script has `set -e`. I suggest to update it to this:
```suggestion
  if [ -z "${CIV_OPTIONS:-}" ]; then
```

test/cases/azure.sh
```diff
@@ -199,9 +199,14 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+  if [ -z "${CIV_OPTIONS}" ]; then
```
jrusz, line 202: Same comment as above, this will fail.

jrusz (changes requested):
Suggested a change.

test/cases/aws.sh
jabia99: But it should fail if `CIV_OPTIONS` is not defined. Some of them are not optional

jabia99:
Extra commit to remove cloud cleaner

jrusz (approved):
LGTM, can you then also remove the pipeline schedules in Gitlab if that's not done already? :)

PR #3281, opened 2023-02-10
Blueprint: add support for custom files and directories in `/etc`

thozza:
Tl;DR; this PR adds the ability to create custom files and directories under `/etc` in the image by specifying them as a BP customization. The functionality is supported for all non-ostree images and for ostree commit and container images.

| | Feature description | Support in this PR | Comment |
| --- | --- | :---: | --- |
| Directories customization |Create new | :white_check_mark: | <ul><li>possible to specify that parent directories should be created as needed</li><li>path policy applies</li></ul> |
| | Modify existing | :x: | <ul><li>Modifying ownership and permissions of existing directories is not supported and will result in failure during image build.</li><li>One can specify only plain custom directory (without explicit ownership and permissions) to ensure that the directory exist (e.g. due to bing a parent of a custom file). In such case, the directory will be created with default permissions and ownership if it does not exist in the image or it will be a noop if it exists.</li></ul>|
| | Delete existing | :x:| |
| |Set user / group ownership| :white_check_mark: | <ul><li>possible to use using user / group name or UID / GID</li><li>possible to set to user / group provided as BP customization</li></ul> |
| | Set mode | :white_check_mark: | <ul><li>possible by providing mode in octal format (such as "0755")</li></ul> |
| | Set additional attributes | :white_check_mark: | <ul><li>such as xattr, custom SELinux labels, etc.</li></ul> |
| Files customization | Create new | :white_check_mark: | <ul><li>path policy applies</li><li>parent directories must exist</li></ul> |
| | Modify existing | :white_check_mark: | <ul><li>all supported file attributes can be set on existing files and their content overridden</li><li>path policy applies</li></ul> |
| | Delete existing | :x: | |
| | Set user / group ownership | :white_check_mark: | <ul><li>possible to use using user / group name or UID / GID</li><li>possible to set to user / group provided as BP customization</li></ul> |
| | Set mode | :white_check_mark: | <ul><li>possible by providing mode in octal format (such as "0755")</li></ul> |
| | Set additional attributes | :x: | <ul><li>such as xattr, custom SELinux labels, etc. </li></ul> |
| | Plain text content | :white_check_mark: | |
| |Binary (base64) content | :x: | <ul><li>would be very easy to add</li></ul>|
| Support in distros | all | :white_check_mark: | |
| Support in image types | non-ostree image types | :white_check_mark: | <ul><li>such as qcow2, cloud images, etc.</li></ul> |
| |<ul><li>edge-commit</li><li>edge-container</li></ul>| :white_check_mark: | <ul><li>files and dir are baked into the commit</li></ul> |
| | <ul><li>edge-installer</li><li>edge-raw-image</li><li>edge-simplified-installer</li></ul> | :x: | <ul><li>not as a deploy-time customization</li></ul>|
| Common | SELinux labels | :grey_question: | <ul><li>Custom directories and files will have the default label applied as specified by the used SELinux policy</li></ul>|
| | Weldr API support | :white_check_mark: | |
| | Cloud API support | :white_check_mark: | |
| | Limits | ❔ | <ul><li>No limit is applied on the number of custom directories and files that can be created</li><li>No limit is applied on the size of custom files</li><li>No assurance that the custom files can fit in the image or partition if it is too small</li></ul>|

Most notable changes:
- Add internal representation for filesystem nodes as `fsnode` package
- Add support for `org.osbuild.chown` stage
- Add helper functions for generating osbuild stages from `fsnode` types with the aim to create them in the image.
- Extend Blueprint customizations with Directories and Files customizations and implement their validation.
- Extend Cloud API to accept Directories and Files customizations and validate them.
- Extend distro definitions and pipeline implementations to handle custom Directories and Files.
- Extend image test manifests with the new customizations and regenerate them.
- Extend the OSTree tests (`OSTree` and `New OSTree`) to use the custom Directories and FIles feature and test its effect on the resulting deployed image.

Details and some implementation choices are described in commit messages.

TODO:

- [x] Determine desired custom directories policy (which directories to protect?)
  - Dirs can be created only under `/etc` and existing dirs can't be modified. 
- [x] Determine desired custom files policy (which files to protect?)
  - Keep it minimal (fstab, groups, passwd, shadow)

This pull request includes:

  - [x] unit tests
  - [x] Weldr API tests (tested via ostree tests, which use it)
  - [x] Cloud API tests
  - [x] ostree functional tests (ostree / ostree-ng)
  - [x] https://github.com/osbuild/guides/pull/103


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 157fc42d0bf2222322a45e7737dfb60f1500120f with the main merge-base 6846f59ff4215ae1a0d923e93782d6e4c190ba1b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3753586209/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3753586209/artifacts/browse

henrywang:
@yih-redhat Could you please take a look at this PR and add test here? Thanks.

yih-redhat:
> @yih-redhat Could you please take a look at this PR and add test here? Thanks.

Tomas will add the test code, we can extend the test if necessary once it's ready.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c273163bd224be222d52562dffe736c4a79344b8 with the main merge-base 6846f59ff4215ae1a0d923e93782d6e4c190ba1b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3761769850/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3761769850/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 12515d79e81ec539829f51869b83eca26b74c832 with the main merge-base 6846f59ff4215ae1a0d923e93782d6e4c190ba1b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3762432617/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3762432617/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD bd9629258e634ea510e2d0efa62bd4f0d22f5666 with the main merge-base 6846f59ff4215ae1a0d923e93782d6e4c190ba1b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3766816983/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3766816983/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4e3a1a52c472f019a9292b4b2304e510e69c9e21 with the main merge-base be48279af8885d7c2c85c075313c491fc132125e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3768123268/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3768123268/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 42597c00f3f7e196729c0c5d2311676ca703365b with the main merge-base be48279af8885d7c2c85c075313c491fc132125e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3770233350/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3770233350/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 7ea059fc8d4637b4f08361912d229f9de7b17239 with the main merge-base be48279af8885d7c2c85c075313c491fc132125e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3771311128/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3771311128/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 04a0370f71b03887650bb91d115ad8ae3330994a with the main merge-base 5a624531906f01b604473814345345782c84d41f).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3777183690/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3777183690/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 29cd82a13b71f04e2b598cbef8a4be46bbfe987c with the main merge-base 5a624531906f01b604473814345345782c84d41f).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3778250607/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3778250607/artifacts/browse

internal/osbuild/copy_stage.go
```diff
@@ -39,6 +39,10 @@ func NewCopyStageSimple(options *CopyStageOptions, inputs Inputs) *Stage {
 	}
 }
 
+type CopyStageFilesInputs map[string]*FilesInput
```
lavocatt, line 42: There's a `TODO: commit message` for this commit :smile:

thozza: Can you post the commit SHA? Github is not good at showing that with a review...

thozza: Got it... https://github.com/osbuild/osbuild-composer/pull/3281/commits/b354e86bbabcbb865a08612401affe12418d941d I'll add a message then :D

internal/distro/rhel7/images.go
```diff
@@ -141,6 +142,21 @@ func osCustomizations(
 		osc.FactAPIType = options.Facts.ApiType
 	}
 
+	var err error
+	osc.Directories, err = blueprint.DirectoryCustomizationsToFsNodeDirectories(c.GetDirectories())
```
lavocatt, line 146: Can you generalize this chunk of code to someting like ?
```
osc.Directoriies, osc.Files, err = bluepring.somename(...)
```
To avoid copy pasting this 3 times.

internal/distro/fedora/images.go
```diff
@@ -385,6 +385,16 @@ func iotRawImage(workload workload.Workload,
 	img.Users = users.UsersFromBP(customizations.GetUsers())
 	img.Groups = users.GroupsFromBP(customizations.GetGroups())
 
+	var err error
+	img.Directories, err = blueprint.DirectoryCustomizationsToFsNodeDirectories(customizations.GetDirectories())
```
lavocatt, line 389: Same here for the generalization

internal/distro/rhel8/images.go
```diff
@@ -466,6 +466,17 @@ func edgeSimplifiedInstallerImage(workload workload.Workload,
 	rawImg.Groups = users.GroupsFromBP(customizations.GetGroups())
 
 	rawImg.KernelOptionsAppend = []string{"modprobe.blacklist=vc4"}
+
+	var err error
```
lavocatt, line 470: Same here for the generalization

lavocatt (commented):
Very nice one! I have a few comments/questions below.

I'm wondering also, has the `etc` customization been ported to every test cases that supports it ? If so, should we also add some tests without `etc` customization for the completeness of the testing scenarios.

internal/distro/rhel7/images.go
thozza: I admit that his is a bit verbose, but this is unfortunately the way how error handling works in Go. I do not consider this to be an issue to be honest.

I do not want to unnecessarily tie Files and Directories customization together when not needed. These are two different customizations. We do not do what you are suggesting for any other customizations (e.g. users / groups).

Lastly, the fact that this is copied three times in each distro is a result of how we define distros, not of the fact that there is a separate function for each customization, as there already is for those existing ones. Instead, we should aim for deduplicating `distro/<distro>/images.go` where it makes sense, because these are very similar. However this was not the goal of the rework that @achilleas-k did in all distros, but it is AFAIK something that we would like to eventually look at.

internal/distro/fedora/images.go
thozza: Commented in the Fedora distro. TL;DR; I'm not in favor of generalizing this.

internal/distro/rhel8/images.go
thozza: Commented in the Fedora distro. TL;DR; I'm not in favor of generalizing this.

thozza:
> Very nice one! I have a few comments/questions below.

Thanks for your review!

> I'm wondering also, has the `etc` customization been ported to every test cases that supports it ? If so, should we also add some tests without `etc` customization for the completeness of the testing scenarios.

So just to clear any confusion, the PR does not add `etc` customization, but **two** new customizations for Files and Directories. That may be helpful when thinking about why these are not handled together unless completely necessary. There is only one case where this is needed, to check that someone didn't specify a custom file, which is a parent of a custom directory. We could just fail during build, because directories are created first, but I decided to do the validation in composer to fail fast.

As for the test cases:

- I didn't want to substantially grow our functional test jobs, because I don't see any value in that (than paying more to AWS). I considered adding a separate test case, like the filesystem customization has. After a discussion with @ondrejbudai and @croissanne, I decided to just extend the API test case.
- I extended test cases which already were testing some customizations (mostly API and ostree tests) to use it. Note that ostree tests do not use it on all distros, only on Fedora all, CentOS Stream 8/9 and rhel-9.2 and rhel-8.8.
- One can consider the vanilla image test manifests without any customizations to cover that case when the customization is not used (IOW that nothing gets added to the manifest). I don't think that we can do any additional testing of "not using the new customizations" that would extend the coverage.

internal/osbuild/copy_stage.go
thozza: I'll wait for other reviews before pushing the fixed commit

internal/distro/fedora/images.go
```diff
@@ -129,6 +129,21 @@ func osCustomizations(
 		)
 	}
 
+	var err error
+	osc.Directories, err = blueprint.DirectoryCustomizationsToFsNodeDirectories(c.GetDirectories())
+	if err != nil {
+		// In theory this should never happen, because the blueprint directory customizations
+		// should have been validated before this point.
+		panic(fmt.Sprintf("failed to convert directory customizations to fs node directories: %v", err))
```
croissanne, line 137: I guess it's fine to panic because this would be a programming error. But why not propagate the error? All functions which invoke osCustomizations can return an error (but none do atm). Maybe the panic would need to happen higher up? idk, not a blocker any way, just wondering.

croissanne (commented):
lgtm other than the TODO commit msg :)

internal/distro/fedora/images.go
thozza: I didn't return the error, because as you mentioned, the `osCustomizations()` does not return any. Personally I prefer to return errors instead of panicking. These customizations were the only ones that could return error while converting them. I think that we may end up returning errors also for other customizations if we start validating them.

I would be happy to rework this to return an error... 😉

croissanne: Maybe in a followup though :D

thozza:
Note that the files customization would currently follow symlinks if the file path already exists in the image and it is a symlink to a file. I'll address this in a follow-up PR once https://github.com/osbuild/osbuild/pull/1241 is merged.

mrguitar:
@thozza looks great! Thanks for tweaking the commit vs installer deployment. Two small questions:

1) if a user creates a unit file, will the ordering in osbuild allow the [customization.services] to enable/disable the new units?
2) We're pasting in plain text in between quotes. I haven't gotten my hands on this, nor am I a TOML person, so apologies if this is a dumb question. Will users have to escape quotes that are contained inside the pasted text or anything like that?

internal/disk/disk.go
```diff
@@ -184,19 +168,3 @@ func NewVolIDFromRand(r *rand.Rand) string {
 	}
 	return hex.EncodeToString(volid)
 }
-
-func CheckMountpoints(mountpoints []blueprint.FilesystemCustomization, mountpointAllowList *pathpolicy.PathPolicies) error {
```
achilleas-k: I think this fits better in the new `pathpolicy` package instead of `blueprint` and should have probably moved with the rest of the policy parts.  This shouldn't create any import issues since the `pathpolicy` is imported in `blueprint` only for these checks.

Since the policies are created and managed in the `pathpolicy` package, it seems more natural that the code to use them for the main purpose (policy checking) would be in that same package instead of where the values being checked are defined (`blueprint`).

internal/disk/disk.go
```diff
@@ -53,19 +50,6 @@ const (
...
-	"/boot": {Exact: true},
-	"/var":  {},
-	"/opt":  {},
-	"/srv":  {},
-	"/usr":  {},
-	"/app":  {},
-	"/data": {},
-	"/home": {},
-	"/tmp":  {},
-})
```
achilleas-k, line 66: Can this change (moving the `MountpointPolicies`) be merged with the commit when the package split happened?  Not important though.

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/osbuild/osbuild-composer/internal/common"
+	"github.com/osbuild/osbuild-composer/internal/fsnode"
+)
+
+// validateModeString checks that the given string is a valid mode octal number
+func validateModeString(mode string) error {
+	// check that the mode string matches the octal format regular expression
+	if regexp.MustCompile(`^0[0-7]{3}$`).MatchString(mode) {
+		return nil
+	}
```
achilleas-k, line 22: Are we sure we want to restrict it to having a leading `0`?  I suspect a lot of users will get tripped up by this from the general habit of `chmod`ing with just the three digits for convenience.

achilleas-k (commented):
Partial review and have to stop for the day.

Reached a69ea5f89cb6cb54737101552fe89b552dd41175

teg:
Only high level review, but looks great!

internal/disk/disk.go
thozza: Probably yes, I just did the split really in steps to not mess things up too much. Thus moving things around in multiple commits.

internal/blueprint/fsnode_customizations.go
thozza: I think I could change the regex to make the leading `0` optional. I agree that people are probably more used to write permissions without the leading zero.

internal/fsnode/file.go
```diff
@@ -0,0 +1,36 @@
...
+func (f *File) Data() []byte {
+	if f == nil {
+		return nil
+	}
+	return f.data
+}
+
+// NewFile creates a new file with the given path, data, mode, user and group.
+// user and group can be either a string (user name/group name), an int64 (UID/GID) or nil.
+func NewFile(path string, data []byte, mode *os.FileMode, user interface{}, group interface{}) (*File, error) {
```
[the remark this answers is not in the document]
thozza: I'm not opposed to it. I considered this during the implementation and went with the current version due to the fact, that one would probably always want to specify `data` for a file, but not necessarily the `mode`, `user`, etc. However, now it feels as if I was thinking about this more in the `Python` terms in the sense that in `Go` you can't specify default values for arguments and thus you must always provide all of them. So having the `data` as the second argument does not help much.

internal/distro/rhel9/imagetype.go
```diff
@@ -360,7 +360,7 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 			}
 
 			// ignition is optional, we might be using FDO
-			if customizations.Ignition.HasIgnition() {
```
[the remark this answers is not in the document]
thozza: Thanks, will check it out... 🤔

internal/disk/disk.go
thozza: While I agree that policies are currently created and managed in the `pathpolicy` package, It could be applied on any object with paths. So in theory it could be applied also on a slice of `disk.Filesystem` or any other.

In the same sense as we do not refer to `osbuild` package structures in the internal API-agnostic code, I do believe that it is also good to not refer to the `blueprint` package structures outside of it unless necessary to convert them. I get that `blueprint` is kind of API-agnostic concept in composer so comparing it to `osbuild` package is maybe a stretch.

Lastly, my thinking was also to keep all helper functions for `blueprint` structures in that package, so those that transform them to the "internal API" structures, but also those that are doing come validation  / (policy) checks on them.

But I do not feel very strongly about keeping this in the `blueprint` package.

internal/fsnode/fsnode.go
```diff
@@ -0,0 +1,132 @@
...
+			return fmt.Errorf("user name %q doesn't conform to validating regex (%s)", user, nameRegex.String())
+		}
+	case int64:
+		if user < 0 {
+			return fmt.Errorf("user ID must be non-negative")
+		}
+	case nil:
+		// user is not set
+	default:
+		return fmt.Errorf("user must be either a string or an int64, got %T", user)
```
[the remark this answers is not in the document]
thozza: It can be both, therefore we can not always `panic`...

1. If one defines a new `fsnode` node type in the code with values which do not validate, then it would be a programming error.
2. The validation is later (ab)used when validating user input from the Blueprint customization (`blueprint` package and `cloudapi` package) and returning an error to the user if they provided invalid input. This was my way of not having to implement the validation in 3 places, but only in one place.

thozza:
> @thozza looks great! Thanks for tweaking the commit vs installer deployment. Two small questions:
> 
> 1. if a user creates a unit file, will the ordering in osbuild allow the [customization.services] to enable/disable the new units?

Yes, this works and it is already used in the ostree test case which is part of this PR. In short the test does this:

1. Create `/etc/custom_file.txt` with known content.
2. Create `/etc/systemd/system/custom.service` with `ExecStart=/usr/bin/false`.
3. Create a directory `/etc/systemd/system/custom.service.d/` and an `override.conf` inside it with `ExecStart=/usr/bin/cat /etc/custom_file.txt`.
4. Enable `custom.service` using BP Services customization.
5. Check that the `custom.service` is enabled on the booted system, that it was overridden by the drop-in and that the content of `/etc/custom_file.txt` is logged in the journal since the last boot.

> 2. We're pasting in plain text in between quotes. I haven't gotten my hands on this, nor am I a TOML person, so apologies if this is a dumb question. Will users have to escape quotes that are contained inside the pasted text or anything like that?

No need to apologize, I think that there are very few "TOML people" if any 😅 

[TOML allows multiple ways to specify strings](https://toml.io/en/v1.0.0#string) and some of them would require to escape quotes. All of these should be equivalent:

```toml
[[customizations.files]]
path = "/etc/file.txt"
data = "key=\"value\"\n"
```

```toml
[[customizations.files]]
path = "/etc/file.txt"
data = 'key="value"\n'
```

```toml
[[customizations.files]]
path = "/etc/file.txt"
data = """key="value"
"""
```

internal/fsnode/file.go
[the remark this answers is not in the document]
achilleas-k: I don't mind either way.  :)

internal/fsnode/fsnode.go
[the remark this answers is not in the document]
achilleas-k: Right.  Sounds good to me.

internal/disk/disk.go
achilleas-k: I see your point.  Let's keep it the way it is now then, in the `blueprint` package.
I'm coming around to your way of thinking of having the "internal" structures never import or refer to neither `blueprint` nor `osbuild`.

internal/disk/disk.go
achilleas-k: No worries then.

ondrejbudai (commented):
This is pure awesomeness! Some small suggestions inline.

internal/fsnode/fsnode.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Check that the mode is valid
+	if f.mode != nil && *f.mode&os.ModeType != 0 {
+		return fmt.Errorf("mode must not contain file type bits")
+	}
+
+	// Check that the user and group are valid
+	switch user := f.user.(type) {
+	case string:
+		nameRegex := regexp.MustCompile(usernameRegex)
```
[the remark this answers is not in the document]
thozza: Since the code is covered by unit tests, I think that it should be fine to keep it as is...

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,284 @@
...
+	switch mode := dataMap["mode"].(type) {
+	case string:
+		dir.Mode = mode
+	case nil:
+		break
+	default:
+		return fmt.Errorf("UnmarshalTOML: mode must be a string")
+	}
+
+	switch ensure_parents := dataMap["ensure_parents"].(type) {
```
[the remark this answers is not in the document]
thozza: Ah yes, you are definitely right.... This is the Golang naming convention...

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,284 @@
...
+	*f = file
+	return nil
+}
+
+// Custom JSON unmarshalling for FileCustomization with validation
+func (f *FileCustomization) UnmarshalJSON(data []byte) error {
+	type fileCustomization FileCustomization
+
+	var filePrivate fileCustomization
+	if err := json.Unmarshal(data, &filePrivate); err != nil {
```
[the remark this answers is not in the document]
thozza: Honestly, I dream of us moving to a more sane TOML library one cay and replacing the `UnmarshalTOML` with something more like `UnmarshalJSON`. The current way TOML unmarshaling way is IMO crazy.

Also it feels cleaner to me to not use TOML unmarshaler inside JSON one...

internal/cloudapi/v2/handler.go
```diff
@@ -187,6 +187,106 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 		}
 	}
 
+	if request.Customizations != nil && request.Customizations.Directories != nil {
```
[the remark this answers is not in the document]
thozza: yeah, that is a leftover, good catch!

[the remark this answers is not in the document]
thozza: Fixed

thozza:
I made changes based on reviews. Rebased on top of `main` and also dropped 4 commits, which were related to deploy-time ostree image types and were leftovers from the times when the customization was aiming for them (it is not any more, it instead targets ostree commit image types).

tools/test-case-generators/format-request-map.json
```diff
@@ -170,114 +170,6 @@
...
-              "name": "group2",
-              "gid": 1050
-            }
-          ]
-        }
-      }
-    },
-    "no-image-info": true,
-    "overrides": {}
-  },
```
achilleas-k, line 280: :tada: :fishsticks:

test/cases/api.sh
```diff
@@ -330,6 +330,43 @@ EndOfMessage
 fi
 export SUBSCRIPTION_BLOCK
 
+# Define the customizations for the images here to not have to repeat them
+# in every image-type specific file.
+case "${IMAGE_TYPE}" in
+  # The Directories and Files customization is not supported for these image types.
+  "$IMAGE_TYPE_EDGE_COMMIT"|"$IMAGE_TYPE_IOT_COMMIT"|"$IMAGE_TYPE_EDGE_INSTALLER"|"$IMAGE_TYPE_IMAGE_INSTALLER")
```
achilleas-k, line 337: Didn't we decide we're supporting commits but not `edge-installer`, `edge-raw-image`, and `edge-simplified-installer`?
I see we're testing it in the ostree tests, so is this just for the API?

Also the image installer should be okay to support.  But maybe we can add that later separately.

achilleas-k (commented):
Looks great!  I have one question but I think everything else is good to go.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 14db83acb1c508cb786b765b2ed77cfcc99ace4d with the main merge-base 004c4e8acc2513b6030d2a73614e5a2f0b2a5d1d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3802144968/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3802144968/artifacts/browse

test/cases/api.sh
thozza: > Didn't we decide we're supporting commits but not `edge-installer`, `edge-raw-image`, and `edge-simplified-installer`?

Yes.

>  I see we're testing it in the ostree tests, so is this just for the API?

So I ended up not focusing on testing the commit image in the API test, since it is tested in its own test case. But thinking about this, it should be possible to inspect the custom files in the commit using `ostree`.

Also there is probably no good reason why should I not at least enable the customization for the commit image type.

> Also the image installer should be okay to support. But maybe we can add that later separately.

I already have the support for the `raw-image` and `simplified-installer` implemented on another branch. I removed all of it from this PR, because the decision is to go with just the commit and container image types for 8.8 / 9.2 which this PR targets.

thozza:
CI Failures:

- ostree / F37 - not caused by this PR, will be fixed by https://github.com/osbuild/osbuild-composer/pull/3292
- vsphere - DNS issues 🤷

test/cases/api.sh
thozza: I added the customization and test for ostree commit in API test.

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,468 @@
...
+// DirectoryCustomization represents a directory to be created in the image
+type DirectoryCustomization struct {
+	// Absolute path to the directory
+	Path string `json:"path" toml:"path"`
+	// Owner of the directory specified as a string (user name), int64 (UID) or nil
+	User interface{} `json:"user,omitempty" toml:"user,omitempty"`
+	// Owner of the directory specified as a string (group name), int64 (UID) or nil
+	Group interface{} `json:"group,omitempty" toml:"group,omitempty"`
+	// Permissions of the directory specified as an octal number
+	Mode string `json:"mode,omitempty" toml:"mode,omitempty"`
```
supakeen, line 36: The `User` and `Group` are optional, but not the mode?

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,468 @@
...
+	// Absolute path to the file
+	Path string `json:"path" toml:"path"`
+	// Owner of the directory specified as a string (user name), int64 (UID) or nil
+	User interface{} `json:"user,omitempty" toml:"user,omitempty"`
+	// Owner of the directory specified as a string (group name), int64 (UID) or nil
+	Group interface{} `json:"group,omitempty" toml:"group,omitempty"`
+	// Permissions of the file specified as an octal number
+	Mode string `json:"mode,omitempty" toml:"mode,omitempty"`
+	// Data is the file content in plain text
+	Data string `json:"data,omitempty" toml:"data,omitempty"`
```
supakeen, line 191: I would prefer to be able to have both `DataAscii` and `DataBytes` to be able to add binary files, we can represent the latter as one of the base encodings in the json/toml?

internal/blueprint/fsnode_customizations.go
```diff
@@ -0,0 +1,468 @@
...
+	if len(duplicatePaths) > 0 {
+		return fmt.Errorf("duplicate files / directory customization paths: %v", duplicatePaths)
+	}
+
+	invalidFSNodes := make([]string, 0)
+	checkedPaths := make(map[string]bool)
+	// Sort the paths so that we always check the longest paths first. This
+	// ensures that we don't check a parent path before we check the child
+	// path.
+	sort.Sort(sort.Reverse(sort.StringSlice(nodesPaths)))
```
supakeen, line 376: I might be overlooking something but does this sort on strlen of each path? If it does it should sort on dirsep-count.

internal/blueprint/fsnode_customizations.go
thozza: Not sure what do you mean. `mode` is optional. The default value for a string (empty string) is considered to be an unset value. I could have used pointer to string, but there would be no real benefit of it, and at the same time the use would get harder when using string literals (e.g. in unit tests).

internal/blueprint/fsnode_customizations.go
thozza: I'm with you on this, but the requirement is to support only plain text data for MVP.

As I mentioned in the description, adding support for base64 encoded (potentially binary) data would be super simple, since internaly it's all bytes. My plan would be to later add `data_base64` property later on to hold these. Since the expectation is that users will more often use the "plain text" input, I didn't add "Ascii" into the property name. There is nothing blocking us from renaming the Structure member name to `DataAscii` once the support for base64 encoded content is added, but I didn't see value in doing so now.

internal/blueprint/fsnode_customizations.go
thozza: You are probably onto something here... 🤔 While the result will be IMO still correct, the optimization does not work as I intended... Will think about this a bit.

internal/blueprint/fsnode_customizations.go
achilleas-k: Might also make sense to keep just `Data` and have a toggle for `Binary`?  Or an enum encoding/format, but probably not as simple.  A boolean for `Binary = true/false` would make it easier to handle a single data field without needing to check or tell users that they can't use both.

internal/blueprint/fsnode_customizations.go
achilleas-k: I think I get what your'e saying but I don't see how it can cause an issue.  We only really care that paths are sorted so child paths are ahead of their own parent paths, which this should do.  A path with fewer components being ahead of a path with more components shouldn't cause issues if the former isn't a parent of the latter, right?

thozza: Yeah, there is no impact on the result being correct, but the logic is not correct. I'll just soft the slice using a custom comparison function which will use length based on dir separator.

test/cases/api.sh
achilleas-k: > > Also the image installer should be okay to support. But maybe we can add that later separately.

> I already have the support for the raw-image and simplified-installer implemented on another branch. I removed all of it from this PR, because the decision is to go with just the commit and container image types for 8.8 / 9.2 which this PR targets.

I mean the non-ostree Anaconda installer that installs a traditional system (`image-installer`).

internal/blueprint/fsnode_customizations.go
supakeen: Understood. It's a Go-ism :)

test/cases/api.sh
achilleas-k: We should instead test it in https://github.com/osbuild/osbuild-composer/blob/main/test/cases/installers.sh in a follow-up.

thozza:
- Rebased on `main` to get the F37 fix for ostree test.
- Fixed path sorting to be based on dir separator count in `ValidateDirFileCustomizations()` in the `blueprint` package.
- Added workaround for 💩 DNS making VSphere tests to fail.
- Added dependency on the latest osbuild release (v80) and made use of the `remove_destination` option in the copy stage to not follow symlinks for files customization.
- Pinned osbuild v80 in Schutzfile, since the release happened only yesterday and there was not enough time for it to get to our RPMRepo snapshots and also stable / nightly repos of respective distributions.

achilleas-k (dismissed):
LGTM!  Great work!  :rocket:

supakeen (dismissed):
Nice.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c56c5f6c4653d67e636381a90b7321aa7ac284bd with the main merge-base b68cf83a8e2d4a1070ff621d26d6c2bf5a98f0f9).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3806639592/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3806639592/artifacts/browse

thozza:
The VSphere workaround has no effect on the test failure and I'm out of ideas, since all three DNS resolvers in `/etc/resolv.conf` are unstable... 😕

thozza:
I dropped the commit with "DNS workaround", since it had no effect. IT is working on fixing the DNS resolver in `us-east-1`.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ec105af38bf723ecba9c152e1d014ad140eaf431 with the main merge-base b68cf83a8e2d4a1070ff621d26d6c2bf5a98f0f9).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3811985109/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3811985109/artifacts/browse

achilleas-k (approved):

PR #3285, opened 2023-02-14
`edge-simplified-installer` allows User & Group customizations

7flying:
This PR is about adding support for User customizations on `edge-simplified-installer` image types, and about giving a basic indication that user customizations on `edge-commit` and `edge-container` are deprecated and will no longer be supported in future RHEL edge versions.

Note the "basic indication" on giving the deprecation warning, currently it simply logs out the deprecation notice. I will add another PR in the near future handling the warning via weldr (working on that), this is simply so that QE can test user customizations on `edge-simplified-installer`.

--- 

Updated to also allow Group customizations and to give the same basic deprecation warning.

--- 

### DIFF
--- a/internal/distro/rhel9/imagetype.go
+++ b/internal/distro/rhel9/imagetype.go
@@ -2,6 +2,7 @@ package rhel9
 
 import (
 	"fmt"
+	"log"
 	"math/rand"
 	"strings"
 
@@ -379,6 +380,17 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 	// check the checksum instead of the URL, because the URL should have been used to resolve the checksum and we need both
 	if t.name == "edge-raw-image" && options.OSTree.FetchChecksum == "" {
 		return fmt.Errorf("edge raw images require specifying a URL from which to retrieve the OSTree commit")
+	}
+
+	// warn that user & group customizations on edge-commit, edge-container are deprecated
+	// TODO(edge): directly error if these options are provided when rhel-9.5's time arrives
+	if t.name == "edge-commit" || t.name == "edge-container" {
+		if customizations.GetUsers() != nil {
+			log.Printf("Please note that user customizations on %q image type are deprecated and will be removed in the near future\n", t.name)
+		}
+		if customizations.GetGroups() != nil {
+			log.Printf("Please note that group customizations on %q image type are deprecated and will be removed in the near future\n", t.name)
+		}
 	}
 
 	if kernelOpts := customizations.GetKernel(); kernelOpts.Append != "" && t.rpmOstree && t.name != "edge-raw-image" && t.name != "edge-simplified-installer" {
--- a/internal/distro/rhel9/imagetype.go
+++ b/internal/distro/rhel9/imagetype.go
@@ -332,7 +332,7 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 		}
 
 		if t.name == "edge-simplified-installer" {
-			allowed := []string{"InstallationDevice", "FDO", "Ignition", "Kernel"}
+			allowed := []string{"InstallationDevice", "FDO", "Ignition", "Kernel", "User", "Group"}
 			if err := customizations.CheckAllowed(allowed...); err != nil {
 				return fmt.Errorf("unsupported blueprint customizations found for boot ISO image type %q: (allowed: %s)", t.name, strings.Join(allowed, ", "))
 			}
--- a/test/cases/ostree-simplified-installer.sh
+++ b/test/cases/ostree-simplified-installer.sh
@@ -415,6 +415,14 @@ groups = []
 [customizations]
 installation_device = "/dev/vda"
 
+[[customizations.user]]
+name = "simple"
+description = "Administrator account"
+password = "\$6\$GRmb7S0p8vsYmXzH\$o0E020S.9JQGaHkszoog4ha4AQVs3sk8q0DvLjSMxoxHBKnB2FBXGQ/OkwZQfW/76ktHd0NX5nls2LPxPuUdl."
+key = "${SSH_KEY_PUB}"
+home = "/home/simple/"
+groups = ["wheel"]
+
 [customizations.fdo]
 manufacturing_server_url="http://${FDO_SERVER_ADDRESS}:8080"
 diun_pub_key_insecure="true"
@@ -509,6 +517,7 @@ check_result
 greenprint "🕹 Get ostree install commit value"
 INSTALL_HASH=$(curl "${PROD_REPO_URL}/refs/heads/${OSTREE_REF}")
 
+# User simple in simplified-installer
 # Add instance IP address into /etc/ansible/hosts
 sudo tee "${TEMPDIR}"/inventory > /dev/null << EOF
 [ostree_guest]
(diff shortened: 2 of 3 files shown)


7flying:
CC @henrywang @runcom

7flying:
(looking into the Unit test failures)

henrywang:
@7flying I'll add test in this PR. We can land this PR if test passed. Thanks!

internal/distro/rhel9/imagetype.go
```diff
@@ -381,6 +382,12 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 		return fmt.Errorf("edge raw images require specifying a URL from which to retrieve the OSTree commit")
 	}
 
+	// warn that user customizations on edge-commit, edge-container are deprecated
+	// TODO(edge): directly error if these options are provided when rhel-9.5's time arrives
+	if (t.name == "edge-commit" || t.name == "edge-container") && customizations.GetUsers() != nil {
+		log.Printf("Please note that user customizations on %q image type are deprecated and will be removed in the near future\n", t.name)
```
runcom, line 388: do we need this? isn't this coming from weldr or something?

7flying: yes, I'm working on it, I've left it as it is for the time being

7flying: but I mean, the thing is that the warning will originate from osbuild-composer, which has to be sent back to weldr

henrywang:
@7flying @runcom Hi, just a double check with you. Is this patch is for RHEL 9 only? Thanks!

internal/distro/rhel9/images.go
```diff
@@ -55,9 +55,13 @@ func osCustomizations(
 		osc.ExcludeDocs = *imageConfig.ExcludeDocs
 	}
 
-	if !t.bootISO {
+	if !t.bootISO || (t.name == "edge-simplified-installer" && !common.VersionLessThan(t.arch.distro.osVersion, "9.2")) {
```
thozza, line 58: What is the actual intention of this condition?

If it is to set the user / group customization for edge-simplified-installer, then this is not the right place to do it, since the image type does not use the `os` pipeline at all. The customization must be instead set in the commit-deploy pipeline. 

Looking at https://github.com/osbuild/osbuild-composer/blob/96b42b57919440b798b2171f45792efcf1c148f1/internal/distro/rhel9/images.go#L409-L410 everything seems to be already hooked up.

internal/distro/rhel9/imagetype.go
```diff
@@ -332,7 +332,7 @@ func (t *imageType) checkOptions(customizations *blueprint.Customizations, optio
 		}
 
 		if t.name == "edge-simplified-installer" {
-			allowed := []string{"InstallationDevice", "FDO", "Ignition", "Kernel"}
+			allowed := []string{"InstallationDevice", "FDO", "Ignition", "Kernel", "User"}
```
thozza, line 335: note: This is the actual reason why the customization "started to work" for simplified-installer.

q
```diff
@@ -0,0 +1,83 @@
+[1;33mdiff --git i/test/cases/ostree-simplified-installer.sh w/test/cases/ostree-simplified-installer.sh[m
```
thozza, line 1: This file looks as if it was added to the commit by mistake... (https://github.com/osbuild/osbuild-composer/pull/3285/commits/5a4b914dd4a4448b8bc77fbfdc935a29f1c77686)

thozza (changes requested):
Thanks for the PR.

I think that some changes are not needed at all or are in the wrong place. In addition, I would like to suggest that the customization is enabled also in RHEL-8 and that we do not selectively disable it for minor releases. While the "feature" will be available on-premise in RHEL only in new releases, there is no need to not allow it for older ones in the service or if using new enough osbuild-composer version. Such selective unavailability of features just complicates the code without any technical reason (given that there is none).

internal/distro/rhel9/images.go
7flying: very true :+1:

7flying:
@henrywang let me know if the updated tests make sense to you, if so, feel free to squash the commit with your original one

7flying:
> Thanks for the PR.
> 
> I think that some changes are not needed at all or are in the wrong place. In addition, I would like to suggest that the customization is enabled also in RHEL-8 and that we do not selectively disable it for minor releases. While the "feature" will be available on-premise in RHEL only in new releases, there is no need to not allow it for older ones in the service or if using new enough osbuild-composer version. Such selective unavailability of features just complicates the code without any technical reason (given that there is none).

Thanks for the review :pray: 

Updated after your comments. And yes, there is no technical effort on allowing Users on RHEL-8 so we might as well add them.

test/cases/ostree-simplified-installer.sh
```diff
@@ -102,24 +102,29 @@ SSH_KEY_PUB=$(cat "${SSH_KEY}".pub)
 # kernel-rt package name (differs in CS8)
 KERNEL_RT_PKG="kernel-rt"
 
+# Only RHEL 9.2 and above support user customizations in simplified installer blueprint
 case "${ID}-${VERSION_ID}" in
     "rhel-8.8")
         OSTREE_REF="rhel/8/${ARCH}/edge"
         OS_VARIANT="rhel8-unknown"
+        USER_IN_SIMPLIFIED="false"
```
thozza: just note that if this change will get into 9.2 before ITM26, then it will end up in 8.8 as well.

test/cases/ostree-simplified-installer.sh
```diff
@@ -102,24 +102,29 @@ SSH_KEY_PUB=$(cat "${SSH_KEY}".pub)
...
     "rhel-9.2")
         OSTREE_REF="rhel/9/${ARCH}/edge"
         OS_VARIANT="rhel9-unknown"
+        USER_IN_SIMPLIFIED="true"
         ;;
     "centos-8")
         OSTREE_REF="centos/8/${ARCH}/edge"
         OS_VARIANT="centos8"
         KERNEL_RT_PKG="kernel-rt-core"
+        USER_IN_SIMPLIFIED="false"
```
thozza: just note that if this change will get into centos-8 as well, so it should be possible to test it there if it is desired.

thozza (dismissed):
LGTM 👍

I added a comment to the tests, that it should be possible to test this also on c8s and el8.8 if this gets into el9.2 before ITM26 (because we are pushing the same upstream releases into el8 and el9).

I also wonder whether you don't want to make the same change (enable it and add deprecation notice) also for **Group** customization... 🤔

test/cases/ostree-simplified-installer.sh
henrywang: > @henrywang let me know if the updated tests make sense to you, if so, feel free to squash the commit with your original one

Thanks! Commit squashed.

thozza (dismissed):

7flying:
Sorry for the updates and the re-updates, also added the Group customization options and squashed the commits
CC' @runcom @henrywang

thozza (approved):

thozza:
rebased on `main`

7flying:
Rebased in main after PR #3292  , since we need that test fixed too

7flying:
We are also experiencing issues with vmware (apart from vsphere) @ondrejbudai

thozza:
> We are also experiencing issues with vmware (apart from vsphere) @ondrejbudai

Yes, we know about the problem. It is caused by flaky DNS resolvers provided from the network.

thozza (approved):

PR #3291, opened 2023-02-16
distro: assign pipeline-specific repos to package sets

### LINKED ISSUE #3290: Non-global package sets are ignored
When a package set in a request is restricted to a specific pipeline, it is ignored and it never shows up in any depsolve request.

See https://issues.redhat.com/browse/COMPOSER-1902 for an example of this causing a failure (compose request: https://download.eng.bos.redhat.com/brewroot/work/tasks/9625/50749625/compose-request.json).
The packages `python3-PyYAML` and `python3-iniparse` are in the repository that is restricted to the `build` pipeline and the depsolve fails because the repository is dropped.

achilleas-k:
Assign pipeline-specific repositories to package sets before returning the package set chains.  This is a workaround for a bug where repositories with pipeline restrictions are ignored.

In the new distro definitions, repositories are assigned to pipelines while building the manifest.  The manifest then returns package set chains with repository information from the GetPackageSetChains() method.  However, the manifest is only initialised with the global repos and the package set specific repos are ignored.

~~A more correct fix for this issue would be to pass all repositories to each pipeline initialisation function (e.g., `NewBuild()`) and have each pick the repositories that matches its own name (+ any that have no restrictions).  However this would require more changes and would take more time to implement.  I think the bug is critical enough to fix quickly and clean up later.~~

## Update

Wrote a cleaner fix and unit test to catch any regressions.

## Bugs
Fixes #3290
Fixes COMPOSER-1902

### DIFF
--- a/internal/distro/distro_test.go
+++ b/internal/distro/distro_test.go
@@ -3,6 +3,7 @@ package distro_test
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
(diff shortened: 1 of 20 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c4d8deaa4267d87076e58196167b9b2de4e45617 with the main merge-base 96b42b57919440b798b2171f45792efcf1c148f1).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3783234184/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3783234184/artifacts/browse

achilleas-k:
Wrote a cleaner fix and unit test to catch any regressions.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 9362c92402d0db24f3911f42fa35883d62bf8802 with the main merge-base 96b42b57919440b798b2171f45792efcf1c148f1).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3792221487/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3792221487/artifacts/browse

achilleas-k:
I broke the `payload_repository` handling in the cloud API.

internal/distro/rhel9/imagetype.go
```diff
@@ -305,7 +288,7 @@ func (t *imageType) PackageSets(bp blueprint.Blueprint, options distro.ImageOpti
 	}
 
 	// create a manifest object and instantiate it with the computed packageSetChains
-	manifest, err := t.initializeManifest(&bp, options, globalRepos, packageSets, containers, 0)
+	manifest, err := t.initializeManifest(&bp, options, repos, packageSets, containers, 0)
```
thozza, line 291: This is relevant for all distros: So from now on, the `packageSets` won't ever include any `Repositories` set when passed to `initializeManifest`, right?

If yes, then we should probably rethink how relevant are things like passing the package sets into specific pipelines such as https://github.com/osbuild/osbuild-composer/blob/96b42b57919440b798b2171f45792efcf1c148f1/internal/distro/rhel9/images.go#L492 then resulting in https://github.com/osbuild/osbuild-composer/blob/96b42b57919440b798b2171f45792efcf1c148f1/internal/image/image_installer.go#L77 and repositories (there should not be any) in https://github.com/osbuild/osbuild-composer/blob/96b42b57919440b798b2171f45792efcf1c148f1/internal/manifest/anaconda.go#L125-L136

If my assessment is correct, then the code in anaconda pipeline (and other that handle repos, such as OS pipeline) handling extra repos should become unnecessary with this change. IOW the `packageSets` should never contain any repos, `ExtraBaseRepos` in pipeline definitions should be probably removed and `packageSets` passed to `initializeManifest()` with repositories set should be probably considered a programming error.

But I'm not 100% sure about all of this... 🤔

thozza (commented):
Overall this looks very good. I added a few nitpicks and one question related to the effect of not setting any repositories in the package sets used when initializing Manifest.

internal/distro/distro_test.go
```diff
@@ -235,3 +236,251 @@ func TestImageTypePipelineNames(t *testing.T) {
...
+					}
+				}
+			}
+		})
+	}
+}
+
+// a very basic implementation of a Set of strings
+type stringSet struct {
+	elems map[string]interface{}
```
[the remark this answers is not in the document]
achilleas-k: Thanks.  I'll change it.

internal/distro/rhel9/imagetype.go
achilleas-k: All good points.  I think you're right.  Need to dig a bit and clarify everything while fixing the payload repositories.

achilleas-k:
Did a rebase and rewrote a few things.

This bug made me realise again that I really don't like the flow of the current manifest generation.  I think the components are all good, but the way we need to initialise a manifest twice makes it very hard to trace things.
On top of that, and more relevant to this bug, the way repositories and package sets are handled is very confusing.  Like @thozza mentioned, there are places where we can attach extra repos or packages to pipelines (like "os" and "anaconda-tree"), but there's no way to add extra repositories to "build" for example, other than passing them through with the globals, which is what cause the bug we're solving here to begin with.

Now we're using both mechanisms to get repos into pipelines:
1. Passing all repos through and having pipelines filter for the repos that specify the pipeline name in their `PackageSets` list.
2. Adding repos explicitly to a pipeline through `ExtraBasePackages` (see [imageInstallerImage()](https://github.com/osbuild/osbuild-composer/blob/eaee7870ba977f3f847c12b3915d9876c21dc54c/internal/distro/rhel9/images.go#L507)).

I'd like to make things more explicit, and I have thoughts about getting rid of the double generation, but that's a much bigger change.

internal/distro/rhel9/imagetype.go
achilleas-k: This really broke my brain for a whole day.  See my [followup comment below](https://github.com/osbuild/osbuild-composer/pull/3291#issuecomment-1440960864).

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 4ce05cb4e6b6a98e3b7841d8b07758f7698d1019 with the main merge-base eaee7870ba977f3f847c12b3915d9876c21dc54c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3819564995/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3819564995/artifacts/browse

achilleas-k:
VMWare/vSphere failures from the DNS issue.
Otherwise green.

ondrejbudai (approved):
LGTM!

thozza (approved):
Looks good, but I completely agree that we need to simplify the manifest instantiation and also the repo and package sets handling a bit. My 🧠  is on 🔥  from trying to figure everything out and I'm still not 100% sure 😵

PR #3297, opened 2023-02-20
Fix CIV_OPTIONS bug

jabia99:
With the previous method, we encountered a lot of problems when sending the options via an array. Separate the options into different variables that can be set in CIV CI

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -199,14 +199,13 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
-  if [ -z "${CIV_OPTIONS:-}" ]; then
-    echo "ERROR: Please provide the variable CIV_OPTIONS"
-    exit 1
-  fi
 else
   # If not, get prod container
   TAG="prod"
-  CIV_OPTIONS=( -r=/tmp/resource-file.json -d -o=/tmp/report.xml -m="not pub" )
+  CIV_OPTION_R="-r=/tmp/resource-file.json"
+  CIV_OPTION_D="-d"
+  CIV_OPTION_O="-o=/tmp/report.xml"
+  CIV_OPTION_M="-m not pub"
 fi
 
 CONTAINER_CLOUD_IMAGE_VAL="quay.io/cloudexperience/cloud-image-val-test:$TAG"
@@ -234,6 +233,8 @@ if [ "$ARCH" == "aarch64" ]; then
     sed -i s/t3.medium/a1.large/ "${TEMPDIR}/resource-file.json"
 fi
 
+# The weird sintax with the CIV_OPTION variables is due to problems with bash quoting and
+# having options with whitespaces in them. Please do not spend more hours trying to fix it
 sudo "${CONTAINER_RUNTIME}" run \
     -a stdout -a stderr \
     -e AWS_ACCESS_KEY_ID="${V2_AWS_ACCESS_KEY_ID}" \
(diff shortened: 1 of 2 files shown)


test/cases/aws.sh
```diff
@@ -241,7 +240,14 @@ sudo "${CONTAINER_RUNTIME}" run \
...
     "${CONTAINER_CLOUD_IMAGE_VAL}" \
-    python cloud-image-val.py "${CIV_OPTIONS[@]}" && RESULTS=1 || RESULTS=0
+    python cloud-image-val.py \
+    ${CIV_OPTION_R:+"$CIV_OPTION_R"} \
+    ${CIV_OPTION_O:+"$CIV_OPTION_O"} \
+    ${CIV_OPTION_T:+"$CIV_OPTION_T"} \
+    ${CIV_OPTION_M:+"$CIV_OPTION_M"} \
+    ${CIV_OPTION_D:+"$CIV_OPTION_D"} \
+    ${CIV_OPTION_S:+"$CIV_OPTION_S"} \
+    && RESULTS=1 || RESULTS=0
```
atodorov, line 250: I don't understand what's the intention here? Do you mean to override the values of these variables based on externally defined values but have some sane defaults ? 

If that's the case I think you should move the defaults in the snippet above (or even have it unconditional) and simply use the values of these variables when calling `cloud-image-val.py`.

jrusz: Yeah I don't get this either, if you want defaults you can just put `${CIV_OPTION_R:-"my default value"} \` this will ensure that if you've set the variable sometime before it's going to use that but if you didn't set any value to it until now it will have the value of the string "my default value"

jabia99: I'm having the cloud-image-val options set into variables. This script is being executed both in osbuild-composer ci and cloud-image-val ci. I'm using this sintax as if I just used "${var}" the command would fail if var is empty

jabia99: There are no default options. If the script is being executed in osbuild-composer ci the options are set in the else clause in line 205. But not all the options are used all the time, so if an option is not set I don't want the command to fail

jrusz: It won't fail if you do something like this for example: `${CIV_OPTION_R:-} \` ... Same syntax was used on the now removed line 202

jabia99: But I have to quote it to get it passed shellcheck. Something like "${CIV_OPTION_R:-}". That when it fails

atodorov (approved):

test/cases/aws.sh
```diff
@@ -234,14 +233,23 @@ if [ "$ARCH" == "aarch64" ]; then
...
     -v "${TEMPDIR}":/tmp:Z \
     "${CONTAINER_CLOUD_IMAGE_VAL}" \
-    python cloud-image-val.py "${CIV_OPTIONS[@]}" && RESULTS=1 || RESULTS=0
+    python cloud-image-val.py \
+    ${CIV_OPTION_R:+"$CIV_OPTION_R"} \
+    ${CIV_OPTION_O:+"$CIV_OPTION_O"} \
+    ${CIV_OPTION_T:+"$CIV_OPTION_T"} \
+    ${CIV_OPTION_M:+"$CIV_OPTION_M"} \
+    ${CIV_OPTION_D:+"$CIV_OPTION_D"} \
+    ${CIV_OPTION_S:+"$CIV_OPTION_S"} \
```
thozza, line 251: `CIV_OPTION_T` and `CIV_OPTION_S` don't seem to be set anywhere. Is this intentional?

test/cases/azure.sh
```diff
@@ -238,7 +239,14 @@ sudo "${CONTAINER_RUNTIME}" run \
...
     -v "${TEMPDIR}":/tmp:Z \
     "${CONTAINER_CLOUD_IMAGE_VAL}" \
-    python cloud-image-val.py "${CIV_OPTIONS[@]}" && RESULTS=1 || RESULTS=0
+    python cloud-image-val.py \
+    ${CIV_OPTION_R:+"$CIV_OPTION_R"} \
+    ${CIV_OPTION_O:+"$CIV_OPTION_O"} \
+    ${CIV_OPTION_T:+"$CIV_OPTION_T"} \
+    ${CIV_OPTION_M:+"$CIV_OPTION_M"} \
+    ${CIV_OPTION_D:+"$CIV_OPTION_D"} \
+    ${CIV_OPTION_S:+"$CIV_OPTION_S"} \
```
thozza, line 248: `CIV_OPTION_T` and `CIV_OPTION_S` don't seem to be set anywhere. Is this intentional?

test/cases/aws.sh
jabia99: It is, we might want to set this variables when the script is being executed from CIV ci

thozza (approved):

thozza:
Admin-merging, because CI is till blocked on DNS issues and test cases affected by this PR passed.

Issue #3309, opened 2023-02-28
Cannot update snapshot because of ansible error on centos8


yih-redhat:
I triggered an update snapshot job for osbuild-composer, https://github.com/osbuild/osbuild-composer/pull/3305 , some ostree test failed on centos8 because of "ERROR: Ansible requires blocking IO on stdin/stdout/stderr. Non-blocking file handles detected: <stdout>",  looks like an ansible issue, details can be found in this job https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3824249899, and:
1. It only happens on Centos Stream 8, it works well on RHEL and Fedora
2. I logon the CI runner machine(of the failed job) and run same test again, it can pass. 
3. I tried same test in my own cs8 machine, it can pass

Raised this issue in ansible project, here is the discussion link https://github.com/ansible/ansible/issues/80108

Filed this bug to track this problem, working on it now.

yih-redhat:
Fixed.

PR #3311, opened 2023-02-28
Update rpmrepo snapshots to 20230223 (including necessary fixes and workarounds)

yih-redhat:
This pull request includes:

1. Fix bug https://github.com/osbuild/osbuild-composer/issues/3309 found in update snapshot pr.
2. Add workaround for bug https://bugzilla.redhat.com/show_bug.cgi?id=2162815

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -46,14 +46,14 @@
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-x86_64-updates-released-20221115"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-x86_64-updates-released-20230223"
           }
         ],
         "aarch64": [
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-aarch64-updates-released-20221115"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-aarch64-updates-released-20230223"
           }
         ]
       },
@@ -63,14 +63,14 @@
           {
             "title": "updates-modular",
             "name": "updates-modular",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-x86_64-updates-released-modular-20221115"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-x86_64-updates-released-modular-20230223"
           }
         ],
         "aarch64": [
           {
             "title": "updates-modular",
             "name": "updates-modular",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-aarch64-updates-released-modular-20221115"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f36/f36-aarch64-updates-released-modular-20230223"
           }
         ]
       }
@@ -123,14 +123,14 @@
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-x86_64-updates-released-20221124"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-x86_64-updates-released-20230223"
           }
         ],
         "aarch64": [
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-aarch64-updates-released-20221124"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-aarch64-updates-released-20230223"
           }
         ]
       },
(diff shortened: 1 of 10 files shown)


tools/fix-ansible.py
```diff
@@ -0,0 +1,10 @@
+import os
+import sys
+
+for handle in (sys.stdin, sys.stdout, sys.stderr):
+    try:
+        fd = handle.fileno()
+    except Exception:
+        continue
+
+    os.set_blocking(fd, True)
```
achilleas-k, line 10: I would rename the script to `ansible-blocking-io` or something similar and add a comment pointing to the ansible issue that's linked in the PR description (`#80108`).  From what I understand, this isn't a bug in anisble itself but a side-effect of a different change (either in ansible or somewhere else), so it'll be good to have some information about why we're doing this for the future.

achilleas-k (changes requested):

Schutzfile
```diff
@@ -123,14 +123,14 @@
           {
             "title": "updates",
             "name": "updates",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-x86_64-updates-released-20221124"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f37/f37-x86_64-updates-released-20230223"
```
achilleas-k, line 126: Let's revert the change in F37.  If all goes green, we can keep the changes for the other snapshots so we're testing newer versions until the F37 bug is fixed.

achilleas-k (changes requested):

tools/fix-ansible.py
yih-redhat: Done, renamed to ansible-blocking-io.py, and added more info in it about why I am doing this.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD a2881e2451610de7395588a96c96c0c2b6694824 with the main merge-base 72b223c06465790fe09313dcb61339f3569a46ab).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3860523169/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3860523169/artifacts/browse

tools/provision.sh
```diff
@@ -32,6 +32,11 @@ else
     sudo dnf install -y ansible koji
 fi
 
+# workaround for bug https://bugzilla.redhat.com/show_bug.cgi?id=2162815
+if [[ "${ID}" == "fedora" || "$VERSION_ID" == "37" ]]; then
```
thozza, line 36: You should have used `&&` instead of `||` here. The package gets now installed on any Fedora distro, which is not desired.

tools/ansible-blocking-io.py
```diff
@@ -0,0 +1,13 @@
+# This python script is to fix ansible error in CI test. It's not a bug of ansible, but a side-effect of a different change
```
thozza, line 1: Please add python shebang`#!/usr/bin/python3`

test/cases/ostree.sh
```diff
@@ -686,6 +686,10 @@ ansible_private_key_file=${SSH_KEY}
 ansible_ssh_common_args="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
 EOF
 
+# Fix ansible error https://github.com/osbuild/osbuild-composer/issues/3309
+greenprint "fix stdio file non-blocking issue"
+sudo python3 /usr/libexec/osbuild-composer-test/ansible-blocking-io.py
```
thozza, line 691: nitpick: You can drop `python3` once you add shebang to the script and make it executable. This applies to all files that include this.

thozza (commented):
Thanks for solving all CI issues. Your work is much appreciated ❤️ 

I added a few comments. I'm also not very happy with shipping `tools/ansible-blocking-io.py` as a separate script in `/usr/libexec`, but there is no better way around this if we do not want to duplicate the code in each script (which does not make sense). I also considered if it would be nicer to set the IO to blocking using a bash script instead, so that it can be just sourced by existing test cases. But that would not improve much (if anything), so it probably does not make sense.

Happy to approve this PR once you fix the missing shebang and condition in the `provision.sh` script.

tools/provision.sh
yih-redhat: Thanks for pointing this out, my mistake, fixed.

tools/ansible-blocking-io.py
yih-redhat: Done.

test/cases/ostree.sh
yih-redhat: Done.

yih-redhat:
> Thanks for solving all CI issues. Your work is much appreciated heart
> 
> I added a few comments. I'm also not very happy with shipping `tools/ansible-blocking-io.py` as a separate script in `/usr/libexec`, but there is no better way around this if we do not want to duplicate the code in each script (which does not make sense). I also considered if it would be nicer to set the IO to blocking using a bash script instead, so that it can be just sourced by existing test cases. But that would not improve much (if anything), so it probably does not make sense.
> 
> Happy to approve this PR once you fix the missing shebang and condition in the `provision.sh` script. Please also rebase on top of `main` while making changes.

Rebased with main and pushed again.
I agree it's better to set IO to blocking using bash, I will take a look of this later.

thozza (approved):
LGTM 👍

croissanne:
Packer is currently broken, see #3320. Admin merge one ostree is green?

PR #3314, opened 2023-03-01
koji: log unsuccessful requests only once

diaasami:
This pull request includes:

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -478,24 +478,17 @@ func customCheckRetry(ctx context.Context, resp *http.Response, err error) (bool
 		}
 	}
 
-	if !shouldRetry {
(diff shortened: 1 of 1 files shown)


thozza:
@diaasami the changes look good, but I would like to ask for more details on the intention of this change. E.g. why we do not want to log each retry any more? Happy to approve afterwards.

diaasami:
> @diaasami the changes look good, but I would like to ask for more details on the intention of this change. E.g. why we do not want to log each retry any more? Happy to approve afterwards.

The motivation for this is 1. not to log successful requests, which are not retried and 2. to not log twice.

we log retries, but this happens inside retryable-http so it's not visible here, what we are logging here is the decision to retry or not (which is based on the library function `rh.DefaultRetryPolicy`), if it's a yes we will see the logs of retrying, but if it's a no, we will not so we need to see that decision was taken and may be later change the retry policy.

PR #3315, opened 2023-03-01
tools/appsre-ansible: don't subscribe machines used for rpmbuild

croissanne:
We actually use rhui/cloud access images for rpmbuild as well. And the sync issue between cdn and rhui repos can cause issues when rpms are build against older packages.


thozza (approved):
Thank you!

diaasami (approved):

PR #3316, opened 2023-03-01
Add worker dockerfile for devel purposes

shaded-enmity:
The original worker dockerfile was installing osbuild from fedora repositories rather than from local checkout which interferes with docker-compose setup

This pull request includes:

### DIFF
--- a/distribution/Dockerfile-worker-devel
+++ b/distribution/Dockerfile-worker-devel
@@ -1,0 +1,35 @@
+FROM fedora:37 AS builder
+ENV GOBIN=/opt/app-root/src/go/bin
+# extra packages are needed
+# to compile osbuild
+RUN dnf install -y golang \
+  krb5-devel \
+  gpgme-devel \
+  libassuan-devel
+WORKDIR /osbuild-composer
+COPY . /osbuild-composer
+ENV GOFLAGS=-mod=vendor
+RUN go install ./cmd/osbuild-worker
+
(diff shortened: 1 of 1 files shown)


thozza:
@supakeen you had some fancy container setup that could use any version of osbuild and composer for worker container IIRC. I would like to check with him and ideally unify our devel containers (if possible) instead of adding another container.

shaded-enmity:
@thozza Aye, especially that the added dockerfile would need to be kept up to date alongside with the original dockerfile 

Not sure what @supakeen has, but I got another idea how to go about this - build arguments, something like:

```
# install osbuild
RUN if -e "$ARG_OSBUILD_REPO"; then git clone ... && make rpm && install"; else dnf install -y osbuild
```

That way, if no argument is supplied it falls back to current behaviour, if the argument is supplied it'll do build from git. In docker-compose I'd set the argument to point to my branch with my changes and my use case should be covered. It's more cumbersome than local setup, but it gets the job done.

supakeen:
My 'osbuild-getting-started' setup uses hardcoded paths to the upstream repositories but can check out any reference from there. I'm indeed using build arguments to set those refs:

https://github.com/osbuild/osbuild-getting-started/blob/main/src/ogsc/build/osbuild-composer/Containerfile#L25

Slightly different approach. I should pick up that project but until then I'd like to land this.

PR #3317, opened 2023-03-02
per-distro rpmmd cache directory

bcl:
This pull request includes:


bcl:
First pass at this. Tests don't pass yet :) But it does work, and dnf-json runs in parallel at startup with metadata for all 3 distros on f38 being populated at the same time.

It's a bit hard to test the cache size checks, so I'm not sure that part is right yet.

And unanswered question is, what do do with the old cache directories under ./rpmmd/ when this is upgraded? Remove them on upgrade, either in the spec file or on startup? Leave them? Expand the cleanup to remove them when they are too old?

achilleas-k:
> And unanswered question is, what do do with the old cache directories under ./rpmmd/ when this is upgraded? Remove them on upgrade, either in the spec file or on startup? Leave them? Expand the cleanup to remove them when they are too old?

An excellent question that I have no real answer for.  Is it normal to delete caches on upgrade through the spec file?  If that's something that's generally done and acceptable, I'd be for it.  composer basically "owns" the `/var/cache/osbuild-*` directories, right?

internal/dnfjson/dnfjson.go
```diff
@@ -102,13 +104,29 @@ type Solver struct {
...
+	// Full distribution string, eg. fedora-38, used to create separate dnf cache directories
+	// for each distribution.
+	distro string
+
 	subscriptions *rhsm.Subscriptions
 }
 
 // Create a new Solver with the given configuration. Initialising a Solver also loads system subscription information.
-func NewSolver(modulePlatformID string, releaseVer string, arch string, cacheDir string) *Solver {
+func NewSolver(modulePlatformID, releaseVer, arch, distro, cacheDir string) *Solver {
```
achilleas-k, line 115: My original thought on this issue was to have multiple solvers attached to the API, each initialised with a different cache root (one for each distro) under the main cache directory.  Changing the `Solver` to manage them itself is certainly cleaner, but I'm not sure how the cache management is affected now.  The `rpmCache` assumes that the top-level elements in it are rpm repository files and directories.  Now, they're distro subdirectories with repo files and directories under those.  In other words, the `rpmCache.repoElements` are now distro elements instead.

If you look at `updateCacheDirInfo`, you'll see this assumption is used to group elements per repo based on filename: https://github.com/osbuild/osbuild-composer/blob/7741fce11bd114ee950e6d8dd47e1950c1ee3b2f/internal/dnfjson/cache.go#L102-L112

The 64 characters are the hash we compute for the repository and depend on it to identify unique repos and reuse the same directories.  I think with this change to distros, the cleanup code will basically consider the distro subdirectories as "unknown" and ignore them.

We should either have separate `rpmCache` objects for each distro or, if we want to keep the cache management centralised, change the cleanup and file tracking code to work with the distro assumption.  With the latter solution, we could also solve our leftover cache issue by detecting and deleting top-level repo entries on cache creation.

achilleas-k (changes requested):

internal/dnfjson/dnfjson.go
bcl: That new function is called from a loop that iterates over the top level directories, the name being hashed should still be the same as it was before. But I'm not sure if my attempt works, I think I need to add some tests, or some prints at least :)

bcl: I've added a function (`CleanupOldCacheDirs`) that is called right before the initial metadata download and cache setup. It removes anything under ./rpmmd/ that it doesn't know about. Including old top-level cache files and directories, and old distros that have been removed from `supportedDistros`.

internal/dnfjson/cache_test.go
```diff
@@ -329,3 +321,64 @@ func TestDNFCacheCleanup(t *testing.T) {
 	_, ok := cache.Get("notreallyahash")
 	assert.False(t, ok)
 }
+
+func TestCleanupOldCacheDirs(t *testing.T) {
+	// Run the cleanup without the cache present and with dummy distro names
+	CleanupOldCacheDirs("/var/tmp/test-no-cache-rpmmd/", []string{"fedora-37", "fedora-38"})
```
achilleas-k, line 327: To avoid any issues with deleting files that aren't ours (unlikely, but stranger things have happened), maybe create another `t.TempDir()` and run this against a subdirectory of that new dir.

achilleas-k (approved):
Wonderful.  Love it!  Some very minor requests below but I'm good to merge as is.

internal/dnfjson/cache_test.go
bcl: Not sure what you mean. This is testing to make sure there is no panic when the directory doesn't exist.

achilleas-k: It's not important though.

achilleas-k (approved):
LGTM

achilleas-k (approved):

achilleas-k (approved):

PR #3319, opened 2023-03-02
weldr+distro: allow to send Manifest warnings on ComposeReply

7flying:
This adds a new field 'Warnings' to the `ComposeReply` struct, allowing to send back any warnings (e.g. deprecation notices, see #3285) when a compose build is requested.

This also changes the `Manifest` function on the `ImageType` interface so that any warnings detected during the `checkOptions` step of the manifest initialization are propagated back to the Welder API.

See also https://github.com/osbuild/weldr-client/pull/99 which handles the welder-client side of things.

This pull request includes:


7flying:
Taking care of the tests that I've missed to update

7flying:
@achilleas-k can you review this while I re-work the `welr-client` side of things with Brian?

internal/distro/fedora/distro.go
```diff
@@ -680,10 +680,11 @@ func (t *imageType) initializeManifest(bp *blueprint.Blueprint,
...
 	packageSets map[string]rpmmd.PackageSet,
 	containers []container.Spec,
-	seed int64) (*manifest.Manifest, error) {
+	seed int64) (*manifest.Manifest, []string, error) {
 
-	if err := t.checkOptions(bp.Customizations, options, containers); err != nil {
-		return nil, err
+	warnings, err := t.checkOptions(bp.Customizations, options, containers)
+	if err != nil {
+		return nil, warnings, err
```
thozza, line 687: I'm not sure if it makes sense to return warnings if there is an error... 🤔  This applies to all places where we check for an error.

I'm thinking whether we should return warnings only if we return `nil` as the error from this function.

internal/distro/fedora/distro.go
```diff
@@ -730,26 +731,27 @@ func (t *imageType) Manifest(customizations *blueprint.Customizations,
 	}
 	bp.Customizations = customizations
 
-	manifest, err := t.initializeManifest(bp, options, repos, nil, containers, seed)
+	manifest, warnings, err := t.initializeManifest(bp, options, repos, nil, containers, seed)
 	if err != nil {
-		return distro.Manifest{}, err
+		return distro.Manifest{}, warnings, err
```
thozza, line 736: Not related to your changed, but this got me thinking, why are we returning `distro.Manifest{}` instead of `nil` if there was an error... 🤔 (probably a question for @achilleas-k)

internal/distro/fedora/distro.go
```diff
@@ -730,26 +731,27 @@ func (t *imageType) Manifest(customizations *blueprint.Customizations,
...
 	}
 
-	return manifest.Serialize(packageSets)
+	ret, err := manifest.Serialize(packageSets)
+	return ret, warnings, err
 }
 
 // checkOptions checks the validity and compatibility of options and customizations for the image type.
-func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) error {
+func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) ([]string, error) {
```
thozza, line 744: Please extend the doc text with description of the first return value.

internal/distro/rhel7/distro.go
```diff
@@ -481,21 +483,21 @@ func overridePackageNames(packages []string) []string {
 }
 
 // checkOptions checks the validity and compatibility of options and customizations for the image type.
-func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) error {
+func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) ([]string, error) {
```
thozza, line 486: Please extend the doc text with description of the first return value.

internal/distro/rhel8/imagetype.go
```diff
@@ -346,31 +348,33 @@ func overridePackageNames(packages []string) []string {
 }
 
 // checkOptions checks the validity and compatibility of options and customizations for the image type.
-func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) error {
+func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) ([]string, error) {
```
thozza, line 351: Please extend the doc text with description of the first return value.

internal/distro/rhel9/imagetype.go
```diff
@@ -314,32 +316,34 @@ func (t *imageType) PackageSets(bp blueprint.Blueprint, options distro.ImageOpti
 }
 
 // checkOptions checks the validity and compatibility of options and customizations for the image type.
-func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) error {
+func (t *imageType) checkOptions(customizations *blueprint.Customizations, options distro.ImageOptions, containers []container.Spec) ([]string, error) {
```
thozza, line 319: Please extend the doc text with description of the first return value.

internal/distro/rhel9/imagetype.go
```diff
@@ -314,32 +316,34 @@ func (t *imageType) PackageSets(bp blueprint.Blueprint, options distro.ImageOpti
...
 
+	// holds warnings (e.g. deprecation notices)
+	warnings := []string{}
```
thozza, line 322: I suggest to define `warnings` as
```go
var warnings []string
```
and then just `append()` to it. This way, the function will return `nil` if there are not warnings, instead of an empty slice. So the behavior is now inconsistent with the same same functions defined in distros that do not return any warnings.

internal/distro/rhel8/imagetype.go
```diff
@@ -346,31 +348,33 @@ func overridePackageNames(packages []string) []string {
...
 
+	// holds warnings (e.g. deprecation notices)
+	warnings := []string{}
```
thozza, line 354: I suggest to define `warnings` as
```go
var warnings []string
```
and then just `append()` to it. This way, the function will return `nil` if there are not warnings, instead of an empty slice. So the behavior is now inconsistent with the same same functions defined in distros that do not return any warnings.

thozza (commented):
Thanks for your PR. In general, this looks OK, but I added a few comments inline.

In addition to that, (in the future) I would suggest splitting the change into multiple commits instead of having just a one huge commit. While in this case it is still manageable to review changes, It would be IMO easier if things were done in steps.

internal/distro/rhel9/imagetype.go
7flying: Right, so if there are no warnings, just return nil; but, if there is an error, even if we have warnings, just return nil since the error takes priority?

7flying: Yes, I should have been consistent between fedora+rhel7/8/9, since we want the same thing; so changing it to `var warnings []string` ensures that `nil` when there are no warnings (as in fedora + rhel7) and it initialises the slice with whatever warning there is due to the `append` when we are on rhel8/9; we agree there :+1:

internal/distro/rhel9/imagetype.go
```diff
@@ -214,22 +215,22 @@ func (t *imageType) initializeManifest(bp *blueprint.Blueprint,
 
 	img, err := t.image(w, t, bp.Customizations, options, packageSets, containers, rng)
 	if err != nil {
-		return nil, err
+		return nil, warnings, err
```
thozza, line 218: I'm not sure if I missed this before, but as in other places, I think that we should return `nil` for `warnings` if there is an error. This comment applies to changes done to `initializeManifest()` in all distros.

internal/distro/fedora/distro.go
thozza: In addition, IMO a `nil` should be returned here, instead of the `warnings` if there was an error. This applies to all distros.

internal/cloudapi/v2/server.go
```diff
@@ -458,7 +458,7 @@ func generateManifest(ctx context.Context, workers *worker.Server, depsolveJobID
 		options.OSTree.URL = result.Specs[0].URL
 	}
 
-	manifest, err := imageType.Manifest(b, options, repos, depsolveResults.PackageSpecs, containerSpecs, seed)
+	manifest, _, err := imageType.Manifest(b, options, repos, depsolveResults.PackageSpecs, containerSpecs, seed)
```
thozza, line 461: I just realized if it wouldn't make sense to extend the `JobResult` with a `Warnings` property and return any warnings back.

Specifically in https://github.com/osbuild/osbuild-composer/blob/d7e2e5d7400dad8a88553c6d23582eaa84ce2b27/internal/worker/json.go#L42-L44

Probably a question for @croissanne if we want to eventually return warnings also from the Cloud API and image builder? 🤔 It probably does not make that much sense, since we have more control over the allowed customizations exposed in image-builder, but it may make sense for Koji integration.

thozza (commented):
Thanks for your changes. I added a few comments, but overall this looks pretty good 👍

internal/distro/fedora/distro.go
achilleas-k: > Not related to your changed, but this got me thinking, why are we returning `distro.Manifest{}` instead of `nil` if there was an error... thinking (probably a question for @achilleas-k)

I think in this case they're probably equivalent?  `nil` vs alias to empty byte slice.
I blame @teg :smile:

thozza: Well, not really but almost... `[]byte{} != nil`

internal/distro/rhel9/imagetype.go
7flying: OK, somehow missed it :thinking: , will update in a bit

internal/distro/fedora/distro.go
7flying: Do you want me to do anything about this or are we good?

internal/cloudapi/v2/server.go
achilleas-k: FWIW I think it makes sense.

internal/distro/distro.go
```diff
@@ -118,9 +118,10 @@ type ImageType interface {
 
 	// Returns an osbuild manifest, containing the sources and pipeline necessary
 	// to build an image, given output format with all packages and customizations
-	// specified in the given blueprint. The packageSpecSets must be labelled in
-	// the same way as the originating PackageSets.
-	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, containers []container.Spec, seed int64) (Manifest, error)
+	// specified in the given blueprint; it also returns any warnings (e.g.
+	// deprecation notices) generated by the manifest.
+	// The packageSpecSets must be labelled in the same way as the originating PackageSets.
+	Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, containers []container.Spec, seed int64) (Manifest, []string, error)
```
achilleas-k, line 124: I don't really like returning an extra value for warnings as a slice of strings though I admit that I don't see another straightforward way to do it right now.  Maybe we can have a result object here that encapsulates warning and error, but that's a bigger change that we might want to look at in a followup.

Not asking for a change here, just sharing thoughts for potential future changes.

achilleas-k (approved):
Approving this on my end, though the cloud API question is still open.  My comment on the return values can be ignored for now, but we should consider it later.

internal/cloudapi/v2/server.go
croissanne: I think we can always add it. I'm all for expanding the JobResult base struct to be a bit more verbose :)

internal/distro/fedora/distro.go
thozza: I think that this is something on "us"... so no change needed from your end 😉

internal/distro/distro.go
thozza: I agree. What could help is to use named return variables to make it obvious what's the purpose of the returned value. So something like:

```go
Manifest(b *blueprint.Customizations, options ImageOptions, repos []rpmmd.RepoConfig, packageSpecSets map[string][]rpmmd.PackageSpec, containers []container.Spec, seed int64) (m Manifest, warn []string, err error)
```

internal/distro/rhel9/imagetype.go
```diff
@@ -238,12 +239,13 @@ func (t *imageType) Manifest(customizations *blueprint.Customizations,
 	}
 	bp.Customizations = customizations
 
-	manifest, err := t.initializeManifest(bp, options, repos, nil, containers, seed)
+	manifest, warnings, err := t.initializeManifest(bp, options, repos, nil, containers, seed)
 	if err != nil {
-		return distro.Manifest{}, err
+		return distro.Manifest{}, warnings, err
```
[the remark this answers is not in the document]
7flying: I haven't changed any  `Manifest` functions since we are getting the nil from the `initializeManifest`if there is an error already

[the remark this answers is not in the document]
thozza: @7flying you are right, but the point I'm trying to make (probably not in the best way) is that basically all functions (I'm OK with making the `checkOptions()` an exception) should return `nil` as the value for `warnings` if they are returning an `error`. This applies to all `imagetype.go` instances. I'm sorry for being such a PITA 😇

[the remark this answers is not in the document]
7flying: no worries, I've simply misunderstood. Will update in a moment.

7flying:
Hopefully nils everywhere now :crossed_fingers:

bcl:
I see that these changes are necessary to get the values to percolate back up to the top, but it seems like quite a bit of change. Looking at the number of places where warning is ignored/nil, I wonder if it would be cleaner to switch to a custom error+warning return value? The drawback there being that it would make error checking harder than ` err != nil`. But if it returned an empty `ErrWarn{}` struct instead of nil the check could be something like `ret.IsError()` and `ret.IsWarning()`

7flying:
> I see that these changes are necessary to get the values to percolate back up to the top, but it seems like quite a bit of change. Looking at the number of places where warning is ignored/nil, I wonder if it would be cleaner to switch to a custom error+warning return value? The drawback there being that it would make error checking harder than ` err != nil`. But if it returned an empty `ErrWarn{}` struct instead of nil the check could be something like `ret.IsError()` and `ret.IsWarning()`

@achilleas-k @thozza I'll be happy to rewrite the error handling as Brian suggests if that makes the change a more "permanent" solution. What do you think?

supakeen (commented):
Since all the codechanges have been gone through I won't get too much into that.

I just have a single question: does it make sense to return `nil, nil, err` even in cases where `warnings` *is* available, e.g. when `InstantiateManifest` fails `warnings` are available but not returned.

supakeen:
> > Since all the codechanges have been gone through I won't get too much into that.
> > I just have a single question: does it make sense to return `nil, nil, err` even in cases where `warnings` _is_ available, e.g. when `InstantiateManifest` fails `warnings` are available but not returned.
> 
> The reasoning that we made in all those instances is that when we have an `err`, we only care about returning the error since we are just going to handle that, and not the warnings. Warnings are returned when everything else is OK and we simply need to bring something to the attention of the user.

Makes sense.

7flying:
rebased on main, fixed file conflicts

thozza (approved):
Still looks good

runcom:
> Still looks good

restarted the test failing

thozza:
> rebased on main and fixed conflicts
> 
> @thozza can I ping you once all the tests are green so that you can push the merge button?

Sure. Just ping me once CI is green and I'll review, approve and merge...

7flying:
@thozza this is the

> > rebased on main and fixed conflicts
> > @thozza can I ping you once all the tests are green so that you can push the merge button?
> 
> Sure. Just ping me once CI is green and I'll review, approve and merge...

@thozza this is the all-green ping

thozza (approved):

7flying:
@thozza thanks a lot!

PR #3325, opened 2023-03-08
Test/ostree: add sysroot permission test

yih-redhat:
Update sysroot permission test with new ostree package


test/cases/ostree-ng.sh
```diff
@@ -562,6 +577,21 @@ for LOOP_COUNTER in $(seq 0 30); do
     sleep 10
 done
 
+# Reboot one more time to make /sysroot as RO by new ostree-libs-2022.6-3.el9.x86_64
```
thozza, line 580: The package NVR is specific to el9, but the same code will be executed also on Fedora. So using just the relevant version and not the fill NVR would be probably better.

In addition to that, could you please explain (e.g. in the commit message) why an additional reboot is need at all? What are you trying to achieve / test, which is not possible to do without this reboot?

test/data/repositories/rhel-88.json
```diff
@@ -2,17 +2,17 @@
   "x86_64": [
     {
       "name": "baseos",
-      "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el8/el8-x86_64-baseos-n8.8-20221025/",
+      "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el8/el8-x86_64-baseos-n8.8-20230301/",
```
thozza, line 5: Please update all snapshots in this repo file if you are updating x86 repos. Otherwise we may run into broken dependencies and also would end up testing different things on different architectures.

test/data/repositories/rhel-92.json
```diff
@@ -38,17 +38,17 @@
   "x86_64": [
     {
       "name": "baseos",
-      "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-baseos-n9.2-20230101/",
+      "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-baseos-n9.2-20230301/",
```
thozza, line 41: Please update all snapshots in this repo file if you are updating x86 repos. Otherwise we may run into broken dependencies and also would end up testing different things on different architectures.

test/cases/ostree-ng.sh
yih-redhat: What's NVR? I suppose it is el9.x86_64? I will remove this, and only use ostree-libs-2022.6-3, please let me know if I am wrong.

And for the reboot thing, with new ostree package, the edge vm needs to reboot to make the /sysroot readonly on some OS. The reboot is not required on all OS, for example, rhel9.2. But instead of adding more if/else to check the os, I reboot the vm for all os, it's quick and will keep the test script simple.
Details can be found in https://docs.google.com/document/d/15fgxZGiQUcEVgefb6s3mP3jQaIBwal9KQIuajK_znU4/edit#  if you are interested.

thozza: Sorry, "NVR" stands for "name-version-release" when referring to package names. It is a commonly used abbreviation, so I didn't realize that you may not be familiar with it. Your deduction is correct, that only the "release" part is not relevant for other distributions.

Thanks for the explanation of the reboot. I would appreciate if this info was added to the comment in the script or at least to the commit message (without the link to the internal document). The reason why I'm asking for it is that the reason is no obvious just from reading the test case script.

yih-redhat: Thank you!
Updated the comment.

test/data/repositories/rhel-92.json
yih-redhat: Done.

test/data/ansible/check_ostree.yaml
```diff
@@ -314,10 +294,10 @@
         - name: failed count + 1
           set_fact:
             failed_counter: "{{ failed_counter | int + 1 }}"
-      when: (edge_type == "none") and ((ansible_facts['distribution'] == 'Fedora' and ansible_facts['distribution_version'] is version('37', '<')) or
-            (ansible_facts['distribution'] == 'CentOS') or (ansible_facts['distribution'] == 'RedHat'))
+      when: sysroot_ro == "false"
 
-    - name: /sysroot should be mount with ro permission for edge-simplified-installer and edge-raw-image
+    # https://fedoraproject.org/wiki/Changes/Silverblue_Kinoite_readonly_sysroot
+    - name: /sysroot should be mount with ro permission on RHEL 9.2 , Centos9 and Fedora 37 above
```
thozza, line 300: nitpick: since the task does not execute conditionally based on OS name or release, but only based on `sysroot_ro` variable, it IMO does not make sense to mention RHEL, CS and Fedora in its name at all. This difference is captured by the test case scripts that execute this playbook

thozza (approved):
Thanks for the changes. I added one nitpick, but other than that everything looks good to me.

test/data/ansible/check_ostree.yaml
yih-redhat: Yes, from the test script people can see the difference there. 
I will take care of it in next pr, as it took so loong to pass CI test.

PR #3329, opened 2023-03-10
distro/rhel: add payload repos to os package set

### LINKED ISSUE #3326: Compose with a package from custom repository fails
I tried creating a RHEL-9 AWS image with a package from epel and it failed pretty quickly after starting. The issue seems to be that it fails on gpg check of the custom repository.

Example compose request
```
{
  "distribution": "rhel-91",
  "image_name": "test-image-epel9",
  "image_requests": [
    {
      "architecture": "x86_64",
      "image_type": "aws",
      "upload_request": {
        "type": "aws",
        "options": {
          "share_with_accounts": [
            "123123123123"
          ]
        }
      }
    }
  ],
  "customizations": {
    "packages": [
      "6tunnel"
    ],
    "payload_repositories": [
      {
        "baseurl": "https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/",
        "rhsm": false,
        "gpgkey": "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBGE3mOsBEACsU+XwJWDJVkItBaugXhXIIkb9oe+7aadELuVo0kBmc3HXt/Yp\nCJW9hHEiGZ6z2jwgPqyJjZhCvcAWvgzKcvqE+9i0NItV1rzfxrBe2BtUtZmVcuE6\n2b+SPfxQ2Hr8llaawRjt8BCFX/ZzM4/1Qk+EzlfTcEcpkMf6wdO7kD6ulBk/tbsW\nDHX2lNcxszTf+XP9HXHWJlA2xBfP+Dk4gl4DnO2Y1xR0OSywE/QtvEbN5cY94ieu\nn7CBy29AleMhmbnx9pw3NyxcFIAsEZHJoU4ZW9ulAJ/ogttSyAWeacW7eJGW31/Z\n39cS+I4KXJgeGRI20RmpqfH0tuT+X5Da59YpjYxkbhSK3HYBVnNPhoJFUc2j5iKy\nXLgkapu1xRnEJhw05kr4LCbud0NTvfecqSqa+59kuVc+zWmfTnGTYc0PXZ6Oa3rK\n44UOmE6eAT5zd/ToleDO0VesN+EO7CXfRsm7HWGpABF5wNK3vIEF2uRr2VJMvgqS\n9eNwhJyOzoca4xFSwCkc6dACGGkV+CqhufdFBhmcAsUotSxe3zmrBjqA0B/nxIvH\nDVgOAMnVCe+Lmv8T0mFgqZSJdIUdKjnOLu/GRFhjDKIak4jeMBMTYpVnU+HhMHLq\nuDiZkNEvEEGhBQmZuI8J55F/a6UURnxUwT3piyi3Pmr2IFD7ahBxPzOBCQARAQAB\ntCdGZWRvcmEgKGVwZWw5KSA8ZXBlbEBmZWRvcmFwcm9qZWN0Lm9yZz6JAk4EEwEI\nADgWIQT/itE0RZcQbs6BO5GKOHK/MihGfAUCYTeY6wIbDwULCQgHAgYVCgkICwIE\nFgIDAQIeAQIXgAAKCRCKOHK/MihGfFX/EACBPWv20+ttYu1A5WvtHJPzwbj0U4yF\n3zTQpBglQ2UfkRpYdipTlT3Ih6j5h2VmgRPtINCc/ZE28adrWpBoeFIS2YAKOCLC\nnZYtHl2nCoLq1U7FSttUGsZ/t8uGCBgnugTfnIYcmlP1jKKA6RJAclK89evDQX5n\nR9ZD+Cq3CBMlttvSTCht0qQVlwycedH8iWyYgP/mF0W35BIn7NuuZwWhgR00n/VG\n4nbKPOzTWbsP45awcmivdrS74P6mL84WfkghipdmcoyVb1B8ZP4Y/Ke0RXOnLhNe\nCfrXXvuW+Pvg2RTfwRDtehGQPAgXbmLmz2ZkV69RGIr54HJv84NDbqZovRTMr7gL\n9k3ciCzXCiYQgM8yAyGHV0KEhFSQ1HV7gMnt9UmxbxBE2pGU7vu3CwjYga5DpwU7\nw5wu1TmM5KgZtZvuWOTDnqDLf0cKoIbW8FeeCOn24elcj32bnQDuF9DPey1mqcvT\n/yEo/Ushyz6CVYxN8DGgcy2M9JOsnmjDx02h6qgWGWDuKgb9jZrvRedpAQCeemEd\nfhEs6ihqVxRFl16HxC4EVijybhAL76SsM2nbtIqW1apBQJQpXWtQwwdvgTVpdEtE\nr4ArVJYX5LrswnWEQMOelugUG6S3ZjMfcyOa/O0364iY73vyVgaYK+2XtT2usMux\nVL469Kj5m13T6w==\n=Mjs/\n-----END PGP PUBLIC KEY BLOCK-----",
        "check_gpg": true
      }
    ]
  }
}
```

I did it over the stage web UI but should be the same via API.  I found a traceback in the worker logs:
```
time="2023-03-08T12:00:17Z" level=info msg="    subprocess.CalledProcessError: Command '['rpmkeys', '--root', '/run/osbuild/tree', '--checksig', 'sha256:39a2833701ced4bc227e5753dd88665ffe8e88ca0b12916586b07ee1e983edcc']' returned non-zero exit status 1." jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="    Traceback (most recent call last):" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="    Failed to open file \"/sys/fs/selinux/checkreqprot\": Read-only file system" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="    All rules containing unresolvable specifiers will be skipped." jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="    /usr/lib/tmpfiles.d/journal-nocow.conf:26: Failed to resolve specifier: uninitialized /etc detected, skipping" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="    " jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="      File \"/usr/lib64/python3.9/subprocess.py\", line 528, in run" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.rpm\", line 399, in <module>" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="      File \"/run/osbuild/bin/org.osbuild.rpm\", line 306, in main" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="        subprocess.run([" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="        raise CalledProcessError(retcode, process.args," jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=info msg="        r = main(args[\"tree\"], args[\"inputs\"], args[\"options\"])" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
time="2023-03-08T12:00:17Z" level=error msg="osbuild job failed: osbuild build failed" jobId=acf9ff79-86d2-450a-b005-547fb4a749d6
```
Note:
Originally reported in image-builder repository but Sanne corrected me that this belongs here. Apparently we don't propagate gpg keys to the rpm stage options.

croissanne:
Before instantiating the manifest, any repositories that contain a package set key contained in `PayloadPackageSets()` should be added to the os package set in order to correctly generate the `org.osbuild.rpm` stage for the os pipeline.

Otherwise options like GPG keys are not set correctly.

Fixes #3326


internal/distro/distro_test.go
```diff
@@ -353,7 +383,7 @@ func TestPipelineRepositories(t *testing.T) {
 				"*": {newStringSet([]string{"global-21", "global-22"})},
 				"os": {
 					// chain with payload repo only in the second set for the blueprint package depsolve
-					newStringSet([]string{"os-1", "os-2"}),
+					newStringSet([]string{"os-1", "os-2", "payload"}),
```
thozza, line 386: This is a problem, the payload repo can't appear in the list for the base package set depsolves, because then we would use the user-provided repository when depsolving the base image pkg set.

internal/distro/fedora/distro.go
```diff
@@ -704,12 +705,29 @@ func (t *imageType) initializeManifest(bp *blueprint.Blueprint,
 	/* #nosec G404 */
 	rng := rand.New(source)
 
+	// the os pipeline filters repos based on the `osPkgsKey` package set, merge the repos which
+	// contain a payload package set into the `osPkgsKey`, so those repos are included when
+	// building the rpm stage in the os pipeline
```
thozza, line 710: This is unfortunately not the right approach.

We do not add payload repositories intentionally under the `osPkgsKey` key to ensure that only the "system" repos are used when depsolving the base image pkg set. And only on top of that already depsolved set, we add user-requested packages and also add the payload repos so the list. This assures that the user repo content can't replace any of the base packages, but can be used for any content on top of that.

If the user wants to use their own repos for the base package set, then they need to override system repositories in `/etc/osbuild-composer/repositories`. This is however possible only on-premise.

This change would allow the user-provided payload repos to be used for image base pkg set even in the service. So someone could potentially provide some other EL-clone repos and have the RHEL image be built only using that content (if versions would be higher than what is in RHEL repos, since we do not set hugher priorities for user repos at the moment).

thozza (changes requested):
Thank you for your work on this issue.

Unfortunately, I don't think that this approach is correct.

We should IMO instead change the OS pipeline implementation to take the workload repositories into account when generating RPM stage options.

Specifically here:
https://github.com/osbuild/osbuild-composer/blob/414424cd3058c9601eef771a196e1319ddf0ce16/internal/manifest/os.go#L304
we should combine it with `p.Workload.GetRepos()` if `p.Workload != nil` when calling `osbuild.NewRPMStageOptions()`.

internal/distro/fedora/distro.go
croissanne: This happens after depsolving though. Only when instantiating the manifest the 2nd time do they get added. But I totally agree that if we can solve it in the manifest package that would be much, much cleaner.

internal/distro/distro_test.go
croissanne: ugh you're right, with this approach I should merge them in InstantiateManifest, not initializeManifest.

internal/distro/fedora/distro.go
thozza: Thinking about this a bit more, we should probably make sure that the Workload repos are taken into account for any pipeline implementation from the `internal/manifest` package.

From a brief look:

```bash
thozza@fedora ~/devel/osbuild-composer/internal/manifest (main =)
$ grep -r 'NewRPMStageOptions('
anaconda.go:    pipeline.AddStage(osbuild.NewRPMStage(osbuild.NewRPMStageOptions(p.repos), osbuild.NewRpmStageSourceFilesInputs(p.packageSpecs)))
build.go:       pipeline.AddStage(osbuild.NewRPMStage(osbuild.NewRPMStageOptions(p.repos), osbuild.NewRpmStageSourceFilesInputs(p.packageSpecs)))
commit_server_tree.go:  pipeline.AddStage(osbuild.NewRPMStage(osbuild.NewRPMStageOptions(p.repos), osbuild.NewRpmStageSourceFilesInputs(p.packageSpecs)))
coreos_installer.go:    pipeline.AddStage(osbuild.NewRPMStage(osbuild.NewRPMStageOptions(p.repos), osbuild.NewRpmStageSourceFilesInputs(p.packageSpecs)))
os.go:  rpmOptions := osbuild.NewRPMStageOptions(p.repos)
```

croissanne: As discussed offline, the workload is only properly instantiated during the first `initializeManifest` call from `PackageSets`, as they're based on these pacakge sets. The 2nd invocation doesn't have the needed information to construct the workload the same way. 

I'm not sure what the best approach is, I moved the repo juggling to `Manifest`, instead of `initializeManifest`, which I think is correct, but ugly. Another option is to construct the workloads differently somehow if PackageSets is nil.

achilleas-k (approved):
LGTM

croissanne:
@thozza If you're ok with the current approach, shall we merge?

thozza (approved):
👍

PR #3340, opened 2023-03-20
Change civ repo name

jabia99:
Update CIV repo to use the new one. This is aimed to reduce CI time


jrusz:
I've rebased this, only the ostree tests were failing which should be fixed now.

jrusz (approved):
How does this reduce CI time? Are there less tests in the new container?

PR #3341, opened 2023-03-21
remotefile: Use io.ReadAll not ioutil.ReadAll

bcl:
Just a quick fix of a deprecated ioutil.ReadAll usage that snuck in.


kingsleyzissou (approved):
Oh, my bad! Thanks

PR #3343, opened 2023-03-22
Add tags to CIV instances

jabia99:
To do this, change how we provide parameters to CIV, we now provide a config file stored in gitlab secrets with all this parameters. This way we can also adapt the parameters in the future without modifying the script.


jabia99:
This PR is related to this other [PR-219](https://github.com/osbuild/cloud-image-val/pull/219/files#diff-b1695375e9693a7bf1e1de0e17c2b4d7ce8466d8489c7cd7dc61e2bf79468d04R169). We want to add tags to CIV instances so we can keep track of their cloud costs.

jrusz (approved):

PR #3357, opened 2023-03-28
containers/osbuild-composer: wait for fluentd in entrypoint

diaasami:
to avoid composer crashing and restarting when fluentd is not up

This pull request includes:

### DIFF
--- a/containers/osbuild-composer/entrypoint.py
+++ b/containers/osbuild-composer/entrypoint.py
@@ -20,6 +20,9 @@ import time
 class Cli(contextlib.AbstractContextManager):
     """Command Line Interface"""
 
+    SYSLOG_RETRIES=5
+    SYSLOG_WAIT=2.0
+
     def __init__(self, argv):
         self.args = None
         self._argv = argv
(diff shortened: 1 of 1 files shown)


croissanne (changes requested):
Thank you! 2 nitpicks

croissanne (approved):
Lgtm, thank you!

containers/osbuild-composer/entrypoint.py
```diff
@@ -241,6 +244,20 @@ def _spawn_worker():
             stderr=subprocess.STDOUT,
         )
 
+    @staticmethod
+    def _wait_for_syslog(addr):
+        print("Waiting for fluentd syslog server", file=sys.stderr)
+        for _ in range(Cli.SYSLOG_RETRIES):
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            result = sock.connect_ex(addr)
```
supakeen, line 252: This can raise exceptions. The `_ex` variant only silences exceptions on the `connect(2)` level but not any of the other calls done. Since you're not exiting otherwise they need to be caught (things like `getaddrinfo` failing, etc).

supakeen (changes requested):
I do have some questions:

- What's supposed to happen when the `retries * timeout` is hit and we can't connect to the socket yet?
- Is the syslog server up as soon as it starts accepting connections or do you need to send some data as well?

containers/osbuild-composer/entrypoint.py
diaasami: I switched to socket.connect() so that different error are handled differently and unexpected errors are reported.

supakeen (approved):
Ack.

croissanne (approved):
lgtm!

PR #3359, opened 2023-03-28
packit: build RPMs also for EPEL 9 and RHEL

ondrejbudai:
We got request for these, so let's enable them. I wonder if we should remove epel and leave only rhel in, since we don't technically need epel for anything.


ochosi (approved):
LGTM!
I hadn't noticed your PR and filed [a duplicate](https://github.com/osbuild/osbuild-composer/pull/3364) :laughing:

PR #3362, opened 2023-03-29
Add CI tags to CIV instances

jabia99:
These are runtime variables that we have to set within the test script.

The file is stored in gitlab secrets, we are substituting the placeholders for the actual values.

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -234,6 +234,12 @@ if [ -z "$CIV_CONFIG_FILE" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


jabia99:
It didn't come to mind that files stored in secrets could expand runtime variables, thanks a lot!!

jrusz:
> It didn't come to mind that files stored in secrets could expand runtime variables, thanks a lot!!

I don't know if this works the way you need it though, it was just an idea :sweat_smile:

jabia99:
> I don't know if this works the way you need it though, it was just an idea ￼

Yes! it works as you proposed

PR #3364, opened 2023-03-29
packit: Add EPEL9 to copr builds

ochosi:
Enable EPEL9 as target for COPR builds for PRs and on `main` to remain consistent with osbuild (see https://github.com/osbuild/osbuild/blob/main/.packit.yaml#L31).


ochosi:
Duplicate of https://github.com/osbuild/osbuild-composer/pull/3359

PR #3369, opened 2023-04-03
internal/worker: log dequeue failures

croissanne:


internal/worker/server.go
```diff
@@ -613,6 +613,7 @@ func (s *Server) requestJob(ctx context.Context, arch string, jobTypes []string,
 		jobId, token, depIDs, jobType, args, err = s.jobs.Dequeue(dequeueCtx, jts, channels)
 	}
 	if err != nil {
+		logrus.Errorf("dequeuing job failed: %v", err)
```
ondrejbudai, line 616: I think we want to check for `err` not being a timeout? Otherwise, there will be a ton of entries.

croissanne: good point, fixed

ondrejbudai (approved):

PR #3370, opened 2023-04-03
dbjobqueue: use a fresh context for post-dequeueing operations

ondrejbudai:
This is needed in order to prevent the post-dequeueing operations from failing due to an expired or cancelled context. In order words, when a job is actually dequeued from the database, we need to go all-in.

This pull request includes:

### DIFF
--- a/pkg/jobqueue/dbjobqueue/dbjobqueue.go
+++ b/pkg/jobqueue/dbjobqueue/dbjobqueue.go
@@ -18,9 +18,10 @@ import (
 	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/sirupsen/logrus"
+
 	"github.com/osbuild/osbuild-composer/internal/common/slogger"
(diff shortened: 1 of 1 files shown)


croissanne (approved):

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -337,6 +338,11 @@ func (q *DBJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []
 		}
 	}
 
+	// The job has been dequeued, we are past the point of no return.
```
bcl, line 341: Doesn't this need a timeout? Otherwise it's possible to get stuck here.

ondrejbudai: I usually dislike random timeouts to be honest. I wonder if putting one here is really necessary. It would just fail the HTTP request if the database is extremely slow for some reason. Is it a realistic scenario, though?

ondrejbudai:
Let's implement proper transactions: https://github.com/osbuild/osbuild-composer/pull/3389

PR #3371, opened 2023-04-03
Add support for VMware ovf image type

croissanne:
For reference:

OVF is the open virtualization format, which  I named the image type after. The OVF can either be distributed as the descriptor (the ovf xml file) alongside the disk, or tarred up in an OVA. Which is why I chose to name the platform format after the OVA, and the image type after the OVF. 

If picking a single name makes more sense, I'd go with `ova`.

---

This pull request includes:


internal/image/live.go
```diff
@@ -93,6 +93,13 @@ func (img *LiveImage) InstantiateManifest(m *manifest.Manifest,
 		}
 		artifactPipeline = vmdkPipeline
 		artifact = vmdkPipeline.Export()
+	case platform.FORMAT_OVA:
+		vmdkPipeline := manifest.NewVMDK(m, buildPipeline, imagePipeline)
+		ovfPipeline := manifest.NewOVF(m, buildPipeline, vmdkPipeline)
+		archivePipeline := manifest.NewTar(m, buildPipeline, ovfPipeline, "archive")
```
bcl, line 99: Shouldn't this be `artifactPipeline` and/or set `artifact`? It also looks like GCE has the same error. Or else I am missing how this works :)

internal/upload/vmware/vmware.go
```diff
@@ -17,9 +16,9 @@ type Credentials struct {
 	Datastore  string
 }
 
-// UploadImage is a function that uploads a stream optimized vmdk image to vSphere
+// ImportVmdk is a function that uploads a stream optimized vmdk image to vSphere
```
bcl, line 19: I don't think `Import` is the right thing to call these. To me that implies importing into osbuild-composer, not uploading to an external service.

croissanne: So govc also has datastore.upload, which is why I renamed it to the govc command it calls, which is import.*. That said, happy to rename it to Upload(Ova|Vmdk). Mostly wanted to make the distinction between the two.

croissanne:
It seems like the python version in rhel 8 does not support the `capture_output` option used in `subprocess.run`. This should be fixed in osbuild I suppose.

internal/upload/vmware/vmware.go
croissanne: Reasonable :) I'll change it after it's unblocked.

internal/image/live.go
thozza: It is called `archive`, because the output is an archive (in GCE case at least). Same goes for XZ compressed images. The pipeline name does not matter as long as the image type specifies the same as the name of the pipeline to export. The name has no meaning other than the fact that we need to pass to `osbuild` the pipeline name which result should be exported (technically it could be even more pipeline names than just one, but composer is not able to handle it nor our image definitions).

test/cases/vmware-ovf.sh
```diff
@@ -0,0 +1,238 @@
+#!/bin/bash
```
thozza, line 1: I'm wondering if it would be simple enough just extend the `vmware.sh` script to accept argument whether we want to test VMDK or OVA import. It seems to me that big portion of the script is the same (except for the image type when building the image and the way the VM is created in VSphere)... 🤔  Basically something like we do with `api.sh`.

internal/image/live.go
```diff
@@ -93,6 +93,13 @@ func (img *LiveImage) InstantiateManifest(m *manifest.Manifest,
...
+		archivePipeline.Format = osbuild.TarArchiveFormatOldgnu
```
thozza, line 100: Out of curiosity, is this really required by VSphere to accept the tarball? 🤔 The reason this is used for the GCE image is that GCP really can't import the tarball otherwise.

thozza (commented):
Nice work! I added a few suggestions. Overall, this looks very good 👍

internal/image/live.go
bcl: What I mean is this function returns `archive`, and in these 2 blocks nothing is setting that to anything so it's going to remain set to `nil`. If compression is selected then it operates on `artifactPipeline` which also isn't set in these blocks. So how does this function return anything for these platform cases? It looks to me like it will just return `nil`.

internal/image/live.go
croissanne: Actually the format in the ovf spec is the `ustar` format, but I haven't actually managed to get that working successfully :grimacing:

test/cases/vmware-ovf.sh
croissanne: I merged them, but i wonder if we should also run both the vmdk and the ova test on all RHELs. Since they're not that different.

achilleas-k:
I merged #3446 earlier today and now I feel bad because this PR is a lot older and I created conflicts. :/

internal/image/live.go
achilleas-k: Brian's right: the `artifactPipeline` is defined outside the switch block and is meant to be set based on each case.  It doesn't cause any issues because we never use the return value of the function.  It's meant to define the export pipeline to pass to osbuild but we never wired it up properly and still use the old method where we define the export pipeline name on the image type.

This should be fixed though because we'll probably be wiring it up soon.

achilleas-k (commented):
Partial review with a couple of comments.

achilleas-k (commented):
Partial review with a couple of comments.

internal/upload/vmware/vmware.go
achilleas-k: FWIW, `Import` doesn't sound strange or wrong to me, but I'm fine either way.

internal/upload/vmware/vmware.go
```diff
@@ -14,20 +14,36 @@ type Credentials struct {
...
+func ImportVmdk(creds Credentials, imagePath string) error {
+	args := []string{
+		"import.vmdk",
+	}
+	args = append(args, commonOptions(creds)...)
+	if creds.Folder != "" {
+		args = append(args, fmt.Sprintf("-folder=%s", creds.Folder))
+	}
+
+	args = append(args, imagePath)
```
achilleas-k, line 46: Is this appending the folder option twice?

croissanne: naw, `-folder=%s` is confusingly not a physical folder/path, it's just a logical folder which vsphere classifies the machine under.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 20d98ddc8d04368688a43d1ab4ddb2c3b77f12cd with the main merge-base 468c63d4333812282325f52602dfb505fbed90c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4318374490/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4318374490/artifacts/browse

internal/upload/vmware/vmware.go
achilleas-k: I mean, `commonOptions` does `args = append(args, ... creds.Folder)` and then after that it does the same, no?

thozza (commented):
Besides the duplicated `-folder=` option everything looks good 👍

thozza (approved):
LGTM 👍

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 543047a58e684faee0c89e6624ad1bd86c7208e0 with the main merge-base 468c63d4333812282325f52602dfb505fbed90c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4324019860/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4324019860/artifacts/browse

.gitlab-ci.yml
```diff
@@ -558,14 +558,29 @@ gcp.sh:
...
     - !reference [.nightly_rules_x86_64, rules]
   variables:
-    SCRIPT: vmware.sh
+    SCRIPT: vmware.sh vmdk
+
+vmware.sh_ova:
+  stage: test
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/vmware.sh ova
```
thozza, line 574: You should instead inherit from `.integration_rhel` and define the `SCRIPT` variable, as it is done in the case above. It may be also enough to inherit directly from `vmware.sh_vmdk` job and redefine the `SCRIPT` variable.

croissanne: Yea but I don't want to run it against all runners ._.

thozza: You can still have that with inheritance. Any field that you specify explicitly will be overridden.

croissanne: I'll run it against all rhel rnuners, if we also want this onprem

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 71a101f9946f0f340145a470513d1bc07c671a09 with the main merge-base 468c63d4333812282325f52602dfb505fbed90c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4334320928/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4334320928/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD a3e9c46fbff6c531741cc9b9b96bae99732fff0c with the main merge-base 468c63d4333812282325f52602dfb505fbed90c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4342896598/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4342896598/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d6cfd24d65d16bfc13a9be9c95eef056d1bea1ea with the main merge-base 468c63d4333812282325f52602dfb505fbed90c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4344077554/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4344077554/artifacts/browse

supakeen (approved):

thozza (commented):
FWIW, I reviewed the PR and it looked good 🙄

PR #3378, opened 2023-04-05
test/simplified-installer: fix yq installation on rhel8

runcom:
Basically pulls in https://github.com/kislyuk/yq/issues/165 to fix the installation issue

This pull request includes:

### DIFF
--- a/test/cases/ostree-simplified-installer.sh
+++ b/test/cases/ostree-simplified-installer.sh
@@ -12,7 +12,7 @@ source /usr/libexec/tests/osbuild-composer/shared_lib.sh
 
 # Start firewalld
 sudo systemctl enable --now firewalld
-sudo pip3 install yq
+sudo pip3 install yq==v3.2.1
 # Install fdo packages (This cannot be done in the setup.sh because fdo-admin-cli is not available on fedora)
 sudo dnf install -y fdo-admin-cli
 # Start fdo-aio to have /etc/fdo/aio folder


bcl:
Could you make the commit message a bit more verbose, explain why it needs to be pinned?

runcom:
> Basically pulls in [kislyuk/yq#165](https://github.com/kislyuk/yq/issues/165) to fix the installation issue

forgot to add this to the commit message

runcom:
not sure what's happening with the c9s test (cc @henrywang)

runcom:
this is still needed to avoid breakages in the future by pinning version - notice main PRs are now passing as they're getting the new version from pip

bcl:
> not sure what's happening with the c9s test

There was a problem with gpg2 defaults that was reverted, but doesn't seem to have made it into whatever packit uses to run things. https://bugzilla.redhat.com/show_bug.cgi?id=2070722

yih-redhat:
The failure on cs9 is because the fdouser was not created when ansible try to run check with fdouser, then I logon the target vm, and found that the fdouser is created after a few minutes.

====
Ansible runs at  [2023-04-10T23:01:42-04:00]
foduser was created at Apr 10 23:03:30  (about two minutes later)
====

I'm not sure why it's taking longer than before to create this fdouser on cs9, but I will add a check in script to wait for this user to present.

thozza (approved):

PR #3379, opened 2023-04-05
COMPOSER-1943: Fix failure in cross-distro.sh

atodorov:
This pull request includes:

### DIFF
--- a/test/cases/cross-distro.sh
+++ b/test/cases/cross-distro.sh
@@ -115,7 +115,14 @@ if [ "$ALL_EXPECTED_DISTROS" != "$INSTALLED_DISTROS" ];then
     echo "Some distros are missing!"
     echo "Missing distros:"
     diff <(echo "${ALL_EXPECTED_DISTROS}") <(echo "${INSTALLED_DISTROS}") | grep "<" | sed 's/^<\ //g'
-    exit 1
+
+    # the check above compares repositories/*.json files from git checkout
+    # vs the files installed from an RPM package in order to find files which are
+    # not included in the RPM. Don't fail when running on nightly CI pipeline b/c
+    # very often the repository will be newer than the downstream RPM.
+    if [[ "${CI_PIPELINE_SOURCE:-}" != "schedule" ]]; then
+        exit 1
+    fi
 fi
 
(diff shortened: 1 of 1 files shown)


atodorov:
Link to nightly results https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/835972693

test/cases/cross-distro.sh
```diff
@@ -104,11 +104,26 @@ fi
...
 ALL_DISTROS=$(find "$REPO_PATH" -name '*.json' -printf '%P\n' | awk -F "." '{ print $1 }')
 ALL_EXPECTED_DISTROS=$(echo "$ALL_DISTROS" | grep -E "$PATTERN" | grep -Ev 'beta|stream' | sort)
-ALL_REMAINDERS=$(echo "$ALL_DISTROS" | grep -v "$RECOGNIZED_DISTROS")
+
+# Warning: filter out the remaining distros one by one instead of using grep b/c
+# the distro value rhel-93 will be matched by the grep pattern rhel-9 and will be
+# filtered out. If we're running on a RHEL 9.2 maching osbuild-composer doesn't
+# know anything about 9.3 images so the value rhel-9.3 should be treated as unrecognized
+# and error out as expected in the test snippet further below
+UNRECOGNIZED_DISTROS=""
```
thozza, line 117: The `UNRECOGNIZED_DISTROS` variable is already defined on line no. 87.  I would suggest to prefix this one with `ALL_`, since its use is different.

Also wouldn't it be sufficient to match whole words or lines with `grep` instead? IOW use `$(echo "$ALL_DISTROS" | grep -vw "$RECOGNIZED_DISTROS")`... From what I tested, it seems to produce the expected results.

thozza (commented):
Thank for the fix, I added a few comments.

test/cases/cross-distro.sh
atodorov: > Also wouldn't it be sufficient to match whole words or lines with grep instead?

Probably. I didn't know about this option. Updated the PR accordingly.

thozza (approved):
Looks great. 🎉 I guess you'll test it with the nightly pipeline before merging...

PR #3382, opened 2023-04-12
Test/fix cs9 edge-simplified-installer test failure

yih-redhat:
Test script ostree-simplified-installer.sh failed on cs9, in these two prs.

1. https://github.com/osbuild/osbuild-composer/pull/3378#issuecomment-1502633131
2. https://github.com/osbuild/osbuild-composer/pull/3379

I reproduced this failure in my local environment (it's not 100% reproducible), and found the root cause:
sometimes, when run ansible tasks with fdouser(which is created during fdo onboarding), fdouser is not created yet, caused ansible task failed. In fact, this user is created about two minutes later.

This PR added a step to verify if fdouser exist before run ansible tasks with fdouser.

### DIFF
--- a/test/cases/ostree-simplified-installer.sh
+++ b/test/cases/ostree-simplified-installer.sh
@@ -663,6 +663,22 @@ for _ in $(seq 0 30); do
     sleep 10
 done
 
+# Workaround to fix edge-simplified-installer test failure (ansible runs before fdouser is created)
+# Bug link: https://github.com/osbuild/osbuild-composer/pull/3378#issuecomment-1502633131
+if [[ "${ANSIBLE_USER}" == "fdouser" ]]; then
+    greenprint "🕹 Check if user 'fdouser' exist in edge vm"
+    for _ in $(seq 0 30); do
+        FDOUSER_EXIST=$(ssh "${SSH_OPTIONS[@]}" -i "${SSH_KEY}" admin@$PUB_KEY_GUEST_ADDRESS "grep fdouser /etc/passwd")
+        if [[ ${FDOUSER_EXIST} =~ "fdouser" ]]; then
+            echo "fdouser has been created"
+            break
+        else
+            echo "fdouser has not been created yet"
+            sleep 10
+        fi
+    done
+fi
+
 # Check image installation result
 check_result
 


yih-redhat:
hi @thozza , could you please help to review/approve this pr? I will make a rebase later.

test/cases/ostree-simplified-installer.sh
```diff
@@ -663,6 +663,22 @@ for _ in $(seq 0 30); do
...
+        FDOUSER_EXIST=$(ssh "${SSH_OPTIONS[@]}" -i "${SSH_KEY}" admin@$PUB_KEY_GUEST_ADDRESS "grep fdouser /etc/passwd")
+        if [[ ${FDOUSER_EXIST} =~ "fdouser" ]]; then
+            echo "fdouser has been created"
+            break
+        else
+            echo "fdouser has not been created yet"
+            sleep 10
+        fi
+    done
+fi
```
thozza, line 680: In general this looks good. Nevertheless I think that this would be made generic (since AFAICT, the same user is used also on RHEL-9.2). So there is no need to do this check only on c9s, but instead we should be checking it when `$ANSIBLE_USER` is set to `fdouser` before running Ansible. Since this seems to be a race condition, I think that we may at some point end up hitting this on RHEL as well.

Our of curiosity, isn't there a better way of checking that FDO finished configuring the system? E.g. something like cloud-init has?

thozza (commented):
Thanks a lot for debugging the issue and submitting a fix.

I would like to use the workaround on all distributions (and their version) on which FDO is used. Otherwise this looks good. 👍

test/cases/ostree-simplified-installer.sh
yih-redhat: Thanks for your comments, now it will check when $ANSIBLE_USER is set to fdouser, so this check will run on all OS where fdouser is used.

And for your question "isn't there a better way of checking that FDO finished configuring the system", currently we check it in ansible playbook tasks, and if ansible find this file /etc/device-credentials is being created, which means fdo configuration finished. I will take a look if there is other way to check it.

I have rebased this pr with main and pushed again, could you please change this pr to auto-merge?

thozza: Sure, thanks for the changes and for the info. Approved 👍

PR #3389, opened 2023-04-14
dbjobqueue: Make dequeuing atomic

ondrejbudai:
We found some issues when a job is dequeued in the database (the `sqlDequeue` query succeeds), but the job is never sent to a worker. Since the logging statement at the end of `Dequeue` is never hit (we don't see anything in the logs), it's probably caused by failures of the other SQL queries. There can be multiple reasons for this, the most probable ones are random network issues or a race condition when the context passed into the function hits time-out or is cancelled before all SQL queries finish.

Let's fix this by putting all SQL queries related to one dequeuing operation into a SQL transaction. This should ensure that this operation is atomic. In order words, either all SQL operations succeed, or they are rolled back.

Some code shuffling was needed in order to achieve that, but it's pretty sane.


croissanne (approved):
Fantastic, thank you

PR #3390, opened 2023-04-14
customizations: create all groups unconditionally

ondrejbudai:
We had this weird condition in code that prevented composer to create groups with the same name as a user has. This unfortunately means that you are not able to create a user with a primary group with a certain GID that has the same name as the user. There's the gid field in the user customization, but it requires that the group already exists.

In order to allow that, we need to remove the condition. From now on, it's possible to create groups with the same name as a user has, which can be used to create primary groups with a custom gid.

Note that the lorax compatibility behaviour was actually wrong. When lorax was given a custom gid for a user, it didn't require the gid to exist. When it didn't, the group was just created. Thus, we still don't have full backward compatibility, but at least we now have support for custom gids.

I tested this change locally and I'm now fully able to create an image with a user that has a group with the same name with a custom gid. See `format-request-map.json`.


bcl:
Interesting. In lorax-composer the default behavior of `useradd` was to create a group named for the user, which is why this filter was needed. It looks like osbuild is also using `useradd` but maybe the environment is different, since this behavior depends on how `/etc/login.defs` is setup.

thozza (approved):
LGTM

Issue #3312, opened 2023-02-28
RPi firmware and Uboot missing from esp partition on aarch64 iot disk images


paulwhalen:
**Describe the bug**
The RPi firmware and Uboot need to be copied to the esp partition on aarch64 Fedora IoT disk images in Fedora 38. Currently the files are on the disk image but not yet copied to the correct location.

**Environment**
 * Fedora 38
 * Koji 

**To Reproduce**
Create an aarch64 Fedora IoT disk image in F38. 

**Expected behavior**
Resulting image should be bootable on the RPi family of SBC's

**Additional context**
Snippet from the Fedora IoT kickstart that copies Uboot:

<pre class="syntaxhighlightblock"><code class="hljs bash">
# Setup Raspberry Pi uboot
if [[ $arch == "aarch64" ]]; then
cp -P /usr/share/uboot/rpi_3/u-boot.bin /boot/efi/rpi3-u-boot.bin
cp -P /usr/share/uboot/rpi_4/u-boot.bin /boot/efi/rpi4-u-boot.bin
cp -P /usr/share/uboot/rpi_arm64/u-boot.bin /boot/efi/rpi-u-boot.bin
fi
</code></pre>

The RPi firmware files are on the disk image in '/usr/lib/ostree-boot/efi/'  but should also be copied to '/boot/efi/' 

RPi Firmware packages:
bcm283x-overlays-20230118-1.2578acb.fc38.aarch64
bcm2835-firmware-20230118-1.2578acb.fc38.aarch64
bcm283x-firmware-20230118-1.2578acb.fc38.aarch64
bcm2711-firmware-20230118-1.2578acb.fc38.aarch64

ondrejbudai:
@nullr0ute please clarify whether we can simplify the copying

nullr0ute:
So confirmed we just need to do the following to support all current generations of the RPi on aarch64:

cp -P /usr/share/uboot/rpi_arm64/u-boot.bin /boot/efi/rpi-u-boot.bin

paulwhalen:
The firmware files also need to end up there as well, currently they are installed to '/usr/lib/ostree-boot/efi/'

supakeen:
I'm implementing this at the moment but had a thought (that might be entirely wrong), why don't the RPMs for these files place them in `/boot/efi`? @paulwhalen, @nullr0ute.

PR #3391, opened 2023-04-14
rpi firmware copy

supakeen:
This copies the `u-boot.bin` firmware file into `/boot/efi` when preparing the image.

Raspberry Pi's need that file to be able to boot, copying it through an RPM is cumbersome see #3312.

I've created a new method on the platform that can be used in the `image`-pipelines to place files where they need to be. This is the right place to do it, though it could also be moved to a separate pipeline.

This PR allows Fedora IoT to build bootable (on raspberry pi) images in Koji.


supakeen:
Ok, this got a bit hairy as we can't do `*` copies. This is all the firmware packages from F37. For F38 I'll still need to add the `bcm2711-rpi-cm4-io.dtb`.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c19c92c6048f21f4ff956937a2a031cd501452cc with the main merge-base d94cea91502feaef655daaa2b501decd298cd38b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4117126329/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4117126329/artifacts/browse

internal/distro/fedora/distro.go
```diff
@@ -1025,11 +1025,44 @@ func newDistro(version int) distro.Distro {
 		imageInstallerImgType,
 	)
 	aarch64.addImageTypes(
-		&platform.Aarch64{
+		&platform.Aarch64_IoT{
 			BasePlatform: platform.BasePlatform{
 				ImageFormat: platform.FORMAT_RAW,
 			},
 			UEFIVendor: "fedora",
+			BootFiles: [][2]string{
```
ondrejbudai, line 1033: I think you should be able to remove `[2]string` from all of these lines, which would make the code cleaner.

nullr0ute:
> This copies the `u-boot.bin` firmware file into `/bin` when preparing the image.

Do you mean /boot/efi rather than /bin?

internal/distro/fedora/distro.go
thozza: +1, there should be no need to specify `[2]string` on each line

supakeen:
> > This copies the `u-boot.bin` firmware file into `/bin` when preparing the image.
> 
> Do you mean /boot/efi rather than /bin?

That's indeed what I mean!

internal/osbuild/ostree_input.go
```diff
@@ -20,7 +20,7 @@ type OSTreeCheckoutInput struct {
 	References OSTreeCheckoutReferences `json:"references"`
 }
 
-func (OSTreeCheckoutInput) isStageInput() {}
+func (OSTreeCheckoutInput) isInput() {}
```
thozza, line 23: From what I checked, this interface https://github.com/osbuild/osbuild-composer/blob/a56c70d5053b1c2e5293d68dc3665eefe694829f/internal/osbuild/input.go#L8-L11 is not really used by any code, there are only two instances of its use in the ostree stage / input implementations. This is IMO useless and I propose to instead delete it (probably all of the interface "implementations" as well as the interface). This is most probably a result of the initial inputs implementation when we didn't have everything figured out.

internal/osbuild/tree_input.go
```diff
@@ -19,7 +19,7 @@ func NewTreeInput(reference string) *TreeInput {
 	return input
 }
 
-type PipelineTreeInputs map[string]TreeInput
+type PipelineTreeInputs map[string]Input
```
thozza, line 22: IMO we should not start using the `Input` interface, but should instead converge the tree input implementation to the equivalent of changes done for files input https://github.com/osbuild/osbuild-composer/blob/main/internal/osbuild/files_input.go.

test/data/manifests/fedora_36-aarch64-iot_raw_image-boot.json
```diff
@@ -2389,6 +2396,130 @@
                 {
                   "from": "input://root-tree/",
                   "to": "mount://root/"
+                },
+                {
+                  "from": "input://ostree-tree///usr/lib/ostree-boot/efi/bcm2710-rpi-2-b.dtb",
```
thozza, line 2401: The triple slash `/` does not look correct to be honest 🤔

test/data/manifests/fedora_38-aarch64-iot_raw_image-boot.json
```diff
@@ -2207,6 +2207,13 @@
           {
             "type": "org.osbuild.copy",
             "inputs": {
+              "ostree-tree": {
+                "type": "org.osbuild.ostree.checkout",
+                "origin": "org.osbuild.source",
+                "references": [
+                  ""
```
thozza, line 2214: Shouldn't there be some reference? An empty string does not look correct.

internal/manifest/raw_ostree.go
```diff
@@ -61,7 +63,24 @@ func (p *RawOSTreeImage) serialize() osbuild.Pipeline {
...
+	bootFiles := p.platform.GetBootFiles()
+	if len(bootFiles) > 0 {
+		for _, paths := range bootFiles {
+			copyOptions.Paths = append(copyOptions.Paths, osbuild.CopyStagePath{
+				From: fmt.Sprintf("input://ostree-tree/%s/%s", commitChecksum, paths[0]),
+				To:   fmt.Sprintf("mount://root/%s", paths[1]),
+			})
+		}
+
+		copyInputs["ostree-tree"] = osbuild.NewOSTreeCheckoutInput("org.osbuild.source", commitChecksum)
```
thozza, line 81: The origin is IMO wrong here, since the input origin is a pipeline, not source. We should really add dedicated function to generate the pipeline origin input, such as it is done for the files input (https://github.com/osbuild/osbuild-composer/blob/a56c70d5053b1c2e5293d68dc3665eefe694829f/internal/osbuild/files_input.go#L136) and not pass the origin as an argument.

thozza (changes requested):
Nice work. I have my reservations to changes done to the tree input, since I don't think this goes into the right direction. Also there seem to be bugs in the generated manifests (triple slash, missing reference and also of wrong origin).

internal/distro/fedora/distro.go
supakeen: Done :)

test/data/manifests/fedora_36-aarch64-iot_raw_image-boot.json
supakeen: The triple slash in the test manifests is because 'normally' the commit ID is in between there making the full path `input://ostree-tree/1337abcd1337//usr/lib/ostree-boot/efi/bcm2710-rpi-2-b.dtb` which still has a *double* slash but that isn't the worst?

test/data/manifests/fedora_38-aarch64-iot_raw_image-boot.json
supakeen: This is empty because there is no commit checksum available.

internal/osbuild/ostree_input.go
supakeen: Ack.

test/data/manifests/fedora_36-aarch64-iot_raw_image-boot.json
thozza: Thanks for the additional info. The double slash still looks weird and I would be interested to know where it is coming from. But If this does not cause issues when building images, I could accept it. It would be however good to not have it there if possible.

test/data/manifests/fedora_38-aarch64-iot_raw_image-boot.json
thozza: Right. As we discussed, it would be good to use some placeholder commit for test manifests, because this looked suspicious at the first sight. Thanks for the explanation.

internal/manifest/raw_ostree.go
thozza: We discussed this off-channel and I didn't realize that the copy stage is used in a bit unexpected way to copy files from an ostree commit which is taken as an input for the stage. The reason for this is that the copy stage can't copy files within the tree and using `Stage.MountOSTree()` won't work on pipeline origin input. So while not ideal, this approach is valid.

thozza (approved):
Thanks for all the changes. All looks good. I would prefer for the last two commits to be squashed, but I can live with the current state as well.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD fbaa08d727cd471d58998b60cee6f96a871027d3 with the main merge-base d94cea91502feaef655daaa2b501decd298cd38b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4132913164/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4132913164/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD b7bfd5059bb28ae6b79faa6e6e68d77d8ba4facf with the main merge-base 10c46999b164f88786b167fee09a1c22c3339b46).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4134125940/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4134125940/artifacts/browse

PR #3395, opened 2023-04-18
use full source as image local name

pmtk:
This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3398, opened 2023-04-19
Set container local names explicitly

achilleas-k:
The `name` field in the blueprint customizations for containers was ignored.  We now use it to set the `LocalName` and fall back to the source value when it's not.

This is a behaviour change for both cases:
- When `name` set, the full string is added to the `LocalName` field in the spec.
- When `name` is not set, the full `source` string is added to the `LocalName` field in the spec.  Previously, the `source.Name()` was added instead, which corresponds to the source container name without any tag or digest.

The `LocalName` is then set in the manifest as the `name` option for the input image in the skopeo source.  This is used when importing the image in the skopeo stage: https://github.com/osbuild/osbuild/blob/7cd36f9797a8de740afdd4f9e22cfa004bf04531/stages/org.osbuild.skopeo#L176-L178

As a result, the `names` array in the `images.json` file is populated with the `name` (or `source`) from the blueprint (or compose request), including any tags or digests the user has specified.

In the specific case where a user specifies a `name`, it gives the user the ability to rename an image or add a tag of their own.
In the case where  a user does not specify a `name`, it preserves any tag that was used to define the image.

The most important behaviour change is that local images are no longer tagged as `latest` **unless** the `latest` tag is specified in the `name` (or `source`, if `name` is empty).  Previously, all images were tagged as `latest` in the image, which is the standard tag given when no tag is specified.

Fixes COMPOSER-1952

----

This pull request includes:


thozza:
Didn't review yet, but I wonder if #3395 is attempting to solve a similar problem and should be closed as duplicate? (although there is literally zero information in that PR and commit messages also don't help much)

achilleas-k:
> Didn't review yet, but I wonder if #3395 is attempting to solve a similar problem and should be closed as duplicate? (although there is literally zero information in that PR and commit messages also don't help much)

Ah yes.  I forgot about that one.  I'll close it in favour of this.

thozza (approved):
Amazing work 👍

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD eeff3666e25d87951676e6ef7da31aa897dfe732 with the main merge-base d8342862d5e86a97953cbfc29ee373aca1b5c4a1).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4140093149/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4140093149/artifacts/browse

PR #3399, opened 2023-04-19
Fixes for the koji integration

ondrejbudai:
The first two commits are just small clean-ups for the koji setup in our tests. The other two commits fix rather nasty issues that we are hitting when integrating with koji. I think that the commit messages explain them well, so please, have a look, and feel free to ask any questions. :)

### DIFF
--- a/internal/upload/koji/koji.go
+++ b/internal/upload/koji/koji.go
@@ -3,10 +3,6 @@ package koji
 import (
 	"bytes"
 	"context"
-	"net"
-	"strings"
-	"time"
-
 	// koji uses MD5 hashes
 	/* #nosec G501 */
 	"crypto/md5"
@@ -17,15 +13,18 @@ import (
 	"fmt"
 	"hash/adler32"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
+	"time"
 
+	rh "github.com/hashicorp/go-retryablehttp"
 	"github.com/kolo/xmlrpc"
 	"github.com/sirupsen/logrus"
 	"github.com/ubccr/kerby/khttp"
 
-	rh "github.com/hashicorp/go-retryablehttp"
 	"github.com/osbuild/osbuild-composer/internal/rpmmd"
 )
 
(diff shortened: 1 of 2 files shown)


bcl (approved):
Looks ok to me, given your description of the problem. Hopefully koji is also working on a better solution? If things fail regularly that's quite an increase in upload traffic.

internal/upload/koji/koji.go
```diff
@@ -284,13 +283,32 @@ func (k *Koji) CGImport(build ImageBuild, buildRoots []BuildRoot, images []Image
...
+		var result CGImportResult
+		err = k.xmlrpc.Call("CGImport", []interface{}{string(metadata), directory, token}, &result)
+
+		if err != nil {
+			// Retry when the error mentions a corrupted upload. It's usually
+			// just because of NFS inconsistency when the kojihub has multiple
+			// replicas.
+			if strings.Contains(err.Error(), "Corrupted upload") {
+				time.Sleep(retryDelay)
+				continue
```
thozza, line 299: I'm sure this works, but I'm wondering it it would be worth to use exponential backoff time instead of a constant for each retry? 🤔 It could result in less retries...

IOW, it would be interesting to see how many retries (seconds) it takes on average to make the import successful.

thozza (dismissed):
Great work! 🙇‍♂️  I added one question / suggestion, but all looks good...

ondrejbudai:
> Looks ok to me, given your description of the problem. Hopefully koji is also working on a better solution?

It's a great idea to start this discussion with the fedora infra team and the koji maintainers. I opened the following ticket: https://pagure.io/fedora-infrastructure/issue/11257

internal/upload/koji/koji.go
ondrejbudai: I added a logging statement, is that enough? Exponential back-off seems nice, but tbh, I feel lazy to implement it. What about merging this, seeing how this behaves in the real world and then adjusting it if needed?

thozza (approved):
LGTM

ondrejbudai:
I dropped the commit modifying `gen-certs.sh`. It's not relevant to this PR, it was just an attempt to clean the tests a bit. However, I'm just not able to make the changes working. :/

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 91512fbe99152c86ce73cfff9f124fa1d18c1aa0 with the main merge-base 128f56169efbce95089a412eb52297b1d421a023).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4302261065/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4302261065/artifacts/browse

thozza (approved):
All looks good, thanks 👍

PR #3402, opened 2023-04-21
Adjust packer build to work with the amazon plugin 1.2.3

ondrejbudai:
It turns out that we might have been using the plugin improperly: `associate_public_ip_address` isn't needed if we use the default VPC (which we do). If this option isn't set, we shouldn't hit the new subnet selection algorithm introduced in https://github.com/hashicorp/packer-plugin-amazon/commit/f1ec287c7710e5b6284d6a637ecf092b9e248886 and rely on the AWS's default one, that doesn't ever select a subnet not supporting the instance type we want.

See [packer's docs](https://developer.hashicorp.com/packer/plugins/builders/amazon/ebs#run-configuration) and [this comment](https://github.com/hashicorp/packer-plugin-amazon/issues/368#issuecomment-1516975441).

### DIFF
--- a/templates/packer/config.pkr.hcl
+++ b/templates/packer/config.pkr.hcl
@@ -1,7 +1,7 @@
 packer {
   required_plugins {
     amazon = {
-      version = "= 1.2.2"
+      version = ">= 1.2.3"
       source = "github.com/hashicorp/amazon"
     }
   }
--- a/templates/packer/worker.pkr.hcl
+++ b/templates/packer/worker.pkr.hcl
@@ -18,7 +18,6 @@ source "amazon-ebs" "image_builder" {
   ami_users = "${var.image_users}"
 
   # Network configuration for the instance building our image.
-  associate_public_ip_address = true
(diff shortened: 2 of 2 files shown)


ondrejbudai:
Marking as a draft because I don't want to deploy this to stage on Friday.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ba12dd4ad5e92bd1a79f20ffca1dda9a132b23ba with the main merge-base 2c23894e2a5b3c38f3aec9340f26e15fd1b6cd3d).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4153787644/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4153787644/artifacts/browse

PR #3413, opened 2023-04-27
internal/manifest: remove redundant insights-client call

croissanne:
RHC automatically connects to insights already.

---
Done with 1 rhc connect call:
image

---

marking as draft, as it does seem like it's doing *something*, I asked some other people for help.
image

---

edit2: clarified the above with the insights devs, rhc already does this. So it's not needed.


bcl (approved):
Thanks for confirming that's not needed.

PR #3418, opened 2023-05-02
cloudapi: custom repos add missing fields

kingsleyzissou:
Some fields set in the blueprints and through the service were being ignored. This PR adds the changes to fix this.

- add missing custom repository filename validation
- change repo gpgcheck variable name and `json` serialization field to match image-builder. This ensures that the json tag is the same as the `payload_repository` and makes things more consistent
- set the value for `priority` in cloudapi


achilleas-k (dismissed):
LGTM thanks

achilleas-k:
There's a tricky situation here now.  If we merge this, it'll get released tomorrow, which will mean that we will have an option in osbuild-composer that isn't supported by osbuild yet, and can potentially create manifests that fail the schema validation.  If we release composer as it is now, we will have a blueprint option that's ignored, which isn't as bad.

kingsleyzissou:
> There's a tricky situation here now. If we merge this, it'll get released tomorrow, which will mean that we will have an option in osbuild-composer that isn't supported by osbuild yet, and can potentially create manifests that fail the schema validation. If we release composer as it is now, we will have a blueprint option that's ignored, which isn't as bad.

Oh good point. I could split this PR out and keep the first two commits? Or just leave it for after the release.

achilleas-k:
> > There's a tricky situation here now. If we merge this, it'll get released tomorrow, which will mean that we will have an option in osbuild-composer that isn't supported by osbuild yet, and can potentially create manifests that fail the schema validation. If we release composer as it is now, we will have a blueprint option that's ignored, which isn't as bad.
> 
> Oh good point. I could split this PR out and keep the first two commits? Or just leave it for after the release.

Yeah, splitting would be good to get the validation in and the names fixed.

achilleas-k (approved):
Love it.  Thank you :)

kingsleyzissou:
Rebased because #3420. Hopefully CI will go green now

PR #3425, opened 2023-05-08
Add tests for blueprint package name globs and fix blueprint freeze with globs

bcl:
This pull request includes:

### DIFF
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -966,6 +966,43 @@ func TestMultilibBlueprintDepsolveV0(t *testing.T) {
 	}
 }
 
+// depsolve a blueprint with package name glob
+func TestBlueprintDepsolveGlobsV0(t *testing.T) {
+	// Depends on real packages, only run as an integration test
+	if testState.unitTest {
+		t.Skip()
+	}
+	bp := `{
+		"name": "test-deps-blueprint-globs-v0",
+		"description": "CheckBlueprintDepsolveGlobsV0",
+		"version": "0.0.1",
+		"packages": [{"name": "tmux", "version": "*"},
+		{"name": "openssh-*", "version": "*"}]
+	}`
+
+	// Push a blueprint
+	resp, err := PostJSONBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "POST blueprint failed with a client error")
+	require.True(t, resp.Status, "POST blueprint failed: %#v", resp)
+
+	// Depsolve the blueprint
+	deps, api, err := DepsolveBlueprintV0(testState.socket, "test-deps-blueprint-globs-v0")
+	require.NoError(t, err, "Depsolve blueprint failed with a client error")
+	require.Nil(t, api, "DepsolveBlueprint failed: %#v", api)
+	require.Greater(t, len(deps.Blueprints), 0, "No blueprint dependencies returned")
+	require.Greater(t, len(deps.Blueprints[0].Dependencies), 2, "Not enough dependencies returned")
+
+	// Did it include tmux? Did it include openssh-clients and openssh-server?
+	var names []string
+	for _, d := range deps.Blueprints[0].Dependencies {
+		names = append(names, d.Name)
+	}
+	sort.Strings(names)
+	assert.True(t, common.IsStringInSortedSlice(names, "openssh-clients"))
+	assert.True(t, common.IsStringInSortedSlice(names, "openssh-server"))
+	assert.True(t, common.IsStringInSortedSlice(names, "tmux"))
+}
+
 // depsolve a non-existent blueprint
 func TestNonBlueprintDepsolveV0(t *testing.T) {
 	resp, api, err := DepsolveBlueprintV0(testState.socket, "test-deps-non-blueprint-v0")
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -1101,6 +1101,46 @@ func TestBlueprintFreezeV0(t *testing.T) {
 	require.NotEqual(t, "*", frozen.Blueprints[0].Blueprint.Modules[0].Version, "Wrong version in frozen blueprint module")
 }
 
+func TestBlueprintFreezeGlobsV0(t *testing.T) {
+	// Test needs real packages, skip it for unit testing
+	if testState.unitTest {
+		t.Skip()
+	}
+
+	bp := `{
+		"name": "test-freeze-blueprint-glob-v0",
+		"description": "TestBlueprintFreezeGlobsV0",
+		"version": "0.0.1",
+		"packages": [{"name": "tmux", "version": "?.*"},
+		{"name": "openssh-*", "version": "*"}]
+	}`
+
+	// Push a blueprint
+	resp, err := PostJSONBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "POST blueprint failed with a client error")
+	require.True(t, resp.Status, "POST blueprint failed: %#v", resp)
+
+	// Freeze the blueprint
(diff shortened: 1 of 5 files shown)


bcl:
Documentation in https://github.com/osbuild/guides/pull/113

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD c4e1eb84b47a6970cdc19fdfef9213f2a409ed3a with the main merge-base a48c9d7b00db402d30deb149b3812134dea30602).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4261047737/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4261047737/artifacts/browse

internal/client/blueprints_test.go
```diff
@@ -1101,6 +1101,46 @@ func TestBlueprintFreezeV0(t *testing.T) {
...
+	// Freeze the blueprint
+	frozen, api, err := FreezeBlueprintV0(testState.socket, "test-freeze-blueprint-glob-v0")
+	require.NoError(t, err, "Freeze blueprint failed with a client error")
+	require.Nil(t, api, "FreezeBlueprint failed: %#v", api)
+	require.Greater(t, len(frozen.Blueprints), 0, "No frozen blueprints returned")
+	require.Greater(t, len(frozen.Blueprints[0].Blueprint.Packages), 0, "No frozen packages returned")
+
+	var names []string
+	for _, p := range frozen.Blueprints[0].Blueprint.Packages {
+		// TODO Test versions for * (none of them should have *)
```
ondrejbudai, line 1132: What's the plan with this TODO note?

internal/weldr/api.go
```diff
@@ -1590,6 +1590,29 @@ func setPkgEVRA(dependencies []rpmmd.PackageSpec, packages []blueprint.Package)
...
+				return nil, err
+			}
+			for _, d := range dependencies {
+				if g.Match(d.Name) {
+					newPackages = append(newPackages, blueprint.Package{Name: d.Name, Version: pkg.Version})
+				}
+			}
+		}
+	}
+	return newPackages, nil
```
ondrejbudai, line 1613: Shall we deduplicate the list before returning? I think that more items from `packages` can match one package from `dependencies`?

internal/client/blueprints_test.go
```diff
@@ -1101,6 +1101,46 @@ func TestBlueprintFreezeV0(t *testing.T) {
...
+	frozen, api, err := FreezeBlueprintV0(testState.socket, "test-freeze-blueprint-glob-v0")
+	require.NoError(t, err, "Freeze blueprint failed with a client error")
+	require.Nil(t, api, "FreezeBlueprint failed: %#v", api)
+	require.Greater(t, len(frozen.Blueprints), 0, "No frozen blueprints returned")
+	require.Greater(t, len(frozen.Blueprints[0].Blueprint.Packages), 0, "No frozen packages returned")
+
+	var names []string
+	for _, p := range frozen.Blueprints[0].Blueprint.Packages {
+		assert.NotContains(t, p.Version, "*")
+		assert.NotContains(t, p.Version, "?")
```
thozza, line 1133: Since you are checking specifically for `?`, I think that it would make sense to use this pattern in the package name listed in the testing blueprint.

internal/weldr/api.go
```diff
@@ -1567,27 +1567,54 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
...
+
+	for _, pkg := range packages {
+		if !strings.ContainsAny(pkg.Name, "*?") {
+			// No glob characters, find an exact match in dependencies
+			i := sort.Search(len(dependencies), func(i int) bool {
+				return dependencies[i].Name >= pkg.Name
+			})
+			if i >= len(dependencies) || dependencies[i].Name != pkg.Name {
+				// Packages should not be missing from the depsolve results
+				return nil, fmt.Errorf("%s missing from depsolve results", pkg.Name)
```
thozza, line 1584: No problem here, but this got me thinking if specifying virtual provides as a package name would work when building images, because that could potentially hit this condition (the virtual provides requested will depsolve to a different package name and the original will be missing from the list).

E.g. requesting `MTA` as a package may depsolve to `exim`, `postfix`, `sendmail`, etc... 

```toml
name = "test"
description = ""
version = "0.0.1"
modules = []
groups = []
distro = ""

[[packages]]
name = "MTA"
version = "*"
```
currently works

```toml
name = "test"
description = ""
version = "0.0.0"
modules = []
groups = []
distro = ""

[[packages]]
name = "MTA"
version = "*.*"
```
produces ```ERROR: BlueprintsError: test: DNF error occurred: MarkingErrors: Error occurred when marking packages for installation: Problems in request:
missing packages: MTA-*.*```. I'm not sure if specifying `*.*` version glob makes sense for virtual provides, but I wanted to point out this behavior.

internal/weldr/api.go
```diff
@@ -1567,27 +1567,54 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
...
-			// Packages should not be missing from the depsolve results
-			return fmt.Errorf("%s missing from depsolve results", pkg.Name)
+			// Add all the packages matching the glob
+			g, err := glob.Compile(pkg.Name)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, d := range dependencies {
+				if g.Match(d.Name) {
```
thozza, line 1598: This won't work for virtual provides using globing patterns, but I have no good solution to that 😕

thozza (commented):
Thanks for the change and added tests. The work looks good to me, but there are some corner cases that I would like to discuss before approving.

internal/weldr/api.go
bcl: Yeah, I don't think that makes sense, since it ends up installing multiple packages. We should document it in the blueprint documentation though. There's really no way to figure out the mapping from installed packages back to the virtual provide.

internal/weldr/api.go
bcl: I don't think there's anything that can be done other than clearly document that freeze won't work with virtual provides.

bcl:
Documented virtual provide limitations here - https://github.com/osbuild/guides/pull/116

internal/weldr/api.go
```diff
@@ -1567,27 +1567,54 @@ func (api *API) blueprintsDepsolveHandler(writer http.ResponseWriter, request *h
...
+			}
+			newPackages[dependencies[i].GetNEVRA()] = blueprint.Package{
+				Name:    dependencies[i].Name,
+				Version: dependencies[i].GetEVRA(),
 			}
 		} else {
-			// Packages should not be missing from the depsolve results
-			return fmt.Errorf("%s missing from depsolve results", pkg.Name)
+			// Add all the packages matching the glob
+			g, err := glob.Compile(pkg.Name)
```
thozza, line 1592: Sorry, but reading this the second time, wouldn't this also cover the case when the package name does not contain any of the `*` or `?` characters? In such case the glob pattern should be an exact match, which would cover the case above. I'm OK with the current version, but simplifying the code is always appreciated.

thozza (approved):
I noticed one thing that could be IMO simplified, but in general all looks good 👍

internal/weldr/api.go
thozza: Yeah, I realized that, but not sure if the speed in this case makes much difference to the user. Anyway, it does not matter, all is good 👍

PR #3430, opened 2023-05-10
internal/cloudapi: new prometheus listener

diaasami:
Listening on another port, while keeping the existing endpoint

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -44,7 +44,7 @@ type Composer struct {
 	weldr   *weldr.API
 	api     *cloudapi.Server
 
-	weldrListener, localWorkerListener, workerListener, apiListener net.Listener
+	weldrListener, localWorkerListener, workerListener, apiListener, promListener net.Listener
 }
 
 func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Composer, error) {
@@ -128,6 +128,10 @@ func (c *Composer) InitWeldr(repoPaths []string, weldrListener net.Listener,
 	// Preload the Metadata for all the supported distros
 	c.weldr.PreloadMetadata()
 	return nil
+}
+
+func (c *Composer) InitMetricsAPI(prometheus net.Listener) {
+	c.promListener = prometheus
 }
 
 func (c *Composer) InitAPI(cert, key string, enableTLS bool, enableMTLS bool, enableJWT bool, l net.Listener) error {
@@ -218,7 +222,7 @@ func (c *Composer) Start() error {
 		logrus.Fatal("neither the weldr API socket nor the composer API socket is enabled, osbuild-composer is useless without one of these APIs enabled")
 	}
 
-	var localWorkerAPI, remoteWorkerAPI, composerAPI *http.Server
+	var localWorkerAPI, remoteWorkerAPI, composerAPI, prometheusAPI *http.Server
 
 	if c.localWorkerListener != nil {
 		localWorkerAPI = &http.Server{
@@ -313,6 +317,25 @@ func (c *Composer) Start() error {
 		}()
 	}
 
+	if c.promListener != nil {
+		// metrics listener on another port
+		metricsMux := http.NewServeMux()
+		metricsMux.Handle("/metrics", promhttp.Handler().(http.HandlerFunc))
+
+		prometheusAPI = &http.Server{
+			ErrorLog:          c.logger,
+			Handler:           metricsMux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+
+		go func() {
+			err := prometheusAPI.Serve(c.promListener)
+			if err != nil && err != http.ErrServerClosed {
+				panic(err)
+			}
+		}()
+	}
+
 	if c.weldrListener != nil {
 		go func() {
 			err := c.weldr.Serve(c.weldrListener)
@@ -336,6 +359,13 @@ func (c *Composer) Start() error {
 		// First, close all listeners and then wait for all goroutines to finish.
 		err := composerAPI.Shutdown(context.Background())
 		c.api.Shutdown()
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	if c.promListener != nil {
+		err := prometheusAPI.Shutdown(context.Background())
 		if err != nil {
 			panic(err)
 		}
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -416,6 +416,20 @@ function collectMetrics(){
     echo "$METRICS_OUTPUT" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' '
 }
 
+function testNewMetricsPort(){
+    METRICS_OUTPUT1=$(curl \
+                          --cacert /etc/osbuild-composer/ca-crt.pem \
+                          --key /etc/osbuild-composer/client-key.pem \
+                          --cert /etc/osbuild-composer/client-crt.pem \
+                          https://localhost/metrics)
+    METRICS_OUTPUT2=$(curl http://localhost:8008/metrics)
+
+    COMPOSES1=$(echo "$METRICS_OUTPUT1" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' ')
+    COMPOSES2=$(echo "$METRICS_OUTPUT2" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' ')
+
+    test "$COMPOSES1" = "$COMPOSES2"
+}
+
 function sendCompose() {
     OUTPUT=$(mktemp)
     HTTPSTATUS=$(curl \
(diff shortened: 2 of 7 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 32d555483ce6c39e99b93541e33ab89eb2f0b582 with the main merge-base a48c9d7b00db402d30deb149b3812134dea30602).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4258080702/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4258080702/artifacts/browse

bcl:
`prom listener`? Did you mean `port`? :)

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ca874a30339d88239e9f3051b308a4ff981d48ec with the main merge-base a48c9d7b00db402d30deb149b3812134dea30602).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4266630543/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4266630543/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD bbfea24b142351793fbba6911c81ec68de7d564c with the main merge-base 584ae47eca0b172a7647b70bc1954dc38981a457).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4292793238/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4292793238/artifacts/browse

test/cases/api.sh
```diff
@@ -413,7 +413,15 @@ function collectMetrics(){
...
-    echo "$METRICS_OUTPUT" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' '
+    COMPOSES1=$(echo "$METRICS_OUTPUT" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' ')
+
+    METRICS_OUTPUT2=$(curl https://localhost:9090/metrics)
+
+    COMPOSES2=$(echo "$METRICS_OUTPUT2" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' ')
+
+    test "$COMPOSES1" = "$COMPOSES2"
+
+    echo "$COMPOSES1"
```
croissanne, line 424: These metrics are actually huge, so better not to echo them.

croissanne (dismissed):
lgtm! All that's needed is to update the service monitor's target ports in app-intf, and then drop the old listener.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 7c2148026bde77919e072dddaa7ebeb7e702ca86 with the main merge-base 603c44061c768d178c8e791592843848259c7a56).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4325614363/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4325614363/artifacts/browse

cmd/osbuild-composer/composer.go
```diff
@@ -311,6 +311,24 @@ func (c *Composer) Start() error {
 				panic(err)
 			}
 		}()
+
+		// metrics listener on another port
+		metricsMux := http.NewServeMux()
+		metricsMux.Handle("/metrics", promhttp.Handler().(http.HandlerFunc))
+
+		addr := ":9090"
```
thozza, line 319: Cockpit uses the same port number number, which is documented as "Web-based System Manager".

Moreover, shouldn't this be treated the same way as any other socket that we currently open? IOW use systemd socket unit and socket activation to get a new socket, instead of hardcoding the port and listening on all available interfaces by default...

thozza (changes requested):

cmd/osbuild-composer/composer.go
croissanne: you're right, +1 on a different port.

So this was mostly intended for use in the container, but we might have other uses for the cloudapi in future. So let's do that. We also need to update the container entrypoint in that case though.

thozza: BTW, I suggest to use `8008` port for prometheus socket, since it is listed as another HTTP-alt port in IANA port assignment list.

thozza (commented):
I added a few comments which I would like to discuss. Specifically, I don't think that we need to tie the prometheus socket so closely with cloud API in the composer server implementation. The rest looks good, thanks for all the changes 👍

cmd/osbuild-composer/composer.go
```diff
@@ -311,6 +313,25 @@ func (c *Composer) Start() error {
...
+				ReadHeaderTimeout: 5 * time.Second,
+			}
+
+			go func() {
+				err := promServer.Serve(c.promListener)
+				if err != nil && err != http.ErrServerClosed {
+					panic(err)
+				}
+			}()
+		}
```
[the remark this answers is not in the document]
diaasami: I like the idea of separating this, it's cleaner.

cmd/osbuild-composer/composer.go
```diff
@@ -311,6 +315,25 @@ func (c *Composer) Start() error {
...
```
thozza, line 336: I thought that this block will be moved outside of the `if c.apiListener != nil` block. 🤔

thozza (commented):
Thanks for your changes. I added one comment, the rest looks good. 👍

cmd/osbuild-composer/composer.go
```diff
@@ -218,7 +222,7 @@ func (c *Composer) Start() error {
 		logrus.Fatal("neither the weldr API socket nor the composer API socket is enabled, osbuild-composer is useless without one of these APIs enabled")
 	}
 
-	var localWorkerAPI, remoteWorkerAPI, composerAPI *http.Server
```
thozza, line 221: Nitpick: sorry for not noticing earlier, but it would be nice to be consistent and use `API` suffix in the `*http.Server` variable name, as others do. But I'm OK with keeping it as is.

test/cases/api.sh
```diff
@@ -413,7 +413,15 @@ function collectMetrics(){
                           --cert /etc/osbuild-composer/client-crt.pem \
                           https://localhost/metrics)
 
-    echo "$METRICS_OUTPUT" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' '
+    COMPOSES1=$(echo "$METRICS_OUTPUT" | grep "^image_builder_composer_total_compose_requests" | cut -f2 -d' ')
+
+    METRICS_OUTPUT2=$(curl https://localhost:9090/metrics)
```
thozza, line 418: Shouldn't the URL use `8008` port instead of `9090`? 🤔  I would expect that this test should fail, but it did not from what I remember, which is weird...

thozza (commented):
I planned to approve this PR, but then noticed  that the test still uses `9090` port and it did not fail from what I remember, so keeping the PR unapproved until CI finishes.

test/cases/api.sh
diaasami: Turns out `test` doesn't work when the output of the function is captured in a variable, so I created a new function just for the test.

thozza: Alternatively, you could have just a simple `if` block and `exit 1` in case the condition is true... I wonder if we use `test` in the same way in other places of the test case 🤔

diaasami: It's now failing and for the right reason: `curl: (7) Failed to connect to localhost port 9090: Connection refused`, might be a bit hard to spot though.
In all cases, this is a short lived change and once prometheus is moved to the new port, this new test function will be removed.
I checked all other `*.sh` scripts and we don't have other cases of `test` inside function with the returned value saved in a variable.

thozza (dismissed):
Looks good, thanks! 👍

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 99928acd9b43c92338030de6a9488bf1eed6efe6 with the main merge-base 76b9906332671ceed583ee0c386d07c28183a4c6).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4424329446/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4424329446/artifacts/browse

thozza (approved):

PR #3432, opened 2023-05-10
internal/manifest: install rhc-worker-playbook when using rhc

croissanne:
This isn't pulled in as a dependency, and is required for services lik remediations.

### DIFF
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -212,7 +212,7 @@ func (p *OS) getPackageSetChain() []rpmmd.PackageSet {
 	if p.Subscription != nil {
 		packages = append(packages, "subscription-manager")
 		if p.Subscription.Rhc {
-			packages = append(packages, "rhc", "insights-client")
+			packages = append(packages, "rhc", "insights-client", "rhc-worker-playbook")
 		} else if p.Subscription.Insights {
 			packages = append(packages, "insights-client")
 		}
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -499,12 +499,16 @@ func (p *OS) serialize() osbuild.Pipeline {
 		if p.Subscription.Rhc {
 			// Use rhc for registration instead of subscription manager
 			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
+			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
+			commands = append(commands, "restorecon -R /root/.gnupg")
 		} else {
 			commands = []string{fmt.Sprintf("/usr/sbin/subscription-manager register --org=%s --activationkey=%s --serverurl %s --baseurl %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl, p.Subscription.BaseUrl)}
 
 			// Insights is optional when using subscription-manager
 			if p.Subscription.Insights {
 				commands = append(commands, "/usr/bin/insights-client --register")
+				// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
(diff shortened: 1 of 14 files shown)


bcl (dismissed):
Ack. But it seems like it should be pulled in by rhc or insights eventually, shouldn't it?

croissanne:
> Ack. But it seems like it should be pulled in by rhc or insights eventually, shouldn't it?

Indeed

supakeen (approved):
OK change but not very fond of us having to manage dependencies like these manually. 

This might possibly theoretically cause issues later when we add support for system-installed marking of packages.

ondrejbudai:
> @bcl "Eventually" is the key thing.
> 
> `rhc connect` registers with subscription-manager, registers with insights-client, and connects to MQTT — that's it. `rhc-worker-playbook` might eventually subsequently be installed. But that depends on a lot of other things all working nicely together: the config-manager application, kafka queues, MQTT queues, etc. And even when it's all working nicely, it still takes a minimum of ~60 seconds.

I think @bcl had something different in mind. Shouldn't `rhc` simply require `rhc-worker-playbook` as an RPM dependency? Are there any downsides to this approach?

Ichimonji10:
RHC is useful without rhc-worker-playbook:

1. RHC ships out of the box on RHEL, with the intent of being used as the default registration tool. The goal is for a user to be able to execute `rhc connect` instead of `subscription-manager register && insights-client --register`.
2. RHC and foreman_rh_cloud (?) are used by Satellite 6.11+ (and optionally 6.10, IIRC) to execute playbooks.

For these two use cases respectively:

1. Giving `rhc` a hard dependency on `rhc-worker-playbook` means that _all_ RHEL images, even minimal ones, will include an additional ~17 packages. This is a form of bloat that will make it incrementally harder for us to validate, ship, and support RHEL releases.
2. I expect Satellite system administrators would be very upset to see a tool trying to run [Ansible playbooks](https://github.com/RedHatInsights/config-manager/tree/master/playbooks) on their servers.

croissanne:
Honestly maybe we shouldn't install this by default either then, and just let rhc do it's thing and let it fetch the playbooks itself. But something to be resolved after, there's reasons rn why this has to land sooner rather than later ._.

Ichimonji10:
My general recommendation is that anyone wishing to execute playbooks with RHC explicitly install rhc-worker-playbook. I'm not familiar with osbuild-composer or its requirements, and am happy to hop on a call to see if that general advice fits this specific application and its requirements.

achilleas-k (approved):
Based on the discussion and the current urgency, this LGTM.
We should rethink some of this at a later date.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 64fcbe56700e49241d949c766a5bd578197b3296 with the main merge-base a48c9d7b00db402d30deb149b3812134dea30602).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4270733521/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4270733521/artifacts/browse

teg (approved):

ochosi (approved):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD cbd114457b6af444717b85c78413ff03312c7a42 with the main merge-base a48c9d7b00db402d30deb149b3812134dea30602).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4273066556/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4273066556/artifacts/browse

PR #3440, opened 2023-05-15
COMPOSER-1389: Fedora Workstation Live ISO

supakeen:
Provide a live-installer for Fedora Workstation.


supakeen:
Related: https://github.com/osbuild/osbuild/pull/1311

supakeen:
What do we think about the `Type` I added to `AnacondaInstaller`, is this a good approach or do we hate it?

achilleas-k (commented):
Did a quick read through and it looks good so far.  We talked about the pipeline unification offline and now that I see it it's a bit tricky.  I haven't made up my mind yet if they're different enough to be separated or if there is enough in common to merge and switch on type for the details.  As it is now it seems like they're close enough that they're okay to be merged (there's 2-3 blocks difference in a rather large function), so I'm okay with keeping it as it is here now.

This is currently blocked on https://github.com/osbuild/osbuild/pull/1303 right?

supakeen:
> Did a quick read through and it looks good so far. We talked about the pipeline unification offline and now that I see it it's a bit tricky. I haven't made up my mind yet if they're different enough to be separated or if there is enough in common to merge and switch on type for the details. As it is now it seems like they're close enough that they're okay to be merged (there's 2-3 blocks difference in a rather large function), so I'm okay with keeping it as it is here now.
> 
> This is currently blocked on [osbuild/osbuild#1303](https://github.com/osbuild/osbuild/pull/1303) right?

Yes, it's blocked on [osbuild/osbuild#1303](https://github.com/osbuild/osbuild/pull/1303) :(

internal/manifest/live_installer.go
```diff
@@ -0,0 +1,280 @@
...
+		version:    version,
+	}
+	buildPipeline.addDependent(p)
+	m.addPipeline(p)
+	return p
+}
+
+// TODO: refactor - what is required to boot and what to build, and
+// do they all belong in this pipeline?
+func (p *LiveInstaller) liveInstallerBootPackageSet() []string {
```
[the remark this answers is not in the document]
thozza: +1, this should be ideally defined only on the platform level as it is for other images. It should be possible to extend the platform package set if needed.

Nevertheless, the existing Anaconda installer does the same thing. So if we end up "fixing" this, the other installer should be part of that effort.

internal/manifest/live_installer.go
```diff
@@ -0,0 +1,280 @@
...
+		Password: &rootPassword,
+	}
+
+	usersStageOptions := &osbuild.UsersStageOptions{
+		Users: map[string]osbuild.UsersStageOptionsUser{
+			"root": rootUser,
+		},
+	}
+
+	pipeline.AddStage(osbuild.NewUsersStage(usersStageOptions))
```
thozza: You should ideally use `osbuild.GenUsersStage()` instead.

thozza (commented):
Nice work. I especially like that you merged the anaconda installer pipeline implementations and iso tree pipeline implementations as part of the PR. I wanted to suggest the same thing while going commit by commit only to fund out that you did it :clap: 

I added few comments inline and agree with some comments from Ondrej. I would be OK with some of them being a follow up.

In addition, the commit https://github.com/osbuild/osbuild-composer/pull/3440/commits/d4b1fa70a0f9f3635721fcf8d73e4aa481baba86 deletes some live installer test manifests for no apparent reason :shrug:

internal/distro/fedora/distro.go
```diff
@@ -80,6 +80,24 @@ var (
 		exports:          []string{"bootiso"},
 	}
 
+	// The live installer image type is a live installer for Fedora.
+	liveInstallerImgType = imageType{
+		name:        "live-installer",
+		nameAliases: []string{"fedora-live-installer"},
```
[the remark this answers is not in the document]
supakeen: Right, this is consistent with some of the other names that were there but it isn't in use currently. I'll remove the alias.

internal/manifest/live_installer.go
supakeen: I thought the `osbuild.GenUsersStage` was (mostly) meant to go from the customization format to the stage options so I omitted it here. I'll replace it.

thozza: I think that we don't have strong guidelines around this, but the function is meant for generating stage options from internal abstraction of the user and group definitions (which is agnostic to osbuild). I'm open to discuss what we want to do in general in such cases, I just suggested this change since we were trying to (ideally) not use stage options outside of the osbuild package. :man_shrugging:

supakeen: Then I will use `osbuild.GenUsersStage` :)

thozza: I know :roll_eyes: :innocent:

internal/common/distro.go
```diff
@@ -81,3 +81,33 @@ func VersionLessThan(a, b string) bool {
...
 	return false
 }
+
+// Returns true if the version represented by the first argument is
+// semantically newer than the second.
+// Meant to be used for comparing distro versions for differences between minor
+// releases.
+// Evaluates to false if a and b are equal.
+// Assumes any missing components are 0, so 8 < 8.1.
+func VersionGreaterThan(a, b string) bool {
```
thozza, line 91: I'm fine with adding this variant of the function to compare versions, but you could have just use `!common.VersionLessThan()` as we do in e.g.:
https://github.com/osbuild/osbuild-composer/blob/9ae8632933f8bd0eec362eaa9b37debcdc6211f5/internal/distro/rhel9/edge.go#L408 (also noticed that this example comparing `9-stream` as a version is not what the function can handle, although the outcome is definitely "correct" in this case)

https://github.com/osbuild/osbuild-composer/blob/9ae8632933f8bd0eec362eaa9b37debcdc6211f5/internal/distro/rhel9/images.go#L266

or multiple other places.

internal/distro/fedora/package_sets.go
```diff
@@ -436,6 +436,47 @@ func iotInstallerPackageSet(t *imageType) rpmmd.PackageSet {
 	return ps
 }
 
+func liveInstallerPackageSet(t *imageType) rpmmd.PackageSet {
+	ps := rpmmd.PackageSet{
+		Include: []string{
+			"@workstation-product-environment", // in ks this is @^, what does that mean?
```
thozza, line 442: nitpick: the comment does not provide much value

internal/distro/fedora/package_sets.go
```diff
@@ -436,6 +436,47 @@ func iotInstallerPackageSet(t *imageType) rpmmd.PackageSet {
...
+			"anaconda-install-env-deps",
+			"anaconda-live",
+			"anaconda-dracut",
+			"dracut-live",
+			"glibc-all-langpacks",
+			"kernel",
+			"kernel-modules",
+			"kernel-modules-extra",
+			"livesys-scripts",
+			"rng-tools", // XXX these should be included right!?
```
thozza, line 453: nitpick related to the comment: when we will know the answer? It looks like a WIP comment, which does not provide much value once we merge it, does it?

thozza (approved):
I added a few comments, but the work looks good overall and I'm OK with addressing my comments in a follow up (based on the discussion)

:rocket:

internal/common/distro.go
supakeen: I feel it's a bit harder to determine inclusivity for `!VersionLessThan` vs `VersionGreaterThan`.

internal/distro/fedora/package_sets.go
supakeen: Yea, this question has been answered in the PR (and by @bcl) so the comment goes.

internal/distro/fedora/package_sets.go
supakeen: The comment isn't relevant anymore, removed :)

internal/manifest/live_installer.go
supakeen: Even after saying it twice, the `GenUsersStage` isn't particularly friendly to explicitly empty passwords. This'll be a followup.

internal/common/distro.go
supakeen: After more coffee I don't feel that way and it'll use `!VersionLessThan`.

thozza (dismissed):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD da63883d1b2e055c3b4fe0aa581c3c35d30f7114 with the main merge-base dda50003cbc86930d36e1737acb966b4efd19ef4).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4540991853/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4540991853/artifacts/browse

internal/manifest/live_installer.go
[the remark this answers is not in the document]
supakeen: I'm going to move this to a followup PR to get this MVP in.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 071151707cb81ade6ad148118004678a0b032fd5 with the main merge-base feaa093ef0d2115fb06e2a62b059c87f191c15ff).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4542735499/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4542735499/artifacts/browse

thozza:
You need to generate test manifests for new image types, specifically:
- `missing image test cases for "fedora-37"/"x86_64": [live-installer]`
- `missing image test cases for "fedora-38"/"x86_64": [live-installer]`

supakeen:
> You need to generate test manifests for new image types, specifically:
> 
> * `missing image test cases for "fedora-37"/"x86_64": [live-installer]`
> * `missing image test cases for "fedora-38"/"x86_64": [live-installer]`

Done :)

thozza (approved):
LGTM

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD af8f0b8b4b0d16ab3ad556e34c711650c252244e with the main merge-base 32c3865507a0b0f0f753a7864efdf192b2f2ed9b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4547107483/artifacts/browse) as `manifests.diff`.

**NOTE:** Manifest generation on merge-base with `main` (32c3865507a0b0f0f753a7864efdf192b2f2ed9b) failed.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4547107483/artifacts/browse

PR #3443, opened 2023-05-17
Dockerfile*: update to ubi9 and chown the files when copying

ondrejbudai:
There's no reason not to update to UBI 9.

However, the latest version of UBI 9 needs also one extra patch:

When `go install` is called, go tries to get the git commit hash and embed it
into the built binary. Internally, go just calls the git executable.

The newer go-toolset seems to be based on RHEL 9.2 that ships a newer version
of git (2.39.1). This version contains the safe directory patch that
disallows git from operating on repositories owned by different users.

Thus, we need to chown the files when copying.

See
https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory

### DIFF
--- a/distribution/Dockerfile-fauxauth
+++ b/distribution/Dockerfile-fauxauth
@@ -1,10 +1,10 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:latest AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS builder
 COPY . .
 ENV GOFLAGS=-mod=vendor
 RUN go install ./cmd/osbuild-mock-openid-provider/
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
-RUN microdnf install python3
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+RUN microdnf install -y python3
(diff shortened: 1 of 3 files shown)


croissanne (approved):
thank you!

thozza (approved):
😍

ondrejbudai:
Pushed a fix to the UBI 9 bump - `microdnf install` in EL9 requires `-y` in order to proceed with the transaction. Without it, it just hangs indefinitely.

croissanne (approved):

atodorov (approved):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD cffc42d039b6f05fd61a6380c61c22cf69599736 with the main merge-base 584ae47eca0b172a7647b70bc1954dc38981a457).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4299510727/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4299510727/artifacts/browse

PR #3444, opened 2023-05-17
Finalise interface between composer and image definitions

achilleas-k:
This PR is the biggest step towards moving the image definitions (the current `internal/distro/` package) into a separate repository.  The main goal of this change was to simplify the manifest generation process and especially to avoid initialising the manifest twice (once in `PackageSets()` and once in `Manifest()`).  This lead to a lot of confusion (and ultimately bugs) because it was hard to understand the what exactly was happening in each separate initialisation.

To this end, the following major changes were made:
1. The `PackageSets()` function of the `ImageType` interface is now gone.
2. The `Manifest()` function returns a `manifest.Manifest` struct instead of the serialized manifest (`[]byte`).
3. The `manifest.Manifest` struct holds all unresolved content specs, i.e., package sets, containers, and ostree commits.

This changes the manifest generation process.

**OLD process:**
1. Call `PackageSets()` to determine which package sets are needed for an image type.
2. Resolve containers from blueprint.
3. Resolve package sets.
4. Call `Manifest()` to generate and serialize manifest with resolved packages and containers.

The container resolution was independent and could be done at any time before `Manifest()`.  The `PackageSets()` function would generate a manifest object, get the packages from it, and discard it.  It needed to "fake" certain things like ostree commit references and container specs.

**NEW process:**
1. Call `Manifest()` to determine which package sets are needed and collect container source specs (unresolved container specifications) and return a `manifest.Manifest`.
2. Get the package sets and container source specs from the `manifest` and resolve them separately.
3. Call `manifest.Serialize()` with the resolved content.

### Future changes

To keep this PR from getting even bigger, I decided to hold off some changes for future PRs.
1. ostree commits and file content resolution: These should eventually be resolved the same way as the package sets and containers.  In other words, the `Manifest()` function will create source specs to return with the `Manifest` object and the caller will resolve them and pass them in during `manifest.Serialize()`.
2. Cloud API manifest generation: Manifest generation in the cloud API is still done the old way, but with the `PackageSets()` call replaced by `Manifest()`.  This will require a lot of work to get right and should be done separately.  We're still initialising the manifest twice, but now at least the manifest generation is run the same way both times and should be easier to work with.

--------

Relevant links:
Image Type interface redesign discussion: https://github.com/osbuild/osbuild-composer/discussions/3386
Finalise interface between composer and image definitions: [COMPOSER-1944](https://issues.redhat.com/browse/COMPOSER-1944)
Split out Image Definitions: [COMPOSER-1853](https://issues.redhat.com/browse/COMPOSER-1853)

--------

This pull request includes:


achilleas-k:
Rebased for container fix.

achilleas-k:
Needs a rebase and conflict fix now as well as some small fixes that I see from the test failures.

cmd/gen-manifests/main.go
```diff
@@ -161,7 +161,13 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
...
 		}
 
-		packageSpecs, err := depsolve(cacheDir, imgType, bp, options, repos, distribution, archName)
+		manifest, _, err := imgType.Manifest(&bp, options, repos, nil, containerSpecs, seedArg)
+		if err != nil {
+			err = fmt.Errorf("[%s] failed: %s", filename, err)
+			return
+		}
+
+		packageSpecs, err := depsolve(cacheDir, manifest.Content.PackageSets, distribution, archName)
```
thozza, line 165: Seeing `manifest.Content.PackageSets`, did you consider adding methods to the `Manifest` struct for getting its content? I'm OK with this, but since we tend to use interfaces a lot, I wanted to ask...

internal/distro/rhel9/imagetype.go
```diff
@@ -251,6 +236,9 @@ func (t *imageType) Manifest(bp *blueprint.Blueprint,
 			cw.DisabledServices = services.Disabled
 		}
 		w = cw
+	} else {
+		// workload packages should use the payload repositories
+		w.SetRepos(payloadRepos)
```
thozza, line 237: I'm not completely sure about this 🤔 For internal workloads like SAP or EAP, I would argue that we don't want to use user-specified repos for depsolving the package set. I think that we will still need to handle Workloads in two ways (and even when merging later). One situation is any internal workload that should be treated as the base image (WRT package set and used repos) and another situation is a custom workload based on user input.

I think that if we would apply customizations (read as workloads) in a separate pipeline per workload, that could remove the problem of merging workloads and would make the separation easier (IMO). I'm not suggesting that you should do it now, but I think that we should keep that in mind for the future...

internal/distro/rhel9/imagetype.go
```diff
@@ -257,6 +257,8 @@ func (t *imageType) Manifest(bp *blueprint.Blueprint,
 	}
 
 	manifest.Content.PackageSets = manifest.GetPackageSetChains()
+	manifest.Content.Containers = manifest.GetContainerSourceSpecs()
+	manifest.Content.OSTreeCommits = manifest.GetOSTreeSourceSpecs()
```
thozza, line 262: I'm wondering what's the advantage of having both, the `GetXXXSourceSpecs()` method and exported `manifest.Content.XXX` provided and containing the same value at the same time? 🤔  Shouldn't this be abstracted from the user of `manifest.Manifest` in its implementation?

thozza (commented):
Nice work! I like these changes and the overall direction. I added a few comments and questions inline...

cmd/gen-manifests/main.go
achilleas-k: The funny thing is that these methods exist and the do the collection from the pipelines.  I made the `Content` field public so that it can be set externally for testing, thinking it might be useful for other purposes as well, but it turned out not to be.  I think I'd rather have the content be private and only accessible through the methods, and the methods can be written to do the collection only once to avoid scanning through every time.  I'm trying out something to get around the testing.

I could turn Manifest into an interface as well, but I'll see if I can avoid that.

internal/distro/rhel9/imagetype.go
achilleas-k: > For internal workloads like SAP or EAP, I would argue that we don't want to use user-specified repos for depsolving the package set.

I'm inclined to agree but not sure if we would need to support it eventually, maybe for testing or for internal builds.  I think I'm going to remove it for now and keep support for payload repos only on custom workloads.  We can revisit it later if a requirement comes around.

cmd/gen-manifests/main.go
achilleas-k: So here's where I'm at:
- I don't want to make the `manifest.Manifest` an interface.
- We have the manifest.Pipeline interface, but it has private methods so you can't override them outside the manifest package.
- I'd like to be able to make simple `Manifest` instances that return fixed content and maybe fixed pipelines for testing.
- Figuring out how to test these things helps us figure out how users who import the packages might interact with them.

What I'm thinking now is to keep `Manifest` as a struct, make `AddPipeline()` public, but simplify the `manifest.Pipeline` interface and make some of its methods public, while keeping other private methods implemented in the `Base`.  This way anyone from outside the `manifest` package can implement custom pipelines by embedding the `Base` pipeline.  This is already possible now, but too many of the private methods in `Base` simply return `nil` and aren't useful.

The basic question is: do we want external users writing their own pipelines?  I think we do.
I think the `Manifest` is simple enough that it doesn't need to be an interface.  The collection and serialization methods should have clear semantics (and I believe they already do) so that it's straightforward to implement one or more custom Pipelines to get a desired output.

achilleas-k: Still fiddling with this.  I think I might leave it as is and go deeper in a follow-up.

internal/distro/rhel9/imagetype.go
achilleas-k: I've removed this and adjusted tests accordingly.  The EAP image type, which is currently the only one that specifies a workload, drops any payload repos now.

cmd/gen-manifests/main.go
achilleas-k: I'm going to keep the `Content` field public on `Manifest` for now, solely in service of keeping `test_distro` mostly unchanged.  I think I will immediately follow this PR up with another one that makes the content private and handles the testing differently.

achilleas-k:
Pushed changes that address all the comments.  Going to work on the rebase and conflict resolution now so the two change sets are easier to review.

achilleas-k:

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ee8f41aa9015a6ad09cfabbabd0feea9445aa861 with the main merge-base 92bd58bd764a07d787dcd4332a017c065c0aa6cf).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4352435261/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4352435261/artifacts/browse

croissanne (approved):
yesssssssssss

croissanne (approved):
yessssssssssssssssssssssssssssssssss

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 651dde328b51b2a4a06cb127ae4397b2afb10dd9 with the main merge-base be6119ccfeb20bab06725400f1393377d83ead81).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4353591225/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4353591225/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD e301d9ee9b32ecc8c73b440736cf6061f6da75a9 with the main merge-base be6119ccfeb20bab06725400f1393377d83ead81).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4354567591/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4354567591/artifacts/browse

achilleas-k:
Fixed.  I was missing the containers in `Serialize()` in the cloud API.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 86d2d399844b9de675587d7b61faf4908c455dda with the main merge-base be6119ccfeb20bab06725400f1393377d83ead81).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4357003980/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4357003980/artifacts/browse

supakeen (approved):

cmd/gen-manifests/main.go
thozza: Your plan sounds good to me. 👍

thozza (dismissed):
Great work! 👍

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD f32127560d24024df0b34eddb5e891ce7d495454 with the main merge-base 686b01d0ba56f018f91c46dd881c3eefe66a644c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4376037738/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4376037738/artifacts/browse

achilleas-k:
We're back to the old bugs where things are getting a little messy from having to initialise the manifest twice.  The cloud API was calling `Manifest()` for package selection without an ostree ref, which meant certain packages weren't being selected when they should.

Feels a bit like it was a mistake to leave the cloud API operation re-ordering for a follow-up and have this inbetween state that has all the issues of the old method, but I still think we shouldn't make this PR any longer.

thozza (approved):

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ce10100197b41a6d93ba3fdf41d34925aef8b93e with the main merge-base 686b01d0ba56f018f91c46dd881c3eefe66a644c).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4384273894/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4384273894/artifacts/browse

PR #3446, opened 2023-05-17
Support hybrid boot mode on x86_64 AMI images and set the AMI boot mode on image registration

thozza:
Main changes introduced by this PR:

- `distro.ImageType` interface has been extended with a new `BootMode()` options, which makes it possible to determine for an image type, which boot modes it supports. In specific distro implementations, this is determined based on the `platform.Platform` associated with the specific image type.
- Internal AWS cloud API has been extended to allow specifying the AMI boot mode when registering the image.
- AWS upload target options have been extended to contain the boot mode to be used by worker when registering the AMI.
- Both, the Cloud API and Weldr API have been extended to set the desired image type boot mode (`distro.ImageType.BootMode() return value) as the desired AMI boot type in the AWS upload target options.
- The x86_64 AMI / EC2 images on all distributions have been modified to support hybrid boot mode. As a result, the RHEL-8 x86_64 AMI / EC2 base partition table changes accordingly (separate `/boot`, added `/boot/efi` partition). To make the partitioning unified across RHEL-9 and RHEL-8 and on all supported RHEL-8 architectures, the aarch64 RHEL-8 AMI / EC2 partition table slightly changed as well.
- Relevant test cases have been extended to test newly added functionality.

Side effect cleanups by this PR:

- `s390x` platform implementation now returns sane values from `GetZiplSupport()` method.
- `rhelX` distro implementations structures have been cleaned up from `legacy` and `bootType` members. In addition, the packages needed to boot the image are now added only by the associated platform, not implicitly to the image's base package set.
- The GCE image type has been made UEFI-only, which was the intention since the beginning.

This pull request includes:

  - extended both, the Cloud API (`api.sh`) and Weldr API (`aws.sh`) test cases 


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD f010d4cf9fb65adc0629895e5550ecd6e46f10e6 with the main merge-base 128f56169efbce95089a412eb52297b1d421a023).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4307100098/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4307100098/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 05e359302ff36a40b8c3499e8aee8442c10d57e7 with the main merge-base bd7f0741b2165f524e4808e04b1bfc52a3de058e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4309828291/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4309828291/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d39ce7d35258d84edd479d747076062d99548376 with the main merge-base bd7f0741b2165f524e4808e04b1bfc52a3de058e).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4313429110/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4313429110/artifacts/browse

achilleas-k (approved):
LGTM!

One small comment: Since the boot mode depends strictly on the platform configuration, would it make sense to put the `BootMode()` logic on the `Platform` struct instead?  We could expose it through the `ImageType` as well (as a little passthrough one-line method) so it would work the same as it does now, but the selection logic would be contained in the `BasePlatform`.

This could be done in a followup.

achilleas-k:
Merging since we're all green.  No need to delay this.  My comment can be handled in a followup.

thozza:
> LGTM!
> 
> One small comment: Since the boot mode depends strictly on the platform configuration, would it make sense to put the `BootMode()` logic on the `Platform` struct instead? We could expose it through the `ImageType` as well (as a little passthrough one-line method) so it would work the same as it does now, but the selection logic would be contained in the `BasePlatform`.
> 
> This could be done in a followup.

I was thinking about the same thing, but omitted it from the PR. Since you suggested it, I'll move it in a follow-up 😉 

Originally I also removed the `bootable` from `imageType` implementations and replaced it with use of `BootMode()`, however it turned out that some ostree images are not marked as bootable, since you need to deploy them first to boot them. This is also something to potentially rethink as part of your rework...

achilleas-k:
> Originally I also removed the `bootable` from `imageType` implementations and replaced it with use of `BootMode()`, however it turned out that some ostree images are not marked as bootable, since you need to deploy them first to boot them. This is also something to potentially rethink as part of your rework...

Somewhere in some branch I was working on for the rework I created `Unbootable` platform variants.  That's meant for containers and the tarball and the point is that it shouldn't have anything boot related and also no kernel.  We can also define variants for ostree commits, where they don't define any bootable parameters (uefi vs bios is undefined for example), but still should contain a kernel.

thozza:
> > Originally I also removed the `bootable` from `imageType` implementations and replaced it with use of `BootMode()`, however it turned out that some ostree images are not marked as bootable, since you need to deploy them first to boot them. This is also something to potentially rethink as part of your rework...
> 
> Somewhere in some branch I was working on for the rework I created `Unbootable` platform variants. That's meant for containers and the tarball and the point is that it shouldn't have anything boot related and also no kernel. We can also define variants for ostree commits, where they don't define any bootable parameters (uefi vs bios is undefined for example), but still should contain a kernel.

I agree, but in reality, we mark e.g. `edge-commit` and `edge-container` image types also as NOT bootable, which is half-way true in the context of unbootable platform. Because the image type itself is not bootable on its own, but once deployed it is bootable. So unbootable platform would not make sense in this context. I think that we overloaded the meaning of `bootable` in `ImageType` a bit. If we redefine what it really means, we could definitely replace it.

IMHO, any existing platform which returns empty string from `GetBIOSPlatform()` and `GetUEFIVendor()` and `false` from `GetZiplSupport()` should be already considered not bootable (which is what I used to determine if an image type is "bootable"...

PR #3448, opened 2023-05-19
internal/manifest: execute rhc's post install script

croissanne:
The post install script customises the selinux policy, but is gated on `selinuxenabled`, which will fail inside the buildroot container.

As a result it's never executed.

### DIFF
--- a/internal/manifest/os.go
+++ b/internal/manifest/os.go
@@ -501,6 +501,8 @@ func (p *OS) serialize() osbuild.Pipeline {
 			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
+			// execute the rhc post install script as the selinuxenabled check doesn't work in the buildroot container
+			commands = append(commands, "/usr/sbin/semanage permissive --add rhcd_t")
 		} else {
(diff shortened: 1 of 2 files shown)


internal/manifest/os.go
```diff
@@ -501,6 +501,8 @@ func (p *OS) serialize() osbuild.Pipeline {
 			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
+			// execute the rhc post install script as the selinuxenabled check doesn't work in the buildroot container
+			commands = append(commands, "/usr/sbin/semanage permissive --add rhcd_t || true")
```
ondrejbudai, line 505: Will `|| true` work? Is the command executed in shell?

croissanne: Ugh of course, I'll just remove that part

thozza (dismissed):
LGTM 👍

thozza:
> The post install script customises the selinux policy, but is gated on `selinuxenabled`, which will fail inside the buildroot container.

If possible, we could ensure or "fake" in the buildroot container so that `selinuxenabled` exists with zero exit value. We could do this based on the fact if the image needs to have SELinux enabled or not (now all of them have and should, but who knows what will come in the future). Didn't look into details, so I may be suggesting something that is not viable or possible...

thozza:
There is a unit test failing which needs to be modified:
https://github.com/osbuild/osbuild-composer/blob/39f42699d25d5862851e4073292d911b7042104c/internal/manifest/os_test.go#L101-L116

thozza:
My GitHub UI was probably a bit stuck, because I didn't see the PR to be rebased (to include the 8.7 test workaround), so I clicked in the UI to rebase it. It looks like Sanne already did that before me, so sorry for another rebase 😇

Issue #3097, opened 2022-10-25
CI: upload build logs as artifacts for failed RPM builds


thozza:
It seems that when we build RPMs in CI, we are instructing `mock` not keep any logs from the build. As a result, there is literally no information to help debugging a build failure.

We should be probably setting `cleanup_on_failure` to `False` when running `mock` and upload all build logs as artifacts.

Example of a build failure:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/3223520188

CC @jrusz

PR #3461, opened 2023-05-25
Rpm build log

jabia99:
Fixes: https://github.com/osbuild/osbuild-composer/issues/3097


thozza (approved):
LGTM. It would be nice if you could post an example from your testing with a failed build and uploaded logs.

PR #3476, opened 2023-06-02
Update Cloud API with new manifest generation process

achilleas-k:
This PR is a follow-up to #3444 and completes the redesign of the image definition interface.
There's two major changes here:
- OSTree option and commit resolution handling
- Cloud API manifest generation process

### OSTree options

Until now, the ostree option handling was a bit weird.  We would take what the user provided (ref, url, parent ref), resolve any commit ID we might need, then put it back on the image options.  Also, when an image required an ostree ref to be specified, we read the default one off the image type and put it in the options if it wasn't already set.  I've changed this now so that the ostree options **always** hold just what the user provided (or the test caller).  Using the default ref when an image type needs it is a responsibility of the image type function (each corresponding `imageFunc` implementation).  The options will never be used for resolved information.  So the process now is:
* Call the `Manifest()` function with ostree options exactly as the user specified them (just like we do with the blueprint for example)
* `checkOptions()` returns an error if an ostree URL was required and not specified (for ostree installers and raw images which require an ostree commit payload to be pulled)
* The `Manifest()` function constructs the ostree commit source specification from the ostree options (similar to how it creates container source specifications from the blueprint customizations)
* The image function is called with the commit source specification argument for image types that require it for content (e.g., ostree installers)
* Image function adds the user-specified ref (or default ref if empty) where needed in the manifest when initialising
* The manifest returns an ostree commit source specification for the caller to resolve (like we do with package sets and containers)
* `manifest.Serialize()` takes a resolved ostree commit spec to serialise into the manifest as content (like we do with package specs and resolved container specs)

This unifies the handling of all three content types and makes it easier to separate user options from manifest values.

### Cloud API manifest generation

The manifest initialisation now only happens once in the Cloud API as well.  The manifest is initialised in the composer handler and the content getters (e.g., `GetPackageSetChains()`) are used to retrieve content source specifications and set up the resolve jobs on the worker queue.  The initialised manifest is passed through to the `serializeManifest()` function (which was previously called `generateManifest()`) along with the dependency job IDs.  Once each resolve job is finished, the manifest is serialised with the content specs.

-----

This pull request includes:

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -124,10 +124,10 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 	options := distro.ImageOptions{Size: 0}
 	if cr.OSTree != nil {
 		options.OSTree = &ostree.ImageOptions{
-			URL:           cr.OSTree.URL,
-			ImageRef:      cr.OSTree.Ref,
-			FetchChecksum: cr.OSTree.Parent,
-			RHSM:          cr.OSTree.RHSM,
+			URL:       cr.OSTree.URL,
+			ImageRef:  cr.OSTree.Ref,
+			ParentRef: cr.OSTree.Parent,
+			RHSM:      cr.OSTree.RHSM,
 		}
 	} else {
 		// use default OSTreeRef for image type
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -8,6 +8,7 @@
 package main
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -192,7 +193,7 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 			Blueprint:    cr.Blueprint,
 			OSTree:       cr.OSTree,
 		}
-		err = save(mf, packageSpecs, containerSpecs, request, path, filename)
+		err = save(mf, packageSpecs, containerSpecs, commitSpecs, request, path, filename)
 		return
 	}
 	return job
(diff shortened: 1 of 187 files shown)


achilleas-k:
I just realised a problem I hadn't considered.  With the package selection the way it is now, there's no way to override package names for older distros (RHEL 7 & 8).  The override functions were a bit of a hack, but we still need to figure out how to do package selection per-distro, because package names change and it can't be distro-agnostic.

achilleas-k:
Quick fix, but it's close to something I've been planning for a bit now with distribution-specific package selection.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 63e5cd184d652f0ee6fcefd9c8013ef045057899 with the main merge-base 3faab2f102cff1af67626ba425d0abb44aee0efb).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4413771843/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4413771843/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD a20ca48fe8712d513c61c6adad9ed7db0ea4beba with the main merge-base 3faab2f102cff1af67626ba425d0abb44aee0efb).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4414394358/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4414394358/artifacts/browse

achilleas-k:
Messed up something in the whole ostree commit resolve and ref fallback bits.  I think I might write some unit tests for this, maybe through the APIs as well.

PR #3480, opened 2023-06-05
Cleanup of Fedora cloud images

ondrejbudai:
Our Fedora support isn't currently the greatest, so this is an attempt to improve it. The resulting state after this PR isn't perfect, but it's the step in the right direction in my opinion.

The main change is that I unified `qcow2`, `ami` and `openstack` images. They now share as much code as possible, since this is exactly what Fedora Cloud does: It uses one image for all these 3 platforms.

After the unification, I aligned the package set to look more like the official one. It's not perfect, we still pull 3 extra packages (kernel, kernel-modules, linux-firmware), but it's the best thing we can get with the way we currently assemble manifests. I also raised the image size to 5GiB to match the official image. Btrfs is still not available, but at least the size matches.

I also did some repositories shenaginans:

- I fixed the GPG keys used for F38-39. The F37 key was used for these, causing the builds to fail in some cases.
- I updated the snapshots (we still used branched F37...)

Finally, I adjusted the `aws.sh` test in a way that it no longer uses a custom blueprint. `ami` built by image builder should be usable on AWS without any customizations.

This pull request includes:


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d2a37e192f1103e05222f7a58d0c0393b6790073 with the main merge-base 5e6c0642cda2b772cee98ebf73c9b287f44ca7d4).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4415267117/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4415267117/artifacts/browse

supakeen (approved):
<3

thrix:
Thanks, when will be this deployed? Or is it already the case?

PR #3483, opened 2023-06-06
osbuild: Add validation error logging

bcl:
osbuild can return a json object with details about manifest validation errors. This adds support for saving those, and printing them when the Write function is called. eg. when using composer-cli compose log UUID

Includes tests for new behavior.

This pull request includes:

### DIFF
--- a/internal/osbuild/result.go
+++ b/internal/osbuild/result.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"sort"
+	"strings"
 )
 
 type Result struct {
@@ -13,6 +14,13 @@ type Result struct {
 	Error    json.RawMessage             `json:"error,omitempty"`
 	Log      map[string]PipelineResult   `json:"log"`
 	Metadata map[string]PipelineMetadata `json:"metadata"`
+	Errors   []ValidationError           `json:"errors,omitempty"`
+	Title    string                      `json:"title,omitempty"`
+}
+
+type ValidationError struct {
+	Message string   `json:"message"`
+	Path    []string `json:"path"`
 }
 
 type PipelineResult []StageResult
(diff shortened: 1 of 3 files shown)


achilleas-k (commented):
I did a quick test of this by renaming the json tag for the required option in the selinux stage from `file_contexts` to `file_contextso`, but the parsing in `ValidationError` failed because the `path` array isn't all strings.

This is the output from osbuild:

```json
{
  "type": "https://osbuild.org/validation-error",
  "title": "JSON Schema validation failed",
  "success": false,
  "errors": [
    {
      "message": "'file_contexts' is a required property",
      "path": [
        "pipelines",
        0,
        "stages",
        1,
        "options"
      ]
    },
    {
      "message": "Additional properties are not allowed ('file_contextso' was unexpected)",
      "path": [
        "pipelines",
        0,
        "stages",
        1,
        "options"
      ]
    },
    {
      "message": "'file_contexts' is a required property",
      "path": [
        "pipelines",
        1,
        "stages",
        9,
        "options"
      ]
    },
    {
      "message": "Additional properties are not allowed ('file_contextso' was unexpected)",
      "path": [
        "pipelines",
        1,
        "stages",
        9,
        "options"
      ]
    }
  ]
}
```
The `path` array is a mix of strings and integers, which admittedly looks weird, but I guess makes sense because it's the keys (object keys and array indices) you have to follow to get to the problem.

bcl:
Interesting. I was renaming the curl source url to test. I thought path was the name of the stage, I didn't see any ints at all. I'll take another look at that, thanks. I was seeing `["sources", "org.osbuild.curl"]`

bcl:
Ends up osbuild uses a union for the path elements :/ But at least its either a string or an int, not anything more complex. So this should handle it.

PR #3486, opened 2023-06-09
HMS-1442: handle RPM errors

lavocatt:
As requested by https://issues.redhat.com/browse/HMS-1442 implement a simple routine to extract the RPM errors and forward them to the user.

This only changes the way the JobError is computed, when the OSBuildJorResult is download, so it'll be compatible with any route to expose the information.

An example of API result when RPM is causing troubles:

```
{
  "href": "/api/image-builder-composer/v2/composes/bfc183dd-326d-4cf7-a812-783585a8dd2e",
  "id": "bfc183dd-326d-4cf7-a812-783585a8dd2e",
  "kind": "ComposeStatus",
  "image_status": {
    "error": {
      "details": {
        "id": 10,
        "reason": "rpm stage failed",
        "details": "/usr/lib/tmpfiles.d/abrt.conf:2: Failed to resolve user 'abrt': No such process\n/usr/lib/tmpfiles.d/abrt.conf:9: Failed to resolve user 'abrt': No such process\n/usr/lib/tmpfiles.d/colord.conf:1: Failed to resolve user 'colord': No such process\n/usr/lib/tmpfiles.d/colord.conf:2: Failed to resolve user 'colord': No such process\n/usr/lib/tmpfiles.d/colord.conf:3: Failed to resolve user 'colord': No such process\n/usr/lib/tmpfiles.d/gluster.conf:2: Failed to resolve user 'gluster': No such process\n/usr/lib/tmpfiles.d/journal-nocow.conf:26: Failed to replace specifiers in '/var/log/journal/%m': No such file or directory\n/usr/lib/tmpfiles.d/openvpn.conf:1: Failed to resolve group 'openvpn'.\n/usr/lib/tmpfiles.d/openvpn.conf:2: Failed to resolve group 'openvpn'.\n/usr/lib/tmpfiles.d/rpcbind.conf:2: Failed to resolve user 'rpc': No such process\n/usr/lib/tmpfiles.d/systemd.conf:23: Failed to replace specifiers in '/run/log/journal/%m': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:25: Failed to replace specifiers in '/run/log/journal/%m': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:26: Failed to replace specifiers in '/run/log/journal/%m/*.journal*': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:29: Failed to replace specifiers in '/var/log/journal/%m': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:30: Failed to replace specifiers in '/var/log/journal/%m/system.journal': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:32: Failed to replace specifiers in '/var/log/journal/%m': No such file or directory\n/usr/lib/tmpfiles.d/systemd.conf:33: Failed to replace specifiers in '/var/log/journal/%m/system.journal': No such file or directory\nFailed to open file \"/sys/fs/selinux/checkreqprot\": Read-only file system\nTraceback (most recent call last):\n  File \"/run/osbuild/bin/org.osbuild.rpm\", line 400, in <module>\n    r = main(args[\"tree\"], args[\"inputs\"], args[\"options\"])\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/run/osbuild/bin/org.osbuild.rpm\", line 345, in main\n    raise Exception(\"ahahah I'm crashing here!\")\nException: ahahah I'm crashing here!\n"
      },
      "id": 10,
      "reason": "at least one rpm failed"
    },
    "status": "failure"
  },
  "status": "failure"
}
```

This pull request includes:

### DIFF
--- a/internal/worker/json.go
+++ b/internal/worker/json.go
@@ -76,6 +76,39 @@ func (j *OSBuildJobResult) TargetErrors() []*clienterrors.Error {
 	return targetErrors
 }
 
+// RpmErrors a *clienterrors.Error gathered
+// from the failed rpm stage within the job's error.
+// If there was no RPM error then nil is returned.
+func (j *OSBuildJobResult) RpmError() *clienterrors.Error {
+	type StageError struct {
+		Type   string `json:"type"`
+		Output string `json:"output"`
+	}
+	type OSBuildErrorDetails struct {
+		Stage *StageError `json:"stage,omitempty"`
+	}
+	type OSBuildError struct {
+		Details *OSBuildErrorDetails `json:"details,omitempty"`
+	}
+
+	if j.OSBuildOutput == nil || j.OSBuildOutput.Error == nil {
+		return nil
+	}
+
+	var stageError OSBuildError
+	err := json.Unmarshal(j.OSBuildOutput.Error, &stageError)
+	if err != nil {
+		return nil
+	}
+	if stageError.Details != nil && stageError.Details.Stage != nil && stageError.Details.Stage.Type == "org.osbuild.rpm" {
+		return clienterrors.WorkerClientError(
+			clienterrors.ErrorBuildJob,
+			"rpm stage failed",
+			stageError.Details.Stage.Output)
+	}
+	return nil
+}
+
 // TargetResultsByName iterates over TargetResults attached to the Job result and
 // returns a slice of Target results of the provided name (type). If there were no
 // TargetResults of the desired type attached to the Job results, the returned
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -309,6 +309,12 @@ func (s *Server) OSBuildJobInfo(id uuid.UUID, result *OSBuildJobResult) (*JobInf
 			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorTargetError, "at least one target failed", result.TargetErrors())
 		}
 	}
+	if result.JobError != nil && result.JobError.ID == clienterrors.ErrorBuildJob {
+		rpmError := result.RpmError()
+		if rpmError != nil {
+			result.JobError = rpmError
+		}
+	}
 	// For backwards compatibility: OSBuildJobResult didn't use to have a
 	// top-level `Success` flag. Override it here by looking into the job.
 	if !result.Success && result.OSBuildOutput != nil {
--- a/internal/worker/server_test.go
+++ b/internal/worker/server_test.go
@@ -994,6 +994,44 @@ func TestJobDependencyChainErrors(t *testing.T) {
 				Reason: "empty manifest received",
 			},
 		},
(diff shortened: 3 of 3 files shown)


bcl (approved):
Looks good to me

internal/worker/server.go
```diff
@@ -309,6 +309,12 @@ func (s *Server) OSBuildJobInfo(id uuid.UUID, result *OSBuildJobResult) (*JobInf
 			result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorTargetError, "at least one target failed", result.TargetErrors())
 		}
 	}
+	if result.JobError != nil && result.JobError.ID == clienterrors.ErrorBuildJob {
+		rpmError := result.RpmError()
+		if rpmError != nil {
+			result.JobError = rpmError
+		}
+	}
```
croissanne, line 317: Can we just call the RpmError before setting the `result.JobError` and then set it in the details directly?

above:
```
result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "osbuild build failed", nil)
```

internal/worker/json.go
```diff
@@ -76,6 +76,39 @@ func (j *OSBuildJobResult) TargetErrors() []*clienterrors.Error {
 	return targetErrors
 }
 
+// RpmErrors a *clienterrors.Error gathered
+// from the failed rpm stage within the job's error.
+// If there was no RPM error then nil is returned.
+func (j *OSBuildJobResult) RpmError() *clienterrors.Error {
```
croissanne, line 82: Let's just return a string and embed that in the details. If we just embed a string in the details image builder will automatically forward it. I'm not sure nested error structs work without explicit changes in https://github.com/osbuild/image-builder/blob/main/internal/v1/handler.go#L262

croissanne (changes requested):
Thanks! looks pretty good, i'd simplify the details  a little though.

internal/worker/server.go
lavocatt: In that particular case, the `JobError` is already set in the DB entry, so we have to check for it I think.

croissanne: I mean setting the string directly in details in line 305, right above this block.

lavocatt: I'm sorry, I still don't understand. Line `305` is in a case where the `JobError` from the DB is `nil`. Where in the opposite case. What am I missing ?

internal/worker/json.go
lavocatt: I'm not sure I understand either, what do you mean by nested ? The function `RpmError` returns a single error with for details the output which is a text string. The same way we generate an error on line 305.

PR #3507, opened 2023-06-22
Packit/copr: Remove EPEL builds in favor of RHEL ones

ondrejbudai:
This PR firstly adds aarch64 builds to RHEL 8. After that, our build matrices for EPEL and RHEL buildroots are equal.

The second part of this PR then proceeds by removing the EPEL buildroots. osbuild-composer doesn't (and actually mustn't) depend on EPEL because it's shipped in RHEL and CentOS Stream. Thus, there isn't basically any value in keeping the EPEL builds alive. By removing them, we will save the COPR team some money.

If you happen to be using the EPEL builds today, please switch to the RHEL ones. If you found any issues with them, please report them to us.


atodorov (approved):

henrywang:
This PR should make a rebase to enable RHEL 8.9 and 9.3 RPM building and test. That caused RHEL 9.3 and 8.9 test failed.

thozza (approved):

PR #3514, opened 2023-06-26
Allow any hashing algorithm in osbuild stage inputs

achilleas-k:
Helper functions that create stage input objects with references always hard-coded `sha256:` as a prefix/algorithm for the checksum.  This prevents the functions from being used in cases where other algorithms are use, like sha1, which is possible with (perhaps older) RPM repositories.  The inputs in osbuild a number of hashing algorithms and we should be able to generate stages with other prefixes when necessary.

This PR removes the `sha256:` prefix in the helper functions and assumes all arguments to these functions provide the correct prefix.

Fixes rhbz#2215043 (RHELPLAN-159871).

This pull request includes:

### DIFF
--- a/internal/distro/distro_test.go
+++ b/internal/distro/distro_test.go
@@ -155,8 +155,8 @@ func TestImageTypePipelineNames(t *testing.T) {
 					// manifest creation doesn't fail.
 					allPipelines := append(imageType.BuildPipelines(), imageType.PayloadPipelines()...)
 					minimalPackageSet := []rpmmd.PackageSpec{
-						{Name: "kernel"},
-						{Name: "filesystem"},
+						{Name: "kernel", Checksum: "sha256:a0c936696eb7d5ee3192bf53b9d281cecbb40ca9db520de72cb95817ad92ac72"},
+						{Name: "filesystem", Checksum: "sha256:6b4bf18ba28ccbdd49f2716c9f33c9211155ff703fa6c195c78a07bd160da0eb"},
 					}
 
 					packageSets := make(map[string][]rpmmd.PackageSpec, len(allPipelines))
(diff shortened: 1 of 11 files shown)


thozza (dismissed):
LGTM, great work!

achilleas-k:
I also did a test locally with a custom repo using the `dummy` rpm we use in integration tests and creating a repo with `createrepo -s sha512` (it wouldn't allow me to create a `sha1` repo on Fedora 38).

Before the fix, grepping through the stage inputs of the manifest in the job store:
```
# jq .args.manifest.pipelines[1].stages[1].inputs.packages.references[].id de5fe558-bb03-4aca-9713-e0a5acdc7009.json  | grep sha512
"sha256:sha512:795ddd246d8e65bb65e3340c27c1c90b937397f3610e7f308fddb5bfc5e5b29903a82b85ca0f2be1903dc982b10037e4f657a1f474465766f4cf7f03d1572ce2"
```
and after
```
# jq .args.manifest.pipelines[1].stages[1].inputs.packages.references[].id c68d6939-6f2e-4cca-89b3-aeb49affce2e.json | grep sha512
"sha512:795ddd246d8e65bb65e3340c27c1c90b937397f3610e7f308fddb5bfc5e5b29903a82b85ca0f2be1903dc982b10037e4f657a1f474465766f4cf7f03d1572ce2"
```

supakeen (dismissed):
Looks good, perhaps we want to assert the comment in `NewFilesInputSourceObjectRef`?

achilleas-k:
> Looks good, perhaps we want to assert the comment in `NewFilesInputSourceObjectRef`?

Yeah good idea as a sanity check.

PR #3373, opened 2023-04-04
minimal-raw: zstd compress the raw image

runcom:
Transferring 3.6+ GB over slow networks is just a no go. This brings down the image size to something around 700MB which is sane.

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3518, opened 2023-06-27
Imagetype: minimal-raw support for rhel8/9 with xz compression

sarmahaj:
This pull request includes:

- minimal-raw image type support for rhel8/9
- fedora already had this image type , just added zstd compression stage.
- add compression with newly added zstd stage in osbuild. (https://github.com/osbuild/osbuild/pull/1322)  
   This brings down the image size to something around 700MB.
- updated README.md under image-types/ folder.
- Closed https://github.com/osbuild/osbuild-composer/pull/3373 and updated this PR


teg:
Out of curiosity, how does the compression time and ratio compare to the alternatives?

thozza (commented):
Thanks a lot for this PR. I added some comments inline, but I have also some general comments (and some to just reiterate):

- I think that `minimalRawImage()` function should go away in all distros and you should set the compression to `"zstd"` in the `imageType` definition instead. It will work just fine with the existing `liveImage()` function.
- Please split the PR into smaller self-contained commits, instead of having just one giant commit. Sometimes it is not possible, but in this case it it and it would make it easier to follow each change. E.g.:
  - The addition of support for `zstd` osbuild stage can be a separate commit.
  - The addition of the `Zstd` pipeline implementation can be a separate commit.
  - The addition of `zstd` compression type in `LiveImage` can be a separate commit.
  - Switching Fedora to use the compression + regenerating manifests can be a separate commit.
  - Adding the image type to el8 and el9 can be a separate commit.
- I would like to ask you for additional context and motivation for switching to the compressed file format (in the commit message). My reasoning is that we tend to produce images ideally in ready-to-use format. This change goes to the opposite direction, since the user needs to decompress the image before it can be used. Also this will be a backward-incompatible change in released Fedora version from user's point of view. Having a good justification for this step in the commit message would help (also for the future if we need to ever get back to this change).

Besides these comments, the PR looks really good, great job on that :+1:

sarmahaj:
> Out of curiosity, how does the compression time and ratio compare to the alternatives?
@teg I have ran zstd and xz on my vm so timing might not be correctbut good for comparison.
`
# time zstd -T0 --keep minimal-raw.img 
minimal-raw.img      : 16.92%   (  3.69 GiB =>    639 MiB, minimal-raw.img.zst) 

real	0m8.300s
user	0m12.950s
sys	0m2.535s
 
# time xz -T0 --keep minimal-raw.img

real	8m58.167s
user	17m14.061s
sys	0m3.840s `

and sizes 

` # ls -al
-rw-------. 1 root root  603381024 Jun 27 18:21 minimal-raw.img.xz
-rw-------. 1 root root  669902862 Jun 27 18:21 minimal-raw.img.zst
`
Image size is smaller by around 66MB when xz is used by for compression time zstd is clearly a winner!

sarmahaj:
> Thanks a lot for this PR. I added some comments inline, but I have also some general comments (and some to just reiterate):
> 
> * I think that `minimalRawImage()` function should go away in all distros and you should set the compression to `"zstd"` in the `imageType` definition instead. It will work just fine with the existing `liveImage()` function.
> * Please split the PR into smaller self-contained commits, instead of having just one giant commit. Sometimes it is not possible, but in this case it it and it would make it easier to follow each change. E.g.:
>   
>   * The addition of support for `zstd` osbuild stage can be a separate commit.
>   * The addition of the `Zstd` pipeline implementation can be a separate commit.
>   * The addition of `zstd` compression type in `LiveImage` can be a separate commit.
>   * Switching Fedora to use the compression + regenerating manifests can be a separate commit.
>   * Adding the image type to el8 and el9 can be a separate commit.
> * I would like to ask you for additional context and motivation for switching to the compressed file format (in the commit message). My reasoning is that we tend to produce images ideally in ready-to-use format. This change goes to the opposite direction, since the user needs to decompress the image before it can be used. Also this will be a backward-incompatible change in released Fedora version from user's point of view. Having a good justification for this step in the commit message would help (also for the future if we need to ever get back to this change).
> 
> Besides these comments, the PR looks really good, great job on that 👍

@thozza Thanks for the review :) , will update accordingly 👍

teg:
Thank you @sarmahaj - those are some compelling numbers!

internal/distro/fedora/package_sets.go
```diff
@@ -532,7 +532,7 @@ func containerPackageSet(t *imageType) rpmmd.PackageSet {
 	return ps
 }
 
-func minimalrpmPackageSet(t *imageType) rpmmd.PackageSet {
+func minimalRawImgPackageSet(t *imageType) rpmmd.PackageSet {
```
thozza, line 535: I just noticed that this package set is used also by the image installer image type, so keeping the old name should be better. This seems to be the case only for Fedora. Sorry for not noticing during my initial review :innocent:

internal/distro/fedora/images.go
```diff
@@ -199,6 +199,32 @@ func osCustomizations(
 
 // IMAGES
 
+func minimalRawImage(workload workload.Workload,
```
thozza, line 202: This function should go away, since it is not used (and should not be).

internal/distro/fedora/distro.go
```diff
@@ -305,19 +305,19 @@ var (
...
 		defaultSize:         2 * common.GibiByte,
 		image:               liveImage,
 		buildPipelines:      []string{"build"},
-		payloadPipelines:    []string{"os", "image"},
-		exports:             []string{"image"},
+		payloadPipelines:    []string{"os", "image", "zstd"},
+		exports:             []string{"zstd"},
 		basePartitionTables: defaultBasePartitionTables,
 	}
 )
```
thozza, line 323: You need to specify the `compression` property in the image type.

thozza (commented):
Thanks for your changes. The PR looks great. There are just a few small things that I commented on.

In addition to those, I think that https://github.com/osbuild/osbuild-composer/pull/3518/commits/3b116aebda9fadbeee86a0864ba11a8a1d94791e should be split, specifically the regenerated test manifest. The commit contains modified manifests for el8 and el9 (including centos stream), but those distros have been chnaged only after that commit, so in reality, those manifest changes should be part of it instead.
The same goes to the test case modification, which would IMO fail if run on that commit (because rhel was not changed yet at that point).

Overall, great work, thanks!

sarmahaj:
> Thanks for your changes. The PR looks great. There are just a few small things that I commented on.
> 
> In addition to those, I think that [3b116ae](https://github.com/osbuild/osbuild-composer/commit/3b116aebda9fadbeee86a0864ba11a8a1d94791e) should be split, specifically the regenerated test manifest. The commit contains modified manifests for el8 and el9 (including centos stream), but those distros have been chnaged only after that commit, so in reality, those manifest changes should be part of it instead. The same goes to the test case modification, which would IMO fail if run on that commit (because rhel was not changed yet at that point).
> 
> Overall, great work, thanks!

@thozza Thanks !
I actually have messed up a bit in the files while rebasing , I am working on it(some tests are failing) .
I will split the manifest commit too, and convert the state of the PR to open till CI turns green ! :D

internal/distro/fedora/package_sets.go
sarmahaj: ahh!!  is it okay if I change the name in all the places its been used as a part of this PR ?

thozza: sure. sorry for the noise...

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD ad324c7cb65ca5f0523f18bc9cf3e8ca3ff19b8a with the main merge-base a2c07ea83a2062f929862a5f3a46cf27ad86796b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4589409627/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4589409627/artifacts/browse

supakeen (changes requested):
After going over this it seems OK and what you want to do. However I need to be a bit annoying and request that you remove all changes in `internal/distro/fedora` and only leave the changes for RHEL.

The issue is that the `minimal-raw` image type for Fedora has been in `osbuild-composer` for a few releases by now. Changing it to output files in a different format and with a different name would break anyone that has built tooling over this (I know I personally have some scripts that would break over this that I use to generate my server images).

Aside from this (but this is up to you so feel free to fully ignore this) it'd likely be better if you have `minimal-raw` and `minimal-raw-zstd` in RHEL. Only offering zstd just adds extra actions before the image can be used.

runcom:
> Aside from this (but this is up to you so feel free to fully ignore this) it'd likely be better if you have `minimal-raw` and `minimal-raw-zstd` in RHEL. Only offering zstd just adds extra actions before the image can be used.

not sure here, when playing with ARM machines remotely, transferring 12gb of stuff is pretty bad anyway, and having it spitting out compressed images is a plus. Decompressing zstd is pretty fast anyway too but the gain is on the transfer side of the world. `zstd` came originally from @ondrejbudai too as I wanted to use `xz` for compression so naming it `-zstd` feels weird.



> The issue is that the `minimal-raw` image type for Fedora has been in `osbuild-composer` for a few releases by now. Changing it to output files in a different format and with a different name would break anyone that has built tooling over this (I know I personally have some scripts that would break over this that I use to generate my server images).

this is a good point indeed - maybe @nullr0ute can comment on what we want to do it here in fedora

nullr0ute:
> > Aside from this (but this is up to you so feel free to fully ignore this) it'd likely be better if you have `minimal-raw` and `minimal-raw-zstd` in RHEL. Only offering zstd just adds extra actions before the image can be used.
> 
> not sure here, when playing with ARM machines remotely, transferring 12gb of stuff is pretty bad anyway, and having it spitting out compressed images is a plus. Decompressing zstd is pretty fast anyway too but the gain is on the transfer side of the world. `zstd` came originally from @ondrejbudai too as I wanted to use `xz` for compression so naming it `-zstd` feels weird.

In Fedora we use .xz compression, the concern I have with using .zstd is that unless it yields massive size reduction over .xz I would actually prefer to use .xz because if a customer is writing out the image using el7 or even something like windows I'm not sure how widely supported .zstd is in tools where .xz is pretty widely supported.

> > The issue is that the `minimal-raw` image type for Fedora has been in `osbuild-composer` for a few releases by now. Changing it to output files in a different format and with a different name would break anyone that has built tooling over this (I know I personally have some scripts that would break over this that I use to generate my server images).
> 
> this is a good point indeed - maybe @nullr0ute can comment on what we want to do it here in fedora

So while it's supported Fedora for a few releases it's not actually used in deliverables, we still use ImageFactory there, it outputs .xz files, so unless there's a massive reduction in size from xz -> zstd I would actually prefer to use .xz

sarmahaj:
After discussion around using zstd or xz , we have decided to go for xz over zstd.
Summary : 
- there is no huge difference in compression size ,infact zstd has around 70MB additional size.
- xz is widely deployed and supported across a bunch of platforms whereas zstd is fairly new. So user & compatibility  pov it is better to use xz.
- I will close this PR and create a new one for `minimal-raw` with xz compression support for fedora & rhel8/9.

Thanks @thozza @supakeen for reviewing this PR! 
cc: @ondrejbudai (Thanks for suggesting to use zstd, Its been added to osbuild stages, even though we are not using it for `minimal-raw` I hope it ll be used for for some other purpose !)

supakeen:
There's no need for a new PR is there? It can be changed to use `xz` in this one.

@nullr0ute @runcom it seems we've evaded the question about doing this to Fedora. While it's now been confirmed that *Fedora* itself doesn't use `minimal-raw` images this is still a breaking change to anyone that is currently using `minimal-raw`.

nullr0ute:
> @nullr0ute @runcom it seems we've evaded the question about doing this to Fedora. While it's now been confirmed that _Fedora_ itself doesn't use `minimal-raw` images this is still a breaking change to anyone that is currently using `minimal-raw`.

Personally I would move Fedora to the xz by default as well. I don't think Fedora is currently widely used, but I would like to start to use it in the main compose process and for that it would need to be xz as that's the artifact we currently produce.. I personally don't see this as breaking, but actually fixing a bug/oversight.

supakeen:
> > @nullr0ute @runcom it seems we've evaded the question about doing this to Fedora. While it's now been confirmed that _Fedora_ itself doesn't use `minimal-raw` images this is still a breaking change to anyone that is currently using `minimal-raw`.
> 
> Personally I would move Fedora to the xz by default as well. I don't think Fedora is currently widely used, but I would like to start to use it in the main compose process and for that it would need to be xz as that's the artifact we currently produce.. I personally don't see this as breaking, but actually fixing a bug/oversight.

In that case I'm ok with approving this PR when it's been swapped over to use `xz` :)

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 9f454062f221ae25b21c3a0a122e76215bb115e4 with the main merge-base d7ed3cae6eb8960aa261900768f7cd56d0c74752).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4605145893/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4605145893/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 1fb3a4bc83675e664010960a0eab4d6b38bbb00f with the main merge-base d7ed3cae6eb8960aa261900768f7cd56d0c74752).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4606407401/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4606407401/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 05bfa964730fa2ea2688bcb5001793760847caf1 with the main merge-base d7ed3cae6eb8960aa261900768f7cd56d0c74752).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4611745554/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4611745554/artifacts/browse

henrywang:
@sarmahaj I found an issue on RHEL8.9. The journald doesn't have persistent logging. It only keeps the current one after reboot. But RHEL 9.3 does not have this issue. Could you please take a look? Thanks.

sarmahaj:
> @sarmahaj I found an issue on RHEL8.9 and CS8. The journald doesn't have persistent logging. It only keeps the current one after reboot. But RHEL 9.3 and CS9 does not have this issue. Could you please take a look? Thanks.

@henrywang Thanks for updating the minimal-raw.sh and notifying the issue I will take a look it at!

thozza:
> In that case I'm ok with approving this PR when it's been swapped over to use `xz` :)

I would be still OK with adding the support for zstd compression in pipeline implementations and also adding the support for the zstd osbuild stage. Even if it is not used by any image type. The rest of the discussion seems to be settled down already...

test/data/ansible/check-minimal.yaml
```diff
@@ -4,6 +4,7 @@
   vars:
     total_counter: "0"
     failed_counter: "0"
+    download_node: ""
```
thozza, line 7: nitpick: `hostname` or `url` would IMO make more sense than `node`, but this is not a blocker from my PoV.

thozza (approved):
Nice work! I agree with @supakeen that changing the artifact format for the image type in Fedora is not nice. However, if Fedora IoT is OK with this, I don't consider this to be a blocker.

test/data/ansible/check-minimal.yaml
sarmahaj: cc @henrywang

sarmahaj:
> @sarmahaj I found an issue on RHEL8.9, 9.3, CS8 and CS9. The journald doesn't have persistent logging. It only keeps the current one after reboot. But Fedora 37/38/rawhide does not have this issue. Could you please take a look? Thanks. Following test log from Fedora:
> 
> ```
> TASK [list boots] **************************************************************
> changed: [192.168.100.51] => changed=true 
>   cmd: journalctl --list-boots
>   delta: '0:00:00.013482'
>   end: '2023-07-07 04:28:59.640266'
>   msg: ''
>   rc: 0
>   start: '2023-07-07 04:28:59.626784'
>   stderr: ''
>   stderr_lines: <omitted>
>   stdout: |-
>     IDX BOOT ID                          FIRST ENTRY                 LAST ENTRY
>      -1 8441a01b1ee145789e97622d385b9a49 Fri 2023-07-07 04:24:50 UTC Fri 2023-07-07 04:27:21 UTC
>       0 aa7e0e675bd3438a90e012de315b1cca Fri 2023-07-07 04:28:03 UTC Fri 2023-07-07 04:28:59 UTC
>   stdout_lines: <omitted>
> ```

@henrywang After some discussions with team we found out that persistent logging is disabled in rhel by default https://access.redhat.com/solutions/696893 , and has to be enabled. We tried with method 1 mentioned in this doc and it works. (whereas for fedora is probably enabled thats why persistent logging works on fedora) .
Could you please update the test files for 'minimal-raw' accordingly.

sarmahaj:
This PR is moved to osbuild/images and merged https://github.com/osbuild/images/pull/29
Test are covered in https://github.com/osbuild/osbuild-composer/pull/3565
Closing this PR.

PR #3522, opened 2023-06-27
cloudapi: openscap integration

kingsleyzissou:
Enable the openscap customizations in the cloudapi.


kingsleyzissou:
Not too sure if/how we want to test this in the api integration tests. I could maybe enable like a debug mode that saves the results of the remediation to the image and we could check if that file was created in the tests? Wdyt?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -863,6 +863,16 @@ components:
           type: boolean
         priority:
           type: integer
+    OpenSCAP:
+      type: object
+      required:
+        - datastream
+        - profile_id
+      properties:
```
teg, line 871: Should this be hard-coded per-distro, and not exposed in the public API?

kingsleyzissou: Yeah good point, I'm just confirming with the openscap team that there aren't any cases where there is more than one datastream per distro

kingsleyzissou: Cool, going to go ahead and hardcode these now.

internal/cloudapi/v2/handler.go
```diff
@@ -329,6 +329,22 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
+					ProfileID:  request.Customizations.Openscap.ProfileId,
+				},
+			}
+		} else {
+			bp.Customizations.OpenSCAP = &blueprint.OpenSCAPCustomization{
+				DataStream: request.Customizations.Openscap.Datastream,
+				ProfileID:  request.Customizations.Openscap.ProfileId,
+			}
+		}
+	}
```
thozza: nitpick: I like the way used by other customizations above nicer. IOW, define the OSCAP customization only once and then use it in the if-else block.

internal/cloudapi/v2/openapi.v2.yml
thozza: I'm not that familiar with OSCAP profiles, but does hard-coding means that the caller would not have an option to specify different profiles? Not sure what is the intended workflow, thus asking... :innocent:

thozza: Yeah, that makes a lot of sense :+1:

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 72b0e5d07b1098cb3ca2b56971ad7d6aa651e8e1 with the main merge-base 36158ab84331cb577c77738b9c214ab28f4eabc9).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4553337180/artifacts/browse) as `manifests.diff`.

**NOTE:** Manifest generation on merge-base with `main` (36158ab84331cb577c77738b9c214ab28f4eabc9) failed.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4553337180/artifacts/browse

internal/distro/rhel8/imagetype.go
```diff
@@ -374,7 +374,7 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 
 	if osc := customizations.GetOpenSCAP(); osc != nil {
 		// only add support for RHEL 8.7 and above. centos not supported.
-		if !t.arch.distro.isRHEL() || common.VersionLessThan(t.arch.distro.osVersion, "8.7") {
+		if common.VersionLessThan(t.arch.distro.osVersion, "8.7") {
```
thozza, line 382: You should also adjust the comment above the line, since it says `centos not supported`

internal/blueprint/customizations.go
```diff
@@ -107,8 +107,7 @@ type ServicesCustomization struct {
 }
 
 type OpenSCAPCustomization struct {
-	DataStream string `json:"datastream,omitempty" toml:"datastream,omitempty"`
```
thozza, line 110: The official RH documentation refers to the `datastream` option. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/composing_a_customized_rhel_system_image/assembly_creating-pre-hardened-images-with-image-builder-openscap-integration_composing-a-customized-rhel-system-imag

While I agree with hardcoding the path, it would be IMO better for backward compatibility to still keep the option and use the hardcoded path if none was provided by the user. Removing it at this point is IMO too late. We can still decide to not expose it in the Cloud API or even limit it only on the image builder API level...

internal/distro/fedora/distro.go
```diff
@@ -32,6 +32,10 @@ const (
 
 	//Kernel options for ami, qcow2, openstack, vhd and vmdk types
 	defaultKernelOptions = "ro no_timer_check console=ttyS0,115200n8 biosdevname=0 net.ifnames=0"
+
+	// hardcode the datastream since this is provided by the scap-security-guide rpm
+	// and is not expected to change.
+	datastream = "/usr/share/xml/scap/ssg/contentssg-fedora-ds.xml"
```
thozza, line 38: nitpick: I think that it would be good to prefix the variable name with `oscap` or something like that. E.g. `defaultOSCAPDatastream`...

internal/distro/fedora/images.go
```diff
@@ -123,7 +123,7 @@ func osCustomizations(
 		}
 		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(
 			osbuild.OscapConfig{
-				Datastream: oscapConfig.DataStream,
+				Datastream: datastream,
```
thozza, line 126: here, you could set it to default only if the user didn't provide any value...

internal/distro/rhel7/images.go
```diff
@@ -129,15 +129,6 @@ func osCustomizations(
...
 
-	if oscapConfig := c.GetOpenSCAP(); oscapConfig != nil {
-		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(
-			osbuild.OscapConfig{
-				Datastream: oscapConfig.DataStream,
-				ProfileID:  oscapConfig.ProfileID,
-			},
-		)
-	}
-
```
thozza, line 140: why was this deleted?

internal/distro/rhel8/imagetype.go
```diff
@@ -38,6 +38,11 @@ const (
 
 	// blueprint package set name
 	blueprintPkgsKey = "blueprint"
+
+	// hardcode the datastream since this is provided by the scap-security-guide rpm
+	// and is not expected to change.
+	centosDatastream = "/usr/share/xml/scap/ssg/content/ssg-centos8-ds.xml"
+	rhelDatastream   = "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
```
thozza, line 45: again, I would like to see these names prefixed with `oscap` to make it obvious what they are for...

internal/distro/rhel9/imagetype.go
```diff
@@ -41,6 +41,11 @@ const (
 
 	// blueprint package set name
 	blueprintPkgsKey = "blueprint"
+
+	// hardcode the datastream since this is provided by the scap-security-guide rpm
+	// and is not expected to change.
+	centosDatastream = "/usr/share/xml/scap/ssg/content/ssg-cs9-ds.xml"
+	rhelDatastream   = "/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml"
```
thozza, line 48: again, I would like to see these names prefixed with `oscap` to make it obvious what they are for...

thozza (commented):
In general, changes look good. However I'm not sure if removing the `datastream` option from the BP customization is a good thing. I still agree with hardcoding the path per distro if none is provided by the user.

internal/blueprint/customizations.go
teg: Yeah, probably don't want to break backwards compatibility, my suggestion was meant to be to only restrict it in the cloud api as that is new.

kingsleyzissou: My bad, I misunderstood that. Will fix

internal/distro/rhel7/images.go
kingsleyzissou: Because it's essentially dead code. We validate this earlier and return and error if a user tries to provide oscap customizations for a RHEL7 image. I can leave it in.

internal/distro/fedora/distro.go
kingsleyzissou: I've added the prefix and moved this to the oscap package... imo it's slightly cleaner

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 30ee243a7b3e11f411baf339941223d55ce7a781 with the main merge-base 815f6c4f163e3aabed8c7ead40cb05d9a5f1cb20).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4558019511/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4558019511/artifacts/browse

internal/distro/rhel7/images.go
thozza: I'm inclining to keep `images.go` as similar as possible across various distros, since I see them as candidates for deduplication. But removing this is not a big deal. I'll leave it up to you, as I'm OK with either way after your explanation.

thozza (approved):
LGTM, nice work!

internal/distro/rhel7/images.go
kingsleyzissou: I left it in the end :)

PR #3539, opened 2023-06-30
Edge vsphere image type

7flying:
This adds the `edge-vsphere` image type for x86_64 and
aarch64. It is based on the 'edge-raw-image' but produces
a .vmdk image in order to upload the artifact to vSphere.

This image has Ignition support.

This pull request includes:

---

**WIP stuff**:
- ~~Hitting an issue with the manifest generation on aarch64, it is trying to pull the grub2-pc, grub2-efi-x64 and shim-x64 packages, which do not make sense from my POV.~~
- ~~I've set a 10GB `defaultSize` in the image type, but the artifact is just 1.6GB, setting 2GB gave me a runtime error regarding LVM stages having  insufficient free space. So don't know if this setting is OK.~~

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -150,21 +150,21 @@ func (impl *OSBuildJobImpl) getAWSForS3Target(options *target.AWSS3TargetOptions
 // getGCP returns an *gcp.GCP object using credentials based on the following
 // predefined preference:
 //
-// 1. If the provided `credentials` parameter is not `nil`, it is used to
-//    authenticate with GCP.
+//  1. If the provided `credentials` parameter is not `nil`, it is used to
+//     authenticate with GCP.
 //
-// 2. If a path to GCP credentials file was provided in the worker's
-//    configuration, it is used to authenticate with GCP.
+//  2. If a path to GCP credentials file was provided in the worker's
+//     configuration, it is used to authenticate with GCP.
 //
-// 3. Use Application Default Credentials from the Google library, which tries
-//    to automatically find a way to authenticate using the following options:
+//  3. Use Application Default Credentials from the Google library, which tries
+//     to automatically find a way to authenticate using the following options:
 //
-//    3a. If `GOOGLE_APPLICATION_CREDENTIALS` environment variable is set, it
-//        tries to load and use credentials form the file pointed to by the
-//        variable.
+//     3a. If `GOOGLE_APPLICATION_CREDENTIALS` environment variable is set, it
+//     tries to load and use credentials form the file pointed to by the
+//     variable.
 //
-//    3b. It tries to authenticate using the service account attached to the
-//        resource which is running the code (e.g. Google Compute Engine VM).
+//     3b. It tries to authenticate using the service account attached to the
+//     resource which is running the code (e.g. Google Compute Engine VM).
 func (impl *OSBuildJobImpl) getGCP(credentials []byte) (*gcp.GCP, error) {
 	if credentials != nil {
 		logrus.Info("[GCP] 🔑 using credentials provided with the job request")
@@ -377,7 +377,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	osbuildJobResult.OSBuildOutput, err = osbuild.RunOSBuild(jobArgs.Manifest, impl.Store, outputDirectory, exports, nil, extraEnv, true, os.Stderr)
 	// First handle the case when "running" osbuild failed
 	if err != nil {
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "osbuild build failed", nil)
+		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "osbuild build failed", err)
 		return err
 	}
 
--- a/internal/image/ostree_raw.go
+++ b/internal/image/ostree_raw.go
@@ -91,16 +91,27 @@ func (img *OSTreeRawImage) InstantiateManifest(m *manifest.Manifest,
 	buildPipeline.Checkpoint()
 
 	var art *artifact.Artifact
-	switch img.Compression {
-	case "xz":
-		ostreeCompressed := ostreeCompressedImagePipelines(img, m, buildPipeline)
-		art = ostreeCompressed.Export()
-	case "":
+	switch img.Platform.GetImageFormat() {
+	case platform.FORMAT_VMDK:
+		if img.Compression != "" {
+			panic(fmt.Sprintf("no compression is allowed with VMDK format for %q", img.name))
+		}
 		ostreeBase := baseRawOstreeImage(img, m, buildPipeline)
-		ostreeBase.Filename = img.Filename
-		art = ostreeBase.Export()
+		vmdkPipeline := manifest.NewVMDK(m, buildPipeline, nil, ostreeBase)
+		vmdkPipeline.Filename = img.Filename
+		art = vmdkPipeline.Export()
 	default:
-		panic(fmt.Sprintf("unsupported compression type %q on %q", img.Compression, img.name))
+		switch img.Compression {
+		case "xz":
+			ostreeCompressed := ostreeCompressedImagePipelines(img, m, buildPipeline)
+			art = ostreeCompressed.Export()
+		case "":
+			ostreeBase := baseRawOstreeImage(img, m, buildPipeline)
+			ostreeBase.Filename = img.Filename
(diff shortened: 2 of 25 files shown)


schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 5647245fdee5e003016a9b3ce8805fa883db9bc7 with the main merge-base a2c07ea83a2062f929862a5f3a46cf27ad86796b).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4573308512/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4573308512/artifacts/browse

internal/manifest/vmdk.go
```diff
@@ -53,3 +53,45 @@ func (p *VMDK) Export() *artifact.Artifact {
...
+
+// A VMDK turns a raw ostree image file into vmdk image.
+type VMDKOstree struct {
+	Base
+	Filename string
+
+	imgPipeline *RawOSTreeImage
+}
+
+func NewVMDKFromOstreeRaw(m *Manifest, buildPipeline *Build, imgPipeline *RawOSTreeImage) *VMDKOstree {
```
thozza, line 65: I would suggest to instead extend the `NewVMDK()` function to accept additional type of `imgPipeline`. Mainly because the rest of the code is basically identical at the first glance. The VMDK pipeline should not care necessarily which type of pipeline produces the raw image, it is just that we didn't have this use case before.

7flying: The problem is that the `VMDK` struct takes a `RawImage` instead of a `RawOSTreeImage`, so I would have to add another field in the struct to differentiate between them, so making a different struct with the associated functions seemed cleaner. What do you think?

yih-redhat:
@7flying 
I created a vmdk image and uploaded to vsphere, but failed to boot it, the error message is "Unsupported or invalid disk type 2 for 'scsi0:0'. Ensure that the disk has been imported."
Then I used vmkfstools in vsphere to convert the image again: vmkfstools -i original.vmdk -d thin new.vmdk
And I can boot with new vmdk image.

7flying:
> @7flying I created a vmdk image and uploaded to vsphere, but failed to boot it, the error message is "Unsupported or invalid disk type 2 for 'scsi0:0'. Ensure that the disk has been imported." Then I used vmkfstools in vsphere to convert the image again: vmkfstools -i original.vmdk -d thin new.vmdk And I can boot with new vmdk image.

I haven't had that issue.
When you are creating the VM you need to add a new (existing) hard disk, add the .vmdk there, and then delete the original one. Then you edit the new hard disk and select the `Virtual Device Node` that has the `New Hard Disk` words on it, sometimes it's `SCSI(0:0)` other times `IDE(0:0)` and so on.

yih-redhat:
> > @7flying I created a vmdk image and uploaded to vsphere, but failed to boot it, the error message is "Unsupported or invalid disk type 2 for 'scsi0:0'. Ensure that the disk has been imported." Then I used vmkfstools in vsphere to convert the image again: vmkfstools -i original.vmdk -d thin new.vmdk And I can boot with new vmdk image.
> 
> I haven't had that issue. When you are creating the VM you need to add a new (existing) hard disk, add the .vmdk there, and then delete the original one. Then you edit the new hard disk and select the `Virtual Device Node` that has the `New Hard Disk` words on it, sometimes it's `SCSI(0:0)` other times `IDE(0:0)` and so on.

I tried again as you suggested.
I deleted original disk and added my vmdk image as existing hard disk, and the virtual device node has the 'New hard disk' words. I tried SCSI(0:0) and SCSI(0:1), both failed to boot with same error message, but IDE(0:0) can boot. 
Will contact you later for this issue.

internal/manifest/vmdk.go
supakeen: You wouldn't need to add another struct, you could make the `imgPipeline` argument of type `Pipeline` which is the interface describing them *however* there is no `Filename` attribute or method in that interface so you indeed would end up with adding a different field, ensuring that not both are set, ensuring that at least one is set, etc.

Because that's (to me) more messy than duplicating these 30 lines of code. I'm ok with the current approach.

supakeen (approved):
While I agree with @thozza that it would make more sense to have an `interface{}` that describes what's actually necessary (which are the `.Name()`-method and `.Filename`-attribute as far as I can see.

For this case it can be done by setting the type of `imgPipeline` to `interface{}` and

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD e8ab26a75ce592029f7ae0d8f424d93177c9b5a3 with the main merge-base d7ed3cae6eb8960aa261900768f7cd56d0c74752).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4622662278/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/4622662278/artifacts/browse

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -377,7 +377,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	osbuildJobResult.OSBuildOutput, err = osbuild.RunOSBuild(jobArgs.Manifest, impl.Store, outputDirectory, exports, nil, extraEnv, true, os.Stderr)
 	// First handle the case when "running" osbuild failed
 	if err != nil {
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "osbuild build failed", nil)
+		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "running osbuild build failed", nil)
```
thozza, line 380: not sure if this change is actually useful :thinking:

internal/image/ostree_raw.go
```diff
@@ -91,16 +91,27 @@ func (img *OSTreeRawImage) InstantiateManifest(m *manifest.Manifest,
...
-		ostreeBase.Filename = img.Filename
-		art = ostreeBase.Export()
+	switch img.Platform.GetImageFormat() {
+	case platform.FORMAT_VMDK:
+		if img.Compression == "" {
+			ostreeBase := baseRawOstreeImage(img, m, buildPipeline)
+			vmdkPipeline := manifest.NewVMDKFromOstreeRaw(m, buildPipeline, ostreeBase)
+			vmdkPipeline.Filename = img.Filename
+			art = vmdkPipeline.Export()
+		}
```
thozza, line 101: The behavior if someone specifies a compression and uses VMDK format will be wrong and won't produce any error, which is not good.

IMO, it would be ideal to refactor the code in some way, so that the compression is applied conditionally regardless of the image format.

But as the bare minimum, I would propose to add an `else` branch with `panic()`.

internal/manifest/vmdk.go
thozza: Actually, the `Pipeline` interface does contain `Export() *artifact.Artifact` method, which contains the `Filename`.

cmd/osbuild-worker/jobimpl-osbuild.go
7flying: there is the same error message below in the same function, I simply want to distinguish where it was failed

internal/manifest/vmdk.go
7flying: mmm I'll take another look

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Would setting the error detail instead work for you (the `nil` value)?

7flying:
Made changes on VMDK so that the `NewVMDK` function also takes a `imgOstreePipeline *RawOSTreeImage`, and changed the VMDK struct accordingly to accommodate that (it has a `Pipeline` instead of a `RawImage`) + cleanups related to that.

On `ostree_raw` we explicitly fail if a compression type is specified with the VMDK format.

Moving this to  the images repo.

yih-redhat:
Added vmdk test script  test/cases/ostree-vsphere.sh, uses govc to upload image and test script passed in local env.

thozza:
> Added vmdk test script test/cases/ostree-vsphere.sh, uses govc to upload image and test script passed in local env.

This PR should have been closed, please create a new one...

thozza:
The support landed in the images repo https://github.com/osbuild/images/pull/24

PR #3541, opened 2023-07-03
Schedule a nightly CI pipeline against each PR

atodorov:
This pull request includes:

### DIFF
--- a/.github/workflows/trigger-gitlab.yml
+++ b/.github/workflows/trigger-gitlab.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       IMAGEBUILDER_BOT_GITLAB_SSH_KEY: ${{ secrets.IMAGEBUILDER_BOT_GITLAB_SSH_KEY }}
+      GITLAB_TOKEN: ${{ secrets.IMAGEBUILDER_BOT_GITLAB_PIPELINE_TRIGGER_TOKEN }}
     steps:
       - name: Report status
         uses: haya14busa/action-workflow_run-status@v1
@@ -36,6 +37,7 @@ jobs:
           per_page: 100
 
       - name: Checkout branch
+        id: pr_data
         run: |
           PR_DATA=$(mktemp)
           # use uuid as a file terminator to avoid conflicts with data content
@@ -45,6 +47,7 @@ jobs:
 
           PR=$(jq -rc '.[] | select(.head.sha | contains("${{ github.event.workflow_run.head_sha }}")) | select(.state | contains("open"))' "$PR_DATA" | jq -r .number)
           if [ ! -z "$PR" ]; then
+            echo "pr_branch=PR-$PR" >> "$GITHUB_OUTPUT"
             git checkout -b PR-$PR
           else
             git checkout ${{ github.event.workflow_run.head_branch }}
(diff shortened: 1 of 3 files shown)


.github/workflows/trigger-gitlab.yml
```diff
@@ -88,3 +91,33 @@ jobs:
...
+            # TODO: create if it doesn't exist (subsequent pushes)
+            NEW_SCHEDULE=$(curl --request POST --header "PRIVATE-TOKEN: $GITLAB_TOKEN" \
+                 --form description="9.3 nightly against ${{ steps.pr_date.outputs.pr_branch }}" \
+                 --form ref="${{ steps.pr_date.outputs.pr_branch }}" \
+                 --form cron="0 1 * * 5" --form cron_timezone="UTC" \
+                 --form active="false" \
+                 "https://gitlab.com/api/v4/projects/$PROJECT_ID/pipeline_schedules")
+            SCHEDULE_ID=$(jq .id <<< $NEW_SCHEDULE)
+
+            # adjust variables for new schedule
```
ondrejbudai, line 110: Isn't this potentially racy? You can pass variable in the trigger I think: https://docs.gitlab.com/ee/ci/triggers/#pass-cicd-variables-in-the-api-call

atodorov: > You can pass variable in the trigger I think: https://docs.gitlab.com/ee/ci/triggers/#pass-cicd-variables-in-the-api-call

I will update for this. 

> Isn't this potentially racy?

However racy with what ?

atodorov: I think I see what you mean here but I don't see how I can specify the pipeline I want to trigger. There's PROJECT_ID and TOKEN, which is used for authentication. I don't see where I can specify the actual pipeline that I want to trigger.

ondrejbudai: Ah, I see, we are running this as a scheduled pipeline. In that case, your method is probably the only viable one, since you cannot pass extra variables to the `play` route: https://docs.gitlab.com/ee/api/pipeline_schedules.html#run-a-scheduled-pipeline-immediately

The alternative would to run it just a plain pipeline and not a scheduled. I'm not sure, though, how much would that change all our scripts.

atodorov: > The alternative would to run it just a plain pipeline and not a scheduled.

The question around changes aside, how does trigger specify which pipeline to execute? I assume it will execute all that are defined, no ?

ondrejbudai: Indeed, we would have to make them running conditionally based on a variable.

atodorov: I've discovered that I can actually pass `--form "variables[CI_PIPELINE_SOURCE]=schedule"` as part of the POST payload. It triggers the nightly pipeline b/c we have conditionals on the CI_PIPELINE_SOURCE variable. However GitLab's UI doesn't show the pipeline as "Scheduled" b/c it isn't. Everything else seems to be as expected:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/920086961

atodorov:
Closed in favor of https://github.com/osbuild/osbuild-composer/pull/3543

PR #3545, opened 2023-07-05
distro/fedora: define fedora as blueprints

supakeen:
The Fedora distribution will be defined in blueprints, to do so I first have to tear everything down before it can be built up again.


achilleas-k:
This is an interesting approach to a problem we're all obviously thinking about, but I don't think blueprints are the solution.  Regardless, I'd hold off on these kinds of changes until after we split out the image definitions (https://github.com/osbuild/images).  I was hoping this would be done already, but the Fedora IoT PRs were blocking.

supakeen:
This will be re-opened with a new idea against the `images` repo.

PR #3551, opened 2023-07-10
split: replace internal packages with images library

achilleas-k:
Remove all the internal package that are now in the github.com/osbuild/images package and vendor it.

A new function in internal/blueprint/ converts from an osbuild-composer blueprint to an images blueprint.  This is necessary for keeping the blueprint implementation in both packages.  In the future, the images package will change the blueprint (and most likely rename it) and it will only be part of the osbuild-composer internals and interface.  The Convert() function will be responsible for converting the blueprint into the new configuration object.

This pull request includes:


croissanne (approved):
😺😺😺😺😺🎉

PR #3558, opened 2023-07-17
blueprint: add sysconfig customization 

supakeen:
This adds partial configuration support for the sysconfig stage.

Note: this needs https://github.com/osbuild/images/pull/30 merged *first* and then the `go.mod` commit removed.


supakeen:
During todays discussion at the team meeting it was decided that the custom files customization will be used instead of adding customizations for each configuration file(group).

PR #3572, opened 2023-07-24
tests: Add a check for valid snapshot urls

bcl:
This pulls the list of snapshots from the rpmrepo API, greps the codebase for all uses of rpmrepo.osbuild.org that look like a snapshot name, and then checks to make sure they are still valid.

This pull request includes:

- [X] adequate testing for the new functionality or fixed issue
- [X] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3574, opened 2023-07-25
test: Fix Fedora 39 snapshot urls in test repo

bcl:
This fixes the bad f39 repo names, found by the new check in #3572 

I didn't combine this with the check because I wanted to make sure it fails properly first :)


ondrejbudai (approved):
Lovely!

PR #3576, opened 2023-07-26
test: Add workaround for BZ#2213660 and #2225667

henrywang:
All RHEL for Edge test on RHEL 9.3 failed due to bug https://bugzilla.redhat.com/show_bug.cgi?id=2225667 which would be related with systemd bug https://bugzilla.redhat.com/show_bug.cgi?id=2213660.
This PR is just a workaround, it'll be removed when issue fixed.

### DIFF
--- a/test/cases/installers.sh
+++ b/test/cases/installers.sh
@@ -71,6 +71,13 @@ EOFKS
     echo "============================"
 }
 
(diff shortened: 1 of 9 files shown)


ondrejbudai (approved):
Yep, got hit by the same bug on Fedora, it's really annoying. Thanks!

PR #3585, opened 2023-07-28
cloudapi: Add ability to skip uploading and save image locally

bcl:
During development it can be very useful to store the results locally instead of uploading to a remote system. This implements a development only option to help with that.

To use it you need to add CLOUD_LOCALSAVE to the server's environment. This can be done by editing /usr/lib/systemd/system/osbuild-composer.service and adding:

Environment="CLOUD_LOCALSAVE=1"

You can then use an 'upload_options' object to skip trying to upload to the default service for the type of image, eg:

    "image_requests": [
    {
      "architecture": "x86_64",
      "image_type": "guest-image",
      "upload_options": {
          "local-save": true
      },
      ...
    }]

The results will be saved to /var/lib/osbuild-composer/artifacts/UUID/ using the default filename for the image type.

This pull request includes:

I chose not to try to add testing for this since it is meant as developer only, and any bugs that crop up will be caught in use by someone who can then fix it.


achilleas-k (commented):
I like this idea.  The only minor comment would be to maybe namespace the env var to something like `IMAGEBUILDER_CLOUD_LOCALSAVE` or something.  I doubt there would be any conflict for anyone running the cloud API, but you never know where these things might end up in the future.

Another idea, to make it really truly a devel option, might be to put it behind a build flag, but that might be a step too far.

bcl:
CLOUD_ is the namespace I randomly picked over OSBUILD_ :) IMAGEBUILDER_ seems pretty awkward though. I considered a cmdline option (plumbing that into the cloud v2 module would be a pain), and buildtime option but I can see a situation where I would want to just switch this on without rebuilding to examine something. Checking an environmental variable is safe and fast so I think it's ok.

teg:
I wonder if we want to have a generic mechanism for these type of developer features to avoid them becoming API by accident (shipped in RHEL, discovered and relied upon by a customer).

A compile time option is one way, another might be to introduce an option in the API that you have to specify to indicate you understand you can't rely on the behavior. `"unstable": true` or something like that.

Specifically for this option (for the future), I wonder if there is a way to make this useful outside of development. On prem the worker API allows you to upload the image to composer, in the cloud we clearly can't support that (composer has no state). Though would it make sense to support a generic HTTP POST target and on prem it just happens to point to composer itself, but we could point it at anything? I'm not sure if that's useful or how to make it generic enough / deal with authentication, but if there is a way to make it work similarly to the generic S3 target I think that might be neat.

bcl:
I agree that we don't want this to become implicit API. As far as extending it to do HTTP POST, etc. I think that's a different issue that could be addressed later along with removing all the state of the weldr API.

I think the requirement for an environmental variable is enough to avoid users depending on it, unless they really know what they are doing. You need local access, and need to edit the .service file and need to re-edit it after upgrading. FWIW I'll change the 'namespace' to OSBUILD to make it more specific to our use.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -883,12 +883,26 @@ components:
...
         This should really be oneOf but AWSS3UploadOptions is a subset of
         AWSEC2UploadOptions. This means that all AWSEC2UploadOptions objects
         are also valid AWSS3UploadOptionas objects which violates the oneOf
         rules. Therefore, we have to use anyOf here but be aware that it isn't
         possible to mix and match more schemas together.
+    LocalUploadOptions:
+      type: object
+      additionalProperties: false
+      required:
+        - local-save
```
ondrejbudai, line 897: Can you make this `local_save`? We are pretty consistent with having `_` as the delimiter in API keys.

ondrejbudai (changes requested):
Making development of composer easier is always nice. I agree that having a unified method for introducing development-only (power user-only?) features is currently a heavily discussed topic, and we should solve it at some point, but I don't want to block this PR on this decision.

Other than my the hyphen in the API (see below), this looks good. :)

internal/cloudapi/v2/openapi.v2.yml
bcl: Sure!

internal/cloudapi/v2/handler.go
```diff
@@ -503,6 +530,13 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
 			if request.Koji == nil {
 				return HTTPError(ErrorJSONUnMarshallingError)
 			}
+		} else if isLocalSave(ir.UploadOptions) {
+			// Override the image type upload selection and save it locally
+			// Final image is in /var/lib/osbuild-composer/artifacts/UUID/
+			irTarget = target.NewWorkerServerTarget()
+			irTarget.ImageName = imageType.Filename()
+			irTarget.OsbuildArtifact.ExportFilename = imageType.Filename()
+			irTarget.OsbuildArtifact.ExportName = imageType.Exports()[0]
```
thozza, line 539: If I'm reading this right, in case one specified the `LocalUploadOptions` in the image request and the environment variable is not set, the `irTarget` will be `nil` and the image won't be uploaded anywhere. While in ideal case, the server should return an error saying that the specified upload target is not valid for the running configuration.

thozza (changes requested):

internal/cloudapi/v2/handler.go
bcl: Right, that's the same way it behaves now if the upload options are an empty object. The problem with the way it determines upload options is that it can't know that it is actually empty, the image type determines which struct it tries to marshal/unmarshal and it will end up being empty.

ondrejbudai: Would it be possible to add an error in that case @thozza mentioned? It might save some nerves developers that forgot to set the right env variable (or made a typo).

bcl: There already is, although I'm not sure if these errors make it back to the client. In the journal they look like:

```
Jun 15 11:35:06 fedora osbuild-worker[792]: time="2023-06-15T11:35:06-07:00" level=error msg="osbuild job failed: at least one target failed" jobId=37532cfb-aebf-45ea-8991-01f2d0af7d96
Jun 15 11:35:06 fedora osbuild-worker[792]: time="2023-06-15T11:35:06-07:00" level=error msg="failure details : [Code: 17, Reason: No AWS bucket provided, Details: org.osbuild.aws.s3]" jobId=37532cfb-aebf-45ea-8991-01f2d0af7d96
```

thozza: These would make it to the client, but too late. Basically after waiting for the image to be built for couple of minutes, since it fails when the worker tries to upload the image.

In the case of `LocalUploadOptions`, the implementation should fail if the ENV variable is not set, without even enqueuing any job. You can specifically check if `local-save` is set to true, since the default value for `bool` type is `false`.

bcl: Doh! good idea!

thozza (approved):
LGTM

thozza:
Out of curiosity, the `adequate documentation informing people about the change such as` is checked, but I'm not sure where is this feature documented :thinking: Do you plan to open a PR for the `guides` repo?

achilleas-k (approved):
LGTM.  Good to merge since guides are external, right?

thozza:
> LGTM. Good to merge since guides are external, right?

Yup... :wink:

PR #3599, opened 2023-08-04
Koji: expose image metadata in build extra metadata, including boot mode

thozza:
The main goal of this PR is to expose the image boot mode in the Koji Build Extra metadata field.

Notable changes:

- Refactor Koji structures to be more consistent with their JSON representation and to what's used in Koji documentation. The changes make data structures more generic, which will be later helpful for the consolidation of information attached to Koji builds.
- Expose the image output extra metadata also in the build extra metadata under the image filename. The same data structure is now used for these two situations.
- Add the image boot mode to the `OSBuildJob` options and `OSBuildJobResult` options and copy the value in worker from job options to the result. Unfortunately, there is no better way (see discussion in this PR) to enable `KojiFinalize` job to access the data.
- Modify Weldr API and Cloud API to set the boot mode in the `OSBuildJob` options.
- Extend the Koji image output extra metadata to contain image boot mode.
- Extend the `koji.sh` test case to verify the Koji buildinfo content, with the focus on the build extra metadata.

This PR partially relates to https://issues.redhat.com/browse/COMPOSER-1801

This pull request includes:

### DIFF
--- a/cmd/osbuild-koji-tests/main_test.go
+++ b/cmd/osbuild-koji-tests/main_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/osbuild/images/pkg/distro"
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/upload/koji"
 )
@@ -180,7 +181,8 @@ func TestKojiImport(t *testing.T) {
 			RPMs:         []rpmmd.RPM{},
 			Extra: koji.BuildOutputExtra{
 				Image: koji.ImageExtraInfo{
-					Arch: "noarch",
+					Arch:     "noarch",
+					BootMode: distro.BOOT_LEGACY.String(),
 				},
 			},
 		},
--- a/cmd/osbuild-koji-tests/main_test.go
+++ b/cmd/osbuild-koji-tests/main_test.go
@@ -141,7 +141,7 @@ func TestKojiImport(t *testing.T) {
 	require.NoError(t, err)
 
 	// Import the build
-	build := koji.ImageBuild{
+	build := koji.Build{
 		TaskID:    1,
 		Name:      buildName,
 		Version:   "1",
@@ -168,18 +168,18 @@ func TestKojiImport(t *testing.T) {
 			RPMs:  []rpmmd.RPM{},
 		},
 	}
-	output := []koji.Image{
+	output := []koji.BuildOutput{
 		{
 			BuildRootID:  1,
 			Filename:     filename,
 			FileSize:     uint64(filesize),
 			Arch:         "noarch",
-			ChecksumType: "md5",
-			MD5:          hash,
-			Type:         "image",
+			ChecksumType: koji.ChecksumTypeMD5,
+			Checksum:     hash,
+			Type:         koji.BuildOutputTypeImage,
 			RPMs:         []rpmmd.RPM{},
-			Extra: koji.ImageExtra{
-				Info: koji.ImageExtraInfo{
+			Extra: koji.BuildOutputExtra{
+				Image: koji.ImageExtraInfo{
 					Arch: "noarch",
 				},
 			},
--- a/cmd/osbuild-koji/main.go
+++ b/cmd/osbuild-koji/main.go
@@ -60,7 +60,7 @@ func main() {
 		return
 	}
 
-	build := koji.ImageBuild{
+	build := koji.Build{
 		TaskID:    uint64(taskID),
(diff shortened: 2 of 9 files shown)


teg:
Love the refactoring of the CGImport structs, looks like we might be able to include more useful info there.

One nit: I don't think you should need to copy the boot mode from the job to the result, if i remember correctly there should be a way to make the job struct available to the finalize job.

thozza:
> One nit: I don't think you should need to copy the boot mode from the job to the result, if i remember correctly there should be a way to make the job struct available to the finalize job.

There is none. `KojiFinalize` can access only the result from the job on which it depends, not the job options. There is no existing way arount this, other than copying the value (which we already do for pipeline names). Of course we could rework the worker API, but I didn't want to do it.

teg:
> > One nit: I don't think you should need to copy the boot mode from the job to the result, if i remember correctly there should be a way to make the job struct available to the finalize job.
> 
> There is none AFAICT. `KojiFinalize` can access only the result from the job on which it depends, not the job options. There is no existing way arount this, other than copying the value (which we already do for pipeline names). Of course we could rework the worker API, but I didn't want to do it.

Ah, you are right. We tried to make sure that you wouldn't need to copy from inputs to outputs, but I misremembered how. If you want to reuse results, you can depend on them multiple times (that's what I remembered), but indeed that's not possible for reusing inputs.

What would be possible though is to simply pass the boot mode as an input to `KojiFinalize` directly, rather than via the `OSBuild` job. Not saying that's necessarily right, just that that's how we initially designed it (to avoid copy-paste as it felt a bit weird to have to do that).

thozza:
> What would be possible though is to simply pass the boot mode as an input to `KojiFinalize` directly, rather than via the `OSBuild` job. Not saying that's necessarily right, just that that's how we initially designed it (to avoid copy-paste as it felt a bit weird to have to do that).

I considered this option (among others such as adding the boot mode to koji target result) and while possible, I dismissed it as not feasible.

My main reason is:
Koji composes commonly contain multiple image builds. This implies multiple `OSBuildJob` instances. Each of these images can have different boot mode. The order of osbuild job results in the dynamic arguments of Koji finalize job will be kept stable from when the job was enqueued, thus it should be possible to add a list of boot modes to Koji finalize job arguments with each entry corresponding to the osbuild job. However, I consider it a potential guesswork and opportunity for future bugs. I find keeping the boot mode information closely associated with the actual build by including it in the osbuild job results much cleaner. With this approach, the Koji finalize job does not need to associate image boot modes from its own arguments with each built image (osbuild job results passed as dynamic arguments of the job). There is basically no doubt which boot mode value corresponds with which image.

Moreover, based on the discussion in the team, we want to keep out options open and be able to eventually expose this (and future) information also in Cloud API or Weldr API. Including the boot mode information only in the Koji finalize job arguments would make the exposure Koji-specific.

FWIW, I also don't like copying of the value from job options to result, but for the lack of better mechanism it is what it is... As I mentioned in respective the commit, we already do the same thing for pipeline names, thus this is not the first instance of such copying.

teg:
Thanks for the explanation :) That makes sense!

thozza:
> Thanks for the explanation :) That makes sense!

Any time :sweat_smile:  I'm glad that you asked, since I would expect others to ask the same thing during the review and some of the thinking I went through is not visible in the commits :wink:

achilleas-k:
It seems this is becoming a general issue that we don't have a great solution for (I wouldn't call the current solution bad though).  The PipelineNames before and the BootMode now are both properties of an image type that we need to have access to after the build.  It used to be that any metadata we needed after a build was available through the osbuild output/metadata (package names and ostree commits), so once a job was sent to osbuild, we didn't need to carry over any build-related information.  The manifest and everything that went into creating it could be completely opaque to the worker and any code reading the job result (in other words, the concept of an "image type" disappears after the build is started).

Will we need a more general solution at some point for maintaining properties from the build job to the result?  The copying of PipelineNames in particular is a pretty ugly solution for being able to read build log metadata and separating them between the build pipeline and the rest.  It requires statically defining the names of pipelines on an image type, and making sure they match the ones that will be generated with instantiating a manifest, so they can be read without actually generating any pipeline.  I'm wondering if it would be better to have some way of reading image type properties from an `ImageType` instance after the build, but I'm not sure I like that idea.
Alternatively, we could generalise this build job metadata so that it's harder to miss by only defining it in one place.  The BootMode case is simpler because we don't need it in weldr, but the PipelineNames need to be set on the job explicitly wherever a build job is defined.  Maybe this will be less of a problem when there's only one API and therefore only one place where a build job is instantiated.

Anyway, this is all discussion for a later time.  This PR should stay as is.

achilleas-k (approved):

Issue #3107, opened 2022-10-31
specifying a parent commit when doing `compose start-ostree` fails to start compose


miabbott:
**Describe the bug**

When trying to compose an `edge-installer` artifact via `composer-cli compose start-ostree`, specifying a parent commit causes the compose operation to fail.

**Environment**
 * OS version (`/etc/os-release` and `/etc/redhat-release`):

```
[root@localhost ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.1 (Plow)
[root@localhost ~]# # cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="9.1 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.1"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.1 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/9/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.1"
[root@localhost ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.1 (Plow)
```

 * osbuild-composer version (`rpm -qi osbuild-composer)`

```
[root@localhost ~]# rpm -qi osbuild-composer
Name        : osbuild-composer
Version     : 62.1
Release     : 1.el9
Architecture: x86_64
Install Date: Tue Oct 18 13:22:04 2022
Group       : Unspecified
Size        : 17612
License     : ASL 2.0
Signature   : RSA/SHA256, Fri Sep 16 11:40:17 2022, Key ID 199e2f91fd431d51
Source RPM  : osbuild-composer-62.1-1.el9.src.rpm
Build Date  : Fri Sep 16 10:08:50 2022
Build Host  : x86-64-01.build.eng.rdu2.redhat.com
[...]

PR #3606, opened 2023-08-08
deps: update osbuild/images to 9548bf0d0140

achilleas-k:
Update osbuild/images dependency to osbuild/images@9548bf0d0140

Relevant PRs:
- https://github.com/osbuild/images/pull/43
- https://github.com/osbuild/images/pull/91

@henrywang It's now possible to specify parent ostree commits using the commit ID, like we could before we introduced changing refs. (see https://github.com/osbuild/osbuild-composer/issues/3107).

Can we add a test case for this to one of the ostree tests?  Building a child commit from a parent, but specify the parent using the commit ID instead of the ref.


henrywang:
@achilleas-k Questions about this change:

1. Does `--parent <parent ref>` still work? If not, the rebase upgrade, like upgrade from rhel 8 to rhel 9 will not work.
2. Commit ID is used as vaule of `--parent`. Which ref will be used if there're two or more refs in ostree repo?
3. If we have issue on 1 or 2, we should add one more arg `--commit`

achilleas-k:
> @achilleas-k Questions about this change:
> 
> 1. Does `--parent <parent ref>` still work? If not, the rebase upgrade, like upgrade from rhel 8 to rhel 9 will not work.

Yes, it still works.

> 2. Commit ID is used as vaule of `--parent`. Which ref will be used if there're two or more refs in ostree repo?

The ref for the new (child) commit will be the `--ref` argument or the default for the distribution of it's not specified.  For pulling the parent, the ref doesn't matter.  It just pulls the commit with the given ID from the repo.

> 3. If we have issue on 1 or 2, we should add one more arg `--commit`

There shouldn't be any issues.

achilleas-k:
Looking good.  Failures are
```
An error occurred (VpcLimitExceeded) when calling the CreateVpc operation: The maximum number of VPCs has been reached.
```

Marking PR as ready.

henrywang:
@achilleas-k Yes, it's not caused by this PR.

henrywang:
@achilleas-k All green now.

ondrejbudai (approved):

PR #3613, opened 2023-08-10
test: update edge-ami test to support aarch64

henrywang:
This PR is to support edge-ami for aarch64 test. It fixed `VpcLimitExceeded` and added available instance checking before deployment.


achilleas-k (approved):
Very nice!

PR #3618, opened 2023-08-12
test: run greenboot rollback test on ostree.sh, ostree-ami-image.sh and ostree-vsphere.sh

henrywang:
Do not need to run greenboot rollback case on all tests to save time.
ostree.sh covered rollback on kvm, ostree-ami-image.sh covered rollback on AWS ec2, and ostree-vsphere.sh covered rollback on ESXi.


achilleas-k (approved):
Nice!  Great test trimming while maintaining coverage.

PR #3619, opened 2023-08-12
test: wait for ami image avaiable to use before tag creation

henrywang:
This PR adds `aws ec2 wait` commands for instance terminate, image register, and instance running to avoid hard code `sleep`. And this PR also re-formats some long shell commands shorter

### DIFF
--- a/test/cases/ostree-ami-image.sh
+++ b/test/cases/ostree-ami-image.sh
@@ -224,6 +224,8 @@ clean_up () {
     if [[ -v INSTANCE_ID ]]; then
         aws ec2 terminate-instances \
             --instance-ids "${INSTANCE_ID}"
+        aws ec2 wait instance-terminated \
+            --instance-ids "${INSTANCE_ID}"
     fi
 
     # Remove bucket content and bucket itself quietly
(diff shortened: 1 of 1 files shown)


henrywang:
The `rpm-build` failed on fedora 39. It should be related with new fedora 39 branched. Rawhide pinned to fedora 40 now.

test/cases/ostree-ami-image.sh
```diff
@@ -690,12 +777,22 @@ for _ in $(seq 0 9); do
...
 
+# Check instance has been deployed correctly
+check_result
+
+INSTANCE_ID=$(jq -r '.Instances[].InstanceId' "${INSTANCE_OUT_INFO}")
+
 # wait for instance running
-sleep 5
+aws ec2 wait instance-running \
+    --instance-ids "$INSTANCE_ID"
```
achilleas-k, line 787: It took me a while to find the functional change here in this PR.  Can we split this into multiple commits so that the formatting changes are in one and the main functional change is in another?

achilleas-k (commented):
LGTM.  Minor comment inline.

test/cases/ostree-ami-image.sh
henrywang: Yes, you're right. Fixed. Thanks.

PR #3636, opened 2023-08-17
Add support for uploading an ostree commit to Pulp

achilleas-k:
A new target and uploader is added `pulp.ostree`.  The `pulp` namespace will most likely be used for more targets in the future.

To make an ostree commit available in Pulp, we perform the following actions:
- Upload the commit.
- If the user-defined repository does not exist, create it.
- If the repository was created, distribute it (make it available for download).
- Import the commit into the repository.

The uploader workflow was based on [Pulp's guide](https://docs.pulpproject.org/pulp_ostree/workflows/import.html#import-workflow) and [demonstration video](https://www.youtube.com/watch?v=gz_C3GtRxjs).
It uses a client API library that I autogenerated and made available here: https://github.com/osbuild/pulp-client

Relevant tickets:
https://issues.redhat.com/browse/HMS-1163
https://issues.redhat.com/browse/HMS-2340

---

About testing:
We can probably add a test in CI but in my local testing using the [One Container](https://pulpproject.org/pulp-in-one-container/) the import and distribution tasks take a few minutes and would be adding extra delays to our ostree integration tests.  Maybe we should add a separate test for this that just pushes as single commit and does a remote repository check (with `ostree remote add ...` and `ostree remote summary ...`).

Also, while testing, I discovered that there might be a bug when multiple commits with different refs to the same repository: https://github.com/pulp/pulp_ostree/issues/277

---

This pull request includes:

Guides PR: https://github.com/osbuild/guides/pull/144


achilleas-k:
Something I wasn't sure about is the decision to return from the upload code before async tasks are complete.
The import and distribute can take a while but I made it so that the URL is returned before they're finished.  Would it be better if we blocked while these tasks run?  I don't know how long it can take for them to finish in realistic scenarios.

achilleas-k:
Converted to draft until we add tests.

internal/weldr/upload.go
```diff
@@ -363,6 +375,10 @@ func uploadRequestToTarget(u uploadRequest, imageType distro.ImageType) *target.
 
 			TlsVerify: options.TlsVerify,
 		}
+	case *pulpOSTreeUploadSettings:
```
thozza, line 378: I'm wondering if this new target should be covered also in `targetsToUploadResponses()` function? :thinking:

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -904,6 +906,21 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
...
 			targetResult.Options = &target.ContainerTargetResultOptions{URL: client.Target.String(), Digest: digest.String()}
 
+		case *target.PulpOSTreeTargetOptions:
+			targetResult = target.NewPulpOSTreeTargetResult(nil)
+			archivePath := filepath.Join(outputDirectory, jobTarget.OsbuildArtifact.ExportName, jobTarget.OsbuildArtifact.ExportFilename)
+
+			creds := pulp.Credentials{
+				Username: targetOptions.Username,
+				Password: targetOptions.Password,
+			}
```
thozza, line 916: We tend to pass credentials in the target options only for the on-premise use case. Is there some plan to allow configuring some credentials for Pulp on workers? I mean, we may not want to be passing secrets from `image-builder` to composer with each request... :shrug:

thozza (commented):
I added a few comments, but the PR looks good. I have additional questions:

1. Is there a plan to add the support for Cloud API for this upload target?
2. Is there any dependency between publishing of the repo and importing the commit, which would mandate a requirement to make them synchronous?
3. Can importing of the commit or publishing of the repo fail after the request for the action is created? I'm asking, because we may run into a situation where we return success for the target from the worker, but the actual import or publishing failed eventually and the content would not be available.

internal/upload/pulp/pulp.go
```diff
@@ -127,3 +127,16 @@ func (cl *Client) ImportCommit(commitHref, repoHref string) (string, error) {
 
 	return result.Task, nil
 }
+
+// Distribute makes an ostree repository available for download. This task is
+// asynchronous. The returned value is the href for the import task.
```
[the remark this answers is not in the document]
achilleas-k: Indeed.  Good catch.

internal/upload/pulp/pulp.go
```diff
@@ -178,3 +191,63 @@ func (cl *Client) TaskWaitingOrRunning(task string) bool {
...
+// UploadAndDistributeCommit uploads a commit, creates a repository if
+// necessary, imports the commit to the repository, and distributes the
+// repository.
+func UploadAndDistributeCommit(archivePath, serverURL, repoName, basePath string, creds *Credentials) (string, error) {
+	client := NewClient(serverURL, creds)
+
+	// Check for the repository before uploading the commit:
+	// If the repository needs to be created but the basePath is empty, we
+	// should fail before uploading the commit.
+	log.Printf("checking if repository %q already exists", repoName)
```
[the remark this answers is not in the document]
achilleas-k: Right.  Thanks.

achilleas-k:
> I added a few comments, but the PR looks good. I have additional questions:
> 
> 1. Is there a plan to add the support for Cloud API for this upload target?

Yes, in fact that's the main goal.  However, since in the cloud API we associate upload targets with image types directly, we need to wait until the internal Pulp service is available before adding the target to the Cloud API.  For now, I'm adding it to weldr only so we can test everything.

> 2. Is there any dependency between publishing of the repo and importing the commit, which would mandate a requirement to make them synchronous?

The two actions can be done in any order and are basically independent of each other.  Publishing/distributing a repository before a commit is imported creates a distribution (a `content` path) that has nothing to serve (it just 404s).  As soon as the commit import is complete, the path becomes available.

> 3. Can importing of the commit or publishing of the repo fail after the request for the action is created? I'm asking, because we may run into a situation where we return success for the target from the worker, but the actual import or publishing failed eventually and the content would not be available.

That's definitely possible.  It's basically the thing I was wondering in my second comment.  It would make sense to wait for the import to finish, but it can take several minutes.  So I'm wondering if it's a good idea to block for several minutes, holding up the worker that's not really doing anything, while a remote server is processing.  I don't know what's better here.

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: Good point.  I didn't think about the service config here much.  I'll fix it.

internal/upload/pulp/pulp.go
```diff
@@ -178,3 +191,63 @@ func (cl *Client) TaskWaitingOrRunning(task string) bool {
...
+		return "", err
+	}
+
+	repoURL, err := url.JoinPath(serverURL, "pulp/content", basePath+"/")
+	if err != nil {
+		// this will happen if the serverURL is malformed, which would have
+		// already been caught by now
+		return "", err
+	}
+	log.Printf("repository url: %s", repoURL)
```
achilleas-k, line 250: I just realised that this is wrong.  The `basePath` can be empty or ignored if the repository already exists and has already been distributed.  We will have to find the distribution for the existing repository to return this.

teg:
> It would make sense to wait for the import to finish, but it can take several minutes. So I'm wondering if it's a good idea to block for several minutes, holding up the worker that's not really doing anything, while a remote server is processing.

This is what happens with AMIs today, AFAIK. A potential solution for both AMIs and ostree commits is to split the job in two pieces: 1) build+upload; and 2) import. Then importing could be done in parallel with future builds, so there is no blocking.

teg:
Questions:
1. Did you consider leaving the repository creation/publishing to the caller? Similarly to how S3 buckets are not set up in composer, but by the caller. Could simplify things potentially?
2. How will we deal with authentication between workers and pulp in crc? Will the pulp instance in crc publish a public API the worker can talk to?

achilleas-k:
> Questions:
> 
> 1. Did you consider leaving the repository creation/publishing to the caller? Similarly to how S3 buckets are not set up in composer, but by the caller. Could simplify things potentially?

Is the caller in this case the customer or the IB service?  We could go this way.  Assume it's all created beforehand.

> 2. How will we deal with authentication between workers and pulp in crc? Will the pulp instance in crc publish a public API the worker can talk to?

I think that's the plan, yes.  @croissanne ?

thozza:
> > It would make sense to wait for the import to finish, but it can take several minutes. So I'm wondering if it's a good idea to block for several minutes, holding up the worker that's not really doing anything, while a remote server is processing.
> 
> This is what happens with AMIs today, AFAIK. A potential solution for both AMIs and ostree commits is to split the job in two pieces: 1) build+upload; and 2) import. Then importing could be done in parallel with future builds, so there is no blocking.

I had a very similar idea, I would just make the import async and add a new job type which would wait for import to finish in parallel with any osbuild job. Basically in the same way as we do for depsolving.

The tricky part is that such a change would be very disruptive, since composer assumes that the UUID of the osbuild job is the actual compose UUID and we use that to determine status, logs, etc in API endpoints. We already have a special case for Koji where the last job is the KojiFinalize and its UUID is used as the compose UUID. We would have to do a similar thing for new non-Koji composes and only for some of them depending on the used target. I'm sure someone can figure out an elegant and backward incompatible way to make it work, I just wanted to point out that this is not a trivial task.

yih-redhat:
Hi @achilleas-k I have tested uploading commit in pulp and it seems works well.
My steps:
1. configure pulp using [one container](https://pulpproject.org/pulp-in-one-container/)
2. build edge-commit image and upload it to pulp. 
3. create local repo, pull content from pulp to local repo.
4. provision vm with local repo
5. build an upgrade edge-commit image, and upload to pulp. (same repo, same distribution, same path). Now pulp repository changed to version 2 with new commit.
6. pull new commit content from pulp to local repo.
7. run 'rpm-ostree upgrade' to upgrade current vm, success.

And I have some questions:
1. Could you please give an example of 'composer-cli' command to build and upload to pulp? 
2. Customer need to configure pulp properly before using composer to build image? like creating repo, creating distribution, etc. 
3. The upload will take a while, if customer check the composer job status, what's the expected result? 'running', 'uploading', 'finished',  etc?
4. If build success but upload failed, can user still be able to download the image?

kingsleyzissou:
@yih-redhat, Achilleas is out until tomorrow so I can try answer some of your questions. He has a small test though that he wrote here: [achilleas-k/pulp-ostree-test](https://github.com/achilleas-k/pulp-ostree-test )

The test does depend on https://github.com/pulp/pulp_ostree/issues/277, but there is an open pull request for this... I built the `all-in-one` container locally with the changes from that PR and I was able to test it successfully using the above tests.

> 1. Could you please give an example of 'composer-cli' command to build and upload to pulp?

For the configuration, here is a link to the guides PR: (https://github.com/osbuild/guides/pull/144). Or you could use the small test Achilleas wrote for reference.

> 2. Customer need to configure pulp properly before using composer to build image? like creating repo, creating distribution, etc.

As it stands the user doesn't need to do any of those steps. A repo will be created if it doesn't exist, the commit will get imported and the repository will get distributed.

> 3. The upload will take a while, if customer check the composer job status, what's the expected result? 'running', 'uploading', 'finished', etc?

As far as I know the job status will be 'running' even for the 'uploading' part... for on-prem we just have 'running', 'waiting', 'finished' & 'failed'. But maybe one of the others could correct me if I'm wrong.

> 4. If build success but upload failed, can user still be able to download the image? 

I don't think this is possible at the moment. Again, someone could correct me here.

yih-redhat:
Thank you @kingsleyzissou

yih-redhat:
What's the final decision of async import job? I found the compose job is at finished status, but importing job is not finished yet, usually need to wait several minutes until I can see content in pulp (At first I thought something was wrong in compose job but the log shows no error, then after a while I checked pulp again and found the content.). Filed a github bug  https://github.com/osbuild/osbuild-composer/issues/3673 to track this.

yih-redhat:
Composer will build the image first even if the pulp configuration in pulp.toml is wrong and job will fail at final upload step.
Composer should check and deny the build request if pulp configuration is wrong, just like what it does for --url parameter, if --url is wrong, the compose job will not start. 
Filed a bug https://github.com/osbuild/osbuild-composer/issues/3674 to track this.

yih-redhat:
https://github.com/pulp/pulp_ostree/issues/279 also blocks this PR.

I just found the child commit cannot be imported into pulp repo where parent commit was imported, and then I found @achilleas-k already filed this bug.

yih-redhat:
Added test case for pulp, and due to bug https://github.com/pulp/pulp_ostree/issues/279 , test case cannot success.

achilleas-k:
> One more question @kingsleyzissou , what OS does this PR support? rhel9.3 and later? how about fedora and centos-stream?

Any releases made after this PR is merged will support this.  We wont be gating the feature on any OS or OS version.  Also, we wont be backporting this to any existing RHEL versions and I don't think we want to go through the exception process to get this into 9.3 or 8.9.

FTR, this feature is primarily intended for the hosted service.

test/cases/ostree.sh
```diff
@@ -773,6 +799,437 @@ sudo ansible-playbook -v -i "${TEMPDIR}"/inventory \
...
+CACHE_ENABLED=True" >> settings/settings.py
+sudo podman run --detach \
+            --publish 9090:80 \
+            --name pulp \
+            --volume "$(pwd)/settings":/etc/pulp:Z \
+            --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z \
+            --volume "$(pwd)/pgsql":/var/lib/pgsql:Z \
+            --volume "$(pwd)/containers":/var/lib/containers:Z \
+            --device /dev/fuse \
+            docker.io/pulp/pulp:latest
```
kingsleyzissou, line 837: @achilleas-k @yih-redhat since the pulp-ostree issue has been closed and the PR merged, the changes are available in their nightly build of the container. I've tested it locally and it appears to work now

yih-redhat: @kingsleyzissou Thanks, let's wait and see the CI test result.

yih-redhat: Seems not fixed in the latest container docker.io/pulp/pulp:latest

........Error: Task /pulp/api/v3/tasks/018a82b0-4ca7-73da-a347-3dfdceaa40fe/ failed: 'The parent commit '37b860943b1b39f1cbade7e0f71016c8e64b34c96510d0b73981a567ac4514d2' could not be loaded'

kingsleyzissou: Yeah when I was testing it looks like only the `nightly` has the fix

yih-redhat: I can change to nightly, to make the test pass. But currently the CI test is not started because of RPM build failed for some os, can rebase fix these failures?

achilleas-k: Clicked the rebase.  Let's see.

achilleas-k:
Marking as ready since we have tests and we should be working to get this through tnow.

yih-redhat:
Hi @achilleas-k , for issue https://github.com/osbuild/osbuild-composer/issues/3673, is there any update?

Currently I can only check pulp to know if the import is completed, it might finish in about 5~20 minutes after compose job finished (importing child commit will take longer).

I suggest to fix this issue, but as this feature is for hosted service, maybe they have different thoughts.

yih-redhat:
Filed bug https://github.com/pulp/pulp_ostree/issues/288 in pulp repo. 

Upload parent commit into pulp and ostree pull can pull the commit content from pulp, then upload child commit into pulp, but ostree pull cannot find the new commit content.

yih-redhat:
Filed bug https://github.com/pulp/pulp_ostree/issues/289 in pulp repo.

Issue description:
Upload parent commit takes around 5 minutes, but upload child commit takes more than 40 minutes, that's too long as there is only minor update content in child commit.

achilleas-k:
> Hi @achilleas-k , for issue #3673, is there any update?

This wont be solved in this PR.  The current behaviour is compatible with what we do in other uploaders and a proper solution would be too big and is left for another time.  The alternative, quick solution would be to make the build wait for the import to finish which we don't want to do.

teg:
> Did I misread @teg's comment ([#3636 (comment)](https://github.com/osbuild/osbuild-composer/pull/3636#issuecomment-1686744109))? Looking at the upload code, it seems we block until the AMI is ready to boot.

That's what I intended to write / suggest we do. But I may have mistyped my comment.

cmd/osbuild-worker/config.go
```diff
@@ -71,6 +76,7 @@ type workerConfig struct {
 	GenericS3      *genericS3Config            `toml:"generic_s3"`
 	Authentication *authenticationConfig       `toml:"authentication"`
 	Containers     *containersConfig           `toml:"containers"`
+	Pulp           *pulpConfig                 `toml:"pulp"`
```
thozza, line 79: It would be good to extend the extend the unit test for config parser:
https://github.com/osbuild/osbuild-composer/blob/eb8b63f3ca166ecad13e10a21b8499344a824f86/cmd/osbuild-worker/config_test.go#L11

a follow up is OK...

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -259,6 +267,34 @@ func (impl *OSBuildJobImpl) getContainerClient(destination string, targetOptions
...
+	if creds != nil {
+		return pulp.NewClient(address, creds), nil
+	}
+
+	// read from worker configuration
+	if impl.PulpConfig.CredsFilePath == "" {
+		return nil, fmt.Errorf("pulp credentials not set")
+	}
+
+	return pulp.NewClientFromFile(impl.PulpConfig.ServerAddress, impl.PulpConfig.CredsFilePath)
```
thozza, line 295: Shouldn't you use `address` instead of `impl.PulpConfig.ServerAddress`? I think that I get your intention to not use user-provided server address with credentials from the configuration. However it can be used the other way around - use server address from the configuration with user-provided credentials.

Is this intentional?

thozza:
@achilleas-k yes, we always block until the image importing is finished. I must admit that I'm puzzled WRT which approach is the best here... Blocking the worker for tens of minutes is definitely not great. However, not having the image (commit) available in the target environment and claiming that everything is done is also not great.

I think that we should block for now, to be consistent with other uploaders and think about the best solution (potentially wait until this becomes an issue?)...

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: I can't remember if I meant to do it this way for the reason you mentioned, but I think you're right and I should change it.

cmd/osbuild-worker/config.go
achilleas-k: I'll do it here.

achilleas-k:
I addressed @thozza's minor comments.  I'll modify the uploaded to block until the repo is ready now.

achilleas-k:
The function that uploads and distributes a commit now waits for any async tasks to finish before returning.

thozza (approved):
LGTM

yih-redhat:
> The function that uploads and distributes a commit now waits for any async tasks to finish before returning. cc @yih-redhat

Thanks, and just let you know, there is a performance issue https://github.com/pulp/pulp_ostree/issues/289 here when uploading child commit, it could take more than 40 minutes. (my last attempt took 48 minutes)

achilleas-k:
> > The function that uploads and distributes a commit now waits for any async tasks to finish before returning. cc @yih-redhat
> 
> Thanks, and just let you know, there is a performance issue [pulp/pulp_ostree#289](https://github.com/pulp/pulp_ostree/issues/289) here when uploading child commit, it could take more than 40 minutes. (my last attempt took 48 minutes)

Thanks.  I think I'd rather get this merged now and then focus on these things later.

yih-redhat:
> > > The function that uploads and distributes a commit now waits for any async tasks to finish before returning. cc @yih-redhat
> > 
> > 
> > Thanks, and just let you know, there is a performance issue [pulp/pulp_ostree#289](https://github.com/pulp/pulp_ostree/issues/289) here when uploading child commit, it could take more than 40 minutes. (my last attempt took 48 minutes)
> 
> Thanks. I think I'd rather get this merged now and then focus on these things later.

Ok, now we have only one blocking bug https://github.com/pulp/pulp_ostree/issues/288, this bug is already fixed by pulp team, and I am verifying it now.

yih-redhat:
Verified pulp bug https://github.com/pulp/pulp_ostree/issues/288 with nightly container in docker.io, fixed.

I updated my test script (which has passed in my local env) and triggered the CI test, once the CI pass this PR is ok to merge.

achilleas-k:
RPM builds are failing on RHEL 8 and CS8 for aarch64.  @kingsleyzissou and I have been trying to reproduce it but in our local builds it's working fine so far.

kingsleyzissou:
I found this in the logs:
```
github.com/osbuild/pulp-client/pulpclient: /usr/lib/golang/pkg/tool/linux_arm64/compile: signal: killed
```

I think it's an OOM killer. When I try compiling `pulp-client`, I get:
```
fatal error: runtime: out of memory
```

achilleas-k:
After some more investigation by @ondrejbudai, he discovered that a 1 GiB swapfile solves the OOM issue for el8 aarch64 builds and we agreed to use that.

achilleas-k:
Rebased and fixed conflicts.

henrywang:
Hi! @achilleas-k @yih-redhat,
I found the Pulp test takes more than 1 hour.
1. Building installation and upload to pulp takes 9 minutes (19:03:48-19:12:35) -- This looks reasonable.
2. Pull from pulp server (local server) to local takes 10 minutes -- 10 minutes doesn't make sense. It's local to local.
3. Building upgrade image and upload to pulp server takes more than 30 minutes (19:26:06-20:07:06 not finished yet) -- It's not reasonable.

@achilleas-k Do you have any suggestion on that? Thank you!

achilleas-k:
> Hi! @achilleas-k @yih-redhat, I found the Pulp test takes more than 1 hour.
> 
> 1. Building installation and upload to pulp takes 9 minutes (19:03:48-19:12:35) -- This looks reasonable.
> 2. Pull from pulp server (local server) to local takes 10 minutes -- 10 minutes doesn't make sense. It's local to local.
> 3. Building upgrade image and upload to pulp server takes more than 30 minutes (19:26:06-20:07:06 not finished yet) -- It's not reasonable.
> 
> @achilleas-k Do you have any suggestion on that? Thank you!

Yih already has an issue open for the upload performance here https://github.com/pulp/pulp_ostree/issues/289.
That issue links to this more general issue about investigating the performance of the pulp ostree plugin: https://github.com/pulp/pulp_ostree/issues/169

achilleas-k:
@henrywang I'd like to get this through without blocking on the Pulp performance issues.  There are other things dependent on this and the slow performance is an external factor.  Maybe it's better to disable the test or run it only nightly with a longer timeout so that it doesn't affect CI runtime for all PRs.

What do you (and everyone else) think?

yih-redhat:
I updated the pulp test case due to performance bug https://github.com/pulp/pulp_ostree/issues/289

Added short pulp test case in a separate ostree_pulp.sh, will build commit, upload to pulp and provision edge vm, it's much faster now.
The upgrade test are removed and can be added back once the performance pulp bug fixed.

yih-redhat:
As a new separate test script, ostree-pulp.sh need to be added into rhel-edge-ci project, then ci can run pulp test.
The problem is, once we added pulp test in rhel-edge-ci, pulp test will run for each new pull request after that, and if this pull request is not merged at that time, pulp test will fail.
The solution is to add pulp test in rhel-edge-ci, and merge this pull request at the same time. Here is the steps:
1. add pulp test in rhel-edge-ci
2. trigger ci test for this pull request.
3. merge this pull request.

We can do it at our day time, so it will not impact any new pull request in osbuild-composer. @achilleas-k @kingsleyzissou Please let us know when this pull request is ready for merge.

achilleas-k:
Rebased and resolved conflict.

test/cases/ostree-pulp.sh
```diff
@@ -0,0 +1,465 @@
...
+/usr/libexec/osbuild-composer-test/provision.sh none
+
+# Set os-variant and boot location used by virt-install.
+case "${ID}-${VERSION_ID}" in
+    "rhel-9.3")
+        IMAGE_TYPE=edge-commit
+        OSTREE_REF="rhel/9/${ARCH}/edge"
+        OS_VARIANT="rhel9-unknown"
+        # Use a stable installer image unless it's the nightly pipeline
+        BOOT_LOCATION="http://download.devel.redhat.com/released/rhel-9/RHEL-9/9.2.0/BaseOS/x86_64/os/"
```
achilleas-k, line 23: @yih-redhat Is this correct?  The distro version ID is 9.3 but we're using 9.2.0 to boot.

yih-redhat: Hi @achilleas-k This is copied from edge commit test case https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree.sh#L111, and I also have the same question as you. I believe this is added by @henrywang , xiaofeng, could you please help to explain a bit more about this? 
Btw, this test script can pass without problem.

achilleas-k: It looks like it's just using the latest stable to avoid issues with the unreleased 9.3.  Resolving the conversation.

achilleas-k:
The PR is green and ready from my side.

yih-redhat:
> The PR is green and ready from my side.

Thanks, @henrywang and I will handle this.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -293,6 +301,44 @@ func (impl *OSBuildJobImpl) getContainerClient(destination string, targetOptions
 	return client, nil
 }
 
+func (impl *OSBuildJobImpl) getPulpClient(targetOptions *target.PulpOSTreeTargetOptions) (*pulp.Client, error) {
+	// Read server configuration and credentials from the target options and
+	// fall back to worker config if they are not set (targetOptions take precedent).
+	// Mixing sources is allowed. For example, the server address can be
+	// configured in the worker config while the targetOptions provide the
+	// credentials (or vice versa).
```
thozza, line 309: This should ideally go before the function definition, so that it shows up as a doc text.

internal/target/pulp_target.go
```diff
@@ -0,0 +1,33 @@
...
+	ServerAddress string `json:"server_address"`
+
+	// Repository to import the ostree commit to
+	Repository string `json:"repository"`
+
+	// BasePath for distributing the repository (if new)
+	BasePath string `json:"basepath,omitempty"`
+
+	Username string `json:"username"`
+	Password string `json:"password"`
```
thozza, line 16: Shouldn't these be marked as `omitempty`, since credentials can be set also in the worker configuration?

internal/target/pulp_target.go
```diff
@@ -0,0 +1,33 @@
+package target
+
+const TargetNamePulpOSTree TargetName = "org.osbuild.pulp.ostree"
+
+type PulpOSTreeTargetOptions struct {
+	// ServerAddress for the pulp instance
+	ServerAddress string `json:"server_address"`
```
thozza, line 7: I have the same comment as with credentials... The address can be set also in the worker config, thus it could be marked as `omitempty` here, no?

thozza (approved):
The PR looks good to me :+1: I added just a few comments on minor issues / nitpicks, which are not blockers from my PoV.

However, it would be nice to see at least some results from `test/cases/ostree-pulp.sh` run, before merging (I've read the comment about testing in the PR, but it would be great if we had a mechanism to run new ostree tests as part of the PR which adds them, without breaking all other PRs in this repo...).

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: Done

internal/target/pulp_target.go
achilleas-k: Done.

internal/target/pulp_target.go
achilleas-k: Done

thozza: Looks like this one is actually not done :sweat_smile:

thozza (approved):

thozza (approved):
:+1: :fishsticks:

achilleas-k:
@yih-redhat [edge-minimal-raw-cs8](https://github.com/osbuild/rhel-edge-ci/actions/runs/6534455519/job/17741696015) is failing on the persistent journal check.  Any idea why that might have changed?

achilleas-k:
Rebased and fixed go.mod conflicts.

thozza (approved):
LGTM

PR #3642, opened 2023-08-22
dependabot: group go package updates

ondrejbudai:
Imho, this is much saner than having so many PRs for all individual dependencies. Taken from osbuild/images.


thozza (approved):
:heart:

PR #3643, opened 2023-08-22
Tests: Remove deprecated --os-type cli argument

atodorov:
test logs print:

[2023-08-22T10:18:14-04:00] 💿 Install image via installer(ISO) on VM WARNING  --os-type is deprecated and does nothing. Please stop using it.


bcl (approved):
Looks ok to me, the docs say that `--os-variant` is a more specific way to tell it what kind of machine to make, and that it will try to guess when using a URL for the install media.

thozza (approved):
LGTM, thanks!

ondrejbudai (approved):
Nice!

ondrejbudai:
ubi-wsl having a bad time, nothing relevant.

PR #3653, opened 2023-08-24
test/cases/ubi-wsl: public ip fixes

croissanne:
Use the "Standard" public ip sku, as "Basic" produces warnings.

Spin until the vm gets assigned a valid ipv4 address.

### DIFF
--- a/test/cases/ubi-wsl.sh
+++ b/test/cases/ubi-wsl.sh
@@ -192,20 +192,33 @@ $AZURE_CMD vm create \
    --attach-os-disk "$AZ_DISK" \
    --os-type "windows" \
    --security-type "TrustedLaunch" \
+   --public-ip-sku Standard \
    --location "$AZURE_WSL_LOCATION" \
     --nic-delete-option delete \
     --os-disk-delete-option delete \
     --size "Standard_D2as_v5"
 
 $AZURE_CMD vm open-port --resource-group "$AZURE_RESOURCE_GROUP" --name "wsl-vm-$TEST_ID" --port 22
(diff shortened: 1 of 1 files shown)


ondrejbudai (dismissed):

thozza (approved):

croissanne:
Looking at the test, the retry counter for the host went all the way up to 9 (out of 10). Maybe more tries would be prudent ._.

thozza (approved):
:+1: 

Just realized that increasing the `sleep` would be also OK :innocent:

test/cases/ubi-wsl.sh
```diff
@@ -192,20 +192,33 @@ $AZURE_CMD vm create \
...
+    HOST=$($AZURE_CMD vm list-ip-addresses \
+       --resource-group "$AZURE_RESOURCE_GROUP" \
+       --name "wsl-vm-$TEST_ID" \
+       --query "[].virtualMachine.network.publicIpAddresses[0].ipAddress" \
+       --output tsv)
+
+    if echo "$HOST" | grep -Eq "^([0-9]{1,3}[\.]){3}[0-9]{1,3}$"; then
+        break
+    fi
+    if [ "$LOOP_COUNTER" = "10" ]; then
```
bcl, line 214: The counter is never going to go past 10 :)

thozza: :facepalm:

croissanne: I'm an idiot :crying_cat_face:

PR #3666, opened 2023-08-29
Handle panics in the osbuild job & fix panic when OCI authentication fails

ondrejbudai:
Two fixes related to each other:

1. When an osbuild job panics, but `osbuild` actually succeeded and no `JobError` was set, we currently report success back to composer. That's wrong. I added a `recover()` to set an error state.
2. Passing wrong credentials when uploading to OCI causes a panic. Seems like the OCI SDK doesn't handle this well. I worked around this by doing more checking when the error is returned.


bcl (approved):
Looks ok to me, other than not having a test -- which of course would be tricky to do with a panic.

PR #3672, opened 2023-09-04
Pulp cloudapi

kingsleyzissou:
This pull request includes:


ondrejbudai:
This is probably not the best place to mention it, but I will do it nevertheless: It would be great to version osbuild/pulp-client in the same way as we version osbuild/images. This way we get automation via dependabot to update composer whenever the client gets an update.

teg:
@achilleas-k mentioned that this is causing tricky issues on RHEL8, just wanted to point out that making this a RHEL9+ feature at first might be a good solution. pulp isn't supported on prem and the service runs on RHEL9...

achilleas-k:
> pulp isn't supported on prem and the service runs on RHEL9...

We could drop it from the weldr api upload options in general, though it'd be nice to have for users that want to have their own local repo management with their own pulp instance if they wanted.

I'm not entirely sure how to go about removing a dependency from certain builds though.

kingsleyzissou:
> @achilleas-k mentioned that this is causing tricky issues on RHEL8, just wanted to point out that making this a RHEL9+ feature at first might be a good solution. pulp isn't supported on prem and the service runs on RHEL9...

Oh that's an interesting idea. That could be a decent solution at first.

achilleas-k:
Adding a build tag to disable pulp integration and will push it to #3636

achilleas-k:
This PR will probably need more work.  One of the requirements of the parity feature is to be able to support both upload targets at the same time (s3 and pulp).  IB includes upload options with an image request, but on the composer side we always assume that only one upload type is supported by each image type, so we'll have to break that assumption here and decouple the image type from the upload target (but still fail on invalid combinations).

kingsleyzissou:
Closing in favour of @achilleas-k's PR

PR #3686, opened 2023-09-15
test/cases/ubi-wsl: set --os-type when creating vm

croissanne:
Solves `invalid usage for storage profile: attach existing managed OS disk`.


thozza (approved):

croissanne:
ubi-wsl succeeded, so going to admin merge.

PR #3698, opened 2023-09-21
templates/compose: add startingDeadlineSeconds to maintenance job

croissanne:
The job won't run if it doesn't get scheduled within 30 minutes. This prevents the job running multiple times in a row if it didn't get scheduled, for instance due to resource limits.


thozza (approved):
AFAICT, these changes look sane.

PR #3716, opened 2023-10-04
cloudapi/v2: add all existing customizations to openapi spec

croissanne:


bcl:
Could you also extend the tests in `compose_test.go` to include the new stuff?

henrywang:
`edge-vsphere-cs9` and `edge-vsphere-9` failure has been fixed by PR https://github.com/osbuild/osbuild-composer/pull/3724. This failure can be ignored in this case or make a rebase.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1156,6 +1156,130 @@ components:
               items:
                 type: string
                 example: "firewalld"
+        hostname:
+          # todo
```
thozza, line 1160: What TODO? :thinking:

thozza (commented):
Changes LGTM.

I agree with @bcl that the `compose_test.go` should be extended to reflect these changes.

ondrejbudai (commented):
Awesome! One small nitpick inline.

internal/cloudapi/v2/openapi.v2.yml
croissanne: todo to remove the todo :duck:

internal/cloudapi/v2/compose_test.go
```diff
@@ -87,6 +87,73 @@ func TestGetBlueprintWithCustomizations(t *testing.T) {
...
 			},
 		},
+		Firewall: &FirewallCustomization{
+			Ports: common.ToPtr([]string{
+				"22/tcp",
+			}),
+		},
+		Hostname:           common.ToPtr("hostname"),
+		InstallationDevice: common.ToPtr("/dev/sda"),
+		Kernel: &struct {
```
bcl, line 97: Why embed structs here instead of using the types from the generator?

croissanne: In that case I'd have to split the sub-objects out in the spec to be a separate entry under schemas, objects which are immediately defined are also nested in the generated code. 

https://github.com/croissanne/osbuild-composer/commit/9cbf8b26d8504aae578b46400aa8d23e2b104d1f#diff-8bbd501fd1c37ce113701845557026187ef15239c63c0a284961a72517f6891bR373-R380

bcl: Oh, I see. This more of the openapi generator not knowing how to make decent Go code :/ But if you have a way to make it behave I'd pick that over this embedded struct stuff. We ought to be able to have properly defined structs in one place instead of needing to copy-paste these embedded things everywhere.

croissanne: will do :)

bcl (approved):
Looks good!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1156,6 +1156,37 @@ components:
               items:
                 type: string
                 example: "firewalld"
+        hostname:
+          type: string
+          description: Configures the hostname
+          example: myhostname
+        kernel:
+          $ref: '#/components/schemas/Kernel'
```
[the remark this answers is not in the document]
croissanne: I don't thin kit's a huge deal currently, since they're only instantiated directly in the tests, but yea, I agree :)

PR #3718, opened 2023-10-05
Container: set the composer git revision during the build

thozza:
When the container with osbuild-composer gets built in our CI or by AppSRE, we do not set the composer version to any value (as we do when we built RPMs). As a result, the version reported by composer is always "devel". This is not useful for debugging and determining the used version of composer. In addition, this information now gets exposed in Koji builds, therefore it makes sense to make it useful.

This pull request includes:


croissanne (approved):

henrywang:
`edge-vsphere-cs9` and `edge-vsphere-9` failure has been fixed by PR https://github.com/osbuild/osbuild-composer/pull/3724. This failure can be ignored in this case or make a rebase.

PR #3730, opened 2023-10-10
internal/weldr: specify architecture of compose

croissanne:
This is useful in environments with multi-arch remote workers. Defaults to the host architecture.

---

Wondering if we should test the cross arch in weldr api, since it queries the state of the remote worker socket now :/ 

---

This pull request includes:

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -52,7 +52,7 @@ type API struct {
 	workers *worker.Server
 
 	solver       *dnfjson.BaseSolver
-	archName     string
+	hostArch     string
 	repoRegistry *reporegistry.RepoRegistry
 
 	logger *log.Logger
@@ -63,7 +63,6 @@ type API struct {
 
 	hostDistroName string                   // Name of the host distro
 	distroRegistry *distroregistry.Registry // Available distros
-	distros        []string                 // Supported distro names
 
 	//  List of ImageType names, which should not be exposed by the API
 	distrosImageTypeDenylist map[string][]string
@@ -97,7 +96,7 @@ func (cs ComposeState) ToString() string {
 // systemRepoNames returns a list of the system repos
 // NOTE: The system repos have no concept of id vs. name so the id is returned
 func (api *API) systemRepoNames() (names []string) {
-	repos, err := api.repoRegistry.ReposByArchName(api.hostDistroName, api.archName, false)
+	repos, err := api.repoRegistry.ReposByArchName(api.hostDistroName, api.hostArch, false)
 	if err == nil {
 		for _, repo := range repos {
 			names = append(names, repo.Name)
@@ -107,14 +106,16 @@ func (api *API) systemRepoNames() (names []string) {
 }
 
 // validDistros returns a list of distributions that also have repositories
-func validDistros(rr *reporegistry.RepoRegistry, dr *distroregistry.Registry, arch string, logger *log.Logger) []string {
+func (api *API) validDistros(arch string) []string {
 	distros := []string{}
-	for _, d := range dr.List() {
-		_, found := rr.DistroHasRepos(d, arch)
+	for _, d := range api.distroRegistry.List() {
+		_, found := api.repoRegistry.DistroHasRepos(d, arch)
 		if found {
 			distros = append(distros, d)
 		} else {
-			logger.Printf("Distro %s has no repositories, skipping.", d)
+			if api.logger != nil {
+				api.logger.Printf("Distro %s has no repositories, skipping.", d)
+			}
 		}
 	}
 
@@ -136,13 +137,12 @@ func NewTestAPI(solver *dnfjson.BaseSolver, arch distro.Arch, dr *distroregistry
 		store:                    store,
 		workers:                  workers,
 		solver:                   solver,
-		archName:                 arch.Name(),
+		hostArch:                 arch.Name(),
 		repoRegistry:             rr,
 		logger:                   logger,
 		compatOutputDir:          compatOutputDir,
 		hostDistroName:           hostDistro.Name(),
 		distroRegistry:           dr,
-		distros:                  validDistros(rr, dr, arch.Name(), logger),
 		distrosImageTypeDenylist: distrosImageTypeDenylist,
 	}
 	return setupRouter(api)
@@ -158,7 +158,7 @@ func New(repoPaths []string, stateDir string, solver *dnfjson.BaseSolver, dr *di
 	if err != nil {
 		return nil, fmt.Errorf("failed to read host distro information")
 	}
-	archName := common.CurrentArch()
+	hostArch := common.CurrentArch()
 
 	rr, err := reporegistry.New(repoPaths)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


internal/weldr/api.go
```diff
@@ -2375,6 +2389,7 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
 		Size          uint64              `json:"size"`
 		OSTree        ostree.ImageOptions `json:"ostree"`
 		Branch        string              `json:"branch"`
+		Arch          *string             `json:"arch"`
```
bcl, line 2392: Wouldn't this make more sense as part of the blueprint?

internal/weldr/api.go
```diff
@@ -2763,11 +2782,22 @@ func (api *API) composeTypesHandler(writer http.ResponseWriter, request *http.Re
 	}
 
 	// Get the distro specific arch so that we can return the correct list of image types
-	arch, err := d.GetArch(api.archName)
+	archName, err := api.parseArch(request.URL.Query())
```
bcl, line 2785: Is parseArch really needed? The only way to tell if the arch name is valid is to make a `d.GetArch` call with it, so I'm not sure any other kind of parsing is needed.

bcl:
In addition to this, if we are going to support other arches we need a way to query what arches are supported for each distro so maybe an API route like `/distros/arch/<distro-name>,...` to list them?

bcl:
> isn't this already implicit in the types endpoint? you can just list the supported types for a sepcific distro/arch?

only for a specific distro, the arch is assumed to be the host arch. Another possibility is to expand the distro naming to include the arch if it isn't the host's arch. eg. `fedora-39:aarch64` or something like that. Not sure how much I like the idea, but there it is :)

ondrejbudai:
ftr, I've noticed while working on other stuff that  the list of distributions available in Weldr is currently filtered by the host architecture: https://github.com/osbuild/osbuild-composer/blob/6267ecf12cec023139c5e152a1edf61bd809fa56/internal/weldr/api.go#L201

I guess that we can remove this filtering in this PR.

internal/weldr/api.go
```diff
@@ -2443,6 +2456,41 @@ func (api *API) composeHandler(writer http.ResponseWriter, request *http.Request
...
+		}
+		units, err := sdconn.ListUnitsByNamesContext(context.Background(), []string{"osbuild-remote-worker.socket"})
+		if err != nil {
+			log.Println("Error listing units:", err)
+		}
+		if len(units) > 1 {
+			log.Println("Found multiple units with the name \"osbuild-remote-worker.socket\"")
+		}
+		if len(units) == 0 {
+			log.Println("Cross-architecture build requested while \"osbuild-remote-worker.socket\" wasn't found, no remote workers can connect")
```
bcl, line 2473: What action would normally fix this? Do they need to manually run `systemctl start osbuild-remote-worker.socket`?

internal/weldr/api.go
```diff
@@ -3354,7 +3408,7 @@ func (api *API) fetchPackageList(distroName string, names []string) (packages rp
 		return nil, err
 	}
 
-	solver := api.solver.NewWithConfig(d.ModulePlatformID(), d.Releasever(), api.archName, d.Name())
+	solver := api.solver.NewWithConfig(d.ModulePlatformID(), d.Releasever(), api.hostArch, d.Name())
```
bcl, line 3411: I think this should be arch now, right? Same goes for `ReposByArchName` and other `solver.NewWithConfig` uses.

internal/weldr/api.go
croissanne: Well the idea is that if the remote socket is inactive, there are no remote workers up, meaning cross-arch builds won't be run. It's a proxy for not having a clear list of which workers have called in. We could also maintain a list of workers which have called into the worker api of course.

internal/weldr/api.go
croissanne: Ah indeed, I left the modules info, project depsolve and metadata preloading based on the host arch. Should I switch over one of the first two?

internal/weldr/api.go
teg: So to fix this, attach a worker, right?

Nitpicks incoming:

Workers could all be on the local socket ... Either by piping it over SSH or by having a local worker on emulated hardware (qemu-user or in a VM).

Not that that's something we should go out of our way to support, but maybe also not something to break on purpose?

Also, we currently allow workers to be attached after jobs have been queued, so even if we have remote workers they could just be "late".

How about: track the workers and add a new endpoint to query the state. Then the cli/UI could show an informational message "So... I did queue that job as you asked, but you better get a worker to connect or you won't have a good time..." (maybe run the phrasing of the message by someone else).

croissanne: So tracking the workers is also a best-effort kinda thing right? it would look at the workers which called in within the last x hours and if so let you queue it. So if a worker is 'late' we'd have the same problem. It is a nicer solution though I agree.

bcl: I was thinking of it more along the lines of me seeing this in the logs and not having any idea how to make sure it has the workers I need, and not knowing how to start them, etc.

internal/weldr/api.go
bcl: I think they should all use the specific arch, it's possible there will be a different set of packages so it should be as precise as possible.

internal/weldr/api.go
teg: Sanne: don't we know all the pending workers at any given time?

Brain: very good point, not very actionable as currently written.

croissanne: Ah right we do know the workers which have called in, but then I wonder how to do his in a distributed setup, like psql really doesn't feel like the place to track this.

croissanne: That said the workers don't have a unique identifier, so that means the list is just 'calls on the request job endpoint'.

croissanne: Looking at it further,  tracking the workers would require breaking the worker api to uniquely identify each worker, unless we simply want to keep track of `$arch job was requested x time ago`, one entry per arch. So I don't even know if we should have that in the db.

croissanne:
> I don't hate the idea of starting tracking connected workers properly. Each worker sends its status to composer every 30 seconds. The status can bring additional useful information like available disk space, running jobs, etc. The database tracks only the latest status from each worker (they need to be identifiable somehow - instance id, hostname, random UUID...). Workers that don't send an update for 5 minutes are considered unavailable, and are dropped from the database after let's say 24 hours.

Would you make a separate endpoint, incorporate it into `-XPOST /jobs`, or into -`XGET /jobs/{token}`?

ondrejbudai:
Your question reminded me that we have heartbeats that solve a similar issue... I wonder if we can somehow fold these two concepts into one.

The first solution that comes to my mind would be to remove the current heartbeats, but instead save which worker is running a particular job (a new column I presume). Instead of the heartbeats call, introduce a new status call (as explained in my previous message) and include a list of jobs that the worker is currently running. Then, we can implement both concepts using the new worker statuses:

1) Job would be restarted if its worker is gone, or if it's no longer listed by its worker (e.g. the job panicked).
2) We can use the list of active workers in weldr to show a warning if someone enqueues a job for an arch that isn't available on that composer node.

However, when weldr is gone, we need to change 2) somehow. I think that the cloud API will return a warning if a job for submitted for an arch without workers.

While this idea is imho awesome (if composer can probably track its workers, imagine the possibilities with metrics :heart_eyes: ), it feels huge just for getting multi-arch into weldr...

bcl (dismissed):
Looks good to me

bcl (approved):
Still looks ok, even if CI has other ideas :)

PR #3734, opened 2023-10-11
cloudapi: Simplify the bp.Customizations code

bcl:
There is a lot of repeated checks for bp.Customization != nil, this simplifies that by creating an empty blueprint.Customizations at the top, and checking to see if it is still empty at the bottom and setting it back to nil.

Includes a new test for calling with an empty (not nil) v2.Customizations set on the request.

This pull request includes:


thozza (approved):
:heart:

PR #3744, opened 2023-10-16
Cloud API: Support selecting upload targets and add Pulp OSTree uploads

achilleas-k:
This PR is a follow-up to #3636 and replaces #3672.  It requires #3636 but isn't based on it (it wont compile until it is).

----

The purpose of this PR is to add the ability to upload edge and iot commits built through the cloud API to a Pulp server that supports ostree repos.  However, at the same time, we need to be able to support the old behaviour of uploading ostree commits to AWS S3.  This is a required part of [HMS-1798](https://issues.redhat.com/browse/HMS-1798).

I also think it's a good idea in general to be able to specify alternative upload targets for certain artifacts.  We could support uploading any artifact to S3 for example.

The main change in this PR is that an image request now includes an (optional) `upload_target` option which is an enum of the names of all the upload targets we support in the cloud API (in the form `org.osbuild.<target>`, e.g., `org.osbuild.aws.s3`).  Omitting the `upload_target` maintains the current behaviour, which means each image type has a default target.  Also, we define a mapping between upload targets and image types that they support and any invalid combination is a bad request.

The final commit that adds the pulp ostree upload target to the cloud API was cherry picked from #3672 (thanks @kingsleyzissou).

----

This pull request includes:


internal/cloudapi/v2/imagerequest.go
```diff
@@ -33,57 +33,232 @@ func (ir *ImageRequest) GetImageOptions(imageType distro.ImageType, bp blueprint
...
+
+	t.ImageName = fmt.Sprintf("composer-api-%s", uuid.New().String())
+	t.OsbuildArtifact.ExportFilename = imageType.Filename()
+	return t, nil
+}
+
+// Returns the name of the default target for a given image type name or error
+// if the image type name is unknown.
+func getDefaultTarget(imageType ImageTypes) (ImageRequestUploadTarget, error) {
+	switch imageType {
```
bcl, line 252: Almost seems like this doesn't need to be a switch anymore, a map would also work.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -767,6 +777,20 @@ components:
             $ref: '#/components/schemas/Repository'
         ostree:
           $ref: '#/components/schemas/OSTree'
+        upload_target:
```
bcl, line 780: Is it possible that we may want to upload to more than one of these? If so then it might be good to just make this a list now instead of dealing with singular/plural problems later :)

internal/cloudapi/v2/imagerequest.go
achilleas-k: yeah, I considered it.  I did the map for the support mapping because that felt easier, but this was already there so I didn't want to rewrite it.  But maybe it's worth the consistency.

internal/cloudapi/v2/openapi.v2.yml
achilleas-k: I think that would require `upload_options` to also be a list, which isn't a compatible change.  You're not wrong though, I'm just not sure how to tackle this.  I'm open to ideas.

bcl: Ah, I see. Then it would need to be a list of objects with both items. Not something that can be done in a backwards compatible way with openapi, so probably better to wait until it's really needed and then give them a new name :)

achilleas-k: Maybe it's time we started thinking about v3 API :thinking: :smile:

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -692,6 +694,14 @@ components:
...
         url:
           type: string
+    PulpOSTreeUploadStatus:
+      type: object
+      required:
+        - repo_url
+      properties:
+        repo_url:
+          type: string
+          example: 'TODO'
```
kingsleyzissou, line 704: I left this in there :|

achilleas-k: Oops :sweat_smile:

achilleas-k:
#3636 has been merged.  Rebased and marked as ready.

internal/cloudapi/v2/openapi.v2.yml
croissanne: And then we can recycle upload_options, and have multiple targets. If upload_targets isn't defined we can just do what we do now and take upload_options straight from the compose request.

achilleas-k: I like it.  Let's do this.  I think multiple targets wont be an issue but we're about to find out.

achilleas-k:
Converted to draft to block this while I work on supporting multiple targets.

achilleas-k:
Sorry this took so long.  Multiple upload targets added.

This now supports what @croissanne suggested and also adds the default upload options with the default target for the image type if specified, so all the following examples are valid:

Single explicit upload target:
```json
{
  "architecture": "x86_64",
  "image_type": "edge-commit",
  "upload_targets": [
    {
      "type": "org.osbuild.aws.s3",
      "upload_options": { ... }
    }
  ]
}

{
  "architecture": "x86_64",
  "image_type": "edge-commit",
  "upload_targets": [
    {
      "type": "org.osbuild.pulp.ostree",
      "upload_options": { ... }
    }
  ]
}

{
  "architecture": "x86_64",
  "image_type": "guest-image",
  "upload_targets": [
    {
      "type": "org.osbuild.aws.s3",
      "upload_options": { ... }
    }
  ]
}
```


Default only (will select the default for the image type):
```json
{
  "architecture": "x86_64",
  "image_type": "edge-commit",
  "upload_options": { ... },
  "upload_targets": []
}

{
  "architecture": "x86_64",
  "image_type": "guest-image",
  "upload_options": { ... },
  "upload_targets": []
}
```


One explicit upload target + a default:
```json
{
  "architecture": "x86_64",
  "image_type": "edge-commit",
  "upload_options": { ... },
  "upload_targets": [
    {
      "type": "org.osbuild.pulp.ostree",
      "upload_options": { ... }
    }
  ]
}

{
  "architecture": "x86_64",
  "image_type": "guest-image",
  "upload_options": { ... },
  "upload_targets": [
    {
      "type": "org.osbuild.aws.s3",
      "upload_options": { ... }
    }
  ]
}
```

Multiples of the same upload target are also supported.

achilleas-k:
We should add some tests for this.  I'm thinking we can change some of the compose requests in the API tests to use the new `UploadTargets` and for types that get uploaded to S3, we can define multiple upload targets and check that all three succeeded.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1083,6 +1127,21 @@ components:
...
+      properties:
+        basepath:
+          type: string
+          description: 'Basepath for distributing the repository'
+        repository:
+          type: string
+          description: 'Repository to import the ostree commit to'
+        server_address:
+          type: string
+          format: uri
```
achilleas-k, line 1144: These are the three options we support in weldr.  Do we want them all to be supported through the API?  Maybe server address should be configured instead, or is it useful to have it so IB can take care of it?

achilleas-k:
Edge test failures are failures to fetch from the rpm repos for the osbuild-composer build.  A glitch, I guess?

achilleas-k:
The response for the compose status is
```yaml
    ComposeStatus:
      allOf:
      - $ref: '#/components/schemas/ObjectReference'
      - type: object
        required:
          - status
          - image_status
        properties:
          status:
            $ref: '#/components/schemas/ComposeStatusValue'
          image_status:
            $ref: '#/components/schemas/ImageStatus'
          image_statuses:
            type: array
            items:
              $ref: '#/components/schemas/ImageStatus'
          koji_status:
            $ref: '#/components/schemas/KojiStatus'
```

which means we can use the `image_statuses` array to send responses for the multiple uploads:
```yaml
    ImageStatus:
      required:
       - status
      properties:
        status:
          $ref: '#/components/schemas/ImageStatusValue'
        upload_status:
          $ref: '#/components/schemas/UploadStatus'
        error:
          $ref: '#/components/schemas/ComposeStatusError'
```

but what should we do with the `image_status` at the top level?  Should it mirror the request?  Meaning, if the request defines `upload_options` in the old top level part of the request, the result of that specific upload request goes into the `image_status`.  Any upload defined in `upload_targets` goes in the `image_statuses`.  Do we need to keep the order the same so they match the request?  Am I overcomplicating this?

thozza (commented):
Nice work, the PR looks good! I have just one small nitpick...

internal/cloudapi/v2/imagerequest.go
```diff
@@ -108,156 +283,148 @@ func (ir *ImageRequest) GetTarget(request *ComposeRequest, imageType distro.Imag
...
+		UploadTargetTypeOrgOsbuildPulpOstree: {
+			ImageTypesEdgeCommit: true,
+			ImageTypesIotCommit:  true,
+		},
+	}
+}
+
+// GetImageTargets returns the targets for the ImageRequest. Merges the
+// UploadTargets with the top-level default UploadOptions if specified.
+func (ir *ImageRequest) GetTargets(request *ComposeRequest, imageType distro.ImageType) ([]*target.Target, error) {
```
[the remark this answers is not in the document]
achilleas-k: Thanks.  Good catch.

achilleas-k:
Added support for multiple upload targets and the response now includes an array of `upload_status` objects under `upload_statuses`.  The top-level `upload_status` is always a copy of the first element of `upload_statuses`, so in the most common case where we only have one upload target, the `upload_statuses` array has only one element that is the same as the existing `upload_status`.

The structure was discussed and agreed with @croissanne.

achilleas-k:
RPM builds keep failing on this PR, which reminds me of the issue in the previous PR https://github.com/osbuild/osbuild-composer/pull/3636#issuecomment-1726093137, but I can't see how this PR would be causing the same issue again.  Other builds in composer CI seem to be fine.

thozza:
@achilleas-k, I just checked some random build log and it seems that some unit tests are failing...

```
github.com/osbuild/osbuild-composer/internal/store
--- FAIL: TestStore (0.00s)
    --- FAIL: TestStore/TestSourceConfigGPGKeysTrueFalse (0.00s)
        store_test.go:508: 
            	Error Trace:	/builddir/build/BUILD/osbuild-composer-93/_build/src/github.com/osbuild/osbuild-composer/internal/store/store_test.go:508
            	Error:      	Should be true
            	Test:       	TestStore/TestSourceConfigGPGKeysTrueFalse
        store_test.go:509: 
            	Error Trace:	/builddir/build/BUILD/osbuild-composer-93/_build/src/github.com/osbuild/osbuild-composer/internal/store/store_test.go:509
            	Error:      	Should be true
            	Test:       	TestStore/TestSourceConfigGPGKeysTrueFalse
        store_test.go:510: 
            	Error Trace:	/builddir/build/BUILD/osbuild-composer-93/_build/src/github.com/osbuild/osbuild-composer/internal/store/store_test.go:510
            	Error:      	Should be false
            	Test:       	TestStore/TestSourceConfigGPGKeysTrueFalse
        store_test.go:511: 
            	Error Trace:	/builddir/build/BUILD/osbuild-composer-93/_build/src/github.com/osbuild/osbuild-composer/internal/store/store_test.go:511
            	Error:      	Should be false
            	Test:       	TestStore/TestSourceConfigGPGKeysTrueFalse
FAIL
exit status 1
FAIL	github.com/osbuild/osbuild-composer/internal/store	0.023s
```

bcl:
Crap. Those are the tests I added in commit b786178077a23bc6aca9f65ee58651bd4dfb1244 and they were passing... but now also fail locally :/

achilleas-k:
That would explain the seemingly random nature of this issue.  Strange that I didn't come across any other failures in main or other PRs, but maybe I didn't look hard enough or they got restarted enough times that they passed.

Thanks for catching this @thozza.  I completely missed it and it was looking like the old OOM issue.

bcl:
https://github.com/osbuild/osbuild-composer/pull/3776 should fix that

croissanne:

achilleas-k:
I think the RHEL 8.8 RPM builds are getting killed again:
```
Nov 08 17:40:46 ip-172-31-22-179.ec2.internal kernel: oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/machine.slice/machine-cb421a227be348e29fe19c158933ca72.scope,task=compile,pid=29577,uid=1000
Nov 08 17:40:46 ip-172-31-22-179.ec2.internal kernel: Out of memory: Killed process 29577 (compile) total-vm:2798720kB, anon-rss:1645056kB, file-rss:0kB, shmem-rss:1088kB, UID:1000 pgtables:640kB oom_score_adj:0
```
From RHEL 8.8 aarch64 rpm build journal log: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5490961946

croissanne (approved):
thank you! this was a big effort ._.

PR #3750, opened 2023-10-18
Add pkg splunk_logging

diaasami:
This pull request includes:

### DIFF
--- a/pkg/splunk_logger/splunk_logger.go
+++ b/pkg/splunk_logger/splunk_logger.go
@@ -1,0 +1,147 @@
+package logger
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+const (
+	PayloadsChannelSize = 1000
+	// in seconds, how often batched events should be sent
+	SendFrequency = 5
+)
+
+type SplunkLogger struct {
+	client   *http.Client
+	url      string
+	token    string
+	source   string
+	hostname string
+
+	payloads chan *SplunkPayload
+}
+
+type SplunkPayload struct {
+	// splunk expects unix time in seconds
+	Time  int64       `json:"time"`
+	Host  string      `json:"host"`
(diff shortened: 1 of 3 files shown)


croissanne (commented):
I would just add the logrus splunk forwarder here, not the insights bits.

pkg/logger/logger.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+var logLevel logrus.Level
+
+var stdLoggerConfigd = false
+
+func ConfigLogger(log *logrus.Logger, level string) error {
```
croissanne, line 14: Is this necessarily? Like I think just the splunk hook should be here, configuration can/should be handled separately.

croissanne (changes requested):
LGTM, thank you! Can we squash the commits?

pkg/splunk_logger/splunk_logger.go
```diff
@@ -0,0 +1,147 @@
...
+		url:      url,
+		token:    token,
+		source:   source,
+		hostname: hostname,
+	}
+
+	ticker := time.NewTicker(time.Second * SendFrequency)
+	sl.payloads = make(chan *SplunkPayload, PayloadsChannelSize)
+
+	go sl.flushPayloads(ticker.C)
```
ondrejbudai, line 55: This line creates a potentially dangling goroutine (there's no way to return from `flushPayloads`). This is usually not an issue, but it may create small annoyances in certain cases (tests).

I suggest passing a `context.Context` into `NewSplunkLogger` and then to `flushPayloads`. When the context gets cancelled, make the function return (possibly after calling `SendPayloads` for one last time?).

ondrejbudai (changes requested):
Looks good, I have just one suggestion regarding a potentially dangiling goroutine. :)

ondrejbudai (approved):
Thanks!

croissanne (approved):
Still good :)

croissanne:

croissanne (approved):

PR #3766, opened 2023-10-27
Worker status heartbeat

croissanne:
Unresponsive workers (>=1 hour of no status update) are cleaned up.
    
    Several things are enabled by keeping track of workers, in future the
    worker server could:
    - keep track of how many workers are active
    - see if a worker for a specific architecture is available

### DIFF
--- a/internal/jobqueue/fsjobqueue/fsjobqueue.go
+++ b/internal/jobqueue/fsjobqueue/fsjobqueue.go
@@ -56,6 +56,15 @@ type fsJobQueue struct {
 	// reported as done.
 	jobIdByToken map[uuid.UUID]uuid.UUID
 	heartbeats   map[uuid.UUID]time.Time // token -> heartbeat
+
+	workerIDByToken map[uuid.UUID]uuid.UUID // token -> workerID
+	workers         map[uuid.UUID]worker
+}
+
+type worker struct {
+	Arch      string    `json:"arch"`
+	Heartbeat time.Time `json:"heartbeat"`
+	Tokens    map[uuid.UUID]struct{}
 }
 
 // On-disk job struct. Contains all necessary (but non-redundant) information
@@ -85,12 +94,14 @@ type job struct {
 // loaded and rescheduled to run if necessary.
 func New(dir string) (*fsJobQueue, error) {
 	q := &fsJobQueue{
-		db:           jsondb.New(dir, 0600),
-		pending:      list.New(),
-		dependants:   make(map[uuid.UUID][]uuid.UUID),
-		jobIdByToken: make(map[uuid.UUID]uuid.UUID),
-		heartbeats:   make(map[uuid.UUID]time.Time),
-		listeners:    make(map[chan struct{}]struct{}),
+		db:              jsondb.New(dir, 0600),
+		pending:         list.New(),
+		dependants:      make(map[uuid.UUID][]uuid.UUID),
+		jobIdByToken:    make(map[uuid.UUID]uuid.UUID),
+		heartbeats:      make(map[uuid.UUID]time.Time),
+		listeners:       make(map[chan struct{}]struct{}),
+		workers:         make(map[uuid.UUID]worker),
+		workerIDByToken: make(map[uuid.UUID]uuid.UUID),
 	}
 
 	// Look for jobs that are still pending and build the dependant map.
@@ -186,7 +197,7 @@ func (q *fsJobQueue) Enqueue(jobType string, args interface{}, dependencies []uu
 	return j.Id, nil
 }
 
-func (q *fsJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []string) (uuid.UUID, uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
+func (q *fsJobQueue) Dequeue(ctx context.Context, wID uuid.UUID, jobTypes, channels []string) (uuid.UUID, uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
@@ -231,6 +242,10 @@ func (q *fsJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []
 	j.Token = uuid.New()
 	q.jobIdByToken[j.Token] = j.Id
 	q.heartbeats[j.Token] = time.Now()
+	if _, ok := q.workers[wID]; ok {
+		q.workers[wID].Tokens[j.Token] = struct{}{}
+		q.workerIDByToken[j.Token] = wID
+	}
 
 	err := q.db.Write(j.Id.String(), j)
 	if err != nil {
@@ -240,7 +255,7 @@ func (q *fsJobQueue) Dequeue(ctx context.Context, jobTypes []string, channels []
 	return j.Id, j.Token, j.Dependencies, j.Type, j.Args, nil
 }
 
-func (q *fsJobQueue) DequeueByID(ctx context.Context, id uuid.UUID) (uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
+func (q *fsJobQueue) DequeueByID(ctx context.Context, id, wID uuid.UUID) (uuid.UUID, []uuid.UUID, string, json.RawMessage, error) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
@@ -268,6 +283,10 @@ func (q *fsJobQueue) DequeueByID(ctx context.Context, id uuid.UUID) (uuid.UUID, 
 	j.Token = uuid.New()
 	q.jobIdByToken[j.Token] = j.Id
 	q.heartbeats[j.Token] = time.Now()
+	if _, ok := q.workers[wID]; ok {
+		q.workers[wID].Tokens[j.Token] = struct{}{}
+		q.workerIDByToken[j.Token] = wID
+	}
 
 	err = q.db.Write(j.Id.String(), j)
 	if err != nil {
@@ -296,6 +315,10 @@ func (q *fsJobQueue) RequeueOrFinishJob(id uuid.UUID, maxRetries uint64, result 
 
 	delete(q.jobIdByToken, j.Token)
 	delete(q.heartbeats, j.Token)
+	if wID, ok := q.workerIDByToken[j.Token]; ok {
+		delete(q.workers[wID].Tokens, j.Token)
+		delete(q.workerIDByToken, j.Token)
+	}
 
 	if j.Retries >= maxRetries {
 		j.FinishedAt = time.Now()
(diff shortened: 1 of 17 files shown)


internal/worker/client.go
```diff
@@ -138,11 +149,101 @@ func NewClientUnix(conf ClientConfig) *Client {
...
+		if err != nil {
+			logrus.Errorf("Unable to update worker status: %v", err)
+		}
+
+		var buf bytes.Buffer
+		resp, err := c.NewRequest("POST", url.String(), map[string]string{"Content-Type": "application/json"}, &buf)
+		if err != nil {
+			logrus.Errorf("Unable to update worker status: %v", err)
+		}
+		if resp.StatusCode != http.StatusOK {
```
ondrejbudai, line 236: `resp` can `nil` here

henrywang:
@croissanne Test got error `Nov 15 13:06:15 runner-rhos-0x-rhel-8-9-large-7619 osbuild-worker[23989]: time="2023-11-15T13:06:15-05:00" level=error msg="Requesting job failed: error requesting job: 500 — Error is not an instance of HTTPError (IMAGE-BUILDER-WORKER-10001)"` when build ostree upgrade image.

croissanne:
> @croissanne Test got error `Nov 15 13:06:15 runner-rhos-0x-rhel-8-9-large-7619 osbuild-worker[23989]: time="2023-11-15T13:06:15-05:00" level=error msg="Requesting job failed: error requesting job: 500 — Error is not an instance of HTTPError (IMAGE-BUILDER-WORKER-10001)"` when build ostree upgrade image.

Hm is the full worker log in the workflow?

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -56,6 +56,15 @@ type fsJobQueue struct {
 	// reported as done.
 	jobIdByToken map[uuid.UUID]uuid.UUID
 	heartbeats   map[uuid.UUID]time.Time // token -> heartbeat
+
+	workerIDByToken map[uuid.UUID]uuid.UUID // token -> worker id
```
kingsleyzissou, line 60: Nitpick: Feel free to ignore this since it might be a pain to implement. But I'm wondering if we could/should start aliasing the `uuids` here because it gets a bit confusing and I had to read over this a few times.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -677,6 +713,102 @@ func (q *DBJobQueue) RefreshHeartbeat(token uuid.UUID) {
...
+
+	var count int
+	err = conn.QueryRow(context.Background(), sqlQueryHeartbeatsForWorker, workerID).Scan(&count)
+	if err != nil {
+		return err
+	}
+
+	// If worker has any active jobs, refuse to remove the worker
+	if count != 0 {
+		return jobqueue.ErrActiveJobs
```
kingsleyzissou, line 796: What happens here if there is a job that is stuck? Will the worker be stuck too or do we have a way of dealing with that?

internal/worker/client.go
```diff
@@ -138,11 +149,97 @@ func NewClientUnix(conf ClientConfig) *Client {
...
 	}
+	err = client.register()
+	if err != nil {
+		panic(err)
+	}
+	go client.workerHeartbeat()
+	return client
+}
+
+func (c *Client) register() error {
```
kingsleyzissou, line 164: Nitpick: this seems to be registering a worker only, so maybe we could call this `registerWorker`?

internal/worker/client.go
```diff
@@ -138,11 +149,97 @@ func NewClientUnix(conf ClientConfig) *Client {
...
+	return nil
+}
+
+func (c *Client) workerHeartbeat() {
+	//nolint:staticcheck // avoid SA1015, this is an endless function
+	for range time.Tick(time.Minute * 1) {
+		if c.workerID == uuid.Nil {
+			err := c.register()
+			if err != nil {
+				logrus.Errorf("Error registering working, %v", err)
```
kingsleyzissou, line 213: Typo :) "working"

internal/worker/client.go
```diff
@@ -138,11 +149,97 @@ func NewClientUnix(conf ClientConfig) *Client {
...
+				continue
+			}
+		}
+
+		url, err := c.server.Parse(fmt.Sprintf("workers/%s/status", c.workerID))
+		if err != nil {
+			logrus.Errorf("Unable to update worker status: %v", err)
```
kingsleyzissou, line 220: Maybe "parsing worker status"?

internal/worker/client.go
```diff
@@ -138,11 +149,97 @@ func NewClientUnix(conf ClientConfig) *Client {
...
+		resp, err := c.NewRequest("POST", url.String(), map[string]string{"Content-Type": "application/json"}, bytes.NewReader(buf.Bytes()))
+		if err != nil {
+			logrus.Errorf("Unable to update worker status: %v", err)
+			continue
+		}
+		if resp.StatusCode == http.StatusBadRequest {
+			logrus.Warning("Registering worker again")
+			err = c.register()
+			if err != nil {
+				logrus.Errorf("Error registering working, %v", err)
```
kingsleyzissou, line 234: Typo :)

pkg/jobqueue/dbjobqueue/dbjobqueue.go
croissanne: So this is just the heartbeats. Only jobs which are running get inserted into heartbeats, meaning that if this job stops responding it will get removed from the heartbeats, and thus the worker will not have this job attached to it anymore.

Jobs which are stuck in pending aren't inserted into heartbeats (yet), so those also don't count.

kingsleyzissou: Cool thanks for the clarification :smiley:

internal/jobqueue/fsjobqueue/fsjobqueue.go
croissanne: Do you think it's better now? There's a whole bunch of casts going on but ig the type fsjobqueue struct is a bit more self-explanatory now.

internal/worker/client.go
croissanne: woops, missed this one

kingsleyzissou (dismissed):

internal/jobqueue/fsjobqueue/fsjobqueue.go
kingsleyzissou: Sorry just saw this comment. Yeah I agree its a bit of a trade-off, but I think its easier to parse.

henrywang:
@croissanne All Edge tests have been moved to RHEL 9.4 and 8.10 with PR https://github.com/osbuild/osbuild-composer/pull/3808. Could you please a rebase to make test green? Thanks.

internal/worker/client_test.go
```diff
@@ -134,5 +134,5 @@ func TestProxy(t *testing.T) {
 	// - request job (succeeds)
 	// - upload artifact
 	// - cancel
-	require.Equal(t, 5, proxy.calls)
+	require.Equal(t, 6, proxy.calls)
```
thozza, line 137: nitpick: you should update the comment about this line (fix the number of calls and add the new call).

thozza (commented):
Overall this looks great! :rocket:  I added just a few comments / questions. Happy to approve once resolved.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/jsondb"
 )
 
+type token uuid.UUID
+type workerID uuid.UUID
```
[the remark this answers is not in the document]
thozza: np, it would be just good to be consistent and if we see value in this, use it also in the `JobQueue` interface, not only in the `fsjobqueue`.

internal/worker/api/openapi.yml
```diff
@@ -203,6 +204,43 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /workers:
+    post:
+      operationId: postWorkers
+      summary: Create a new worker
```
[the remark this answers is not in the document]
achilleas-k: I found this weird too and had the exact same thought (not a REST API designer, so maybe there are conventions I'm not aware of).  Maybe naming the endpoints `worker` would imply that only one worker is interacting?

Probably not that important for the worker API anyway.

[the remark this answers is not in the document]
kingsleyzissou: Thought it was a rest convention to use the plural.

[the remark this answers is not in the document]
thozza: Based on an off-channel discussion, this really is a convention.

internal/jobqueue/fsjobqueue/fsjobqueue.go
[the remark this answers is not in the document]
croissanne: up to you, I have no strong opinion.

internal/test/helpers.go
```diff
@@ -120,6 +120,10 @@ func TestRouteWithReply(t *testing.T, api http.Handler, external bool, method, p
 		return
 	}
 
+	if len(replyJSON) == 0 {
+		return
+	}
```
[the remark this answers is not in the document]
croissanne: When the worker reports its status it receives an empty body in the response, and then the unmarshal fails since it's not valid json. The alternative is to return `{}` instead of nothing in the worker server, which might actually be more correct.

internal/worker/api/openapi.yml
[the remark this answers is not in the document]
croissanne: indeed, it's convention to use plural for items which are collections, since there are multiple workers (we just don't have a listing endpoint), the endpoint should be plural.

internal/test/helpers.go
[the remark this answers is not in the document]
croissanne:
```diff --git a/internal/worker/server.go b/internal/worker/server.go
index 254c1421c..58ce6c824 100644
--- a/internal/worker/server.go
+++ b/internal/worker/server.go
@@ -1068,7 +1068,7 @@ func (h *apiHandlers) PostWorkerStatus(ctx echo.Context, workerIdstr string) err
                return api.HTTPErrorWithInternal(api.ErrorUpdatingWorkerStatus, err)
        }
 
-       return ctx.NoContent(http.StatusOK)
+       return ctx.JSON(http.StatusOK, map[string]string{})
 }
 
``` 

would fix this ig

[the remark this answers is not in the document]
croissanne: oh it would only fails because the expected json is `"{}"` :facepalm: , fixing

internal/jobqueue/fsjobqueue/fsjobqueue.go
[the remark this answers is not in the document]
croissanne: actually I agree, if we do this it should be in the jobqueue interface itself. I reverted this part, and maybe we can do it as a follow-up ?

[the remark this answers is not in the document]
thozza: ACK, a follow-up, if we think it would be helpful, sounds good.

thozza (approved):
LGTM, thanks for the changes!

achilleas-k (approved):

kingsleyzissou (approved):

PR #3775, opened 2023-11-01
Switch testing to RHEL-8.10 and RHEL-9.4 nightly

jrusz:
This should get merged after we don't need to do any testing on 9.3 and 8.9 which would probably be after they're GA.


croissanne:

henrywang:
@jrusz There's no RPM build job in this PR. How to make  this PR run gitlab jobs? Thanks.

jrusz:
> @jrusz There's no RPM build job in this PR. How to run gitlab job in this PR? Thanks!

I've already added rpmbuild jobs in a previous PR. This is just adding the new runners to the test runs. I've rebased this PR which will re-run it.

henrywang:
@jrusz All RHEL 8.10 and 9.4 tests are green!

test/cases/upgrade8to9.sh
```diff
@@ -103,7 +103,7 @@ sudo virt-install --name rhel-test \
                   --memory 3072 \
                   --vcpus 2 \
                   --disk size=20 \
-                  --location http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.8.0/compose/BaseOS/x86_64/os/ \
+                  --location http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.9.0/compose/BaseOS/x86_64/os/ \
```
atodorov, line 106: Shouldn't these actually be the 8.9 GA repository ?

test/cases/cross-distro.sh
```diff
@@ -110,8 +110,8 @@ ALL_DISTROS=$(find "$REPO_PATH" -name '*.json' -printf '%P\n' | awk -F "." '{ pr
 ALL_EXPECTED_DISTROS=$(echo "$ALL_DISTROS" | grep -E "$PATTERN" | grep -Ev 'beta|stream' | sort)
 # Warning: filter out the remaining distros by matching whole words to avoid matching
 # the value rhel-93 by the pattern rhel-9!
```
atodorov, line 112: ^^^ `rhel-93` above. Maybe available in other files as well.

test/cases/cross-distro.sh
```diff
@@ -110,8 +110,8 @@ ALL_DISTROS=$(find "$REPO_PATH" -name '*.json' -printf '%P\n' | awk -F "." '{ pr
...
-# If we're running on a RHEL 9.2 osbuild-composer doesn't know anything about 9.3
-# images so the value rhel-9.3 should be treated as unrecognized and error out as
+# If we're running on a RHEL 9.3 osbuild-composer doesn't know anything about 9.4
+# images so the value rhel-9.4 should be treated as unrecognized and error out as
```
atodorov, line 114: OTOH why not update this to something like `rhel-X.Y` so that we can avoid updating comment in the future ? WDYT ?

test/cases/minimal-raw.sh
```diff
@@ -86,10 +86,10 @@ SSH_KEY_PUB=$(cat "${SSH_KEY}".pub)
 EDGE_USER_PASSWORD=foobar
 
 case "${ID}-${VERSION_ID}" in
-    "rhel-8.9")
+    "rhel-8.10")
         OS_VARIANT="rhel8-unknown"
         ;;
-    "rhel-9.3")
+    "rhel-9.4")
```
atodorov, line 92: Can this just be a match on `rhel-8*`, `rhel-9*`, etc. If we need specifics for different minor versions we can always add them.

test/cases/oscap.sh
```diff
@@ -362,6 +362,7 @@ clean_up
 ###############################
 
 # Write a blueprint for hardened image.
+# TODO: Remove firewalld rules from tailoring once https://github.com/ComplianceAsCode/content/issues/11275 is fixed
```
atodorov, line 365: Do you have a COMPOSER ticket for this? I think this TODO will get forgotten very easily.

test/cases/ostree-ami-image.sh
```diff
@@ -80,7 +80,7 @@ SSH_KEY_PUB=$(cat "${SSH_KEY}".pub)
 IGNITION_USER=core
 
 case "${ID}-${VERSION_ID}" in
-    "rhel-9.3")
+    "rhel-9.4")
```
atodorov, line 83: Similar `rhel-9*` match here. WDYT?

test/cases/ostree-ignition.sh
```diff
@@ -116,7 +116,7 @@ IGNITION_USER_PASSWORD=foobar
 SYSROOT_RO="true"
 
 case "${ID}-${VERSION_ID}" in
-    "rhel-9.3")
+    "rhel-9.4")
```
atodorov, line 119: Similar `rhel-9*` match here. WDYT?

test/cases/ostree-pulp.sh
```diff
@@ -15,7 +15,7 @@ fi
 
 # Set os-variant and boot location used by virt-install.
 case "${ID}-${VERSION_ID}" in
-    "rhel-9.3")
+    "rhel-9.4")
```
atodorov, line 18: Similar `rhel-9*` match here. WDYT?

test/cases/ostree.sh
```diff
@@ -66,7 +66,7 @@ case "${ID}-${VERSION_ID}" in
             BOOT_LOCATION="${COMPOSE_URL:-}/compose/BaseOS/x86_64/os/"
         fi
         ;;
-    "rhel-8.9")
+    "rhel-8.10")
```
atodorov, line 69: Doesn't update `BOOT_LOCATION` several lines below.

test/cases/ostree.sh
```diff
@@ -108,7 +108,7 @@ case "${ID}-${VERSION_ID}" in
             BOOT_LOCATION="${COMPOSE_URL:-}/compose/BaseOS/x86_64/os/"
         fi
         ;;
-    "rhel-9.3")
+    "rhel-9.4")
```
atodorov, line 111: Same thing, BOOT_LOCATION needs update.

test/cases/ostree.sh
```diff
@@ -108,7 +108,7 @@ case "${ID}-${VERSION_ID}" in
...
         IMAGE_TYPE=edge-commit
```
atodorov, line 112: Another comment: not sure why we've kept references to previous minor releases but I think you need to do either one of these:

1) Remove them to keep this file clean or 
2) Keep the existing entries and add new ones for the new distros

atodorov (changes requested):
Overall looks good. I've added a few comments but only a couple seem to be real issues, the rest are nice-to-have.

test/cases/upgrade8to9.sh
jrusz: That's a good question. It depends on what we're actually testing here, I thought we're testing 8.10 nightly > 9.4 nightly upgrade path in which case I made a mistake and it should be 8.10 actually... However if we're testing latest RHEL-8-GA > latest RHEL-9-nightly then it should be GA here instead. I tried searching for an old ticket we had for this but I can't find this exact information anywhere, I'll ask the LEAPP team on slack.

test/cases/oscap.sh
jrusz: COMPOSER-2076

test/cases/ostree.sh
jrusz: Yeah it's probably redundant as we're only running this test on the latest nightly in the edge-ci so I'll remove all the older entries.

test/cases/upgrade8to9.sh
jrusz: I've ammended the related upgrade commit with the link to the slack thread and it's using the latest rhel-8 nightly as base now.

henrywang:
@jrusz All Edge tests(RHEL 9.4 and 8.10 included) are passed.

PR #3776, opened 2023-11-01
store: Fix test for CheckGPG

bcl:
Forgot the #1 rule of map iteration. Don't depend on the order. This uses a slice instead, which should be just as good for testing the loop variable alias behavior.

This pull request includes:


achilleas-k (approved):
Thanks for the quick fix!!

PR #3780, opened 2023-11-02
templates/packer: configure pulp creds on startup

kingsleyzissou:
Configure pulp credentials on worker startup.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_pulp_creds.sh
+++ b/templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_pulp_creds.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -eo pipefail
+source /tmp/cloud_init_vars
+
+echo "Deploy Pulp credentials."
+
+if [[ -z "$PULP_PASSWORD_ARN" ]]; then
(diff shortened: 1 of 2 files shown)


yih-redhat:
hi, with this pr, do we need to change/update test cases in https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree-pulp.sh ?
And one more question, I can see the pulp server, username and password in this pr, so there will be an embedded pulp server here? because from my understanding, customer is supposed to setup their own pulp server environment.

kingsleyzissou:
> hi, with this pr, do we need to change/update test cases in https://github.com/osbuild/osbuild-composer/blob/main/test/cases/ostree-pulp.sh ? And one more question, I can see the pulp server, username and password in this pr, so there will be an embedded pulp server here? because from my understanding, customer is supposed to setup their own pulp server environment.

This PR is so we can eventually use pulp in the service, at the moment there is a pulp instance in staging. This kind of goes together with the #3744 pr. As for users who want to use this on-prem, they will have to configure a pulp server themselves.

Regarding the test, I'm not sure if we run any tests against stage, maybe @croissanne can correct me?

achilleas-k (approved):
Sweet!!

PR #3783, opened 2023-11-06
Blueprints: Fix TOML filesystem `minsize` keyword

kingsleyzissou:
Due to an oversight, the toml and json tags for the `MinSize`field had different keywords. This commit fixes this by creating
a `minsize` toml tag and ensuring backwards compatability by checking the old `size` tag.

If both `minsize` & `size` are set in the toml block, the custom unmarshal function validates the input for inconsistencies.

Guides PR: https://github.com/osbuild/guides/pull/152

### DIFF
--- a/internal/blueprint/filesystem_customizations.go
+++ b/internal/blueprint/filesystem_customizations.go
@@ -9,7 +9,12 @@ import (
 
 type FilesystemCustomization struct {
 	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
-	MinSize    uint64 `json:"minsize,omitempty" toml:"size,omitempty"`
+	MinSize    uint64 `json:"minsize,omitempty" toml:"minsize,omitempty"`
+
+	// Note: The TOML `size` tag has been deprecated in favor of `minsize`.
+	// we check for it in the TOML unmarshaler and use it as `minsize`.
+	// However due to the TOML marshaler implementation, we can omit adding
+	// a field for this tag and get the benifit of not having to export it.
 }
 
 func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
@@ -22,17 +27,58 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
 		return fmt.Errorf("TOML unmarshal: mountpoint must be string, got %v of type %T", d["mountpoint"], d["mountpoint"])
 	}
 
+	var size uint64
+	var minsize uint64
+
+	// `size` is an alias for `minsize. We check for the `size` keyword
+	// for backwards compatibility. We don't export a `Size` field as
+	// we would like to discourage its use.
 	switch d["size"].(type) {
 	case int64:
-		fsc.MinSize = uint64(d["size"].(int64))
+		size = uint64(d["size"].(int64))
 	case string:
-		size, err := common.DataSizeToUint64(d["size"].(string))
+		s, err := common.DataSizeToUint64(d["size"].(string))
 		if err != nil {
 			return fmt.Errorf("TOML unmarshal: size is not valid filesystem size (%w)", err)
 		}
-		fsc.MinSize = size
+		size = s
+	case nil:
+		size = 0
 	default:
 		return fmt.Errorf("TOML unmarshal: size must be integer or string, got %v of type %T", d["size"], d["size"])
+	}
+
+	switch d["minsize"].(type) {
+	case int64:
+		minsize = uint64(d["minsize"].(int64))
+	case string:
+		s, err := common.DataSizeToUint64(d["minsize"].(string))
+		if err != nil {
+			return fmt.Errorf("TOML unmarshal: minsize is not valid filesystem size (%w)", err)
+		}
+		minsize = s
+	case nil:
+		minsize = 0
+	default:
+		return fmt.Errorf("TOML unmarshal: minsize must be integer or string, got %v of type %T", d["minsize"], d["minsize"])
+	}
+
+	if size == 0 && minsize == 0 {
+		return fmt.Errorf("TOML unmarshal: minsize must be greater than 0, got %v", minsize)
+	}
+
+	if size > 0 && minsize == 0 {
+		fsc.MinSize = size
+		return nil
+	}
+
+	if size == 0 && minsize > 0 {
+		fsc.MinSize = minsize
+		return nil
+	}
+
+	if size > 0 && minsize > 0 {
+		return fmt.Errorf("TOML unmarshal: size and minsize cannot both be set (size is an alias for minsize)")
 	}
 
 	return nil
--- a/internal/blueprint/filesystem_customizations_test.go
+++ b/internal/blueprint/filesystem_customizations_test.go
@@ -1,0 +1,159 @@
+package blueprint
+
+import (
+	"testing"
+
+	"github.com/BurntSushi/toml"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetFilesystems(t *testing.T) {
+
+	expectedFilesystems := []FilesystemCustomization{
+		{
+			MinSize:    1024,
+			Mountpoint: "/",
+		},
+	}
+
+	TestCustomizations := Customizations{
+		Filesystem: expectedFilesystems,
+	}
+
+	retFilesystems := TestCustomizations.GetFilesystems()
+
+	assert.ElementsMatch(t, expectedFilesystems, retFilesystems)
+}
+
+func TestGetFilesystemsMinSize(t *testing.T) {
+
+	expectedFilesystems := []FilesystemCustomization{
+		{
+			MinSize:    1024,
+			Mountpoint: "/",
+		},
+		{
+			MinSize:    4096,
+			Mountpoint: "/var",
+		},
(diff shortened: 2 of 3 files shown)


internal/blueprint/filesystem_customizations.go
```diff
@@ -21,20 +22,62 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
...
 		return fmt.Errorf("TOML unmarshal: size must be integer or string, got %v of type %T", d["size"], d["size"])
 	}
 
+	switch d["minsize"].(type) {
+	case int64:
+		minsize = uint64(d["minsize"].(int64))
+	case string:
+		s, err := common.DataSizeToUint64(d["size"].(string))
+		if err != nil {
+			return fmt.Errorf("TOML unmarshal: size is not valid filesystem size (%w)", err)
```
thozza, line 49: nitpick: the msg should probably use `minsize` instead of `size`.

internal/blueprint/filesystem_customizations.go
```diff
@@ -21,20 +22,62 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
...
+		size = 0
 	default:
 		return fmt.Errorf("TOML unmarshal: size must be integer or string, got %v of type %T", d["size"], d["size"])
 	}
 
+	switch d["minsize"].(type) {
+	case int64:
+		minsize = uint64(d["minsize"].(int64))
+	case string:
+		s, err := common.DataSizeToUint64(d["size"].(string))
```
thozza, line 47: c&p error? Shouldn't you use `d["minsize"]` here? :thinking:

internal/blueprint/filesystem_customizations.go
```diff
@@ -21,20 +22,62 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
...
+		return nil
+	}
+
+	if size == 0 && minsize > 0 {
+		fsc.MinSize = minsize
+		return nil
+	}
+
+	if size == minsize {
+		fsc.MinSize = size
```
thozza, line 73: nitpick: I would probably take `minsize` here, since it is the preferred option, although the end result is exactly the same.

I know that I suggested this, but wondering what others think - if we should not error in case both options are set to a value. Basically discouraging people to have both set...

internal/blueprint/filesystem_customizations_test.go
```diff
@@ -0,0 +1,173 @@
...
+[[customizations.filesystem]]
+mountpoint = "/var"
+size = 2048
+minsize = 1024
+`
+
+var tomlSizeAndMinSizeNotSet = `
+[[customizations.filesystem]]
+mountpoint = "/var"
+`
```
thozza, line 108: I would find it nicer if you just in-lined these in the `TestGetFilesystemsMinSizeTOML()` test cases. Also, please add a case when the size is specified as a string, which would reveal the c&p error that I found out in the TOML unmarshaling code.

thozza (commented):
Thanks a lot.

I found one issue in the unmarshaling code (which should be covered by the unit test, which I appreciate a lot!). Otherwise this looks great to me...

internal/blueprint/filesystem_customizations.go
kingsleyzissou: Whoops, thought I got all of them

internal/blueprint/filesystem_customizations.go
kingsleyzissou: Yeah I would agree with you, it seems very weird to be able to specify both... so maybe we should discourage this.
I will default to an error and if anyone feels strongly about it we can discuss.

internal/blueprint/filesystem_customizations_test.go
kingsleyzissou: Sure

internal/blueprint/filesystem_customizations.go
achilleas-k: I think making it an error is better than silently overriding one or the other.  It can create a lot of confusion.  An error message explaining that both can't be used and that one is an alias for the other is clearer and tells the user  everything they need to know.

internal/blueprint/filesystem_customizations.go
```diff
@@ -22,19 +23,57 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
...
+		return nil
+	}
+
+	if size == 0 && minsize > 0 {
+		fsc.MinSize = minsize
+		return nil
+	}
+
+	if size > 0 && minsize > 0 {
+		return fmt.Errorf("TOML unmarshal: size and minsize cannot both be set, got %v & %v", size, minsize)
```
achilleas-k, line 74: We could mention that one is an alias for the other or that the two options have the same meaning.  The current message might imply that they're in conflict but that they have different meaning (for example, one could think that `size` sets a strict exact size for the partition whereas `minsize` sets a flexible size with a lower bound).

Something like
```
size and minsize cannot be both set (size is an alias for minsize)
```
but feel free to rephrase.

achilleas-k (changes requested):

internal/blueprint/filesystem_customizations.go
kingsleyzissou: Good insight, wording seems good to me too :)

internal/blueprint/filesystem_customizations.go
```diff
@@ -9,7 +9,8 @@ import (
 
 type FilesystemCustomization struct {
 	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
-	MinSize    uint64 `json:"minsize,omitempty" toml:"size,omitempty"`
+	MinSize    uint64 `json:"minsize,omitempty" toml:"minsize,omitempty"`
+	Size       uint64 `toml:"size,omitempty"`
```
bcl, line 13: I'd add a comment to Size making it clear it should never be used outside UnmarshalTOML because it only there as an alias for MinSize. I can see someone (totally not me!) looking at the struct in 6 months and saying 'obviously Size is what I want to use' :)

internal/blueprint/filesystem_customizations.go
```diff
@@ -22,19 +23,57 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
...
+		}
+		minsize = s
+	case nil:
+		minsize = 0
+	default:
+		return fmt.Errorf("TOML unmarshal: minsize must be integer or string, got %v of type %T", d["minsize"], d["minsize"])
+	}
+
+	if size == 0 && minsize == 0 {
+		return fmt.Errorf("TOML unmarshal: minsize must be specified, got %v", minsize)
```
bcl, line 60: I think `must be > 0` is clearer than `must be specified`.

internal/blueprint/filesystem_customizations.go
thozza: I just realized (and successfully tested), that thanks to the poor implementation of custom unmarshalers in the TOML library that we use, we can omit the `Size` from the struct completely.

The thing is that the parsed data from the TOML are passed to the custom unmarshaler as `interface{}`, but in reality it is `map[string]interface{}`, thus it does not really relies on the actual struct members at all. The code of the custom unmarshaler can stay as is and handle the potential `size` key and assign it to `MinSize`.

It would be still good to add a comment to the custom unmarshaler, that the handling of `size` is kept there for backward compatibility.

kingsleyzissou: Yeah this is great, since we won't have to export another variable (which could lead to more confusion)

bcl (approved):
Looks good to me. I really like not needing to change the actual struct.

internal/blueprint/blueprint.go
```diff
@@ -301,7 +301,8 @@ func Convert(bp Blueprint) iblueprint.Blueprint {
 		if fs := c.Filesystem; fs != nil {
 			ifs := make([]iblueprint.FilesystemCustomization, len(fs))
 			for idx := range fs {
-				ifs[idx] = iblueprint.FilesystemCustomization(fs[idx])
+				ifs[idx].Mountpoint = fs[idx].Mountpoint
+				ifs[idx].MinSize = fs[idx].MinSize
```
thozza, line 305: nitpick: this should not be needed, since the struct didn't change in the end, right?

internal/blueprint/filesystem_customizations_test.go
```diff
@@ -0,0 +1,159 @@
...
+			`,
+			Want:  []FilesystemCustomization{{MinSize: 1024, Mountpoint: "/var"}},
+			Error: false,
+		},
+		{
+			Name: "minsize set (string), no size",
+			TOML: `
+[[customizations.filesystem]]
+mountpoint = "/var"
+size = "1KiB"
```
thozza, line 115: Most probably a Copy&Paste mistake here - you should be assigning to `minsize` instead.

thozza (changes requested):
Thanks for all the changes. I found one typo in the unit test which should be fixed. Otherwise, the PR looks great :+1:

thozza (approved):
LGTM, thanks a lot :rocket:

achilleas-k (approved):
:rocket:

PR #3799, opened 2023-11-15
pkg/splunk_logger: make it a module that can be imported seprately

diaasami:
Fixes COMPOSER-2051

This pull request includes:


pkg/splunk_logger/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/osbuild/osbuild-composer/pkg/splunk_logger
+
+go 1.20
```
bcl, line 3: Is it a good idea to have a different go version here? The main package is still on 1.19

diaasami: Good point, makes sense to stay on the same version.

achilleas-k (approved):
I don't know if it would be more convenient to separate this into a different repo, but I guess it's fine to have it here for now.

PR #3821, opened 2023-11-27
dnf-json: Allow passing module_hotfixes

ezr-ondrej:
Allow passing [module_hotfixes](https://dnf.readthedocs.io/en/latest/modularity.html#hotfix-repositories) as repo option,
it will allow disabling of modularity filtering per repository.

WHY
This would obviously be only the first step for enabling this.
I'm happy to continue pursuing this, but first I'd like to hear if we even want to do this.
I'd believe so, as modularity might go away, ppl might just want to force packages in without enabling modularity.

### DIFF
--- a/dnf-json
+++ b/dnf-json
@@ -119,6 +119,11 @@ class Solver():
         # we set the expiration to a short time period, rather than 0.
         repo.metadata_expire = desc.get("metadata_expire", "20s")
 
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
Now part of https://github.com/osbuild/osbuild-composer/pull/3831 closing this one.

PR #3831, opened 2023-12-01
cloudapi: Add module_hotfixes flag

ezr-ondrej:
Allow passing module_hotfixes flag through the cloudapi. This will enable depsolving on repositories that might be affected by modularity filtering.

Refs HMS-3202

Needs https://github.com/osbuild/images/pull/284 to have this flag on all the repositories.
:warning: currently this includes `osbuild/images` bump, which we want to wait for the release I'm guessing, but apart of that, this is ready for review :)


thozza:
I think that it would be good to enable this repo setting also for user-defined sources (repos) on-premise. This means extending:
https://github.com/osbuild/osbuild-composer/blob/main/internal/store/store.go#L59
https://github.com/osbuild/osbuild-composer/blob/main/internal/store/store.go#L622
https://github.com/osbuild/osbuild-composer/blob/main/internal/weldr/json.go#L138
https://github.com/osbuild/osbuild-composer/blob/main/internal/weldr/json.go#L203

Without this, one won't be able to define 3rd party repos with this option set, which are used for the actual image build. They could only embed repo definitions inside the final image with the option set (this is handled by the BP customization).

test/cases/api.sh
```diff
@@ -438,6 +438,36 @@ EOF
 )
 export FIREWALL_CUSTOMIZATION_BLOCK
 
+if [ "$TEST_MODULE_HOTFIXES" = 1 ]; then
+  if [ "$ARCH" = "x86_64"]; then
+    NGINX_REPO_URL="https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-x86_64-nginx-20231207"
+  else
+    NGINX_REPO_URL="https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-aarch64-nginx-20231207"
+  fi
```
thozza, line 446: nitpick:  You could also just use
```bash
NGINX_REPO_URL="https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-${ARCH}-nginx-20231207"
```

tools/test-case-generators/repos.json
```diff
@@ -274,6 +274,12 @@
         "image_type_tags": [
           "azure-eap7-rhui"
         ]
+      },
+      {
+        "name": "nginx",
+        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-x86_64-nginx-20231207",
+        "check_gpg": false,
+        "module_hotfixes": true
```
thozza, line 282: This should not be needed. for two reasons:

1. We do not generate any manifests in this repo any more, so these repo definitions are not used any more.
2. More importantly: the repo is not needed for any image manifest that we generate (or used to) generate for testing purposes.

tools/test-case-generators/repos.json
```diff
@@ -302,6 +308,12 @@
         "image_type_tags": [
           "azure-rhui"
         ]
+      },
+      {
+        "name": "nginx",
+        "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-aarch64-nginx-20231207",
+        "check_gpg": false,
+        "module_hotfixes": true
```
thozza, line 316: This should not be needed. for two reasons:

1. We do not generate any manifests in this repo any more, so these repo definitions are not used any more.
2. More importantly: the repo is not needed for any image manifest that we generate (or used to) generate for testing purposes.

test/cases/api.sh
```diff
@@ -582,9 +612,16 @@ function waitForImgState() {
...
 #
+
+if [ "$TEST_MODULE_HOTFIXES" != 1 ]; then
+  jq_edit = '.customizations.packages = [ "jesuisunpaquetquinexistepas" ]'
+else
+  jq_edit = 'del(.customizations.payload_repositories[] | select(.baseurl | match("*public/el8/el8-*-nginx-*")) | .module_hotfixes)'
+fi
 REQUEST_FILE2="${WORKDIR}/request2.json"
-jq '.customizations.packages = [ "jesuisunpaquetquinexistepas" ]' "$REQUEST_FILE" > "$REQUEST_FILE2"
+jq $jq_edit "$REQUEST_FILE" > "$REQUEST_FILE2"
```
thozza, line 624: I don't have a strong preference here, but you could have just add the `module_hotfixes` as an additional test case, instead of conditionally testing that or the failure when selecting non-existing package. Running both should be reasonably fast, since the compose will fail on depsolving and no image will be built.

test/cases/api.sh
```diff
@@ -438,6 +438,36 @@ EOF
 )
 export FIREWALL_CUSTOMIZATION_BLOCK
 
+if [ "$TEST_MODULE_HOTFIXES" = 1 ]; then
```
thozza, line 441: Maybe I'm blind, but I don't see the `TEST_MODULE_HOTFIXES` being set anywhere :thinking:

test/cases/api.sh
ezr-ondrej: I did, but the repo url validator complains xD

tools/test-case-generators/repos.json
ezr-ondrej: Ah, gotcha, I forgot I've added this, will remove it ;)

test/cases/api.sh
ezr-ondrej: You're right, I need to set it in the gitlab test config

ezr-ondrej: Done, but I dunno if correctly xD

internal/store/store.go
```diff
@@ -628,6 +629,7 @@ func (s *SourceConfig) RepoConfig(name string) rpmmd.RepoConfig {
 	repo.RHSM = s.RHSM
 	repo.CheckRepoGPG = common.ToPtr(s.CheckRepoGPG)
 	repo.GPGKeys = s.GPGKeys
+	repo.ModuleHotfixes = s.ModuleHotfixes
```
thozza, line 632: You need to make a similar change also in `NewSourceConfig()`

henrywang:
All FIPS tests failure has been fixed by PR #3849. Could you please make a rebase? Thanks.

henrywang:
All `edge-simplified` test failure is not caused by this PR. Issue has been fixed by PR #3854

test/cases/api/aws.sh
```diff
@@ -47,11 +47,11 @@ function createReqFile() {
     "payload_repositories": [
       {
         "baseurl": "$PAYLOAD_REPO_URL"
-      }
+      }${EXTRA_PAYLOAD_REPOS_BLOCK}
```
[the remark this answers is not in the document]
ezr-ondrej: Done :+1:

ezr-ondrej:
@thozza can you check out the vsphere failures if you see something in there? https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1109228638 I can't figure out what's the issue.
The good news is we seem to be quite close :stuck_out_tongue:

test/cases/api.sh
ezr-ondrej: Done :+1:

internal/store/store.go
```diff
@@ -583,10 +584,11 @@ func (s *Store) GetAllDistroSources(distro string) map[string]SourceConfig {
...
 	sc := SourceConfig{
-		Name:    repo.Name,
-		System:  system,
-		RHSM:    repo.RHSM,
-		GPGKeys: repo.GPGKeys,
+		Name:           repo.Name,
+		System:         system,
+		RHSM:           repo.RHSM,
+		GPGKeys:        repo.GPGKeys,
+		ModuleHotfixes: repo.ModuleHotfixes,
```
thozza, line 591: So, this change should have been added to https://github.com/osbuild/osbuild-composer/pull/3831/commits/33a430104b145100b8b89020d3043e26beb7a6d1 instead of https://github.com/osbuild/osbuild-composer/pull/3831/commits/aa6bc9c669b7e61d153bdf28b8b183d52775dcf3, but I won't insist on moving it. Please be more careful next time.

test/cases/api.sh
```diff
@@ -438,6 +439,39 @@ EOF
...
+        "rhsm": false,
+        "module_hotfixes": true
+      }
+EOF
+)
+
+  EXTRA_PACKAGES_BLOCK=$(cat <<EOF
+,
+      "nginx",
+      "nginx-module-njs"
```
thozza, line 463: I realized that it would be good to extend https://github.com/osbuild/osbuild-composer/blob/9b53d3e6a66ae8f2dd9845a160475a20f1617161/test/cases/api/common/common.sh#L18-L78 to check for these packages on the booted image, in case `[ "$TEST_MODULE_HOTFIXES" = "1" ]`

A follow-up is OK.

thozza (approved):
LGTM.

The nightly pipeline failures are unrelated to this PR.

test/cases/api.sh
ezr-ondrej: given I'm not sure how to do this, I'd rather work on it separatelly, so we can work on image builder integration in parallel

internal/store/store.go
ezr-ondrej: ugh, let me fix that.

test/cases/api.sh
thozza: It would be very similar to the code on line 23...

ezr-ondrej: gotcha, it will take few trials tho, so I'll open a follow-up for this: https://github.com/osbuild/osbuild-composer/pull/3863

internal/store/store.go
ezr-ondrej: Fixed

ezr-ondrej:
Thank you for guiding me @thozza ! :+1:

PR #3839, opened 2023-12-06
cloudapi/v2: fix verbiage of customizations

croissanne:
cloudapi/v2: tweak customizations
    
    Fix verbiage of groups customization, fields which accept an array
    should be plural.
    
    Remove the sshkey customization, sshkeys are merged into user
    customizations anyway, so users should use the "users" customization
    instead.
    
    Since these customizations aren't in use yet, this edit should be fine.
    
    See #3716


diaasami (dismissed):

thozza (approved):

achilleas-k:
This is the kind of customization I think we should also drop in the images lib.  Composer should do the conversion early, like it does here, and pass through only users to the `Manifest()`.

diaasami (approved):

PR #3857, opened 2023-12-18
add pulp test case: upload child commit to pulp server

yih-redhat:
added pulp child commit upload test case as bug https://github.com/pulp/pulp_ostree/issues/289 fixed.


henrywang:
@yih-redhat Does this PR have same issue as we found in downstream?

PR #3858, opened 2023-12-18
Adjust expected response based on osbuild-composer version

atodorov:
Will make cross-distro.sh test work on nightlies.

### DIFF
--- a/test/cases/cross-distro.sh
+++ b/test/cases/cross-distro.sh
@@ -6,6 +6,7 @@
 APISOCKET=/run/weldr/api.socket
 
 source /etc/os-release
+source /usr/libexec/tests/osbuild-composer/shared_lib.sh
 
 # Build a grep pattern that results in an empty string when the expected distros are installed
 case $ID in
@@ -146,7 +147,11 @@ EOF
 
     # there is a different reponse if legacy composer-cli is used
     if rpm -q --quiet weldr-client; then
-        EXPECTED_RESPONSE="ERROR: BlueprintsError: '$REMAINING_DISTRO' is not a valid distribution (architecture '$(uname -m)')"
+        if nvrGreaterOrEqual "osbuild-composer" "97"; then
+            EXPECTED_RESPONSE="ERROR: BlueprintsError: '$REMAINING_DISTRO' is not a valid distribution (architecture '$(uname -m)')"
+        else
+            EXPECTED_RESPONSE="ERROR: BlueprintsError: '$REMAINING_DISTRO' is not a valid distribution"
+        fi
     else
         EXPECTED_RESPONSE="'$REMAINING_DISTRO' is not a valid distribution"
         RESPONSE=${RESPONSE#*: }
--- a/test/cases/cross-distro.sh
+++ b/test/cases/cross-distro.sh
@@ -116,6 +116,14 @@ ALL_EXPECTED_DISTROS=$(echo "$ALL_DISTROS" | grep -E "$PATTERN" | grep -Ev 'beta
 # expected in the test snippet further below
 ALL_REMAINDERS=$(echo "$ALL_DISTROS" | grep -vw "$RECOGNIZED_DISTROS")
 
+echo "DEBUG: ===== ALL_DISTROS ===="
+echo "$ALL_DISTROS"
+echo "DEBUG: ===== ALL_EXPECTED_DISTROS ===="
+echo "$ALL_EXPECTED_DISTROS"
+echo "DEBUG: ===== ALL_REMAINDERS ===="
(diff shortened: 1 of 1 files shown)


jrusz (dismissed):

jrusz:
For some reason the test is failing everywhere, but I don't see why.

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD 75415d6ef488f3a2bb65143cf3511c0ea49931d0 with the main merge-base 3c3445c83212e8736eb67e6343e251f0099c6b12).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5776971970/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5776971970/artifacts/browse

schutzbot (commented):
⚠️ This PR introduces changes in at least one manifest (when comparing PR HEAD d847cf3c4da500a5546d6efb910a05553022c62f with the main merge-base 3c3445c83212e8736eb67e6343e251f0099c6b12).  Please review the changes.  The changes can be found in the [artifacts of the `Manifest-diff` job [0]](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5782144800/artifacts/browse) as `manifests.diff`.

[0] https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5782144800/artifacts/browse

jrusz:
I got curious and did some debugging and it seems that the offending line is:
```
 RESPONSE=$(sudo composer-cli blueprints push $TEST_BP 2>&1)
```
If I surround it with `set +e` and `set -e` the test seems to behave as expected. I guess this is because there are actually no `set` statements in the script but in the function `nvrGreaterOrEqual` there is `set -e` at the end and I guess that's why the first iteration works and the second one fails...

atodorov:
FTR from https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5782461753

```
INFO: Start interating over ALL_REMAINDERS
INFO: iterating over centos-8
DEBUG: centos-8, RESPONSE=ERROR: BlueprintsError: 'centos-8' is not a valid distribution
Blueprint push with centos-8 distro failed as expected.
INFO: iterating over centos-9
96 < 97
```

From the log messages it's clear that the script dies somehwhere between these lines, on the second iteration of the for loop:
https://github.com/osbuild/osbuild-composer/pull/3858/files#diff-e027796660fd9f47f7b2a38bf7234860367c7ab6a4daf21f6a83fb749cd121e7R145-R159

which makes no sense to me at all :-(.

jrusz:
> FTR from https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/5782461753
> 
> ```
> INFO: Start interating over ALL_REMAINDERS
> INFO: iterating over centos-8
> DEBUG: centos-8, RESPONSE=ERROR: BlueprintsError: 'centos-8' is not a valid distribution
> Blueprint push with centos-8 distro failed as expected.
> INFO: iterating over centos-9
> 96 < 97
> ```
> 
> From the log messages it's clear that the script dies somehwhere between these lines, on the second iteration of the for loop: https://github.com/osbuild/osbuild-composer/pull/3858/files#diff-e027796660fd9f47f7b2a38bf7234860367c7ab6a4daf21f6a83fb749cd121e7R145-R159
> 
> which makes no sense to me at all :-(.

See my comment above, I think that explains it.

atodorov:
> If I surround it with `set +e` and `set -e` the test seems to behave as expected. 

Thanks for finding this out. 

> I guess this is because there are actually no `set` statements in the script but in the function `nvrGreaterOrEqual` there is `set -e` at the end and I guess that's why the first iteration works and the second one fails...

There's also a `set +e` inside the `nvrGreaterOrEqual` function so the trailing `set -e` is supposed to be doing its job. The answer here is not that we are missing an opening `set +e` statement.

Quick google search reveals that [calling set +/-e in this way may be unreliable](https://stackoverflow.com/questions/44014857/function-invocation-in-for-loop-does-not-fail-even-with-set-e)! I will redo the if condition and move it outside the loop because we don't need to evaluate it every single time. And I'm not really in the mood to debug the intricacies of Bash and find out when it works and when it doesn't!

jrusz (changes requested):
Can you remove the debug commits? I don't think they need to be kept here.

jrusz (approved):

PR #3869, opened 2023-12-22
Test: delete "diff manifests" test case

thozza:
This is not adding any value any more, since all image definitions are defined in the osbuild/images repository and just vendored in this one.

This pull request includes:


jrusz (commented):
So, since composer should not introduce changes to manifests I wonder what was even the original motivation for this test case? 

Is some similar test running in the Images repository?

achilleas-k:
I wasn't sure if the diff test was providing any value after all, but let's add it there now and see if we can make it more useful after.

achilleas-k (approved):

jrusz (approved):

PR #3878, opened 2024-01-02
test: weldr api module_hotfixes flag

ezr-ondrej:
Adds testing for module_hotfixes on weldr API

### DIFF
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -13,6 +13,7 @@ package client
 import (
 	"fmt"
 	"os/exec"
+	"runtime"
 	"sort"
 	"strconv"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/client/blueprints_test.go
```diff
@@ -1006,6 +1006,35 @@ func TestBlueprintDepsolveGlobsV0(t *testing.T) {
...
+		t.Skip()
+	}
+	bp := `{
+		"name": "test-deps-blueprint-module-hotfixes-v0",
+		"description": "CheckBlueprintDepsolveModuleHotfixesV0",
+		"version": "0.0.1",
+		"packages": [{"name": "nginx", "version": "*"},
+		{"name": "nginx-module-njs", "version": "*"}]
+		"customizations": {"repositories": [{
+			"id": "nginx", "name": "nginx", "baseurls": ["https://rpmrepo.osbuild.org/v2/mirror/public/el8/el8-x86_64-nginx-20231207"], "gpgcheck"": false}
```
bcl, line 1022: You should set module_hotfixes to false to make it clear what's happening here.
Also, shouldn't you also test it with it set to true?

ezr-ondrej: oh, sorry it was not really ready, I supposed to have drafted it :)
Now it's what I believe we should add, but it still doesn't work locally, I think it uses a my local dnfjson maybe? :)

ezr-ondrej: Well, I've pushed next rebase, let see what happens :shrug:
If it stays red, I'm out of ideas :thinking:

ezr-ondrej:
It's still very red :( xD

PR #3886, opened 2024-01-09
aws: allow the user to customize their AMI name

lavocatt:
In the frontend a new field is allowing user to setup customized names
for their images (see
https://github.com/RedHatInsights/image-builder-frontend/pull/1136).

To allow the customization to effectively take place, this commit is
slightly changing how EC2 images are activated.

The Name tag and the AMI Name are not always sharing the same value
exactly anymore. If the user provides a SnaphsotName, the raw value is
set in the `Name` tag and the AMI name is set to `$SnaphsotName-$UUID`
(to make it unique).

This means that the `Name` tag can't be used anymore for the maintenance
tooling. The tooling will have to use the new Build-By tag to identify
the images built by osbuild-composer.

**Example**
The request:
image

Ended up producing this AMI:
image

This pull request includes:

https://issues.redhat.com/browse/HMS-2959

### DIFF
--- a/internal/cloud/awscloud/awscloud.go
+++ b/internal/cloud/awscloud/awscloud.go
@@ -212,7 +212,7 @@ func WaitUntilImportSnapshotTaskCompletedWithContext(c *ec2.EC2, ctx aws.Context
 // The caller can optionally specify the boot mode of the AMI. If the boot
 // mode is not specified, then the instances launched from this AMI use the
 // default boot mode value of the instance type.
-func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch string, bootMode *string) (*string, error) {
+func (a *AWS) Register(imageName, bucket, s3key, tagName string, shareWith []string, rpmArch string, bootMode *string) (*string, error) {
 	rpmArchToEC2Arch := map[string]string{
 		"x86_64":  "x86_64",
 		"aarch64": "arm64",
@@ -229,15 +229,18 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 		}
 	}
 
-	logrus.Infof("[AWS] 📥 Importing snapshot from image: %s/%s", bucket, key)
-	snapshotDescription := fmt.Sprintf("Image Builder AWS Import of %s", name)
+	// Note that the key value is only used in the upload transition phase in S3. It's deleted once the snapshot is done
+	// importing. So it's content has not much impact. The key is not conveyed anymore, compute a unique one based on
+	// the nameTag's value.
+	logrus.Infof("[AWS] 📥 Importing snapshot from image: %s/%s", bucket, s3key)
+	snapshotDescription := fmt.Sprintf("Image Builder AWS Import of %s", imageName)
 	importTaskOutput, err := a.ec2.ImportSnapshot(
 		&ec2.ImportSnapshotInput{
 			Description: aws.String(snapshotDescription),
 			DiskContainer: &ec2.SnapshotDiskContainer{
 				UserBucket: &ec2.UserBucket{
 					S3Bucket: aws.String(bucket),
-					S3Key:    aws.String(key),
+					S3Key:    aws.String(s3key),
 				},
 			},
 		},
@@ -261,10 +264,10 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 	}
 
 	// we no longer need the object in s3, let's just delete it
-	logrus.Infof("[AWS] 🧹 Deleting image from S3: %s/%s", bucket, key)
+	logrus.Infof("[AWS] 🧹 Deleting image from S3: %s/%s", bucket, s3key)
 	_, err = a.s3.DeleteObject(&s3.DeleteObjectInput{
 		Bucket: aws.String(bucket),
-		Key:    aws.String(key),
+		Key:    aws.String(s3key),
 	})
 	if err != nil {
 		return nil, err
@@ -290,7 +293,11 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 			Tags: []*ec2.Tag{
 				{
 					Key:   aws.String("Name"),
-					Value: aws.String(name),
+					Value: aws.String(tagName),
+				},
+				{
+					Key:   aws.String("Build-by"),
+					Value: aws.String("osbuild-composer"),
 				},
 			},
 		},
@@ -306,7 +313,7 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 			Architecture:       aws.String(ec2Arch),
 			BootMode:           bootMode,
 			VirtualizationType: aws.String("hvm"),
-			Name:               aws.String(name),
+			Name:               aws.String(imageName),
 			RootDeviceName:     aws.String("/dev/sda1"),
 			EnaSupport:         aws.Bool(true),
 			BlockDeviceMappings: []*ec2.BlockDeviceMapping{
@@ -332,7 +339,11 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 			Tags: []*ec2.Tag{
 				{
 					Key:   aws.String("Name"),
-					Value: aws.String(name),
+					Value: aws.String(tagName),
+				},
+				{
+					Key:   aws.String("Build-by"),
+					Value: aws.String("osbuild-composer"),
 				},
 			},
 		},
--- a/internal/cloudapi/v2/imagerequest.go
+++ b/internal/cloudapi/v2/imagerequest.go
@@ -47,11 +47,8 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
 		return nil, HTTPError(ErrorJSONUnMarshallingError)
 	}
 
-	// For service maintenance, images are discovered by the "Name:composer-api-*"
-	// tag filter. Currently all image names in the service are generated, so they're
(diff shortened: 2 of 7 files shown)


lavocatt:
What kind of test would be adequate for this functionality do you think? I could change the aws test to add a snapshotName in the request and see if I can get access to the proper name after upload? Wdyt?

internal/cloudapi/v2/imagerequest.go
```diff
@@ -47,10 +47,8 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
 		return nil, HTTPError(ErrorJSONUnMarshallingError)
 	}
 
-	// For service maintenance, images are discovered by the "Name:composer-api-*"
-	// tag filter. Currently all image names in the service are generated, so they're
-	// guaranteed to be unique as well. If users are ever allowed to name their images,
-	// an extra tag should be added.
+	// For service maintenance, images are discovered by the "Name:composer-api-*" tag filter. The `Name` tag differs
+	// from the AMI Name. The later being user customizable.
```
achilleas-k, line 51: The previous version of this comment said that if we ever allow users to name their images, we should add an extra tag.  The way I read it is that we should tag images with something composer-specific for maintenance and set the name to what the user wanted.  @croissanne?  (I think you wrote the original comment).

I think that would make more sense to set the name to what the user provided, but since you mentioned the tags aren't shared, and the "Name" field (in your screenshot) is controlled by the Name tag (is that right?), then maybe it's fine to just name the AMI and leave the rest the same.

bcl:
What benefit is there in using a new UUID to make the AMI Name unique? Why not use the UUID of the image so that no matter which field you look at it is clear (ha!) that it is the same thing?

lavocatt:
> What benefit is there in using a new UUID to make the AMI Name unique? Why not use the UUID of the image so that no matter which field you look at it is clear (ha!) that it is the same thing?

Thanks for this input, I haven't thought about it, I'll make the two UUIDs the same.

internal/cloud/awscloud/awscloud.go
```diff
@@ -283,14 +283,14 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	snapshotID := importOutput.ImportSnapshotTasks[0].SnapshotTaskDetail.SnapshotId
 
-	// Tag the snapshot with the image name.
+	// Tag the snapshot with the image name and the key.
```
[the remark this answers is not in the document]
lavocatt: Thanks for pointing out this comment, I need to revert it as I'm only sending the name tag anymore.

lavocatt:
The tests are failing and I'm starting to understand what's happening.

We decided to reuse the `snapshot_name` because we thought that this field in the request wasn't really used anywhere. But it turns out that this field is being used in the tests:
https://github.com/osbuild/osbuild-composer/blob/6735e74897d2eaed771f43b35ed33b257aca29e5/test/cases/api/aws.sh#L69C21-L79

https://github.com/osbuild/osbuild-composer/blob/6735e74897d2eaed771f43b35ed33b257aca29e5/test/cases/api/aws.sh#L103-L110

The test suite it using this field to find out what is the image_id of the image it has just built. But since now we're making the `snapshot_name` unique by appending to it some UUID to it, we're breaking the way the test suite had to recover the image.

I can go for an easy fix right now, which is to simply change how the test suite is finding the image by updating the filter from `being equal` to `containing`. For that I'm just adding an asterix to the filter.

But this is opening a new question:

Since the `snapshot_name` was originally thought to be used to create a link between the image and its snapshot with a customizable name, shouldn't we keep that behavior intact?

I'm proposing to:
* use the `snapshot_name` to set a `SnapshotName` tag in the AMI. (only when a user wanted to customize that)
* set the `Name` tag to the generated value `composer-api-*`
* set the AMI name to the generated value `composer-api-*`  if no further customization is asked from the user
* use a new field `custom_ami_name` to set the AMI name

What do you think is best here?

internal/cloudapi/v2/imagerequest.go
croissanne: Maybe explicitly separating them makes sense for clarity, but there's no functional difference.

 @achilleas-k when I wrote the comment I didn't realise names are inferred from the tags, and thus aren't shared. So we're stuck with setting the ami name (which has to be unique).

internal/cloud/awscloud/awscloud.go
```diff
@@ -290,7 +290,7 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 			Tags: []*ec2.Tag{
 				{
 					Key:   aws.String("Name"),
-					Value: aws.String(name),
+					Value: aws.String(key),
```
croissanne, line 293: Hm so this does kinda break on-prem behaviour, since the key set in the target file now becomes the ami name (instead of the 'name' set in the target file). Maybe having a 2nd tag makes sense after all.

bcl (commented):
The consistent UUID part of things looks good.

internal/cloud/awscloud/awscloud.go
```diff
@@ -301,12 +306,13 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
...
 
 	logrus.Infof("[AWS] 📋 Registering AMI from imported snapshot: %s", *snapshotID)
+	uniqueName := fmt.Sprintf("%s-%s", name, uuid.New().String())
 	registerOutput, err := a.ec2.RegisterImage(
 		&ec2.RegisterImageInput{
 			Architecture:       aws.String(ec2Arch),
 			BootMode:           bootMode,
 			VirtualizationType: aws.String("hvm"),
-			Name:               aws.String(name),
+			Name:               aws.String(uniqueName),
```
lavocatt: This change is breaking onPrem.
OnPrem needs to have the `ImageName` set in AMI Name (without any UUID appended to it)

internal/cloudapi/v2/imagerequest.go
```diff
@@ -69,10 +67,14 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
...
 	})
+	// There's two possibilities for the ImageName: The user have provided a customization or not. In both cases we're
+	// going to need to append a UUID to make the name unique. Otherwise AWS will reject the image registration. The
+	// UUID is appended in the Register function so that the raw value of the name can be used for
+	// other purpose (Tags for instance).
 	if awsUploadOptions.SnapshotName != nil {
 		t.ImageName = *awsUploadOptions.SnapshotName
 	} else {
-		t.ImageName = key
+		t.ImageName = "composer-api"
```
lavocatt: There's only two fields that I can use to pass down data to the `Register` function. `Key` and `ImageName`.

I've got a problem with making the ImageName unique:
If I append the uuid in this function, the Name tag and the AMI name, both have it, it breaks onPrem.
If I append the uuid in the `Register` function, it breaks onPrem.

I'll try to repurpose the `Key` field to use it as a mean to transmit a tagName. Since in the registration process the key is only used for the S3 upload which gets deleted after the image got registered.

lavocatt: Repurposing was a bad idea, instead I'm adding a new optional `TagName` field in the `AWSTargetOptions`.
Then if it's provided, the `Register` function will use it, if not it'll default to the `ImageName`.
That should hopefully work.

internal/cloud/awscloud/awscloud.go
```diff
@@ -212,7 +212,7 @@ func WaitUntilImportSnapshotTaskCompletedWithContext(c *ec2.EC2, ctx aws.Context
 // The caller can optionally specify the boot mode of the AMI. If the boot
 // mode is not specified, then the instances launched from this AMI use the
 // default boot mode value of the instance type.
-func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch string, bootMode *string) (*string, error) {
+func (a *AWS) Register(imageName string, bucket string, s3key string, tagName string, shareWith []string, rpmArch string, bootMode *string) (*string, error) {
```
thozza, line 215: nitpick: there's no need to add type to each argument. This makes the line just longer...

internal/cloud/awscloud/awscloud.go
```diff
@@ -332,7 +339,11 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 			Tags: []*ec2.Tag{
 				{
 					Key:   aws.String("Name"),
-					Value: aws.String(name),
+					Value: aws.String(tagName),
+				},
+				{
+					Key:   aws.String("Build-by"),
+					Value: aws.String("osbuild-composer"),
```
thozza, line 346: I'm wondering if the maintenance service could not leverage this new tag instead of the image name for filtering? This would allow us to keep the image name and the value of the `Name` tag consistent. Since you are appending an UUID to the user-provided image name, it will be unique and usable as the single image name.

@croissanne WDYT about using this new tag for the maintenance service?

thozza (commented):
The change looks reasonable, but I'm wondering if we can't instead extend the maintenance service to use additional / different tag for filtering images? This would allow us to use the same value for the `Name` tag and as image name and would reduce the complexity and required changes in osbuild-composer a lot. :thinking:

croissanne:
> The change looks reasonable, but I'm wondering if we can't instead extend the maintenance service to use additional / different tag for filtering images? This would allow us to use the same value for the `Name` tag and as image name and would reduce the complexity and required changes in osbuild-composer a lot. 🤔


I think that's already implied in the description. Might be useful to do this in the same PR though, would just require a tiny change.

PR #3887, opened 2024-01-09
Use dot to separate distro major and minor version and replace distro registry with factory

thozza:
Related to: https://issues.redhat.com/browse/COMPOSER-1962

Main changes:

- Delete commands which are not relevant to osbuild-composer after the images repo split. Removing them meant no need to port them to dot-notation. These are:
  - `cmd/osbuild-playground`
  - `cmd/osbuild-package-sets`
  - `cmd/osbuild-pipeline`
  - `cmd/osbuild-composer-image-definitions`
  - tools: delete old Python manifest generator scripts, since the `gen-manifests` is the tool to use and that will be removed from this repo eventually as well.
- Port `osbuild/images` `v0.33.0` to osbuild-composer. This is unfortunately one giant commit, because there is no other good way to ensure that composer compiles and unit tests pass after the update of the `osbuild/images` package. It includes:
  - Replace all uses of distroregistry by distrofactory.
   - Delete local version of reporegistry and use the one from the osbuild/images.
   - Weldr: unify `createWeldrAPI()` and `createWeldrAPI2()` into a single `createTestWeldrAPI()` function`.
   - store/fixture: rework fixtures to allow overriding the host distro name and host architecture name. A cleanup function to restore the host distro and arch names is always part of the fixture struct.
   - Delete `distro_mock` package, since it is no longer used.
   - Bump the required version of osbuild to 98, because the OSCAP customization is using the 'compress_results' stage option, which is not available in older versions of osbuild.
- Drop implementations of functions in composer, which are provided by `osbuild/images`. Specifically, remove:
  - `common.GetHostDistroName()`
  - `common.CurrentArch()`
- Extend osbuild-composer configuration to allow specifying distro name aliases. This is useful to define e.g. that `rhel-8` actually translates to `rhel-8.10`, etc. The default configuration has these aliases defined for the next RHEL GA versions (the on-prem scenario). The service can override these to point to the current GA in its deployment.
- Rename all repo configurations (testing and those shipped in the regular RPM) to use dot to separate major and minor version. Modify all test cases to use the "dot notation".
- Update testing repo configs to:
  - Use CDN repos for RHEL 8.9 and 9.3, which are already GA.
  - Add SAP repos to all RHEL repo configs, to make it possible to build SAP images as part of test cases.
- Add new test for Weldr API and extend the `api.sh` test case to enable testing of dot-notation, backward compatibility (not using dots) and distro aliases. The distro alias test always uses SAP image and inspects `/etc/dnf/vars/releasever` in it. This image type is the only which contains is and the content of the file is always the release version of the distro definition used to generate the image manifest (in the form of "X.Y"). Otherwise, there would not be any other way to verify which distro definition was used to generate the manifest used to build the image. The content of `/etc/os-release` always comes from the provided repositories and it is not anyhow related to the actual distro definition release version used to generate the manifest.
- Make a few modifications, which are a result of the port and improve various things noticed along the way. Specifically:
  - Test/old-worker-new-composer: don't wait for composer indefinitely. It happened to me that if the container didn't start, the test would loop and wait for it for 5 hours in CI.
  - cloudapi: pass depsolve job error details to the manifest job error
  - worker/osbuild: provide more details and logs when osbuild build fails. Previously, any manifest validation errors would be thrown away and not logged or sent to composer. This issue made debugging such issues extremely hard, since the only message in the worker log would be `osbuild build failed`. This is not the case any more and we get a lot more details about the actual problem.

This pull request includes:

  - [x] Unit test are modified or added where it made sense
  - [x] Functional tests for both Cloud API and Weldr API testing the dot-notation, backward compatibility and distro aliases are added.
  - **TBD**


cmd/gen-manifests/main.go
```diff
@@ -208,6 +208,14 @@ func makeManifestJob(name string, imgType distro.ImageType, cr composeRequest, d
 
 type DistroArchRepoMap map[string]map[string][]repository
 
+func (darm DistroArchRepoMap) ListDistros() []string {
+	distros := make([]string, 0, len(darm))
+	for d := range darm {
+		distros = append(distros, d)
+	}
+	return distros
```
bcl, line 216: Sorting these would make it more consistent :)

cmd/osbuild-composer/config_test.go
```diff
@@ -90,6 +98,13 @@ func TestConfig(t *testing.T) {
 	require.Equal(t, []string{"https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs"}, config.Koji.JWTKeysURLs)
 	require.Equal(t, "", config.Koji.JWTKeysCA)
 	require.Equal(t, "/var/lib/osbuild-composer/acl", config.Koji.JWTACLFile)
+
+	expectedDistroAliases := map[string]string{
+		"rhel-7": "rhel-7.9", // this value is from the default config
+		"rhel-8": "rhel-8.9",
```
bcl, line 104: These don't match what's above. And I wonder if it would be possible to set the alias to the latest of each X release automatically? Or is there ever a reason for the alias to lag behind?

cmd/gen-manifests/main.go
thozza: I'll do it, but I don't see why it would matter in this case.

cmd/osbuild-composer/config_test.go
thozza: One note to the question of if we can set the tag to the latest release automatically:
Previously, this was hard coded in the distro definition. There is no longer any notion of what is the latest release in the distro / image definitions. The idea is that distro versions with defined repositories are treated as tested and supported. We could probably check the available repositories for the on-prem and set the aliases to the latest release, but not for the service. In Cloud API, the list of repositories is provided with the compose request. Moreover, as mentioned above, there is a need to have different target defined for these aliases on-prem and in the service deployment.

cmd/gen-manifests/main.go
bcl: Oh ha. I now notice this in in main.go not in distrofactory, sorry about that :)

henrywang:
The `edge-installer-f38` test failed because ostree name changed from `fedora` to `fedora-iot`. Anything changed in this PR caused this failure? Thanks.

thozza:
> The `edge-installer-f38` test failed because ostree name changed from `fedora` to `fedora-iot`. Anything changed in this PR caused this failure? Thanks.

Not by my changes, but by the latest version of `osbuild/images`. Specifically https://github.com/osbuild/images/commit/3776160b1fb9b92f4717a2b027541bcc0e59298e @paulwhalen should have more info on why this changed.

henrywang:
@thozza Thanks for the info. I'll push the fix into this PR.

thozza:
> @thozza Thanks for the info. I'll push the fix into this PR.

@henrywang thanks, but I think that it would be better if you could just point me to the right place to make the change. This PR is still in heavy development and I'm regularly force pushing it. And I don't want to drop your changes by accident... :innocent:

henrywang:
@thozza Sure. The change is simple.
``` diff
diff --git i/test/cases/ostree-ng.sh w/test/cases/ostree-ng.sh
index 4a87bc137..27fc3d436 100755
--- i/test/cases/ostree-ng.sh
+++ w/test/cases/ostree-ng.sh
@@ -125,7 +125,7 @@ case "${ID}-${VERSION_ID}" in
         CONTAINER_TYPE=iot-container
         INSTALLER_TYPE=iot-installer
         OSTREE_REF="fedora/${VERSION_ID}/${ARCH}/iot"
-        OSTREE_OSNAME=fedora
+        OSTREE_OSNAME=fedora-iot
         OS_VARIANT="fedora-unknown"
         EMBEDED_CONTAINER="false"
         DIRS_FILES_CUSTOMIZATION="true"
```

internal/weldr/api.go
```diff
@@ -3559,17 +3564,16 @@ func (api *API) payloadRepositories(distroName string) []rpmmd.RepoConfig {
...
 // which are needed to build the specific image type. The allRepositories() can't do this, because
 // it is used in places where image types are not considered.
-func (api *API) allRepositoriesByImageType(imageType distro.ImageType) ([]rpmmd.RepoConfig, error) {
-	repos, err := api.repoRegistry.ReposByImageTypeName(
-		imageType.Arch().Distro().Name(),
-		imageType.Arch().Name(),
-		imageType.Name(),
-	)
+//
+// Note that the distro name is not determined frim the image type, but must be provided as an argument.
```
achilleas-k, line 3568: Typo: `frim`

achilleas-k (commented):
This is an incredible amount of work and it all looks great.  Amazing work!

thozza:
So, the Nightly pipeline looks "good", because the only failure is on Azure, which is caused by https://github.com/osbuild/images/issues/398. That will be tackled separately and I consider it out of scope. So that won't turn greed on this PR.

I'll fix up some errors in the regular pipeline.

internal/weldr/api.go
thozza: Fixed.

achilleas-k (dismissed):
Let's go!!

achilleas-k (approved):
Fantastic!  I can't believe we're finally getting so many dots ............
:smiling_face_with_tear:

PR #3906, opened 2024-01-22
deps: updates osbuild/images to v0.31.0

achilleas-k:
The images dependency update needs to be done manually because of breaking changes in the API.  Updating here without introducing any functional changes to unblock dependency updates.

The Container struct in the Blueprint was expanded as part of local containers-storage feature [1].  Adding and testing support for this feature will require more work.  For now, let's explicitly ignore the fields during the Blueprint copying so we can update the images dependency.

The Pipeline constructor signatures in osbuild/images were changed as part of a recent PR [2]. Updating the examples in osbuild-composer to match.

[1] https://github.com/osbuild/images/pull/306
[2] https://github.com/osbuild/images/pull/346


achilleas-k:
Of course, we'll have to add the build flags to disable the container storage backends, or install the dependencies for the unit tests.  Not sure what the linter is complaining about though.

thozza:
Just FYI, all of the issues should be resolved in #3887

achilleas-k:
> Just FYI, all of the issues should be resolved in #3887

I knew most of it was covered there (even the Container customization?) which is why I pinged on this.  I though we might need to unblock dependency updates while that's still being worked on.  I'm fine with leaving it up to your PR though.  There's no real rush to update deps right now.

thozza:
> > Just FYI, all of the issues should be resolved in #3887
> 
> I knew most of it was covered there (even the Container customization?) which is why I pinged you specifically on this. I though we might need to unblock dependency updates while that's still being worked on. I'm fine with leaving it up to your PR though. There's no real rush to update deps right now.

Right.

I'm currently reworking the distro alias test cases in my PR to use the SAP image with the pinned down release in DNF config. Otherwise it should be good for review. My expectation is 1-2 more days. I have been rebasing on top of `osbuild/images` `main` regularly, so the PR is based on almost the latest version of `images`.

So from my PoV, I would prefer get #3887 reviewed once CI passes, merge the `images` PR, make a release and then update #3887 to the new `images` release. :innocent:

thozza:
And yeah, the container customization is updated in my PR. But only to make everything work / compile.

achilleas-k:
> So from my PoV, I would prefer to get #3887 reviewed once CI passes, merge the `images` PR, make a release and then update #3887 to the new `images` release. 😇

SGTM.  Closing.

PR #3907, opened 2024-01-22
composer: glitchtip integration

diaasami:
Initial glitchtip/sentry integration, covers panics, a subsequent PR logrus integration will include severe errors and HTTP 500s as well.

https://issues.redhat.com/browse/HMS-1986

### DIFF
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/coreos/go-systemd/activation"
+	"github.com/getsentry/sentry-go"
 	slogger "github.com/osbuild/osbuild-composer/pkg/splunk_logger"
 	"github.com/sirupsen/logrus"
 )
@@ -68,6 +69,17 @@ func main() {
 			panic(err)
 		}
 		logrus.AddHook(hook)
+	}
+
+	if config.GlitchTipDSN != "" {
+		err = sentry.Init(sentry.ClientOptions{
+			Dsn: config.GlitchTipDSN,
+		})
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		logrus.Warn("GLITCHTIP_DSN not configured, skipping initializing Sentry/Glitchtip")
 	}
 
 	stateDir, ok := os.LookupEnv("STATE_DIRECTORY")
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -12,6 +12,8 @@ import (
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/getkin/kin-openapi/routers"
 	legacyrouter "github.com/getkin/kin-openapi/routers/legacy"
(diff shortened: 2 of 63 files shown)


cmd/osbuild-composer/main.go
```diff
@@ -70,6 +71,17 @@ func main() {
...
 
+	if config.GlitchTipDSN != "" {
+		err = sentry.Init(sentry.ClientOptions{
+			Dsn: config.GlitchTipDSN,
+		})
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		logrus.Warn("Sentry/Glitchtip was not initialized")
```
ondrejbudai, line 82: Maybe something like "GLITCHTIP_DSN not configured, skipping initializing Sentry/Glitchtip"? Your variant sounds like composer attempted to initialize it, but wasn't successful in doing so.

internal/cloudapi/v2/server.go
```diff
@@ -87,6 +89,12 @@ func (s *Server) Handler(path string) http.Handler {
 	e.Use(middleware.Recover())
 	e.Logger = common.Logger()
 
+	if sentry.CurrentHub().Client() == nil {
+		logrus.Warn("Sentry/Glitchtip not initialized, echo middleware was not enabled")
```
ondrejbudai, line 93: I think I would apply a similar suggestion to my previous comment.

templates/composer.yml
```diff
@@ -118,6 +118,12 @@ objects:
                 optional: true
           - name: SPLUNK_HEC_PORT
             value: "${SPLUNK_HEC_PORT}"
+          - name: GLITCHTIP_DSN
+            valueFrom:
+              secretKeyRef:
+                key: dsn
```
ondrejbudai, line 124: `glitchtip_dsn` for consistency with the env variable name? `dsn` isn't super-descriptive.

internal/cloudapi/v2/server.go
diaasami: I agree about the other one, but this one makes sense since it checks if sentry itself, but I agree it can be improved, may be "Sentry/Glitchtip not initialized, skipping echo middleware".

templates/composer.yml
diaasami: this is the secret imported from the namespace, so we do not have control over the key name, but I can add comments to explain DSN.

henrywang:
All RHEL 8 failure has been fixed by PR #3933. Could you please make a rebase? Thanks!

internal/cloudapi/v2/server.go
ondrejbudai: Right, makes sense, I misread the code initially.

templates/composer.yml
ondrejbudai: Ahhh, makes sense. :) Thanks for the explanation.

ondrejbudai:
rebased

ondrejbudai (approved):
Thanks!

PR #3910, opened 2024-01-23
tools/fedora-worker-packer: fix packer only/except

croissanne:
It should only build fedora, not everything except fedora.


thozza (approved):
Ups... :slightly_smiling_face:

PR #3920, opened 2024-01-30
templates/packer: deal with unbound variables

croissanne:
Don't allow unbound variables, but for the variables that are used to determine whether or not that part of the setup should continue, default to false.


ochosi (approved):
Looks ok!

PR #3927, opened 2024-02-02
Support configuring distro aliases via ENV and set aliases in the composer template.

thozza:
SSIA, check commit messages for more details.

Also added unit tests for the functionality.

### DIFF
--- a/cmd/osbuild-composer/config.go
+++ b/cmd/osbuild-composer/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/BurntSushi/toml"
 )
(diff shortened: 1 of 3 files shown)


thozza:
@croissanne I'm not 100% sure that my approach with the template is correct and sufficient. Please let me know if I need to do it differently or if I missed some place that required modification. Thanks!

cmd/osbuild-composer/config.go
```diff
@@ -14,7 +15,7 @@ type ComposerConfigFile struct {
 	Koji          KojiAPIConfig     `toml:"koji"`
 	Worker        WorkerAPIConfig   `toml:"worker"`
 	WeldrAPI      WeldrAPIConfig    `toml:"weldr_api"`
-	DistroAliases map[string]string `toml:"distro_aliases"`
+	DistroAliases map[string]string `toml:"distro_aliases" env:"DISTRO_ALIASES"`
```
bcl, line 18: This should have an `OSBUILD_` prefix. Just in case.

thozza: I considered it, but then saw that the other configuration options also do not use any prefix. :shrug:

bcl: Odd. When I added `OSBUILD_LOCALSAVE` that was one of the changes folks requested. In practice it likely doesn't really matter.

croissanne (approved):
thank you!

ondrejbudai (approved):

PR #3928, opened 2024-02-03
Use distro repos in cloudapi

bcl:
This pull request includes:

### DIFF
--- a/internal/cloudapi/v2/compose.go
+++ b/internal/cloudapi/v2/compose.go
@@ -4,16 +4,17 @@ package v2
 import (
 	"crypto/rand"
 	"fmt"
-	"github.com/osbuild/images/pkg/distrofactory"
-	"github.com/osbuild/images/pkg/rhsm/facts"
-	"github.com/osbuild/osbuild-composer/internal/target"
 	"math"
 	"math/big"
 	"reflect"
 
 	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/distrofactory"
+	"github.com/osbuild/images/pkg/reporegistry"
+	"github.com/osbuild/images/pkg/rhsm/facts"
 	"github.com/osbuild/images/pkg/subscription"
 	"github.com/osbuild/osbuild-composer/internal/blueprint"
+	"github.com/osbuild/osbuild-composer/internal/target"
 )
 
 // Return the string representation of the partitioning mode
@@ -944,7 +945,7 @@ func (request *ComposeRequest) GetPartitioningMode() (disk.PartitioningMode, err
 
 // GetImageRequests converts a composeRequest structure from the API to an intermediate imageRequest structure
 // that's used for generating manifests and orchestrating worker jobs.
-func (request *ComposeRequest) GetImageRequests(distroFactory *distrofactory.Factory) ([]imageRequest, error) {
+func (request *ComposeRequest) GetImageRequests(distroFactory *distrofactory.Factory, repoRegistry *reporegistry.RepoRegistry) ([]imageRequest, error) {
 	distribution := distroFactory.GetDistro(request.Distribution)
 	if distribution == nil {
 		return nil, HTTPError(ErrorUnsupportedDistribution)
@@ -1000,6 +1001,15 @@ func (request *ComposeRequest) GetImageRequests(distroFactory *distrofactory.Fac
 		repos, err := convertRepos(ir.Repositories, payloadRepositories, imageType.PayloadPackageSets())
 		if err != nil {
 			return nil, err
+		}
+
+		// If no repositories are included with the imageRequest use the defaults for the distro
+		if len(ir.Repositories) == 0 {
+			dr, err := repoRegistry.ReposByImageTypeName(request.Distribution, arch.Name(), imageType.Name())
+			if err != nil {
+				return nil, err
+			}
+			repos = append(repos, dr...)
 		}
 
 		// Get the initial ImageOptions with image size set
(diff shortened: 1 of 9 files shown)


thozza (commented):
Overall, this looks good. :+1:  I added one question, but nothing serious.

internal/cloudapi/v2/handler.go
```diff
@@ -212,6 +212,15 @@ func (h *apiHandlers) PostCompose(ctx echo.Context) error {
...
 		}
 
+		// If no repositories are included with the imageRequest use the defaults for the distro
+		if len(ir.Repositories) == 0 {
+			dr, err := h.server.repos.ReposByImageTypeName(distribution.Name(), arch.Name(), imageType.Name())
+			if err != nil {
+				return err
+			}
+			repos = append(repos, dr...)
+		}
```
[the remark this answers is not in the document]
bcl: I think this is ok as is, an empty repositories list is still required (without it openapi returns an error because of the missing keyword) so it's pretty clear what the intention is. And I hate adding more switches :)

internal/cloudapi/v2/v2_test.go
```diff
@@ -34,11 +35,16 @@ func newV2Server(t *testing.T, dir string, depsolveChannels []string, enableJWT
 	distros := distrofactory.NewTestDefault()
 	require.NotNil(t, distros)
 
+	// TODO pass it a real path?
```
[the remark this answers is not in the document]
bcl: oh whoops, forgot to remove the TODO after I discovered the `reporegistry.NewTestedDefault` function.

internal/cloudapi/v2/compose.go
```diff
@@ -1002,6 +1003,15 @@ func (request *ComposeRequest) GetImageRequests(distroFactory *distrofactory.Fac
 			return nil, err
 		}
 
+		// If no repositories are included with the imageRequest use the defaults for the distro
+		if len(ir.Repositories) == 0 {
+			dr, err := repoRegistry.ReposByImageTypeName(distribution.Name(), arch.Name(), imageType.Name())
```
thozza, line 1008: I think that you should use `request.Distribution` instead of `distribution.Name()` here. If the DistroFactory has an alias set for the distro name specified in the ImageRequest, then the `distribution.Name()` would return the distro name of the target distro of the alias.

This is done also in Weldr, see https://github.com/osbuild/osbuild-composer/blob/8f28da5214f4078593dea051d1eee25727ab1473/internal/weldr/api.go#L3568-L3571

thozza (commented):
I found one problem that I commented on.

Other than that, I'm wondering if it would be worth to extend some of the existing functional test cases to also cover this feature, or at least adding a new unit test for it.

The rest looks good :+1:

internal/cloudapi/v2/compose.go
bcl: Are you sure that will work? I see obvious alias code in distrofactory but not in the reporegistry code.

thozza: I'm pretty sure, since it is tested in Weldr by:
https://github.com/osbuild/osbuild-composer/blob/073e304978acade8bfa059f00a005402aa037c99/test/cases/weldr-distro-dot-notation-and-aliases.sh#L318-L341

The idea is that if one specifies a distro name in the request, we always need to use it as is. That means that we use that string to get the distro instance and repositories.

In DistroFactory, this means first translating the potential alias to the target distro name and getting the Distro object (however, that won't reflect the aliased name on its own).

For repositories, this means grabbing the repos that correspond to the `<distro_name>.json`. If one specified an alias, there must be repositories defined for that alias distro name. If there are no repos, one can't build that distro (which is consistent with e.g. missing el9 repos on el8 - the distro definition is there, but you can't build it because of missing repos). For this reason, there's no aliasing code in repositories, since repo definitions must be always explicit. This is how it worked before DistroFactory and I kept the behavior intentionally.

I admit that it would be good to capture this information somewhere, since it is not obvious. Happy to do so if you have any suggestions. :wink:

bcl: Ok, I think I understand now. The bit I was missing was that the alias is just an alias for the distro definition, NOT for the repository json file. I was thinking, for example, `rhel-9` would then get aliased to `rhel-9.3` and use the `rhel-9.3.json` repo file. But it doesn't, it uses `rhel-9.json` which points to the latest release on the cdn.

bcl:
> I'm wondering if it would be worth to extend some of the existing functional test cases to also cover this feature, or at least adding a new unit test for it.

Added a unit test to check it. I had to hard-code a path traversal so I could load the repos, it seems stable, but let me know if there's a better way.

supakeen (approved):
Neat.

thozza (approved):
Thanks for adding the test case.

internal/cloudapi/v2/compose_test.go
```diff
@@ -657,16 +659,42 @@ func TestGetImageRequests_ImageTypeConversion(t *testing.T) {
...
+		Distribution: "fedora-40",
+		ImageRequest: &ImageRequest{
+			Architecture:  "x86_64",
+			ImageType:     ImageTypesAws,
+			UploadOptions: &uo,
+			Repositories:  []Repository{},
+		},
+	}
+	// NOTE: current directory is the location of this file, back up so it can use ./repositories/
+	rr, err := reporegistry.New([]string{"../../../"})
```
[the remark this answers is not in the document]
bcl: Yeah, it works so far :)

PR #3939, opened 2024-02-09
osbuild-runner: run osbuild in an ec2 vm

croissanne:


templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_builder.sh
```diff
@@ -0,0 +1,13 @@
...
+
+RUN_BUILDER=${RUN_BUILDER:-}
+if [[ -z "$RUN_BUILDER" ]] || [ "$RUN_BUILDER" != true ]; then
+  echo "RUN_BUILDER not defined or not set to 'true', not starting builder."
+  exit 0
+fi
+
+echo "Starting osbuild-jobsite-builder."
+/usr/libexec/osbuild-composer/osbuild-jobsite-builder &
+disown $!
```
ondrejbudai, line 13: What about running this in its own service? Enabled in the playbook, but running only when `ConditionPathExists` is true (the fleet would create a file, something like `/tmp/run_builder_service`).

internal/osbuildrunner/osbuild-runner.go
```diff
@@ -0,0 +1,11 @@
+package osbuildrunner
+
+import (
+	"io"
+
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+type Runner interface {
```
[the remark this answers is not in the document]
croissanne: yes! perfect

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,370 @@
...
+			aws.String(sgID),
+		},
+	})
+	if err != nil {
+		return sgID, err
+	}
+	if len(describeSGOutput.SecurityGroups) != 1 {
+		return sgID, fmt.Errorf("Expected 1 security group")
+	}
+	if len(describeSGOutput.SecurityGroups[0].IpPermissions) != 1 || len(describeSGOutput.SecurityGroups[0].IpPermissionsEgress) != 2 {
```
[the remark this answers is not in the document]
croissanne: there's also a default rule allowing all outgoing traffic. Since we're still pulling sources we need it. In future we can remove it when the worker is a full proxy for the sources.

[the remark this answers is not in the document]
ondrejbudai: Maybe let's add a comment about it? It's not obvious on the first glance.

internal/osbuildexecutor/osbuild-executor.go
```diff
@@ -0,0 +1,11 @@
+package osbuildexecutor
```
bcl, line 1: Commit title calls this osbuild**er**executor. I'd comment on the commit message, but you can't :)

internal/cloud/awscloud/awscloud.go
```diff
@@ -21,6 +22,7 @@ import (
 type AWS struct {
 	uploader *s3manager.Uploader
 	ec2      *ec2.EC2
+	ec2md    *ec2metadata.EC2Metadata
```
bcl, line 25: Might be easier to read later if `ec2md` was `ec2metadata` since that's what it actually is.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	if len(descrInstancesOutput.Reservations) != 1 || len(descrInstancesOutput.Reservations[0].Instances) != 1 {
+		return nil, fmt.Errorf("Expected exactly one instance")
+	}
+	vpcID := *descrInstancesOutput.Reservations[0].Instances[0].VpcId
+	imageID := *descrInstancesOutput.Reservations[0].Instances[0].ImageId
+
+	var secureInstance *SecureInstance
+	defer func() {
+		if secureInstance != nil && secureInstance.Instance == nil {
```
bcl, line 48: This is pretty long for a defer :) I think it would be better as a cleanup function that's called. Or at least turn the wrapping nil checks into an early return so the actual code can be indented less.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,378 @@
...
+					GroupId: aws.String(secureInstance.SGID),
+				})
+				if err != nil {
+					logrus.Errorf("Error deleting security group %s: %v", secureInstance.SGID, err)
+				}
+			}
+		}
+	}()
+
+	sgID, err := a.createOrReplaceSG(identity.InstanceID, identity.PrivateIP, vpcID)
```
bcl, line 83: I had to dig into this a little before I understood that sdID could be set, and need cleaning up, even when there is an error. Maybe a comment about that is needed ?
Also, it seems (from the outside of the function) fragile to depend on an if block to create the SecureInstance, is there any reason not to just make an empty one and fill it in?

croissanne: I tried to make it a bit clearer, and construct a default SI now.

croissanne: I tried to restructure the cleanup a bit, and added a small comment on top of the fn.

cmd/osbuild-worker/config.go
```diff
@@ -113,5 +123,9 @@ func parseConfig(file string) (*workerConfig, error) {
 		}
 	}
 
+	if config.OSBuildExecutor.Type != "host" && config.OSBuildExecutor.Type != "aws.ec2" && config.OSBuildExecutor.Type != "qemu.kvm" {
```
supakeen, line 126: I'd prefer a `.Contains` here but eh.

supakeen (dismissed):

cmd/osbuild-worker/config.go
croissanne: is that in go 1.19?

croissanne: Introduced in 1.21, otherwise only available in experimental pkgs

supakeen: Oof, ok.

templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
```diff
@@ -0,0 +1,18 @@
...
+  --query SecretList[0].ARN --output text
+)
+
+/usr/local/bin/aws secretsmanager get-secret-value \
+  --secret-id "${SUBSCRIPTION_MANAGER_COMMAND_ARN}" | jq -r ".SecretString" > /tmp/subscription_manager_command.json
+jq -r ".subscription_manager_command" /tmp/subscription_manager_command.json | bash
+rm -f /tmp/subscription_manager_command.json
+
+echo "Starting osbuild-jobsite-builder."
+/usr/libexec/osbuild-composer/osbuild-jobsite-builder &
```
ondrejbudai, line 17: I wanted a separate service for the builder, so we don't have to run it in background and disown it. Now we can just run it as a part of the system service. Since it's a *oneshot* one, it will be run just once, which is AFAIK exactly what we want. :) IOW, I suggest removing `&` and disown. :)

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	for _, exp := range exportPaths {
+		args = append(args, "--export-file", exp)
+	}
+	for _, env := range extraEnv {
+		args = append(args, "--environment", env)
+	}
+	args = append(args, "--output", outputDirectory)
+
+	cmd := exec.Command(
```
bcl, line 59: I think this should have a timeout context, just in case something gets stuck. I'm not sure how long it should be though, so I'd err on the side of longer to make sure it doesn't accidentally interrupt a running build.

bcl (changes requested):
All my comments except one are cosmetic :) The only actual change I think it needs is a worst-case timeout on the `exec.Command`

templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
croissanne: Oh ok, yea that makes total sense!

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,396 @@
...
+	ltID, err := a.createOrReplaceLT(identity.InstanceID, imageID, sgID, iamProfile)
+	if ltID != "" {
+		secureInstance.LTID = ltID
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	createFleetOutput, err := a.ec2.CreateFleet(&ec2.CreateFleetInput{
+		LaunchTemplateConfigs: []*ec2.FleetLaunchTemplateConfigRequest{
```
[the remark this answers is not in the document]
croissanne: I think this is just a feature of the aws sdk :/ I'm not sure if it would be that much better considering it's already in the private `createOrReplaceLT` function.

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: So the `/usr/libexec/osbuild-composer/osbuild-jobsite-manager` already had a bunch of timeouts included inside of it. That waits 60 minutes.

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -0,0 +1,97 @@
...
+		"/usr/libexec/osbuild-composer/osbuild-jobsite-manager",
+		args...,
+	)
+
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+
+	err = cmd.Start()
```
mvo5, line 69: A bit of a meta-question - given that osbild-jobsite-manager is part of this repo and written in go, why is it not used directly from go code, why the extra indirection via a os/exec and the json? is it for robustness, i.e. a crashing osbuild-jobsite-manager would not affect the service? or is there a different/deeper reason?

mvo5 (commented):
Thank you, this looks very nice. I really like the attention of detail in the security setup! I add a whole bunch of ideas/suggestions and also have some questions. All is just nitpicks or personal preferences, feel free to ignore them if you have different preferences or opinions :) Hope this is useful.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func (a *AWS) createOrReplaceSG(hostInstanceID, hostIP, vpcID string) (string, error) {
+	sgName := fmt.Sprintf("SG for %s (%s)", hostInstanceID, hostIP)
+	descrSGOutput, err := a.ec2.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
+		GroupNames: []*string{
+			aws.String(sgName),
+		},
+	})
+	if len(descrSGOutput.SecurityGroups) == 1 {
```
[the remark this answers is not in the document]
croissanne: I don't think it can no, considering we have to give a name and the name needs to be unique, AWS would have to be bugged.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,404 @@
...
+
+	createLaunchTemplateOutput, err := a.ec2.CreateLaunchTemplate(input)
+	if err != nil {
+		return "", err
+	}
+	return *createLaunchTemplateOutput.LaunchTemplate.LaunchTemplateId, nil
+}
+
+func (a *AWS) deleteFleetIfExists(si *SecureInstance) error {
+	if si.FleetID == "" {
```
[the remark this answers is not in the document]
croissanne: Ah interesting, yea maybe. I don't think there's currently a scenario where that could happen unless you terminate the instance twice. But why not!

internal/osbuildexecutor/osbuild-executor.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"io"
+
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+type Executor interface {
+	RunOSBuild(manifest []byte, store, outputDirectory string, exports, exportPaths, checkpoints, extraEnv []string, result bool, errorWriter io.Writer) (*osbuild.Result, error)
```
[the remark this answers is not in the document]
croissanne: Yea absolutely, but let's do that in a followup, don't want to change this API here. The result bool is super weird.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -0,0 +1,404 @@
...
+	if len(descrInstancesOutput.Reservations) != 1 || len(descrInstancesOutput.Reservations[0].Instances) != 1 {
+		return nil, fmt.Errorf("Expected exactly one instance")
+	}
+	vpcID := *descrInstancesOutput.Reservations[0].Instances[0].VpcId
+	imageID := *descrInstancesOutput.Reservations[0].Instances[0].ImageId
+
+	secureInstance := &SecureInstance{}
+	defer func() {
+		if secureInstance.Instance == nil {
+			err := a.deleteFleetIfExists(secureInstance)
```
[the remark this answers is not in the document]
croissanne: yes, this one will try to clean up as much as possible. Though I suppose it doesn't really matter tbh, like it doesn't even do retries or whatever. I can merge them.

internal/osbuildexecutor/runner-impl-aws-ec2.go
bcl: ok, as long as you think those are reliable then this should be ok.

croissanne: I'll keep an eye on it, if they turn out not to be we can of course add it.

internal/osbuildexecutor/runner-impl-aws-ec2.go
supakeen: It's a layer of separation here; the worker might/will spawn multiple managers for multiple builders.

supakeen (approved):

PR #3945, opened 2024-02-14
worker-executor followups: fix oneshot service and secure instance for non-default vpcs

croissanne:
cloud/awscloud: specify subnets when creating secure instance
    
 For non-default VPCs, AWS needs the subnets it can launch the instance
 in, otherwise it will try to launch the instance in the default VPC,
 even if the supplied security groups are attached to a non-default VPC.
    
 Furthermore there can only be 1 subnet specified per availability zone,
 so query the subnets in the VPC of the host (as the instance needs to be
 launched in the same network), and pick 1 of the VPC's subnets per AZ.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
+++ b/templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-/usr/local/bin/aws secretsmanager get-secret-value \
-  --secret-id executor-subscription-manager-command | jq -r ".SecretString" > /tmp/subscription_manager_command.json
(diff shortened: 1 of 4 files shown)


templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-/usr/local/bin/aws secretsmanager get-secret-value \
-  --secret-id executor-subscription-manager-command | jq -r ".SecretString" > /tmp/subscription_manager_command.json
-jq -r ".subscription_manager_command" /tmp/subscription_manager_command.json | bash
-rm -f /tmp/subscription_manager_command.json
+# Don't subscribe on fedora
```
ochosi, line 4: Sorry for the naive question, but do we subscribe on CentOS Stream?

PR #3947, opened 2024-02-14
cloud/awscloud: take instance type from host

croissanne:
InstanceRequirements is very flakey, the create fleet request fails almost consistently with the same error.

To continue with testing use a fixed instance type for now. As a followup we can expand the instance type selection logic or figure out what was wrong with the InstanceRequirements.


ochosi (approved):
Sounds like a sane step 1. Thanks!

achilleas-k (approved):

supakeen (approved):

PR #3950, opened 2024-02-14
templates/packer: set -builder-path to /var/cache/osbuild-builder

croissanne:
The builder uses `/run/osbuild` as a default path for this argument. Yet this directory doesn't exist when the builder writes the manifest. But osbuild should own this directory, not the builder.

Furthermore `/run` is a tmpfs, so the executor might run into memory issues if we use `/run` as the store and output directory (on the "host" workers these are in `/var/cache`).

While `/tmp` might seem like a good candidate on RHEL, it's a tmpfs on Fedora, so it's also to be avoided.


achilleas-k (approved):
LGTM

Should we make this the default in the builder?

croissanne:
> Should we make this the default in the builder?

probably

PR #3953, opened 2024-02-15
cmd/osbuild-jobsite: capture osbuild's stdout

croissanne:
Write osbuild's stdout in the progress step. The manager can just copy it to stdout and the executor will be able to parse the output into an osbuild result.


supakeen (approved):
Looks good to me, in a followup (I have a local branch for this) we will likely start chunking out the buffer with each progress request.

PR #3960, opened 2024-02-16
jobsite: handle builder progress errors

supakeen:
This slightly improves handling of builder progress errors, the builder now communicates through a separate status code that the build process exited and includes stderr in its body before exiting.

The manager takes this output and outputs it to its own stderr.

Note: this isn't *entirely* needed; we can get the same information from the logs.


cmd/osbuild-jobsite-builder/main.go
```diff
@@ -119,45 +118,42 @@ func (builder *Builder) GetState() State {
...
+			w.WriteHeader(http.StatusConflict)
+			logrus.Fatal("Received request in invalid state. Exiting")
+		}
+
+		if r.Method != m {
+			w.WriteHeader(http.StatusBadRequest)
+			logrus.Fatal("Received invalid request method. Exiting")
+		}
+
 		if err := h(w, r); err != nil {
```
mvo5, line 141: Might be useful to also log here in what state the error happened, e.g. `logrus.Fatalf("error in %v: %v", s, err)`

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -119,45 +118,42 @@ func (builder *Builder) GetState() State {
...
-	builder.GuardState(StateProvision)
-
-	if r.Method != "PUT" {
-		return fmt.Errorf("Builder.HandleProvision: Unexpected request method")
-	}
-
+func (b *Builder) HandleProvision(w http.ResponseWriter, r *http.Request) error {
 	logrus.WithFields(logrus.Fields{"argBuildPath": argBuildPath}).Debug("Builder.HandleProvision: Opening manifest.json")
 
 	dst, err := os.OpenFile(
```
mvo5, line 159: I think it's fine to just use `os.Create()` here, AFAICT there are no security benefits from creating non-writable files or doing O_EXCL, the inputs come from trusted sources and if someone owns the worker-worker site 0400 will not really stop any attackers. Or am I missing something here?

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -166,13 +162,8 @@ func (builder *Builder) HandleProvision(w http.ResponseWriter, r *http.Request)
 		0400,
 	)
 
-	defer func() {
```
mvo5, line 169: Why would we remove `defer dst.Close()` here ? It's a stanard pattern and if we don't do it we potentially leak file descriptors  in the error handling of e.g. io.Copy(). Sure it does not matter here because the program dies but I feel it's still worthwhile (code gets refactored etc)

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -166,13 +162,8 @@ func (builder *Builder) HandleProvision(w http.ResponseWriter, r *http.Request)
...
-		if cerr := dst.Close(); cerr != nil {
-			err = cerr
```
mvo5, line 171: This pattern is potentially cnfusing as a close error would shadow the original error, I think this should check if there is a pre-existing error and only if that is nil override (I know this got removed but still wanted to point it out)

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -181,58 +172,42 @@ func (builder *Builder) HandleProvision(w http.ResponseWriter, r *http.Request)
...
 		return fmt.Errorf("Builder.HandleProvision: Failed to write manifest.json")
 	}
 
-	w.WriteHeader(http.StatusCreated)
-
-	if _, err := w.Write([]byte(`done`)); err != nil {
-		return fmt.Errorf("Builder.HandleProvision: Failed to write response")
+	if err = dst.Close(); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return fmt.Errorf("Builder.HandleProvision: Failed to close manifest.json")
```
mvo5, line 181: Here (and above) discarding the original error means we close information, I think this should be something like: `		return fmt.Errorf("Builder.HandleProvision: Failed to close manifest.json: %v", err)`

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -253,91 +228,82 @@ func (builder *Builder) HandleBuild(w http.ResponseWriter, r *http.Request) erro
 	envs := os.Environ()
 	envs = append(envs, buildRequest.Environments...)
 
-	builder.Build = &BackgroundProcess{}
-	builder.Build.Process = exec.Command(
+	b.Build = &BackgroundProcess{}
+	b.Build.Process = exec.Command(
 		"/usr/bin/osbuild",
```
mvo5, line 233: (nipitck) if we just use "osbuild" and rely on PATH this will be easier testable by providing a mock binary in a test controlled PATH. It should have no security implications because if someone overrides PATH on our machine we have bigger problems. OTOH there are other ways to test this so fine to ignore.

docs/architecture.md
```diff
@@ -0,0 +1,42 @@
...
+
+Workers are separate processes that look at the jobqueue and take any jobs that they can perform. Jobs are differentiated on architecture and/or operating system.
+
+#### Jobsite
+
+When a worker has accepted a job it will setup a temporary area to do this work in. This area is called the jobsite. The area takes the form of a temporary network in which a few processes are ran that are optionally ran in separate virtual machines.
+
+The jobsite goes through several distinct phases which happen in lockstep between the `manager` and the `builder`, these phases are:
+
+- Claim -- The manager ensures the builder has started up and is available.
```
mvo5, line 25: I would love if we could revisit these states at some point. I feel `/build`, `/progress`, `export` is all we need, having a single upload endpoint has the advantage that it's a single transfer, it's very simple to drive via e.g. curl and we can still do all checks and include sources by doing e.g. multipart/form-data and while streaming validating the content.

mvo5 (commented):
Thanks for the cleanups! A few quick drive-by comments, I have not looked in great detail (sorry for that!) but hope some of my suggestions are still useful :)

cmd/osbuild-jobsite-builder/main.go
supakeen: Agreed, I've changed all error conversions to include the original `err`.

cmd/osbuild-jobsite-builder/main.go
supakeen: It wasn't much meant as a security measure but more as an assertion that the machine the builder is running on is in a correct state (no output has been written yet).

However since this makes running the builder locally more annoying and the gain is small I've switched this to `os.Create`.

cmd/osbuild-jobsite-builder/main.go
supakeen: Agreed however since the idiom has changed now this part can be discussed in the other comment for the `defer dst.Close()`.

cmd/osbuild-jobsite-builder/main.go
supakeen: It makes the error handling slightly more finicky and less linear to read. Since we always exit we won't be leaking. If this gets refactored into using a deferred close again then yes it can be brought up then.

cmd/osbuild-jobsite-builder/main.go
supakeen: It's definitely a habit of mine to always specify full paths to prevent any `PATH` shenanigans. I'm going to leave it in because as you said there are other ways to test this or we can alternatively make it part of some configuration that can be set by the test cases.

cmd/osbuild-jobsite-builder/main.go
supakeen: Done.

cmd/osbuild-jobsite-builder/main.go
mvo5: Thanks for the rational. I personally like the go-pattern of "open/check err/defer close" and it's so common that every time I see something that diverges from the pattern my flow is interrupted. Sorry, that's a winded way of saying that I really like consistency in resource/error handling :) (plus if it gets refactored we may not remember about this anymore as it diverges from the usual pattern).

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -119,120 +118,94 @@ func (builder *Builder) GetState() State {
...
-
+func (b *Builder) HandleBuild(w http.ResponseWriter, r *http.Request) error {
 	var buildRequest BuildRequest
 
-	var err error
-
-	if err = json.NewDecoder(r.Body).Decode(&buildRequest); err != nil {
+	if err := json.NewDecoder(r.Body).Decode(&buildRequest); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
 		return fmt.Errorf("HandleBuild: Failed to decode body")
```
mvo5, line 204: It would be nice to include the origin error here, e.g. `fmt.Errorf("HandleBuild: Failed to decode body: %w", err)`

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -253,91 +226,82 @@ func (builder *Builder) HandleBuild(w http.ResponseWriter, r *http.Request) erro
...
-		w.WriteHeader(http.StatusOK)
+	// If there was an error in the build process we respond with an appropriate error status code and include
+	// stderr in the body of the response before exiting.
+	if b.Build.Error != nil {
+		w.WriteHeader(http.StatusConflict)
 
-		if builder.Build.Error != nil {
-			return fmt.Errorf("Builder.HandleBuild: Buildprocess exited with error: %s", builder.Build.Error)
+		if _, err := w.Write(b.Build.Stderr.Bytes()); err != nil {
+			return fmt.Errorf("Builder.HandleBuild: Failed to write stderr response")
```
mvo5, line 280: Same here, would be nice to include the original error here as well

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -253,91 +226,82 @@ func (builder *Builder) HandleBuild(w http.ResponseWriter, r *http.Request) erro
...
+	}
 
-		builder.SetState(StateExport)
-	} else {
-		w.WriteHeader(http.StatusAccepted)
+	// Otherwise we respond with OK and include stdout instead.
+	w.WriteHeader(http.StatusOK)
+
+	if _, err := w.Write(b.Build.Stdout.Bytes()); err != nil {
+		return fmt.Errorf("Builder.HandleBuild: Failed to write stdout response")
```
mvo5, line 290: and here

cmd/osbuild-jobsite-builder/main.go
mvo5: Is a push missing here maybe? The comment says "done" but the code is still the same?

mvo5 (commented):
Thanks for considering my comments and updating the code, appreciated! I added a few more nitpicks comments/suggestion. Out of curiosity, what is missing to get it out of "draft" state?

supakeen:
@mvo5 it indeed seems something went weird with some git stashes, I've re-pushed with them applied which should address the latest comments.

There was no definition of done for the quality of life improvements but I've marked it as ready now.

ochosi:
I presume we want both a small unit test and a proper integration test that spins up the virtual machines, no? I wonder if both of these things make this PR balloon too much :shrug:

supakeen:
> I presume we want both a small unit test and a proper integration test that spins up the virtual machines, no? I wonder if both of these things make this PR balloon too much 🤷

Yes, it's OK to add them to this PR since this PR doesn't change anything functionally :)

PR #3828, opened 2023-11-30
switch to images/pkg/dnfjson and remove internal copy

diaasami:
https://issues.redhat.com/browse/COMPOSER-2068

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3962, opened 2024-02-16
Add minimal-raw and clean up the compose handler a bit

ondrejbudai:
Firstly, this adds the minimal-raw image type to the cloud API.

However, I wasn't sure how to test it - we surely shouldn't build all images in the composer's CI - that's what the images' CI is for. Since adding an image type to cloud API is about implementing a translation between cloud API and the image definitions API, I decided to clean up the translation code a bit to reduce its API surface and cover it with a smoke test. The test itself is pretty dumb, but that's just how tests for translation layers look like. :shrug:


internal/cloudapi/v2/server.go
```diff
@@ -126,6 +126,9 @@ func (s *Server) enqueueCompose(distribution distro.Distro, bp blueprint.Bluepri
 	ir := irs[0]
 
 	ibp := blueprint.Convert(bp)
+	// shortcut
+	arch := ir.imageType.Arch()
```
achilleas-k: The comment here doesn't really help much.  I think it's clear what's going on from the code.

achilleas-k (approved):
I love the API simplifications here.  More of this!! <3

LGTM

thozza:
The nightly pipeline failure is a known issue tracked in https://issues.redhat.com/browse/RHEL-24346

thozza:
The failure in the edge test case is a known issue, as confirmed by https://github.com/osbuild/osbuild-composer/pull/3828#issuecomment-1952384801.

The failure in the nightly pipeline is a known issue, tracked in https://issues.redhat.com/browse/RHEL-24346

All the required tests passed. Merging manually (@nullr0ute asked to get this in ASAP), since there is an unresolved discussion, which I want to keep open.

PR #3964, opened 2024-02-19
RHEL-16006: improve error on missing package name

schuellerf:
Improves the error message if a name or (name and version) is missing.

Also introduces a test for this.

This pull request includes:

- [✔] adequate testing for the new functionality or fixed issue
- [✔] adequate documentation informing people about the change such as
  - [✔] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

https://issues.redhat.com/browse/RHEL-16006

### DIFF
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -105,7 +105,7 @@ func (b *Blueprint) Initialize() error {
 		if len(p.Name) == 0 {
 			var errMsg string
 			if len(p.Version) == 0 {
-				errMsg = fmt.Sprintf("Entry #%d has neither version nor name.", i+1)
+				errMsg = fmt.Sprintf("Entry #%d has no name.", i+1)
 			} else {
 				errMsg = fmt.Sprintf("Entry #%d has version '%v' but no name.", i+1, p.Version)
 			}
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -101,13 +101,13 @@ func (b *Blueprint) Initialize() error {
 		return fmt.Errorf("Error hashing passwords: %s", err.Error())
 	}
 
-	for i, p := range b.Packages {
-		if len(p.Name) == 0 {
+	for i, pkg := range b.Packages {
+		if pkg.Name == "" {
 			var errMsg string
-			if len(p.Version) == 0 {
+			if pkg.Version == "" {
 				errMsg = fmt.Sprintf("Entry #%d has no name.", i+1)
 			} else {
-				errMsg = fmt.Sprintf("Entry #%d has version '%v' but no name.", i+1, p.Version)
+				errMsg = fmt.Sprintf("Entry #%d has version '%v' but no name.", i+1, pkg.Version)
 			}
 			return fmt.Errorf("All package entries need to contain the name of the package. %s", errMsg)
 		}
--- a/internal/blueprint/blueprint.go
+++ b/internal/blueprint/blueprint.go
@@ -101,6 +101,18 @@ func (b *Blueprint) Initialize() error {
 		return fmt.Errorf("Error hashing passwords: %s", err.Error())
 	}
 
+	for i, p := range b.Packages {
+		if len(p.Name) == 0 {
+			var errMsg string
+			if len(p.Version) == 0 {
+				errMsg = fmt.Sprintf("Entry #%d has neither version nor name.", i+1)
+			} else {
+				errMsg = fmt.Sprintf("Entry #%d has version '%v' but no name.", i+1, p.Version)
+			}
+			return fmt.Errorf("All package entries need to contain the name of the package. %s", errMsg)
+		}
+	}
+
 	return nil
 }
 
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -1229,3 +1229,56 @@ func TestBlueprintChangeV1(t *testing.T) {
 	require.NotNil(t, bp, "GET blueprint change failed: missing blueprint")
 	assert.Equal(t, bp.Version, "0.0.1")
 }
+
+func TestEmptyPackageNameBlueprintJsonV0(t *testing.T) {
+	bp := `{
+		"name": "test-emptypackage-blueprint-v0",
+		"description": "TestEmptyPackageNameBlueprintV0",
+		"version": "0.0.1",
+		"packages": [{"name": "", "version": "1.32.1"}]
+	}`
+
+	// Push a blueprint
+	resp, err := PostJSONBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "failed with a client error")
+	require.False(t, resp.Status, "Negative status expected.")
+	require.Equal(t, len(resp.Errors), 1, "There should be exactly one error")
+	assert.True(t, strings.HasPrefix(resp.Errors[0].String(), "BlueprintsError"), "I expect a BlueprintsError")
+}
+
+func TestEmptyPackageNameBlueprintV0(t *testing.T) {
+	bp := `name = "EMPTY-PACKAGE-NAME"
+           description = "empty package name"
+           version = "0.0.1"
+           modules = []
+           groups = []
+           distro = "test-distro"
+
+           [[packages]]
+           version = "*"
+           `
+
+	resp, err := PostTOMLBlueprintV0(testState.socket, bp)
+	require.NoError(t, err, "failed with a client error")
+	require.False(t, resp.Status, "Negative status expected.")
+	require.Equal(t, len(resp.Errors), 1, "There should be exactly one error")
+	assert.True(t, strings.HasPrefix(resp.Errors[0].String(), "BlueprintsError"), "I expect a BlueprintsError")
+}
+
+func TestEmptyPackageNameAndVersionBlueprintV0(t *testing.T) {
+	bp := `name = "EMPTY-PACKAGE-NAME"
+           description = "empty package name"
+           version = "0.0.1"
+           modules = []
+           groups = []
+           distro = "test-distro"
+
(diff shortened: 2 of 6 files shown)


schuellerf:
the last commit is obviously not directly related to this PR, so we can remove it or add it to another PR

internal/weldr/api.go
```diff
@@ -2086,6 +2086,23 @@ func (api *API) blueprintsNewHandler(writer http.ResponseWriter, request *http.R
 		}
 	}
 
+	for i, p := range blueprint.Packages {
```
bcl, line 2089: Good idea! I think it makes sense to make this part of the blueprint's `Initialize` function so that it isn't just checked by the weldr api.

internal/client/blueprints_test.go
```diff
@@ -1229,3 +1229,56 @@ func TestBlueprintChangeV1(t *testing.T) {
 	require.NotNil(t, bp, "GET blueprint change failed: missing blueprint")
 	assert.Equal(t, bp.Version, "0.0.1")
 }
+
```
bcl, line 1232: The (loose) idea in the blueprint testing is to test the same things as JSON and TOML to make sure they return the same result/error and it's easier to compare things if the content of the blueprint is the same in the TOML as it is in the JSON.

bcl (changes requested):
Pretty good, just a couple tweaks, and I'm fine with leaving the Makefile change -- no need to go through another PR for something so simple.

internal/client/blueprints_test.go
schuellerf: So my tests are fine or how should they be changed? 🤔

schuellerf: ... when I moved the implementation to `internal/blueprint/blueprint.go` probably the tests should also go to `internal/blueprint/blueprint_test.go`?

schuellerf:
As I had problems with at least `./tools/prepare-source.sh` in my first push I added "make push-check" to check for as many problems locally before pushing as possible. Do you think this makes sense?

internal/blueprint/blueprint.go
```diff
@@ -101,6 +101,18 @@ func (b *Blueprint) Initialize() error {
 		return fmt.Errorf("Error hashing passwords: %s", err.Error())
 	}
 
+	for i, p := range b.Packages {
+		if len(p.Name) == 0 {
+			var errMsg string
+			if len(p.Version) == 0 {
+				errMsg = fmt.Sprintf("Entry #%d has neither version nor name.", i+1)
```
ondrejbudai, line 108: No version is probably fine, isn't it? I don't remember ever putting a version into a blueprint... Thus, I would remove this, because it suggests that the version is a required field.

schuellerf: It's just "no version" AND "no name" - I'll improve the wording

schuellerf:
@ondrejbudai  I improved the wording and when implementing the "error message wording check" I realized that the test failed before due to wrong distribution but the test has been fine 😮
Thanks for this hint 😉

internal/client/blueprints_test.go
```diff
@@ -1229,3 +1229,61 @@ func TestBlueprintChangeV1(t *testing.T) {
...
+func TestEmptyPackageNameBlueprintJsonV0(t *testing.T) {
```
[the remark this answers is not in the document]
schuellerf: I'll still keep JSON and TOML tests separate. That seems more readable for me

internal/client/blueprints_test.go
```diff
@@ -1229,3 +1229,48 @@ func TestBlueprintChangeV1(t *testing.T) {
...
+		`"packages": [{"name": "", "version": "*"}]`,
+		`"packages": [{"name": ""}]`,
+	} {
+		bp := `{
+		"name": "test-emptypackage-blueprint-v0",
+		"description": "TestEmptyPackageNameBlueprintV0",
+		"version": "0.0.1",
+        ` + packageSnippet + `}`
+
+		expectedErrorPrefix := "BlueprintsError: All package entries need to contain the name of the package."
```
mvo5, line 1244: This is nice, thank you! It seems with a tiny bit of extra effort we can test the exact error here too, I had a quick look and something like this might work? This would give us this nice extra confidence that both cases are handled correctly and in case we refactor that we still are confident about t it :)
```go
index b96dc466e..997a048f7 100644
--- a/internal/client/blueprints_test.go
+++ b/internal/client/blueprints_test.go
@@ -1231,25 +1231,32 @@ func TestBlueprintChangeV1(t *testing.T) {
 }
 
 func TestEmptyPackageNameBlueprintJsonV0(t *testing.T) {
-	for _, packageSnippet := range []string{
-		`"packages": [{"name": "", "version": "*"}]`,
-		`"packages": [{"name": ""}]`,
+	for _, tc := range []struct {
+		packageSnippet string
+		specificErr    string
+	}{
+		{
+			`"packages": [{"name": "", "version": "*"}]`,
+			"Entry #1 has version '*' but no name.",
+		}, {
+			`"packages": [{"name": ""}]`,
+			"Entry #1 has no name.",
+		},
 	} {
 		bp := `{
 		"name": "test-emptypackage-blueprint-v0",
 		"description": "TestEmptyPackageNameBlueprintV0",
 		"version": "0.0.1",
-        ` + packageSnippet + `}`
+        ` + tc.packageSnippet + `}`
 
 		expectedErrorPrefix := "BlueprintsError: All package entries nee
d to contain the name of the package."
+		expectedErr := expectedErrorPrefix + " " + tc.specificErr
 
 		resp, err := PostJSONBlueprintV0(testState.socket, bp)
 		require.NoError(t, err, "failed with a client error")
 		require.False(t, resp.Status, "Negative status expected.")
 		require.Equal(t, len(resp.Errors), 1, "There should be exactly o
ne error")
-		reasoning := "I expected an error message starting with\n" + exp
ectedErrorPrefix +
-			"\nnot\n" + resp.Errors[0].String()
-		assert.True(t, strings.HasPrefix(resp.Errors[0].String(), expect
edErrorPrefix), reasoning)
+		assert.Equal(t, resp.Errors[0].String(), expectedErr)
 	}
 }
 
```

internal/client/blueprints_test.go
```diff
@@ -1229,3 +1229,48 @@ func TestBlueprintChangeV1(t *testing.T) {
...
+
+		resp, err := PostJSONBlueprintV0(testState.socket, bp)
+		require.NoError(t, err, "failed with a client error")
+		require.False(t, resp.Status, "Negative status expected.")
+		require.Equal(t, len(resp.Errors), 1, "There should be exactly one error")
+		reasoning := "I expected an error message starting with\n" + expectedErrorPrefix +
+			"\nnot\n" + resp.Errors[0].String()
+		assert.True(t, strings.HasPrefix(resp.Errors[0].String(), expectedErrorPrefix), reasoning)
```
mvo5, line 1252: Related to the above, the `assert.Equal()` output is really nice, so if we do the above we also get nice diffs in case of failure :)

mvo5 (commented):
Thanks! I still like this a lot! Another tiny suggesiton inline about the tests for your consideration

bcl (dismissed):
Looks good to me, thanks!

thozza (dismissed):
LGTM

bcl (approved):
Looks good, thanks!

thozza:
The Installer failure on RHEL-8.10 seems to be an issue in other PRs as well, but I can't tell what's the problem just from inspecting logs.

The failures in the nightly pipeline are caused by:

1. Base tests - the test added in this PR is failing on nightly compose, since the composer version there does not have the fix which is included in this PR.
2. aws.sh - known issue caused by NetworkManager AVCs.

The edge-commit-cs9 failure is known issue.

internal/client/blueprints_test.go
```diff
@@ -1229,3 +1229,66 @@ func TestBlueprintChangeV1(t *testing.T) {
...
+		expectedErrorPrefix := "BlueprintsError: All package entries need to contain the name of the package."
+		expectedErr := expectedErrorPrefix + " " + tc.specificErr
+
+		resp, err := PostJSONBlueprintV0(testState.socket, bp)
+		require.NoError(t, err, "failed with a client error")
+		require.False(t, resp.Status, "Negative status expected.")
+		require.Equal(t, len(resp.Errors), 1, "There should be exactly one error")
+		reasoning := "I expected the error message\n" + expectedErr +
+			"\nnot\n" + resp.Errors[0].String()
+		assert.True(t, strings.HasPrefix(resp.Errors[0].String(), expectedErr), reasoning)
```
mvo5, line 1261: (nitpick) - any specific reason to not just use `assert.Equal()` here? it's very nice and  gives a full diff when things are different (and makes the "reasoning" string obsolete)

schuellerf: no, sorry, copy'n'paste and too little experience with go :-/

schuellerf:
>     1. Base tests - the test added in this PR is failing on nightly compose, since the composer version there does not have the fix which is included in this PR.

@thozza - how to fix this chicken and egg problem you mentioned?

btw - just pushed a rebase again

thozza:
> @thozza - how to fix this chicken and egg problem you mentioned?

We tend to workaround it, but in this case it would be tricky... https://github.com/osbuild/osbuild-composer/blob/242df9ab750103ddc2a1967e4c04a0a41a0f3647/test/cases/base_tests.sh#L54-L59 So the solution would be to selectively disable the test for old versions of composer.

But I'd be OK with ignoring the failure for now, since it will get resolved once the new composer release gets to RHEL nightly compose.

ondrejbudai (approved):
Let's get this in, then. :)

thozza (approved):

PR #3967, opened 2024-02-21
Skip Image Tests tests for image types already covered in osbuild/images

atodorov:
Related: COMPOSER-2127

This pull request includes:


atodorov:
8.10 job now taking 217 minutes instead of timing out:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/6222764883

atodorov:
Not quite sure how `image_installer-boot` and `image_installer_with_users-boot` relate to `installers.sh` which uses the `image-installer` image type. At least one of them can be filtered out.

atodorov:
After the latest changes:
9.4 job down to 93 minutes -> https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/6230989840
8.10 job down to 143 minutes -> https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/6231371569

jrusz:
> Not quite sure how `image_installer-boot` and `image_installer_with_users-boot` relate to `installers.sh` which uses the `image-installer` image type. At least one of them can be filtered out.

Yeah, the blueprint for that test has an `$SSH_USER` as a regular user being added so I suppose the one with users can be removed? But maybe it would cover both, IDK, I'm just guessing, I imagine @thozza would know better.

jrusz (approved):
Very impressive time saving! LGTM

thozza (approved):
Thanks for this PR and sorry for the delay.

I would suggest to also delete the corresponding files from `test/data/manifests` and also the definitions from https://github.com/osbuild/osbuild-composer/blob/main/tools/test-case-generators/format-request-map.json.

This should be enough and we would not need to add any filtering to `image_tests.sh`.

WRT the `image_installers.sh`, the `image_installer_with_users-boot` adds a user and groups https://github.com/osbuild/osbuild-composer/blob/main/tools/test-case-generators/format-request-map.json#L713-L759. The `image_installers.sh` only adds a user used to SSH to the image https://github.com/osbuild/osbuild-composer/blob/main/test/cases/installers.sh#L310-L316.

AFAIK, we do not boot test the installer images in `image_tests.sh` anyway, so we test only that it builds. So we could IMHO stop testing the `image_installer*` images in the image tests and just keep the `image_tests.sh`. Optionally, we could extend the BP in the test case to also create a custom group and use it.

All of these can be follow-ups.

jrusz (approved):

thozza (approved):
Thanks! :broom:

thozza:
The only failure in the regular Gitlab pipeline is caused by Packer job, which was a known issue fixed by https://github.com/osbuild/osbuild-composer/pull/3987

PR #3971, opened 2024-02-22
tools/appsre-build-fedora: wait until rpms are built upstream

croissanne:
The fedora build script should wait until the rpms are released upstream before starting the build, otherwise the rpms will come from the fedora repos.

### DIFF
--- a/tools/appsre-build-fedora-worker-packer.sh
+++ b/tools/appsre-build-fedora-worker-packer.sh
@@ -6,6 +6,23 @@ export SKIP_CREATE_AMI=false
 # Use prebuilt rpms for the fedora images
 export BUILD_RPMS=false
 export SKIP_TAGS="rpmcopy,subscribe"
-export PACKER_ONLY_EXCEPT=--only=amazon-ebs.fedora-38-x86_64,amazon-ebs.fedora-38-aarch64
+FEDORA=fedora-38
+export PACKER_ONLY_EXCEPT=--only=amazon-ebs."$FEDORA"-x86_64,amazon-ebs."$FEDORA"-aarch64
+
+# wait until the rpms are built upstream
+COMMIT_SHA="${COMMIT_SHA:-$(git rev-parse HEAD)}"
+while true; do
+    RET=$(curl -w "%{http_code}" -s -o /dev/null http://osbuild-composer-repos.s3.amazonaws.com/osbuild-composer/"$FEDORA"/x86_64/"$COMMIT_SHA"/state.log)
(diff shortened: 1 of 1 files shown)


supakeen (dismissed):
Approved though I'd like timeouts/max counts on the loops. Shellcheck will need to be fixed.

croissanne:
> Approved though I'd like timeouts/max counts on the loops. Shellcheck will need to be fixed.

Isn't it ok to rely on the timeout of the jenkins job/pipeline? Like the build shouldn't complete if the rpms aren't ready. The timeout is currently 2h.

tools/appsre-build-fedora-worker-packer.sh
```diff
@@ -8,4 +8,20 @@ export BUILD_RPMS=false
 export SKIP_TAGS="rpmcopy,subscribe"
 export PACKER_ONLY_EXCEPT=--only=amazon-ebs.fedora-38-x86_64,amazon-ebs.fedora-38-aarch64
 
+# wait until the rpms are built upstream
+COMMIT_SHA="${COMMIT_SHA:-$(git rev-parse HEAD)}"
+while true; do
```
supakeen, line 13: Can we do this with a for loop so we wait for example for a max of 10 iterations or 5 minutes? (or another sensical value).

tools/appsre-build-fedora-worker-packer.sh
```diff
@@ -8,4 +8,20 @@ export BUILD_RPMS=false
 export SKIP_TAGS="rpmcopy,subscribe"
 export PACKER_ONLY_EXCEPT=--only=amazon-ebs.fedora-38-x86_64,amazon-ebs.fedora-38-aarch64
```
supakeen, line 9: 1. Why is this `fedora-38`?
2. Can we put the `fedora-38` in a variable and use it here and down below?

supakeen:
> > Approved though I'd like timeouts/max counts on the loops. Shellcheck will need to be fixed.
> 
> Isn't it ok to rely on the timeout of the jenkins job/pipeline? Like the build shouldn't start if the rpms aren't ready. The timeout is currently 2h.

Sure, that's OK as well.

tools/appsre-build-fedora-worker-packer.sh
croissanne: considering upstream rpm builds might take an hour or sometimes more, I'd prefer to just use the jenkins timeout.

supakeen (approved):

PR #3972, opened 2024-02-22
Enable masked systemd services in cloudapi (HMS-3661)

kingsleyzissou:
This pull request includes:
This PR adds support for `masked` systemd services via the cloud API.

Depends on:
https://github.com/osbuild/images/pull/464


internal/cloudapi/v2/compose_test.go
```diff
@@ -78,6 +78,7 @@ func GetTestBlueprint() blueprint.Blueprint {
 		Services: &blueprint.ServicesCustomization{
 			Enabled:  []string{"sshd"},
 			Disabled: []string{"cleanup"},
+			Masked:   []string{"firewalld"},
```
bcl, line 81: The `testBlueprint` above also needs to have `"masked"` ... added to it.

internal/weldr/api_test.go
```diff
@@ -629,6 +629,7 @@ func TestBlueprintsCustomizationInfoToml(t *testing.T) {
 			Services: &blueprint.ServicesCustomization{
 				Enabled:  []string{"sshd", "cockpit.socket", "httpd"},
 				Disabled: []string{"postfix", "telnetd"},
+				Masked:   []string{"firewalld"},
```
bcl, line 632: The `testBlueprint` above needs to have `"masked"` ... added to it.

thozza:
Converted to draft, since it needs the change to land in `images` first.

internal/blueprint/blueprint.go
```diff
@@ -48,9 +48,8 @@ type Container struct {
 	Source string `json:"source,omitempty" toml:"source"`
 	Name   string `json:"name,omitempty" toml:"name,omitempty"`
 
-	TLSVerify           *bool   `json:"tls-verify,omitempty" toml:"tls-verify,omitempty"`
```
bcl, line 51: I think the container related changes and the mask changes should be split into separate commits. I also have other concerns about the blueprint field name changes, but since it's already in images that boat may have sailed ;)

thozza (commented):
Changes look good to me, but I would propose to merge the first two commits into a single one (update of images to v0.41.0 and changes to the blueprint). The reason is that otherwise, the commit updating the images version will result in composer not being able to compile. This could impact potential bisecting in the future.

internal/blueprint/blueprint.go
kingsleyzissou: I had that initially. But then it meant that one of the commits would still be broken and wouldn't compile 🫠 so I squashed them together. Happy to change it back though

thozza (approved):

PR #3976, opened 2024-02-26
README: Add osbuild-composer's principles

ochosi:
This pull request includes adds the principles section to the README, analogous to how we did it for osbuild and the image definitions.

Please review and scrutinize!

I've shortened the bits about the API, but am more than happy to decompress those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,28 @@ instance, includes a
 [Cockpit Module](https://github.com/osbuild/cockpit-composer) or using the
 [composer-cli](https://weldr.io/lorax/composer-cli.html) command-line tool.
 
-### Project
+## Project
 
  * **Website**: <https://www.osbuild.org>
  * **Bug Tracker**: <https://github.com/osbuild/osbuild-composer/issues>
- * **matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
-*  **Mailing List**: image-builder@redhat.com
+ * **Matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
+ * **Mailing List**: image-builder@redhat.com
  * **Changelog**: <https://github.com/osbuild/osbuild-composer/releases>
 
-#### Contributing
+### Principles
 
-Please refer to the [developer guide](https://www.osbuild.org/guides/developer-guide/index.html) to learn about our workflow, code style and more.
+1. OSBuild Composer shall only allow users to do what generally makes sense.
+2. Blueprints are the policy layer where we decide what to expose to end users.
+3. If a blueprint can be built, it should also boot.
+4. It should be obvious why a blueprint doesn’t build.
+5. The [cloud API](./internal/cloudapi) is never broken.
+6. In the hosted service, OSBuild Composer is an orchestrator of image builds.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,17 +17,28 @@ instance, includes a
...
 
-### Project
+## Project
 
  * **Website**: <https://www.osbuild.org>
  * **Bug Tracker**: <https://github.com/osbuild/osbuild-composer/issues>
- * **matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
-*  **Mailing List**: image-builder@redhat.com
+ * **Matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
+ * **Mailing List**: image-builder@redhat.com
```
bcl, line 25: Is this the right list to direct people to? The last message I have is from April 2023, and according to the headers it is a private list (the archives are private at least).

README.md
```diff
@@ -17,17 +17,28 @@ instance, includes a
...
  * **Changelog**: <https://github.com/osbuild/osbuild-composer/releases>
 
-#### Contributing
+### Principles
 
-Please refer to the [developer guide](https://www.osbuild.org/guides/developer-guide/index.html) to learn about our workflow, code style and more.
+1. OSBuild Composer shall only allow users to do what generally makes sense.
+2. Blueprints are the policy layer where we decide what to expose to end users.
+3. If a blueprint can be built, it should also boot.
```
ondrejbudai, line 32: I'm still not sure about this one. It's so easy to break an image using a custom RPM/file.

ochosi: Right. I mean it's a principle and as such aspirational. What it means to me is: we don't hand out footguns, but it's  not unbreakable.

README.md
ochosi: It is. There's just almost no traffic there since we don't use it internally and there's no community.

README.md
```diff
@@ -17,17 +17,28 @@ instance, includes a
...
 [composer-cli](https://weldr.io/lorax/composer-cli.html) command-line tool.
 
-### Project
+## Project
 
  * **Website**: <https://www.osbuild.org>
  * **Bug Tracker**: <https://github.com/osbuild/osbuild-composer/issues>
- * **matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
-*  **Mailing List**: image-builder@redhat.com
+ * **Matrix**: https://matrix.to/#/#image-builder:fedoraproject.org
```
bcl, line 24: In https://github.com/osbuild/images/pull/454 we agreed to start using a longer matrix url:
`https://matrix.to/#/#image-builder:fedoraproject.org?web-instance[element.io]=chat.fedoraproject.org`

README.md
achilleas-k: I think the word "should" carries the right amount of weight for this (and is probably correct in the RFC meaning of the word too).

ochosi: @ondrejbudai As no further concerns have been raised - do you have an alternative suggestion or are you good with this as is for now?

ondrejbudai (approved):

achilleas-k (approved):

PR #3681, opened 2023-09-12
Enable testing in OpenShift

atodorov:
b/c I want to expand it further

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

PR #3977, opened 2024-02-27
Add CI job to detect possible problematic usage of `trap`

atodorov:
see for example this change:
https://github.com/osbuild/osbuild-composer/pull/3681/commits/13a3ca24ceef05164a77db1c71b929e262098f41

This pull request includes:

### DIFF
--- a/test/cases/filesystem.sh
+++ b/test/cases/filesystem.sh
@@ -21,6 +21,13 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
 COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
 COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
+function cleanup_on_exit() {
+    greenprint "== Script execution stopped or finished - Cleaning up =="
+    # kill dangling journalctl processes to prevent GitLab CI from hanging
+    sudo pkill journalctl || echo "Nothing killed"
+}
+trap cleanup_on_exit EXIT
+
 # Workaround the problem that 'image-info' can not read SELinux labels unknown to the host from the image
 OSBUILD_LABEL=$(matchpathcon -n "$(which osbuild)")
 sudo chcon "$OSBUILD_LABEL" /usr/libexec/osbuild-composer-test/image-info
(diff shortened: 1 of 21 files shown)


atodorov:
@thozza , @mvo5 , @jrusz should I continue further in this direction and adjust the offending test scripts or do we want a different solution ?

thozza:
> @thozza , @mvo5 , @jrusz should I continue further in this direction and adjust the offending test scripts or do we want a different solution ?

I'm not sure if it makes sense to run this as a GH action for every PR, since we are hitting the runner limits a lot recently. Nevertheless, the general idea of this check is good. Maybe it would be better to add this as a separate script to `tools/` directory and/or to the Makefile as a target.

Fixing all the test scripts as part of this effort is IMO desired.

jrusz (commented):
Looks like a good idea, but if we are having issues with Github runners being overwhelmed, might be a good idea what Tomas suggests to have some custom linting script and run it in Gitlab CI instead...

Also since you've found out it can be a problem I believe it's worth following up and updating the scripts.

atodorov:
> > I'm not sure if it makes sense to run this as a GH action for every PR, since we are hitting the runner limits a lot recently.
> 
> I will rework this into a step in one of the existing CI jobs (Shellcheck), instead of creating a new one. That will not increase the number of separate jobs being executed in GitHub.

As a follow up we can stagger some of the jobs, e.g. shellcheck/lint first and then the rest of the tests. WDYT ?

jrusz (approved):
LGTM. 

I was just wondering, in some of the scripts there are two functions for cleanup like a regular `cleanup` and then `cleanup_on_exit` I assume that's because your intention is to keep current behavior and just take care of killing the journal the proper way with just one trap so you added a `cleanup_on_exit` function for that. Is my assumption correct?

atodorov:
> LGTM.
> 
> I was just wondering, in some of the scripts there are two functions for cleanup like a regular `cleanup` and then `cleanup_on_exit` I assume that's because your intention is to keep current behavior and just take care of killing the journal the proper way with just one trap so you added a `cleanup_on_exit` function for that. Is my assumption correct?

Yes. See commit log for some more context.

test/cases/filesystem.sh
```diff
@@ -21,6 +21,13 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
 COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
 COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
+function cleanup_on_exit() {
+    greenprint "== Script execution stopped or finished - Cleaning up =="
+    # kill dangling journalctl processes to prevent GitLab CI from hanging
+    sudo pkill journalctl || echo "Nothing killed"
+}
```
thozza, line 28: nitpick: It would probably make sense to call `clean_up()` as part of this function and not at the end of the script.

thozza (dismissed):
Thanks!

I commented on one nitpick, but since it was so before, I consider it to be nice to have.

thozza (approved):

PR #3978, opened 2024-02-27
ci: skip ostree-remount check in CS9 due to bug RHEL-25249

mcattamoredhat:
This pull request includes:

The remount of /var and /sysroot mount points fails in edge-commit CS9 due to bug https://issues.redhat.com/browse/RHEL-25249 

This is causing several workflow jobs to fail and affects several PRs.

We suggest to skip the playbook task that makes sure mentioned folders are remounted, until we find out the root cause of the issue. 


7flying (approved):
LGTM

henrywang (approved):
LGTM. Thanks!

PR #3981, opened 2024-03-01
osbuild-jobsite: push sources to executor (HMS-3633)

croissanne:
This pull request includes:

### DIFF
--- a/cmd/osbuild-jobsite-builder/main.go
+++ b/cmd/osbuild-jobsite-builder/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"archive/tar"
 	"bufio"
 	"bytes"
 	"context"
@@ -12,6 +13,8 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -205,6 +208,46 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
 
 	if r.Method != "POST" {
 		return fmt.Errorf("Builder.HandlePopulate: unexpected request method")
+	}
+	storePath := path.Join(argBuildPath, "store")
+	err := os.Mkdir(storePath, 0755)
+	if err != nil {
+		return fmt.Errorf("Builder.HandlePopulate: failed to make store directory: %v", err)
+	}
+
+	tarReader := tar.NewReader(r.Body)
+	for header, err := tarReader.Next(); err != io.EOF; header, err = tarReader.Next() {
+		if err != nil {
+			return fmt.Errorf("Builder.HandlerPopulate: failed to unpack sources: %v", err)
+		}
+
+		// gosec seems overly zealous here, as the destination gets verified
+		dest := filepath.Join(storePath, header.Name) // #nosec G305
+		if !strings.HasPrefix(dest, filepath.Clean(storePath)) {
+			return fmt.Errorf("Builder.HandlerPopulate: name not clean: %v doesn't have %v prefix", dest, filepath.Clean(storePath))
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.Mkdir(dest, header.FileInfo().Mode()); err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to make dir in sources: %v", err)
+			}
+		case tar.TypeReg:
+			file, err := os.Create(dest)
+			if err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to open file in sources: %v", err)
+			}
+			defer file.Close()
+
+			// the inputs are trusted so ignore G110
+			_, err = io.Copy(file, tarReader) // #nosec G110
+			if err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to write file in sources: %v", err)
+			}
+			file.Close()
+		default:
+			return fmt.Errorf("Builder.HandlerPopulate: unexpected tar header type: %v", header.Typeflag)
+		}
 	}
 
 	w.WriteHeader(http.StatusOK)
--- a/cmd/osbuild-jobsite-manager/main.go
+++ b/cmd/osbuild-jobsite-manager/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"archive/tar"
 	"bytes"
 	"encoding/json"
 	"errors"
@@ -14,6 +15,7 @@ import (
 	"os/signal"
 	"path"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -57,6 +59,7 @@ var (
 	argEnvironments ArgumentList
 	argExports      ArgumentList
 	argOutputPath   string
+	argStore        string
 )
 
 type BuildRequest struct {
@@ -87,6 +90,7 @@ func init() {
 
 	flag.Var(&argEnvironments, "environment", "Environments to add. Can be passed multiple times.")
 	flag.StringVar(&argOutputPath, "output", "/dev/null", "Output directory to write to.")
+	flag.StringVar(&argStore, "store", "", "Store to send to builder.")
 
 	flag.Parse()
 
@@ -170,11 +174,11 @@ func Dance(done chan<- struct{}, errs chan<- error) {
 	close(done)
 }
 
-func Request(method string, path string, body []byte) (*http.Response, error) {
+func Request(method string, path string, body io.Reader) (*http.Response, error) {
 	cli := &http.Client{}
 	url := fmt.Sprintf("http://%s:%d/%s", argBuilderHost, argBuilderPort, path)
 
-	req, err := http.NewRequest(method, url, bytes.NewBuffer(body))
+	req, err := http.NewRequest(method, url, body)
 
 	if err != nil {
 		return nil, err
@@ -220,7 +224,7 @@ func Wait(timeout int, fn Step) error {
 
 func StepClaim() error {
 	return Wait(argTimeoutClaim, func(done chan<- struct{}, errs chan<- error) {
-		res, err := Request("POST", "claim", []byte(""))
+		res, err := Request("POST", "claim", nil)
 
 		if err != nil {
 			errs <- err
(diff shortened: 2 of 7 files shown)


cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +208,42 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+	storePath := path.Join(argBuildPath, "store")
+	err := os.Mkdir(storePath, 0755)
+	if err != nil {
+		return fmt.Errorf("Builder.HandlePopulate: failed to make store directory")
+	}
+
+	tarReader := tar.NewReader(r.Body)
+	for header, err := tarReader.Next(); err != io.EOF; header, err = tarReader.Next() {
+		if err != nil {
+			return fmt.Errorf("Builder.HandlerPopulate: failed to unpack sources")
```
mvo5, line 220: Maybe worth adding the original error here with `%w` (also in the other places we generate errors)?

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +208,42 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+		}
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.Mkdir(path.Join(storePath, header.Name), 0755); err != nil {
```
mvo5, line 224: I wonder if we should respect the original permissions from the tar header here instead of hardcoding 0755?

mvo5, line 224: Also I wonder if we should be a little bit paranoid and double check that Name is not funny, i.e. contains `..` or something. More for robustness than security (as the data comes from the trusted side)

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +208,42 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
 	if r.Method != "POST" {
 		return fmt.Errorf("Builder.HandlePopulate: unexpected request method")
 	}
+	storePath := path.Join(argBuildPath, "store")
+	err := os.Mkdir(storePath, 0755)
+	if err != nil {
+		return fmt.Errorf("Builder.HandlePopulate: failed to make store directory")
+	}
+
+	tarReader := tar.NewReader(r.Body)
```
mvo5, line 217: I think we need to `defer r.Body.Close()` here, yes?

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +208,42 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+			}
+		case tar.TypeReg:
+			fmt.Println("Making file: %v", path.Join(storePath, header.Name))
+			err = os.MkdirAll(filepath.Dir(path.Join(storePath, header.Name)), 0755)
+			if err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to make directory %v under sources %v", filepath.Dir(path.Join(storePath, header.Name)), err)
+			}
+
+			file, err := os.Create(path.Join(storePath, header.Name))
+			defer file.Close()
```
mvo5, line 235: The usual pattern here is to first error check, then defer the close. I think in this case it also makes sense to add an explicit close after the `io.Copy` or we have many open file descriptors for big tar archives

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +208,42 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+			if err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to open file %s in sources", path.Join(storePath, header.Name))
+			}
+			if _, err := io.Copy(file, tarReader); err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to write file in sources")
+			}
+
+		default:
+			return fmt.Errorf("Builder.HandlerPopulate: unexpected tar header type")
```
mvo5, line 244: probably good to include here what the unexpected type is/was for easier debugging/logging

mvo5 (commented):
A quick drive by review, happy to do a more detailed one or help in any other way

cmd/osbuild-jobsite-builder/main.go
croissanne: good point

cmd/osbuild-jobsite-manager/main.go
```diff
@@ -264,13 +269,54 @@ func StepProvision(manifest []byte) error {
...
 func StepPopulate() error {
 	return Wait(argTimeoutPopulate, func(done chan<- struct{}, errs chan<- error) {
-		res, err := Request("POST", "populate", []byte(""))
-
+		file, err := ioutil.TempFile(filepath.Dir(argStore), "store")
+		if err != nil {
+			errs <- err
+			return
+		}
+		tw := tar.NewWriter(file)
```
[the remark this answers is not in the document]
croissanne: I don't think it makes all that much of a difference here, as you say it's simple enough. And currently the happy path is covered under the integration test.

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +209,46 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+			return fmt.Errorf("Builder.HandlerPopulate: name not clean: %v != %v", filepath.Clean(header.Name), header.Name)
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.Mkdir(path.Join(storePath, header.Name), header.FileInfo().Mode()); err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to make dir in sources: %v", err)
+			}
+		case tar.TypeReg:
+			err = os.MkdirAll(filepath.Dir(path.Join(storePath, header.Name)), header.FileInfo().Mode())
```
mvo5, line 234: I'm not sure we need this, technically yes but I think we can assume that tar gives us a "well" behaviring tarfile. What I don't like about it is that we now need to make assumptions about dir permissions. Using the permissions of the file (as it's done here) is probably not giving us what we want. I personally would just drop this and mandate that tar files contain the directories for the files :)

mvo5 (commented):
Thank you! Just a quick quick comments, this looks very close!

cmd/osbuild-jobsite-builder/main.go
croissanne: Indeed, this is still there, will fix.

cmd/osbuild-jobsite-builder/main.go
```diff
@@ -206,6 +209,47 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
...
+				return fmt.Errorf("Builder.HandlerPopulate: unable to make dir in sources: %v", err)
+			}
+		case tar.TypeReg:
+			file, err := os.Create(path.Join(storePath, header.Name))
+			if err != nil {
+				return fmt.Errorf("Builder.HandlerPopulate: unable to open file in sources: %v", err)
+			}
+			defer file.Close()
+			for {
+				_, err := io.CopyN(file, tarReader, 4096)
```
mvo5, line 240: Why not just `io.Copy()` here? Is the concern  the buffersize of io.Copy() ? If so, maybe io.CopyBuffer but it should be okay, iirc io.Copy() uses a 32kb internal buffer which is not amazing but should be okay? or am I missing something?

croissanne: gosec complains :(

croissanne: https://pkg.go.dev/github.com/securego/gosec/v2#section-readme g110 specifically

mvo5: Thanks! I see, idea below. But I also think that gosec is silly here, we trust the inputs and also we cannot know how big or small a file in the tar should be so it's hard to guard against this.

maybe just: 
```go
diff --git a/cmd/osbuild-jobsite-builder/main.go b/cmd/osbuild-jobsite-builder/main.go
index 9037a6601..7e880b6ae 100644
--- a/cmd/osbuild-jobsite-builder/main.go
+++ b/cmd/osbuild-jobsite-builder/main.go
@@ -239,16 +239,12 @@ func (builder *Builder) HandlePopulate(w http.ResponseWriter, r *http.Request) e
                        }
                        defer file.Close()
                        // copy in defined chunks to avoid gosec#G110
-                       for {
-                               _, err := io.CopyN(file, tarReader, 4096)
-                               if err != nil {
-                                       if err == io.EOF {
-                                               break
-                                       }
-                                       return fmt.Errorf("Builder.HandlerPopulate: unable to write file in sources: %v", err)
-                               }
+                       if _, err := io.CopyN(file, tarReader, header.Size); err != nil {
+                               return fmt.Errorf("Builder.HandlerPopulate: unable to write file in sources: %v", err)
+                       }
+                       if err := file.Close(); err != nil {
+                               return fmt.Errorf("Builder.HandlerPopulate: unable to close file in sources: %v", err)
                        }
-                       file.Close()
                default:
                        return fmt.Errorf("Builder.HandlerPopulate: unexpected tar header type: %v", header.Typeflag)
                }

````
to avoid the loop? internally CopyN is just Copy with a limited reader so that should be fine (not tested but happy to write a test).

Comment (croissanne) [reply cmd/osbuild-jobsite-builder/main.go:243]: we can also just turn it off, maybe that's best. Will do that

File: cmd/osbuild-jobsite-manager/main.go [hunk 9]
```diff
@@ -264,13 +268,68 @@ func StepProvision(manifest []byte) error {
...
+				return err
+			}
+
+			// FileInfo only contains the basename, see https://pkg.go.dev/archive/tar#FileInfoHeader.
+			header.Name = strings.TrimPrefix(strings.Replace(filePath, argStore, "", -1), string(filepath.Separator))
+			if err := tw.WriteHeader(header); err != nil {
+				return err
+			}
 
+			if fi.Mode().IsRegular() {
```
supakeen, line 302: Error if it isn't?

croissanne: directories just need the header :) so this is fine I think.

supakeen (approved):

bcl (approved):
Looks ok as far as I understand things (I don't feel qualified to comment on the aws egress stuff). Just one typo found :)

PR #3995, opened 2024-03-06
templates/packer: vector expects a yaml configuration now

croissanne:
When switching to the new vector repos, the version was also bumped. The newer versions expect a yaml config.


thozza (approved):
LGTM

thozza:
Packer job in CI passed, we need to fix Fedora workers ASAP

PR #4003, opened 2024-03-11
Test/aws.sh: disable no AVCs test in CIV, which is a known issue

thozza:
The test case is consistently failing due to selinux-policy not being updated for what NM does.

Once https://issues.redhat.com/browse/RHEL-24346 is resolved, the filer should be removed.

This pull request includes:


test/cases/aws.sh
```diff
@@ -258,13 +258,15 @@ cp "${CIV_CONFIG_FILE}" "${TEMPDIR}/civ_config.yml"
...
 if nvrGreaterOrEqual "osbuild-composer" "83"; then
+    # TODO: remove test filtering using "-t" once https://issues.redhat.com/browse/RHEL-24346 is resolved
     sudo "${CONTAINER_RUNTIME}" run \
         -a stdout -a stderr \
         -e AWS_ACCESS_KEY_ID="${V2_AWS_ACCESS_KEY_ID}" \
         -e AWS_SECRET_ACCESS_KEY="${V2_AWS_SECRET_ACCESS_KEY}" \
         -e AWS_REGION="${AWS_REGION}" \
         -e JIRA_PAT="${JIRA_PAT}" \
         -v "${TEMPDIR}":/tmp:Z \
+        -t "not (TestsGeneric and test_no_avc_denials)" \
```
atodorov, line 269: That's one way of dealing with it although not my preferred one.

Do you mind opening a ticket for us to revert this change once the blocker is fixed? 

Otherwise LGTM.

thozza:
Hey @narmaku. I would appreciate your help with disabling a single test case in CIV... passing the filter in the format accepted by pytest results in `Error: invalid reference format`, see https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/6366005991#L3852

I tried to have a look at CIV's code, but have no clue where it is coming from and the output is pretty sparse even with us enabling the debug in the config used to run CIV...

narmaku:
> Hey @narmaku. I would appreciate your help with disabling a single test case in CIV... passing the filter in the format accepted by pytest results in `Error: invalid reference format`, see https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/6366005991#L3852
> 
> I tried to have a look at CIV's code, but have no clue where it is coming from and the output is pretty sparse even with us enabling the debug in the config used to run CIV...

Hi @thozza , the issue comes from the docker/podman command. The "-t" option added should be passed to the cloud-image-val, after the python command call. Right now the container is trying to interpret the -t option and it fails with that message.

Anyways, I think the best to do is to skip the test case based on the corresponding JIRA ticket (to be done in our side). Once resolved, we will remove that skip.

Example: https://github.com/osbuild/cloud-image-val/blob/c787db5d92980d78d2a1446d35294adbd1ea740b/test_suite/generic/test_generic.py#L829

Doc: https://github.com/osbuild/cloud-image-val/blob/c787db5d92980d78d2a1446d35294adbd1ea740b/README.md?plain=1#L176

test/cases/aws.sh
narmaku: I agree. As commented [here](https://github.com/osbuild/osbuild-composer/pull/4003#issuecomment-1990419242) I think the best is for us (cloudx team) to add the jira_skip marker with the corresponding bug in JIRA.

PR #4021, opened 2024-03-18
worker: drop backwards compatibility for DepsolveJob serialisation (COMPOSER-1612)

achilleas-k:
We've kept this around a lot longer than we should have.


internal/worker/json.go
```diff
@@ -177,81 +177,12 @@ func (pn *PipelineNames) All() []string {
 // into a single PackageSpec list in the result.  Each PackageSet defines the
 // repositories it will be depsolved against.
 type DepsolveJob struct {
-	PackageSets      map[string][]rpmmd.PackageSet `json:"grouped_package_sets"`
+	PackageSets      map[string][]rpmmd.PackageSet `json:"package_sets"`
```
ondrejbudai, line 180: Isn't this incompatible change? I might be missing something...

achilleas-k: Yes, thanks.  I read the compat function wrong.  The `package_sets` key is the name in the old format.

thozza (approved):
Makes sense :broom:

PR #4032, opened 2024-03-21
osbuild-jobsite-manager: close temp store file before deleting it

croissanne:


croissanne:
draft, as the http client closes the request body once sent, so we need to reimplement the retry

PR #4046, opened 2024-03-28
Create a nightly pipeline for GA/EUS releases

tkoscieln:
Separate RHEL GA/EUS releases testing to nightly pipeline from the pipeline ran on every PR.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -60,11 +60,15 @@ init:
 
 .upstream_rules_all:
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
 
 .upstream_rules_x86_64:
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
+
+.upstream_and_ga_rules_all:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER'
 
 .nightly_rules_all:
   rules:
@@ -73,6 +77,14 @@ init:
 .nightly_rules_x86_64:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.5-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+
+.ga_rules_all:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $NIGHTLY == "false"'
+
+.ga_rules_x86_64:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "false"'
 
 
 .RPM_RUNNERS_RHEL: &RPM_RUNNERS_RHEL
@@ -87,6 +99,7 @@ RPM:
   extends: .terraform
   rules:
     - !reference [.build_rules, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - sh "schutzbot/mockbuild.sh"
   parallel:
@@ -123,7 +136,8 @@ Container:
   stage: rpmbuild
   extends: .terraform
   rules:
-    - !reference [.build_rules, rules]
+    - !reference [.upstream_and_ga_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - sh "schutzbot/containerbuild.sh"
   parallel:
@@ -135,7 +149,8 @@ Packer:
   stage: test
   extends: .terraform
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - !reference [.upstream_and_ga_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - tools/ci-build-worker-packer.sh
   variables:
@@ -167,6 +182,7 @@ Base:
   rules:
     - !reference [.upstream_rules_all, rules]
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/base_tests.sh
@@ -192,6 +208,7 @@ Base:
   rules:
     - !reference [.upstream_rules_all, rules]
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/${SCRIPT}
@@ -229,23 +246,28 @@ regression-composer-works-behind-satellite:
 
 regression-excluded-dependency:
   extends: .regression
-  rules: 
-    # WHITELIST
-    - if: $RUNNER =~ "/^.*(rhel-8.*|rhel-9.*|centos-stream-9).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+  rules:
+    # WHITELIST & BLACKLIST: excluding GA runners from the PR pipeline
+    - if: $RUNNER =~ "/^.*(rhel-8.*|rhel-9.*|centos-stream-9).*$/" && $CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   variables:
     SCRIPT: regression-excluded-dependency.sh
 
 regression-include-excluded-packages:
   extends: .regression
   rules: 
-    # BLACKLIST: Skipped on fedora systems
-    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    # BLACKLIST: Skipped on fedora systems & excluding GA runners from the PR pipeline
+    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   variables:
     SCRIPT: regression-include-excluded-packages.sh
 
 regression-old-worker-new-composer:
+  rules:
+    - !reference [.upstream_and_ga_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   parallel:
     matrix:
       - RUNNER:
@@ -259,9 +281,10 @@ regression-old-worker-new-composer:
 regression-insecure-repo:
   extends: .regression
   rules:
-    # WHITELIST
-    - if: $RUNNER =~ "/^.*(rhel-*).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    # WHITELIST & BLACKLIST: excluding GA runners from the PR pipeline
+    - if: $RUNNER =~ "/^.*(rhel-*).*$/" && $CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   variables:
     SCRIPT: regression-insecure-repo.sh
 
@@ -269,9 +292,10 @@ regression-insecure-repo:
 regression-no-explicit-rootfs-definition:
   extends: .regression
   rules: 
-    # BLACKLIST: Skipped on fedora systems
-    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    # BLACKLIST: Skipped on fedora systems & excluding GA runners from the PR pipeline
+    - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   variables:
     SCRIPT: regression-no-explicit-rootfs-definition.sh
 
@@ -308,6 +332,7 @@ Trigger-rhel-edge-ci:
   rules:
     - !reference [.upstream_rules_all, rules]
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/${SCRIPT}
@@ -348,7 +373,8 @@ koji.sh (cloud upload):
   stage: test
   extends: .terraform
   rules:
-    - !reference [.upstream_rules_all, rules]
+    - !reference [.upstream_and_ga_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   script:
     - schutzbot/deploy.sh
     - /usr/libexec/tests/osbuild-composer/koji.sh cloud-upload ${CLOUD_TARGET} ${IMAGE_TYPE}
@@ -377,7 +403,8 @@ koji.sh (cloudapi):
   extends: .integration
   # Not supported in nightly pipelines
   rules:
-    - !reference [.upstream_rules_all, rules]
+    - !reference [.upstream_and_ga_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
   variables:
     SCRIPT: koji.sh
   parallel:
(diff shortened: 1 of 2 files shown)


tkoscieln:
Sample succesfull run of the GA/EUS Nightly pipeline (debug):
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1258487788

.gitlab-ci.yml
```diff
@@ -127,11 +136,28 @@ Build -tests RPM for RHEL:
...
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/rhel-8.4-ga-x86_64
+          - aws/rhel-8.4-ga-aarch64
+          - aws/rhel-8.9-ga-x86_64
+          - aws/rhel-8.9-ga-aarch64
+          - aws/rhel-9.3-ga-x86_64
+          - aws/rhel-9.3-ga-aarch64
+
```
jrusz, line 155: I don't think we need a new GA rpm build, we can just use the regular one with modified rules.

jrusz (changes requested):
Can you please rename it to just GA pipeline? The nightly pipeline is called that way because it's testing nightly composes while this is supposed to be testing GA composes which is something else, nightly GA doesn't make sense. So please also update all instances of "ga" and "nightly" used together to just "ga".

The rules don't seem to work, I see many jobs on GA runners on the regular PR pipeline.

atodorov (changes requested):

.gitlab-ci.yml
jrusz: @atodorov Why do we even build the rpms for GA? In case we need to do a z-stream release with some bug so to make sure the latest rpm from main still works on the given GA? Would we even ship such a new rpm? I suppose we should rather make a fix on the respective GA branch no? Should we maybe do the same as for nightly and that is use the package rpms that are shipped there and rebuild the `-tests` package only?

tkoscieln: @jrusz I named it nightly simply because of the fact they will run at night, but that can be ofc changed. I overlooked the rules not working for some of the jobs since in a few cases, they are overriden directly inside the jobs. Will investigate and fix that.

jrusz: The reason they run at night is because it's expected to have the least impact on daily productivity and also to have the results in the morning. The main reason behind the name is because it's running on nightly composes.

atodorov: > Why do we even build the rpms for GA? 

These are RPMs built from the software under test, regardless of which runner OS it is using.  My understanding is that with these new pipelines we're testing "latest osbuild-composer on already released distros". So you need to build them, no? 

> Would we even ship such a new rpm? 

Yes, it will go out as an advisory. 

> I suppose we should rather make a fix on the respective GA branch no? 

I think the fix should land first into `main` and then be cherry-picked into the relevant downstream branches. However this opens the question of what's the development model here and where do we want to be testing stuff. 

OTOH there is still value in knowing that latest osbuild-composer releases can be dropped into an older distro version. 

I think unltimately we'll need testing both ways but that's a bigger discussion than a GH PR review comment I think.

.gitlab-ci.yml
```diff
@@ -60,11 +60,11 @@ init:
 
 .upstream_rules_all:
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
 
 .upstream_rules_x86_64:
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
```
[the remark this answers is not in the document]
atodorov: > It is differentiated by having NIGHTLY_GA variable instead of NIGHTLY.

I see only 2 states: either testing latest nightly compose or a GA compose and there is already an environment variable which can represent 2 states (3 actualy). Unless we really need another variable let's not add one and reuse whatever already exists. `NIGHTLY == "false"` is the same as `GA == "true"`, isn't it ?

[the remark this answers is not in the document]
atodorov: >  NIGHTLY_GA

On naming: this is mighty confusing. The names are chosen based on which is the subkect under test, not when the pipelines execute. As a matter of fact they seem to execute around 4-6 AM in the morning which you can argue in the summer is during the day.

.gitlab-ci.yml
```diff
@@ -143,7 +158,7 @@ Packer:
   stage: test
   extends: .terraform
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule"'
+    - !reference [.ga/eus_rules_all, rules]
```
atodorov, line 161: shouldn't this also include    `- !reference [.upstream_rules_all, rules]` ?

.gitlab-ci.yml
```diff
@@ -269,10 +290,14 @@ regression-include-excluded-packages:
...
     - if: $RUNNER !~ "/^.*(fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
     - !reference [.nightly_rules_all, rules]
+    - !reference [.ga/eus_rules_all, rules]
+    - !reference [.ga/eus_rules_exclude, rules]
   variables:
     SCRIPT: regression-include-excluded-packages.sh
 
 regression-old-worker-new-composer:
+  rules:
+    - !reference [.ga/eus_rules_all, rules]
```
atodorov, line 300: Missing `    - !reference [.upstream_rules_all, rules]` ? or rather shouldn't this be the same as other `regression-` jobs ?

.gitlab-ci.yml
```diff
@@ -723,14 +782,35 @@ NIGHTLY_SUCCESS:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule" && $NIGHTLY == "true"'
   script:
-    - schutzbot/slack_notification.sh SUCCESS ":partymeow:"
+    - schutzbot/slack_notification.sh SUCCESS ":partymeow:" rhel
```
atodorov, line 785:
```suggestion
    - schutzbot/slack_notification.sh SUCCESS ":partymeow:" nightly
```

atodorov (changes requested):
Minor comments

schutzbot/slack_notification.sh
```diff
@@ -9,7 +9,11 @@ fi
 
 COMPOSE_ID=$(cat COMPOSE_ID)
 COMPOSER_NVR=$(cat COMPOSER_NVR)
-MESSAGE="\"Nightly pipeline execution on *$COMPOSE_ID* with *$COMPOSER_NVR* finished with status *$1* $2 \n QE: @atodorov, @jrusz\n Link to results: $CI_PIPELINE_URL\n For edge testing status please see https://url.corp.redhat.com/edge-pipelines \""
+if [ "$3" == "ga" ]; then
+    MESSAGE="\"Nightly GA/EUS releases pipeline execution finished with status *$1* $2 \n QE: @atodorov, @jrusz, @tkosciel\n Link to results: $CI_PIPELINE_URL \""
```
jrusz, line 13:
```suggestion
    MESSAGE="\"GA/EUS composes pipeline execution finished with status *$1* $2 \n QE: @atodorov, @jrusz, @tkosciel\n Link to results: $CI_PIPELINE_URL \""
```

jrusz (changes requested):
I'm thinking that it would be enough to just call it `GA` not `GA/EUS` because there is also `ELS` and idk what else and it's basically all a subset of GA. Every released RHEL version no matter the support level can still be considered GA. What do you think @atodorov ?

Also the rules don't seem to work, I see many GA jobs in the PR pipeline. The way these rules work by default is that whenever it matches the conditions it will schedule the job. Since you've added these new rules it will enabled to run it in the scheduled GA pipeline but there is nothing preventing it from running on PRs for that I think you'll need to modify the upstream rules to exclude these jobs

jrusz:
So what seems to work is if you take your exclude rule and add it to the upstream rules and build rules for example:
```
.build_rules:
  rules:
    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "false" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
    - if: '$CI_PIPELINE_SOURCE != "schedule" && $SKIP_CI == "true" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
      when: manual

.upstream_rules_all:
  rules:
    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
    ```
    There is also the upstream rule for x86_64 which will require a little bit different regex and then there are a few jobs which have their own rules defined like `aws.sh` and some regression jobs. Use https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/ci/lint to check out if it works, just paste the whole CI config there, tick the checkbox to `Simulate a pipeline created for the default branch` and then just search for any `-ga` jobs that are still there.

**Comment by atodorov** [comment]
> I'm thinking that it would be enough to just call it `GA` not `GA/EUS` because there is also `ELS` and idk what else and it's basically all a subset of GA. Every released RHEL version no matter the support level can still be considered GA. What do you think @atodorov ?

+1 for the proposed simplified naming. Let's use what you propose for now and update it later if need be.

File: .gitlab-ci.yml (hunk 3 above)
Comment (atodorov) [reply .gitlab-ci.yml:161]: hmmm, I don't really know what this job does but the old rule means "execute on PRs" (e.g. the job wasn't scheduled via pipeline schedules) while the new one means the opposite: "execute on this particular scheduled pipeline"

File: .gitlab-ci.yml [hunk 7]
```diff
@@ -131,7 +146,7 @@ Container:
   stage: rpmbuild
   extends: .terraform
   rules:
-    - !reference [.build_rules, rules]
+    - !reference [.ga/eus_rules_all, rules]
```
atodorov, line 149: Remove this change and keep the old rules.

We want to build the container image on pull requests to make sure that the build isn't broken, but we don't need the container when exercising the pipelines against a nightly RHEL compose or against the GA composes.

jrusz (changes requested):
I've added some comments in the code. One of the main issues I see right now is that I don't think it will run on `main` after a PR is merged. It's defined as one of the goals in the Jira ticket, we just want to remove these jobs from running on each PR.

@thozza I wanted to ask you if you agree with this change. The goal here is to not run tests on GA releases on PRs but instead run them only on `main` and in a schedule similar to current nightly pipelines. This will make PR testing faster with minimum impact to quality if and an issue does occur on some of the GA releases we'll see it reported on `main` and in Slack along with nightly pipelines.

thozza:
> @thozza I wanted to ask you if you agree with this change. The goal here is to not run tests on GA releases on PRs but instead run them only on `main` and in a schedule similar to current nightly pipelines. This will make PR testing faster with minimum impact to quality if and an issue does occur on some of the GA releases we'll see it reported on `main` and in Slack along with nightly pipelines.

We should keep testing on PRs at least for GA releases used in our SaaS deployment. Otherwise, this LGTM.

@ondrejbudai @croissanne WDYT?

.gitlab-ci.yml
```diff
@@ -76,6 +76,16 @@ init:
...
+.ga_rules_all:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $NIGHTLY== "false"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $NIGHTLY == "false"'
+
+.ga_rules_x86_64:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY== "false"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "false"'
+
```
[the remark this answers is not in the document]
atodorov: > Why not add a new variable like `GA == true` and check that instead?

@jrusz see https://github.com/osbuild/osbuild-composer/pull/4046#discussion_r1570998094. 

> If we add a new schedule which will also set NIGHTLY=false but will do something else it would trigger these jobs as well.

That's true but let's worry about it when it happens. Currently this is not what we intend to do.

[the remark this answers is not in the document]
jrusz: Ok fair enough, let's keep that part as it is now.

thozza:
So, for the SaaS version, we need to keep testing the **latest RHEL-9 GA release (9.4)**.

WRT the specific test cases, the following ones are relevant for the Saas:

- [api.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/api.sh)
- [koji.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/koji.sh)
- [multi-tenancy.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/multi-tenancy.sh)
- [regression-old-worker-new-composer.sh](https://github.com/osbuild/osbuild-composer/blob/main/test/cases/regression-old-worker-new-composer.sh)

@ondrejbudai @croissanne feel free to add / correct me.

atodorov:
@tkoscieln can you execute the new so called GA pipeline and post a link here so we can compare the results ?

tkoscieln:
I have updated the pipeline to re-add the tests specified by @thozza earlier.
How the pipeline looked before: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1347409921
How it looks now after I re-added some of the test cases: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1347426312
How the "GA" pipeline looks now:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1348619307

I have left the re-added test which now run in the PR pipeline again in the GA pipeline as well. What do you think @atodorov @jrusz ? Do we want to test them in both pipelines or just in the PR one?

atodorov:
> How the pipeline looked before: 

https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1348643668 (from pull #4232)

> How it looks now after I re-added some of the test cases: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1347426312 

* `Packer` job is missing, see comment above. I'm not quite certain if it needs to be executed on PRs but the original conditions indicate so and you are making it the opposite. 

* Not sure about `Container` either, need decision there.


The rest of the jobs in the list are the same, need to examine a bit better whether or not the expected runners are used.


> How the "GA" pipeline looks now: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1348619307

What is the expected result here? That test jobs are executed only on GA runners and not on nightly runners, isn't it ?

tkoscieln:
> > How the pipeline looked before:
> 
> https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1348643668 (from pull #4232) - **217 test jobs**
> 
> > How it looks now after I re-added some of the test cases: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1347426312
> 
> **^^^ 271 test jobs** - quite a lot more actually.

This number includes some cancelled jobs, so I checked and it is actually lower than 217.

>
> * `Packer` job is missing, see comment above. I'm not quite certain if it needs to be executed on PRs but the original conditions indicate so and you are making it the opposite.
> * Not sure about `Container` either, need decision there.
> 
> The rest of the jobs in the list are the same, need to examine a bit better whether or not the expected runners are used.
> 

I left `Packer` and `Container` jobs as they were not specified to be re-added, but can add them back as well.

> > How the "GA" pipeline looks now: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1348619307
> 
> What is the expected result here? That test jobs are executed only on GA runners and not on nightly runners, isn't it ?

Yes, it should contain jobs run on GA runners only, with some of the tests cases being extracted from the PR pipeline entirely to be run only in this GA pipeline.

atodorov:
> > > How the pipeline looked before:
> > **^^^ 271 test jobs** - quite a lot more actually.
> 
> This number includes some cancelled jobs, so I checked and it is actually lower than 217.

Can you git push this PR again so we can see the real number of jobs being scheduled. 

> I left `Packer` and `Container` jobs as they were not specified to be re-added, but can add them back as well.
> 

Container job should be executed on PRs to detect changes which break container builds


from `templates/packer/README.md`:

```
This directory contains a packer configuration for building osbuild-composer
worker AMIs based on RHEL.
```

So it looks like this job needs to be executed on pull requests as well, seems important to me.

tkoscieln:
@atodorov it seems that with this optimalization we are down to 136 jobs in the PR pipeline from 217 including the `Packer` and `Container` jobs.

atodorov (dismissed):
LGTM

.gitlab-ci.yml
```diff
@@ -54,17 +54,21 @@ init:
...
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
 
 .upstream_rules_x86_64:
   rules:
-    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/"'
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER =~ "/^.*(x86_64).*$/" && $RUNNER !~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/'
+
+.upstream_rules_ga_all:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE != "schedule" && $RUNNER'
```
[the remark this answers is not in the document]
jrusz: After looking through the file I realized it's upstream+ga, can you update the naming to make it more obvious?

jrusz (changes requested):
Looks good, just some cosmetic changes requested.

.gitlab-ci.yml
```diff
@@ -74,6 +78,16 @@ init:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.5-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
 
+.ga_rules_all:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $NIGHTLY== "false"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-[\S]+-(?:(?:ga)|(?:eus))[\S]+/ && $NIGHTLY == "false"'
```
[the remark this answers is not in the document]
tkoscieln: Propably none :D will delete

atodorov (dismissed):

atodorov (approved):

jrusz (approved):

PR #4047, opened 2024-03-28
main: journald support with external correlation

lzap:
This patch adds native journald support through logrus which enables to work with structured logs. It uses a dependency that is already in the project (`coreos/systemd-go`) which is a pure Go implementation. New log type is introduced named "journal" and it is now the default with fallback to "text" if journald is not present on the system.

Structured logging enabled me to work with logrus fields that can help with correlating log records on standalone or hosted deployments. Already existing "operation_id" field (random string) is now fully propagated via echo context logger and new logrus hook extracts it from Go standard context.

A new field called "external_id" is extracted from HTTP header `X-External-Id` when present. The idea is to set this to X-Edge-Request-Id on hosted image builder so logs can be easily correlated across both services.

Additionally, each cloud API request is now logged using the built-in echo logging middleware which creates entries like `Processed request GET /path`. Together with journald this can be tested end-to-end:

```
# curl -H "X-External-Id: XXXX" --unix-socket /run/cloudapi/api.socket -XGET http://localhost/api/image-builder-composer/v2/openapi
```

This will create the following entries:

```
# journalctl -e -o json-pretty --output-fields=OPERATION_ID,MESSAGE,EXTERNAL_ID
{
        "OPERATION_ID" : "2eJwe1mdLa9s2rB3PVZbVAttQi1",
        "EXTERNAL_ID" : "XXXX",
        "MESSAGE" : "Returning OpenAPI document"
}
{
        "MESSAGE" : "Processed request GET /api/image-builder-composer/v2/openapi",
        "EXTERNAL_ID" : "XXXX",
        "OPERATION_ID" : "2eJwe1mdLa9s2rB3PVZbVAttQi1",
}
```

Users can easily dig logs per request or operation:

```
# journalctl -u osbuild-composer OPERATION_ID=2eJwe1mdLa9s2rB3PVZbVAttQi1 -o cat
Returning OpenAPI document
Processed request GET /api/image-builder-composer/v2/openapi
```

TODO

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -6,6 +6,8 @@
 onto a system. We recommend doing this by building rpms, with:
 
 ```
+dnf install fedora-packager
+dnf builddep osbuild-composer
 make rpm
 ```
 
@@ -94,4 +96,22 @@ To rebuild the containers after a change, add the `--build` flag to the `docker-
 
 ```
 docker-compose up --build
+```
+
+## Shortening the loop
+
+For some components, it is possible to install distribution packages first and then only to replace binaries which may or may not work for smaller changes.
+
+```
+systemctl stop osbuild-composer.service osbuild-composer.socket osbuild-local-worker.socket
+make build && sudo install -m755 bin/osbuild-composer bin/osbuild-worker /usr/libexec/osbuild-composer/
+systemctl start osbuild-composer.socket osbuild-local-worker.socket
+```
+
+## Accessing Cloud API
+
+You can use curl to access the Cloud API:
+
+```
+curl --unix-socket /run/cloudapi/api.socket -XGET http://localhost/api/image-builder-composer/v2/openapi
 ```
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -3,11 +3,15 @@ package main
 import (
 	"context"
 	"flag"
+	"io"
+	"log"
 	"os"
 
 	"github.com/coreos/go-systemd/activation"
+	"github.com/coreos/go-systemd/journal"
 	"github.com/getsentry/sentry-go"
 	sentrylogrus "github.com/getsentry/sentry-go/logrus"
+	"github.com/osbuild/osbuild-composer/internal/common"
 	slogger "github.com/osbuild/osbuild-composer/pkg/splunk_logger"
 	"github.com/sirupsen/logrus"
 )
@@ -27,6 +31,11 @@ func main() {
 	var verbose bool
 	flag.BoolVar(&verbose, "verbose", false, "Print access log")
 	flag.Parse()
+
+	// Redirect Go standard logger into logrus before it's used by other packages
+	log.SetOutput(common.Logger())
+	// Ensure the Go standard logger does not have any prefix or timestamp
+	log.SetFlags(0)
 
 	if !verbose {
 		logrus.Print("verbose flag is provided for backward compatibility only, current behavior is always printing the access log")
@@ -49,6 +58,16 @@ func main() {
 	}
 
 	switch config.LogFormat {
+	case "journal":
+		// If we are running under systemd, use the journal. Otherwise,
+		// fallback to text formatter.
+		if journal.Enabled() {
+			logrus.SetFormatter(&logrus.JSONFormatter{})
+			logrus.AddHook(&common.JournalHook{})
+			logrus.SetOutput(io.Discard)
+		} else {
+			logrus.SetFormatter(&logrus.TextFormatter{})
+		}
 	case "text":
 		logrus.SetFormatter(&logrus.TextFormatter{})
 	case "json":
--- a/cmd/osbuild-composer/main.go
+++ b/cmd/osbuild-composer/main.go
@@ -50,6 +50,8 @@ func main() {
 	logLevel, err := logrus.ParseLevel(config.LogLevel)
 
 	logrus.SetReportCaller(true)
+	// Add context hook to log operation_id and external_id
+	logrus.AddHook(&common.ContextHook{})
 
 	if err == nil {
 		logrus.SetLevel(logLevel)
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -87,9 +87,36 @@ func (s *Server) Handler(path string) http.Handler {
 	e := echo.New()
 	e.Binder = binder{}
 	e.HTTPErrorHandler = s.HTTPErrorHandler
-	e.Pre(common.OperationIDMiddleware)
-	e.Use(middleware.Recover())
 	e.Logger = common.Logger()
+
+	// OperationIDMiddleware - generates OperationID random string and puts it into the contexts
+	// ExternalIDMiddleware - extracts ID from HTTP header and puts it into the contexts
+	// LoggerMiddleware - creates context-aware logger for each request
+	e.Pre(common.OperationIDMiddleware, common.ExternalIDMiddleware, common.LoggerMiddleware)
+	e.Use(middleware.Recover())
+
+	e.Use(middleware.RequestLoggerWithConfig(middleware.RequestLoggerConfig{
+		LogURI:     true,
+		LogStatus:  true,
+		LogLatency: true,
+		LogMethod:  true,
+		LogValuesFunc: func(c echo.Context, values middleware.RequestLoggerValues) error {
+			fields := logrus.Fields{
+				"uri":          values.URI,
+				"method":       values.Method,
+				"status":       values.Status,
+				"latency_ms":   values.Latency.Milliseconds(),
+				"operation_id": c.Get(common.OperationIDKey),
+				"external_id":  c.Get(common.ExternalIDKey),
+			}
+			if values.Error != nil {
+				fields["error"] = values.Error
+			}
+			logrus.WithFields(fields).Infof("Processed request %s %s", values.Method, values.URI)
+
+			return nil
+		},
+	}))
 
 	if sentry.CurrentHub().Client() == nil {
 		logrus.Warn("Sentry/Glitchtip not initialized, echo middleware was not enabled")
--- a/internal/common/echo_logrus.go
+++ b/internal/common/echo_logrus.go
@@ -1,39 +1,42 @@
 package common
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 
-	"github.com/labstack/gommon/log"
+	lslog "github.com/labstack/gommon/log"
 	"github.com/sirupsen/logrus"
 )
 
 // EchoLogrusLogger extend logrus.Logger
 type EchoLogrusLogger struct {
 	*logrus.Logger
+	Ctx context.Context
 }
 
 var commonLogger = &EchoLogrusLogger{
 	Logger: logrus.StandardLogger(),
+	Ctx:    context.Background(),
 }
 
 func Logger() *EchoLogrusLogger {
 	return commonLogger
 }
 
-func toEchoLevel(level logrus.Level) log.Lvl {
+func toEchoLevel(level logrus.Level) lslog.Lvl {
 	switch level {
 	case logrus.DebugLevel:
-		return log.DEBUG
+		return lslog.DEBUG
 	case logrus.InfoLevel:
-		return log.INFO
+		return lslog.INFO
 	case logrus.WarnLevel:
-		return log.WARN
+		return lslog.WARN
 	case logrus.ErrorLevel:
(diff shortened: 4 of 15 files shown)


cmd/osbuild-composer/composer.go
```diff
@@ -402,7 +402,7 @@ func (c *Composer) Start() error {
 func (c *Composer) ensureStateDirectory(name string, perm os.FileMode) (string, error) {
 	d := path.Join(c.stateDir, name)
 
-	err := os.Mkdir(d, perm)
+	err := os.MkdirAll(d, perm)
```
lzap, line 405: Can probably drop this, I was trying to run composer from source locally but I gave up.

HACKING.md
```diff
@@ -94,4 +96,22 @@ To rebuild the containers after a change, add the `--build` flag to the `docker-
...
+systemctl start osbuild-composer.socket osbuild-local-worker.socket
+```
+
+## Accessing Cloud API
+
+You can use curl to access the Cloud API:
+
+```
+curl --unix-socket /run/cloudapi/api.socket -XGET http://localhost/api/image-builder-composer/v2/openapi
+```
```
lzap, line 117: These things I had to find the hard way, I believe we should put them somewhere.

cmd/osbuild-composer/main.go
```diff
@@ -41,6 +49,7 @@ func main() {
 	logLevel, err := logrus.ParseLevel(config.LogLevel)
 
 	logrus.SetReportCaller(true)
+	logrus.AddHook(&common.ContextHook{})
```
lzap, line 52: This hooks puts operation_id and external_id fields into the event.

cmd/osbuild-composer/main.go
```diff
@@ -49,6 +58,14 @@ func main() {
...
 
 	switch config.LogFormat {
+	case "journal":
+		if journal.Enabled() {
+			logrus.SetFormatter(&logrus.JSONFormatter{})
+			logrus.AddHook(&common.JournalHook{})
+			logrus.SetOutput(io.Discard)
+		} else {
+			logrus.SetFormatter(&logrus.TextFormatter{})
+		}
```
lzap, line 68: Fallback to "text" setting if systemd is not present.

internal/cloudapi/v2/handler.go
```diff
@@ -43,6 +43,8 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
 }
 
 func (h *apiHandlers) GetOpenapi(ctx echo.Context) error {
+	ctx.Logger().Info("Returning OpenAPI document")
+
```
lzap, line 47: I used this message for debugging, not very useful, can drop.

internal/cloudapi/v2/server.go
```diff
@@ -87,9 +87,31 @@ func (s *Server) Handler(path string) http.Handler {
...
+				"external_id":  c.Get(common.ExternalIDKey),
+			}
+			if values.Error != nil {
+				fields["error"] = values.Error
+			}
+			logrus.WithFields(fields).Infof("Processed request %s %s", values.Method, values.URI)
+
+			return nil
+		},
+	}))
```
lzap, line 114: This is echo middleware helper for info logs, this is used already in the image builder and quite useful.

internal/common/echo_logrus.go
```diff
@@ -1,39 +1,42 @@
...
 
 // EchoLogrusLogger extend logrus.Logger
 type EchoLogrusLogger struct {
 	*logrus.Logger
+	Ctx context.Context
 }
 
 var commonLogger = &EchoLogrusLogger{
 	Logger: logrus.StandardLogger(),
+	Ctx:    context.Background(),
```
lzap, line 20: Key change of this patch - context is added to the EchoLogrusLogger via middleware so it can be passed into all logging functions. Unfortunately, echo logger is not context aware so this is the only way to do it.

internal/common/echo_logrus.go
```diff
@@ -63,113 +66,117 @@ func (l *EchoLogrusLogger) SetPrefix(p string) {
...
 	if err != nil {
 		panic(err)
 	}
-	l.Logger.Panicln(string(b))
+	l.Logger.WithContext(l.Ctx).Panicln(string(b))
+}
+
+func (l *EchoLogrusLogger) Write(p []byte) (n int, err error) {
+	return l.Logger.WithContext(l.Ctx).Writer().Write(p)
 }
```
lzap, line 182: The Writer interface is needed to be able to forward Go log entries (from images library) into logrus directly.

internal/common/journal_hook.go
```diff
@@ -0,0 +1,72 @@
...
+func (hook *JournalHook) Levels() []logrus.Level {
+	return []logrus.Level{
+		logrus.PanicLevel,
+		logrus.FatalLevel,
+		logrus.ErrorLevel,
+		logrus.WarnLevel,
+		logrus.InfoLevel,
+		logrus.DebugLevel,
+	}
+}
```
lzap, line 72: This is all of the journald implementation, it is from github.com/wercker/journalhook with small changes, I do not think these 70 lines are worth introducing a new dependency.

internal/common/external_id.go
```diff
@@ -0,0 +1,30 @@
...
+
+		eid = strings.TrimSuffix(eid, "\n")
+		c.Set(ExternalIDKey, eid)
+
+		ctx := c.Request().Context()
+		ctx = context.WithValue(ctx, externalIDKeyCtx, eid)
+		c.SetRequest(c.Request().WithContext(ctx))
+
+		return next(c)
+	}
```
lzap, line 29: Both operation and external ids must be passed into both echo context and Go standard context. This is because logrus only supports Go context while echo middleware logger only supports echo context.

internal/common/journal_hook.go
```diff
@@ -0,0 +1,72 @@
...
+	case r >= 'a' && r <= 'z':
+		return r - 32
+	default:
+		return rune('_')
+	}
+}
+
+func stringifyKey(key string) string {
+	key = strings.Map(stringifyOp, key)
+	if strings.HasPrefix(key, "_") {
```
ezr-ondrej, line 42: Linter disagrees here :)

lzap: Sure, I will update linter nitpicks after first round of review. THanks.

HACKING.md
achilleas-k: IMO, this is the right place.

cmd/osbuild-composer/main.go
achilleas-k: Since this isn't obvious, the comment would be more useful in the code instead of the in-line PR comments.

cmd/osbuild-composer/main.go
achilleas-k: The comment would be more useful in the code instead of the in-line PR comments.

internal/cloudapi/v2/server.go
```diff
@@ -87,9 +87,31 @@ func (s *Server) Handler(path string) http.Handler {
 	e := echo.New()
 	e.Binder = binder{}
 	e.HTTPErrorHandler = s.HTTPErrorHandler
-	e.Pre(common.OperationIDMiddleware)
+	e.Pre(common.OperationIDMiddleware, common.ExternalIDMiddleware, common.LoggerMiddleware)
```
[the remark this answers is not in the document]
achilleas-k: This comment reads more to me like a descriptive commit message for a commit that only includes this change.

achilleas-k (commented):
I think this is a decent idea and will make log reading a lot less unpleasant.
I'd like to see the PR broken into more readable smaller commits though.  Some of the inline comments could be in-code since they clarify things that are happening implicitly and aren't going to be very useful as PR comments in the long run.
Also, some of the in-line comments are nice explanations for _why_ a change happened, which is often very useful to have as a commit message for that standalone patch.

cmd/osbuild-composer/composer.go
lzap: Dropping.

lzap:
Moved all comments into the codebase, broken into three commits:

* Documentation and gitignore update
* Adding journald native support
* Propagating operation and external id through context to logs (this one is hard to break up into two as I had to do updates to existing operation id in order to efficiently implement external id)

HACKING.md
```diff
@@ -94,4 +96,22 @@ To rebuild the containers after a change, add the `--build` flag to the `docker-
...
\ No newline at end of file
+```
+
+## Shortening the loop
+
+For some components, it is possible to install distribution packages first and then only to replace binaries which may or may not work for smaller changes.
+
+```
+systemctl stop osbuild-composer.service osbuild-composer.socket osbuild-local-worker.socket
+go build ./cmd/osbuild-composer && cp osbuild-composer /usr/libexec/osbuild-composer/
```
achilleas-k, line 107: TEENY TINY NITPICK: we have a `build` target in the `Makefile` that builds all the binaries under `bin/`.  It's annoying to use in this case because it builds a lot more binaries than we need here, but maybe it'd be nice for consistency to do `go build -o bin/ ...`.  That would also remove the need for adding the binaries to `.gitignore`.

And then also using `sudo install -m 755 ...` instead of `cp` to make sure perms are correct (the user's umask might have group and other modes blank, for example).

Not really necessary though.  We can change it later.

achilleas-k (dismissed):
Thanks for tidying up.  LGTM!

Tiny note about the documantation, but can be changed in a follow-up, or even ignored.

HACKING.md
lzap: Done

HACKING.md
```diff
@@ -94,4 +96,22 @@ To rebuild the containers after a change, add the `--build` flag to the `docker-
...
\ No newline at end of file
+```
+
+## Shortening the loop
+
+For some components, it is possible to install distribution packages first and then only to replace binaries which may or may not work for smaller changes.
+
+```
+systemctl stop osbuild-composer.service osbuild-composer.socket osbuild-local-worker.socket
+make build && sudo install -m755 osbuild-composer /usr/libexec/osbuild-composer/
```
achilleas-k, line 107: The install now should be
```
sudo install -m755 bin/osbuild-composer ...
```
Also would probably be useful to also `install ... bin/osbuild-worker` at the same time.

achilleas-k (changes requested):

achilleas-k (approved):
LGTM

achilleas-k:
@lzap We don't do merge commits in these repos.  Can you drop the merge commit and rebase instead?

achilleas-k (changes requested):
@lzap We don't do merge commits in these repos.  Can you drop the merge commit and rebase instead?

lzap:
Damn that was a misclick, amended a rebase.

achilleas-k (dismissed):

achilleas-k:
> and then an AWS test which I don’t understand why fails

`iot-bootable-container` failed with
```
dial tcp: lookup cdn03.quay.io: no such host
```
which looks like a plain old network hiccup.  Annoying.

lzap:
@ezr-ondrej please do not rebase, @achilleas-k plans to only restart failed tests. is that right? thanks.

achilleas-k:
> @ezr-ondrej please do not rebase, @achilleas-k plans to only restart failed tests. is that right? thanks.

It's a tricky decision.  It's good to rebase when there's a lot of changes in main and there's a chance the rebase might break things, but OTOH the CI runtimes are long and annoying.

A merge queue would be great but that doesn't work with flaky testing.

ezr-ondrej:
Yeah, sorry! I've got new buttons in Gitlab now, I'll push those instead xD

lzap:
Only schulzbot is red, seems only 8.4 is red, likely not caused by this PR.

achilleas-k:
There's a known issue with that test but I thought we fixed it already.

lzap:
Rebased.

ezr-ondrej:
The main pipeline is green!! Woooot!

lzap:
Well bunch of tests failed, someone help me to understand if these are relevant. These logs are hard to read.

achilleas-k:
Since all the edge tests are failing but the main gitlab CI pipeline succeeded, it might something with the edge tests themselves or something in the PR changed something the tests rely on.

I'm seeing at least one instance of the issue described in #4094

lzap:
Just `rpm-build:centos-stream-10-ppc64le` failed

achilleas-k (approved):

ezr-ondrej:
Thank you @achilleas-k and @lzap ! :orange_heart:

PR #4056, opened 2024-04-04
Makefile: implement "fully source containers" HMS-3883

schuellerf:
Getting the stack up and running locally but all from source

also contains a followup of COMPOSER-2185 just to get it compiling


schuellerf:
Note: unit tests fail due to inconsistencies with the `images` repo - "fixed" in 
https://github.com/osbuild/osbuild-composer/pull/4056/commits/eef36f74e0f5919a710fcf550953d07ef38a127c
but the testframework apperently uses the old version of `images`

Makefile
```diff
@@ -326,3 +330,96 @@ $(PROCESSED_TEMPLATE_DIR)/%.yml: $(PROCESSED_TEMPLATE_DIR) $(OPENSHIFT_TEMPLATES
 .PHONY: process-templates
 process-templates: $(addprefix $(PROCESSED_TEMPLATE_DIR)/, $(OPENSHIFT_TEMPLATES))
 
+# either "docker" or "sudo podman"
+# podman needs to build as root as it also needs to run as root afterwards
+CONTAINER_EXECUTABLE ?= docker
```
bcl, line 335: Like I mentioned in the weldr-client version of this, we should be using podman by default.

bcl:
Hm, so just quickly trying `make container_composer.dev` here and this appears to assume things about my repo layouts -- it is trying to run `make container.dev` in ../osbuild which fails because I don't have the PR for osbuild checked out. I also don't really like the idea of one repo touching paths outside of it (whether or not it is dangerous in practice, it is unexpected).

schuellerf:
The starting point of building the stack "fully from source" and the layout is described [here](https://github.com/schuellerf/osbuild-getting-started/tree/HMS-3883-fully-source-containers/service#setup)
but you are right all layers already depending on outside source need this description too.
That's why I wanted/need to merge them in-order (as mentioned [here](https://issues.redhat.com/browse/HMS-3883?focusedId=25515771&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25515771) ) 
so I'll revert this to draft and continue the process including the discussions in-order

PR #4061, opened 2024-04-08
templates/packer: set http(s)_proxy environment variable in unit HMS-3798

croissanne:
The snapshotted repositories on stage are behind a VPN + a proxy.


supakeen (approved):
Looks good to me.

PR #4068, opened 2024-04-10
rm dnf-json

achilleas-k:
The depsolver is now in osbuild and packaged as osbuild-depsolve-dnf. Let's remove the old copy from here to avoid confusion and potential divergence (which we already had happen before).


croissanne (approved):

bcl (approved):
Ack!

thozza (approved):
Thanks! :broom:

supakeen (approved):
remov snek

PR #4083, opened 2024-04-17
osbuildexecutor/ec2: pass extraEnv only to sources invocation

croissanne:
The extra env only contains paths to secrets which are needed to pull sources. Since the sources don't get pulled on the executor, don't pass the env along to the executor.


achilleas-k (approved):

ochosi (approved):

croissanne:
workerexecutor test worked, so merged

PR #4086, opened 2024-04-18
osbuild-worker: fix nil pointer in depsolve job

croissanne:
Yesterday in an effort to get the snapshotting to work quickly, I ignored the results of the api tests, turns out that was a bad idea.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -52,7 +52,7 @@ type DepsolveJobImpl struct {
 // (matching map keys).
 func (impl *DepsolveJobImpl) depsolve(packageSets map[string][]rpmmd.PackageSet, modulePlatformID, arch, releasever string) (map[string][]rpmmd.PackageSpec, map[string][]rpmmd.RepoConfig, error) {
 	solver := impl.Solver.NewWithConfig(modulePlatformID, releasever, arch, "")
-	if impl.RepositoryMTLSConfig.Proxy != nil {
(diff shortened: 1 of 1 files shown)


schuellerf (approved):

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -52,7 +52,7 @@ type DepsolveJobImpl struct {
 // (matching map keys).
 func (impl *DepsolveJobImpl) depsolve(packageSets map[string][]rpmmd.PackageSet, modulePlatformID, arch, releasever string) (map[string][]rpmmd.PackageSpec, map[string][]rpmmd.RepoConfig, error) {
 	solver := impl.Solver.NewWithConfig(modulePlatformID, releasever, arch, "")
-	if impl.RepositoryMTLSConfig.Proxy != nil {
+	if impl.RepositoryMTLSConfig != nil && impl.RepositoryMTLSConfig.Proxy != nil {
```
schuellerf, line 55:
```suggestion
	if impl.RepositoryMTLSConfig?.Proxy != nil {
```
Would work as well, right?

croissanne: does that operator exist in go?

schuellerf: sorry - mixed it up...

thozza (approved):
LGTM

cmd/osbuild-worker/jobimpl-depsolve.go
croissanne: ye, js only :>

PR #4089, opened 2024-04-19
Active worker gauge (COMPOSER-2215)

croissanne:
Adds an active worker gauge, increments when workers register, decrements when workers timeout. This will be useful to create an alert around (if no workers are connecting), and would be cool to have on the dashboard ($x active workers).


kingsleyzissou (approved):
Nice!

PR #4097, opened 2024-04-23
dashboards/worker-general: rearrange worker status (COMPOSER-2215)

schuellerf:
...rearrange viz like the others: statistic first, then the time series

also show architecture info in the "active worker" time series


schuellerf:
... just some idea to get the view more uniform

PR #4100, opened 2024-04-24
templates/dashboards: fix community-stage tenant variable

croissanne:
A space is necessary before and after the colon separating the key and the value.

[skip ci]


schuellerf (approved):
LGTM

PR #4101, opened 2024-04-24
templates/packer: add failure script

croissanne:
In case the service failed, set the instance to unhealthy.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization.service
+++ b/templates/packer/ansible/roles/common/files/worker-initialization.service
@@ -21,6 +21,7 @@ ExecStart=/usr/local/libexec/worker-initialization-scripts/get_oci_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/get_pulp_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/get_ldap_sa_mtls_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/worker_service.sh
+ExecStopPost=/usr/local/libexec/worker-initialization-scripts/on_exit.sh
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


templates/packer/ansible/roles/common/files/worker-initialization.service
```diff
@@ -21,6 +21,7 @@ ExecStart=/usr/local/libexec/worker-initialization-scripts/get_oci_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/get_pulp_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/get_ldap_sa_mtls_creds.sh
 ExecStart=/usr/local/libexec/worker-initialization-scripts/worker_service.sh
+ExecStopPost=/usr/local/libexec/worker-initialization-scripts/on_failure.sh
```
ondrejbudai, line 24: Are you sure that this won't be run on successful runs? `man systemd.service` kinda implies that services with `Type=oneshot` are considered dead if they don't have `RemainAfterExit=` set.

> Behavior of oneshot is similar to simple; however, the service manager will consider the unit up after the main process exits. It will then start follow-up units.  RemainAfterExit= is particularly useful for this type of service.  Type=oneshot is the implied default if neither Type= nor
               ExecStart= are specified. Note that if this option is used without RemainAfterExit= the service will never enter "active" unit state, but will directly transition from "activating" to "deactivating" or "dead", since no process is configured that shall run continuously. In particular this means
               that after a service of this type ran (and which has RemainAfterExit= not set) it will not show up as started afterwards, but as dead.

I suggest trying this locally. If you already did, please ignore me. :)

croissanne: egh no you're right
I think OnFailure with a separate service might be the way to go. Or checking the service_result.

thozza (approved):
Looks good :+1:

ondrejbudai (approved):
Splendid, thanks!

PR #4102, opened 2024-04-24
templates/openshift: split worker from composer maintenance

croissanne:
The main reason for this split is that we can start scheduling maintenance jobs per tenant we have in app-intf.  Currently the composer maintenance job also maintains the cloud accounts for the CRC tenant, while composer should be tenant-agnostic. The composer maintenance job should just maintain the DB, and then the specific worker maintenance job should maintain the worker resources.

As soon as this is split, we can start deploying it and disabling the aws/gcp maintenance in the composer namespace, as well as expanding the aws maintenance to look for leftover resources related to the worker-executor stuff.


schuellerf:
Can't really "test" but looks sane

schuellerf (approved):

schuellerf (approved):

PR #4103, opened 2024-04-24
Makefile for templates/openshift

schuellerf:
I just wanted to check #4102 locally.
maybe that's helpful for someone

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -276,15 +276,7 @@ jobs:
     - uses: redhat-actions/oc-installer@v1
     - name: Process template
       run: |
-        mkdir processed-templates
-        oc process -f templates/openshift/composer.yml \
-          -p IMAGE_TAG=image_tag \
-          --local \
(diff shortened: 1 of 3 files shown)


croissanne (commented):
Hmm I wonder if this makes sense to put in the makefile. The only reason we process them in the github workflow is so that kubelinter can do a very basic check. The processing of the template when actually deploying is something that is out of our hands. If we put it in the makefile, then i'd just consider it as a helper for the github workflow.

schuellerf:
We could use it from GitHub Workflow but I just wanted to call it so see if there are syntax problems locally (avoid GitHub Action problems before pushing - the usual problem)

bcl:
I like having the ability to run things from the Makefile. I'd also add a commit that converts the workflow to use the Makefile target so that things don't drift.

bcl (approved):
Looks Good!

bcl (approved):
Looks good!

PR #4105, opened 2024-04-25
test/old-worker-new-composer: reset osbuild repo for image-info

croissanne:
Image-info is installed via the latest test rpms, which need a newer osbuild version.


achilleas-k (approved):
The repo shuffling is a bit confusing but LGTM

croissanne:
> The repo shuffling is a bit confusing but LGTM

wonder if we could just use old image-info?

achilleas-k:
> > The repo shuffling is a bit confusing but LGTM
> 
> wonder if we could just use old image-info?

It doesn't tend to change that often, so it could.  But this is safer.

PR #4112, opened 2024-04-29
templates/openshift: make the maintenance template generic

croissanne:
We could deploy this job for both composer and each tenant's workers that's present in app-intf. Then we can remove the maintenance bits from the composer template.

### DIFF
--- a/templates/openshift/maintenance.yml
+++ b/templates/openshift/maintenance.yml
@@ -1,23 +1,23 @@
 apiVersion: v1
 kind: Template
 metadata:
-  name: composer
+  name: image-builder-maintenance
   annotations:
-    openshift.io/display-name: Image Builder CRC worker services
+    openshift.io/display-name: Image Builder maintenance
     description: |
-      Services related to maintaining the CRC workers.
+      Cronjob related to maintaining both composer and the workers.
     tags: golang
     iconClass: icon-shadowman
     template.openshift.io/provider-display-name: Red Hat, Inc.
 labels:
-  template: composer-worker-maintenance
+  template: image-builder-maintenance
 objects:
 - apiVersion: batch/v1
   kind: CronJob
   metadata:
     labels:
       service: image-builder
-    name: crc-worker-maintenance
+    name: image-builder-maintenance
   spec:
     # run maintenance job at midnight
     schedule: 0 0 * * *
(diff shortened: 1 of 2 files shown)


templates/openshift/maintenance.yml
```diff
@@ -1,23 +1,23 @@
...
-  template: composer-worker-maintenance
+  template: image-builder-maintenance
 objects:
 - apiVersion: batch/v1
   kind: CronJob
   metadata:
     labels:
       service: image-builder
-    name: crc-worker-maintenance
+    name: image-builder-maintenance
```
schuellerf, line 20: also rename this in line 35 ?

schuellerf (changes requested):
Will the DB maintenance then be remove from the composer.yml?

croissanne:
> Will the DB maintenance then be remove from the composer.yml?

exactly, we'll just deploy the maintenance template twice, once for composer and once for the workers.

schuellerf:
probably with
```
- name: maintenance
  url: https://github.com/osbuild/osbuild-composer
  path: /templates/maintenance.yml
  targets:
  - namespace:
      $ref: /services/image-builder/namespaces/workers-stage.yml
    upstream:
      instance:
        $ref: /dependencies/ci-ext/ci-ext.yml
      name: osbuild-osbuild-composer-gh-build-main
    ref: main
    parameters:
      PGSSLMODE: "require"
      MAINTENANCE_DRY_RUN: "true"
      ENABLE_AWS_MAINTENANCE: "true"
      ENABLE_GCP_MAINTENANCE: "true"
>      ENABLE_DB_MAINTENANCE: "false"
  - namespace:
      $ref: /services/image-builder/namespaces/composer-stage.yml
    upstream:
      instance:
        $ref: /dependencies/ci-ext/ci-ext.yml
      name: osbuild-osbuild-composer-gh-build-main
    ref: main
    parameters:
      PGSSLMODE: "require"
      MAINTENANCE_DRY_RUN: "true"
      ENABLE_DB_MAINTENANCE: "true"
```
as ENABLE_DB_MAINTENANCE is mandatory, right? 🤔

schuellerf (approved):
LGTM

croissanne:
> 
> as ENABLE_DB_MAINTENANCE is mandatory, right? 🤔

yea but the template sets it to false by default (required & default value just prevent it from being set to "")

PR #4113, opened 2024-04-29
osbuild-worker: improve error "reason" in case of stage failures (HMS-1442)

schuellerf:
This should improve the error in the frontend from

stage_error_before

to

stage_error_after

There is no test framework to test changes like this, right?

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -77,23 +77,31 @@ func workerClientErrorFrom(err error) (*clienterrors.Error, error) {
 	switch e := err.(type) {
 	case dnfjson.Error:
 		// Error originates from dnf-json
+		reason := fmt.Sprintf("DNF error occurred: %s", e.Kind)
+		details := e.Reason
 		switch e.Kind {
 		case "DepsolveError":
-			return clienterrors.WorkerClientError(clienterrors.ErrorDNFDepsolveError, err.Error(), e.Reason), nil
+			return clienterrors.WorkerClientError(clienterrors.ErrorDNFDepsolveError, reason, details), nil
 		case "MarkingErrors":
-			return clienterrors.WorkerClientError(clienterrors.ErrorDNFMarkingErrors, err.Error(), e.Reason), nil
+			return clienterrors.WorkerClientError(clienterrors.ErrorDNFMarkingErrors, reason, details), nil
 		case "RepoError":
-			return clienterrors.WorkerClientError(clienterrors.ErrorDNFRepoError, err.Error(), e.Reason), nil
+			return clienterrors.WorkerClientError(clienterrors.ErrorDNFRepoError, reason, details), nil
 		default:
 			err := fmt.Errorf("Unhandled dnf-json error in depsolve job: %v", err)
 			// This still has the kind/reason format but a kind that's returned
 			// by dnf-json and not explicitly handled here.
-			return clienterrors.WorkerClientError(clienterrors.ErrorDNFOtherError, err.Error(), e.Reason), err
+			return clienterrors.WorkerClientError(clienterrors.ErrorDNFOtherError, reason, details), err
 		}
 	default:
-		err := fmt.Errorf("rpmmd error in depsolve job: %v", err)
+		reason := "rpmmd error in depsolve job"
+		details := "<nil>"
+		if err == nil {
+			err = fmt.Errorf("workerClientErrorFrom expected an error to be processed. Not nil")
+		} else {
+			details = err.Error()
+		}
 		// Error originates from internal/rpmmd, not from dnf-json
-		return clienterrors.WorkerClientError(clienterrors.ErrorRPMMDError, err.Error(), nil), err
+		return clienterrors.WorkerClientError(clienterrors.ErrorRPMMDError, reason, details), err
 	}
 }
 
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -351,6 +351,33 @@ func (impl *OSBuildJobImpl) getPulpClient(targetOptions *target.PulpOSTreeTarget
 	return pulp.NewClientFromFile(address, impl.PulpConfig.CredsFilePath)
 }
 
+func makeJobErrorFromOsbuildOutput(osbuildOutput *osbuild.Result) *clienterrors.Error {
+	var osbErrors []string
+	if osbuildOutput.Error != nil {
+		osbErrors = append(osbErrors, fmt.Sprintf("osbuild error: %s", string(osbuildOutput.Error)))
+	}
+	if osbuildOutput.Errors != nil {
+		for _, err := range osbuildOutput.Errors {
+			osbErrors = append(osbErrors, fmt.Sprintf("manifest validation error: %v", err))
+		}
+	}
+	var failedStage string
+	for _, pipelineLog := range osbuildOutput.Log {
+		for _, stageResult := range pipelineLog {
+			if !stageResult.Success {
+				failedStage = stageResult.Type
+				break
+			}
+		}
+	}
+
+	reason := "osbuild build failed"
+	if len(failedStage) > 0 {
+		reason += " in stage:\n" + failedStage
+	}
+	return clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, reason, osbErrors)
+}
+
 func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 	logWithId := logrus.WithField("jobId", job.Id().String())
 	// Initialize variable needed for reporting back to osbuild-composer.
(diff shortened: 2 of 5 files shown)


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -553,7 +555,16 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				osbErrors = append(osbErrors, fmt.Errorf("manifest validation error: %v", err))
 			}
 		}
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "osbuild build failed", osbErrors)
+
+		reason := "osbuild build failed"
+		if len(failedStages) > 0 {
```
bcl, line 560: It feels a bit awkward to be pluralizing things manually :) You could just use `in stage(s):` instead.

schuellerf: As this ends up in the frontend at the customers - I thought it nicer there...
although I really expected that there is a simple library for this (I'm not the golang guy :-/ )

mvo5: Fwiw, this is usually handled via something like https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html as plural forms are complex in different languages. Not sure if we support i18n in the frontend (yet) but even if we don't preparing for it and getting "free" plural handling might be worth it (or not, idk). The topic of what gettext (if we go this route) to use in go is a bit more complex as there is no clear winner (at least last I looked).

schuellerf: I'd also prefer a library (not sure which one, though) but I got the feeling we only support English for now.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -401,7 +401,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	osbuildVersion, err := osbuild.OSBuildVersion()
 	if err != nil {
-		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "Error getting osbuild binary version", err)
+		osbuildJobResult.JobError = clienterrors.WorkerClientError(clienterrors.ErrorBuildJob, "Error getting osbuild binary version", err.Error())
```
schuellerf, line 404: btw. these lines (this one and 519) are maybe a fix for some cases where the details have not been propagated...

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: Can there be actually more than one failed stage? I don't think so... :thinking:

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -534,6 +535,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				logWithId.Infof("  %s success", stageResult.Type)
 			} else {
 				logWithId.Infof("  %s failure:", stageResult.Type)
+				failedStages = append(failedStages, stageResult.Type)
```
thozza, line 538: I think that the information about the failed stage name should be instead taken from the `osbuildJobResult.OSBuildOutput.Error`, which will contain it if the failure is caused by a stage failure.

See https://github.com/osbuild/osbuild/blob/1b5b015fef9da025ee872b63f46e0dc69f0c108b/osbuild/formats/v2.py#L426-L439

BTW, this also means that all of this is already part of the `osbErrors` added as a detail to `JobError`.

schuellerf: Probably it still would make sense to add in the the reason (additionally) as it's shown more prominently in the UI  - but I'll check to use `osbuildJobResult.OSBuildOutput.Error`

cmd/osbuild-worker/jobimpl-osbuild.go
schuellerf: I wasn't sure if there will be support for running parallel stages soonish - so I'd rather support it here?

cmd/osbuild-worker/jobimpl-osbuild.go
schuellerf: @thozza at that level I get `osbuildJobResult.OSBuildOutput.Error` only as `json.RawMessage` not sure if I should _try_ to guess the structure of it - or where could I get the type of the structure for sure? (couldn't find something like the `ClientErrorCode` or similar in that context)

Screenshot_20240502_113019

thozza: I think that a custom unmarshaler would be needed, which would check the value of the `"type"` property and unmarshal to the appropriate struct. We never implemented proper osbuild errors unmarshaling, because it was not needed. Until now.

I agree with putting more information into the "reason" of the error :wink:

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: OK, in that case up to you. I was just looking at the current implementation, which can't return failures in more than one stage.

cmd/osbuild-worker/jobimpl-osbuild.go
schuellerf: @thozza ... and I thought that the type is not very strict so it would be more some kind of guessing and trying to cover all types of structs?

thozza: We control what osbuild returns and it currently returns exactly one type of a struct in case of the stage failure. IMHO, it makes sense to standardize this in osbuild if it is not nice and easy to interpret in osbuild-composer (or any other place that consumes osbuild's JSON output).

We do this type of unmarshaling already in multiple places, e.g.:
https://github.com/osbuild/osbuild-composer/blob/8dece19342d14476a17f2a0a5c9807fed7e24320/internal/weldr/upload.go#L133-L175
https://github.com/osbuild/osbuild-composer/blob/8dece19342d14476a17f2a0a5c9807fed7e24320/internal/target/target.go#L62-L134

schuellerf:
Note: related issues: #4137 and #4145

bcl (approved):
Looks good

mvo5 (commented):
Thank you for improving this! I wonder if we should take the opportunity (maybe in a followup if you prefer) to extract a helper here and write a test. The current Run() function is very long and does a lot of things so extracting a helper will not helper *that* much but it's a start. Wdyt about something like:

[0001-osbuild-worker-extract-makeJobErrorFromOsbuildOutput.patch.txt](https://github.com/user-attachments/files/15884380/0001-osbuild-worker-extract-makeJobErrorFromOsbuildOutput.patch.txt)

as a starting point, there is a lot of room for improvement here but this feels like we could build more on top (and also could be used as an example to e.g. extract the logging too etc.

But happy to do as a followup of course!

schuellerf:
Also fixes https://github.com/osbuild/images/issues/727

cmd/osbuild-worker/jobimpl-osbuild_test.go
```diff
@@ -0,0 +1,81 @@
...
+bad-stage, Details: []`,
+		},
+		{
+			inputData: &osbuild.Result{
+				Success: false,
+				Log: map[string]osbuild.PipelineResult{
+					"fake-os": []osbuild.StageResult{},
+				},
+			},
+			expected: `Code: 10, Reason: osbuild build failed, Details: []`,
```
mvo5, line 47: I hope errors like this do not happen for real, i.e. failure with no further details :) but testing it here makes sense of course.

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -77,23 +77,31 @@ func workerClientErrorFrom(err error) (*clienterrors.Error, error) {
...
+		reason := "rpmmd error in depsolve job"
+		details := "<nil>"
+		if err == nil {
+			err = fmt.Errorf("workerClientErrorFrom expected an error to be processed. Not nil")
+		} else {
+			details = err.Error()
+		}
 		// Error originates from internal/rpmmd, not from dnf-json
-		return clienterrors.WorkerClientError(clienterrors.ErrorRPMMDError, err.Error(), nil), err
+		return clienterrors.WorkerClientError(clienterrors.ErrorRPMMDError, reason, details), err
```
mvo5, line 104: Totally unreleated to this PR which is great - I'm not too happy with the interface of workerClientErrorFrom() and I realize the irony that I probably added it /o\ - my issue is that the convention in go usually is that when returning some result and an error then when an error is returned the result is usually a null value (nil, empty etc). here this is not the case and that is non-idiomatic. Nothing related to your changes, but I wanted to mention it and I think we should fix it (happy to do in a followup).

mvo5 (approved):
Let's merge it, I would like to get back to workerClientErrorFrom() but this PR improves things a lot so let's get it merged :)

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -351,6 +351,33 @@ func (impl *OSBuildJobImpl) getPulpClient(targetOptions *target.PulpOSTreeTarget
...
+			if !stageResult.Success {
+				failedStage = stageResult.Type
+				break
+			}
+		}
+	}
+
+	reason := "osbuild build failed"
+	if len(failedStage) > 0 {
+		reason += " in stage:\n" + failedStage
```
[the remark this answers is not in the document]
schuellerf: Hmm was just stuck with the version expecting multiple stages to fail - can be removed

PR #4114, opened 2024-04-29
weldr/api: indicate preloading progress in the log

schuellerf:
With this you can see if the preloading is already finished or not

### DIFF
--- a/internal/weldr/api.go
+++ b/internal/weldr/api.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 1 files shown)


bcl:
I'm not understanding what additional info this adds. It already prints when it is done, and when I'm looking at the logs I don't care how many there are, just whether or not the one I'm interested in has failed to load the metadata.

schuellerf:
For the service to be up and running and responsive they should _all_ be loaded, right?
The current logs didn't indicate if it's still busy preloading.
If this doesn't provide any value also for running the service to check if the startup is very slow or still not fully done, we can as well drop this

croissanne (commented):
I'm also not sure what this contributes. I don't think the preload is technically required, it just makes it slightly faster when depsolving as it would have fetched the metadata.

schuellerf:
Hmm - it's used with the getting started "service" mode but you are right it's not aws 🤔

PR #4116, opened 2024-04-30
templates/openshift/maintenance: service account & parameter fixes

croissanne:
Current runs fail because the ssl mode isn't set, and the service account is missing (in the workers-stage ns).


schuellerf (changes requested):

schuellerf (approved):
LGTM (although some description where this change comes from, would be good 😉)

schuellerf (approved):

PR #4124, opened 2024-05-02
gen-manifests: derrive seedArg from filename and env

mvo5:
To ensure that manifests get random(ish) and stable UUIDs we set the seed arg based on the filename and the environment (`OSBUILD_GEN_MANIFEST_RNG_SEED`).

This should fix the issue discovered in
https://github.com/osbuild/manifest-db/pull/124

that duplicated UUIDs for xfs/btrfs can trigger random(ish) and hard to diagnose errors.

[draft as it lacks testing, I don't know yet how to do that]

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -12,10 +12,12 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/osbuild/images/pkg/blueprint"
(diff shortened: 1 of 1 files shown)


achilleas-k:
I think we might want to get rid of this cmd and only keep the one in images.  We'd have retool some things in manifest-db maybe so let's discuss it tomorrow and do it all then.

achilleas-k:
Unless this is urgent

mvo5:
If we get rid of this command I will close this PR and open one that removes the command  - otherwise we could simply import the helper from https://github.com/osbuild/images/pull/674 here now

mvo5:
Closing in favor of https://github.com/osbuild/osbuild-composer/pull/4152

PR #4129, opened 2024-05-06
Use copr rpms for the community workers

croissanne:
---

Currently the appsre job for the community workers hangs a lot, as it waits for rpmrepo rpms which are often not there. Either due to cancelled jobs (after fast merges to main) or [skip-ci] commits.

This is a short-term solution until we have a merge queue in composer (which should guarantee rpms for each PR merge).


tools/appsre-build-worker-packer.sh
```diff
@@ -67,7 +68,7 @@ EOF
         # write osbuild_commit variable if defined in Schutzfile
         # if it's not defined, osbuild will be installed from distribution repositories
         if [[ $osbuild_commit != "null" ]]; then
-            tee -a "$item/group_vars/all.yml" null >dev <<EOF
+            tee -a "$item/group_vars/all.yml" <<EOF
```
schuellerf, line 71: Then you could as well just `cat` it - or should it be visible in the logs?
```suggestion
            cat - >> "$item/group_vars/all.yml" <<EOF
```

templates/packer/ansible/roles/common/tasks/packages.yml
```diff
@@ -104,11 +105,28 @@
...
+    dnf copr enable -y @osbuild/osbuild-composer
+    COMPOSER_COMMIT_SHORT=$(echo {{ COMPOSER_COMMIT }} | head -c 9)
+    COMPOSER_RPMS=$(dnf search -q --showduplicates osbuild-composer | grep -E "osbuild-composer-worker-[0-9]+.*g$COMPOSER_COMMIT_SHORT.*{{ ansible_architecture }}" | cut -f 1 -d ':')
+    if [ -z $COMPOSER_RPMS ]; then
+      echo $COMPOSER_COMMIT_SHORT worker rpms not available for on @osbuild/osbuild-composer copr
+      exit 1
+    fi
+    echo "installing rpms $COMPOSER_RPMS"
+    dnf install -y $COMPOSER_RPMS
+
```
schuellerf, line 129: I guess it's not worth rewriting this more "ansible like" (using the builtin vars and dnf modules)

croissanne: I mean search + grep + cut, it's a hack, idk how you'd rewrite this in the ansible dnf module.

schuellerf: Just using the native dnf modules and set-fact (instead of bash-env) as much as possible but probably no real benefit here...

schuellerf (approved):
can't really test but LGTM

PR #4134, opened 2024-05-07
tools/fedora-worker-packer: increase timeout further

croissanne:
It can take a long time before packer picks up the jobs and the jobs are actually finished.


schuellerf (approved):
LGTM

PR #4135, opened 2024-05-07
Add support for arbitrary kickstart file injection into ISOs (HMS-3879)

achilleas-k:
Pulling in (primarily):
- https://github.com/osbuild/images/pull/631
- https://github.com/osbuild/images/pull/700
- https://github.com/osbuild/images/pull/705


achilleas-k:
Rebased on #4157

thozza (dismissed):
LGTM

achilleas-k:
Seems the issue is actually
```
  - nothing provides selinux-policy >= 38.33-1.fc38 needed by osbuild-selinux-119-1.20190819git4dbf8e2.fc38.noarch
  - nothing provides selinux-policy-base >= 38.33-1.fc38 needed by osbuild-selinux-119-1.20190819git4dbf8e2.fc38.noarch
```

thozza:
> Seems the issue is actually
> 
> ```
>   - nothing provides selinux-policy >= 38.33-1.fc38 needed by osbuild-selinux-119-1.20190819git4dbf8e2.fc38.noarch
>   - nothing provides selinux-policy-base >= 38.33-1.fc38 needed by osbuild-selinux-119-1.20190819git4dbf8e2.fc38.noarch
> ```

The osbuild repo (and CI building RPMs) probably uses newer rpmrepo snapshots with newer selinux-policy package.

achilleas-k:
Pulled in #4147

thozza (approved):

ondrejbudai (approved):
:heart_eyes:

achilleas-k:
Also found the issue that I noticed here: https://github.com/osbuild/osbuild-composer/pull/4147#issuecomment-2125045271

thozza:
@achilleas-k https://github.com/osbuild/images/pull/705 breaks the "weldr api distro alias" test case, because it relies on the existence of `/etc/dnf/vars/releasever` in the 8.10 SAP image. You may try to change the release it uses to 8.8 or some other EUS release which still creates the file.

achilleas-k:
> @achilleas-k [osbuild/images#705](https://github.com/osbuild/images/pull/705) breaks the "weldr api distro alias" test case, because it relies on the existence of `/etc/dnf/vars/releasever` in the 8.10 SAP image. You may try to change the release it uses to 8.8 or some other EUS release which still creates the file.

Would it make sense to change the alias test to read /etc/os-release?  That way we could build any image type, not rely on SAP, and read the `VERSION_ID`.

thozza:
> > @achilleas-k [osbuild/images#705](https://github.com/osbuild/images/pull/705) breaks the "weldr api distro alias" test case, because it relies on the existence of `/etc/dnf/vars/releasever` in the 8.10 SAP image. You may try to change the release it uses to 8.8 or some other EUS release which still creates the file.
> 
> Would it make sense to change the alias test to read /etc/os-release? That way we could build any image type, not rely on SAP, and read the `VERSION_ID`.

IMO no, because that file comes from the RPM and it will not really tell you which distro version composer used when generating the manifest. The DNF vars file in SAP images is really the only thing that gets baked into the image and reflects the distro version used by composer.

achilleas-k:
Right, we want to make sure the internal distro object and code path is correct regardless of repos.  Gotcha.

achilleas-k:
Added a config to the test that overrides the RHEL-8 alias.

Small question for follow-up, but I wonder if it's necessary to be building the images here.  I realise end-to-end testing is good, but maybe for this particular feature it's enough to just generate a manifest and check that we have a dnf config stage with the releasever set appropriately and save ourselves some testing time.

thozza:
> Small question for follow-up, but I wonder if it's necessary to be building the images here. I realise end-to-end testing is good, but maybe for this particular feature it's enough to just generate a manifest and check that we have a dnf config stage with the releasever set appropriately and save ourselves some testing time.

Distro aliases is a feature of composer, so you can't really test it in the `images` repo just by generating manifest. I'm not sure how would we reasonably generate just the manifest when testing a behavior in composer :thinking: If we could do that, then yeah, it would be OK to just inspect the manifest :wink:

achilleas-k:
Right.  I wasn't thinking of moving it to images, but maybe we could do something like cancel the build and inspect the manifest.

Anyway, a discussion for another time.

thozza (dismissed):
:+1:

achilleas-k:
Pushed again.  Had to fix the RHEL 8.8 runner name for the alias test (`ga` instead of `nightly`).

thozza (approved):
:crossed_fingers:

PR #4137, opened 2024-05-08
clientdetails: handle error interfaces in `Details` somewhat

mvo5:
The clienterror report has no tests right now, that is not ideal so this PR adds some.

While doing that it fixes an issue that @schuellerf discovered, i.e. that when an `error` interface is passed into `clienterrors.Error.Details` the details get lost because the json.Marshaler will not know how to handler an
error interface.

So I think this is fine to merge and will improve the current state. However I think we should rework the
clienterrors.Error so that go types will help us detect issues at compile time. With this PR we will only
detect nested errorrs at runtime and error and it will not deal with nested interface of other types.

We should probably:
1. make `clienterrors.Error` a real error type, it's kinda unusual that an error does not implement the `error` interface.
2. Rethink `Details` - AFAICT all the details will be "flattended" (via `%v`) after the de-serialization anyway, so we might just make details a "string". Of course testing this requires a bit of digging into the source and maybe writting some tests that validate that we indeed just use do the flattening on the other side.

### DIFF
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -3,6 +3,7 @@ package clienterrors
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 )
 
 const (
@@ -59,13 +60,57 @@ func (e *Error) String() string {
 	return fmt.Sprintf("Code: %d, Reason: %s, Details: %v", e.ID, e.Reason, e.Details)
 }
 
+func ensureNoErrs(v reflect.Value) error {
+	switch v.Kind() {
+	case reflect.Interface:
+		if errIf, ok := v.Interface().(error); ok {
+			if errIf != nil {
+				return fmt.Errorf("%v", errIf.Error())
+			}
+		}
+	case reflect.Ptr:
+		if err := ensureNoErrs(v.Elem()); err != nil {
+			return err
+		}
+	case reflect.Struct:
+		for i := 0; i < v.NumField(); i++ {
+			if err := ensureNoErrs(v.Field(i)); err != nil {
+				return err
+			}
+		}
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < v.Len(); i++ {
+			if err := ensureNoErrs(v.Index(i)); err != nil {
+				return err
+			}
+		}
+	case reflect.Map:
+		for _, key := range v.MapKeys() {
+			if err := ensureNoErrs(v.MapIndex(key)); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func (e *Error) MarshalJSON() ([]byte, error) {
 	var details interface{}
 	switch v := e.Details.(type) {
 	case error:
 		details = v.Error()
+	case []error:
+		details = make([]string, 0, len(v))
+		for _, err := range v {
+			if err != nil {
+				details = append(details.([]string), err.Error())
+			}
+		}
 	default:
-		details = v
+		if err := ensureNoErrs(reflect.ValueOf(v)); err != nil {
+			return nil, fmt.Errorf("found nested error in %+v: %v", e.Details, err)
+		}
+		details = e.Details
 	}
 
 	return json.Marshal(&struct {
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -1,6 +1,7 @@
 package clienterrors
 
 import (
+	"encoding/json"
 	"fmt"
 )
 
@@ -56,6 +57,26 @@ type Error struct {
 
 func (e *Error) String() string {
 	return fmt.Sprintf("Code: %d, Reason: %s, Details: %v", e.ID, e.Reason, e.Details)
+}
+
+func (e *Error) MarshalJSON() ([]byte, error) {
+	var details interface{}
+	switch v := e.Details.(type) {
+	case error:
+		details = v.Error()
+	default:
+		details = v
+	}
+
+	return json.Marshal(&struct {
+		ID      ClientErrorCode `json:"id"`
+		Reason  string          `json:"reason"`
+		Details interface{}     `json:"details,omitempty"`
+	}{
+		ID:      e.ID,
+		Reason:  e.Reason,
+		Details: details,
+	})
 }
 
 const (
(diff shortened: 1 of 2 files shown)


internal/worker/clienterrors/errors.go
```diff
@@ -49,15 +50,35 @@ const (
 type ClientErrorCode int
 
 type Error struct {
-	ID      ClientErrorCode `json:"id"`
-	Reason  string          `json:"reason"`
-	Details interface{}     `json:"details,omitempty"`
+	ID      ClientErrorCode
+	Reason  string
+	Details interface{}
```
schuellerf, line 55: why did you remove the `json:..` here? isn't that needed for decoding?

mvo5: Probably because I'm a Muppet (sometimes at least :) - if we also decode I need to re-add it, it is a bit annoying that it's duplicated but I am not sure there is a nicer way with go, I will look though.

internal/worker/clienterrors/errors_test.go
```diff
@@ -0,0 +1,38 @@
...
+	} {
+		wce := clienterrors.WorkerClientError(2, "details", tc.err)
+		assert.Equal(t, fmt.Sprintf("Code: 2, Reason: details, Details: %s", tc.expectedStr), wce.String())
+	}
+}
+
+func TestErrorJSONMarshal(t *testing.T) {
+	err := fmt.Errorf("some-error")
+
+	json, err := json.Marshal(clienterrors.WorkerClientError(2, "details", err))
```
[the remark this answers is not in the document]
mvo5: Thanks! That is a really good point. If we are certain about the inputs tht naive way of course would be something like
```go
diff --git a/internal/worker/clienterrors/errors.go b/internal/worker/clienterrors/errors.go
index 8a2587749..fac804302 100644
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -64,6 +64,8 @@ func (e *Error) MarshalJSON() ([]byte, error) {
        switch v := e.Details.(type) {
        case error:
                details = v.Error()
+       case []error:
+               details = fmt.Sprintf("%v", v)
        default:
                details = v
        }
diff --git a/internal/worker/clienterrors/errors_test.go b/internal/worker/clienterrors/errors_test.go
index 2e7eb9401..4ac7b19f7 100644
--- a/internal/worker/clienterrors/errors_test.go
+++ b/internal/worker/clienterrors/errors_test.go
@@ -30,9 +30,15 @@ func TestErrorInterface(t *testing.T) {
 }
 
 func TestErrorJSONMarshal(t *testing.T) {
-       err := fmt.Errorf("some-error")
-
-       json, err := json.Marshal(clienterrors.WorkerClientError(2, "details", err))
-       assert.NoError(t, err)
-       assert.Equal(t, `{"id":2,"reason":"details","details":"some-error"}`, string(json))
+       for _, tc := range []struct {
+               err         interface{}
+               expectedStr string
+       }{
+               {fmt.Errorf("some-error"), "some-error"},
+               {[]error{fmt.Errorf("err1"), fmt.Errorf("err2")}, "[err1 err2]"},
+       } {
+               json, err := json.Marshal(clienterrors.WorkerClientError(2, "details", tc.err))
+               assert.NoError(t, err)
+               assert.Equal(t, fmt.Sprintf(`{"id":2,"reason":"details","details":"%s"}`, tc.expectedStr), string(json))
+       }
 }
```
but of course if it's arbitrary nested it's a bit more work. I think it's also worth questing if "Details" should be this arbitrary, might be worthwhile to check how it's used.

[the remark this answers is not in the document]
mvo5: I pushed a bit of a compromise, needs more tests as it does not test for all the `ensureNoErrs` cases and also bails on the first nested error (but prints the whole structure so it should be okay).

internal/worker/clienterrors/errors.go
```diff
@@ -58,6 +60,65 @@ func (e *Error) String() string {
...
+func (e *Error) MarshalJSON() ([]byte, error) {
+	var details interface{}
+	switch v := e.Details.(type) {
+	case error:
+		details = v.Error()
+	case []error:
+		details = fmt.Sprintf("%v", v)
+	default:
+		if err := ensureNoErrs(reflect.ValueOf(v)); err != nil {
+			return nil, fmt.Errorf("found nested error in %+v: %v", e.Details, err)
```
mvo5, line 106: This is probably a bit too terse, maybe `found nested error in Details field` or something

schuellerf: For the rare cases (none?) where this will happen, I think it's good :-)

schuellerf (approved):
looks pretty complete

croissanne (approved):
thank you 😍

internal/worker/clienterrors/errors.go
```diff
@@ -58,6 +60,70 @@ func (e *Error) String() string {
...
+			}
+		}
+	case reflect.Ptr:
+		if err := ensureNoErrs(v.Elem()); err != nil {
+			return err
+		}
+	case reflect.Struct:
+		for i := 0; i < v.NumField(); i++ {
+			if err := ensureNoErrs(v.Field(i)); err != nil {
+				return err
```
schuellerf, line 78: @mvo5 Do I get this right, that only the first will be returned (same for array and map)? Is this on purpose?

mvo5: Yes, only the first is detected right now. We can change that of course, I'm not sure if it's worth it (also I'm not sure this entire PR is the right approach, it's nice in the sense that it does detect the issue now but it still will not catch other cases where we pass something to `Details interface{}` that is in fact an interface that is not properly serializable to json. 

I still wonder if we a better appraoch would be:
a) merge https://github.com/osbuild/osbuild-composer/pull/4145 to fix the immedidate issue
b) change `Details inteface{}` -> `Details string` or `Details []string` or `Details json.RawMessage` (the later would force the caller to serialize).

schuellerf: Not sure where the idea comes from that Details should be flexible but then it would be a) and b) with `json.RawMessage`
Starting with #4145 for now sound perfect 😏

mvo5: it's probably worth to salvage (some of) the test cases from this PR but otherwise I'm fine closing it (or closing it after (a) and (b) got done).

PR #4143, opened 2024-05-13
clienterrors: rename `WorkerClientError` to `clienterrors.New`

mvo5:
The usual convention to create new object is to prefix `New*` so this commit renames the `WorkerClientError`. Initially I thought it would be `NewWorkerClientError()` but looking at the package prefix it seems unneeded, i.e. `clienterrors.New()` already provides enough context it seems and it's the only error we construct.

We could consider renaming it to `clienterror` (singular) too but that could be a followup.

I would also like to make `clienterror.Error` implement the `error` interface but that should be a followup to make this (mechanical) rename trivial to review.

Note that the rename was done via
```
$ go fmt -w 'WorkerClientError -> NewWorkerClientError' ./...
```
so choosing a different name it "cheap".

### DIFF
--- a/cmd/osbuild-worker/jobimpl-awsec2.go
+++ b/cmd/osbuild-worker/jobimpl-awsec2.go
@@ -36,31 +36,31 @@ func (impl *AWSEC2CopyJobImpl) Run(job worker.Job) error {
 	var args worker.AWSEC2CopyJob
 	err := job.Args(&args)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


croissanne (approved):
yes please!

croissanne:
@mvo5 do you mind if I push to your branch with the rebase?

mvo5:
> @mvo5 do you mind if I push to your branch with the rebase?

Sorry for the slow reply, I rebased and pushed now, sorry that I did not see the conflicts earlier. And thanks for the kind offer to do it for me, that is of course always welcome too :)

PR #4145, opened 2024-05-13
osbuild-worker: do not use `error` in clienterror.Error.Details

mvo5:
This is an alternative/complementary fix for https://github.com/osbuild/osbuild-composer/pull/4137. It is very simple so should be uncontroversial.

It fixes an issue that @schuellerf discovered, i.e. that when an error interface is passed into clienterrors.Error.Details the details get lost because the json.Marshaler will not know how to handler an error interface.

To find the problematic uses of `error` a custom vet checker was build in https://github.com/mvo5/osbuild-cvet. With that the result is:
```
$ go run github.com/mvo5/osbuild-cvet@latest ./...
/home/mvogt/devel/osbuild/osbuild-composer/cmd/osbuild-worker/jobimpl-depsolve.go:93:26: do not pass 'error' to WorkerClientError() details, use error.Error() instead
/home/mvogt/devel/osbuild/osbuild-composer/cmd/osbuild-worker/jobimpl-osbuild.go:404:31: do not pass 'error' to WorkerClientError() details, use error.Error() instead
/home/mvogt/devel/osbuild/osbuild-composer/cmd/osbuild-worker/jobimpl-osbuild.go:519:31: do not pass 'error' to WorkerClientError() details, use error.Error() instead
/home/mvogt/devel/osbuild/osbuild-composer/cmd/osbuild-worker/jobimpl-osbuild.go:556:31: do not pass '[]error' to WorkerClientError() details, use []string instead
```
and once this commit is in no more errors.

Just like PR#4137 this is not perfect because it will not do a recursive check for the passed argument.

If we are happy with this approach we should add something like
```yaml
diff --git a/.github/workflows/tests.yml b/.github/workflows/tests.yml
index 635b1af0b..88f3d831c 100644
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,6 +146,9 @@ jobs:
           version: v1.54.2
           args: --verbose --timeout 5m0s
 
+      - name: Run osbuild-cvet
+        run: go run github.com/mvo5/osbuild-cvet@latest ./...
+
   packit-config-lint:
     name: "📦 Packit config lint"
     runs-on: ubuntu-latest
```
(and move to the vet to the osbuild org)

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -90,7 +90,7 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 					for _, baseurlstr := range repo.BaseURLs {
 						match, err := impl.RepositoryMTLSConfig.CompareBaseURL(baseurlstr)
 						if err != nil {
-							result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidRepositoryURL, "Repository URL is malformed", err)
+							result.JobError = clienterrors.WorkerClientError(clienterrors.ErrorInvalidRepositoryURL, "Repository URL is malformed", err.Error())
 							return err
 						}
 						if match {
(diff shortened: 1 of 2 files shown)


schuellerf (approved):
LGTM

PR #4152, opened 2024-05-16
[RFC] cmd: remove `gen-manifests` in favor of the `images` version

mvo5:
This is a result of the outcome of
https://github.com/osbuild/osbuild-composer/pull/4124

Achilleas suggested that instead of having a copy of `gen-manifests`
in osbuild-composer we could use the one from images. AFAICT nothing
in the code references `gen-manifests` except the `test/README` so
it seems safe to remove and just use images.

The README got updated to point to `images` and we will also
need to remove `test/data/manifest` at a later point (after
checking that nothing else depends on this anymore).


mvo5:
> I'd still like to remove the command here. Maybe we can update the docs to instruct people to clone `osbuild/images` and run the command from there, because the configs are also needed.
> 
> We should also delete `test/data/manifests/` while we're at it, but I think some tests look for them so maybe we should look at that separately.

Thanks! I reworked the README changes based on your feedback and reopened. I also added Alexander as the last two commits in `test/data/manifests/` were done by him (just to be sure that everyone involved is notified :)

atodorov:
> We should also delete `test/data/manifests/` while we're at it, but I think some tests look for them so maybe we should look at that separately.

Let me dig around the various repositories and test suites a bit more to remind myself what was there. I remember we analyzed this at the beginning of the year, removed some test scenarios which were duplicate b/w osbuild-composer and images repositories and kept the rest (there is a ticket with more details, I'll find it and we can talk it through on Monday). 

My only concern ATM is that if we happen to need to update `test/data/manifests/`, especially b/c we'll be adding new major/minor versions to the test matrix very soon, and the tools in the 2 repos are incompatible how we'll be able to update the manifests. Sounds like a test blocking situation could happen very easily so we may have to hold off on merging this PR for some time. 

I'll get back to you with more details.

test/README.md
```diff
@@ -101,19 +101,9 @@ Each file contains a sample manifest for an image configuration and some
...
-
-Manifest generation can be restricted to only some distributions,
-architectures, or image types using command line flags.
-
-The command uses the configurations in
-`tools/test-case-generators/format-request-map.json` and repositories defined
-in `tools/test-case-generators/repos.json`.
+Those are legacy now and all manifest testing is happening in
+github.com/osbuild/images - there is a similar `test/data/manifest` directory
+in this repo.
```
atodorov, line 106: Please see #3967 and the corresponding COMPOSER-2127 and COMPOSER-2122. 

From what I am reading on the above 2 tickets + 1 PR there are still multiple image types which aren't covered by osbuild/images and it is unclear (as of now) when they will be fully covered. 

Based on this we've removed duplicate test scenarios and kept ones which were only covered in the osbuild-composer repository. Based on this and the fact that the format between the 2 repositories has changed I cannot approve this PR. 

Let's discuss it more before introducing any more changes.

atodorov (changes requested):

test/README.md
achilleas-k: Thanks for this.  I admit, I lost track of how the manifests in osbuild-composer were being used.

mvo5: Thank you! Based on this I will close this PR for now until the tests are really fully transferred into a single repository.

PR #4162, opened 2024-05-23
templates/packer: switch to fedora-40

croissanne:
Fedora 38 is EOL, and packit no longer builds rpms for it.


schuellerf (changes requested):
should we also remove this from `Schutzfile`, `tools/check-runners`  and 
`internal/cloudapi/v2/compose_test.go`?
Would this save some time during the checks?

also in `test/cases/minimal-raw.sh` it seems odd that `fedora-37` maps to `fedora37` and 38 to `fedora-unknown` 🤔

croissanne:
> should we also remove this from `Schutzfile`, `tools/check-runners` and `internal/cloudapi/v2/compose_test.go`? Would this save some time during the checks?
> 
> also in `test/cases/minimal-raw.sh` it seems odd that `fedora-37` maps to `fedora37` and `fedora-38` to `fedora-unknown` 🤔

yea, I just wanted to fix the packer job first >.> Also yea that's not great in the minimal raw test ._.

schuellerf (approved):

croissanne:
Ugh ok so ansible on fedora 39 is broken. I might try fedora 40.

thozza (approved):
Looks good.

PR #4174, opened 2024-05-30
templates/dashboards: revert active worker count

croissanne:
It's misleading since it counts the amount of workers that have
    registered to the current composer pods, it doesn't actually keep track
    of the active workers.
    
Remove it and keep the worker-api stats as a proxy for active workers.


schuellerf (approved):
LGTM

PR #4184, opened 2024-06-03
Switch to oaas / osbuild-worker-executor

croissanne:
This simplifies the way the worker-executor, there's only 3 http calls happening: wait until ready, post for a build, get for the results.


mvo5 (commented):
Just as a quick drive-by comment, it might be nice if the first commit with the copy of oaas also points to the original repo url in the description (that will then be archived). Just so that the git history is complete.

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,42 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	client := http.Client{
+		Timeout: time.Second * 10,
+	}
+
+	for {
+		resp, err := client.Get(fmt.Sprintf("%s:8001/api/v1/", host))
+		if err != nil {
+			logrus.Debugf("Waiting for secure instance continues: %v", err)
+		}
+		if resp.StatusCode != 200 {
```
mvo5, line 43: It might be slightly more easier to read when checking on `if resp.StatusCode == 200 { return true}`, this way the flow is slightly more linear, maybe something like?
```go
	for {
		resp, err := client.Get(fmt.Sprintf("%s:8001/api/v1/", host))
		if err != nil {
			logrus.Debugf("Waiting for secure instance continues: %v", err)
		}
                defer resp.Body.Close()
		if resp.StatusCode == 200 {
			return true
		}
		body, err := io.ReadAll(resp.Body)
		if err != nil {
			logrus.Warningf("Unable to read body waiting for secure instance: %v", err)
		}
		logrus.Debugf("Waiting for secure instance continues: %s", body)

		select {
		case <-ctx.Done():
			logrus.Error("Timeout waiting for secure instance to spin up")
			return false
		}
	}
```

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,42 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+func waitForSI(ctx context.Context, host string) bool {
+	client := http.Client{
+		Timeout: time.Second * 10,
+	}
+
+	for {
+		resp, err := client.Get(fmt.Sprintf("%s:8001/api/v1/", host))
+		if err != nil {
+			logrus.Debugf("Waiting for secure instance continues: %v", err)
+		}
```
mvo5, line 42: AFAICT we should `defer resp.Body.Close()` here

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,42 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+		} else {
+			return true
+		}
+
+		select {
+		case <-ctx.Done():
+			logrus.Error("Timeout waiting for secure instance to spin up")
+			return false
+		default:
+			continue
```
mvo5, line 59: Is this `default` actually needed? given that it's a `for {}` loop, that's the default, yes?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -53,52 +95,133 @@ func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory s
...
-		"--builder-host",
-		*si.Instance.PrivateIpAddress,
-		"--store",
-		store,
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*10)
+	defer cancel()
+	if !waitForSI(ctx, *si.Instance.PrivateIpAddress) {
+		return nil, err
 	}
+	controlFile, err := os.Open(filepath.Join("/var/tmp/osbuild-composer", "control.json"))
```
mvo5, line 98: It might be slightly shorter to use `os.WriteFile()` here, something like:
```go
	tmpdir := os.MkdirTemp("/var/tmp/osbuild-composer", "executor-")
	if err != nil {
		return nil, fmt.Errorf("cannot create tempdir for oaas run: %w", err)
	}
	defer os.RemoveAll(tmpdir)
	
	controlFilePath := filepath.Join(tmpdir, "control.json")
	if err = os.WriteFile(controlFilePath, []byte(fmt.Sprintf("{\"exports\": [\"%s\"]}", exports[0])), 0644); err != nil {
		return nil, fmt.Errorf("cannot write control file: %w", err)
	}
```

mvo5, line 98: I'm always a bit nervous about using hardcoded paths like this, if this code moves and two instances of it run at the same time this becomes racy and they will (potentially) override each others files. Maybe worth creating a tempdir here using `os.MkdirTemp()`?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -53,52 +95,133 @@ func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory s
...
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(controlFile.Name())
+	_, _ = controlFile.Write([]byte(fmt.Sprintf("{\"exports\": [\"%s\"]}", exports[0])))
```
mvo5: This is a super silly concern but if exports contain any special for json it would break the build or even allow injection of data. But I guess exports are not user controlled? Out of paranoia I would still include a validator here that errors if exports is not e.g. `[a-z0-9]+` or something similar (probably too strict but if we can start with an allow list that would be ideal).

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -53,52 +95,133 @@ func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory s
...
+	_ = manifestFile.Sync()
+
+	cmd := exec.Command("tar",
+		"--strip-components=3",
+		"-cf",
+		filepath.Join("/var/tmp/osbuild-composer", "input.tar"),
+		filepath.Join("/var/tmp/osbuild-composer", "control.json"),
+		filepath.Join("/var/tmp/osbuild-composer", "manifest.json"),
+	)
+	if err = cmd.Run(); err != nil {
```
mvo5, line 134: Maybe worth using `cmd.Output()` or `cmd.CombinedOutput()` here and returning it when an error happens? Something like:
```go
	if output, err := cmd.CombinedOutput(); err != nil {
		logrus.Errorf("Unable to create input tar: %v output:\n%s", err, output)
		return nil, err
	}
```
there is often useful context in the output like "disk-full" and the standard behavior of exec.Command.Run() is to just include the error code which is often not very helpful

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -53,52 +95,133 @@ func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory s
...
+	if err != nil {
+		logrus.Errorf("Unable to write executor result tarball :%v", err)
+		return nil, err
+	}
+	cmd = exec.Command("tar",
+		"-Sxf",
+		"output.tar",
+	)
+	cmd.Dir = outputDirectory
+	err = cmd.Start()
```
mvo5, line 224: Could this be slightly shorter by just using `cmd.Run()` (or `cmd.CombinedOutput()`) ? Something like:
```go
	cmd.Dir = outputDirectory
	if output, err := cmd.CombinedOutput(); err != nil {
		logrus.Errorf("Untarring image failed: %v output:\n%s", err, output)
		return nil, err
	}
```

mvo5 (commented):
Thanks for working on this! I'm very excited! I did a quick runthrough and added some small sugestions/ideas, mostly drive-by. Super happy to help out here too, I would love for example to write some tests for runner-impl-aws-ec2.go :)

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: I think it's necessary for the loop to keep calling the executor, otherwise it'll just wait on the context being done after 1 call.

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +31,173 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
 	return err
 }
 
+func waitForSI(ctx context.Context, host string) bool {
```
mvo5, line 34: We have some similar code in https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker-executor/main_test.go#L19 - I wonde if its worthwhile to extract/reuse/unify them?

croissanne: Hmm, like something in internal/common?

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -84,11 +84,12 @@ func runOsbuild(buildDir string, control *controlJSON, output io.Writer) (string
 	// #nosec G204
 	cmd = exec.Command(
 		"tar",
+		"-C",
```
mvo5, line 87: This change will conflicts with https://github.com/osbuild/osbuild-composer/pull/4193 afaict (and that PR probably needs adjusting)  - (sorry!) it's not really important, the fix is trivial

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -188,8 +189,8 @@ func handleIncludedSources(atar *tar.Reader, buildDir string) error {
 		if filepath.Clean(name) != strings.TrimSuffix(name, "/") {
 			return fmt.Errorf("name not clean: %v != %v", filepath.Clean(name), name)
 		}
-		if !strings.HasPrefix(name, "store/") {
-			return fmt.Errorf("expected store/ prefix, got %v", hdr.Name)
+		if !strings.HasPrefix(hdr.Name, "osbuild-store/") {
```
mvo5, line 192: This should probably be `hdr.Name` -> `name` for consistency with the lines around it.

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -84,11 +84,12 @@ func runOsbuild(buildDir string, control *controlJSON, output io.Writer) (string
 	// #nosec G204
 	cmd = exec.Command(
 		"tar",
+		"-C",
+		buildDir,
```
mvo5, line 88: I'm curious why we serve the entire build dir, honest question there is nothing wrong with it (as such), but I wonder what the rational is. Does it make some internal things easier on the consumer side ?  

The thing that makes me slightly uncomfortable is that in some way anything that comes out of osbuild-worker-executor is potentially tainted/toxic so the less we touch it/process it the better. But again, not a blocker, mostly curiosity.

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -84,11 +84,12 @@ func runOsbuild(buildDir string, control *controlJSON, output io.Writer) (string
 	// #nosec G204
```
mvo5, line 84: Maybe a small comment here like: `// the result is put into a tar because we get sparse file support for free this way` (and/or if there is more maybe also list that?).

cmd/osbuild-worker-executor/handler_result.go
```diff
@@ -38,7 +38,7 @@ func handleResult(logger *logrus.Logger, config *Config) http.Handler {
 				return
 			}
 
-			fss := http.FileServer(http.Dir(filepath.Join(config.BuildDirBase, "build/output")))
```
mvo5, line 41: Same comment as above, maybe a very short coment why we do not serve output only but everything

cmd/osbuild-worker-executor/main.go
```diff
@@ -76,7 +76,7 @@ func run(ctx context.Context, args []string, getenv func(string) string) error {
 
 func main() {
 	ctx := context.Background()
-	if err := run(ctx, os.Args, os.Getenv); err != nil {
+	if err := run(ctx, os.Args[1:], os.Getenv); err != nil {
```
mvo5, line 79: *cough* nice catch!

internal/osbuildexecutor/runner-aws-ec2_test.go
```diff
@@ -0,0 +1,135 @@
...
+	require.False(t, waitForSI(ctx, server.URL))
+
+	server.Start()
+	ctx2, cancel2 := context.WithTimeout(context.Background(), time.Second*1)
+	defer cancel2()
+	require.True(t, waitForSI(ctx2, server.URL))
+}
+
+func TestWriteInputArchive(t *testing.T) {
+	cacheDir, err := os.MkdirTemp("", "")
```
mvo5, line 36: (nitpick) you could use `t.TempDir()` here which will not require error handling or cleanup :) (and in some similar places)

internal/osbuildexecutor/runner-aws-ec2_test.go
```diff
@@ -0,0 +1,135 @@
...
+	defer cancel()
+	require.False(t, waitForSI(ctx, server.URL))
+
+	server.Start()
+	ctx2, cancel2 := context.WithTimeout(context.Background(), time.Second*1)
+	defer cancel2()
+	require.True(t, waitForSI(ctx2, server.URL))
+}
+
+func TestWriteInputArchive(t *testing.T) {
```
mvo5, line 35: :heart:

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+		archive,
+		filepath.Base(control),
+		filepath.Base(manifest),
+	)
+	if output, err := cmd.CombinedOutput(); err != nil {
+		logrus.Errorf("Unable to create input tar: %v\n%s", err, output)
+		return "", err
+	}
+	/* #nosec G204 */
+	cmd = exec.Command("tar",
```
mvo5, line 101: Maybe a small comment here like: `// this is a separate tar call because here we need to switch into the "store" directory`

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	resp, err := client.Post(fmt.Sprintf("%s/api/v1/build", host), "application/x-tar", inputFile)
+	if err != nil {
+		logrus.Errorf("Unable to request build from executor instance: %v", err)
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 201 {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			logrus.Warningf("Unable to read body waiting for build to run: %v", err)
```
mvo5, line 135: Maybe worth including the status code in the error message here ?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+		}
+		return nil, fmt.Errorf("Something went wrong during executor build: %v\n%s", err, body)
+	}
+
+	var osbuildResult osbuild.Result
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("Something went wrong during executor build: %v\n%s", err, body)
+	}
+	err = json.Unmarshal(body, &osbuildResult)
```
mvo5, line 146: (super nitpick) we could use a `json.Decoder` which would avoid the io.ReadAll() as it would stream directly from the body into the json decoding (but fine to leave as is, this is a super nitpick). Unless I miss something of course, maybe there is a good reason to first get it all and then unmarshal :)

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
 
+func waitForSI(ctx context.Context, host string) bool {
+	client := http.Client{
+		Timeout: time.Second * 1,
+	}
+
+	for {
+		resp, err := client.Get(fmt.Sprintf("%s/api/v1/", host))
+		if err != nil {
+			logrus.Debugf("Waiting for secure instance continues: %v", err)
```
mvo5, line 41: (super nitpick) we probably want to get into the habbit of writing `%w` for errors so that they can be unwrapped potentially (here and in similar places but that is probably it's own followup so that we are consistent :)

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	if err != nil {
+		return nil, fmt.Errorf("Unable to unmarshal response body into osbuild result: %v\nbody: %s", err, body)
+	}
+
+	return &osbuildResult, nil
+}
+
+func fetchOutputArchive(cacheDir, host string) (string, error) {
+	client := http.Client{
+		Timeout: time.Minute * 15,
```
mvo5, line 156: (earnest questions) will this be enough for multi gigabyte archives? what values are we currently using?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+
+	resp, err := client.Get(fmt.Sprintf("%s/api/v1/result/output.tar", host))
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return "", err
```
mvo5, line 167: Maybe worthwhile to include more context here, in some cases this may just return `EOF` which will not be super helpful :)

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+		}
+		return "", fmt.Errorf("%v, body: %s", err, body)
```
mvo5, line 169: Maybe also a bit more context here like: `cannot fetch output archive: %w, body: %s", err, body` or something similar?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	}
+	file, err := os.Create(filepath.Join(cacheDir, "output.tar"))
+	if err != nil {
+		logrus.Errorf("Unable to write executor result tarball :%v", err)
+		return "", err
```
mvo5, line 174: Maybe also a bit more context here?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	if resp.StatusCode != 200 {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return "", err
+		}
+		return "", fmt.Errorf("%v, body: %s", err, body)
+	}
+	file, err := os.Create(filepath.Join(cacheDir, "output.tar"))
+	if err != nil {
+		logrus.Errorf("Unable to write executor result tarball :%v", err)
```
mvo5, line 173: I notice we logrus here but not on the other error places, should we logrus in more places or will it be logged on the call side already?

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,6 +30,174 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+	defer file.Close()
+	_, err = io.Copy(file, resp.Body)
+	if err != nil {
+		logrus.Errorf("Unable to write executor result tarball :%v", err)
+		return "", err
+	}
+	return file.Name(), nil
+}
+
+func extractOutputArchive(outputDirectory, outputTar string) error {
```
mvo5, line 185: Do we need to worry about about directory traversal attacks here? Given that things coming from the osbuild-worker-executor can be potentially tainted/toxic should we do a pre-check that there are no ".." or other unclean bits in the tar? Or is this already implicitly ensured? (*maybe* we could even extract as a different user but that would also be a followup).

mvo5 (commented):
Thank you! This looks very nice! Some ideas/comments inline, the only relevant question (that prevented me from setting "approve") is if we need to be checking for directory traversal attachs in the resulting tar file download.

cmd/osbuild-worker-executor/handler_build.go
croissanne: oh right, rebase fallout

cmd/osbuild-worker-executor/handler_build.go
croissanne: the idea was to serve the build log through this as well, but in the end we just pull the output.

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: currently it's 30 minutes, i'll bump it

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: looking at it I think we should log in runosbuild, but not lower than that. Will keep them in WaitForSI, but otherwise just move them up.

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -53,61 +227,49 @@ func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory s
...
-		logrus.Errorf("Starting osbuild-jobsite-manager failed: %v", err)
+		logrus.Errorf("Something went wrong handling the executor's build: %v", err)
 		return nil, err
 	}
-	err = cmd.Wait()
+	if !osbuildResult.Success {
+		return osbuildResult, nil
+	}
+
+	outputArchive, err := fetchOutputArchive(ec2e.tmpDir, executorHost)
```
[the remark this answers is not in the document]
croissanne: hmm, but we already define exports in control.json no? I never really understood the difference between exportPaths and exports here.

cmd/osbuild-worker-executor/handler_build.go
croissanne: oh right, because otherwise we're creating a tarball from the output directory, outputted to output/output.tar, and tar isn't always happy with creating an archive inside the directory you're archiving. So that's why it's one up. Either we server the entire build dir or we move the tarball after.

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: hmm, like first call "tar -tf" and see what's in it?

mvo5: Yeah, ideally we would be fine with a totally hostile remote side and my worry would be that they send us a tar full of ../../../../../../usr/bin/sshd with `#!/bin/sh -e p0wned`

internal/osbuildexecutor/runner-impl-aws-ec2.go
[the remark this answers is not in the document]
mvo5: Maybe I phrased my question bad (sorry!) I wonder if we could (in theory, not saying we should) fetch only exactly what we asked the remote site to build for us? I.e. do we have all the information for that here?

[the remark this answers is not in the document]
croissanne: hmm but we're just pulling the output tarball right? And whta's in there is determined by the exports were demanded.

internal/osbuildexecutor/export_test.go
```diff
@@ -0,0 +1,8 @@
+package osbuildexecutor
+
+var ExtractOutputArchive = extractOutputArchive
```
mvo5, line 3: (comment, no action needed) this could also be written as:
```go
var (
   ExtractOutputArchive = extractOutputArchive
   FetchOutputArchive = fetchOutputArchive
   ...
)
```

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -24,27 +34,231 @@ func prepareSources(manifest []byte, store string, extraEnv []string, result boo
...
+
+}
+
 func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, store, outputDirectory string, exports, exportPaths, checkpoints,
 	extraEnv []string, result bool, errorWriter io.Writer) (*osbuild.Result, error) {
 
 	err := prepareSources(manifest, store, extraEnv, result, errorWriter)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Failed to prepare sources: %w", err)
```
mvo5, line 246: :heart: for this and all the other contexts added (and %w!)

mvo5 (approved):
Thank you! This looks really nice!

mvo5 (approved):
Still looks good (after the rebase)

internal/osbuildexecutor/runner-impl-aws-ec2_test.go
```diff
@@ -0,0 +1,194 @@
...
+		"store/subdir/",
+		"store/subdir/contents",
+		"store/contents",
+		"",
+	}, strings.Split(string(out), "\n"))
+}
+
+func TestHandleBuild(t *testing.T) {
+	buildServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusCreated)
```
mvo5, line 71: Maybe for a followup - it might be nice to read the input there and doulbe check that we got the expected inputArchive, this might be as simple as `inp := io.ReadAll(r.Body); c.AssertEqual(t, []byte("test"), inp)` or something.

mvo5 (approved):
Thank you!

internal/osbuildexecutor/runner-impl-aws-ec2_test.go
croissanne: https://github.com/osbuild/osbuild-composer/pull/4204

PR #4188, opened 2024-06-05
obuild-worker: extract workerClientErrorFrom() helper and add tests

mvo5:
Tiny commit to extract a helper from DepsolveJobImpl.Run() that can then be unit tested.

This should help with https://github.com/osbuild/images/issues/727


achilleas-k (approved):

schuellerf (approved):
Good starting point - looking forward to the subsequent PR removing the duplication 😅😇
I'll align with @mvo5 so we don't do things at the same time 😉

PR #4189, opened 2024-06-05
cmd/osbuild-worker-executor: add oaas as worker executor

mvo5:
This PR is a split out of https://github.com/osbuild/osbuild-composer/pull/4184

It adds the new `osbuild-worker-executor` without actually using it (yet). This way we can retire the old https://github.com/mvo5/oaas repo and work in the osbuild-composer repo together


croissanne (approved):
lgtm!

PR #4191, opened 2024-06-06
osbuild-worker-executor: make test output silent again

mvo5:
Do not use the global logger but pass instead the locally created logger. This means the test output is silent again.

Sadly using the global logger is difficult because it is a global resource so replacing it in tests means all tests (that are potentially run in parallel) will write to it which makes testing specific log output hard.


croissanne (approved):
Woops, sorry, force of habit :)

PR #4192, opened 2024-06-06
osbuild-worker-executor: fix order of `assert.Equal()` in tests

mvo5:
The `assert.Equal()` expects that the "expected" value is put first. Which is not what I'm used to. It's also slightly inconsistent because `assert.EqualError()` expects the "actual" err first and then the expected string. But this commit is not about ranting :)

This commit fixes the order in the tests assert.Equal() so that mismatches actually are displayed correctly.


bcl (approved):
Looks good to me, this one trips me up as well.

PR #4193, opened 2024-06-06
osbuild-worker-executor: fix tar warning and log unexpected  output

mvo5:
This commit fixes a warning from tar that the archive cannot contain
itself. It also makes any tar output a warning (maybe even an error?)
as we do not expect anything from the tar command. The test is updated
to also check this.

To review just look at https://github.com/osbuild/osbuild-composer/commit/b85ea62feb72d4d2ac2abf5ead4f79560d85b4ab

Build on top of https://github.com/osbuild/osbuild-composer/pull/4191 as it requires the logger to be accessible from the test :)


cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -84,6 +84,7 @@ func runOsbuild(logger *logrus.Logger, buildDir string, control *controlJSON, ou
 	// #nosec G204
 	cmd = exec.Command(
 		"tar",
+		"--exclude=output/output.tar",
```
thozza, line 87: Nitpick: it would be IMO better to not hardcode the output directory and join `outputDir` with `"output.tar"` as it is done below.

thozza (commented):
I've added one comment, the rest looks good :+1:

mvo5:
This got merged as part of https://github.com/osbuild/osbuild-composer/pull/4184/commits

PR #4196, opened 2024-06-06
main: rework the way the mock logger is passed

mvo5:
Pass the mock logger directly to `run()` instead of mocking `logrus.New`. Doing the later leads to a data race when multiple parallel tests modify the (global) `var logrusNew logrus.New`.

Thanks to Tomas Hozza for reporting.


thozza (approved):
:+1:

PR #4205, opened 2024-06-12
osbuildexecutor: allow file type `tar.TypeGNUSparse` too

mvo5:
We need to allow files of type `tar.TypeGNUSparse` in the result that we get from the osbuild-worker-executor too.


schuellerf (approved):
LGTM

PR #4208, opened 2024-06-13
osbuild-worker-executor: `job-id` in control.json as hostname

mvo5:
This commit adds support to set the hostname to the job-id that is part of the control.json.

This is the first part, the second part will actually set this in the `internal/osbuildexecutor/runner-impl-aws-ec2.go`


croissanne (approved):
lgtm! ty :penguin:

PR #4214, opened 2024-06-18
golangci: fix integration tests for images

schuellerf:
`images` also calls this linter but with `path-prefix`.
This should fix PR tests in the `images` repo


mvo5 (approved):
Thank you

PR #4217, opened 2024-06-19
tools/check-snapshots: implement retries

schuellerf:
as there seem to be some problems in the GH-runners, I think retrying might be a valid solution.
At least for HTTP-504 which I got e.g. in #4103


croissanne (changes requested):
Thanks! I think we should just disable it and fix the root cause.

bcl:
> Thanks! I think we should just disable it and fix the root cause.

Yeah, I agree :/ The core problem is that the endpoint takes too long to respond. The previous fix for that was increasing the memory of the runner IIRC, but it might be better to clean up old snapshots? According to my notes when I wrote the checker I was getting responses in about 8s! So things have slowed way down.

PR #4218, opened 2024-06-19
osbuild-worker-executor/main_test: use random port for tests

schuellerf:
I got `error listening and serving: listen tcp 127.0.0.1:18002: bind: address already in use` during copr/packit a build.

(e.g. [this build](https://download.copr.fedorainfracloud.org/results/packit/osbuild-osbuild-composer-4103/centos-stream-10-s390x/07628035-osbuild-composer/builder-live.log.gz) )

This implementation for sure is racy but hopefully better than colliding with other concurrent tests with a fixed port for sure

### DIFF
--- a/cmd/osbuild-worker-executor/main_test.go
+++ b/cmd/osbuild-worker-executor/main_test.go
@@ -3,8 +3,10 @@ package main_test
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


schuellerf:
@mvo5 any other suggestion to avoid `address already in use`? 🤔

mvo5 (approved):
Thank you! Very nice!

PR #4220, opened 2024-06-20
internal/awscloud: retry CreateFleet (HMS-4098)

croissanne:
When receiving the "UnfillableCapacity" error from CreateFleet, retry the request with an OnDemand instance.

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -104,7 +104,7 @@ func (a *AWS) RunSecureInstance(iamProfile, keyName, CloudWatchGroup string) (*S
 		return nil, fmt.Errorf("Expected at least 1 subnet in the VPC, got 0")
 	}
 
-	createFleetOutput, err := a.ec2.CreateFleet(&ec2.CreateFleetInput{
+	createFleetOutput, err := a.createFleet(&ec2.CreateFleetInput{
 		LaunchTemplateConfigs: []*ec2.FleetLaunchTemplateConfigRequest{
 			&ec2.FleetLaunchTemplateConfigRequest{
 				LaunchTemplateSpecification: &ec2.FleetLaunchTemplateSpecificationRequest{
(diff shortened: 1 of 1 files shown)


internal/cloud/awscloud/secure-instance.go
```diff
@@ -436,3 +422,35 @@ func (a *AWS) deleteSGIfExists(si *SecureInstance) error {
...
+		return nil, fmt.Errorf("Unable to create fleet: %w", err)
+	}
+
+	if len(createFleetOutput.Errors) > 0 && createFleetOutput.Errors[0].ErrorCode == aws.String("UnfillableCapacity") {
+		logrus.Warn("Received UnfillableCapacity from CreateFleet, retrying CreateFleet with OnDemand instance")
+		input.SpotOptions = nil
+		createFleetOutput, err = a.ec2.CreateFleet(input)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Unable to create fleet: %w", err)
```
schuellerf, line 438: Can we change the error string, so we know which one failed?
```suggestion
		return nil, fmt.Errorf("Unable to create OnDemand fleet: %w", err)
```

internal/cloud/awscloud/secure-instance.go
```diff
@@ -436,3 +422,35 @@ func (a *AWS) deleteSGIfExists(si *SecureInstance) error {
 	}
 	return err
 }
+
+func (a *AWS) createFleet(input *ec2.CreateFleetInput) (*ec2.CreateFleetOutput, error) {
+	createFleetOutput, err := a.ec2.CreateFleet(input)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to create fleet: %w", err)
```
schuellerf, line 429: same here
```suggestion
		return nil, fmt.Errorf("Unable to create spot fleet: %w", err)
```

schuellerf (changes requested):

schuellerf (approved):
LGTM

PR #4221, opened 2024-06-20
Set executor hostname via cloud-init

croissanne:
This way much more of the journal will be captured under the new hostname.

### DIFF
--- a/internal/cloud/awscloud/secure-instance_test.go
+++ b/internal/cloud/awscloud/secure-instance_test.go
@@ -8,6 +8,7 @@ import (
 func TestSecureInstanceUserData(t *testing.T) {
 	type testCase struct {
 		CloudWatchGroup  string
+		Hostname         string
 		ExpectedUserData string
 	}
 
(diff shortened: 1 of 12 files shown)


templates/packer/ansible/roles/common/files/worker-initialization-scripts/set_executor_hostname.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -euo pipefail
```
mvo5, line 2: (slightly unrelated) Fwiw, `pipefail` has it's own set of weirdness , see https://mvogt.wordpress.com/2018/07/23/bash-o-pipefail-can-have-pitfailswpitfalls/ - not that it matters here.

internal/cloud/awscloud/secure-instance_test.go
```diff
@@ -8,6 +8,7 @@ import (
 func TestSecureInstanceUserData(t *testing.T) {
 	type testCase struct {
 		CloudWatchGroup  string
+		Hostname         string
```
mvo5, line 11: :heart:

mvo5 (approved):
Thank you!

templates/packer/ansible/roles/common/files/worker-initialization-scripts/set_executor_hostname.sh
croissanne: oh interesting, had no idea

PR #4225, opened 2024-06-25
internal/weldr/api_test: fix unit tests for images v0.68.0

schuellerf:
Unittests within `images` like this
https://github.com/osbuild/images/actions/runs/9648467656/job/26609722237?pr=760
fail as the API changed and `osbuild-composer` needs to use the updated images to have the same environment

should we also update to go 1.21 with a vendor update like this?


schuellerf:
… ok we'll stick to go 1.20 for now :-D

bcl (approved):
Looks Good

thozza:
This PR is most probably not relevant any more, since the `osbuild/images` version used by composer is now `v0.70.0`. Feel free to reopen if I'm mistaken.

PR #4226, opened 2024-06-25
awscloud: try to terminate previous secure instance (HMS-4341)

croissanne:
In case the previous executor SI belonging to the worker did not get shut down properly, attempt to do it again when starting a new one, otherwise replacing the SG or LT will not work.

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -78,6 +78,15 @@ func (a *AWS) RunSecureInstance(iamProfile, keyName, cloudWatchGroup, hostname s
 		}
 	}()
 
+	previousSI, err := a.terminatePreviousSI(identity.InstanceID)
+	if err != nil {
+		logrus.Errorf("Unable to terminate previous secure instance %s (parent instance ID: %s): %v", previousSI, identity.InstanceID, err)
+		return nil, fmt.Errorf("Unable to terminate previous secure instance %s (parent instance ID: %s): %v", previousSI, identity.InstanceID, err)
+	}
+	if previousSI != "" {
+		logrus.Warningf("Previous instance (%s) terminated by parent instance (%s)", previousSI, identity.InstanceID)
+	}
+
 	sgID, err := a.createOrReplaceSG(identity.InstanceID, identity.PrivateIP, vpcID)
 	if sgID != "" {
 		secureInstance.SGID = sgID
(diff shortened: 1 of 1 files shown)


internal/cloud/awscloud/secure-instance.go
```diff
@@ -78,6 +78,11 @@ func (a *AWS) RunSecureInstance(iamProfile, keyName, cloudWatchGroup, hostname s
 		}
 	}()
 
+	previousSI, err := a.terminatePreviousSI(identity.InstanceID)
+	if err != nil {
+		logrus.Warningf("Unable to terminate previous secure instance with instance ID (%s): %v, this job is now likely to fail", previousSI, err)
```
schuellerf, line 83: logging `previousSI` in this case is not really useful in all cases? 
or maybe ... (see below)

internal/cloud/awscloud/secure-instance.go
```diff
@@ -194,6 +199,42 @@ func (a *AWS) TerminateSecureInstance(si *SecureInstance) error {
...
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(descrInstancesOutput.Reservations) == 0 || len(descrInstancesOutput.Reservations[0].Instances) == 0 {
+		return "", nil
+	}
+
+	if *descrInstancesOutput.Reservations[0].Instances[0].State.Name == ec2.InstanceStateNameTerminated {
+		return "", nil
```
schuellerf, line 219: … maybe change this for better error messages
```suggestion
		return "instance not yet terminated", nil
```

internal/cloud/awscloud/secure-instance.go
```diff
@@ -194,6 +199,42 @@ func (a *AWS) TerminateSecureInstance(si *SecureInstance) error {
...
+				Name:   aws.String("tag:parent"),
+				Values: []*string{aws.String(hostInstanceID)},
+			},
+		},
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(descrInstancesOutput.Reservations) == 0 || len(descrInstancesOutput.Reservations[0].Instances) == 0 {
+		return "", nil
```
schuellerf, line 215: … maybe change this for better error messages
```suggestion
		return "instance has no reservations", nil
```

internal/cloud/awscloud/secure-instance.go
```diff
@@ -194,6 +199,42 @@ func (a *AWS) TerminateSecureInstance(si *SecureInstance) error {
...
+	descrInstancesOutput, err := a.ec2.DescribeInstances(&ec2.DescribeInstancesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name:   aws.String("tag:parent"),
+				Values: []*string{aws.String(hostInstanceID)},
+			},
+		},
+	})
+	if err != nil {
+		return "", err
```
schuellerf, line 212: … maybe change this for better error messages
```suggestion
		return "no instance found", err
```

schuellerf (changes requested):

internal/cloud/awscloud/secure-instance.go
```diff
@@ -78,6 +78,11 @@ func (a *AWS) RunSecureInstance(iamProfile, keyName, cloudWatchGroup, hostname s
 		}
 	}()
 
+	previousSI, err := a.terminatePreviousSI(identity.InstanceID)
+	if err != nil {
+		logrus.Warningf("Unable to terminate previous secure instance with instance ID (%s): %v, this job is now likely to fail", previousSI, err)
+	}
```
schuellerf, line 84: would it be also interesting as a metric to print the terminated `previousSI` in the `else` case too?
```suggestion
	} else {
	  logrus.Infof("Previous secure instance with instance ID (%s) terminated", previousSI)
	}
```

internal/cloud/awscloud/secure-instance.go
croissanne: ah right, I think we should replace it with the host/worker id maybe

internal/cloud/awscloud/secure-instance.go
croissanne: sure

internal/cloud/awscloud/secure-instance.go
croissanne: Well in this case the describe instances call failed, so it's not that there were no instances found.

internal/cloud/awscloud/secure-instance.go
croissanne: n this case the instance is already terminated, which is good

internal/cloud/awscloud/secure-instance.go
croissanne: in this case there are no active executor instances, which is fine. I don't think it should be logged.

schuellerf (approved):
LGTM

PR #4228, opened 2024-06-25
Introduce make lint

schuellerf:
trying to reproduce `golangci-lint` errors with the new `make lint` and fix them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,11 @@ SRCDIR ?= .
 
 RST2MAN ?= rst2man
 
+# v1.55 to get golang 1.21 (1.21.3)
+# v1.53 to get golang 1.20 (1.20.5)
+GOLANGCI_LINT_VERSION=v1.53
+GOLANGCI_LINT_CACHE_DIR=$(HOME)/.cache/golangci-lint/$(GOLANGCI_LINT_VERSION)
+GOLANGCI_COMPOSER_IMAGE=composer_golangci
 #
 # Automatic Variables
 #
@@ -82,6 +87,7 @@ help:
 	@echo "    unit-tests:         Run unit tests"
 	@echo "    push-check:         Replicates the github workflow checks as close as possible"
 	@echo "                        (do this before pushing!)"
+	@echo "    lint:               Runs linters as close as github workflow as possible"
 
 $(BUILDDIR)/:
 	mkdir -p "$@"
@@ -89,6 +95,8 @@ $(BUILDDIR)/:
 $(BUILDDIR)/%/:
 	mkdir -p "$@"
 
+$(GOLANGCI_LINT_CACHE_DIR):
+	mkdir -p "$@"
 #
 # Documentation
 #
@@ -156,9 +164,10 @@ install: build
 clean:
 	rm -rf $(BUILDDIR)/bin/
 	rm -rf $(CURDIR)/rpmbuild
+	rm -rf container_composer_golangci_built.info
 
 .PHONY: push-check
-push-check: build unit-tests srpm man
+push-check: lint build unit-tests srpm man
 	./tools/check-runners
 	./tools/check-snapshots --errors-only .
 	rpmlint --config rpmlint.config $(CURDIR)/rpmbuild/SRPMS/*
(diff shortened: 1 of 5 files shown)


tools/install_golangci_lint_deps.sh
achilleas-k, line 1: These aren't `golangci-lint` dependencies, they're osbuild-composer (and obsuild-worker) deps.  I'd actually prefer to emphasise the fact that it's a script that uses `apt` and ubuntu package names and name it something like `apt-install-deps.sh`, so people don't try to use it to install dependencies in other scenarios.

tools/install_golangci_lint_deps.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# just a warning - as "sudo" is not in the container we are using
+echo "This script should be run as root"
+
+apt-get update
+
+# This is needed to lint internal/upload/koji package
+apt-get install -y libkrb5-dev
```
mvo5, line 9: (quick drive-by comment) doing this in one long command line (like `apt install -y libkrb5-dev libgpgme-dev libbtrfs-dev libdevmapper-dev` is slightly more efficient as it will read/rebuild the apt cache only once (but of course writing comments is a little bit harden then.

Makefile
```diff
@@ -275,3 +284,11 @@ scratch: $(RPM_SPECFILE) $(RPM_TARBALL)
...
 		$(RPM_SPECFILE)
 
+container_composer_golangci_built.info: Makefile Containerfile_golangci_lint tools/install_golangci_lint_deps.sh
+	podman build -f Containerfile_golangci_lint -t $(GOLANGCI_COMPOSER_IMAGE) --build-arg "GOLANGCI_LINT_VERSION=$(GOLANGCI_LINT_VERSION)"
+	echo "Image last built on" > $@
+	date >> $@
+
+.PHONY: lint
+lint: $(GOLANGCI_LINT_CACHE_DIR) container_composer_golangci_built.info
+	podman run -t --rm -v $(SRCDIR):/app:z -v $(GOLANGCI_LINT_CACHE_DIR):/root/.cache:z -w /app $(GOLANGCI_COMPOSER_IMAGE) golangci-lint run -v
```
achilleas-k, line 294: I'm not entirely sure what benefit this has over just installing (or downloading) the golangci-lint version we care about and running it without a container.  I'm not against it, but I personally will never run this instead of `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53 run`.

tools/install_golangci_lint_deps.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
mvo5, line 1: All the other tools use the explicit `#!/bin/bash` here it seems, unless there is a good reason I would keep it this way. I personally prefer the `#!/bin/sh`  (or `#!/bin/bash`) because it ensures it's always the distro provided bash [0] that is used, not some random one installed in $PATH. But then I'm coming from a distro background so take it with a grain of salt :)

[0] assuming no "make install prefix=/" was used

mvo5 (commented):
Some quick dry-by comments (that may actually be a bit silly, freel free to ignore)

Makefile
schuellerf: The environment would be more similar to the GH workflow, which is the one rejecting your PR later on 🤔

tools/install_golangci_lint_deps.sh
schuellerf: yes, I just took it over (was like this) due to the comments - I also didn't really like calling apt that often

tools/install_golangci_lint_deps.sh
schuellerf: I thought `/usr/bin/env` is the most generic way but the main reason was, that my IDE suggested and autocompleted that line 🙊

schuellerf: I'd rather have a small chat in the team about the way to go and I'd be happy to align all then.

schuellerf: you are right about the `sh` though - `bash` is not needed in this script ;)

Makefile
achilleas-k: I guess that's fair.

achilleas-k:
Fedora 38 RPM builds are failing but F38 is EOL so let's just drop them.

schuellerf:
how to re-run the gitlab pipelines? I think the PR should be fine

achilleas-k (approved):

PR #4239, opened 2024-07-02
Added arch and storage_account, needed by CIV to deploy the resources…

narmaku:
… without having to check for vhd uri regex.

The regex check is now not needed since we create other shared image metadata on the fly. Plus it was not a good solution for an upstream project, as the regex was only affecting our RH teams internally.

This pull request includes:


narmaku:
See related CIV PR here: https://github.com/osbuild/cloud-image-val/pull/327

atodorov (approved):

PR #4242, opened 2024-07-02
internal/prometheus: more human-readable time buckets

croissanne:
Let's make them slightly easier to query and reason about. From 100 ms to 24 hours.

---

idk why I thought weird powers of 2 and sums of different powers of 2 was a good idea.


kingsleyzissou (approved):
=)

PR #4246, opened 2024-07-04
templates/dashboards: human readable job duration targets

croissanne:
Also makes the default 40m, which is the new slo target for osbuild jobs.

---

Now that #4242 is in, let's update the dashboards.


schuellerf (approved):
LGTM

PR #4247, opened 2024-07-04
Schutzfile: move to latest osbuild with `curl --parallel` support

mvo5:
This commit updates osbuild so that we benefit from osbuild PR#1573 to enable `curl --parallel`. This should speed up downloading especially on fast connections as less time is spend with the connection setup/handshake etc.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -2,7 +2,7 @@
   "fedora-38": {
     "dependencies": {
       "osbuild": {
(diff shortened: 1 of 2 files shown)


ondrejbudai (dismissed):
Go!

mvo5:
Setting to draft as there are some CI failures that need investigation first.

mvo5:
This needs more investigation, the "gcp" test is not working and it looks like there is a bug in the way the config file is generated.

PR #4254, opened 2024-07-09
osbuild-worker: rework the `workerClientErrorFrom()` error

mvo5:
The workerClientErrorFrom() was returning an `*clienterrors.Error` and an `error` (if something with the conversation goes wrong.

But the calling code was expecting that even if an `error` is returned the `*clienterrors.Error` is still valid. The caller would then just log the error. As returning a valid `value` even when there is an `error` is an unexpected pattern this commit changes the code to always return a `*clienterrors.Error` and log any issue via the logger.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -73,7 +73,11 @@ func (impl *DepsolveJobImpl) depsolve(packageSets map[string][]rpmmd.PackageSet,
 	return depsolvedSets, repoConfigs, nil
 }
 
-func workerClientErrorFrom(err error) (*clienterrors.Error, error) {
+func workerClientErrorFrom(err error, logWithId *logrus.Entry) *clienterrors.Error {
+	if err == nil {
+		logWithId.Errorf("workerClientErrorFrom expected an error to be processed. Not nil")
+	}
+
 	switch e := err.(type) {
 	case dnfjson.Error:
 		// Error originates from dnf-json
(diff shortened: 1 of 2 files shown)


cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -73,35 +73,39 @@ func (impl *DepsolveJobImpl) depsolve(packageSets map[string][]rpmmd.PackageSet,
...
-			err := fmt.Errorf("Unhandled dnf-json error in depsolve job: %v", err)
+			logWithId.Errorf("Unhandled dnf-json error in depsolve job: %v", err)
 			// This still has the kind/reason format but a kind that's returned
 			// by dnf-json and not explicitly handled here.
-			return clienterrors.WorkerClientError(clienterrors.ErrorDNFOtherError, reason, details), err
+			//
+			// XXX: it seems slightly dangerous to assume
+			// that any "error" we get there is coming
+			// from rpmmd, can we generate a more typed
+			// error from dnfjson here for rpmmd errors?
```
thozza, line 101: Not sure if I get this comment, specifically why it mentions `rpmmd`? I assume that the `case dnfjson.Error:` above ensures that the "error" comes from `dnfjson`... :thinking: 

Maybe you wanted to add it to the `defalt:` below, which assumes that the error comes from `rpmmd`?

thozza (commented):
I've added one comment, but the PR technically LGTM. :+1:

cmd/osbuild-worker/jobimpl-depsolve.go
mvo5: Thank you for pointing this out, indeed the comment was put at the wrong place and I meant `default` - I will fix it!

mvo5: Fwiw, I was curious and looked around a bit in this area, and it seems there are multiple ways  that `depsolve()` can fail that are not `dnfjson.Error` or errors from rpmmd, e.g. set proxy might fail, json error might happen, errors while setting up the communication with dnfjson. Nothing critical but in some case we probably give misleading errors. 

I wonder if something like:
```go
diff --git a/cmd/osbuild-worker/jobimpl-depsolve.go b/cmd/osbuild-worker/jobimpl-depsolve.go
index e79ec8983..983b2e561 100644
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -97,15 +97,9 @@ func workerClientErrorFrom(err error, logWithId *logrus.Entry) *clienterrors.Err
                        return clienterrors.WorkerClientError(clienterrors.ErrorDNFOtherError, reason, details)
                }
        default:
-               reason := "rpmmd error in depsolve job"
+               reason := "unknown error in depsolve job, see details"
                details := fmt.Sprintf("%v", err)
-               // Error originates from internal/rpmmd, not from dnf-json
-               //
-               // XXX: it seems slightly dangerous to assume that any
-               // "error" we get there is coming from rpmmd, can we
-               // generate a more typed error from dnfjson here for
-               // rpmmd errors?
-               return clienterrors.WorkerClientError(clienterrors.ErrorRPMMDError, reason, details)
+               return clienterrors.WorkerClientError(clienterrors.ErrorDepsolveUnknown, reason, details)
        }
 }
 
diff --git a/internal/worker/clienterrors/errors.go b/internal/worker/clienterrors/errors.go
index e575c5741..3b44b794d 100644
--- a/internal/worker/clienterrors/errors.go
+++ b/internal/worker/clienterrors/errors.go
@@ -44,6 +44,7 @@ const (
        ErrorJobPanicked          ClientErrorCode = 37
        ErrorGeneratingSignedURL  ClientErrorCode = 38
        ErrorInvalidRepositoryURL ClientErrorCode = 39
+       ErrorDepsolveUnknown      ClientErrorCode = 40
 )
 
 type ClientErrorCode int
```
is appropriate but it will break backwards compatibility (obviously)

thozza: AFAIK, nothing in the backend should depend on the use of `ErrorRPMMDError`. @regexowl  are you aware of anything that could break in the frontend?

Other than that, I like this version / suggestion. It is the cleanest IMO.

thozza (approved):
LGTM, thanks :+1:

cmd/osbuild-worker/jobimpl-depsolve.go
regexowl: Not sure if I completely understand the changes, but frontend doesn't use `ErrorRPMMDError` anywhere, unless API schema changes all should be good on that front. As for image-builder, quick search for `ErrorRPMMDError` didn't find anything :sweat_smile:

thozza: I think that the question is more about the error codes (number), than the actual name of the error in the implementation. Sorry for not being more specific :innocent: 

Now it is `23`. With the proposed change, it would be `40`. I think that the worry is that the error code `23` may be presented in a specific way to users and changing the code for the same type of errors could break that.

regexowl: I see, frontend should be fine then. As for image-builder, from a quick search it also looks mostly fine. There is one test that might be using the error code in two places: https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/v1/handler_post_compose_test.go#L359

thozza: Thanks. I think that the test case should be fine.

mvo5:
> LGTM, thanks 👍

Sorry, had to rebase because of conflicts, could you please re-review?

thozza (approved):
Still looks good.

I realized that we should eventually replace `dnf-json` in the error strings with `osbuild-depsolve-dnf`...

PR #4276, opened 2024-07-29
test/api: silent dump_db()

achilleas-k:
When cleaning up api tests, the build job results are dumped from the database to store in build-result.txt and are printed to the runner log at the same time.  This makes it very difficult to read results as the database dump prints very long lines that flood the job log.  The result isn't really readable (or useful) unless the file is downloaded separately.

Silence the `tee` command so that the file is created but no output is printed from the db dump.

See for example [this job log](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/7433608651) which starts with the last line of the database dump.  To see previous lines, you need to switch to the complete raw form.


jrusz (approved):
Makes total sense, it's saved into the `ARTIFACTS` folder so it will be available in S3 to look at.

thozza (approved):
Yes please! Thank you! :tada:

PR #4278, opened 2024-07-30
clienterrors: rename `WorkerClientError` to `clienterrors.New`

mvo5:
The usual convention to create new object is to prefix `New*` so this commit renames the `WorkerClientError`. Initially I thought it would be `NewWorkerClientError()` but looking at the package prefix it seems unneeded, i.e. `clienterrors.New()` already provides enough context it seems and it's the only error we construct.

We could consider renaming it to `clienterror` (singular) too but that could be a followup.

I would also like to make `clienterror.Error` implement the `error` interface but that should be a followup to make this (mechanical) rename trivial to review.

[this was https://github.com/osbuild/osbuild-composer/pull/4143 but for some reason I couldn't reopen it]


croissanne (approved):
lgtm! thank you :)

PR #4279, opened 2024-07-30
Change log_format for the service to json

schuellerf:
This is to be inline with `image-builder` and to
enable decoding in splunk

@croissanne I'm still somewhat in the dark with the deployment setup :-/ is this the correct place to change the logs, going to splunk, to be json? (to support `|spath input=message` which works for `image-builder`)


croissanne (approved):
lgtm! And I think so :)

PR #4285, opened 2024-08-01
logrus: add deployment channel as field to the logs

schuellerf:
As far as I saw in the openshift deployment, the parameter `"CHANNEL"` is already set implicitly by the environment (`…/environments/stage.yml` or `…/environments/production.yml`)
If something like this worked, this would be awesome for filtering logs

### DIFF
--- a/cmd/osbuild-composer/config.go
+++ b/cmd/osbuild-composer/config.go
@@ -12,17 +12,18 @@ import (
 )
 
 type ComposerConfigFile struct {
-	Koji          KojiAPIConfig     `toml:"koji"`
-	Worker        WorkerAPIConfig   `toml:"worker"`
-	WeldrAPI      WeldrAPIConfig    `toml:"weldr_api"`
-	DistroAliases map[string]string `toml:"distro_aliases" env:"DISTRO_ALIASES"`
-	LogLevel      string            `toml:"log_level"`
-	LogFormat     string            `toml:"log_format"`
-	DNFJson       string            `toml:"dnf-json"`
-	SplunkHost    string            `env:"SPLUNK_HEC_HOST"`
-	SplunkPort    string            `env:"SPLUNK_HEC_PORT"`
-	SplunkToken   string            `env:"SPLUNK_HEC_TOKEN"`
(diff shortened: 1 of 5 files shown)


schuellerf:
probably it would also work if we remove the default `value` ("local" now) from `templates/openshift/composer.yml` to really skip it if undefined?

schuellerf:
Hmm - if this works - I should probably move this to the `slogger` subdir so it can be reused in image-builder?

thozza (approved):
Thanks! The change looks good AFAICT and it could really make the logs easier to filter.

schuellerf:
@croissanne any concerns merging it?

cmd/osbuild-composer/config.go
```diff
@@ -12,17 +12,18 @@ import (
...
+	WeldrAPI          WeldrAPIConfig    `toml:"weldr_api"`
+	DistroAliases     map[string]string `toml:"distro_aliases" env:"DISTRO_ALIASES"`
+	LogLevel          string            `toml:"log_level"`
+	LogFormat         string            `toml:"log_format"`
+	DNFJson           string            `toml:"dnf-json"`
+	SplunkHost        string            `env:"SPLUNK_HEC_HOST"`
+	SplunkPort        string            `env:"SPLUNK_HEC_PORT"`
+	SplunkToken       string            `env:"SPLUNK_HEC_TOKEN"`
+	GlitchTipDSN      string            `env:"GLITCHTIP_DSN"`
+	DeploymentChannel string            `env:"CHANNEL"`
```
croissanne, line 26: Not sure how I feel about the name Channel vs something like Environment, but otherwise lgtm.

croissanne (approved):

cmd/osbuild-composer/config.go
schuellerf: I was also hesitant but it's supposedly called "channel" in openshift!? so I didn't want to rename to keep it readable
if i works I'll firstly move the code to `slogger` and can rename if you think this is better

croissanne (approved):

PR #4287, opened 2024-08-02
cloud/awscloud: aws sdk v2

croissanne:


go.mod
```diff
@@ -17,6 +17,11 @@ require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.13
 	github.com/BurntSushi/toml v1.4.0
 	github.com/aws/aws-sdk-go v1.55.2
+	github.com/aws/aws-sdk-go-v2 v1.26.0
```
mvo5: (super-nitpick) I wonder how hard it woud be to split the commit that just add dependencies to vendor/* and the one that actually chnage the code. I like to skip commits that are essentially um-reviewable (like adding the new aws sdk with hundrets of thousands of lines) and by splitting between imported code and changed code this would be slightly eaiser. But I understand the argument to combine them (plus it's a bit of extra work) so fine to ignore my comment :) 

P.S. What I would *actually* want is to not include depednecies like this in our git tree at all but that is a bigger argument :)

croissanne: No that's fair, I'll rebase

internal/boot/aws.go
```diff
@@ -190,85 +165,111 @@ func WithBootedImageInEC2(e *ec2.EC2, securityGroupName string, imageDesc *image
...
 		// We need to terminate the instance now and wait until the termination is done.
 		// Otherwise, it wouldn't be possible to delete the image.
-		_, err = e.TerminateInstances(&ec2.TerminateInstancesInput{
-			InstanceIds: []*string{
-				res.Instances[0].InstanceId,
+		_, err = awscl.EC2().TerminateInstances(
+			context.Background(),
+			&ec2.TerminateInstancesInput{
+				InstanceIds: []string{
+					*res.Instances[0].InstanceId,
```
mvo5, line 236: I assume we can be sure instanceId is set if err == nil? Or do we need to be careful here to avoid null-pointer dereferences?

internal/boot/aws.go
```diff
@@ -190,85 +165,111 @@ func WithBootedImageInEC2(e *ec2.EC2, securityGroupName string, imageDesc *image
...
-		err = e.WaitUntilInstanceTerminated(describeInstanceInput)
+		instTermWaiter := ec2.NewInstanceTerminatedWaiter(awscl.EC2())
+		err = instTermWaiter.Wait(
+			context.Background(),
+			describeInstanceInput,
+			time.Hour,
+		)
 		if err != nil {
-			retErr = wrapErrorf(retErr, "waiting for the instance termination failed: %v", err)
+			retErr = wrapErrorf(retErr, "cannot terminate the instance: %v", err)
```
mvo5, line 252: Maybe %w here too? (and maybe errors.Join() but that is not very important)

internal/boot/aws.go
```diff
@@ -190,85 +165,111 @@ func WithBootedImageInEC2(e *ec2.EC2, securityGroupName string, imageDesc *image
...
 	// actually can do something useful for us.
-	err = e.WaitUntilInstanceRunning(describeInstanceInput)
+	instWaiter := ec2.NewInstanceRunningWaiter(awscl.EC2())
+	err = instWaiter.Wait(
+		context.Background(),
+		describeInstanceInput,
+		time.Hour,
+	)
 	if err != nil {
 		return fmt.Errorf("waiting for the instance to be running failed: %v", err)
```
mvo5, line 268: Maybe `%w`  here?

internal/cloud/awscloud/autoscaling.go
```diff
@@ -0,0 +1,45 @@
...
+	descrASG, err := a.asg.DescribeAutoScalingInstances(
+		context.Background(),
+		&autoscaling.DescribeAutoScalingInstancesInput{
+			InstanceIds: []string{
+				identity.InstanceID,
+			},
+		},
+	)
+	if err != nil {
+		return nil
```
mvo5, line 26: Do we want to return nil here? Or should this be `return err` ? If `return nil` maybe a comment why?

internal/cloud/awscloud/awscloud.go
```diff
@@ -139,10 +182,15 @@ func NewForEndpoint(endpoint, region, accessKeyID, accessKey, sessionToken, caBu
 // "AWS_SHARED_CREDENTIALS_FILE" env variable or will default to
 // $HOME/.aws/credentials.
 func NewForEndpointFromFile(filename, endpoint, region, caBundle string, skipSSLVerification bool) (*AWS, error) {
-	return newAwsFromCredsWithEndpoint(credentials.NewSharedCredentials(filename, "default"), region, endpoint, caBundle, skipSSLVerification)
+	return newAwsFromCredsWithEndpoint(config.WithSharedCredentialsFiles([]string{filename, "default"}), region, endpoint, caBundle, skipSSLVerification)
 }
 
-func (a *AWS) Upload(filename, bucket, key string) (*s3manager.UploadOutput, error) {
+// This is used by the internal/boot test, which access the ec2 apis directly
+func (a *AWS) EC2() EC2 {
```
mvo5, line 189: If this is only used for tests, maybe `EC2ForTestsOnly` or something to make sure users know this is not a public api? In the past I did checks in cases like this that would look at os.Argv[0] and see if it ends with _test to ensure that helpers like this are not misused but that is probably overkill here :)

internal/cloud/awscloud/awscloud.go
```diff
@@ -264,72 +265,73 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
...
-	// we no longer need the object in s3, let's just delete it
-	logrus.Infof("[AWS] 🧹 Deleting image from S3: %s/%s", bucket, key)
-	_, err = a.s3.DeleteObject(&s3.DeleteObjectInput{
-		Bucket: aws.String(bucket),
-		Key:    aws.String(key),
-	})
-	if err != nil {
-		return nil, err
+	snapshotTaskStatus := *snapWaitOutput.ImportSnapshotTasks[0].SnapshotTaskDetail.Status
+	if snapshotTaskStatus != "completed" {
```
mvo5, line 285: (nitpick) does the aws sdk provide a constant for this so that we don't need to compare to an untyped string?

internal/cloud/awscloud/awscloud.go
```diff
@@ -385,61 +388,47 @@ func (a *AWS) CopyImage(name, ami, sourceRegion string) (string, error) {
...
+			Resources: []string{*result.ImageId},
+			Tags: []ec2types.Tag{
+				{
+					Key:   aws.String("Name"),
+					Value: aws.String(name),
+				},
 			},
 		},
-	})
+	)
```
mvo5, line 421: (super nitpick) I would remove the `\n` here to have the error handling directly next to the error

mvo5 (commented):
Thanks! This looks very interesting! I did a initial review up to internal/cloud/awscloud/awscloud_test.go and then I got tired (sorry!). Some initially comments/questions inline I will try to pick it up later again. Thanks for working on this, looks like it's going to be a really nice improvement.

mvo5 (approved):
Thanks, it's nice to see that this is now unit testable, great!

cmd/osbuild-worker/main.go
```diff
@@ -98,59 +94,24 @@ func WatchJob(ctx context.Context, job worker.Job) {
...
+		logrus.Debugf("Error getting the instance region: %v", err)
 		return
 	}
 
-	// get the identity for the instanceID
-	identity, err := ec2metadata.New(awsSession).GetInstanceIdentityDocument()
+	aws, err := awscloud.NewDefault(region)
 	if err != nil {
-		logrus.Debugf("Error getting the identity document, %s", err)
+		logrus.Infof("Unable to get default aws client: %v", err)
```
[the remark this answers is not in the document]
croissanne: Well we always try to protect the instance from scale-in, regardless of where the worker runs. And it's annoying if people get warning or error messages in their logs that they're not running in AWS as part of an ASG, most users won't.

Really we should turn the protection thing off when we're not running in AWS as part of an ASG but maybe we do that in a followup?

internal/boot/aws.go
```diff
@@ -190,85 +165,111 @@ func WithBootedImageInEC2(e *ec2.EC2, securityGroupName string, imageDesc *image
...
-		})
+		_, err = awscl.EC2().DeleteSecurityGroup(
+			context.Background(),
+			&ec2.DeleteSecurityGroupInput{
+				GroupId: securityGroup.GroupId,
+			},
+		)
 
 		if err != nil {
 			retErr = wrapErrorf(retErr, "cannot delete the security group: %v", err)
```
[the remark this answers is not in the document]
croissanne: let's do it later if ok, the internal/boot code is just used in testing.

internal/cloud/awscloud/autoscaling.go
croissanne: Looked at it again, and at this point we'd know we're running in AWS, but not that we're running as part of an ASG. So let's return the error and let the worker log an info message

```
...
		logrus.Infof("Unable to protect host, if the host isn't running as part of an autoscaling group, this can safely be ignored: %v", err)
```

cmd/osbuild-worker/main.go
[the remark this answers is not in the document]
mvo5: Totally fine for a followup, maybe a small comment for now (or ignore if the followup comes soon). I guess my wondering is exactly this, can we somehow detect when this is an actual issue and we need to raise it to the user or if we are just not part of an ASG and in this case either ignore or not even do the operation?

internal/boot/aws.go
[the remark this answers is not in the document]
mvo5: Totally fine of course :)

internal/cloud/awscloud/mocks_test.go
```diff
@@ -0,0 +1,404 @@
...
+				InstanceStatus: &ec2types.InstanceStatusSummary{
+					Status: ec2types.SummaryStatusOk,
+				},
+			},
+		},
+	}, nil
+}
+
+func (m *ec2mock) RunInstances(ctx context.Context, input *ec2.RunInstancesInput, optfns ...func(*ec2.Options)) (*ec2.RunInstancesOutput, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
croissanne: Rather "RunInstances" += 1; but we don't spin up instances using RunInstances, we spin them up using CreateFleet. RunInstances is only used during the boot tests.

internal/cloud/awscloud/mocks_test.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/stretchr/testify/require"
+)
+
+type s3mock struct {
```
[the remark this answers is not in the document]
croissanne: Hand coded ;_;

internal/boot/aws.go
croissanne: So generally yes, it's just a quirk from the way the SDK is generated I think. Also considering this bit is only used during tests I think it's fine.

cmd/osbuild-worker/main.go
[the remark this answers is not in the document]
croissanne: I added some small comments + returning the error if it failed to describe, lmk if that looks ok to you or if it needs more comments.

internal/cloud/awscloud/awscloud.go
mvo5: Thanks for checking, this is sightly sad indeed

internal/boot/aws.go
mvo5: Thanks! I see, that was not clear to me that this is only used in tests, maybe the surrounding function should also be called `TestOnlyWithBootedImageInEC2` ?

mvo5 (approved):

PR #4289, opened 2024-08-05
osbuildexecutor: show full osbuild exector on json decode errors

mvo5:
This is a short term workaround to get better visibility why the osbuild executor sometimes sends non-json data. When reading the result from the executor the entire output is now read and if the json parsing goes wrong it will use the entire body in the error message for better debug visibility.


croissanne (approved):
Thank you!

PR #4291, opened 2024-08-06
templates/packer: increase net.core.(w/r)mem_max

croissanne:
In conjunction with osbuild/osbuild#1833 this should fix building very large manifests.

---

- [x] https://github.com/osbuild/osbuild/pull/1833

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -2,7 +2,7 @@
   "fedora-38": {
     "dependencies": {
       "osbuild": {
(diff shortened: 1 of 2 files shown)


mvo5 (approved):

mvo5:
With https://github.com/osbuild/osbuild/pull/1836 merge this should no longer be needed because we decided to use FDs for large data transfers. If this this PR still wants to be used we would tweak the {r,w}mem_default instead.

PR #4299, opened 2024-08-12
osbuild-composer: activate deployment-channel reporting for splunk

schuellerf:
followup of PR #4285 

the "deployment channel" does not seem to be ending up in splunk.
@croissanne is it possible that it does not work just due to the order of logrus hooks being executed?


croissanne (approved):
Yea that's probably it :) Upstream it's a map of slices https://github.com/sirupsen/logrus/blob/dd1b4c2e81afc5c255f216a722b012ed26be57df/hooks.go#L14

PR #4301, opened 2024-08-13
splunk_logger: move environment hook to splunk_logger pt1

schuellerf:
this has to be in two steps:
- duplicate `environment_hook` to `splunk_logger` first
- change and use the reference from within `go.mod` and the main code

so this PR just duplicates the code to `pkg/splunk_logger` to be able to reference it from there after being on `main` (from composer as well as image-builder later on)

or can this be done in one go? (...in one "go" 😁 🙄 )

### DIFF
--- a/pkg/splunk_logger/environment_hook.go
+++ b/pkg/splunk_logger/environment_hook.go
@@ -1,0 +1,26 @@
+package logger
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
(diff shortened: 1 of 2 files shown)


schuellerf:
Thanks for the hints @croissanne (in chat), I'll remove the reference from `go.mod` in this case and try it in one PR

croissanne (approved):

PR #4308, opened 2024-08-15
osbuild-composer: add `Requires: osbuild-dnf-json-api = 7`

mvo5:
Use an exact version dependency on the `osbuild-dnf-json-api` to ensure incompatible json protocol changes cannot break composer.

See also https://github.com/osbuild/osbuild/pull/1849


thozza (approved):
LGTM

bcl (approved):
Looks like an excellent idea :)

PR #4317, opened 2024-08-21
tools/appsre-ansible: retry subscribing rpmbuild machines

croissanne:
Also unsubscribe after the build is done.

---

Packer builds are failing with
```
10:18:52 TASK [Subscribe] ***************************************************************
10:19:18 fatal: [44.203.59.246]: FAILED! => {"changed": false, "cmd": "/sbin/subscription-manager register --org **** --activationkey VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "msg": "Remote server error. Please check the connection details, or see /var/log/rhsm/rhsm.log for more information.", "rc": 70, "stderr": "Remote server error. Please check the connection details, or see /var/log/rhsm/rhsm.log for more information.\n", "stderr_lines": ["Remote server error. Please check the connection details, or see /var/log/rhsm/rhsm.log for more information."], "stdout": "", "stdout_lines": []}
```
sometimes. Let's just retry like we do in the other playbook.

### DIFF
--- a/tools/appsre-ansible/rpmbuild.yml
+++ b/tools/appsre-ansible/rpmbuild.yml
@@ -29,6 +29,9 @@
 
   # RHEL mock templates don't work on RHUI, they use the CDN repos
   - name: Subscribe
+    register: result
+    retries: 5
+    until: result is success
     community.general.redhat_subscription:
       activationkey: "{{ RH_ACTIVATION_KEY }}"
       org_id: "{{ RH_ORG_ID }}"
@@ -167,3 +170,10 @@
       mode: pull
       src: /home/ec2-user/rpmbuild/RPMS
       dest: /osbuild-composer/templates/packer/ansible/roles/common/files/rpmbuild/{{ ansible_architecture }}
+
+  - name: Unsubscribe
+    register: result
+    retries: 5
+    until: result is success
+    community.general.redhat_subscription:
+      state: absent


tools/appsre-ansible/rpmbuild.yml
```diff
@@ -29,6 +29,9 @@
 
   # RHEL mock templates don't work on RHUI, they use the CDN repos
   - name: Subscribe
+    register: result
+    retries: 5
+    until: result is success
```
schuellerf, line 34: I think that line is not needed but would probably work

schuellerf (approved):

tools/appsre-ansible/rpmbuild.yml
schuellerf: …and then also `register: …` is not needed - `retries: 5` should actually be sufficient if we don't need to have a special condition other than "not failing"

croissanne: I think you need the register / until for the retries to work? Not sure.

schuellerf: Seems to be "new" in v2.16
https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met 
> If until is not specified, the task will retry until the task succeeds but at most `retries` times (New in version 2.16).

PR #4319, opened 2024-08-22
test/cases: adapt upgrade to support rhel9to10

jrusz:
This pull request includes:


test/data/upgrade/upgrade_prepare.sh
```diff
@@ -0,0 +1,57 @@
...
+# Install SUT
+dnf install -y osbuild-composer composer-cli
+
+# Expecting either 8 or 9
+if [[ "$VERSION_ID" == "8" ]]; then
+  # Prepare the upgrade
+  curl -k -o /etc/yum.repos.d/oam-group-leapp-rhel-8.repo https://gitlab.cee.redhat.com/oamg/upgrades-dev/oamg-rhel8-vagrant/-/raw/main/roles/init/files/leapp-copr.repo
+  # install the leapp upgrade tool + other dependencies
+  dnf install -y leapp-upgrade-el8toel9 vdo jq rpmdevtools
+else
```
bcl, line 22: Maybe also test for 9 and error if it isn't either to future-proof this against being run on 10 before it's ready?

test/data/upgrade/upgrade_verify.sh
```diff
@@ -67,6 +69,29 @@ tee /etc/osbuild-composer/repositories/rhel-9.5.json > /dev/null << EOF
     ]
 }
 EOF
+else
```
bcl, line 72: Same thing here, check for 10. Also not sure about being specific about 9.5 in the above block, which will need to get bumped with every new release won't it?

jrusz: Yes, but we always get a ticket to test a specific upgrade path so I'd like to be explicit here. If I try to get the latest compose or rely on any other automation I'm not sure I could assure that we test what's expected.

jrusz:
I've tried to separate the rename into a separate commit but Github still didn't get it. All the important changes are in the first commit. @bcl @thozza could you give me a review please?

jrusz:
There is some bash trickery going on that's failing the nightly pipelines, need to investigate.

jrusz:
:facepalm: So apparently when you execute a script with an argument and then source another script within that script it gets that argument passed as well which broke my new logic in `define-compose-url.sh` ... I've expanded the check to also check that it's float, so should be fine now.

test/data/rhel-upgrade/upgrade_verify.sh
```diff
@@ -67,6 +69,32 @@ tee /etc/osbuild-composer/repositories/rhel-9.5.json > /dev/null << EOF
     ]
 }
 EOF
+elif [[ ${VERSION_ID%.*} == 10 ]]; then
+  tee /etc/osbuild-composer/repositories/rhel-10.0.json > /dev/null << EOF
+{
+    "x86_64": [
+        {
+            "name": "baseos",
+            "baseurl": "http://download.devel.redhat.com/rhel-10/nightly/RHEL-10-Public-Beta/latest-RHEL-10.0/compose/BaseOS/x86_64/os/",
```
thozza, line 78: Nitpick: there are already non-Beta RHEL-10.0 composes. Would it make sense to use those, instead of Beta? or do you want to explicitly test Beta?

thozza (approved):
Looks good from my PoV. I added one question, but it is not a blocker for merging, since it can be a follow-up.

test/data/rhel-upgrade/upgrade_verify.sh
jrusz: Hmmm, good question, thanks for the heads up. The ticket I'm working on RHELMISC-5648 specifies to test upgrade from 9.5 to 10.0-Beta but it might be obsolete at this time. I'll post a question on the ticket.

thozza: Sure, this is fine from my PoV.

PR #4320, opened 2024-08-22
cmd/osbuild-worker/main: fix worker "crashing" on connection issues during startup

schuellerf:
Rationale:
When the worker can't reach composer _during runtime_ it just retries.
When it can't reach composer on startup it "crashes" (`logrus.Fatalf()` terminates the worker)

With this patch the worker stays alive and retries.
This should be more reliable and faster than restarting completely.
Also for unit tests and development it's nicer to just retry.

(this is loosely connected to the tests of https://github.com/osbuild/ansible-osbuild-worker/pull/6 )

### DIFF
--- a/internal/worker/client.go
+++ b/internal/worker/client.go
@@ -121,9 +121,17 @@ func NewClient(conf ClientConfig) (*Client, error) {
 		clientId:     conf.ClientId,
 		clientSecret: conf.ClientSecret,
 	}
-	err = client.registerWorker()
+
+	// get a token first to avoid an "unauthorized" reply
+	// essentially saving one request on startup
+	err = client.refreshAccessToken()
 	if err != nil {
-		return client, err
+		logrus.Warnf("Error getting access token on startup, %v. Trying again later…", err)
+	} else {
+		err = client.registerWorker()
+		if err != nil {
+			logrus.Warnf("Error registering worker on startup, %v. Trying again later…", err)
+		}
 	}
 	go client.workerHeartbeat()
 	return client, nil
@@ -152,9 +160,14 @@ func NewClientUnix(conf ClientConfig) *Client {
 		serverURL: serverURL,
 		requester: requester,
 	}
-	err = client.registerWorker()
+	err = client.refreshAccessToken()
 	if err != nil {
-		panic(err)
+		logrus.Warnf("Error getting access token on startup, %v. Trying again later…", err)
+	} else {
+		err = client.registerWorker()
+		if err != nil {
+			logrus.Warnf("Error registering worker on startup, %v. Trying again later…", err)
+		}
 	}
 	go client.workerHeartbeat()
 	return client
--- a/internal/worker/client_test.go
+++ b/internal/worker/client_test.go
@@ -2,12 +2,13 @@ package worker_test
 
 import (
 	"encoding/json"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
-
-	"github.com/stretchr/testify/require"
 
 	"github.com/osbuild/osbuild-composer/internal/jobqueue/fsjobqueue"
 	"github.com/osbuild/osbuild-composer/internal/worker"
@@ -42,14 +43,13 @@ func newTestWorkerServer(t *testing.T) (string, string, string) {
 
 	/* Check that the worker supplies the access token  */
 	calls := 0
-	proxySrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if calls == 0 {
-			require.Equal(t, "Bearer", r.Header.Get("Authorization"))
+	proxySrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if req.Header.Get("Authorization") == "Bearer" {
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
-		require.Equal(t, "Bearer "+accessToken, r.Header.Get("Authorization"))
-		handler.ServeHTTP(w, r)
+		require.Equal(t, "Bearer "+accessToken, req.Header.Get("Authorization"))
+		handler.ServeHTTP(w, req)
 	}))
 	t.Cleanup(proxySrv.Close)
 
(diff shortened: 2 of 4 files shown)


cmd/osbuild-worker/main.go
```diff
@@ -269,7 +269,13 @@ func main() {
 			ProxyURL:     proxy,
 		})
 		if err != nil {
-			logrus.Fatalf("Error creating worker client: %v", err)
+			var logLevel logrus.Level
+			if client == nil {
+				logLevel = logrus.FatalLevel
+			} else {
+				logLevel = logrus.ErrorLevel
+			}
```
thozza, line 277: What's your rationale behind using a different log level based on `client` being `nil`?

schuellerf: Because Fatalf terminates the application and if client is nil we really can't continue.
(So a more severe thing than some connection problem)

cmd/osbuild-worker/main.go
```diff
@@ -269,7 +269,13 @@ func main() {
 			ProxyURL:     proxy,
 		})
 		if err != nil {
-			logrus.Fatalf("Error creating worker client: %v", err)
+			var logLevel logrus.Level
```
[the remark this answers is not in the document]
schuellerf: Hmm - IMHO it should just start, go to the "main loop" and _try_ registering there as this is the behavior later on anyway.
You are right that's odd - I'll try to refactor this code just maybe to the "main loop" somewhere

internal/worker/client.go
```diff
@@ -154,10 +150,6 @@ func NewClientUnix(conf ClientConfig) *Client {
 		server:    server,
 		requester: requester,
 	}
-	err = client.registerWorker()
-	if err != nil {
-		panic(err)
```
schuellerf, line 159: (…less panic 😅)

schuellerf:
`registerWorker()` is done with the first request and also in `workerHeartbeat()`, so no need to register (and possibly crash) on startup

thozza (approved):
LGTM, thanks

internal/worker/proxy_test.go
```diff
@@ -72,11 +74,24 @@ func (p *proxy) ServeHTTP(wr http.ResponseWriter, req *http.Request) {
...
+
 	resp, err := client.Do(req)
 	if err != nil {
 		http.Error(wr, "Server Error", http.StatusInternalServerError)
 	}
-	defer resp.Body.Close()
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
+	p.paths = append(p.paths, strconv.Itoa(p.calls)+": "+req.URL.Path+" Body: "+strings.TrimSpace(bodyString)+" (Response: "+resp.Status+")")
```
croissanne, line 94: Nitpick, fmt.Sprintf would be cleaner I think.

internal/worker/client_test.go
```diff
@@ -128,12 +128,11 @@ func TestProxy(t *testing.T) {
...
-	// we expect 6 calls to go through the proxy:
-	// - register worker
+	// we expect 5 calls to go through the proxy:
 	// - request job (fails, no oauth token)
 	// - oauth call
 	// - request job (succeeds)
 	// - upload artifact
 	// - cancel
-	require.Equal(t, 6, proxy.calls)
+	require.Equal(t, 5, proxy.calls, "We got those:\n"+strings.Join(proxy.paths, "\n"))
```
croissanne, line 137: Can we somehow test that the client still registers in a unit test?

croissanne (commented):
lgtm, 2 nitpicks

internal/worker/proxy_test.go
schuellerf: hmm - no idea why I didn't use Sprintf in the first place :-/

internal/worker/client_test.go
schuellerf: I would have registered only after 1 Minute. So I'll revise the implementation.

croissanne: Ahhh ic, yea no that's ok then, no point waiting a minute in a unit test.

schuellerf: I'll suggest a more straight forward implementation. (already started implementing)
It should not be the "usual behavior" to start the interaction with a failed request due to missing authentication in the first place ;)

mvo5 (commented):
Thank you! I really like the new LogrusTest helper. Some musings and comments inline for your consideration (but please take it all with a grain of salt, @croissanne is the authority in this codebase, follow her lead.

internal/worker/client.go
```diff
@@ -121,9 +121,14 @@ func NewClient(conf ClientConfig) (*Client, error) {
 		clientId:     conf.ClientId,
 		clientSecret: conf.ClientSecret,
 	}
-	err = client.registerWorker()
+	err = client.refreshAccessToken()
```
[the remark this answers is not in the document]
schuellerf: You are right, could be explained better - avoids always failing on the first request (as we don't have a token)

internal/worker/client.go
```diff
@@ -344,14 +355,8 @@ func (c *Client) RequestJob(types []string, arch string) (Job, error) {
...
 
-	workerID := func() uuid.UUID {
-		c.workerIDMu.RLock()
-		defer c.workerIDMu.RUnlock()
-		return c.workerID
-	}()
-
-	if workerID != uuid.Nil {
-		reqBody.WorkerId = common.ToPtr(workerID.String())
+	if c.getWorkerID() != uuid.Nil {
```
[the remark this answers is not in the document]
schuellerf: `getWorkedIDLocked` sounds to me like there could be another version without a lock - but there shouldn't be any (without a lock) but it could be renamed if that makes it more clear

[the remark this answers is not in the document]
schuellerf: … and yes - it's kind of racy - I'll change that

internal/worker/client_test.go
```diff
@@ -42,14 +43,13 @@ func newTestWorkerServer(t *testing.T) (string, string, string) {
 
 	/* Check that the worker supplies the access token  */
 	calls := 0
-	proxySrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if calls == 0 {
```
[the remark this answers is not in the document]
schuellerf: That's why I changed the behavior and the test. Why should we want a client to always start without a token? Causing an "error" which is not necessary, I guess.
I'm testing the (error-) case if a token was lost and re-fetching it, also.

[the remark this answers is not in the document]
schuellerf: @croissanne is there a real reason why we want a lazy token refresh _on startup_ ?
It is repeated if needed anyway, but that feels more like "in the case of an error/reconnect" to me

https://github.com/osbuild/osbuild-composer/pull/2433
https://github.com/osbuild/osbuild-composer/commit/42e5947b76784dde3609152c9dfe9c4535292205

internal/worker/client_test.go
```diff
@@ -102,38 +109,152 @@ func TestOAuth(t *testing.T) {
...
+			r := strings.NewReader("artifact contents")
+			require.NoError(t, job.UploadArtifact("some-artifact", r))
+			c, err := job.Canceled()
+			require.False(t, c)
+			require.NoError(t, err)
+
+			// we expect 5 or 7 calls to go through the proxy:
+			// depending, if we "loose" authentication or not
+			// * oauth call
+			// * register worker
```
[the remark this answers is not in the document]
schuellerf: not sure if this is "unrelated" as this client startup behavior would then tested be more explicitly when written as code, you are right.
I just added the "paths" to at least "see" anomalies there (but no reshuffling…). I'll translate this comment into code…

internal/worker/client_test.go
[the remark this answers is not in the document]
mvo5: Thanks! I did not read this careful enough. I wonder if we could make this explicit in the test then to keep the same spirit in the test. Before it was testing "the first call from the client to the proxxy is unauthorized", what about we now test "the first call from the client to the proxy is authorized"? Something like
```go
diff --git a/internal/worker/client_test.go b/internal/worker/client_test.go
index d330aecae..28235dba7 100644
--- a/internal/worker/client_test.go
+++ b/internal/worker/client_test.go
@@ -41,9 +41,15 @@ func newTestWorkerServer(t *testing.T) (string, string, string) {
        offlineToken := "someOfflineToken"
        accessToken := "accessToken!"
 
-       /* Check that the worker supplies the access token  */
-       calls := 0
+       /* Check that the worker supplies the access token */
+       oauthApiCalls := 0
+       proxySrvCalls := 0
        proxySrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+               // the oauth token is fetched before any proxy calls
+               if proxySrvCalls == 0 {
+                       require.Equal(t, oauthApiCalls, 1)
+               }
+               proxySrvCalls++
                if req.Header.Get("Authorization") == "Bearer" {
                        w.WriteHeader(http.StatusUnauthorized)
                        return
@@ -55,7 +61,7 @@ func newTestWorkerServer(t *testing.T) (string, string, string) {
 
        /* Start oauth srv supplying the bearer token */
        oauthSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-               calls += 1
+               oauthApiCalls += 1
                require.Equal(t, "POST", r.Method)
                err := r.ParseForm()
                require.NoError(t, err)

```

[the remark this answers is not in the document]
schuellerf: This is done implicitly - the order is given now in the tests
or didn't I understand you correctly here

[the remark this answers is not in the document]
mvo5: Aha, sorry, I did not see that latest diff, sorry for that. Yeah, with the latest diff it's even nicer because it's much more explicit (and arguably a better place).

internal/worker/client.go
```diff
@@ -121,9 +121,17 @@ func NewClient(conf ClientConfig) (*Client, error) {
...
+	// essentially saving one request on startup
+	err = client.refreshAccessToken()
 	if err != nil {
-		return client, err
+		logrus.Warnf("Error getting access token on startup, %v. Trying again later…", err)
+	} else {
+		err = client.registerWorker()
+		if err != nil {
+			logrus.Warnf("Error registering worker on startup, %v. Trying again later…", err)
+		}
```
croissanne, line 134: There's a lot more stuff here than before now. I think it's fine to let the lazy token do it's thing, a single 401 request at the start is really cheap. Getting the worker heartbeat to register immediately is tricky? Maybe just move the registerWorker call into the `workerheart` fn right before the loop?

internal/worker/client.go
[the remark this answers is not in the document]
croissanne: This is still racy, assign it to a variable first.

internal/worker/client_test.go
```diff
@@ -101,39 +108,171 @@ func TestOAuth(t *testing.T) {
...
+}
+
+func (h *LogCheckerHook) Fire(e *logrus.Entry) error {
+	require.Less(h.t, h.counter, len(h.checkMsgs), "Got %s, probably unexpected", e.Message)
+	require.Contains(h.t, e.Message, h.checkMsgs[h.counter])
+	h.counter += 1
+	return nil
+}
+
+func muteLogrusDuringTest(t *testing.T) {
```
croissanne, line 220: Is this actually needed? Without `-v` go test eats all the stdout and stderr I think.

internal/worker/export_client_test.go
```diff
@@ -0,0 +1,8 @@
+// helper functions for client_test.go
+package worker
+
+func (c *Client) InvalidateAccessToken() {
```
croissanne, line 4: I like testing this scenario, but maybe this is for another PR? Doesn't feel related to the crashing.

croissanne (commented):
There's a lot going on in the PR now, could we fix the crash separately and just change the logrus level on that one call? And then deal with the registering rework separately?

internal/worker/client.go
schuellerf: for the tests this is not really deterministic I guess - otherwise this could work

schuellerf: I thought it's cleaner to do the token and registration in sequence, in the normal "good case" and retry and do lazy otherwise

croissanne: Hmm we do lazy refreshes basically everywhere. And if it registers at the very start it should be pretty deterministic no? Like 1 call with a 401, and 1 successful? The accesstoken is behind a mutex so as long as the client does one operation at a time.

internal/worker/export_client_test.go
schuellerf: This is mainly because the tests did cover this before, so I wanted to restore/keep parity here

schuellerf: Should I mention that the test behavior changed and split to another PR? no big deal, if needed

cmd/osbuild-worker/main.go
```diff
@@ -121,12 +121,12 @@ func setProtection(protected bool) {
...
 }
 
-// Requests and runs 1 job of specified type(s)
+// RequestAndRunJob Requests and runs 1 job of specified type(s)
 // Returning an error here will result in the worker backing off for a while and retrying
 func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls map[string]JobImplementation) error {
 	logrus.Debug("Waiting for a new job...")
 	job, err := client.RequestJob(acceptedJobTypes, arch.Current().String())
-	if err == worker.ErrClientRequestJobTimeout {
+	if errors.Is(err, worker.ErrClientRequestJobTimeout) {
```
schuellerf, line 129: This part is unrelated - I'll remove from the PR

internal/worker/client.go
schuellerf: I mean doing it before the loop in `workerHeartbeat` might be non-deterministic with the main-loop who ever gets dispatched first (e.g. during tests)

schuellerf:
I removed the unrelated changes (e.g. to #4351 )
Hopefully this PR is now more readable

internal/worker/client.go
mvo5: It the possible option:

- do "registerWorker()" as before and instead of returning an error just log a warning

covered above? That one seems fine, it's a small change, it does stop the startup crash. The call will register and as part of the lazy refresh a token. if that fails the heartbeat will try again in a minute. or am I missing something?

croissanne: yea I'd lean towards that one as well, keeps the `New*()` simple and does the lazy token refresh as we do elsewhere.

schuellerf: I think there are three options
* explicitly get a token on startup, register, do lazy/token recovery later (current implementation of this PR)
   no 401 expected - only in case of network/deployment issues (during startup or later)
   also the behavior in case we can't contact composer is the same on startup or during runtime (just retry)
* get a token before going into workerHeartbeat-loop
   possibly racy if the client will do this first or the main loop will try a request first and lazy get a token
   here the startup might generate 401 or not
* just start the heartbeat (acting in 1min) and going to main-loop
   creates a 401 first for sure, which could be ok and works,
   but not necessary to generate an "error" in the "happy path" IMHO

Did I get them right?

regarding the behavior in general and related tests I think the first option is the cleanest and most predictable startup

mvo5: I thought about this a little bit and one nice aspect of doing the lazy refresh is that we ensure it works when doing it this way. It may sound silly but let's think about it: if we somehow break lazy refresh of the token we currently get errors left and right at startup because nothing will work and we will know immediately. If however we get the token explicitly and lazy-refresh-on-401 is broken somehow we will notice that only after the 900s time that suddenly something start failing and there are probably not that many things doing on after 900s so it maybe way later. So the pattern to spot what went wrong is harder to find.

schuellerf: (Ok - out of context - but still my point 😇)
> we currently get errors left and right at startup

But that's exactly my point. In the current implementation we _always_ get a 401 (incl. an warning which you assert in your test) even in the happy path.

Otherwise we only get a 401 and a refresh in case of connection errors, that seems more straight forward to me.

schuellerf:
Main problem solved by #4357

PR #4328, opened 2024-08-26
worker: move `api.BasePath` setup to the start of the funcs

mvo5:
I find it slightly eaiser to read this code when
`api.BasePath = conf.BasePath` is right at the top as it's unrelated to the parsing code below.

Note that the code itself is problematic:
- api.BasePath is global but client is not, this means that multiple client with different configs will result in api.BasePath being potentially wrong
- api.BasePath is set in a non-thread safe manner

Changing is a bigger job but we might consider it (IMHO).

### DIFF
--- a/internal/worker/client.go
+++ b/internal/worker/client.go
@@ -75,13 +75,12 @@ type tokenResponse struct {
 }
 
 func NewClient(conf ClientConfig) (*Client, error) {
+	api.BasePath = conf.BasePath
+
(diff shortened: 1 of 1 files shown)


schuellerf:
despite this flaw in design, I think we can as well "assert" on startup somehow that there is no second client and no second config, as this won't happen I guess
This would be a worker connected to two composers right?

schuellerf (approved):

PR #4329, opened 2024-08-26
client: extract endpoint() and newServerURL() helper

mvo5:
Small refactor, I am actually not sure it's worth it, so feel free to close as long term we want to move to konflux and a lot of this will probably go away. But maybe it is still useful. 

---

worker: add `endpoint()` helper that cannot fail and use it

Small commit to (hopefully) make the intent of the code more
clear when `serverURL.Parse()` is used. In most cases it is
used to generate an endpoint URL. As this is a realtive path
we do not need `Parse()` but can use `JoinPath()` instead which
has the nice benefit that it cannot error or panic.

----

worker: extract `newServerURL()` helper and add tests

Also explain why serverURL.JoinPath() cannot be used here.

### DIFF
--- a/internal/worker/client.go
+++ b/internal/worker/client.go
@@ -75,13 +75,12 @@ type tokenResponse struct {
 }
 
 func NewClient(conf ClientConfig) (*Client, error) {
+	api.BasePath = conf.BasePath
+
 	serverURL, err := url.Parse(conf.BaseURL)
 	if err != nil {
 		return nil, err
 	}
-
-	api.BasePath = conf.BasePath
-
 	serverURL, err = serverURL.Parse(api.BasePath + "/")
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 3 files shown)


internal/worker/client.go
```diff
@@ -222,13 +226,8 @@ func (c *Client) workerHeartbeat() {
 			}
 		}
 
-		url, err := c.serverURL.Parse(fmt.Sprintf("workers/%s/status", workerID()))
-		if err != nil {
-			logrus.Errorf("Error parsing worker status: %v", err)
```
mvo5, line 227: Even if we close this PR I think we should still fix this log message, because I think it's quite misleading. Unless I miss something here this can only happen if workerUUID() return something that creates an error from an URL parse - which should never happen but even if it does the error should be something like: `cannot construct URL for worker status` or similar and not be about parsing the worker status. When I read this I assumed `c.server.Parse()` would actually reach out to the worker which it does not (which partly promoted me to this flurry of PRs that (hopefully) make the code more explicit.

PR #4334, opened 2024-08-28
Worker: move GCE image guest OS features to upload target options

thozza:
Previously, the worker was determining the GCE image guest OS Features on its own, based on the OS name. This caused problems, in case the osbuild-composer was of a newer version than the worker.

Example:
osbuild-composer contained support for c10s GCE image type and its implementation also contained the proper guest OS Features list for it. However, when the worker got the osbuild job, it built it and tried to fetch the guest OS Features for the distro. Since its implementation was too old, it didn't contain the code that added the actual support for c10s GCE images and got no guest OS features list (which is the default for unsupported distros). The image was successfully uploaded and shared, but it does not boot in GCP, because it does not know that it should use UEFI to boot it.

This behavior could be considered a bug. The worker should be dumb. It should not be making decisions about the image features, but instead it should take them from the upload target options. And composer should be the authoritative source of truth for this. Because otherwise, we basically have two components that need to be updated in sync to add support for GCE images on a new distro.

Move the GCE image guest OS features to the GCP upload target options. The worker will just take what is specified there and use it when importing the image to GCP. As a compatibility layer for the case when the composer would be older than the worker (unlikely, but still), worker will try to determine the image guest OS features in case the list in the upload target options is empty.

Extend the GCP functional tests to check that the imported image has at least some guest OS features set.

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -807,9 +807,17 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 				break
 			}
 
+			guestOSFeatures := targetOptions.GuestOsFeatures
+			// TODO: Remove this after "some time"
+			// This is just a backward compatibility for the old composer versions,
+			// which did not set the guest OS features in the target options.
+			if len(guestOSFeatures) == 0 {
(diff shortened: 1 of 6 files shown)


bcl (approved):
Err, wow. That looks incredibly clean!

PR #4336, opened 2024-08-28
cloudapi: Add /distributions/ to cloudapi

bcl:
This pull request includes:

### DIFF
--- a/internal/cloudapi/v2/openapi.v2.yml
+++ b/internal/cloudapi/v2/openapi.v2.yml
@@ -526,6 +526,27 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /distributions:
+    get:
+      operationId: getDistributionList
+      summary: |-
+        Get all of the supported distribution repository details
+      security:
+        - Bearer: []
+      responses:
+        '200':
+          description: Distribution repositories
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DistributionList'
+        '500':
+          description: Unexpected error occurred
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
   /errors/{id}:
     get:
       operationId: getError
@@ -688,6 +709,25 @@ components:
               type: array
               items:
                 $ref: '#/components/schemas/ComposeStatus'
+
+    DistributionList:
+      properties:
+        map:
+          type: object
+          description: Distribution name
+          additionalProperties:
+            map:
+              type: object
+              description: Architecture name
+              additionalProperties:
+                map:
+                  type: object
+                  description: Image type name
+                  additionalProperties:
+                    type: array
+                    description: Repository used for this distro:arch:image-type
+                    items:
+                      $ref: '#/components/schemas/BlueprintRepository'
 
     ComposeStatus:
       allOf:
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,7 @@ require (
 	github.com/osbuild/images v0.112.0
 	github.com/osbuild/osbuild-composer/pkg/splunk_logger v0.0.0-20240814102216-0239db53236d
 	github.com/osbuild/pulp-client v0.1.0
+	github.com/pkg/errors v0.9.1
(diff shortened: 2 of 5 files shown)


bcl:
I'm thinking I want to extend this to return more, so converting to draft for now.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -548,6 +569,21 @@ components:
...
+          additionalProperties:
+            map:
+              type: object
+              additionalProperties:
+                map:
+                  type: object
+                  additionalProperties:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/BlueprintRepository'
```
thozza, line 585: This is IMHO impossible to read and determine the meaning of returned values. I'd suggest at least adding a description to each property.

thozza (commented):
Thanks for the change.

I don't understand the reason behind returning the full repository configuration for each distro/arch/image_type combination from this endpoint. I think that I'm missing the bigger picture behind this change, especially since the linked Jira task talks only about listing the matrix of supported distro:arch:image_type and nothing about repositories.

In addition to that, The summary of the PR is in contradiction to what this PR does - adding only `/distributions/` endpoint and not `/distributions/{arch}`.

Lastly, I'm not sure if it is a good practice for API to use values as object property names in the returned JSON. IOW, the implemented approach vs. to return a flattened array of objects such as:

```json
[
    {
        "distro": "test-distro-1",
        "arch": "test_arch",
        "image_types": [
            "test_type"
        ]
    },
    ...
]
```

bcl:
I'm trying to remember why I included the actual repo data, and I think it was because we've had some questions about exactly what repos are being used for a build and no way to query that info. It's useful info, and easily available,  so I included it.

As for the openapi layout, I couldn't figure out any other way to do it and have variable named keys. The reason for that is that it makes more sense to process the output using keys than it is to have to search through all the results for the distro+arch.

The output from this will be used with composer-cli in order to let the user explore what's available to use for the builds, so more detail is better.

kingsleyzissou (commented):
Regarding the repositories, this is how I'm consuming it on the frontend:

https://github.com/osbuild/image-builder-frontend/blob/8ad4b6a41836819c23602b6ea9c6ac610a532f77/src/store/cockpitApi.ts#L92-L99

In my case, a set of repositories would be a lot cleaner to consume rather than the list of repositories for each image type, but I'm not sure if this is desirable from the cli POV. 

Also a flattened array of objects like @thozza suggested would be nice but this is also workable

bcl:
> In my case, a set of repositories would be a lot cleaner to consume rather than the list of repositories for each image type, but I'm not sure if this is desirable from the cli POV.

It's possible to have different repositories for each image type so you have to list them, even if they are normally identical. eg. `gce` has google specific repos added to it.

mvo5:
We have a discussion about a similar feature in https://github.com/osbuild/images/pull/1167 and https://github.com/osbuild/images/pull/1173 for the CLI. The idea is very similar, show the available distros in a map like format but there the keys are slightly different. I would like to make sure we are aligned here (or at least talked about it). 

For a user visible output `distro->imgType->arch` produces (arguably) nicer output, would a similar map style work for this API too or is there a deeper reason to use `distro->arch->imgType`? While its two different things I would like to align things when possible (also cc @schuellerf )

bcl:
It makes more sense to me to have `distro->arch->imageType` for (at least) 2 reasons. Inputs to the compose/depsolve/search require distro and arch as part of the query, so in order to get the list of images you need to know distro and arch first. And the image types can add their own repositories, so each image type's repo list depends on knowing distro and arch first.

schuellerf:
The structure I implemented was more from a user perspective where the possible image types for a given distro might be more interesting to visually see first, but I'm totally for consistency and I'll adjust the listing in images to match your structure ( being `distro->arch->imageType` ) it will still be readable I guess.
Thanks @mvo5 for closing the loop and starting this discussion.

thozza (approved):
LGTM

kingsleyzissou (approved):

PR #4341, opened 2024-09-02
osbuild-worker: handle error wrapping from dnfjson package

croissanne:
osbuild/images#751 wrapped the errors in the images/dnfjson package to provide more details, the depsolve job should take this into account to map the dnfjson error to the correct worker client error.

This caused user input errors errors to be misclassified as internal errors, triggering depsolve job failure alerts.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


schuellerf (approved):
/LGTM
Thanks for finding this (for me 😑)!

mvo5 (approved):
Thank you! Out of curiosity, what error causes this issue? is it the one from `makeDepsolveRequest()` that is now wrapped?

croissanne:
> Thank you! Out of curiosity, what error causes this issue? is it the one from `makeDepsolveRequest()` that is now wrapped?

The one in `dnfjson.Solver.Depsolve`.

PR #4348, opened 2024-09-05
tests: fix api.sh gcp ssh connection timeout

jrusz:
This pull request includes:

### DIFF
--- a/test/cases/api/common/common.sh
+++ b/test/cases/api/common/common.sh
@@ -19,6 +19,20 @@ function _instanceCheck() {
   echo "✔️ Instance checking"
   local _ssh="$1"
 
+  # Retry loop to wait for instance to be ready
+  # This is here especially because of gcp test
+  RETRIES=10
+  for i in $(seq 1 $RETRIES); do
+    echo "Attempt $i of $RETRIES: Checking instance status..."
+    if eval "$_ssh true"; then
+      echo "Instance is up and ready!"
+      break
+    else
+      echo "Instance is still booting or SSH key not propagated, retrying in 30 seconds..."
+      sleep 30
+    fi
+  done
+
   # Check if postgres is installed
   $_ssh rpm -q postgresql dummy
 


bcl:
Can you just pass it after the `--`? Otherwise I think you need to escape the `"`s in order to get them to pass to the command instead of being evaluated by bash.

jrusz:
> Can you just pass it after the `--`? Otherwise I think you need to escape the `"`s in order to get them to pass to the command instead of being evaluated by bash.

I've been trying different variations over the week while working on other stuff and I've finally found something that seems to work :sweat_smile:

jrusz:
Unfortunately it looks like it doesn't help resolve our issue https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/7807093869 
I'm going to try something else.

jrusz:
I was unable to come into agreement with the `gcloud` command so I've decided to implement our own retry mechanism which worked when I tested it on the side, I hope other tests are fine with it as well. Here is where it helped prevent the timeout https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/7809695856

test/cases/api/common/common.sh
```diff
@@ -19,6 +19,20 @@ function _instanceCheck() {
...
+      echo "Instance is up and ready!"
+      break
+    else
+      echo "Instance is still booting or SSH key not propagated, retrying in 30 seconds..."
+      sleep 30
+    fi
+  done
+
   # Check if postgres is installed
   $_ssh rpm -q postgresql dummy
```
bcl, line 37: I suppose you can now drop this line

test/cases/api/common/common.sh
```diff
@@ -19,6 +19,20 @@ function _instanceCheck() {
   echo "✔️ Instance checking"
   local _ssh="$1"
 
+  # Retry loop to wait for instance to be ready
```
bcl, line 22: I'd be tempted to add an initial delay shorter than 30s so that there is some chance of catching it on the first try and not always waiting 30s -- but I also don't know how long it takes these things to startup :)

jrusz: Well it will try it once right away before waiting which should work in the vast majority of cases. I'm aware of only the gcp test on 8.10-ga where it takes so long that it fails. The alternative there was to wait for the whole test run again which is ~30 minutes, so 30 seconds compared to that is nothing and I'd prefer to not complicate this :)

test/cases/api/common/common.sh
jrusz: Yeah, I was thinking that it might be more explicit this way, but it does check it in the loop and now it's ran twice in the test runs, I'll remove it.

test/cases/api/common/common.sh
```diff
@@ -19,8 +19,20 @@ function _instanceCheck() {
...
 
-  # Check if postgres is installed
-  $_ssh rpm -q postgresql dummy
+  # Retry loop to wait for instance to be ready
+  # This is here especially because of gcp test
+  RETRIES=10
+  for i in $(seq 1 $RETRIES); do
+    echo "Attempt $i of $RETRIES: Checking instance status..."
+    # Check if postgres is installed
+    if eval "$_ssh rpm -q postgresql dummy"; then
```
thozza, line 28: Would it make sense to use just something like `"$_ssh true"` instead? This could fail also if the packages are not installed and it would retry multiple times, no?

jrusz: Riiight, I didn't think of that, I was happy to have something that works after trying out the different variations of `gcloud` ssh flags while fighting with bash :sweat_smile: ... I'll update it, thanks!

test/cases/api/common/common.sh
thozza: This is not relevant any more.

thozza (approved):
LGTM, thanks :heart:

PR #4349, opened 2024-09-05
Openscap compliance facts (HMS-2836)

croissanne:
Also adds the policy id to the blueprint, this doesn't have any effect
    on the openscap step, it just puts in place the rhsm fact so instances
    registered to insights will appear under that policy.

---


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -975,6 +975,12 @@ components:
       required:
         - profile_id
       properties:
+        policy_id:
```
croissanne, line 978: I'm unsure if this belongs here, could add a separate 'Compliance' customisation.

kingsleyzissou: Yeah this is a fair point. I don't have strong feelings though

kingsleyzissou (approved):

PR #4351, opened 2024-09-06
internal/worker/client.go: refactor reading worker ID

schuellerf:
Just refactor the usage of the mutex to remove code duplication

### DIFF
--- a/internal/worker/client.go
+++ b/internal/worker/client.go
@@ -203,16 +203,18 @@ func (c *Client) registerWorker() error {
 	return nil
 }
 
+func (c *Client) getWorkerID() uuid.UUID {
+	c.workerIDMu.RLock()
+	defer c.workerIDMu.RUnlock()
+	return c.workerID
+}
+
 func (c *Client) workerHeartbeat() {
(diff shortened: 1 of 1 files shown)


croissanne (changes requested):
LGTM, but the commit message isn't very clear, there are two mutexes in the worker client, and it's not calling to the mutex, it's reading the worker ID.

Can you change the commit message to something like:

`internal/worker/client:` or `worker(client):`  + `refactor reading worker ID`

```
Adds a helper function to the worker client instead of redeclaring the same inline function everywhere.
```

schuellerf:
> LGTM, but the commit message isn't very clear, there are two mutexes in the worker client, and it's not calling to the mutex, it's reading the worker ID.
> 
> Can you change the commit message to something like:
> 
> `internal/worker/client:` or `worker(client):` + `refactor reading worker ID`
> 
> ```
> Adds a helper function to the worker client instead of redeclaring the same inline function everywhere.
> ```

Thanks - the message was really bad :-/
please check the updated one

mvo5 (approved):
Nice work!

PR #4353, opened 2024-09-06
upload/azure: turn off public access on storage accounts

croissanne:
When creating a storage account we're currently allowing public blob access, users might have compliance policies active on their azure accounts which forbid this.

---

Even though the default should be false, we're still getting
```
Storage account Allow Blob Public Access should be disallowed
```

so let's just do it explicitly.

### DIFF
--- a/internal/upload/azure/azure.go
+++ b/internal/upload/azure/azure.go
@@ -111,7 +111,8 @@ func (ac Client) CreateStorageAccount(ctx context.Context, resourceGroup, name, 
 			tag.Name: &tag.Value,
 		},
 		Properties: &armstorage.AccountPropertiesCreateParameters{
-			MinimumTLSVersion: common.ToPtr(armstorage.MinimumTLSVersionTLS12),
+			AllowBlobPublicAccess: common.ToPtr(false),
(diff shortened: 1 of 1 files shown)


croissanne:
For testing I disabled public blob access on all our storage accounts, so if the azure tests are green it should be fine.

ondrejbudai (approved):
Awesome! :) Btw, they can also create the storage account themselves if they want, they just need to tag it correctly so osbuild-composer finds it.

achilleas-k (approved):

achilleas-k:
> Awesome! :) Btw, they can also create the storage account themselves if they want, they just need to tag it correctly so osbuild-composer finds it.

Should this be documented in the release notes?  We can add instructions for how to get the old behaviour by creating the storage beforehand.

croissanne:
> Awesome! :) Btw, they can also create the storage account themselves if they want, they just need to tag it correctly so osbuild-composer finds it.

Indeed, but this particular user is having trouble with that. I don't fully understand why because it looks properly tagged ` "imageBuilderStorageAccount": "location=eastus"`, but still it was trying to create a storage account.

PR #4359, opened 2024-09-11
[WIP] Add initial support for generating SPDX SBOM documents (COMPOSER-2274)

thozza:
**Still an early WIP.**

This pull request includes:


internal/cloudapi/v2/handler.go
```diff
@@ -888,6 +891,167 @@ func (h *apiHandlers) getComposeManifestsImpl(ctx echo.Context, id string) error
...
+	}
+
+	var imageSBOMs []ImageSBOM
+	for _, manifestDepUUID := range manifesJobInfo.Deps {
+		depJobType, err := w.JobType(manifestDepUUID)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to get job type for dependency %q: %v", manifestDepUUID, err)
+		}
+
+		if depJobType == worker.JobTypeDepsolve {
```
achilleas-k, line 927: NITPICK: Flipping the condition with `continue` would decrease the indentation for the largest part of the loop body, which is nicer IMO.

internal/cloudapi/v2/handler.go
```diff
@@ -888,6 +891,167 @@ func (h *apiHandlers) getComposeManifestsImpl(ctx echo.Context, id string) error
...
+				imageSBOMs = append(imageSBOMs, ImageSBOM{
+					PipelineName:    pipelineName,
+					PipelinePurpose: purpose,
+					Sbom:            sbomDoc.Document,
+					SbomType:        sbomType,
+				})
+			}
+
+			// There should be only one depsolve job per OSBuild job
+			break
```
achilleas-k, line 961: Should this `break` be inside the `depJobType == worker.JobTypeDepsolve` block?  The way it is now, only one the first `manifestJobInfo.Deps` element is considered.  Is that always a depsolve job and nothing else?  I guess no, otherwise we'd just be using `manifestJobInfo.Deps[0]` right?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -215,6 +215,43 @@ paths:
...
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+            example: 123e4567-e89b-12d3-a456-426655440000
+          required: true
+          description: ID of compose SBOMs to get
+      description: |-
+        Get the manifests of a running or finished compose.
```
achilleas-k, line 232: Copy-paste leftover?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -215,6 +215,43 @@ paths:
...
+      summary: Get the SBOMs for a compose.
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+            example: 123e4567-e89b-12d3-a456-426655440000
+          required: true
+          description: ID of compose SBOMs to get
```
achilleas-k, line 230: NITPICK: This is a bit awkwardly phrased and might be misunderstood as "ID of SBOMs".  In other words, it kinda sounds like "compose-SBOM" is the object being fetched, so it wants the ID of that, but instead it's the compose ID it needs.

Maybe something like:
> ID of compose for which to get SBOMs

though I'm not too happy about that one either.

achilleas-k (commented):
Great work.  This is _a lot_.

Couple of notes and nitpicks.  I think there's a bug in the `sbomsFromOSBuildJob()` function (noted below) but I'm not entirely sure so not Requesting changes.

internal/cloudapi/v2/openapi.v2.yml
thozza: yes.

internal/cloudapi/v2/openapi.v2.yml
thozza: I agree that this is not great by any means :smile: I'll happily use your suggestion.

internal/cloudapi/v2/handler.go
thozza: Yeah, I don't see any problem with it.

internal/cloudapi/v2/handler.go
thozza: > Should this break be inside the depJobType == worker.JobTypeDepsolve block?
It actually is inside block as far as I can tell :thinking: 

So the break is at the correct place, but this is a bit of a premature optimization based on the current implementation. Let me explain.

I implemented `sbomsFromOSBuildJob()` in a way, so that it uses the same approach to get SBOMs from `OSBuildJob`, regardless if it is from a Koji or non-Koji compose.

Just for the completeness, for Koji composes, the `DepsolveJob` is a dependency of the `OSBuildJob`, so that it can access SBOMs and upload them to Koji. For non-Koji composes, the `DepsolveJob` is not a direct dependency of the `OSBuildJob` and one must go through the `ManifestJob` as to get to `DepsolveJob`.

There's always only a single `DepsolveJob` in the compose job chain, thus the break. But there may be more than one item in `manifestJobInfo.Deps` and `DepsolveJob` does not have a fixed index.

achilleas-k: Right.  Thanks for the explanation.  I think I just read it wrong :sweat_smile:

achilleas-k (approved):
All good!  :shipit:

internal/cloudapi/v2/openapi.v2.yml
thozza: fixed.

kingsleyzissou (approved):
Also LGTM

internal/cloudapi/v2/handler.go
```diff
@@ -888,6 +891,169 @@ func (h *apiHandlers) getComposeManifestsImpl(ctx echo.Context, id string) error
...
+	}
+
+	pipelineNameToPurpose := func(pipelineName string) (ImageSBOMPipelinePurpose, error) {
+		if slices.Contains(osbuildJobResult.PipelineNames.Payload, pipelineName) {
+			return ImageSBOMPipelinePurposeImage, nil
+		}
+		if slices.Contains(osbuildJobResult.PipelineNames.Build, pipelineName) {
+			return ImageSBOMPipelinePurposeBuildroot, nil
+		}
+		return "", fmt.Errorf("Pipeline %q is not listed as either a payload or build pipeline", pipelineName)
```
schuellerf, line 909: Why return `""` here and `nil` in other cases?

schuellerf (approved):
lgtm

PR #4360, opened 2024-09-11
tools/build-rpms: increase rpm builder instance disk size

croissanne:
The instance created for rpm builds is running out of space on the default 10G disk, let's double it.

### DIFF
--- a/tools/build-rpms.py
+++ b/tools/build-rpms.py
@@ -106,11 +106,22 @@ def create_ec2_instances(cleanup_actions, args, keypair):
             }
         ]
 
+        img = ec2.describe_images(ImageIds=[arch_info[a]["ImageId"]])
         instance = ec2.create_instances(
             ImageId=arch_info[a]["ImageId"],
             MinCount=1,
             MaxCount=1,
             InstanceType=arch_info[a]["InstanceType"],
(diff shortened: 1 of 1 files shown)


schuellerf (approved):
LGTM

tools/build-rpms.py
```diff
@@ -106,11 +106,22 @@ def create_ec2_instances(cleanup_actions, args, keypair):
...
 
+        img = ec2.describe_images(ImageIds=[arch_info[a]["ImageId"]])
         instance = ec2.create_instances(
             ImageId=arch_info[a]["ImageId"],
             MinCount=1,
             MaxCount=1,
             InstanceType=arch_info[a]["InstanceType"],
+            BlockDeviceMappings=[
+                {
+                    "DeviceName": img['Images'][0]['RootDeviceName'],
```
schuellerf, line 117: It probably is fine to just crash here if the arrays don't match up, right?

croissanne: yea but also every image should have at least 1 block device attached to it by default.

schuellerf: Sure, I approved anyway "as it should not happen" but rather if it would - is a stacktrace ok (I guess so…) or a more controlled error

PR #4374, opened 2024-09-19
cloudapi: Add architecture support to the optional Blueprint

bcl:
If included it overrides the architecture in the compose image request.

This pull request includes:


bcl:
Changed this to address @thozza 's comments about arch handling in #4390

achilleas-k (approved):

thozza (dismissed):
While this PR is OK from the implementation point of view, I don't like it from the architecture point of view...

I'm still not fully on board with the plan to add the `architecture` to the CloudAPI version of BP. The reason is that it duplicates the information already specified in the  `ImageRequest` https://github.com/osbuild/osbuild-composer/blob/d531f62488cae6b1dc475cb238a7e7c7153a05ac/internal/cloudapi/v2/openapi.v2.gen.go#L833-L856

Personally, I think that the fact that `arch` was added to the Weldr API BP version was a mistake. Instead, it should have been added to the `ComposeRequest` https://github.com/osbuild/osbuild-composer/blob/d531f62488cae6b1dc475cb238a7e7c7153a05ac/internal/weldr/api.go#L2417-L2424 struct for the Compose API call option when building an image for a specific BP.

Hardcoding the `arch` in the BP itself IMHO makes it less flexible to use for building various image types, and potentially for various arches. Architecture should be IMO treated in the same way as image type - you specify the BP name and when building the image, you pick the image type and an architecture.

I'm worried that we will make things more confusing and architecturally wrong if we add architecture to the BP in Cloud API.

I'd like to have a broader dicsussion and a consensus before we merge this.

bcl:
All I'm trying to do here is make things consistent. Arch ALREADY exists in the blueprint, except that it wasn't added to cloudapi, so the current situation is that it's more confusing.
I don't think this is the place to discuss whether or not it should have been added, it is already there.
Also note that the arch is optional, so it only makes things less flexible if it has been set and that's under the control of the user. Maybe they have a blueprint they only want to use on a specific arch.

PR #4378, opened 2024-09-23
composer: don't create RepoRegistry using reporegistry.New()

thozza:
The `reporegistry.New()` has been enhanced to return an error, in case there were no repositories loaded. This was to fix the situation in many unit tests, which were previously not loading any repositories and silently not running any tests.

This however broke our SaaS deployment, where we actually do not configure any repositories on the filesystem. As a result, osbuild-composer started to fail on it.

Workaround this situation in osbuild-composer by reverting to the old behavior by loading the repo configs separately and then using the loaded repos (which could be empty map) to initialize the RepoRegistry.

Also update the `tools/provision.sh` to delete all repositories in the Service scenario.

This pull request includes:

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -77,10 +77,11 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Compos
 		return nil, fmt.Errorf("failed to configure distro aliases: %v", err)
 	}
 
-	c.repos, err = reporegistry.New(repositoryConfigs)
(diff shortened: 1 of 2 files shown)


thozza:
FTR, I'm happy to fix this in a different way in `images`, just wanted to fix composer ASAP :innocent:

achilleas-k (approved):
Good to merge if it fixes prod.  For the _real_ fix, maybe we could return a special error from `New()` of type `reporegistry.NoReposError` or something and then callers can check for that.

PR #4369, opened 2024-09-17
Add fedora-41

croissanne:
Adds fedora 41 to repositories, and enables some of the tests.

PR #4380, opened 2024-09-24
Revert "Add fedora-41"

croissanne:
Reverts osbuild/osbuild-composer#4369

---

Opening a PR to revert this, because the fedora jobs currently take an absurd amount of time (130 minutes). It's unclear why but fetching the metadata from rpmrepo takes way too long.


jrusz (approved):
LGTM

PR #4385, opened 2024-09-26
cloud: fixed typo UnfulfillableCapacity (HMS-4676)

lzap:
While I was investigating a customer report I noticed this small typo which could actually cause the fallback not to work at all. The correct terms are linked from AWS Go library:

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

Related issue: https://issues.redhat.com/browse/HMS-4676


croissanne (approved):
thank you :heart:

PR #4387, opened 2024-09-27
cloud: new AWS metrics for spot instances

lzap:
This is a followup for https://github.com/osbuild/osbuild-composer/pull/4385

The goal here is to increase observability of AWS composer integration. Couple of histograms and counters from the most important one is to me the create fleet one which allows us to measure how many of fleet runs were successful and out of these how many of these were actually spot or on-demand.

I am doing this because previously there was a type and it appears that the hosted composer were not able to do a proper on-demand fallback. We can verify the behavior after promotion push or even start actively monitoring this metric on the dashboard.

### DIFF
--- a/internal/cloud/awscloud/awscloud.go
+++ b/internal/cloud/awscloud/awscloud.go
@@ -18,6 +18,7 @@ import (
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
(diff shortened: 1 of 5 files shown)


croissanne (changes requested):
Note that the aws cloud / secure-instance part is run on the worker. There's nothing scraping prometheus data there.

lzap:
> Note that the aws cloud / secure-instance part is run on the worker. There's nothing scraping prometheus data there.

Hmmm, I assume some of these metrics will not be exported. I guess createFleet function is called by the scheduler because worker runs on that instance?

I don’t have time to work on pushgateway at the moment, but we can discuss this next week.

PR #4389, opened 2024-09-27
repositories: add fedora 41

croissanne:
This way users could at least build fedora 41, there is currently an issue in rpmrepo where the fedora 41 branched repositories are very slow, so enabling CI is currently not possible.

https://github.com/osbuild/rpmrepo/issues/111


repositories/fedora-41.json
```diff
@@ -0,0 +1,114 @@
...
+      "check_gpg": true
+    },
+    {
```
ondrejbudai, line 16: Can you remove `modular`? This repo doesn't exist since F39. See https://fedoraproject.org/wiki/Changes/RetireModularity

croissanne: ah sure

ondrejbudai (approved):

PR #4390, opened 2024-09-27
Add depsolving and package search to cloudapi

bcl:
This pull request includes:


bcl:
Note, this builds on the depsolving PR so I've closed it in favor of this one. Last 2 commits are package search and the previous ones lay the groundwork and implement depsolve.

internal/cloudapi/v2/compose.go
```diff
@@ -1144,7 +1147,12 @@ func (request *ComposeRequest) GetImageRequests(distroFactory *distrofactory.Fac
 	}
 	var irs []imageRequest
 	for _, ir := range *request.ImageRequests {
-		arch, err := distribution.GetArch(ir.Architecture)
+		reqArch := ir.Architecture
+		// If there is an architecture in the blueprint it overrides the request's arch
+		if len(bp.Arch) > 0 {
+			reqArch = bp.Arch
+		}
```
thozza, line 1158: Just overriding the value does not seem like the best solution. We should ensure that the values in the image request and BP match if both are set or produce an error if they don't (as it is an invalid request).

internal/cloudapi/v2/compose_test.go
```diff
@@ -848,6 +848,33 @@ func TestGetImageRequests_BlueprintDistro(t *testing.T) {
...
+			Architecture:  "x86_64",
+			ImageType:     ImageTypesAws,
+			UploadOptions: &uo,
+			Repositories:  []Repository{},
+		},
+		Blueprint: &Blueprint{
+			Name:         "arch-test",
+			Architecture: common.ToPtr("aarch64"),
+		},
+	}
```
thozza, line 867: IMHO, this compose request should not be treated as valid. The architecture values don't match.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -397,6 +397,50 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /depsolve/blueprint:
+    post:
+      operationId: postDepsolveBlueprint
+      summary: Depsolve one or more blueprints
```
thozza, line 403: This endpoint depsolves only a single BP.

internal/cloudapi/v2/compose.go
```diff
@@ -35,18 +35,18 @@ func (bcpm BlueprintCustomizationsPartitioningMode) String() string {
...
-// GetCustomizationsFromBlueprintRequest populates a blueprint customization struct
-// with the data from the blueprint section of a ComposeRequest, which is similar but
+// GetCustomizationsFromBlueprint populates a blueprint customization struct
+// with the data from request Blueprint, which is similar but
 // slightly different from the Cloudapi's Customizations section
 // This starts with a new empty blueprint.Customization object
 // If there are no customizations, it returns nil
-func (request *ComposeRequest) GetCustomizationsFromBlueprintRequest() (*blueprint.Customizations, error) {
-	if request.Blueprint.Customizations == nil {
+func (rbp *Blueprint) GetCustomizationsFromBlueprintRequest() (*blueprint.Customizations, error) {
```
thozza, line 43: The function name should be `GetCustomizationsFromBlueprint`, judging from the comment above it.

thozza (commented):
While this PR delivers on its promises, I don't like the approach from an architectural point of view.

The server handles the depsolving and package search functionality directly, which is wrong. We already have a `DepsolveJob` type that the worker handles. So, the correct approach should be for the server to enqueue the `DepsolveJob` and use its result.

The same applies to package search—a new worker Job type should be added, and the search should occur on the worker.

internal/cloudapi/v2/compose.go
bcl: No, the value in the blueprint should override the other (or maybe the request should override the optional arch in the bp, but they are not expected to match). This is how it works in the weldr API, if arch is set in a blueprint it overrides the default host arch.

bcl:
> So, the correct approach should be for the server to enqueue the `DepsolveJob` and use its result.

What's the benefit of doing that? We already have the solver available in the server.

internal/cloudapi/v2/compose.go
thozza: The client should ensure a valid API call. This would mean extracting the architecture from the BP and making the correct compose request. The question is whether the architecture property should be added to the BP structure. Specifically, the API is already structured so that the requests are for specific environments and architecture, while BP is agnostic of these properties.

The Cloud API has no notion of the host arch, and everything must be specified in the request.

thozza:
> > So, the correct approach should be for the server to enqueue the `DepsolveJob` and use its result.
> 
> What's the benefit of doing that? We already have the solver available in the server.

I can think of these:

- To maintain reasonable architecture of Cloud API and not taint it with how things used to be done in Weldr API (which made a lot of assumptions of everything running on the same host).
- To enable the SaaS version, eventually take advantage of any functionality you'll add (e.g., the package search).
- The fact that the solver and the worker are on the same host as the API server in the on-prem scenario is an implementation detail in this case.

kingsleyzissou (commented):
Would it be possible to split the package search out into another PR? 

Merely for the fact that package search is more pressing than depsolve for the frontend

bcl:
Replaced by #4564

PR #4395, opened 2024-10-02
tests/regression: Add config for v3 certificates

jrusz:
When generating x509 v3 certs we need to explicitely set "CA:TRUE" otherwise they're not trusted to be used.

This pull request includes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -246,16 +246,16 @@ Base:
 regression-composer-works-behind-satellite-fallback:
   extends: .regression
   rules:
-    # BLACKLIST: Skipped on subscribed RHEL machines
-    - if: $RUNNER !~ "/^.*(rhel-.*-ga|centos|fedora).*$/" && $RUNNER !~ "/^.*(rhel-9.5|rhel-10.0).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    # WHITELIST: Run on RHEL-nightly only
+    - if: $RUNNER =~ "/^.*(rhel-.*-nightly).*$/" && $CI_PIPELINE_SOURCE != "schedule"
   variables:
     SCRIPT: regression-composer-works-behind-satellite-fallback.sh
 
 regression-composer-works-behind-satellite:
   extends: .regression
   rules: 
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -247,15 +247,15 @@ regression-composer-works-behind-satellite-fallback:
   extends: .regression
   rules:
     # BLACKLIST: Skipped on subscribed RHEL machines
```
schuellerf, line 249: Is this still valid?
It skips all except nightly?

.gitlab-ci.yml
```diff
@@ -247,15 +247,15 @@ regression-composer-works-behind-satellite-fallback:
...
-    - if: $RUNNER !~ "/^.*(rhel-.*-ga|centos|fedora).*$/" && $RUNNER !~ "/^.*(rhel-9.5|rhel-10.0).*$/" && $CI_PIPELINE_SOURCE != "schedule"
+    - if: $RUNNER !~ "/^.*(rhel-.*-ga|centos|fedora).*$/" && $CI_PIPELINE_SOURCE != "schedule"
```
schuellerf, line 250: Is it easier and more readable to make it non-inverted?
```suggestion
    - if: $RUNNER ~ "/^.*(rhel-.*-nightly).*$/" && $CI_PIPELINE_SOURCE != "schedule"
```

.gitlab-ci.yml
jrusz: I believe so, it's been like this ever since, here is a commit from 2 years ago that setup these rules and removed them from the scripts. https://github.com/osbuild/osbuild-composer/commit/1873c561cf799c0cb3552361b8f7e7e4745830d2

.gitlab-ci.yml
jrusz: Yeah sure that looks better, thanks!

.gitlab-ci.yml
schuellerf: I was just wondering if we should update the comment to match the implementation?
The comment should describe the condition, right? Is this the functionality that "nightly" is subscribed and the other aren't?

schuellerf (approved):
lgtm

PR #4398, opened 2024-10-04
internal/worker/server: return an error on depsolve timeout (HMS-2989)

schuellerf:
Fixes the special case that if no worker is available and we generate an internal timeout and cancel the depsolve, including all followup jobs, no error was propagated.

For details, see https://issues.redhat.com/browse/HMS-2989

Now all jobs (in the DB) are set to "failed":
 * All `jobs` entries get a "job_error" in the `result` as well as
 * `started_at` and `failed_at` are set to "now" and
 * `token` is set to a random UUID to comply with the DB schema

This pull request includes:

This is set to draft for now until:
* I get feedback that this approach is ok
* tests are implemented
* we decided on if we should "create a separate metrics" (see comment in the code)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
       run: sudo apt-get update
       # gpgme-devel is needed for container upload dependencies
     - name: Install test dependencies
-      run: sudo apt-get install -y libgpgme-dev
+      run: sudo apt-get install -y libgpgme-dev libbtrfs-dev
     - uses: actions/checkout@v4
     - uses: actions/setup-go@v5
       with:
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,6 @@ push-check: lint build unit-tests srpm man
 	./tools/check-runners
 	./tools/check-snapshots --errors-only .
 	rpmlint --config rpmlint.config $(CURDIR)/rpmbuild/SRPMS/*
-	./tools/prepare-source.sh
 	@if [ 0 -ne $$(git status --porcelain --untracked-files|wc -l) ]; then \
 	    echo "There should be no changed or untracked files"; \
 	    git status --porcelain --untracked-files; \
@@ -350,6 +349,7 @@ SHELLCHECK_FILES=$(shell find . -name "*.sh" -not -regex "./vendor/.*")
 
 .PHONY: lint
 lint: $(GOLANGCI_LINT_CACHE_DIR) container_composer_golangci_built.info
+	./tools/prepare-source.sh
 	podman run -t --rm -v $(SRCDIR):/app:z -v $(GOLANGCI_LINT_CACHE_DIR):/root/.cache:z -w /app $(GOLANGCI_COMPOSER_IMAGE) golangci-lint run -v
 	echo "$(SHELLCHECK_FILES)" | xargs shellcheck --shell bash -e SC1091 -e SC2002 -e SC2317
 
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -264,7 +264,7 @@ func (s *Server) enqueueCompose(irs []imageRequest, channel string) (uuid.UUID, 
 
 	s.goroutinesGroup.Add(1)
 	go func() {
-		serializeManifest(s.goroutinesCtx, manifestSource, s.workers, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID, id, ir.manifestSeed)
+		serializeManifest(s.goroutinesCtx, manifestSource, s.workers, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID, ir.manifestSeed)
 		defer s.goroutinesGroup.Done()
 	}()
 
@@ -424,7 +424,7 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
 		// copy the image request while passing it into the goroutine to prevent data races
 		s.goroutinesGroup.Add(1)
 		go func(ir imageRequest) {
-			serializeManifest(s.goroutinesCtx, manifestSource, s.workers, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID, buildID, ir.manifestSeed)
+			serializeManifest(s.goroutinesCtx, manifestSource, s.workers, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID, ir.manifestSeed)
 			defer s.goroutinesGroup.Done()
 		}(ir)
 	}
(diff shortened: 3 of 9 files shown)


schuellerf:
On the frontend this looks like this:
Screenshot_20241004_120837

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -94,6 +94,12 @@ const (
 		WHERE id = $1 AND finished_at IS NULL
 		RETURNING type, started_at`
 
+	sqlFailJob = `
```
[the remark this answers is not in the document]
croissanne: Actually this is fine to have a separate query.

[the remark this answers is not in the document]
schuellerf: From how I understood the behavior of the DB, the token, started_at, finished_at have to be set in order to get to a "failed" state where no jobs have been started (our case here)
otherwise it might appear in "requeue" and other queries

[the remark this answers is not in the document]
schuellerf: Hopefully I understood this correctly

internal/cloudapi/v2/server.go
```diff
@@ -435,23 +436,80 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
...
 	var dynArgs []json.RawMessage
 	var err error
+	token := uuid.Nil
 	logWithId := logrus.WithField("jobId", manifestJobID)
+
+	defer func() {
+		// token == uuid.Nil indicates that no worker even started processing
+		if token == uuid.Nil {
+			if jobResult.JobError != nil {
+				// set all jobs to "failed"
```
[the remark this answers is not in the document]
schuellerf: The error message in `result` column would be something else right? (not indicating the originating problem?)

schuellerf:
@croissanne last question for me is:
should we create separate metrics? maybe in a followup PR?
```
func (s *Server) SetFailed(id uuid.UUID, error *clienterrors.Error) error {
	/* create a separate metrics?
…
```

internal/cloudapi/v2/server.go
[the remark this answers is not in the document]
croissanne: Well that's why we chain the errors when building the error message. If done correctly the errors of all child jobs should show up, and in image builder we have some logic to surface nested errors based on error codes https://github.com/osbuild/image-builder/blob/de52a87ef277cc09d2728b49554d0da4166cc9a9/internal/v1/handler.go#L387-L419

So depending on the error codes set, the error we attach to the manifest job should surface.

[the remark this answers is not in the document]
croissanne: Also see https://github.com/osbuild/osbuild-composer/blob/6ccfc7f81884eedbf3290bb2118d0c6e31c676d5/internal/worker/server.go#L307-L320

croissanne:
> @croissanne last question for me is: should we create separate metrics? maybe in a followup PR?
> 
> ```
> func (s *Server) SetFailed(id uuid.UUID, error *clienterrors.Error) error {
> 	/* create a separate metrics?
> …
> ```

I don't think that's necessary, the jobs will be marked as failed like any other job. As long as the job isn't marked as "userinputerror" it should trigger alerts and all that. Plus at that point presumable the depsolve job duration alert would also fire.

https://github.com/osbuild/osbuild-composer/blob/6ccfc7f81884eedbf3290bb2118d0c6e31c676d5/internal/worker/clienterrors/errors.go#L73-L111

Makefile
```diff
@@ -304,6 +304,7 @@ container_composer_golangci_built.info: Makefile Containerfile_golangci_lint too
 
 .PHONY: lint
 lint: $(GOLANGCI_LINT_CACHE_DIR) container_composer_golangci_built.info
+	go fmt ./...
```
croissanne, line 307: Might need a specific go version here, sometimes between major versions fmt works differently, and people might have different default go versions than the one used by `tools/prepare-source`

.github/workflows/tests.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       run: sudo apt-get update
       # gpgme-devel is needed for container upload dependencies
     - name: Install test dependencies
-      run: sudo apt-get install -y libgpgme-dev
+      run: sudo apt-get install -y libgpgme-dev libbtrfs-dev
```
croissanne, line 70: Seems like an unrelated change? Could we split this out to a separate commit?

internal/cloudapi/v2/server.go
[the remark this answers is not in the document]
croissanne: Hmm in that case I think it should be sufficient to set the resolve jobs and manifest job as failed, and let the build job take care of itself.

And then set the relevant timeout waiting for … errors on each of the resolve job, the manifest job will fail with a dependency error, and the osbuild will also fail with a dependency error. That way the chained errors will work correctly.

.github/workflows/tests.yml
schuellerf: can be even part of #4431

Makefile
schuellerf: Nice catch, thanks!

.github/workflows/tests.yml
schuellerf: The DB tests seem to fail if this is missing :-/
I'll add it for now (although I don't understand why this is needed now)

schuellerf:
…rebased

internal/cloudapi/v2/server.go
```diff
@@ -445,22 +446,79 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
...
+				// osbuild job will fail as dependency
+				jobs := map[string]uuid.UUID{
```
croissanne, line 472: I wonder if it's ok to use a map here, since that means the loop below could be out of order. I'm not sure if there's a problem with failing the manifest job before the resolve jobs.

croissanne (commented):
1 last comment, otherwise lgtm!

internal/cloudapi/v2/server.go
```diff
@@ -445,22 +446,79 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
...
 			select {
 			case <-ctx.Done():
-				logWithId.Warning("Manifest job dependencies took longer than 5 minutes to finish, or the server is shutting down, returning to avoid dangling routines")
+				logWithId.Warning(fmt.Sprintf("Manifest job dependencies took longer than %d minutes to finish,"+
+					" or the server is shutting down, returning to avoid dangling routines", depsolveTimeout))
+
+				jobResult.JobError = clienterrors.New(clienterrors.ErrorDepsolveTimeout,
+					"Timeout while waiting for package dependency resolution",
+					"There may be a temporary issue with compute resources. "+
+						"We’re looking into it, please try again later.",
```
croissanne, line 520: Not sure about the error message here, plus the alignment looks off to me? why is the `We're` more indented than the rest.

Also we don't tend to say that we're looking into it, people should be encouraged to open tickets when necessary. I'd just keep the first line really. Also technically it's not just package dependency, it's all resolve jobs.

We could ask the user to make sure all included sources are reachable and responsive etc…

schuellerf: I removed the last line for now, but ` please try again later.` is at least some hint what to do - I'm not sure if it's good to instruct them to contact support or leave them without a help "what to do"

croissanne (approved):

PR #4400, opened 2024-10-07
tests/filesystem: increase /usr size

jrusz:
The test started failing on 8.10 GA. It seems that something changed in the system repos and the size we had originally set was not enough.

This pull request includes:

### DIFF
--- a/test/cases/filesystem.sh
+++ b/test/cases/filesystem.sh
@@ -144,7 +144,7 @@ size = "125 MiB"
 
 [[customizations.filesystem]]
 mountpoint = "/usr"
-size = 2147483648
+size = 4294967296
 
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Thank you!!

achilleas-k:
Summarising some of our chat discussions for future context:
It's strange that something changed so dramatically in a GA (presumably in a z-stream update?) that it made the size of `/usr` big enough to not fit into the partition.  It's possible the size was barely enough and the change wasn't really that big, but without looking at the old vs new to compare it's hard to say.

It's probably not worth spending time investigating this too deeply unless we see the same issue cropping up in other distro versions.

PR #4401, opened 2024-10-07
cloud/aws: add a third secure instance fallback across AZs

croissanne:
In case the on demand option failed as well, retry one more time across availability zones. This significantly increases the pool of available instances, but increases network related costs, as transferring data between AZs is not free.


croissanne:
> lgtm, seems to do what it should do. Does it make sense to introduce prometheus metrics/counter for those cases to have better overview/statistics? (probably a separate implementation/PR)

The problem is that this runs on the worker, so there's no way to get the data into prometheus unless we want to somehow forward that to composer.

PR #4403, opened 2024-10-08
deps: update osbuild/images to v0.91.0

achilleas-k:
- Update osbuild/images dependency
- Update Go version to v1.22
- Update golangci-lint version v1.61.0 (for Go v1.22)

The updated linter detects a few new issues, which have now been fixed or annotated to be ignored.


cmd/osbuild-worker/jobimpl-koji-finalize.go
```diff
@@ -305,13 +311,16 @@ func (impl *KojiFinalizeJobImpl) Run(job worker.Job) error {
 		}
 	}
 
+	if args.StartTime > math.MaxInt64 {
```
schuellerf, line 314: when would that happen? StartTime _is_ uint64?

internal/cloudapi/v2/handler.go
```diff
@@ -448,7 +453,11 @@ func (h *apiHandlers) getComposeStatusImpl(ctx echo.Context, id string) error {
 			ImageStatuses: &buildJobStatuses,
 			KojiStatus:    &KojiStatus{},
 		}
-		buildID := int(initResult.BuildID)
+
+		if initResult.BuildID < math.MaxInt {
```
schuellerf, line 457: will this ever be false?

schuellerf (changes requested):
Those range checks seem strange or is there a reason for that?
(also for the new if-branches some tests would be missing, if I would nitpick …)

cmd/osbuild-worker/jobimpl-koji-finalize.go
achilleas-k: Right, but it's comparing with max *int* 64 (not u).

internal/cloudapi/v2/handler.go
achilleas-k: My mistake.  That check should be `< 0`.

cmd/osbuild-worker/jobimpl-koji-finalize.go
schuellerf: You are totally right - I overlooked that :-/

cmd/osbuild-worker/jobimpl-koji-finalize.go
```diff
@@ -102,6 +103,11 @@ func (impl *KojiFinalizeJobImpl) Run(job worker.Job) error {
 		// Fail the Koji build if the job error is set and the necessary
 		// information to identify the job are available.
 		if kojiFinalizeJobResult.JobError != nil && initArgs != nil {
+			if initArgs.BuildID > math.MaxInt {
+				logWithId.Errorf("job BuildID %d can't be converted to 'int'", initArgs.BuildID)
```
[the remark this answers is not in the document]
schuellerf: same oversight on my end here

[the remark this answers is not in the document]
achilleas-k: That's a good question.  I'm not entirely sure either.  I (incorrectly) reflexively read this as checking if BuildID is > `MaxUint64`.  I'm not sure if we ever considered what would happen if osbuild-composer was built for (or ran on) a 32-bit system.

achilleas-k:
Looks like RHEL 9.4 is unhappy (Go version issue).

PR #4413, opened 2024-10-15
cloud/awscloud: retry CreateFleet regardless of the error code

croissanne:
The errors returned by create fleet are not entirely clear. It seems it also returns `InsufficientInstanceCapacity` in addition to `UnfulfillableCapacity`. Let's just retry three times regardless of the create fleet error, that way there's no need to chase error codes which aren't clearly defined.

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -547,14 +547,14 @@ func (a *AWS) createFleet(input *ec2.CreateFleetInput) (*ec2.CreateFleetOutput, 
 		return nil, fmt.Errorf("Unable to create spot fleet: %w", err)
 	}
 
-	if len(createFleetOutput.Errors) > 0 && *createFleetOutput.Errors[0].ErrorCode == "UnfulfillableCapacity" {
-		logrus.Warn("Received UnfulfillableCapacity from CreateFleet, retrying CreateFleet with OnDemand instance")
+	if len(createFleetOutput.Errors) > 0 {
+		logrus.Warnf("Received error %s from CreateFleet, retrying CreateFleet with OnDemand instance", *createFleetOutput.Errors[0].ErrorCode)
 		input.SpotOptions = nil
 		createFleetOutput, err = a.ec2.CreateFleet(context.Background(), input)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/cloud/awscloud/secure-instance.go
```diff
@@ -547,14 +547,14 @@ func (a *AWS) createFleet(input *ec2.CreateFleetInput) (*ec2.CreateFleetOutput,
 		return nil, fmt.Errorf("Unable to create spot fleet: %w", err)
 	}
 
-	if len(createFleetOutput.Errors) > 0 && *createFleetOutput.Errors[0].ErrorCode == "UnfulfillableCapacity" {
-		logrus.Warn("Received UnfulfillableCapacity from CreateFleet, retrying CreateFleet with OnDemand instance")
+	if len(createFleetOutput.Errors) > 0 {
+		logrus.Warn("Received error from CreateFleet, retrying CreateFleet with OnDemand instance")
```
schuellerf, line 551: …but may be log them? so we know later?

schuellerf (changes requested):

internal/cloud/awscloud/secure-instance.go
croissanne: good idea!

lzap (approved):
Yeah I wish we had prometheus on workers. This way we could have some long-term stats on how well we are doing on this front.

I am willing to work on this if there is someone who could help me to set the infra up.

internal/cloud/awscloud/secure-instance.go
```diff
@@ -547,14 +547,14 @@ func (a *AWS) createFleet(input *ec2.CreateFleetInput) (*ec2.CreateFleetOutput,
 		return nil, fmt.Errorf("Unable to create spot fleet: %w", err)
 	}
 
-	if len(createFleetOutput.Errors) > 0 && *createFleetOutput.Errors[0].ErrorCode == "UnfulfillableCapacity" {
-		logrus.Warn("Received UnfulfillableCapacity from CreateFleet, retrying CreateFleet with OnDemand instance")
+	if len(createFleetOutput.Errors) > 0 {
+		logrus.Warnf("Received error %s from CreateFleet, retrying CreateFleet with OnDemand instance", *createFleetOutput.Errors[0].ErrorCode)
```
schuellerf, line 551: Why only the first error?

schuellerf (approved):
LGTM for starters, I guess (see comment)

PR #4419, opened 2024-10-21
tests/packer: Disable Packer job in scheduled GA pipelines

tkoscieln:
Disable the Packer test in scheduled GA pipelines.


ondrejbudai (approved):
Thanks! :heart:

jrusz:
Why are we disabling it? Some commit description or link to ticket or anything would be nice...

PR #4420, opened 2024-10-21
Remove leftover secure instances in service maintenance

croissanne:
Now and then there are leftover secure instances, probably when worker instances get terminated during builds, this is possible in ASGs. 2 hours as a cutoff should be enough, since the build times out after 60 minutes, and fetching the output archive after 30 minutes, so that leaves 30 minutes for booting and connection.

### DIFF
--- a/cmd/osbuild-service-maintenance/aws.go
+++ b/cmd/osbuild-service-maintenance/aws.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


cmd/osbuild-service-maintenance/aws.go
```diff
@@ -82,5 +83,22 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
 
+	// Terminate leftover secure instances
+	reservations, err := a.DescribeInstancesByTag("parent", "i-*")
+	if err != nil {
+		return fmt.Errorf("Unable to describe instances by tag %w", err)
+	}
+	var instanceIDs []string
+	for _, res := range reservations {
+		for _, i := range res.Instances {
+			if i.LaunchTime.Before(time.Now().Add(-time.Hour * 2)) {
```
schuellerf, line 94: should we make this configurable?

internal/cloud/awscloud/maintenance_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"testing"
+
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/osbuild-composer/internal/cloud/awscloud"
+)
+
+func TestEC2RemoveSnapshotAndDeregisterImage(t *testing.T) {
```
schuellerf, line 12: a test for the 2h timing is missing, right?
just to know if the calculation is correct for sure

schuellerf (dismissed):
despite my comments, might be fine to merge

cmd/osbuild-service-maintenance/aws.go
croissanne: Well currently the timeout would have to change together with the timeouts in internal/executor, so I think it's fine.

internal/cloud/awscloud/maintenance_test.go
croissanne: Added a test to filter the reservations properly.

PR #4423, opened 2024-10-22
worker: fix OSBUILD_SOURCES_CURL_SSL_CA_CERT variable

lzap:
This looks like a copy and paste error.

The variable name was not correct:

https://github.com/osbuild/osbuild/blob/4b4b5cea950910ed9e1bc9d2fc1c618e0596b2b7/sources/org.osbuild.curl#L14

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -518,7 +518,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	if impl.RepositoryMTLSConfig != nil {
 		if impl.RepositoryMTLSConfig.CA != "" {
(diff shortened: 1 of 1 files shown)


supakeen (changes requested):
Why is this PR introducing a second environment variable that isn't used elsewhere? It's not mentioned in the title, description, or commit.

lzap:
Ah that was a leftover, I am working on this code in a different branch I was splitting git hunk. Fixed.

supakeen (approved):
Ok, thank you.

achilleas-k (approved):
Good catch.  How long has that been there?

Thanks!

achilleas-k:
Some testing would be nice for this.  Could you add one in a follow-up?

lzap:
Sure if you point me to the right place.

lzap:
How exactly I should be testing this? I just verified that this variable is properly used in `ostree` and I actually made a change that ensures it works properly:

https://github.com/osbuild/osbuild/pull/1910

This code only passes the variable into the worker, as long as there is no typo it will work. I believe the cert is enrolled at the OS level in `/etc/pki/ca-certs` thus `curl` gets it from there, that should explain why it worked until now.

PR #4425, opened 2024-10-22
List cloudapi compose jobs

bcl:
This pull request includes:


bcl:
This is just the compose job listing part of #4154 with the output using ComposeStatus instead of just the ids. I'll rewrite the other PR to just do the delete part after this is merged and the questions over there are closed.

achilleas-k (commented):
LGTM.  Sorry it took long to get around to reviewing and now there are merge conflicts :/

Any reason we can't do the dbjobqueue here too?  Would it make the PR too big?

I guess the reasoning is that this is for unifying the weldr API into the cloud API and on-prem the fsjobqueue is used...?

bcl:
> Any reason we can't do the dbjobqueue here too? Would it make the PR too big?

Yeah, it was too large and has some questions about the data structures so I'm trying to do things in smaller steps.

internal/cloudapi/v2/handler.go
```diff
@@ -314,6 +315,30 @@ func targetResultToUploadStatus(t *target.TargetResult) (*UploadStatus, error) {
...
+	if err != nil {
+		return HTTPErrorWithInternal(ErrorGettingComposeList, err)
+	}
+
+	// Gather up the details of each job
+	var stats []ComposeStatus
+	for _, jid := range jobs {
+		s, err := h.getJobIDComposeStatus(jid)
+		if err != nil {
+			// TODO log this error?
```
mvo5, line 330: Under what circumstance could this fail? Fwiw, I think we should log the errors but I do wonder under what circumstance AllRootJobIDs() and getJobIDComposeStatus(id) can get out of sync

bcl: Since it is gathering the jobs as a separate step it is possible something could come along and delete them before it is done iterating the list. Maybe not in practice, especially now before adding a `delete` route but it seems remotely possible. I'm not sure it is worth logging though, there's nothing that can be done if it vanished, other than skipping it,

bcl: This is another case of what would it do with the error? I don't want that to block listing other non-error jobs so it's better to ignore them and carry on.

bcl: I almost agree :) But I think that would end up logging useless errors. I'll leave it as-is for now and we can revisit later if it ends up that a message could help.

achilleas-k (approved):

PR #4428, opened 2024-10-23
cmd/osbuild-service-maintenance: respect dry run

croissanne:
Respect dry run when terminating leftover SIs.

---

Forgot about this when adding #4420. This isn't urgent, currently dryrun isn't enabled anywhere.


schuellerf (approved):
LGTM, although I'd prefer an output in case of `dryRun` to assure we are in dryRun and not terminating anything :-D

bcl:
I agree. It's normal to say something like 'Skipping cleanup of BIGMESS due to dry-run` so that you have a good idea of what will happen when you don't pass it.

schuellerf (approved):
nice, thanks

PR #4430, opened 2024-10-23
cloud/awscloud: rework create fleet retry logic

croissanne:
The current path sometimes launches two instances, which is problematic because the rest of the secure instance code expects exactly one instance. A security group could be attached to both instances and would block the worker from launching any more SIs as it tries to delete the old security group first, which is still held by one of the surplus SIs which didn't get terminated.

Only retry if:
- there are "UnfulfillableCapacity" or "InsufficientInstanceCapacity" error codes;
- there wasn't an instance launched anyway.

If either of these checks fail, do not try to launch another one, and just fail the job.

### DIFF
--- a/internal/cloud/awscloud/export_test.go
+++ b/internal/cloud/awscloud/export_test.go
@@ -1,3 +1,4 @@
 package awscloud
 
 var NewForTest = newForTest
+var DoCreateFleetRetry = doCreateFleetRetry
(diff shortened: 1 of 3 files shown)


schuellerf (approved):
looks ok - I'll think about a solution to create a loop, not to repeat too much

PR #4437, opened 2024-10-24
cloud/awscloud: fix retrying to create secure instances

croissanne:
Set the correct target capacity specification type, just setting the spot options to nil doesn't result in an on demand instance.


ondrejbudai (approved):
Thanks!

PR #4441, opened 2024-10-25
templates/packer: allow setting executor type in worker config

croissanne:
Currently the worker images always have to use aws.ec2, this way we can use the host executor for fedora.


supakeen (approved):
The change is fine; however do we still want to switch Fedora to a host executor?

ondrejbudai (approved):
Yeah, I would love to see the latest fixes deployed to Fedora to see whether they help, or not. :-) We can always keep this one as a backup.

PR #4435, opened 2024-10-24
test/helpers: fix false positives of `make lint`

schuellerf:
with this changes `make lint` passes without complaints.
Those are all false positives

PR #4444, opened 2024-10-28
Makefile: clean golangci-lint cache on `make clean`

schuellerf:
This is a followup of #4435.

To speedup `make lint` we use a local cache.
It seems that there is no sane way to check and just update the cache, so we'll just provide an option to wipe it.

Using the cache does reduce the time of `make lint` from 1 minute 15sec to 1 sec on my PC. So for consecutive runs, the cache still makes sense.

Thanks @mvo5 for the tests in #4435


bcl (approved):
Ack

PR #4446, opened 2024-10-28
cloud/awscloud: don't specify max spot price

croissanne:
The current spot price could be limiting the available instance pool significantly. ARM instances specifically are experiencing a lot of capacity errors.

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -419,7 +419,6 @@ func (a *AWS) createOrReplaceLT(hostInstanceID, imageID, sgID, iamProfile, keyNa
 				NetworkInterfaceCount: &ec2types.NetworkInterfaceCountRequest{
 					Min: aws.Int32(1),
 				},
-				SpotMaxPricePercentageOverLowestPrice: aws.Int32(200),
 				VCpuCount: &ec2types.VCpuCountRangeRequest{
 					Min: aws.Int32(2),
 				},


internal/cloud/awscloud/secure-instance.go
```diff
@@ -419,7 +419,6 @@ func (a *AWS) createOrReplaceLT(hostInstanceID, imageID, sgID, iamProfile, keyNa
 				NetworkInterfaceCount: &ec2types.NetworkInterfaceCountRequest{
 					Min: aws.Int32(1),
 				},
-				SpotMaxPricePercentageOverLowestPrice: aws.Int32(200),
```
schuellerf, line 422: So we expect that the algorithm below does not behave differently or "try harder" if we set a spot price?
Where we could leave the limit here and have another retry where we remove the limit?
(or we do expect it but accept the possibly raised costs?)

schuellerf (approved):
Implementation wise that's ok for me

internal/cloud/awscloud/secure-instance.go
croissanne: https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_InstanceRequirements.html the default behaviour is less strict than what we have now. We already know this because of CI really, they rarely fail to provision spot instances.

schuellerf: The question was more, if we get cheaper spot instances when we search first with a limit, then retry without instead of having no limit in the first try. I didn't find infos if this could happen.

croissanne: Right, since we're using `SpotAllocationStrategyPriceCapacityOptimized` we're already not searching for the absolute cheapest option, but rather a mix between cheap and unlikely to get culled. We could of course do retries based on this but I'm also leaning towards just letting AWS's algorithms figure things out.

schuellerf: Thanks for your thoughts & replies. Consider this resolved!

ondrejbudai (approved):
:dollar: :dollar: :dollar:

PR #4449, opened 2024-10-29
composer: simplify loading the repositories

mvo5:
Move to use `reporegistry.New(repositoryConfigs)` which will do the equivivalent of the existing code and is shorter.

With this merged we can unexport `LoadAllRepositories()` and remove `NewFromDistrosRepoConfigs` from the "images" library as this is the only place using them.

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -77,11 +77,10 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Compos
 		return nil, fmt.Errorf("failed to configure distro aliases: %v", err)
 	}
 
-	repoConfigs, err := reporegistry.LoadAllRepositories(repositoryConfigs)
+	c.repos, err = reporegistry.New(repositoryConfigs)
 	if err != nil {
 		return nil, fmt.Errorf("error loading repository definitions: %v", err)
 	}
-	c.repos = reporegistry.NewFromDistrosRepoConfigs(repoConfigs)
 
 	c.solver = dnfjson.NewBaseSolver(path.Join(c.cacheDir, "rpmmd"))
 	c.solver.SetDNFJSONPath(c.config.DNFJson)


bcl (approved):
Seems simple enough ;)

cmd/osbuild-composer/composer.go
```diff
@@ -77,11 +77,10 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Compos
 		return nil, fmt.Errorf("failed to configure distro aliases: %v", err)
 	}
 
-	repoConfigs, err := reporegistry.LoadAllRepositories(repositoryConfigs)
+	c.repos, err = reporegistry.New(repositoryConfigs)
 	if err != nil {
 		return nil, fmt.Errorf("error loading repository definitions: %v", err)
 	}
-	c.repos = reporegistry.NewFromDistrosRepoConfigs(repoConfigs)
```
thozza, line 84: This is not the same, because previously, the `c.repos` would contain a `RepoRegistry` instance, which is empty and does not contain any repositories, but now it is `nil`.

WeldrAPI requires repositories to be loaded and CloudAPI may as well in some cases.

thozza (changes requested):

cmd/osbuild-composer/composer.go
thozza: I mean, this is the case with the currently bundled `images` version, because `LoadAllRepositories` won't return `error` if there were no repositories loaded. This changed in images v0.95.0. So once PR#4448 lands, your simplification could be adjusted and work, but `c.repos` still can't be `nil` AFAICT.

mvo5: It seems this is fine now - unless I miss something now that we just call reporegistry.New() (c.f. https://github.com/osbuild/images/blob/main/pkg/reporegistry/reporegistry.go#L21) we always get either a valid RepoRegistry or an error. For the error we return and otherwise `c.repos` is always initialized.

mvo5: Well, this is all moot as the new code looks quite different :)

mvo5:
Closing for now as the new code make this no longer feasible (i.e. just calling New() like before is no longer an option)

PR #4452, opened 2024-10-30
bib/composer integration: add manifest job for osbuild processing (HMS-4843)

amirfefer:
With this change:
- bib is used to generate a manifest without initiating a full build.
- Composer’s osbuild job can now process the manifest independently.

This setup improves modularity and prepares for future enhancements in the osbuild processing pipeline.

This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-bootc-manifest.go
+++ b/cmd/osbuild-worker/jobimpl-bootc-manifest.go
@@ -1,0 +1,126 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"regexp"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
+)
+
+type BootcManifestJobImpl struct{}
+
+type ErrorInvalidImageType string
+
+func (e ErrorInvalidImageType) Error() string { return string(e) }
+
+type ErrorCommandExecution string
+
+func (e ErrorCommandExecution) Error() string { return string(e) }
+
+type ErrorInvalidArguments string
+
+func (e ErrorInvalidArguments) Error() string { return string(e) }
+
+type ErrorUnknown string
+
+func (e ErrorUnknown) Error() string { return string(e) }
+
+func setResponseError(err error, result *worker.BootcManifestJobResult) {
+	switch e := err.(type) {
+	case ErrorInvalidImageType:
+		result.JobError = clienterrors.New(
+			clienterrors.ErrorInvalidImageType,
+			"Unsupported image type",
+			e.Error(),
+		)
+	case ErrorCommandExecution:
+		result.JobError = clienterrors.New(
+			clienterrors.ErrorBootcManifestExec,
+			"Error running bootc-image-builder command",
+			e.Error(),
+		)
+	default:
+		result.JobError = clienterrors.New(
+			clienterrors.ErrorInvalidManifestArgs,
+			"Invalid bootc parameters or parameter combination",
+			err.Error(),
+		)
+	}
+}
+
+func buildManifestCommand(args *worker.BootcManifestJob) *exec.Cmd {
+	baseArgs := []string{
+		"sudo", "podman", "run",
+		"--privileged",
+		"--rm",
+		"--pull=missing",
+		"--storage-driver=vfs",
+		"--cgroup-manager=cgroupfs",
+		"--runtime=runc",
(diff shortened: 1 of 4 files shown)


cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+			"Invalid bootc parameters or parameter combination",
+			err.Error(),
+		)
+	}
+}
+
+func buildManifestCommand(args *worker.BootcManifestJob) *exec.Cmd {
+	baseArgs := []string{
+		"sudo", "podman", "run",
+		"--privileged",
```
amirfefer, line 60: Using --privileged mode with Podman is a *security concern Currently, rootless Podman fails to run bib in unprivileged mode for manifest creation due to limitations with nested containers.

Two potential approaches on my mind to address this:

Investigate fixes for the nested container environment to support rootless, non-privileged execution.
Embed the manifest creation process directly within the worker during the build, reducing the need for privileged Podman in the runtime environment.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+			clienterrors.ErrorInvalidManifestArgs,
+			"Invalid bootc parameters or parameter combination",
+			err.Error(),
+		)
+	}
+}
+
+func buildManifestCommand(args *worker.BootcManifestJob) *exec.Cmd {
+	baseArgs := []string{
+		"sudo", "podman", "run",
```
ondrejbudai, line 59: Note that the worker runs as a root, so `sudo` is not needed for rootful. If we want to run podman rootless, we need to find a way to do it.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+		"--privileged",
+		"--rm",
+		"--pull=missing",
+		"--storage-driver=vfs",
```
ondrejbudai, line 63: Why `vfs`?

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+		"--cgroup-manager=cgroupfs",
```
ondrejbudai, line 64: Is this really needed? If it is, a comment would be nice explaining it.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+		"--runtime=runc",
```
ondrejbudai, line 65: Is this really needed? If it is, a comment would be nice explaining it.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+		"--security-opt=label=type:unconfined_t",
+		"quay.io/centos-bootc/bootc-image-builder:latest",
+		"manifest",
```
ondrejbudai, line 68: Adding `--local`, and running `podman pull REF` before bib is the preferred way, see https://github.com/osbuild/bootc-image-builder/pull/423

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,126 @@
...
+		)
+	}
+}
+
+func buildManifestCommand(args *worker.BootcManifestJob) *exec.Cmd {
+	baseArgs := []string{
+		"sudo", "podman", "run",
+		"--privileged",
+		"--rm",
+		"--pull=missing",
```
ondrejbudai, line 62: `missing` is the default, I suggest dropping it.

ondrejbudai (commented):
Awesome! :) Some comments about the `podman` integration inline.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
amirfefer: I've been trying to enable bib for manifest generation in a rootless unprivileged  podman with no success so far, I asked @mvo5 for an advice here. A different solution would be to able to embed bib into the worker (just for the manifest part), or using a secure instance to run this rootful and privileged.

cmd/osbuild-worker/jobimpl-bootc-manifest.go
amirfefer: I use docker for loca devl setup, so the  worker runs on a docker instance. using crun ended with bunch of errors, so I switched it to runc with cgroupfs v2.  In our production's  k8s environment  we probably can use crun, but this needs a verification.

PR #4461, opened 2024-11-05
Bump on requeue

croissanne:
worker/server: update metrics on requeue
    
    When requeuing a job the next worker requesting the job would decrement
    pending counter, but the pending counter only ever got incremented once,
    when the job was first enqueued. Thus make sure to increment the pending
    counter when a job is requeued.


internal/worker/server.go
```diff
@@ -718,6 +718,14 @@ func (s *Server) RequeueOrFinishJob(token uuid.UUID, maxRetries uint64, result j
 		}
 	}
 
+	if requeued {
+		jobInfo, err := s.jobInfo(jobId, nil)
+		if err != nil {
+			return fmt.Errorf("error requeueing job: %w", err)
```
schuellerf, line 724: more like `error getting job details for prometheus metrics`
should we skip prometheus metrics here and try to continue, rather than fail?
although this will then probably fail some lines below…

schuellerf (approved):
LGTM, There is no explicit test for the metrics right?
Not sure if this is doable with a sane effort.
And might be a separate PR?

internal/worker/server.go
croissanne: I think it's very unlikely that this call will fail if the job exists.

croissanne:
It's definitely possible to add unit tests, I'll do that.

schuellerf:
Sidenote - I think the `running_jobs` metric also has some problems (causing it to go below 0) so we could see this in the tests if you can implement them 🥳

schuellerf (approved):
LGTM, I'll check additionally if adding more tests to `RunningJobs` metric will bring up problems in some corner cases (as I think there should be some)

PR #4462, opened 2024-11-05
Introduce grafana number of pending jobs

schuellerf:
Adds the charts "Pending Jobs Max" and "Pending Jobs"
both "by type".
To see how many jobs are pending and where we might need more workers.

Note that the second commit to enable crosshair is just convenience and could be removed from this PR or split off if needed.

### DIFF
--- a/templates/dashboards/grafana-dashboard-image-builder-worker-general.configmap.yml
+++ b/templates/dashboards/grafana-dashboard-image-builder-worker-general.configmap.yml
@@ -33,7 +33,7 @@ data:
       },
       "editable": true,
       "fiscalYearStartMonth": 0,
-      "graphTooltip": 0,
+      "graphTooltip": 1,
       "links": [],
       "liveNow": false,
       "panels": [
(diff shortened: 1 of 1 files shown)


templates/dashboards/grafana-dashboard-image-builder-worker-general.configmap.yml
```diff
@@ -1564,6 +1564,179 @@ data:
...
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${datasource}"
+              },
+              "disableTextWrap": false,
+              "editorMode": "code",
+              "exemplar": false,
+              "expr": "sum(image_builder_worker_pending_jobs) by (type)",
+              "fullMetaSearch": false,
```
croissanne, line 1629: Hmm do we need to do a max instead of a sum here?

schuellerf: acutally no - it's a `"type": "stat"` with `"options": { "reduceOptions": { "calcs": [ "max" ] } }`

schuellerf: So taking the same query as the graph but just a "Stat"-Visualization

croissanne (approved):
thanks!

PR #4482, opened 2024-11-20
worker: report cashes directly to logrus

mvo5:
This is a bit of an RFC commit, I noticed that when we discussed a crash from the worker we lookged at individual message from syslog/journald for the stacktrace deatils. I was wondering if having a more direct crash report would be more useful? We can of course also add more logrus features to flag those with tags like "crash" or something (I did not do that in this PR, I don't know much about the operational side, sorry).

Feel free to close if we already have similar means to get the comprehensive stacktrace, I looked into how we deploy this but it seems we run the worker directly via a systemd unit so this might be still useful.

### DIFF
--- a/cmd/osbuild-worker/export_test.go
+++ b/cmd/osbuild-worker/export_test.go
@@ -3,4 +3,13 @@ package main
 var (
 	WorkerClientErrorFrom         = workerClientErrorFrom
 	MakeJobErrorFromOsbuildOutput = makeJobErrorFromOsbuildOutput
+	Main                          = main
 )
+
+func MockRun(new func()) (restore func()) {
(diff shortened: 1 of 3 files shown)


croissanne (approved):
Neat!

croissanne (approved):

lzap (approved):
This is indeed useful, we do this as well in provisioning. The only drawback is that the message is logged without a context as it cannot be passed into recover block effectively. So it will be missing job_id for example, unless it is carried in the error itself. We could build an `ErrorWithContext` wapper type that would have the context which could be then extracted, but that is definitely something for the future as context is not being passed in the app yet.

PR #4484, opened 2024-11-20
cloudapi: git sha version api and journal for workers

lzap:
Adds a runtime information similarily as in image builder and a logrus hook that is used to put git sha to every single log message. This way we can tell which version was log created with since composer and workers sometimes are running different versions.

This data is then leveraged in a new simple /version cloud API call that returns the data similarily to IB. My plan is to call this from IB too so IB reports both its and composer version info.

Finally, I noticed that workers create pretty unreadable logs, it is a string-encoded JSON:

```
{"message":"{\"channel\":\"staging\",\"file\":\"/opt/app-root/src/internal/common/slogger/logrus.go:29\",\"func\":\"github.com/osbuild/osbuild-composer/internal/common/slogger.(*simpleLogrus).log\",\"job_dependencies\":\"[8402b9c0-ac94-4952-8609-1ad4e575a28a e30715cd-0a59-4993-a530-fdb2189ad2bb]\",\"job_id\":\"be17e941-d8f5-4de2-9eb3-e7a1b3ee41d8\",\"job_type\":\"manifest-id-only\",\"level\":\"info\",\"msg\":\"Dequeued job\",\"time\":\"2024-11-20T12:10:37Z\"}\n","ident":"osbuild-composer","host":"composer-5dcd8bfdf9-bhrv6"}
```

This patch builds on top of native journald capability which I added last year and is used in on-prem mode. Instead using standard output/err worker now detects if there is a systemd-journald running and uses its native API to send logs in structured form.

This means all logs are native now, it can be inspected with journalctl and what is more important they should get into Kibana/Splunk as regular JSON improving our searching capabilities:

```
journalctl -o verbose
...
    BUILD_TIME=2024-11-20T11:24:49Z
    BUILD_COMMIT=ff2660
    PRIORITY=3
    MESSAGE=Requesting job failed: Post "https://0:8881/api/worker/v1/jobs": dial tcp 0.0.0.0:8881: connect: connection refused
    _PID=254883
    _COMM=__debug_bin3968
    _EXE=/home/lzap/osbuild-composer/cmd/osbuild-worker/__debug_bin396826732
    _CMDLINE=/home/lzap/osbuild-composer/cmd/osbuild-worker/__debug_bin396826732 0:8881
    _SOURCE_REALTIME_TIMESTAMP=1732108805201853
Wed 2024-11-20 14:20:05.201871 CET [s=83cfd176837d4a15b19c452be25c8593;i=1ed8;b=3d1a7f8752454afda1a56b25debf50d9;m=664c2be936;t=62758020be16c;x=299933d4e616281a]
    _SELINUX_CONTEXT=kernel
    _BOOT_ID=3d1a7f8752454afda1a56b25debf50d9
    _MACHINE_ID=90304ac7a1274d6ca3abfee98b7975c1
    _HOSTNAME=dev.home.lan
    _RUNTIME_SCOPE=system
...
```

I know this is sort of two features in one PR but these are tiny changes, the API endpoint is super small and I wrote a test.


cmd/osbuild-worker/main.go
```diff
@@ -188,6 +197,39 @@ func main() {
...
+			logrus.SetFormatter(&logrus.TextFormatter{})
+		}
+	case "text":
+		logrus.SetFormatter(&logrus.TextFormatter{})
+	case "json":
+		logrus.SetFormatter(&logrus.JSONFormatter{})
+	default:
+		logrus.Infof("Failed to set logging format from config, '%s' is not a valid option", config.Logging.Format)
+	}
+
```
lzap, line 232: FYI this was copy-pasted from `composer/main.go` we already use it there.

internal/cloudapi/v2/handler.go
```diff
@@ -45,6 +45,19 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
...
+		return HTTPError(ErrorFailedToLoadOpenAPISpec)
+	}
+	version := Version{
+		Version:     spec.Info.Version,
+		BuildCommit: common.ToPtr(common.BuildCommit),
+		BuildTime:   common.ToPtr(common.BuildTime),
+	}
+	return ctx.JSON(http.StatusOK, version)
+}
+
```
lzap, line 60: We have the very same code in image-builder too, this is a copy-paste literally.

cmd/osbuild-worker/main.go
ezr-ondrej: Would it make sense to have some common logging setup function to actually benefit from having both composer and worker in the same repo?

lzap: I would say no, these are completely different binaries, separated by design. Little bit of copying is sometimes better than a dependency.

internal/common/runtime.go
```diff
@@ -0,0 +1,36 @@
+package common
+
+import "runtime/debug"
+
+var (
+	// Git SHA commit (first 4 characters)
```
mvo5, line 6: This comment (4 chars) seems to not match the implementaion AFACIT

internal/common/runtime.go
```diff
@@ -0,0 +1,36 @@
...
+
+func init() {
+	bi, ok := debug.ReadBuildInfo()
+
+	if !ok {
+		BuildTime = "N/A"
+		BuildCommit = "HEAD"
+	}
+
+	BuildGoVersion = bi.GoVersion
```
mvo5, line 24: Is `bi` actually defined (i.e. non-nil) if we get a `!ok` before (if not we would crash here)? I wonder if it wouldn't be cleaner to define the defaults directly in `var BuildCommit = "N/A"` and just return on `if !ok { return }`(?)

mvo5 (commented):
Quick drive-by wondering inline

schuellerf (commented):
No major comment additionally to the already mentioned ones.
In image builder we changed the way some paths are handled so they don't need authentication, which might be neat for `/version` but I'd approve as this might be easier in a followup PR

internal/cloudapi/v2/server.go
```diff
@@ -130,17 +130,33 @@ func (s *Server) Handler(path string) http.Handler {
...
-		prometheus.HTTPDurationMiddleware(prometheus.ComposerSubsystem),
+	mws = append(mws, prometheus.HTTPDurationMiddleware(prometheus.ComposerSubsystem),
+		prometheus.MetricsMiddleware, s.ValidateRequest)
+	mwsna = append(mwsna, prometheus.HTTPDurationMiddleware(prometheus.ComposerSubsystem),
 		prometheus.MetricsMiddleware, s.ValidateRequest)
+
 	RegisterHandlers(e.Group(path, mws...), &handler)
 
+	// no auth endpoints
+	e.GET("/status", func(c echo.Context) error {
```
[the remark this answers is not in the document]
lzap: I know right, but this is what is in the image-builder :-) Do you want me to break this pattern? I am fine with that, can also create both maybe?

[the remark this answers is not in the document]
schuellerf: I think `/version` exists as it's in the `openapi.v2.yml` so this might already result in both existing (which is good)
Rather somehow assure `/version` not to be authenticated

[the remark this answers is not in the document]
lzap: Ok added `/version` with non-auth middleware and also adding status/ready as fixed paths that simply return 200 without ANY middleware because we do not this to be logged or have metrics as k8s calls this every now and then.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -17,6 +17,21 @@ servers:
   description: current domain
 
 paths:
+  /version:
```
schuellerf, line 20: I think this is good, but when thinking about it … that's the version of `osbuild-composer` not the worker, right? (in contrast to the PR title and commit subject :-/ )

schuellerf: The API is "push" in this case - so the worker would have to "push" it's version, maybe?
https://github.com/osbuild/osbuild-composer/blob/64ff0e3dadd4b1e654d095518485591a0c0056d6/internal/worker/api/openapi.yml#L349

cmd/osbuild-worker/main.go
```diff
@@ -189,6 +198,39 @@ var run = func() {
 		logrus.Fatalf("Could not load config file '%s': %v", configFile, err)
 	}
 
+	logrus.SetReportCaller(true)
+	logrus.AddHook(&common.BuildHook{})
+	logrus.SetOutput(os.Stdout)
+	logLevel, err := logrus.ParseLevel(config.Logging.Level)
+	if err == nil {
```
mvo5, line 205: (nitpick/drive-by) maybe switch this around to `if err != nil {` first as this the most common way in go to deal with errors? I know it's minor but when I read code that does not follow this pattern I immediately stop in my flow as this signal "something here is special" (but here it seems it is not (but feel free to ignore of course)

lzap: Amended.

internal/cloudapi/v2/openapi.v2.yml
schuellerf: So this could be either split in two commits or at least the description (PR and commit) should denote that this is touching worker _and_ composer?

lzap:
I have split it into multiple commits, rebased.

schuellerf (dismissed):
lgtm, thanks!

lzap:
The only test failing is edge-farm, rebasing.

lzap:
@croissanne @ezr-ondrej @mvo5 please re-review tests are good I want to get this one merged to increase readability of logs from composer and continue further improving logs with standardized correlation thanks!

internal/common/runtime.go
```diff
@@ -0,0 +1,35 @@
...
+	BuildCommit = "HEAD"
+
+	if bi, ok := debug.ReadBuildInfo(); ok {
+		BuildGoVersion = bi.GoVersion
+
+		for _, bs := range bi.Settings {
+			switch bs.Key {
+			case "vcs.revision":
+				if len(bs.Value) > 6 {
+					BuildCommit = bs.Value[0:6]
```
mvo5, line 27: (nitpick/drive-by) I wonder if we should use a slightly longer string here for the rev -  for the osbuild-composer repo with many commits `git log --oneline` gives me 9 chars for the git commit, this is dynamically calculated  but usually the minimum is 7. So maybe just follow git and use 9 to avoid possible collisions?

thozza: I'd suggest looking at https://github.com/osbuild/osbuild-composer/blob/8c99478a691a87e64f66003756170c4dda71da98/internal/common/constants.go#L30 and ideally unify how we do things in various places.

I'm wondering if the `vcs` values are set with the binary is built from SRPM using `rombuild`?

internal/cloudapi/v2/handler.go
```diff
@@ -45,6 +45,19 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
 	return nil
 }
 
+func (h *apiHandlers) GetVersion(ctx echo.Context) error {
+	spec, err := GetSwagger()
+	if err != nil {
+		return HTTPError(ErrorFailedToLoadOpenAPISpec)
+	}
+	version := Version{
+		Version:     spec.Info.Version,
```
thozza, line 54: I'm sorry, but the API version is not what I expect under `Version` in the `/version` endpoint. Instead, I'd expect the release version. The API version is already included in the `/openapi` endpoint. Could you tell me what your reason is for picking the API version?

internal/common/runtime.go
lzap: Increasing this to 7. I am about to start integrating the newly created shared code published as https://github.com/osbuild/logging and I have just added the very same functionality there.

@thozza good point, it will not be set, the code needs to be built with -X linker flag, I actually implemented a custom build variable that can be overriden and provided an example in the upcoming library which I plan to integrate into both ib and composer:

https://github.com/osbuild/logging/commit/e2079252554bb0136a85c7ce4a8d70692ee52e9e

internal/cloudapi/v2/handler.go
lzap: I can rename this to APIVersion, is there any version information provided currently I could return in here?

ezr-ondrej: I don't really see anywhere where we could read the release version from code. Reading it from https://github.com/osbuild/osbuild-composer/blob/8b0a1d1714f66a72869d3b556ce756794aa19797/osbuild-composer.spec#L15 might be clunky at best. Is there a better way to read it from the system? :thinking: 

I think we could just omit the release for now and rename this field to APIVersion.

internal/common/build_hook.go
```diff
@@ -0,0 +1,26 @@
...
+		logrus.WarnLevel,
+		logrus.ErrorLevel,
+		logrus.FatalLevel,
+		logrus.PanicLevel,
+	}
+}
+
+func (h *BuildHook) Fire(e *logrus.Entry) error {
+	e.Data["build_commit"] = BuildCommit
+	e.Data["build_time"] = BuildTime
```
ezr-ondrej, line 23: do we need build_time for every log line? It sounds a little excessive :thinking:
Wouldn't commit be enough? I'm just bit worried about overpolluting the logs.

internal/cloudapi/v2/handler.go
thozza: No, we should not be reading it from the SPEC file or from the system :wink: 

It should be (and in most of the cases is) embedded in the binary itself, see https://github.com/osbuild/osbuild-composer/pull/4484#discussion_r1900581533

> I think we could just omit the release for now and rename this field to APIVersion.

In that case, I'd suggest entirely omitting the API version. What is the use case for listing it there?

lzap:
The work I am doing here and at other places distilled to a common logging library which I plan to integrate into composer. So closing as it should solve all the problems I have with this logging. I will start integrating it with image-builder first tho:

https://github.com/osbuild/logging

Filed a different PR with the linter improvement: #4547

PR #4485, opened 2024-11-20
awscloud/secure-instance: retry on error in terminated waiter

schuellerf:
Terminated waiter sometimes responded
with `waiter state transitioned to Failure`
where we want to retry waiting for the termination

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -228,6 +228,9 @@ func (a *AWS) TerminateSecureInstance(si *SecureInstance) error {
 }
 
 func (a *AWS) terminatePreviousSI(hostInstanceID string) (string, error) {
+	var retryCount int
+	const maxRetries = 3
+
 	descrInstancesOutput, err := a.ec2.DescribeInstances(
 		context.Background(),
 		&ec2.DescribeInstancesInput{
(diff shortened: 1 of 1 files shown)


croissanne (approved):

internal/cloud/awscloud/secure-instance.go
```diff
@@ -228,6 +228,9 @@ func (a *AWS) TerminateSecureInstance(si *SecureInstance) error {
 }
 
 func (a *AWS) terminatePreviousSI(hostInstanceID string) (string, error) {
+	var retryCount int
+	const maxRetries = 3
```
croissanne, line 232: This number will need to be way higher actually, like idk how long each try takes but I think we should wait more than 3 times. Like up to 10 minutes or so.

croissanne (changes requested):

internal/cloud/awscloud/secure-instance.go
croissanne: I'm not even sure if we should do it or if we should let the waiter do it?

looking at https://github.com/aws/aws-sdk-go-v2/blob/service/ec2/v1.191.0/service/ec2/api_op_DescribeInstances.go#L1383 ; the thing that calls the retryable takes care of everything there.

schuellerf: we can also increase the `MinDelay`
so `MinDelay = 60 * time.Second` and `maxRetries = 10` ?

croissanne: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html I think we should basically do the same except allow "stopping", since I think we might ebs-backed instances sometimes? Not sure. This honestly seems like a bug in the sdk.

croissanne: Hmm, let's maybe try without the custom retryfunction but just increase the mindelay and maxretries?

internal/cloud/awscloud/secure-instance.go
```diff
@@ -260,13 +263,46 @@ func (a *AWS) terminatePreviousSI(hostInstanceID string) (string, error) {
...
+			return false, err
+		}
+		logrus.Warnf("retry waiting for termination %d/%d", retryCount, maxRetries)
+		return true, nil
+
+	}
+
+	TerminatedWaiterOptions := func(o *ec2.InstanceTerminatedWaiterOptions) {
+		o.Retryable = retryFunction
+		o.LogWaitAttempts = true
```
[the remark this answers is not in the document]
schuellerf: probably 🤷‍♂️
removing those is easy …

internal/cloud/awscloud/secure-instance.go
schuellerf: as in theory it should wait for 1 hour (as given by the `maxWaitDur` of `.Wait()`) we probably need to retry on "Failure" (what ever that is)

croissanne (approved):
Let's try.

For future reference, can you format the commit messages a bit better? Like punctuation in the first. Otherwise lgtm!

PR #4489, opened 2024-11-22
awscloud/secure-instance: fix cleaning up fleets with creation errors

croissanne:
By surfacing the output even in case of an error, the fleet ID and
    instance ID can be extracted if present. Thus the instance can be
    terminated before its dependencies are deleted.

---

We're seeing some behaviour where create fleet is not retried and subsequently the SI cleanup fails due to the security group already being tied to an existing instance. There is no error that an instance was launched anyway.


schuellerf (dismissed):
just minor things

schuellerf (approved):
nice, thanks!

PR #4490, opened 2024-11-22
Schutzfile: update osbuild commit (COMPOSER-2406)

croissanne:
Newer osbuild uses dnf 4 on Fedora 41, as we've seen Fedora 41 in combination with dnf 5 (and SBOMs) failing.


thozza (approved):
:+1:

thozza:
Funnily enough, our Fedora workers seem to be still using F40 :facepalm:

PR #4495, opened 2024-11-26
cmd: extra env logging for osbuild worker

lzap:
It looks like extra env is not passed correctly for ostree builds, this should verify it. There are no secrets, certs and keys are passed via filenames, so this should be safe.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -518,6 +518,9 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	// Both curl and ostree input share the same MTLS config
 	if impl.RepositoryMTLSConfig != nil {
+		// Setting a CA cert with hosted Pulp with break the build since Pulp redirects HTTPS requests to AWS S3 which has
+		// a different CA which is part of OS cert bundle. Both curl and ostree commands only support either explicit CA file
+		// or OS cert bundle, but not both. To verify hosted Pulp CA, enroll its CA into the OS cert bundle instead.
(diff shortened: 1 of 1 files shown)


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -560,6 +563,7 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		exportPaths = append(exportPaths, path.Join(jobTarget.OsbuildArtifact.ExportName, jobTarget.OsbuildArtifact.ExportFilename))
 	}
 
+	logWithId.Infof("Extra env: %s", strings.Join(extraEnv, " "))
```
mvo5, line 566: Maybe just `logWithId.Infof("Extra env: %q", extraEnv)` as this will give us automatic quoting if things like spaces are part of the env?

lzap: It will not: https://go.dev/play/p/2gehrpitrhh

mvo5: Hey @lzap - thanks for the golang play link - I'm slightly confused though, the link uses `%s` but I suggested `%q` which will add quote AFAICT (cf. https://go.dev/play/p/4tZBweD8jFC) - or am I misunderstanding/missing something maybe?

lzap: Oh I have missed that difference completely, sure, this is indeed more readable. Amended the change.

mvo5 (approved):
Thank you

lzap:
Schulz is red, not relevant. Merging.

PR #4496, opened 2024-11-26
cloud/awscloud: exclude really old instance types

croissanne:
RHEL 9.6 (nightly) builds fail on stage with "Fatal glibc error: CPU does not support x86-64-v3", this is most likely due to very old instance types not supporting a specific instruction set.


ondrejbudai:
This is weird, [RHEL 9 requires only x86_64-v2](CPU does not support x86-64-v3), it might be beneficial to investigate this and report this.

However, RHEL 10 requires x86_64-v3, so this PR absolutely makes sense: https://access.redhat.com/solutions/7066628

x86_64-v3 is Haswell and newer, see https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

Looking at the [AWS specs for old generations](https://docs.aws.amazon.com/ec2/latest/instancetypes/pg.html#pg_hardware), I think we should also remove: c3*, i2*, m2*, m3* and r3* (all Ivy Bridge or older)

ondrejbudai (approved):
Thanks!

PR #4504, opened 2024-12-02
cloud/awscloud: give secure instances a name

croissanne:
That way you can just enter the parent instance id into the search bar and get both the worker and its executor.


ondrejbudai (approved):
Awesome, thanks! ❤️

lzap (approved):
Nice, thanks.

PR #4507, opened 2024-12-02
cloud/awscloud: use any instance create fleet returns

croissanne:
Even in case of errors, as long as create fleet returns an instance,
    attempt to use it.
    
 In some cases AWS returns `InsufficientInstanceCapacity` but still
    creates an instance:

```
    msg="Won't retry CreateFleet with OnDemand instance, retry: false, errors: InsufficientInstanceCapacity: There is no Spot capacity available that matches your request.; Already launched instance ([i-...]), aborting create fleet"
    msg="doCreateFleetRetry: returning retry: false, msg: [InsufficientInstanceCapacity: There is no Spot capacity available that matches your request. Already launched instance ([i-...]), aborting create fleet]"
    msg="doCreateFleetRetry: cancelling retry, instance already exists: [i-...]"
    msg="doCreateFleetRetry: setting retry to true"
    msg="Checking to retry fleet create on error InsufficientInstanceCapacity (msg: There is no Spot capacity available that matches your request.)"
```

### DIFF
--- a/internal/cloud/awscloud/secure-instance.go
+++ b/internal/cloud/awscloud/secure-instance.go
@@ -166,6 +166,7 @@ func (a *AWS) RunSecureInstance(iamProfile, keyName, cloudWatchGroup, hostname s
 		},
 		Type: ec2types.FleetTypeInstant,
 	})
+
 	// retrieve any instance information even if there's an error, that way the instance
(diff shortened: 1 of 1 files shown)


schuellerf (approved):
LGTM, let's check…

PR #4513, opened 2024-12-03
Clean up orphaned instances and security groups (HMS-3632)

schuellerf:
This pull request includes:

Implement removal of orphaned items as described in https://issues.redhat.com/browse/HMS-3632
removal of launch templates is not yet implemented (that's why it's a draft for now)

### DIFF
--- a/cmd/osbuild-service-maintenance/aws.go
+++ b/cmd/osbuild-service-maintenance/aws.go
@@ -122,6 +122,14 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
 		}
 	}
 
+	errLaunchTemplates := searchLTAndCleanup(ctx, a, dryRun)
+	if errLaunchTemplates != nil {
+		logrus.Errorf("Error in cleaning up launch templates: %v", errLaunchTemplates)
+		if err != nil {
+			err = fmt.Errorf("Multiple errors while processing AWSCleanup: %w and %w.", err, errLaunchTemplates)
+		}
+	}
+
 	return err
 }
 
@@ -268,3 +276,39 @@ func allTerminated(reservations []ec2types.Reservation) bool {
 	return true
 }
 
+func searchLTAndCleanup(ctx context.Context, a *awscloud.AWS, dryRun bool) error {
+	launchTemplates, err := a.DescribeLaunchTemplatesByPrefix(ctx, "launch-template-for-i-")
+	if err != nil {
+		return err
+	}
+
+	for _, lt := range launchTemplates {
+		if lt.LaunchTemplateName == nil || lt.LaunchTemplateId == nil {
+			logrus.Errorf(
+				"Launch template needs to have a LaunchTemplateName (%v) and a LaunchTemplateId (%v).",
+				lt.LaunchTemplateName,
+				lt.LaunchTemplateId)
+			continue
+		}
+
+		reservations, err := a.DescribeInstancesByLaunchTemplateID(*lt.LaunchTemplateId)
+		if err != nil {
+			logrus.Errorf("Failed to describe launch template %s: %v", *lt.LaunchTemplateId, err)
+			continue
+		}
+
+		if allTerminated(reservations) {
+			logrus.Infof("Deleting launch template: %s (%s)\n", *lt.LaunchTemplateName, *lt.LaunchTemplateId)
+			if !dryRun {
+				err := a.DeleteLaunchTemplateById(ctx, lt.LaunchTemplateId)
+
+				if err != nil {
+					logrus.Errorf("Failed to delete launch template %s: %v", *lt.LaunchTemplateId, err)
+				}
+			}
+		} else {
+			fmt.Printf("Launch template %s has non terminated instances associated with it.\n", *lt.LaunchTemplateId)
+		}
+	}
+	return nil
+}
--- a/cmd/osbuild-service-maintenance/aws.go
+++ b/cmd/osbuild-service-maintenance/aws.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sync"
@@ -104,33 +105,29 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
 		wg.Wait()
 	}
 
-	// using err to collect both errors as we want to
-	// continue execution if one cleanup fails
-	err = nil
-	errSecureInstances := terminateOrphanedSecureInstances(a, dryRun)
-	// keep going with other cleanup even on error
-	if errSecureInstances != nil {
-		logrus.Errorf("Error in terminating secure instances: %v, continuing other cleanup.", errSecureInstances)
-		err = errSecureInstances
+	// using `errs` to collect all errors as we want to
+	// continue execution if only one cleanup fails
+	var errs []error
+
+	err = terminateOrphanedSecureInstances(a, dryRun)
+	if err != nil {
+		logrus.Errorf("Error in terminating secure instances: %v, continuing other cleanup.", err)
+		errs = append(errs, err)
 	}
 
-	errSecurityGroups := searchSGAndCleanup(ctx, a, dryRun)
-	if errSecurityGroups != nil {
-		logrus.Errorf("Error in cleaning up security groups: %v", errSecurityGroups)
-		if err != nil {
-			err = fmt.Errorf("Multiple errors while processing AWSCleanup: %w and %w.", err, errSecurityGroups)
-		}
+	err = searchSGAndCleanup(ctx, a, dryRun)
+	if err != nil {
+		logrus.Errorf("Error in cleaning up security groups: %v", err)
+		errs = append(errs, err)
 	}
 
-	errLaunchTemplates := searchLTAndCleanup(ctx, a, dryRun)
-	if errLaunchTemplates != nil {
-		logrus.Errorf("Error in cleaning up launch templates: %v", errLaunchTemplates)
-		if err != nil {
-			err = fmt.Errorf("Multiple errors while processing AWSCleanup: %w and %w.", err, errLaunchTemplates)
-		}
+	err = searchLTAndCleanup(ctx, a, dryRun)
+	if err != nil {
+		logrus.Errorf("Error in cleaning up launch templates: %v", err)
+		errs = append(errs, err)
 	}
 
-	return err
+	return errors.Join(errs...)
 }
 
 func terminateOrphanedSecureInstances(a *awscloud.AWS, dryRun bool) error {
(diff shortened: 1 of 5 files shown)


cmd/osbuild-service-maintenance/aws.go
```diff
@@ -84,13 +99,58 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+	if errSecureInstances != nil {
+		logrus.Errorf("Error in terminating secure instances: %v, continuing other cleanup.", errSecureInstances)
+		err = errSecureInstances
+	}
+
+	errSecurityGroups := searchSGAndCleanup(ctx, a, dryRun)
+	if errSecurityGroups != nil {
+		logrus.Errorf("Error in cleaning up security groups: %v", errSecurityGroups)
+		if err != nil {
+			err = fmt.Errorf("Multiple errors while processing AWSCleanup: %w and %w.", err, errSecurityGroups)
```
schuellerf, line 116: Is this a valid pattern, to try to continue cleanup and collect the errors?

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -103,14 +163,164 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+	}
+	if len(parent) != 1 {
+		logrus.Errorf("Instance %s has %d parents. That should never happen, not changing anything here.", childId, len(parent))
+		return true
+	}
+	if len(parent[0].Instances) == 0 {
+		logrus.Infof("Instance %s has no parent instance, removing it", childId)
+		return false
+	}
+	if len(parent[0].Instances) != 1 {
```
schuellerf, line 220: not sure if all those checks are necessary?

internal/cloud/awscloud/maintenance.go
```diff
@@ -84,6 +85,31 @@ func (a *AWS) DescribeInstancesByTag(tagKey, tagValue string) ([]ec2types.Reserv
...
+func (a *AWS) DescribeInstancesByTag(tagKey, tagValue string) ([]ec2types.Reservation, error) {
+	return a.describeInstancesByKeyValue(fmt.Sprintf("tag:%s", tagKey), tagValue)
+}
+
+func (a *AWS) DescribeInstancesBySecurityGroupID(securityGroupID string) ([]ec2types.Reservation, error) {
+	return a.describeInstancesByKeyValue("instance.group-id", securityGroupID)
+}
+
+func (a *AWS) DescribeInstancesByLaunchTemplateID(launchTemplateID string) ([]ec2types.Reservation, error) {
+	return a.describeInstancesByKeyValue("tag:aws:ec2launchtemplate:id", launchTemplateID)
```
schuellerf, line 97: This looks weird - I'll check if there is another option to get the instances (or do you know by heart how it works @croissanne ?)

croissanne: Eh, looks ok to me. https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/ec2/describe-instances.html#options are the filters, and i don't see launchtemplate

cmd/osbuild-service-maintenance/aws.go
croissanne: I tihnk it's ok

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -84,33 +99,230 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+				inUse = true
+				break
+			}
+		}
+	}
+	return inUse
+}
+
+func searchLTAndCleanup(ctx context.Context, a *awscloud.AWS, dryRun bool) error {
+
```
croissanne, line 291: rm newline here

croissanne (commented):
In general we don't put the jira ticket in the commit message itself, just the PR title.

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -112,17 +128,73 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+				if !slices.Contains(instanceIDs, *i.InstanceId) {
+					instanceIDs = append(instanceIDs, *i.InstanceId)
+				}
 			}
 		}
 	}
 	return instanceIDs
 }
+
+func getChildAndParentInstances(reservations []ec2types.Reservation) []struct {
```
croissanne, line 182: I think defining this struct up to instead of an anonymous struct would make it more readable.

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -97,6 +99,28 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+	// continue execution if one cleanup fails
+	err = nil
+	errSecureInstances := terminateOrphanedSecureInstances(a, dryRun)
+	// keep going with other cleanup even on error
+	if errSecureInstances != nil {
+		logrus.Errorf("Error in terminating secure instances: %v, continuing other cleanup.", errSecureInstances)
+		err = errSecureInstances
+	}
+
+	errSecurityGroups := searchSGAndCleanup(ctx, a, dryRun)
```
croissanne, line 112: I'm not sure if we should share the same context here? It probably doesn't matter that much but I don't get the rationale for doing so here.

cmd/osbuild-service-maintenance/aws.go
croissanne: I think generally error wrapping is used to wrap deeper errors, not really make a list of them. So maybe we should just add the errors to a slice and then join them together in the end? I think it might make it more readable.

cmd/osbuild-service-maintenance/aws.go
schuellerf: Ups - thanks!

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -192,9 +215,69 @@ func checkValidParent(childId string, parent []ec2types.Reservation) bool {
...
+		} else {
+			logrus.Debugf("Security group %s has non terminated instances associated with it.", *sg.GroupId)
+		}
+	}
+	return nil
+}
+
+// checkIfInUse returns true if any instance of the reservations is not terminated
+// then it's considered "in use"
+func checkIfInUse(reservations []ec2types.Reservation) bool {
```
croissanne, line 276: can we just call this "isInstanceRunning"? I thinkit makes it a bit clearer.

croissanne (changes requested):
thanks, looks good overall

Could you reword this commit message though?
```
Usually the parent does not exist when also the secure instance is
older than 2h, but this an extra safeguard that secure instances
without a valid parent instance does not make sense to keep.
```

And then the jira tickets tend to just go in the PR descriptions.

cmd/osbuild-service-maintenance/aws.go
croissanne: Like the context isn't passed from up above, but created in this function.

schuellerf:
> And then the jira tickets tend to just go in the PR descriptions.

I think they should be part of the commit message
https://osbuild.org/docs/developer-guide/general/workflow
but I'll move it to the "git commit subject" that's more readable for sure

cmd/osbuild-service-maintenance/aws.go
schuellerf: I didn't want to change all of the functions from above in this PR, but prepare as much as possible.
As far as I understood that's the idea of the context, that it should always be traversed from top to bottom to be able to hand over cancel, timeouts and other properties…

cmd/osbuild-service-maintenance/aws.go
schuellerf: It's more than running. Also pending, shutting-down.… but you are right, it sounds odd…

schuellerf: or invert and `allTerminated`

cmd/osbuild-service-maintenance/aws.go
schuellerf: please check

cmd/osbuild-service-maintenance/aws.go
schuellerf: please check

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -84,33 +126,212 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
 		wg.Wait()
 	}
 
+	// using `errors` to collect all errors as we want to
+	// continue execution if only one cleanup fails
+	var errors MultiError
```
mvo5, line 131: Quick drive-by, there is `errors.Join()` now, the more common way to write this (IME) is:
```go
var errs []error
...
errors = append(errs, err)
...
return errors.Join(errs)
```

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -13,10 +15,50 @@ import (
...
 )
 
-func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey string, cutoff time.Time) error {
-	a, err := awscloud.New("us-east-1", accessKeyID, accessKey, "")
+type ChildToParentAssociation struct {
+	Child  string
+	Parent string
+}
+
+type MultiError struct {
```
mvo5, line 23: I think we want to use `errors.Join()` instead which gives us the stdlib "joinErrors" struct which does the same expect it also implements Unwrap (so that errors.{Is,As} keeps working)

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -84,33 +126,212 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
+
+// allTerminated returns true if any instance of the reservations is not terminated
+// then it's considered "in use"
+func allTerminated(reservations []ec2types.Reservation) bool {
+	allAreTerminated := true
+
+	for _, reservation := range reservations {
+		for _, instance := range reservation.Instances {
+			if instance.State != nil && (instance.State.Name != ec2types.InstanceStateNameTerminated) {
+				allAreTerminated = false
```
mvo5, line 294: This could just `return false` here, this would avoid the need for the `allAreTerminated` var (and the last return of the function would just be `return true`)

mvo5 (commented):
Thanks for working on this - quick drive-by comments/suggestions (purely on the tactical level :)

cmd/osbuild-service-maintenance/aws.go
croissanne: ack

croissanne (commented):
other than  @mvo5's comments lgtm :)

cmd/osbuild-service-maintenance/aws.go
schuellerf: Thanks! …refactoring mess I left there :-/

croissanne (approved):
LGTM, ty!

PR #4518, opened 2024-12-06
tools/tests: Update rhel10 compose url

tkoscieln:
Update a RHEL 10 compose URL to point to nightly instead of public beta. This PR should fix a failing RHEL10 nightly pipeline as the url will point now to correct repositories.


thozza (approved):
LGTM, thanks

PR #4530, opened 2024-12-13
internal/cloudapi/v2/handler: use compose id for filename

schuellerf:
For easier tracing and finding the result in AWS,
can we use the compose id in the filename?

Is this a valid suggestion?

If this might not be unique enough, we could also append some random string?
```
TargetName:   fmt.Sprintf("composer-api-%s-%s", finalJob, uuid.New().String()[:8]),
```

### DIFF
--- a/internal/cloudapi/v2/handler.go
+++ b/internal/cloudapi/v2/handler.go
@@ -1233,7 +1233,7 @@ func (h *apiHandlers) postCloneComposeImpl(ctx echo.Context, id string) error {
 					Ami:          options.Ami,
 					SourceRegion: options.Region,
 					TargetRegion: img.Region,
-					TargetName:   fmt.Sprintf("composer-api-%s", uuid.New().String()),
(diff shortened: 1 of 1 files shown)


internal/cloudapi/v2/handler.go
```diff
@@ -1233,7 +1233,7 @@ func (h *apiHandlers) postCloneComposeImpl(ctx echo.Context, id string) error {
 					Ami:          options.Ami,
 					SourceRegion: options.Region,
 					TargetRegion: img.Region,
-					TargetName:   fmt.Sprintf("composer-api-%s", uuid.New().String()),
+					TargetName:   fmt.Sprintf("composer-api-%s", finalJob),
```
croissanne, line 1236: This only change the clones (so when copying the image over to different regions). 

In general I'm not at all opposed to it but you'll probably need to pick the osbuild job because in koji jobs the final job is the koji-finalize job, which depends on up to 4 osbuild jobs, and they can't all have the same name.

PR #4535, opened 2024-12-18
Add support for advanced partitioning customizations

achilleas-k:
This PR enables the `disk` customizations in osbuild-composer.

Tests for the new blueprint configurations have been added to filesystem.sh.

---

This pull request includes:


internal/blueprint/customizations.go
```diff
@@ -19,6 +19,7 @@ type Customizations struct {
 	Firewall           *FirewallCustomization    `json:"firewall,omitempty" toml:"firewall,omitempty"`
 	Services           *ServicesCustomization    `json:"services,omitempty" toml:"services,omitempty"`
 	Filesystem         []FilesystemCustomization `json:"filesystem,omitempty" toml:"filesystem,omitempty"`
+	Disk               *DiskCustomization        `json:"disk,omitempty" toml:"disk,omitempty"`
```
mvo5, line 22: I have a very stupid (sorry!) question - could we simply do:
```go
iff --git a/internal/blueprint/customizations.go b/internal/blueprint/customiza
tions.go
index 9ec43401f..5efe3eb40 100644
--- a/internal/blueprint/customizations.go
+++ b/internal/blueprint/customizations.go
@@ -5,33 +5,34 @@ import (
        "reflect"
        "strings"
 
+       "github.com/osbuild/images/pkg/blueprint"
        "github.com/osbuild/images/pkg/disk"
 )
...
-       Disk               *DiskCustomization        `json:"disk,omitempty" toml:"disk,omitempty"`
+       Disk *blueprint.DiskCustomization `json:"disk,omitempty" toml:"disk,omitempty"`
...
````
here (and remove disk_customizations.go in this PR)?

And then we would add a "Validate()" that simply disallows the "dc.Type" (as it seems that is the only thing we do not support compared to the images disk_blueprint)?

File: internal/blueprint/disk_customizations.go [hunk 2]
```diff
@@ -0,0 +1,356 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/osbuild/images/pkg/datasizes"
+)
+
+type DiskCustomization struct {
```
mvo5, line 11: See my question above, I really would like to avoid this duplication, I'm worried it will lead to drift and I am also not sure why we need it (but maybe/surely I'm missing something here?)

mvo5 (commented):
Thanks for this PR, I have a very silly question inline, it comes down to if we can find a way to avoid some of the duplication with the images code that are currently present in this PR.

internal/blueprint/customizations.go
achilleas-k: I think with the work we're doing to unify blueprints, it might be a good idea to just drop the whole copy from here and make it use the one from images directly.

achilleas-k:
I need to update image-info to handle btrfs subvolumes and ignore swap partitions when mounting filesystems.

internal/blueprint/customizations.go
thozza: While it makes sense, since we made the `images` version of blueprints used for user-facing (de)serialization from JSON, we keep a few backward compatibility layers in the composer's BPs that are missing in the `images` version. So it may not be doable cleanly. But let's see what you're able to do.

To @mvo5 point about duplication. Given our current situation with the BP structures, I agree that this is a code duplication. However, the intention (and we should try to get back to it) was to separate the user-facing representation of the data structures for (un)marshaling from the data structures used by the code. The goal is to refactor, change, and simplify those "internal" data structures while keeping the user-facing data structures backward compatible. We are not there anymore with BPs, but it is still, in my opinion, a worthwhile goal.

thozza:
> I need to update image-info to handle btrfs subvolumes and ignore swap partitions when mounting filesystems.

Let's try to standardize on the `osbuild-image-info` and remove the `image-info` from this repository.

test/cases/filesystem.sh
```diff
@@ -191,9 +192,271 @@ size = 131072000
...
+  type = "lvm"
+  name = "testvg"
+  minsize = 10_737_418_240
+
+    [[customizations.disk.partitions.logical_volumes]]
+    name = "homelv"
+    mountpoint = "/home"
+    label = "home"
+    fs_type = "ext4"
+    minsize = "2 GiB"
```
thozza, line 309: Is this indentation semantically significant? I've never seen this in TOML :innocent:

.gitlab-ci.yml
```diff
@@ -519,12 +519,29 @@ filesystem.sh:
...
+          - aws/rhel-9.6-nightly-x86_64
+      - CUSTOMIZATION_TYPE: "disk-plain"
+        RUNNER:
+          - aws/rhel-10.0-nightly-x86_64
+      - CUSTOMIZATION_TYPE: "disk-lvm"
+        RUNNER:
+          - aws/centos-stream-9-x86_64
+      - CUSTOMIZATION_TYPE: "filesystem"
+        RUNNER:
+          - aws/centos-stream-10-x86_64
```
thozza, line 542: The resulting combination of which test variants are being tested on which OS is weird. Is there any intention behind it? I'd expect to see all variants being run on all runners :thinking:

test/cases/filesystem.sh
achilleas-k: It's not.  I originally wrote this as json and converted to toml to make sure I did it right (those nested arrays can get weird) and the converter I used added the indentation and I kept it.  I think it makes it easier to read.

.gitlab-ci.yml
achilleas-k: I didn't want the test matrix to grow too much so I added one configuration per distro version.  You're right, it might be a bit confusing.  Also, thinking about it, I think this setup might cause gaps in the future when an old distro version is removed from testing and a new one gets a different configuration.

test/cases/filesystem.sh
thozza: I agree. And I thought that TOML can't get any more "obvious" :sweat_smile:

achilleas-k:
> > I need to update image-info to handle btrfs subvolumes and ignore swap partitions when mounting filesystems.
> 
> Let's try to standardize on the `osbuild-image-info` and remove the `image-info` from this repository.

#4543

internal/blueprint/customizations.go
mvo5: Thanks @thozza - I started a PR in https://github.com/osbuild/images/pull/983 about separating the data structures/json a while ago, might be something to work on further I suppose (it is on halt currently because of the blueprint v2 discussions but it seems code-wise we probably still want/need something like this).

achilleas-k: If you don't mind, I'd like to open a separate PR that drops all duplication here and have osbuild-composer only define blueprint parts that differ for backwards compatibility.

achilleas-k:
With #4568 merged, this should be good to go now.

I'll open a new PR later this week for the blueprint (de)duplication and we can discuss the approach there.

achilleas-k:
Rebased (onto #4543) and fixed conflicts.

achilleas-k:
Rebased on @thozza's #4592, which updates the osbuild dependency but also fixes the spec file.  Thanks, I missed that.

achilleas-k:
https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1653730001
Three failures:
- 2x repository issues:
    - RHEL 9.6 nightly: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9034350641#L2125
    - RHEL 10.0 nightly: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9034350701#L2128
- 1x stupid typo: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9034350685#L2736

thozza (approved):
I've added one comment, but won't block on it...

internal/blueprint/disk_customizations.go
```diff
@@ -0,0 +1,356 @@
...
+	// addition, certain mountpoints have required minimum sizes. See
+	// https://osbuild.org/docs/user-guide/partitioning for more details.
+	// (optional, defaults depend on payload and mountpoints).
+	MinSize uint64 `json:"minsize" toml:"minsize"`
+
+	BtrfsVolumeCustomization
+
+	VGCustomization
+
+	FilesystemTypedCustomization
```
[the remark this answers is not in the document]
achilleas-k: Yeah, there's no really nice way to do this kind of switching I suppose.  I'm happy to reconsider, but any change would have to happen in images.

I'm planning to address the issues @mvo5 brought up earlier about duplicating structures in a later PR, which will remove the copies from osbuild-composer and only keep the bits that are different.  It's a separate topic but related.

Long term, I'm hoping we can autogenerate these from the spec we expect to get from the blueprint unification work.  The autogenerated code is rarely pretty in these cases, but at least that's easier to sweep under the rug when we're designing based on schema and letting the generator deal with the ugliness.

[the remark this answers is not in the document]
thozza: Fair enough :+1:

achilleas-k:
The repository issues were probably because I didn't mark any runners with the internal network variable in `.gitlab-ci.yml`.  I also noticed that not all runner configurations were showing up in the actual pipeline run, but I think I got that sorted now too.

thozza (approved):
:crossed_fingers:

achilleas-k:
Filesystem tests are now passing.  But I'm still seeing fewer jobs than the ones I listed in the CI config:
```yaml
filesystem.sh:
  extends: .integration
  parallel:
    matrix:
      - CUSTOMIZATION_TYPE: "disk-btrfs"
        RUNNER:
          - aws/fedora-40-x86_64
      - CUSTOMIZATION_TYPE: "filesystem"
        RUNNER:
          - aws/rhel-8.10-ga-x86_64
        INTERNAL_NETWORK: ["true"]
      - CUSTOMIZATION_TYPE: "disk-plain"
        RUNNER:
          - aws/rhel-9.4-ga-x86_64
          - aws/rhel-10.0-nightly-x86_64
        INTERNAL_NETWORK: ["true"]
      - CUSTOMIZATION_TYPE: "disk-lvm"
        RUNNER:
          - aws/rhel-9.6-nightly-x86_64
        INTERNAL_NETWORK: ["true"]
      - CUSTOMIZATION_TYPE: "disk-lvm"
        RUNNER:
          - aws/centos-stream-9-x86_64
      - CUSTOMIZATION_TYPE: "filesystem"
        RUNNER:
          - aws/centos-stream-10-x86_64
  variables:
    SCRIPT: filesystem.sh ${CUSTOMIZATION_TYPE}
```

but in [the pipeline](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1655044946/) I'm only seeing:
- Fedora 40 with btrfs
- CentOS 9 with lvm
- RHEL 9.6 with lvm
- RHEL 10.0 with plain
- CentOS 10 with filesystem (old configs)

achilleas-k:
I spoke with @ondrejbudai and we decided that the gitlab CI config could use a lot of simplification and also that we should merge this as is and sort out any other config combinations separately.

internal/blueprint/disk_customizations.go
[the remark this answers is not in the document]
achilleas-k: Resolving this now.

PR #4547, opened 2025-01-06
Exclude vendor directory ci: do not perform linting on vendor/from linting

lzap:
Prevent linting processes from running on the vendor directory to improve efficiency.


bcl (approved):
Looks like a good idea!

mvo5 (approved):

PR #4551, opened 2025-01-09
build(deps): bump to Go 1.22 and simplify prepare-sources.sh

lzap:
Updates Go to 1.22, this is pretty straightforward.

Modifies prepare-sources.sh to avoid downloading of a specific Go version, if I [understand this correctly](https://go.dev/blog/toolchain) it is no longer necessary. This [document explains it even better](https://go.dev/doc/toolchain). Let's try and revert this back if it does not work?

Updated all deps, oapi-gen now prints a warning that v1->v2 upgrade is needed. Since these upgrades can bite, [created an issue for this](https://issues.redhat.com/browse/HMS-5297).

Other than that, it is must a massive `vendor/` update.


mvo5 (commented):
quick drive-by: I think it might be a bit nicer to split this in two commits, one for prepare-sources.sh and one with the rest of the (mostly mechanical?) changes.

lzap:
Good idea, we might need to revert that change if it turns to be problematic. Done.

bcl (approved):
Looks good to me, restarted the failing schutzbot tests...

lzap:
RPM builds were failing:

```
go.mod requires go >= 1.22.10 (running go 1.22.9; GOTOOLCHAIN=local)
```

Downgraded minor version requirement.

lzap:
Fixed RPM builds but still failing.

lzap:
Note for myself: rebase after https://github.com/osbuild/osbuild-composer/pull/4559 is merged

achilleas-k (approved):

PR #4554, opened 2025-01-10
many: take repositories from the "images" library

mvo5:
This commit changes the spec file so that most of the repositories are taken from the `images` library. See images PR#1112 for details.

Note that we still need the -no-auth-keys and the centos-stream symlinks - and we probably need to think about the handling of -no-auth-keys and if it could move under a different "tree" in images or something (like github.com/osbuild/images/data/repositories for the general case and github.com/osbuild/images/data/repositories/no-aux-keys subdir with the same file names or even repositories/for-rhel10 or something like this) but this should be ready for a first round of reviews now.

(see https://github.com/osbuild/images/pull/1112)


repositories/README.md
```diff
@@ -0,0 +1,5 @@
+All repository data is imported via the "images" library, see
+https://github.com/osbuild/images/tree/main/data/repositories
+
+Most are imported via the "vendor" mechanism, but some special
+repositories are kept here.
```
mvo5, line 5: :thinking: Maybe this should also mention that we keep the symlinks here as go:embed does not support those

repositories/README.md
```diff
@@ -0,0 +1,10 @@
+All repository data is imported via the "images" library, see
+https://github.com/osbuild/images/tree/main/data/repositories
+
+Most are imported via the "vendor" mechanism, but some special
+repositories are kept here, like the `*-no-aux-key.json` that
+is only installed on rhel-8.
```
mvo5, line 6: This is not true and needs to be "-no-aux-keys.json" is only installed on rhel-10

thozza: True, this is a typo. Those are el8 repos installed only on el10.

mvo5: Fixed this now with a better explaination.

osbuild-composer.spec
```diff
@@ -159,20 +159,21 @@ install -m 0755 -vd                                                %{buildroot}%
...
+cp -a repositories/centos-stream-%{centos}*          %{buildroot}%{_datadir}/osbuild-composer/repositories/
 %endif
 
 %else
 
 %if 0%{?rhel}
 # RHEL 10 supports building all RHEL versions
 %if 0%{?rhel} >= 10
-for REPO_FILE in $(ls repositories/rhel-* | grep -v 'no-aux-key'); do
+for REPO_FILE in $(ls vendor/github.com/osbuild/images/data/repositories/rhel-* | grep -v 'no-aux-key'); do
```
thozza, line 176: Calling `grep` is no longer needed since there are no `-no-aux-key` repos in `images, right?

osbuild-composer.spec
```diff
@@ -183,13 +184,13 @@ done
 
 %else
 # All other RHEL versions support building for the same version
-for REPO_FILE in $(ls repositories/rhel-%{rhel}* | grep -v 'no-aux-key'); do
+for REPO_FILE in $(ls vendor/github.com/osbuild/images/data/repositories/rhel-%{rhel}* | grep -v 'no-aux-key'); do
```
thozza, line 187: Calling `grep` is no longer needed since there are no `-no-aux-key` repos in `images, right?

osbuild-composer.spec
```diff
@@ -183,13 +184,13 @@ done
...
     install -m 0644 -vp ${REPO_FILE}                               %{buildroot}%{_datadir}/osbuild-composer/repositories/$(basename ${REPO_FILE})
 done
 
 # RHEL 9 supports building also for RHEL 8
 %if 0%{?rhel} == 9
-for REPO_FILE in $(ls repositories/rhel-8* | grep -v 'no-aux-key'); do
+for REPO_FILE in $(ls vendor/github.com/osbuild/images/data/repositories/rhel-8* | grep -v 'no-aux-key'); do
```
thozza, line 193: Calling `grep` is no longer needed since there are no `-no-aux-key` repos in `images, right?

mvo5:
Note that we added fedora-42 to the repositories in "images" recently (c.f. https://github.com/osbuild/images/pull/1183) - having this PR would mean we get the update automatically :)

thozza (approved):
LGTM

thozza:
The Azure test has already failed multiple times. Ive reported https://github.com/osbuild/cloud-image-val/issues/360 :shrug:

supakeen (approved):

mvo5:
I would love to merge this but the current failure make me a bit nervous, afaict none are related to this PR but ideally this would be merged only when main is green - but someone more familiar with the day-to-day woes of composer can chime in here maybe(?).

thozza:
> I would love to merge this but the current failure make me a bit nervous, afaict none are related to this PR but ideally this would be merged only when main is green - but someone more familiar with the day-to-day woes of composer can chime in here maybe(?).

[Schutzbot on GitLab](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1663475030) is the only test that must pass from those that failed...

PR #4562, opened 2025-01-17
README: remove mailing list

schuellerf:
The mailing list was sunset by the IT department and was rarely used, so we'll replace it with matrix & discussions.


ondrejbudai (approved):
Thanks!

PR #4563, opened 2025-01-17
Fix non-constant log strings

bcl:
Newer versions of the go compiler (1.24 in this case) fail when running go test during a mock rebuild of the srpm created by 'make srpm' on Fedora 42.

Even though we currently don't support go1.24, fix these so they don't become an issue when we do.

This pull request includes:


bcl:
Ah, ha! Caused by this 1.24 change - https://tip.golang.org/doc/go1.24#vet

mvo5 (approved):
Nice catch, thank you.

PR #4564, opened 2025-01-18
Add depsolve to cloudapi

bcl:
This pull request includes:

This replaces #4390 and search will be added as a followup PR.


internal/cloudapi/v2/server.go
```diff
@@ -37,6 +37,9 @@ import (
 	"github.com/osbuild/osbuild-composer/internal/worker/clienterrors"
 )
 
+// How long to wait for a depsolve job to finish
+const depsolveTimeout = 5
```
achilleas-k, line 41: The tiniest nitpick:
Should mention this is in minutes since it's now a few hundred lines away from its usage.  Can be part of the comment
```
How long to wait for a depsolve to finish (in minutes)
```
or the variable name
```
depsolveTimeoutMins = 5
```

achilleas-k (dismissed):
Very nice.  LGTM!

achilleas-k:
Bigger topic:

Would it make the depsolve call more useful if we supported a full request as if generating a manifest?  In other words, we can return the full list of packages for a supposed build by asking for the image type as well.  We can keep the old functionality if needed by supporting `imageType = nil`.

I think both use cases are useful tbh.  The existing one is useful for putting some packages in a blueprint and seeing what they'll pull in just by themselves.  Being able to add an image type will tell you what to expect from a proper build using the blueprint.

bcl:
> Would it make the depsolve call more useful if we supported a full request as if generating a manifest?

Yes, I can see cases where you'd want to see the added packages (and repos) needed for the full image. I'd like to add that later though. Next step is to write a search request which will build on this.

achilleas-k:
> I'd like to add that later though.

Yeah, of course.  Topic for a different time.  Figured it was relevant enough to bring up here and see what you think :)

croissanne (approved):
lgtm!

PR #4559, opened 2025-01-15
Update snapshots to 20250115

schutzbot:
Results of the snapshot jobs:
Job(s) succeeded: 206
Job(s) failed: 1

If these are false, rebuild the enumerate cache manually:
Enumerate cache job succeeded: true
Check snapshot succeeded: true

Workflow run: https://github.com/osbuild/rpmrepo/actions/runs/12779805650

PR #4569, opened 2025-01-21
CI: don't test on EOL F39 and update rpmrepos to 20250115

thozza:
- Replace all F39 CI workers with F41.
- Update all rpmrepo snapshots to `20250115`.
- Don't use `which` to check for executables in our test cases, because it is not available by default on F41 any more.
- Adjust the `worker-executor.sh` test case to open `tcp/8001` port on the firewall in the executor VM.


bcl (approved):
Good Idea! Now if Schutzbot will just play along...

achilleas-k:
Like you said, I ran into the executor problem as well in #4559.  I don't know much about the new(ish) worker/executor setup so I couldn't really dig into it quickly.  If you need any help though, let me know and I can start digging.

test/cases/worker-executor.sh
```diff
@@ -151,8 +151,14 @@ subprocessPIDs+=( $! )
 ssh -oStrictHostKeyChecking=no -i "$KEYPAIR" "fedora@$EXECUTOR_IP" sudo dnf install -y osbuild-composer osbuild
 
 # no internet access during the build
+# TODO [thozza]: while debugging the test case, it turned out that the worker executor instance in fact has Internet access!
```
thozza, line 154: @croissanne I didn't spend any time on this thing, because it was unrelated to the test case failure. I'd be happy to fix it, but I wanted to first check with you if you have any ideas because you've spent much more time working with this AWS feature.

thozza:
Note that the failed edge test cases are NOT blocking the merge and are also failing on other PRs. I reached out to EDGE folks and asked for help with debugging...

achilleas-k (approved):
THANK YOU for figuring this out

test/cases/worker-executor.sh
thozza: I've created https://github.com/osbuild/osbuild-composer/issues/4586

PR #4571, opened 2025-01-22
cloudapi/v2: support local upload target

croissanne:
The target validation rework broke the local upload target, which is needed for cockpit-image-builder.


croissanne:
Ugh turns out this target is kinda broken, will open a followup to fix.

PR #4574, opened 2025-01-23
cloudapi: support worker server target result options

croissanne:
Adds results for the worker server target, the full path to the artefact. This is useful for cockpit-image-builder, which can then show the user the path to the image they have built in case they selected a local target.


bcl (approved):
Looks good to me. There is one small issue, that may not really matter, but to save it locally you now need an entry in `upload_targets` where before you could just set `upload_options` on the `image_request`. If you do that now it defaults to an S3 upload and then fails because it isn't setup -- but doesn't fail until the end. So we may want to add a check for this at some point.

```
      "upload_targets": [
        {
          "type": "local",
          "upload_options": {
            "local_save": true
          }
        }
      ],
```

internal/cloudapi/v2/imagerequest.go
```diff
@@ -464,6 +464,9 @@ func getTarget(targetType UploadTypes, options UploadOptions, request *ComposeRe
 
 	case UploadTypesLocal:
 		irTarget = target.NewWorkerServerTarget()
+		irTarget.ImageName = imageType.Filename()
+		irTarget.OsbuildArtifact.ExportFilename = imageType.Filename()
+		irTarget.OsbuildArtifact.ExportName = imageType.Exports()[0]
```
thozza, line 469: This assignment is already handled on line no. 476, so it should not be here. And it would make sense to move there also the `irTarget.OsbuildArtifact.ExportFilename = imageType.Filename()` line, because all of the `new*Target()` functions do the same thing.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -612,7 +612,9 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 		artifact := jobTarget.OsbuildArtifact
 		switch targetOptions := jobTarget.Options.(type) {
 		case *target.WorkerServerTargetOptions:
-			targetResult = target.NewWorkerServerTargetResult(&artifact)
+			targetResult = target.NewWorkerServerTargetResult(&target.WorkerServerTargetResultOptions{
+				Filename: path.Join(jobTarget.OsbuildArtifact.ExportFilename),
```
thozza, line 616: Nitpick: so this is not really a `filename`, but a path to the artifact relative to the job artifacts directory. I think that it would make sense to rename the `Filename` struct member to something that expresses this fact.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -792,6 +793,13 @@ components:
       properties:
         repo_url:
           type: string
+    LocalUploadStatus:
+      type: object
+      required:
+        - filename
+      properties:
+        filename:
+          type: string
```
thozza, line 802: This is not an actual filename but AFAICT, an absolute path to the artifact. I'd suggest renaming it to something that expresses that fact.

thozza (commented):
Thanks. The PR looks good in general. :+1: 

However, I do have a few nitpicks... :innocent:  Specifically, I'd like to request changes to the naming of the newly added struct member, so that it expresses what value it actually holds. IOW don't use "filename" if it holds relative or absolute paths to the artifact. We use "filename" when it is really just the name of the file.

thozza (approved):
Thank you!

PR #4577, opened 2025-01-27
tests/CI: Re-enable the vmware tests on RHEL 10

tkoscieln:
WIP to see if the vmware tests continue to fail on RHEL 10


ondrejbudai (approved):
Thanks! ❤️

thozza (approved):
Nice work!

PR #4579, opened 2025-01-27
Add /search/packages to cloudapi

bcl:
This pull request includes:

### DIFF
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -24,6 +24,13 @@ type RepositoryMTLSConfig struct {
 	Proxy          *url.URL
 }
 
+// SetupRepoSSL copies the CA, Key, and Cert to the RepoConfig
+func (rmc *RepositoryMTLSConfig) SetupRepoSSL(repo *rpmmd.RepoConfig) {
+	repo.SSLCACert = rmc.CA
+	repo.SSLClientKey = rmc.MTLSClientKey
+	repo.SSLClientCert = rmc.MTLSClientCert
+}
+
 func (rmc *RepositoryMTLSConfig) CompareBaseURL(baseURLStr string) (bool, error) {
 	baseURL, err := url.Parse(baseURLStr)
 	if err != nil {
@@ -139,9 +146,7 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 							return err
 						}
 						if match {
-							args.PackageSets[pkgsetsi][pkgseti].Repositories[repoi].SSLCACert = impl.RepositoryMTLSConfig.CA
-							args.PackageSets[pkgsetsi][pkgseti].Repositories[repoi].SSLClientKey = impl.RepositoryMTLSConfig.MTLSClientKey
-							args.PackageSets[pkgsetsi][pkgseti].Repositories[repoi].SSLClientCert = impl.RepositoryMTLSConfig.MTLSClientCert
+							impl.RepositoryMTLSConfig.SetupRepoSSL(&args.PackageSets[pkgsetsi][pkgseti].Repositories[repoi])
 						}
 					}
 				}
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -127,13 +127,21 @@ func workerClientErrorFrom(err error, logWithId *logrus.Entry) *clienterrors.Err
 
 func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	logWithId := logrus.WithField("jobId", job.Id())
+
+	var result worker.DepsolveJobResult
+	// ALWAYS return a result
+	defer func() {
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
 	var args worker.DepsolveJob
 	err := job.Args(&args)
 	if err != nil {
 		return err
 	}
-
-	var result worker.DepsolveJobResult
 
 	if impl.RepositoryMTLSConfig != nil {
 		for pkgsetsi, pkgsets := range args.PackageSets {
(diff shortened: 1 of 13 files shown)


internal/cloudapi/v2/search.go
```diff
@@ -0,0 +1,85 @@
...
+				break
+			}
+		}
+
+		select {
+		case <-ctx.Done():
+			return nil, HTTPErrorWithInternal(ErrorFailedToDepsolve, fmt.Errorf("Search job %q timed out", searchJobID))
+		default:
+		}
+	}
```
thozza, line 82: I'll let other REST API experts comment, but I'm unsure if this POST should wait on the result instead of returning an operation UUID and expecting the caller to GET the result. But maybe this is acceptable. We may end up waiting on many package searches in goroutines, which will take non-trivial amount of time.

cmd/osbuild-worker/main.go
```diff
@@ -440,6 +440,10 @@ var run = func() {
 				Solver:               solver,
 				RepositoryMTLSConfig: repositoryMTLSConfig,
 			},
+			worker.JobTypeSearch: &SearchJobImpl{
```
thozza, line 443: Nitpick: IMHO, it would make sense to include "package" in the job name since we may end up with more than just package search jobs in the future. The API endpoint also contains "package" in the path.

internal/worker/json.go
```diff
@@ -217,6 +217,26 @@ type DepsolveJobResult struct {
 	JobResult
 }
 
+// SearchJob defines the parameters for a dnf metadata search
+// It will search the included repositories for packages matching the
+// package strings
+// TODO -- Does this support globs of any kind?
+type SearchJob struct {
```
thozza, line 224: Nitpick for both `SearchJob` and `SearchJobResult`: I suggest including "package" in the struct name to clarify the job purpose. While we will probably be ever depsolving only packages, we may end up searching for more things just for packages.

thozza (commented):
Thanks for the PR. I've added a few comments, but overall, I like it.

cmd/osbuild-worker/main.go
bcl: Yeah, I thought about that but opted for the shorter string :) I'll take a poll, if people want `SearchPackageJob` instead click the button :)

internal/cloudapi/v2/search.go
bcl: Making the front end have to poll makes things harder. In general once the system is 'warmed up' the depsolve and search should return fairly quickly.

kingsleyzissou: @croissanne can correct me if I'm wrong. I think the POST/GET would be the technically correct way of doing things. But I think there are examples of this pattern in the service already (POST with the result).

Polling with a GET request in the frontend does sounds like a bit of a pain.

cmd/osbuild-worker/main.go
kingsleyzissou: Which button should I click? :) I'm +1 for `SearchPackageJob`, `SearchPackageJobResult` etc

cmd/osbuild-worker/jobimpl-search.go
```diff
@@ -0,0 +1,73 @@
...
+
+	var result worker.SearchJobResult
+
+	if impl.RepositoryMTLSConfig != nil {
+		for repoi, repo := range args.Repositories {
+			for _, baseurlstr := range repo.BaseURLs {
+				match, err := impl.RepositoryMTLSConfig.CompareBaseURL(baseurlstr)
+				if err != nil {
+					result.JobError = clienterrors.New(clienterrors.ErrorInvalidRepositoryURL, "Repository URL is malformed", err.Error())
+					return err
```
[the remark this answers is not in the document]
bcl: Good catch. Totally missed that.

internal/worker/json.go
bcl: Let's see what folks think.

cmd/osbuild-worker/main.go
thozza: Or `PackageSearchJob` :innocent:

cmd/osbuild-worker/jobimpl-depsolve.go
```diff
@@ -127,6 +134,13 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	}
 
 	var result worker.DepsolveJobResult
+	// ALWAYS return a result
+	defer func() {
+		err = job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
```
thozza, line 143: Technically, this should go before line number 132. But it can be a follow-up.

cmd/osbuild-worker/jobimpl-search.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	var result worker.SearchJobResult
+	// ALWAYS return a result
+	defer func() {
+		err := job.Update(&result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
```
thozza, line 45: Technically, this should go before line number 34. But it can be a follow-up.

internal/worker/server.go
```diff
@@ -33,6 +33,7 @@ const (
 	JobTypeKojiInit         string = "koji-init"
 	JobTypeKojiFinalize     string = "koji-finalize"
 	JobTypeDepsolve         string = "depsolve"
+	JobTypeSearch           string = "search"
```
thozza, line 36: Nitpick: this should ideally have "package" in the variable name and value.

thozza (approved):
Thanks for all the changes. I've added a few nitpicks.

The job type value change has the potential to break server <-> worker integration if merged, deployed and then changed. However, this job type won't be used in any deployment just yet, so it should be OK to change it in a follow-up.

cmd/osbuild-worker/jobimpl-depsolve.go
bcl: Crap. No, let's get it right first :)

kingsleyzissou (dismissed):

bcl:
Ok, with this update I've fixed the location of the deferred update functions, renamed all the various `Search` things to `SearchPackages`

kingsleyzissou (approved):

thozza (approved):
LGTM, thanks for your patience!

PR #4582, opened 2025-01-29
composer: use `reporegistry.New()` again (see PR#4378)

mvo5:
This commit changes cmd/composer to use the `reporegistry.New` call again. This is a preparation for
https://github.com/osbuild/images/pull/1179

And undoes parts of https://github.com/osbuild/osbuild-composer/pull/4378 but that is no longer necessary because in
https://github.com/osbuild/images/pull/946 the error is now passed on from `reporegistry.New()` in the same way as from `LoadRepositories()`.


thozza (approved):
Thanks, LGTM.

PR #4587, opened 2025-01-31
internal/cloud/gcp/compute: Add TDX_CAPABLE guest OS feature

bgartzi:
Latest RHEL images (from 9.6 on) should fully support running as TDX guests, as well as latest CentOS images.

See: https://issues.redhat.com/browse/COS-3111
See: https://github.com/coreos/coreos-assembler/pull/4006
See: https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/5979

This pull request includes:

I couldn't find anything relevant to be changed to address the points above.

### DIFF
--- a/internal/cloud/gcp/compute.go
+++ b/internal/cloud/gcp/compute.go
@@ -29,6 +29,18 @@ var GuestOsFeaturesRHEL9 []*computepb.GuestOsFeature = []*computepb.GuestOsFeatu
 	{Type: common.ToPtr(computepb.GuestOsFeature_GVNIC.String())},
 	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_SNP_CAPABLE.String())},
 	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_LIVE_MIGRATABLE_V2.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_TDX_CAPABLE.String())},
+}
+
+// Guest OS Features for RHEL images up to RHEL9.5.
+// The TDX support was added since RHEL-9.6.
+var GuestOsFeaturesRHEL95 []*computepb.GuestOsFeature = []*computepb.GuestOsFeature{
+	{Type: common.ToPtr(computepb.GuestOsFeature_UEFI_COMPATIBLE.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_VIRTIO_SCSI_MULTIQUEUE.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_CAPABLE.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_GVNIC.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_SNP_CAPABLE.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_LIVE_MIGRATABLE_V2.String())},
 }
 
 // Guest OS Features for RHEL9.1 images.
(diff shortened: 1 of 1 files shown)


bgartzi:
Hi @thozza, @ondrejbudai, based on other similar PRs I think you could provide some feedback about this. Would any of you mind having a look?
Thanks! :)

internal/cloud/gcp/compute.go
```diff
@@ -29,6 +29,18 @@ var GuestOsFeaturesRHEL9 []*computepb.GuestOsFeature = []*computepb.GuestOsFeatu
 	{Type: common.ToPtr(computepb.GuestOsFeature_GVNIC.String())},
 	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_SNP_CAPABLE.String())},
 	{Type: common.ToPtr(computepb.GuestOsFeature_SEV_LIVE_MIGRATABLE_V2.String())},
+	{Type: common.ToPtr(computepb.GuestOsFeature_TDX_CAPABLE.String())},
+}
+
+// Guest OS Features for RHEL9.2 images.
+// The TDX support was added since RHEL-9.6.
+var GuestOsFeaturesRHEL92 []*computepb.GuestOsFeature = []*computepb.GuestOsFeature{
```
thozza, line 37: Nitpick: I think that the comment should state `for RHEL images up to 9.5` and I'd suggest to also name the variable `GuestOsFeaturesRHEL95`

thozza (commented):
Thank you!

I have one tiny suggestion. In addition, I'd like to ask what about the support in c10s? Should the capability be added there as well?

internal/cloud/gcp/compute.go
bgartzi: Addressed this in the last version.

thozza (approved):
LGTM, thanks!

bgartzi:
Hi @thozza!

I asked the centos-devel mailing list whether centos-stream 10 images should be marked as TDX_CAPABLE or not. However, I haven't received any reply.

However, there are many reasons that make me think we should keep this going:
- TDX guest enablement patches are included in the centos  stream 10 kernel, if I understood correctly.
- centos stream 9 is already marked as TDX_CAPABLE.
- Both RHEL 9 (from 9.6 on) and 10 are marked as TDX_CAPABLE.
- I created a custom centos stream 10 image in GCP marked as `TDX_CAPABLE`. I booted a TDX confidential computing machine with such guest. It booted correctly and traces indicated that TDX was enabled successfully.

thozza:
Thanks @bgartzi .

Would you mind extending this PR or submitting a new one?

bgartzi:
Sorry @thozza, I might be missing something. Doesn't the actual patch already imply that c10s images will carry `TDX_CAPABLE`? It currently inherits the guest OS features from rhel 10, which at the same time inherits them from RHEL-9's, right?

thozza:
> Sorry @thozza, I might be missing something. Doesn't the actual patch already imply that c10s images will carry `TDX_CAPABLE`? It currently inherits the guest OS features from rhel 10, which at the same time inherits them from RHEL-9's, right?

Sorry, I didn't check the complete code, only the PR diff chunks. You are right.

bgartzi:
Thanks for confirming @thozza!

ondrejbudai (commented):
Shouldn't then we have a case for RHEL-9.6 as well, since it's mentioned in the comment?

thozza:
> Shouldn't then we have a case for RHEL-9.6 as well, since it's mentioned in the comment?

No.

RHEL 9.6 and later uses `GuestOsFeaturesRHEL9`. `GuestOsFeaturesRHEL95` is used only up to 9.5, including it, unless there is a different condition.

ondrejbudai (approved):
Right, I missed the first line of the diff, thanks, and sorry for my confusion.

PR #4590, opened 2025-01-31
cloudapi download

bcl:
This pull request includes:


croissanne (commented):
So I wonder if we even need this. Currently we implemented downloading / copying etc… the artefact just by using cockpit-files (https://github.com/osbuild/image-builder-frontend/pull/2820).

I think it might make more sense to expand the local target to accept a target artefact location / output directory instead (like ~/.cache/cockpit-image-builder/artefacts) or something.

bcl:
I think we should have both. The cloudapi is incomplete without this, and may be accessed by things that can't use the local filesystem or cockpit plugins. And doing it that way feels like a hack -- it is breaking the separation between the internals and the client needing to know details like file paths.

croissanne:
> I think we should have both. The cloudapi is incomplete without this, and may be accessed by things that can't use the local filesystem or cockpit plugins. And doing it that way feels like a hack -- it is breaking the separation between the internals and the client needing to know details like file paths.

Hm I think the separation is maintained as long as the path to the artefacts are resolved by composer, it's in the upload status, part of the osbuild job result. So it's not like the user is aware of where exactly the artefacts are saved, and if the artefact directory would change, the cockpit plugin wouldn't have to be updated.  So I'm not sure if it's a hack. Though I agree it would be nicer to save it somewhere else, so that we don't have to point the user to /var/lib/osbuild-composer, which is an internal directory.

bcl:
I'm not sure what you mean by it being in the upload status. But either way I need this in order to finish the move from weldr api to cloud api, and for composer-cli. The cloudapi is incomplete without this.

bcl:
> I'm not sure what you mean by it being in the upload status.

LOL. I figured it out :)

lucasgarfield:
@croissanne we'll also want this for composer-cli as part of our work towards deprecating the weldr api, won't we?

internal/cloudapi/v2/handler.go
```diff
@@ -276,7 +276,7 @@ func (h *apiHandlers) targetResultToUploadStatus(jobId uuid.UUID, t *target.Targ
 		workerServerOptions := t.Options.(*target.WorkerServerTargetResultOptions)
 		absPath, err := h.server.workers.JobArtifactLocation(jobId, workerServerOptions.ArtifactRelPath)
 		if err != nil {
-			return nil, fmt.Errorf("unable to find job artefact: %w", err)
+			return nil, fmt.Errorf("unable to find job artifact: %w", err)
```
kingsleyzissou, line 279: Think this is just UK vs US English 😉

kingsleyzissou (commented):
Thanks this look good to me, maybe @croissanne can have another look first

kingsleyzissou (approved):

internal/cloudapi/v2/handler.go
bcl: Right, but everywhere we use it we use the US version. Consistency is important for grepping ;)

internal/cloudapi/v2/handler.go
```diff
@@ -1514,3 +1514,54 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+		return HTTPErrorWithInternal(ErrorArtifactNotFound, err)
+	}
+
+	// Building only supports one target, but that may change, so make sure to check.
+	// NOTE: TargetResults isn't populated until it is finished
+	if len(osbuildResult.TargetResults) != 1 {
+		msg := fmt.Errorf("%#v", osbuildResult.TargetResults)
+		//return HTTPError(ErrorSeveralUploadTargets)
+		return HTTPErrorWithInternal(ErrorSeveralUploadTargets, msg)
+	}
```
thozza, line 1555: Building supports multiple targets. We do use it at least for Koji builds
https://github.com/osbuild/osbuild-composer/blob/bdd2014c4467e9325b29624439c98584addc681f/internal/cloudapi/v2/server.go#L398-L410
and Weldr API already uses it as well
https://github.com/osbuild/osbuild-composer/blob/bdd2014c4467e9325b29624439c98584addc681f/internal/weldr/api.go#L2496-L2506

You could just filter for the appropriate target type using https://github.com/osbuild/osbuild-composer/blob/bdd2014c4467e9325b29624439c98584addc681f/internal/worker/json.go#L103-L111

internal/worker/server.go
```diff
@@ -651,7 +651,7 @@ func (s *Server) JobArtifactLocation(id uuid.UUID, name string) (string, error)
 
 	p := path.Join(s.config.ArtifactsDir, id.String(), name)
 	if _, err := os.Stat(p); errors.Is(err, os.ErrNotExist) {
-		return "", fmt.Errorf("Artifact not found: %s", p)
+		return p, fmt.Errorf("Artifact not found: %s", p)
```
thozza, line 654: The `p` as the path that was checked is part of the error message. I'd say that the correct way would be to return a structured error that has the path stored in a property, so that it can be accessed.

thozza (commented):
Thanks, I've added a few suggestions.

thozza (approved):
I mean, all of my suggestions can be a follow-up

internal/worker/server.go
bcl: This is on purpose so that when the file isn't present the path that it is expected to be at can be used to write it in `TestDownload`, this is much cleaner than creating a new error structure.

internal/cloudapi/v2/handler.go
bcl: The structure allows multiple downloads, but there is a limit on them someplace in the compose path. This is also limited to the `JobTypeOSBuild` so koji, etc. doesn't matter here. I can't seem to find the check though, all I am seeing right now are the ones in `PostCloneCompose` which isn't what I'm thinking of. My plan is to return a tar if there are multiple download items, so that would be fine as a later change.

thozza: There is no limit on the number of targets in the server-worker implementation, only in CloudAPI where `ImageRequest` can contain only single `upload_options`. Additional targets are added only implicitly for special situations by the API handler code, specifically the Koji use case. To make cloud uploads work in addition to the local target, you'll have to either extend the API to support multiple upload options or add the local target implicitly when using the Cloud API on-prem with `composer-cli`. In any case, you'll almost certainly end up filtering for the "worker server" target result to get the path to the local artifact, so that you can return it.

The endpoint should be returning only "worker server" target results and I don't expect that we will ever support builds with more than one such target in it. So while the check that there is just one such target result would be necessary, I don't think that there is any use case for creating a tarball and returning it.

internal/worker/server.go
thozza: Although I disagree on what is cleaner, I can accept the current version. Defining a custom error struct seems easy and clean to me. :shrug: Especially if the only use case is in tests.

If you keep it as is, it would be good to note the meaning of the returned value in the function doc text in case an error is also returned.

PR #4599, opened 2025-02-06
go.mod: update to latest images@v0.115.0

achilleas-k:
Primarily for https://github.com/osbuild/images/pull/1195

---
**tools: revert prepare-source.sh to use downloaded go binary**

The new script can cause changes to the go.mod when executing
`go get go@version`, downgrading packages that we want to bump, which
can revert changes we want the tool to be tidying.

We can work around this for now, but with the script being part of our
PR checks, we also need to revert the script so we can get module
updates through.

This reverts 8554d6202dff0f887d0f8bec838f15c24af02e8d

---

**go.mod: update to latest images@v0.115.0**

---


runcom:
thanks, lgtm

thozza (approved):
LGTM

runcom:
don't we report back edge CI results? I can see the failures in osbuild/rhel-edge-ci but they're not reported here nor it blocks the PR if those fails cc @yih-redhat

thozza:
> don't we report back edge CI results? I can see the failures in osbuild/rhel-edge-ci but they're not reported here nor it blocks the PR if those fails cc @yih-redhat

They are reported, but we don't block on them. The reason is mostly that nobody on the team is deeply familiar with those tests, and they proved to be unstable without us changing any EDGE-related code. :roll_eyes:

runcom:
> EDIT: in this case, the status is really still "in progress".

they're actually done in the rhel-edge-ci repo :/ so maybe something isn't working?

runcom:
right, ok I see them now :smile: 

although the tests are done afaict https://artifacts.osci.redhat.com/testing-farm/8876a623-b410-499c-affd-727dbb89054f/

but they don't come back here

something for @yih-redhat (once he's back I guess)

PR #4600, opened 2025-02-06
tools: use go toolbox instead of installing a binary

achilleas-k:
Repeat of 8554d6202dff0f887d0f8bec838f15c24af02e8d.

This is a much nicer (and newer) way of handling the Go version for the
project.  The issue that caused the previous revert was that the toolbox
version was being set to 1.22.0, which is older than what some
dependencies require, specifically osbuild/images, so the 'go mod tidy'
would downgrade images to match the required toolbox version.

Setting the go and toolbox version to 1.22.6 should make it all work as
expected.

This reverts commit 7e87d1e124c1bbd49f7a0afcd732fd83357f4e57.

Co-authored-by: Lukáš Zapletal <lukas@zapletalovi.com>

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -1,9 +1,20 @@
 #!/bin/sh
 
-set -eux
+set -eu
 
-# Pin Go and toolbox versions at a reasonable version
-go get go@1.22.6 toolchain@1.22.6
+GO_MINOR_VERSION="1.22"
+GO_VERSION="${GO_MINOR_VERSION}.6"
+
+# Check latest Go version for the minor we're using
+LATEST=$(curl -s https://endoflife.date/api/go/"${GO_MINOR_VERSION}".json  | jq -r .latest)
+if test "$LATEST" != "$GO_VERSION"; then
+    echo "NOTE: A new minor release is available (${LATEST}), consider bumping the project version (${GO_VERSION})"
+fi
+
+set -x
+
+# Pin Go and toolbox versions
+go get "go@${GO_VERSION}" "toolchain@${GO_VERSION}"
 
 # Update go.mod and go.sum:
 go mod tidy
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -2,20 +2,15 @@
 
 set -eux
 
-GO_VERSION=1.22.6
-GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
+# Pin Go and toolbox versions at a reasonable version
+go get go@1.22.6 toolchain@1.22.6
 
-# this is the official way to get a different version of golang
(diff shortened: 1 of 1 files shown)


achilleas-k:
Rebased on main and updated commit ID for the revert in the commit message.

tools/prepare-source.sh
```diff
@@ -2,20 +2,15 @@
...
-$GO_BINARY mod vendor
+# Generate all sources (skip vendor/):
+go generate ./cmd/... ./internal/... ./pkg/...
 
-# ... and all code has been regenerated from its sources.
-$GO_BINARY generate ./...
-
-# ... the code is formatted correctly, ...
-$GO_BINARY fmt ./...
+# Generate all sources (skip vendor/):
```
bcl, line 15: Typo :) should be 'Format all sources'

bcl (approved):

tools/prepare-source.sh
achilleas-k: Oh no!!

bcl (dismissed):
That's better! ;)

tools/prepare-source.sh
```diff
@@ -2,20 +2,15 @@
 
 set -eux
 
-GO_VERSION=1.22.6
-GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
+# Pin Go and toolbox versions at a reasonable version
```
lzap, line 5:
```suggestion
# Pin Go and toolbox versions at the latest stable version
```

lzap (approved):
Yeah I really hope this will work, cos once we update to 1.24 with its [tool version fix](https://tip.golang.org/doc/go1.24#go-command) that will be a complete and hassle-free experience, finally!

tools/prepare-source.sh
lzap: There is an API that returns N and N-1 latest stable release: `curl https://go.dev/VERSION?m=text` and https://go.dev/dl/?mode=json tho I think having the version hardcoded is probably better.

lzap: Pseudo-code (not tested):

```bash

if $(curl 'https://go.dev/dl/?mode=json'|jq -r '.[-1].version') != "v$GO_VERSION"; then
 echo "WARNING: A new minor release is available, consider bumping the version"
fi
```

lzap (approved):
Awesome!

achilleas-k:
Had to downgrade to Go 1.22.11 because Fedora 40 doesn't have .12

achilleas-k:
I dropped the last commit that updates the Go version.  We still have RHEL 8.10 and 9.4 RPM builds going on here, so let's just get the new script in and look at distro and compiler versions separately.

thozza (approved):

PR #4545, opened 2025-01-03
build(deps): bump deps specifically CVE-2024-45338

lzap:
SSIA

PR #4603, opened 2025-02-06
many: update for new reporegistry.New() api (c.f. pr#1179)

mvo5:
This commit updates osbuild-composer for the new API in images for the `reporegistry.New()`. The main incompatible change is that the `/repositories` part is not longer automatically added inside the library so we need to add it on the call-site.

This needs https://github.com/osbuild/images/pull/1179


supakeen (approved):

internal/cloudapi/v2/compose_test.go
```diff
@@ -812,7 +812,7 @@ func TestGetImageRequests_NoRepositories(t *testing.T) {
 		},
 	}
 	// NOTE: current directory is the location of this file, back up so it can use ./repositories/
-	rr, err := reporegistry.New([]string{"../../../"})
+	rr, err := reporegistry.New([]string{"../../../repositories"}, nil)
```
thozza, line 815: Could we use the embedded repositories from `images` for tests? This will, IMO break once https://github.com/osbuild/osbuild-composer/pull/4554 lands, and if we are not careful and don't rebase the PR that lands as the last, it could pass CI and break `main`.

internal/cloudapi/v2/compose_test.go
```diff
@@ -838,7 +838,7 @@ func TestGetImageRequests_BlueprintDistro(t *testing.T) {
...
```
thozza, line 841: ditto using repos from `images`

internal/cloudapi/v2/v2_test.go
```diff
@@ -247,7 +247,7 @@ func newV2Server(t *testing.T, dir string, enableJWT bool, fail bool) (*v2.Serve
 	distros := distrofactory.NewTestDefault()
 	require.NotNil(t, distros)
 
-	repos, err := reporegistry.New([]string{"../../../test/data"})
+	repos, err := reporegistry.New([]string{"../../../test/data/repositories"}, nil)
```
thozza, line 250: ditto using repos from `images`

internal/cloudapi/v2/compose_test.go
mvo5: Sure thing, I updated the code to pull in the repos from images now,

internal/cloudapi/v2/compose_test.go
mvo5: Same as above

internal/cloudapi/v2/v2_test.go
mvo5: This one is slightly different, in #4545 we will not remove the test repos (yet) from composer so this should be fine. I think we still want to also remove the test/data/repositories from composer but just a naive:
```go
diff --git a/internal/cloudapi/v2/v2_test.go b/internal/cloudapi/v2/v2_test.go
index f6d249923..1c375b33b 100644
--- a/internal/cloudapi/v2/v2_test.go
+++ b/internal/cloudapi/v2/v2_test.go
@@ -12,21 +12,20 @@ import (
        "github.com/google/uuid"
        "github.com/stretchr/testify/require"
 
-       "github.com/osbuild/osbuild-composer/pkg/jobqueue"
-
        "github.com/osbuild/images/pkg/distro/test_distro"
        "github.com/osbuild/images/pkg/distrofactory"
        "github.com/osbuild/images/pkg/osbuild"
        "github.com/osbuild/images/pkg/ostree/mock_ostree_repo"
-       "github.com/osbuild/images/pkg/reporegistry"
        "github.com/osbuild/images/pkg/rpmmd"
        "github.com/osbuild/images/pkg/sbom"
+       testrepos "github.com/osbuild/images/test/data/repositories"
        v2 "github.com/osbuild/osbuild-composer/internal/cloudapi/v2"
        "github.com/osbuild/osbuild-composer/internal/jobqueue/fsjobqueue"
        "github.com/osbuild/osbuild-composer/internal/target"
        "github.com/osbuild/osbuild-composer/internal/test"
        "github.com/osbuild/osbuild-composer/internal/worker"
+       "github.com/osbuild/osbuild-composer/pkg/jobqueue"
 )
 
 var sbomDoc = json.RawMessage(`{
@@ -247,8 +246,8 @@ func newV2Server(t *testing.T, dir string, enableJWT bool, fail bool) (*v2.Serve
        distros := distrofactory.NewTestDefault()
        require.NotNil(t, distros)
 
-       repos, err := reporegistry.New([]string{"../../../test/data/repositories"}, nil)
-       require.Nil(t, err)
+       repos, err := testrepos.New()
+       require.NoError(t, err)
        require.NotNil(t, repos)
        require.Greater(t, len(repos.ListDistros()), 0)
 

````
leads to many errors because the test/data/repositories dir between "images" and "osbuild-composer" has drifted and just doing the above fails. I *think* its just a matter of adding the "test-distro.json" to the images test distros but I have not investigated fully yet. If you prefer it here over a followup I can do this but keeps this as is should be okay and we do the test repos in a followup(?)

thozza: Right...
Note that `test/data/repositories` are used by all (I think) test cases to override repos installed by the RPM, so removing them completely won't be easy. It should be IMHO OK to use the embedded test repos for unit tests, but we can do it in steps...

thozza (approved):
Thanks, LGTM.

thozza:
@mvo5, there is now a merge conflict. Can you please take a look? Thanks!

mvo5:
@thozza Sorry, this grew again, the removal of fedora-39 from images caused another test failure in the cross-arch.sh so I had to bump images to the current version and vendor all dependencies etc.

achilleas-k:
I amended the commit that updates images to v0.117.0 to also update the Go version in `tools/prepare-source.sh` to 1.22.8, which is required by the new dependency version.

achilleas-k (approved):

bcl (approved):
Looks good to me!

PR #4607, opened 2025-02-11
Switch github workflows to ubuntu 22.04

bcl:
Changes the github workflows to use ubuntu-22.04 everywhere.

### DIFF
--- a/.github/workflows/coverity_scan.yml
+++ b/.github/workflows/coverity_scan.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   coverity_scan:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   tag-and-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Even or odd week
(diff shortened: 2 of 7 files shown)


mvo5 (commented):
A bit of a meta-question (I guess its not very relevant for this PR but I'm curious): why do we pick `ubuntu-22.04` instead of `ubuntu-24.04` as the "stable" relase? AIUI ubuntu-latest points to ubuntu-24.04 since a few weeks [0]. 

The change itself looks fine and can be merged of course.
[0] https://github.com/actions/runner-images/issues/10636

mvo5 (approved):

thozza:
> Why not use "latest"? Do we have a reason to pin to a specific version? This way we automatically get updates and if there are problems with the actions we need to fix them anyway…

Generally, we pin things in CI to keep it stable and know the reason for a potential failure. If the underlying infra changes automatically and causes the CI to fail, you may wonder if it is related to the PR. And when we update things related to CI infrastructure, we can be sure that the failure is related to that change.

thozza (approved):

achilleas-k:
> @thozza I'm not sure if I get the "pin things in CI" thing with `sudo apt-get update` ;)
> 
> https://github.com/osbuild/osbuild-composer/blob/53d50e875d5474dbdb516f8f894de5d5448c8654/.github/workflows/tests.yml#L67

That just pulls in repo metadata so we can `apt install` the dependencies.

Dependency version changes during an LTS release is unlikely to cause issues (though not impossible).  The distro release changing from under your feet however can bring much larger package changes (major versions, even package names can change).

bcl:
The original thoughts on this are in commit https://github.com/osbuild/osbuild-composer/commit/e8ae7e7cae074363708566ffe480669fa24a5f79 and I read that as trying to keep it from jumping to a new version of the os, not pinning every last package.

PR #4612, opened 2025-02-14
Use golangci-lint v1.60 and cleanup the errors it returns

bcl:
Currently a draft because I am not sure how to handle:
```
internal/cloudapi/v2/handler.go:467:17: G115: integer overflow conversion uint64 -> int (gosec)
                buildID := int(initResult.BuildID)
                              ^
cmd/osbuild-worker/jobimpl-koji-finalize.go:314:19: G115: integer overflow conversion uint64 -> int64 (gosec)
                StartTime: int64(args.StartTime),
                                ^
cmd/osbuild-worker/jobimpl-koji-finalize.go:105:40: G115: integer overflow conversion uint64 -> int (gosec)
                        err = impl.kojiFail(args.Server, int(initArgs.BuildID), initArgs.Token)
```

Ideally these would just use the proper type from the start, but I don't know exactly how this interacts with koji, what values it expects or returns.

### DIFF
--- a/internal/cloudapi/v2/handler.go
+++ b/internal/cloudapi/v2/handler.go
@@ -464,7 +464,13 @@ func (h *apiHandlers) getJobIDComposeStatus(jobId uuid.UUID) (ComposeStatus, err
 			ImageStatuses: &buildJobStatuses,
 			KojiStatus:    &KojiStatus{},
 		}
+		/* #nosec G115 */
 		buildID := int(initResult.BuildID)
+		// Make sure signed integer conversion didn't underflow
+		if buildID < 0 {
+			err := fmt.Errorf("BuildID integer underflow: %d", initResult.BuildID)
+			return ComposeStatus{}, HTTPErrorWithInternal(ErrorMalformedOSBuildJobResult, err)
+		}
 		if buildID != 0 {
 			response.KojiStatus.BuildId = &buildID
 		}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,3 @@
-exclude-dirs:
-  - vendor
-
 linters-settings:
   govet:
     disable:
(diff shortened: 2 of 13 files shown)


achilleas-k (approved):
Very nice.  Thank you!  LGTM

lzap:
Note for myself: integrate this into image-builder-crc, I like how linter version is extracted from the makefile that is `ka-wa-ii`! :-D

internal/cloudapi/v2/handler.go
```diff
@@ -464,7 +464,13 @@ func (h *apiHandlers) getJobIDComposeStatus(jobId uuid.UUID) (ComposeStatus, err
 			ImageStatuses: &buildJobStatuses,
 			KojiStatus:    &KojiStatus{},
 		}
+		/* #nosec G115 */
```
schuellerf, line 467: I'd also add a comment for the rationale here … but no showstopper I think

schuellerf (dismissed):
LGTM

lzap:
Hold on I am trying this myself and tag `v1.60` is no more in that docker repo. What is going on?

https://hub.docker.com/r/docker/golangci-lint/tags

bcl:
I don't see the tags there, did they go and delete all the ones without the go version included?
Locally it is still working for me despite my best efforts to clean up caches and things.
I'll switch it to `1.60.3-go1.23.0` to make sure it works for everyone. Super annoying how the thing that's supposed to be reproducible isn't...

internal/cloudapi/v2/handler.go
bcl: Actually the description is right below, it could underflow so we have to ignore it, but the underflow is checked right after.

bcl:
Updated the comment in the Makefile to hopefully prevent tag confusion. I got tangled up in that as well, despite Flo's reply.

internal/cloudapi/v2/handler.go
schuellerf: … right - 2 LOC - that was beyond my attention span 🫣 😁

schuellerf (approved):
LGTM

achilleas-k (approved):

PR #4614, opened 2025-02-14
cloudapi: silence logrus in tests

mvo5:
This commit discards the logrus output when the tests are run. That might be a controversial change but the amount of redundant output that drowns the actual failures makes it very hard to work with these tests.

I want to paste the output here in full but github won't let me because it's too long for the limits that the body can have!
It is 1135 lines and exceeds the 65536 bytes limit of GH.

### DIFF
--- a/internal/cloudapi/v2/v2_internal_test.go
+++ b/internal/cloudapi/v2/v2_internal_test.go
@@ -1,8 +1,10 @@
 package v2
 
 import (
+	"io"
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Just a quick thought: would it be valueable to go over the actual log statements, and find a better level for them so the default behaviour is sane?

thozza (approved):
LGTM

PR #4629, opened 2025-02-27
Preparation for RHEL 9.7 and 10.1 (COMPOSER-2451)

thozza:
- update `osbuild/images` to v1.120.0, which adds RHEL 10.1 and 9.7 repositories
- bump the `rhel-9` and `rhel-10` distro aliases used on-prem to 9.7 and 10.1, because we don't plan any new releases to 9.6 and 10.0.


ondrejbudai (approved):
Go!

PR #4644, opened 2025-03-11
dnf modularity wip

croissanne:
- **weldr: adapt tests**
- **weldr: add enabled modules to depsolve handler**
- **blueprint: add enabled modules to blueprint**
- **cloudapi: support dnf modules**
- **cloudapi: small changes to unit tests, maybe squash**
- **api test**


supakeen:
> Adds enabled_modules as a list of strings with format "$name:$stream".
> 
> Alternatively we could switch to `[{"name": …, "stream": …}]`, not sure what ppl prefer.

Without looking at the code (I'll do that in a bit) it might *feel* like `$name:stream` would also accept a `/profile`; that issue is less apparent with structured data? I don't know :)

achilleas-k (approved):

supakeen (approved):

PR #4647, opened 2025-03-13
test: migrate the openshift virt test to a new cluster (COMPOSER-2458)

ondrejbudai:
The old one is going to be decommissioned. I only changed:
- extracted the storage class to a variable
- adjusted the openshift yaml file to what I was given in the UI
  - most importantly, we now use an instancetype to specify the
    resource requirements instead of doing it manually
  - the network is called default, instead of nic0 on this cluster
- we are downloading the oc and virtctl clients from the new cluster
  so the versions match

+ I removed wget, and introduced retries to help with flakiness.


tkoscieln (approved):

ondrejbudai:
@tkoscieln verified that this will make the nightly pipeline happy: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1725888571 Let's get it in.

PR #4650, opened 2025-03-17
Advanced partitioning customizations in cloud API

achilleas-k:
Update osbuild/images to v0.124.0 and add the new Disk customizations (advanced partitioning) to the clouadpi.

/jira-epic COMPOSER-2399


internal/common/pointers.go
```diff
@@ -3,3 +3,13 @@ package common
 func ToPtr[T any](x T) *T {
 	return &x
 }
+
+// DerefDefault returns the dereferenced value of the given pointer or the
+// default value for the type if unset.
+func DerefDefault[T any](p *T) T {
```
mvo5, line 9: Maybe `DerefOrDefault` (i.e. *Or* in the middle?) and maybe a small unit test. 

Fwiw, I wish we wonder if we could consider a more holistic approach to this problem, I opened https://github.com/osbuild/images/commit/fc3af96416cb9a3ee47938d7494843ceb74f8c72 a while ago for discussion which tries to tackle a similar issue. But its fine of course, there are tradeoffs and this is nice and simple.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -2548,6 +2554,148 @@ components:
...
+        type:
+          type: string
+          enum:
+            - gpt
+            - dos
+          description: |
+            Type of the partition table
+        minsize:
+          x-go-type: uint64
+          example: 2147483648
```
mvo5, line 2571: IIRC we support strings like "20 GiB" here, would it make sense to use them in the example(s)?

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -2548,6 +2554,148 @@ components:
         license:
           type: string
 
+    Disk:
```
mvo5, line 2557: This looks very nice and straightforward

mvo5 (commented):
Very nice! A few questions/ideas inline

internal/common/pointers.go
achilleas-k: Yeah, that is much nicer.  Not sure why it stalled back then tbh.

internal/cloudapi/v2/openapi.v2.yml
achilleas-k: I see that other sizes are also just uints in the cloud API.  I think we just don't support strings here and the IB middleware probably takes care of the string parsing?

croissanne: We could do this either way, either accept an int or string, or both. Will require some manual parsing if it's string though.

croissanne (changes requested):
LGTM except the osbuild bump required in the schutzfile.

internal/cloudapi/v2/openapi.v2.yml
achilleas-k: Let me try to do it in the spec and see what the code looks like.

achilleas-k: Answer: Not great.

The following:
```yaml
    minsize:
      oneOf:
        - name: minsize
          x-go-type: uint64
          example: 2147483648
          description: 'size in bytes'
        - name: minsize
          type: string
          description: 'size with data units'
```
works, but in Go it just creates a `Minsize` type alias to `interface{}`.  It's not impossible to work with, but it needs some fiddling.

achilleas-k: Used that `oneOf` and handled parsing.  PTAL.

achilleas-k:
I dropped the osbuild/images version update to avoid conflicts with #4651 which takes priority right now.

achilleas-k:
> I dropped the osbuild/images version update to avoid conflicts with #4651 which takes priority right now.
> 
> This will break the PR as it is now.

Brought it back because the other PR is going to be delayed a bit.

croissanne (dismissed):
lgtm

Schutzfile
```diff
@@ -2,7 +2,7 @@
   "fedora-40": {
     "dependencies": {
       "osbuild": {
-        "commit": "e4333f87ba14f85ba3d9aa5b070190d82d06a297"
+        "commit": "74d6eb37bdd6e95475c9e0b778195697922c570c"
```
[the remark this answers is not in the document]
achilleas-k: Yes.

achilleas-k:
Resolved conflicts.

achilleas-k:
All the failures in GitLab are vsphere related, which requires a config change.  I'll restart and we should get them passing.

croissanne:
Rebase and then happy to ignore the 500s we're getting from vsphere?

achilleas-k:
Done!  All good now.

croissanne (approved):
yes

PR #4652, opened 2025-03-18
[RHEL-9.6.0] Fix osbuild-composer is not able to build RHEL-8 images since RHEL-9.5 (RHEL-83781)

thozza:
* Don't trigger EDGE CI since its tests do not run on RHEL-9.5, and they have been failing for a very long time and have become useless at this point.
* Update RHEL-8 repo configurations to contain only the RH release key 2 signed with RSA-SHA256, fixing the cross-distro building of RHEL-8 images on RHEL-9.
* Extend cross-distro.sh test case to:
  * Test actual cross-distro image builds of older GA releases for supported combinations of Enterprise Linux variants. The repositories shipped by the RPM are used, instead of the testing repo configs, which ensures that we test what we actually ship in RPMs.
  * On RHEL-10, test building of the latest GA of RHEL-9 and RHEL-8 GCE image.
  * On RHEL-9, test building of the latest GA of RHEL-8 GCE image.
  * On c10s, test building of the latest c9s GCE image.
  * GCE image has been chosen deliberately because it imports GPG keys from all repositories specified in the repo config.


thozza:
RHEL-8 image build passed on RHEL-9 https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9555723521#L1533

achilleas-k (approved):
Very nice!

LGTM

PR #4654, opened 2025-03-20
cloudapi/v2: mark md5 signature as required in package metadata

croissanne:
Technically osbuild/osbuild-composer#4564 broke the api spec by marking
a required field as non-required. Fix this by using allOf.


achilleas-k:
What would break if we made sigmd5 required in packagemetadata everywhere?

croissanne:
> What would break if we made sigmd5 required in packagemetadata everywhere?

The depsolving job apparently can't return sigmd5 as it is currently. I wouldn't mind including it somehow though.

achilleas-k (approved):

achilleas-k (approved):

PR #4661, opened 2025-03-28
[RHEL-9.6.0] Reduce CI only to RHEL-9.6 and on-prem scenario (RHEL-83781)

thozza:
SSIA + cherry-picks from `main` to make CI actually work.


achilleas-k (approved):
:heart:

PR #4662, opened 2025-03-28
[RHEL-9.5.0] Reduce CI only to RHEL-9.5 and on-prem scenario (RHEL-83780)

thozza:
SSIA + cherry-picks from main to make CI actually work.


achilleas-k (dismissed):

thozza:
I had to disable some test cases:

- which are not supported on RHEL-9.5 (the satellite regression tests)
- which are not supported on `aarch64`, but the rules didn't restrict them

And also move some tests that require KVM for virtualization from `aws` runner to `rhos-01` runner.

achilleas-k (approved):
Thanks again

PR #4664, opened 2025-03-28
Test/api.sh/vsphere: configure cloud-init via VM's extraConfig

thozza:
We used to provide the cloud-init user-data and meta-data to the VM as key/value in the VM's extraConfig as guestinfo keys. This replaces bunch of code that used to create an ISO, upload it to VSphere and attach it to the VM.

We recently started to observe issues when uploading the ISO to our testing VSphere instance in CI. This commit aims to circumvent this issue by using a different and simpler way to provide the cloud-init data to the VM.


tkoscieln (approved):
LGTM, thank you!

achilleas-k (approved):
Awesome!  ty

PR #4665, opened 2025-03-28
Test/repositories/el8*: use rhui-4 instead of rhui-3

thozza:
RHUI-3 has been EOL on RHEL-8 and it is no longer being updated and we do not take new snapshots of it. Use RHUI-4 instead. More info in [1].

It turns out, that the previously used RHUI-3 snapshot was missing amazon-libdnf-plugin package, wich is required by the RHUI client. As a result, the "weldr dot-notation" test has been failing in the GA pipeline on RHEL-8. We do not run the test case on RHEL-8 in PRs of `main` pipeline, because we test only on nightly composes and there are no new releases being developed for RHEL-8. Therefore this was manifesting only in the GA pipeline runs.

This change should fix the failures in the GA pipeline (such as https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/9550422284#L10090)

[1] https://github.com/osbuild/rpmrepo/pull/119


achilleas-k (approved):
Thanks for the full explanation.
LGTM

mvo5 (approved):

achilleas-k:
Rebasing to get fixed vsphere tests.

PR #4693, opened 2025-04-23
[RHEL-9.2] go.mod: update jwt-go to v4.5.2 fix CVE-2025-30204

diaasami:
/jira-epic COMPOSER-2510


osbuild-composer.spec
```diff
@@ -22,7 +22,7 @@ It is compatible with composer-cli and cockpit-composer clients.
 }
 
 Name:           osbuild-composer
-Release:        1%{?dist}
```
thozza, line 25: The release in the upstream spec is ignored. Please add `.1` to the version instead.

thozza (approved):

thozza:
There are multiple issues with GitLab CI, that I unfortunately don't have time to fix before my vacation and this issue needs to be fixed ASAP. The issues are:

- VMWare tests are known to fail and should be disabled.
- I probably forgot to fix another occurrence of `CI_BUILD_ID` in `aws.sh`.
- The Base tests fail on rpmrepo snapshots for c8s, which is already EOL. These should be probably updated.
- OSTreee tests are failing, but this PR could not have possibly affect them.

Given the actual change in osbuild-composer code, running unit tests should be enough to verify that things still work.

I'm admin-merging this PR.

PR #4694, opened 2025-04-23
Test/cross-distro: fix one issue and make the test more robust on "ZStream" branches

thozza:
Test/cross-distro: fix processing of repo configs 

Previously, the code would remove everything after the first dot in the
repo config filename. So 'rhel-10.0.json' -> 'rhel-10'. This means that
there was just a bunch of 'rhel-X' repo config names to compare. This is
probably a leftover from the days before we introduced dot-notation for
distro names.

Modify the test to effectivelly strip only the filename extension (the
part after the last dot, including the dot).

-----------------------------------------------------------------------------

Test/cross-distro: take the images version used to compile composer 

After the move of the repo configs to `osbuild/images`, we changed the
way to determine the all available distro repo configs by taking the
last release of 'images'. This turns out to be fragile, especially when
the test case is being run on a "rhel-x.y.0" branch or basically with
any old osbuild-composer binary. It could happen also in Nightly test
pipelines after the devel freeze.

We always compile osbuild-composer binary with debug information, so we
can determine the vendored 'images' version from the binary. We also
always ship repo configs from the same 'images' version in the RPMs.

Therefore modify the test case to check out the 'images' version that
osbuild-composer was compiled with and use repo configs from that
version.

### DIFF
--- a/test/cases/cross-distro.sh
+++ b/test/cases/cross-distro.sh
@@ -99,12 +99,19 @@ else
     echo "All installed distros are recognized by composer."
 fi
 
-# set path to all "images" repositories, do not use "main" here but the
-# last release version (this assume 200 commits since the last tag)
-git clone --depth 200 http://github.com/osbuild/images
-( cd images &&
-      git checkout "$(git describe --tags "$(git rev-list --tags --max-count=1)")"
-)
+# determine the 'osbuild/images' repository version used by the osbuild-composer
+sudo dnf install -y golang
+COMPOSER_DEPS=$(go version -m /usr/libexec/osbuild-composer/osbuild-composer)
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):

achilleas-k:
Good stuff!

thozza:
> Good stuff!

Thanks! Unfortunately, I'll have to amend the commit, because of
```
/usr/libexec/tests/osbuild-composer/cross-distro.sh: line 103: go: command not found
```
:-/

achilleas-k:
My Approve finger is ready!

thozza:
```
[2025-04-24T08:10:13+00:00] INFO: Using osbuild/images version to check repo configs: v0.137.0
```
is in the `cross-distro` job log :+1:

achilleas-k (approved):

PR #4706, opened 2025-04-30
Test/cross-distro: add version check for cross-building el8 on el9

thozza:
The issue has been fixed in 9.6 0day ZStream in v132.1, so running the test against anything older will make the test fail. This is currently the case for RHEL-9-nightly pipeline.

Let's solve this annoyance.


achilleas-k (approved):
Neat!

bcl (approved):
Ack!

PR #4709, opened 2025-05-02
cloudapi: Fix missing specs for koji ContainerResolve

bcl:
It looks like this got lost in a refactor back in commit 6e4efabf246561673fc7334c0e4fcb6157c48d1b when it used to populate the Specs list in a loop.


bcl:
@achilleas-k I stumbled across this while comparing the koji enqueue code with the regular one, seems like a bug?

achilleas-k (approved):
Oh dear.  Funny how didn't catch this until now.

Thank you!

croissanne (approved):

Issue #4705, opened 2025-04-30
Unable to override image size in cloud API when using disk customizations


achilleas-k:
https://github.com/osbuild/osbuild-composer/blob/55cd6a9df5c879e77122a2455168c46729fe03f5/internal/cloudapi/v2/imagerequest.go#L27-L39

The function above makes it impossible to set an image size when using only disk customizations, because it only takes into account the filesystem customizations, the image request, and the image type's default size.  In brew, this makes it impossible to build certain configurations, namely the Azure raw image which should be 32 GiB.  The total disk size always gets overridden by the image type's default size (64 GiB).

Long term, I think the correct fix is we should map out the full interaction between `imageOptions.Size`, `customizations.Filesystem`, and `customizations.Disk`, write it up in images in one place (in the disk package preferably), and remove any other logic from the cloud api, osbuild-composer, and the distro definitions.

osbuild-composer should be passing options and customizations to images without doing any pre-processing of the options.

PR #4711, opened 2025-05-05
cloudapi: drop ImageRequest.GetImageOptions() method

achilleas-k:
This method is not particularly useful anymore.  Its purpose was to initialise the ImageOptions from an ImageRequest with the appropriate size and partitioning mode.  However, the partitioning mode was also being set later using request.GetPartitioningMode().  More importantly, setting the size on the ImageOptions caused issues with the interaction between filesystem and partitioning customizations as well as the image request size (see #4705).  The correct thing to do here is to map the ImageRequest.Size directly onto ImageOptions.Size, without taking into account ImageType or the Blueprint Customizations.  The rest are considered when generating the manifest in images, either when preparing the Manifest() call or when generating the partition table.  This makes it easier to trace and reason about the effect of each option.  This kind of decision making in the API layer makes it difficult to maintain the logic, since it requires duplicating the decision making or, as we had now, making certain specific combinations impossible.

### DIFF
--- a/internal/cloudapi/v2/compose.go
+++ b/internal/cloudapi/v2/compose.go
@@ -14,6 +14,7 @@ import (
 	"github.com/osbuild/images/pkg/customizations/subscription"
 	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/distro"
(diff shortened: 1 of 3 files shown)


mvo5 (approved):
Thank you

croissanne (approved):

PR #4714, opened 2025-05-06
internal: drop `internal/fsnode` package

mvo5:
This package has an identical copy in `images` and is also (AFAICT) unused. Hence this commit removes it.


supakeen (approved):

achilleas-k (approved):
Very nice 🧹

PR #4716, opened 2025-05-06
tests/CI: Add runners for RHEL 9.7 and 10.1 (HMS-6278)

tkoscieln:
Add runners for RHEL 9.7 and 10.1 and remove 9.6 and 10.0 runners.

JIRA: [HMS-6278](https://issues.redhat.com/browse/HMS-6278)


.gitlab-ci.yml
```diff
@@ -72,13 +72,13 @@ init:
...
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9\.(6|7)-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10\.(0|1)-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
 
 .nightly_rules_x86_64:
   rules:
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.6-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10.0-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9\.(6|7)-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10\.(0|1)-[^ga][\S]+/ && $RUNNER =~ "/^.*(x86_64).*$/" && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
```
thozza, line 81: ditto as above

test/cases/api.sh
```diff
@@ -580,7 +580,13 @@ function sendCompose() {
                  --output "$OUTPUT" \
                  https://localhost/api/image-builder-composer/v2/compose)
 
+    if [ "$HTTPSTATUS" != "201" ]; then
+        redprint "Sending compose request failed:"
+        cat "$OUTPUT"
+    fi
+
     test "$HTTPSTATUS" = "201"
+
```
thozza, line 589: This change seems to be unrelated to the commit https://github.com/osbuild/osbuild-composer/pull/4716/commits/347f1e5258fb196eb0d4c7a691f3a983af6357e9

test/data/repositories/rhel-10.1.json
```diff
@@ -4,79 +4,55 @@
...
...ID6RLfi59B9lmAbekecnsMIFMx7qR7ZKyQb3GOuZwNYOaYFevuxusSwCHv\n4FtLDIhk+Fge+EbPdEva+VLJeMOb02gC4V/cX/oFoPkxM1A5LHjkuAM+aFLAiIRd\nNp/tAPWk1k6yc+FqkcDqOttbP4ciiXb9JPtmzTCbJD8lgH0rGp8ufyMXC9x7/dqX\nTjsiGzyvlMnrkKB4GL4DqRFl8LAR02A3846DD8CAcaxoXggL2bJCU2rgUQARAQAB\ntDVSZWQgSGF0LCBJbmMuIChhdXhpbGlhcnkga2V5IDMpIDxzZWN1cml0eUByZWRo\nYXQuY29tPokCUgQTAQgAPBYhBH5GJCWMQGU11W1vE1BU5KRaY0CzBQJiKSKeAhsD\nBQsJCAcCAyICAQYVCgkICwIEFgIDAQIeBwIXgAAKCRBQVOSkWmNAsyBfEACuTN/X\nYR+QyzeRw0pXcTvMqzNE4DKKr97hSQEwZH1/v1PEPs5O3psuVUm2iam7bqYwG+ry\nEskAgMHi8AJmY0lioQD5/LTSLTrM8UyQnU3g17DHau1NHIFTGyaW4a7xviU4C2+k\nc6X0u1CPHI1U4Q8prpNcfLsldaNYlsVZtUtYSHKPAUcswXWliW7QYjZ5tMSbu8jR\nOMOc3mZuf0fcVFNu8+XSpN7qLhRNcPv+FCNmk/wkaQfH4Pv+jVsOgHqkV3aLqJeN\nkNUnpyEKYkNqo7mNfNVWOcl+Z1KKKwSkIi3vg8maC7rODsy6IX+Y96M93sqYDQom\naaWue2gvw6thEoH4SaCrCL78mj2YFpeg1Oew4QwVcBnt68KOPfL9YyoOicNs4Vuu\nfb/vjU2ONPZAeepIKA8QxCETiryCcP43daqThvIgdbUIiWne3gae6eSj0EuUPoYe\nH5g2Lw0qdwbHIOxqp2kvN96Ii7s1DK3VyhMt/GSPCxRnDRJ8oQKJ2W/I1IT5VtiU\nzMjjq5JcYzRPzHDxfVzT9CLeU/0XQ+2OOUAiZKZ0dzSyyVn8xbpviT7iadvjlQX3\nCINaPB+d2Kxa6uFWh+ZYOLLAgZ9B8NKutUHpXN66YSfe79xFBSFWKkJ8cSIMk13/\nIfs7ApKlKCCRDpwoDqx/sjIaj1cpOfLHYjnefg==\n=UZd/\n-----END PGP PUBLIC KEY BLOCK-----"
```
thozza, line 35: Not a problem, but these changes are unnecessary :thinking:

thozza (commented):
Thank you for the PR!

The first commit contains unrelated or unneeded changes.

I have two additional questions:

1. Did you consider fixing the satellite regression test to work with the rpmrepo snapshot URLs that are missing the last backslash? I'm asking because other test cases seem to work just fine without it (at least you didn't express that they don't), so fixing the test case seems like a good idea, compared to keeping this unexpected behavior in the test and risking that we will have to spend time debugging the same thing in the future.
2. Do you plan to submit a follow-up to move the GA runners from 9.5 to 9.6 and also adding 10.0 ga runners?

.gitlab-ci.yml
```diff
@@ -72,13 +72,13 @@ init:
 
 .nightly_rules_all:
   rules:
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9.6-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10.0-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-9\.(6|7)-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "9"'
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $RUNNER =~ /[\S]+rhel-10\.(0|1)-[^ga][\S]+/ && $NIGHTLY == "true" && $RHEL_MAJOR == "10"'
```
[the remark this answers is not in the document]
tkoscieln: Oh, this is a leftover from when I started working on this PR when the plan was to have 9.6/10.0 and 9.7/10.1 running at the same time. Will delete it.

test/data/repositories/rhel-10.1.json
tkoscieln: This was done in order to fix the [API tests](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/1816661625). The issue was caused by [this line](https://github.com/osbuild/osbuild-composer/blob/d08cedf73ba47148e074dce4a9c995236ded18f2/test/cases/api/aws.sh#L73) in the test, so this maybe circles back to your first question if the test case should be changed or the resource it uses.

test/cases/api.sh
tkoscieln: Hmm yes, that's true. Should I extract it to a seperate commit? This was partly debug thing so it can be dropped as well, but it improves the logs?

tkoscieln:
> 1. Did you consider fixing the satellite regression test to work with the rpmrepo snapshot URLs that are missing the last backslash? I'm asking because other test cases seem to work just fine without it (at least you didn't express that they don't), so fixing the test case seems like a good idea, compared to keeping this unexpected behavior in the test and risking that we will have to spend time debugging the same thing in the future.

Well since all of the other distros use the slash in the base_url, I thought I will just edit these to contain them as well. But on the other hand yes, I do not like the way the test handles the url creation. So we can just leave it as is or we can add a condition into the test (or edit the url creation itself).

> 2. Do you plan to submit a follow-up to move the GA runners from 9.5 to 9.6 and also adding 10.0 ga runners?

Yes, I plan to do that asap.

test/cases/api.sh
thozza: A separate commit would be great, thank you.

tkoscieln: Created a separate commit df4ee8360a3de44f75490ae6beae17dd85cf71ae

thozza (approved):
LGTM

PR #4717, opened 2025-05-07
Add job delete support

bcl:
This pull request includes:


internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -726,3 +726,48 @@ func (q *fsJobQueue) AllRootJobIDs() ([]uuid.UUID, error) {
...
+	if len(j.Dependents) > 0 {
+		return q.db.Write(id.String(), j)
+	}
+	// Recursively delete the dependencies of this job
+	for _, dj := range j.Dependencies {
+		_ = q.deleteJob(id, dj)
+	}
+
+	return q.db.Delete(id.String())
+}
```
lzap, line 773: This should be in a single transaction, but I guess the DB API does not provide this abstraction level.

internal/jobqueue/fsjobqueue/fsjobqueue.go
```diff
@@ -726,3 +726,48 @@ func (q *fsJobQueue) AllRootJobIDs() ([]uuid.UUID, error) {
...
+	j.Dependents = deps
+
+	// This job can only be deleted when the Dependents list is empty
+	// Otherwise it needs to be saved with the new Dependents list
+	if len(j.Dependents) > 0 {
+		return q.db.Write(id.String(), j)
+	}
+	// Recursively delete the dependencies of this job
+	for _, dj := range j.Dependencies {
+		_ = q.deleteJob(id, dj)
```
lzap, line 769: I suggest to use the new feature `errors.Join`: https://pkg.go.dev/errors#Join

internal/jobqueue/jobqueuetest/jobqueuetest.go
```diff
@@ -814,5 +819,107 @@ func testFail(t *testing.T, q jobqueue.JobQueue) {
...
 		require.Greater(t, endTime, tmr)
 	}
+}
+
+// sortUUIDs is a helper to sort a list of UUIDs
+func sortUUIDs(entries []uuid.UUID) {
+	sort.Slice(entries, func(i, j int) bool {
+		return entries[i].String() < entries[j].String()
+	})
+}
```
lzap, line 829: I would probably define the `func(i, j int)` as a helper and use it on multiple callsites but whatever, this works too :-D

internal/jsondb/db.go
```diff
@@ -79,6 +79,14 @@ func (db *JSONDatabase) List() ([]string, error) {
 	return names, nil
 }
 
+// Delete will delete the file from the database
+func (db *JSONDatabase) Delete(name string) error {
+	if len(name) == 0 {
+		return fmt.Errorf("missing jsondb document name")
+	}
+	return os.Remove(filepath.Join(db.dir, name+".json"))
```
lzap, line 87: This is vulnerable to path traversal, however, I only see UUIDs passed in. So I am assuming we do not pass anything else in here. We are good.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -904,9 +913,148 @@ func (q *DBJobQueue) jobDependents(ctx context.Context, conn connection, id uuid
 	return dependents, nil
 }
 
+// deleteJob removes the job from the database
+// the CASCADE constraint will also delete any entries from the job_dependencies table
+func (q *DBJobQueue) deleteJob(jobID uuid.UUID) error {
+	conn, err := q.pool.Acquire(context.Background())
```
lzap, line 919: This API is a mixed bag when it comes to context passing, some methods do support it, some don't. Thing is, I am working on migrating from `logrus` to `log/slog` and I will be refactoring everything to make use of context passing. Would you mind adding `ctx context.Context` function argument for all newly introduced methods already here so I have less work later? Just pass the context down the stack to all helpers, the benefit will be that all the logging coming from the SQL driver will have correlation id so we can search more easily for detailed logs.

internal/jsondb/db.go
bcl: Right, and with it appending .json it's not terribly useful.

internal/jobqueue/fsjobqueue/fsjobqueue.go
bcl: It's ignoring the errors on purpose, there's no need to capture them.

internal/jobqueue/fsjobqueue/fsjobqueue.go
bcl: This is a filesystem based queue, so there is no concept of transactions.

bcl:
Note that Snyk is wrong, no directory traversal is possible since it uses uuid.Parse() to make sure the id is a uuid not some arbitrary string.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -904,9 +913,148 @@ func (q *DBJobQueue) jobDependents(ctx context.Context, conn connection, id uuid
...
+func (q *DBJobQueue) DeleteJob(ctx context.Context, id uuid.UUID) error {
+	conn, err := q.pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+
+	// Start it off with an empty parent
+	return q.deleteJobs(ctx, conn, uuid.UUID{}, id)
+}
```
lzap, line 1025: Consider creating an explicit transaction here just like in `Enqueue`.

lzap (approved):
Thanks for adding those contexts.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
bcl: I don't think I can. The functions all take connections or acquire their own from the pool, etc. But I agree that it needs something to lock access. Delete won't run unless the job is finished, so the only thing to worry about is other delete operations so a lock should work.

bcl: How is the mutex going to break? You'd have to call DeleteJob from inside DeleteJob to double lock it. I admit it is a bit of a hack, but so is making complex new interface type.

bcl: If someone knows how to do this with pure SQL I'd be open to that. My SQL-foo is really rusty.

lzap: I am not sure what exactly do you mean by pure SQL, all I am asking is to avoid mutex and just use a regular SQL transaction across multiple helper functions. Feel free to ignore, I can do a follow-up.

bcl: So it ends up there is already an interface for using conn or tx but it was buried in the middle of the code. I added Exec to it and I think this new version is one you'll like. Updated PR incoming...

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -849,13 +867,6 @@ func (q *DBJobQueue) DeleteWorker(workerID uuid.UUID) error {
 	return nil
 }
 
-// connection unifies pgxpool.Conn and pgx.Tx interfaces
-// Some methods don't care whether they run queries on a raw connection,
-// or in a transaction. This interface thus abstracts this concept.
-type connection interface {
-	Query(ctx context.Context, sql string, args ...interface{}) (pgx.Rows, error)
-}
-
```
lzap, line 858: Ha, little guy. I thing it would not hurt search-n-replacing all connection arguments to use this interface so next time someone copy-paste a method it is already there and ready :-) Great find.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -904,9 +915,152 @@ func (q *DBJobQueue) jobDependents(ctx context.Context, conn connection, id uuid
...
+// no dependency loops. Shared Dependents are ok, but a job cannot have a dependency
+// on any of its parents (this should never happen).
+func (q *DBJobQueue) DeleteJob(ctx context.Context, id uuid.UUID) error {
+	conn, err := q.pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+
+	tx, err := conn.Begin(context.Background())
```
lzap, line 986: Few overlooks, this will carry the context over into the transaction so we can cancel it or put a deadline on it etc.

```suggestion
	tx, err := conn.Begin(ctx)
```

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -904,9 +915,152 @@ func (q *DBJobQueue) jobDependents(ctx context.Context, conn connection, id uuid
...
+	if err != nil {
+		return fmt.Errorf("error starting database transaction: %w", err)
+	}
+	defer func() {
+		err := tx.Rollback(context.Background())
```
lzap, line 991:
```suggestion
		err := tx.Rollback(ctx)
```

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -904,9 +915,152 @@ func (q *DBJobQueue) jobDependents(ctx context.Context, conn connection, id uuid
...
+			q.logger.Error(err, "Error rolling back delete transaction")
+		}
+	}()
+
+	// Start it off with an empty parent
+	err = q.deleteJobs(ctx, tx, uuid.UUID{}, id)
+	if err != nil {
+		return fmt.Errorf("Error deleting job %s: %w", id.String(), err)
+	}
+	err = tx.Commit(context.Background())
```
lzap, line 1002:
```suggestion
	err = tx.Commit(ctx)
```

lzap (dismissed):
Yeah this is great! Small nitpick, but this is ready.

Someone else please review!

thozza (changes requested):
Sorry, I misclicked. I wanted to request changes, instead of approving :innocent:

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -960,8 +969,98 @@ func (q *DBJobQueue) AllRootJobIDs(ctx context.Context) (rootJobs []uuid.UUID, e
...
+// missing dependencies are ignored, it deletes as much as it can.
+// A missing parent (the first call) will be returned as an error
+func (q *DBJobQueue) deleteJobs(ctx context.Context, conn connection, parent, id uuid.UUID) error {
+	// Delete parent:id dependencies if they exist
+	if len(parent.String()) > 0 {
+		err := q.deleteJobDependencies(ctx, conn, parent, id)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
bcl: It essentially is, the first call is with parent set to an empty UUID so it skips this.

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -134,6 +134,62 @@ paths:
...
+        - in: path
+          name: id
+          schema:
+            type: string
+            format: uuid
+            example: '123e4567-e89b-12d3-a456-426655440000'
+          required: true
+          description: ID of compose to delete
+      description: |-
+        Delete a compose and all of its independent jobs.
```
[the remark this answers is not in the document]
bcl: Maybe? I like having it there because it makes it obvious what's going on.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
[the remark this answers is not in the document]
thozza: While this seems to be true, IMO it would make sense to keep this method's logic in sync with `fsjobqueue`. The current flow suggests that 
> IMHO, we should not delete a job's dependencies if we can't delete the job itself because something else depends on it.

can happen.

internal/cloudapi/v2/openapi.v2.yml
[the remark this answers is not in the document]
thozza: While I understand that, this does not comply with how `image-builder-crc` handles `DELETE` operation on objects, see https://github.com/osbuild/image-builder-crc/blob/main/internal/v1/api.yaml.

From what I was able to research, in REST API, the URL should identify the resource on which you can do CRUD (Create, Read, Update, Delete) operations, not the operation itself.

internal/jobqueue/jobqueuetest/jobqueuetest.go
```diff
@@ -814,5 +819,107 @@ func testFail(t *testing.T, q jobqueue.JobQueue) {
...
+		}
+	})
+
+	// root with 2 jobs depending on another (simulates Koji jobs)
+	t.Run("koji job simulation", func(t *testing.T) {
+		kojiOSTree := pushTestJob(t, q, "ostree", nil, nil, "KojiJob")
+		kojiDepsolve := pushTestJob(t, q, "depsolve", nil, nil, "KojiJob")
+		kojiManifest := pushTestJob(t, q, "manifest", nil, []uuid.UUID{kojiOSTree, kojiDepsolve}, "KojiJob")
+		kojiInit := pushTestJob(t, q, "init", nil, nil, "KojiJob")
+		kojiRoot := pushTestJob(t, q, "final", nil, []uuid.UUID{kojiInit, kojiManifest, kojiDepsolve}, "KojiJob")
```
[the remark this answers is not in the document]
bcl: I swear every time I read this code it looks different. Fixed.

lzap (approved):

internal/worker/server.go
```diff
@@ -349,8 +349,53 @@ func (s *Server) JobDependencyChainErrors(id uuid.UUID) (*clienterrors.Error, er
...
+		if err != nil {
+			continue
+		}
+
+		// Is there a job with this UUID?
+		if _, _, _, _, err := s.jobs.Job(id); err != nil {
+			// No associated job, it is safe to remove the unused artifact directory
+			// and everything under it, and the ComposeRequest (if it exists)
+			_ = os.Remove(path.Join(s.config.ArtifactsDir, "ComposeRequest", id.String()+".json"))
+			err = os.RemoveAll(path.Join(s.config.ArtifactsDir, id.String()))
```
thozza, line 378: Snyk complains about `Path Traversal vulnerability` on these two lines.

I think that this is a false positive. The ArtifactsDir comes from the configuration. And the `id` coming from the request must be a valid UUID, and it must correspond to a valid job ID. So there is, IMHO, no space for traversing paths at all.

PR #4719, opened 2025-05-09
Update check-snapshots to include an update functionality

PicklesMcDougan:
Added a new functionality to the tools/check-snapshots utility to allow automatic updates of any files with newer results.

`./tools/check-snapshots -u .`


achilleas-k:
Hey.  Thanks for the PR.  I haven't read through it in detail yet.  I invited you to one of the org teams with write access so that your PR checks run without needing approval.

Could you please clean up the git history a bit?  Please have a look at our commit message guidelines: https://osbuild.org/docs/developer-guide/general/workflow/#preferred-commit-message-format

Some of the commit message summaries are a bit too long.  Also some commits could be squashed for cleaner history.  The `update` script for example is added and removed in the PR, so it shouldn't appear at all.  The last commit message ("final version ...") isn't very informative.  It should describe what changed in the commit itself.

bcl:
Also, please separate changes to the tool from changes to the json files into their own commits.

bcl:
Err. And now I'm confused, this PR is opened against rhel-9.2.0 not the main branch. Was that on purpose?

PicklesMcDougan:
@bcl My understanding was that the more recent branches had a method for updating the snapshot urls. People were out on leave so I was just working on how to fix the failing tests for the 9.2 branch, since 9.4 was working. I am unaware if the newer branches are working simply because the snapshots are still valid, or if there was another method to update those repo urls. Outside of whatever I've done to upset Schultzbot, I'm happy to update or cancel this request in order to do whatever would be best

bcl:
I'm not seeing the update option in the current version of the script on main, did you add that yourself or copy it from somewhere? I'm likely missing some context here related to the older branches though.

PicklesMcDougan:
@achilleas-k I see a pr where you [manually triggered schultzbot](https://github.com/osbuild/rpmrepo/actions/runs/14706682982) to update the files for you.

Should I even work on this check-snapshots -u feature or is there a way I can trigger Schultz to update these files for me on older branches like 9.2?

achilleas-k:
I'm copying some chat messages I sent today here for better visibility:

The current workflow doesn't work against older branches.  But there might be room for improvement here.

The current action in the rpmrepo repository does two things:
1. Creates snapshot mirrors of all the configured repositories
2. Opens PRs against all our projects to update the snapshot URLs in the test repo configs

I suppose what you need to do for an older branch is only 2.  We probably have recent enough repository snapshots for the distros we're talking about.  You can get a full list of snapshot names at https://rpmrepo.osbuild.org/v2/enumerate/ (RH VPN connection required).

Usually, when we need to do an snapshot update for a specific branch or repo, we just do it manually.  It's a little tedious, but not egregious.

As for improvements: I have considered, and I think it might be nice to have, making a GH action in each project repository to update all snapshots to the latest one that it finds in the rpmrepo.osbuild.org.  There's a bit of an issue with the current action that performs step 2. above that misses some repo updates somehow.  I haven't looked into it, but I think @croissanne has a good idea about what's wrong.  Having an action that looks at the existing ones and opens a PR in a project to update things would be a nice way to make sure things are up-to-date when we want them and it would be a good way to do it on branches too.

achilleas-k:
The linter failed to pull the base runner container of all things.

PR #4723, opened 2025-05-14
templates/packer: set wanted-by to cloud-init.target

croissanne:
The `cloud-init.target` in 9.6 has `After=multi-user.target` in its unit config. The worker initialization service was set to run before `multi-user.target`, but after `cloud-final.service`. This created an impossible situation and systemd just disabling the initialization service.

So this changes:
`multi-user.target -> worker-*.service -> cloud-final.service -> multi-user.target`
to
`cloud-init.target -> worker-*.service -> cloud-final.service -> multi-user.target`.

Thus resolving the loop.

---

on the graph you can see the worker-initialization -> cloud-final -> multi target -> worker-initialization loop.

worker-cycle

---

fixed loop
worker-cycle-fixed


ondrejbudai (approved):

kingsleyzissou (approved):

lzap (approved):

lzap:
Shall we merge to unblock the queue?

PR #4727, opened 2025-05-21
weldr: add memory stats in status API

lzap:
It was brought to our attention that the API is consuming memory in a way that is suspicious. Since Linux memory testing tools can get deceptive results, the definitive answer can be shown via the Go runtime. This patch adds an opt-in feature of the status call that shows some key heap memory and GC cycles stats directly from the runtime. I also added a benchmark which I used to measure the performance of such thing. Before:

```
goos: linux
goarch: amd64
pkg: github.com/osbuild/osbuild-composer/internal/weldr
cpu: Intel Xeon Processor (SierraForest)
BenchmarkStatus
BenchmarkStatus-10    	   87990	     13373 ns/op	   11358 B/op	     120 allocs/op
PASS
ok  	github.com/osbuild/osbuild-composer/internal/weldr	1.506s
```

After:

```
goos: linux
goarch: amd64
pkg: github.com/osbuild/osbuild-composer/internal/weldr
cpu: Intel Xeon Processor (SierraForest)
BenchmarkStatus
BenchmarkStatus-10    	   45235	     26846 ns/op	   11623 B/op	     134 allocs/op
PASS
ok  	github.com/osbuild/osbuild-composer/internal/weldr	1.678s
```

This is indeed 50% slower and since this status call is often called every second I am making this opt-in only via an env variable.


lzap:
@mvo5 had a good idea to allow the possibility to run the GC, so I am amending one more small change. The status API endpoint now roll a pseudo dice and if a number between (0.0, 1.0] is lower than number defined via `OSBUILD_COMPOSER_GC_STATUS_RATE` the GC will run. Setting this variable to 1 will cause the GC to be executed on every request, 0.1 about 10% requests and 0 or not setting it all (the default behavior) will not run GC at all.

lzap:
I have found the problem, this PR got way too much and as it turns out those stats and calling GC was not useful at all during analysis of the problem. I will break few things from this PR into a new one (upgrade to v22 systemd library) and I will separate the actual problem too (unclosed pipe).

Issue #4726, opened 2025-05-20
Memory leak when polling API endpoints


stefbenoist:
**Describe the bug**
Hi, we observe memory usage increasing dramatically when invoking **/api/status** or **api/v1/compose/queue** endpoints for respectively liveness probe and polling compose queue (1s period for both) concerns.

**Environment**
 * OS version (`/etc/os-release` and `/etc/redhat-release`):
 
```
# cat /etc/os-release 
Red Hat Enterprise Linux release 9.5 (Plow)
[root@localhost ~]# cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="9.5 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.5"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.5 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://issues.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.5
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.5"

# cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.5 (Plow)
```

 * osbuild-composer version (`rpm -qi osbuild-composer)`
```
# rpm -qi osbuild-composer
Name        : osbuild-composer
Version     : 132.2
Release     : 1.el9_6
Architecture: x86_64
Install Date: Tue May 20 10:58:57 2025
Group       : Unspecified
Size        : 19031
License     : Apache-2.0
Signature   : RSA/SHA256, Wed Apr 23 03:09:40 2025, Key ID 199e2f91fd431d51
Source RPM  : osbuild-composer-132.2-1.el9_6.src.rpm
Build Date  : Tue Apr 22 12:58:39 2025
Build Host  : x86-64-04.build.eng.rdu2.redhat.com
[...]

PR #4728, opened 2025-05-22
common: fix unclosed logrus logging pipes

lzap:
It looks like both CloudAPI and WeldrAPI consume memory, process can go up to several gigabytes pretty quickly just by running a simple script. This was reported few days ago: https://github.com/osbuild/osbuild-composer/issues/4726

```
while sleep 0.001; do
    curl --unix-socket /run/weldr/api.socket -XGET http://localhost/api/
    curl --unix-socket /run/cloudapi/api.socket -XGET http://localhost/api/
done
```

I was able to quickly reproduce on Fedora 42, extracting core with `gcore PID` and running `dlv core /usr/libexec/osbuild-composer/osbuild-composer core.PID` revealed that there are tens of thousands of blocked goroutines all with the following stack:

```
(dlv) goroutine 70207
Switched from 67162 to 70207 (thread 885812)

(dlv) stack
0  0x00005607664f026e in runtime.gopark
   at /usr/lib/golang/src/runtime/proc.go:436
1  0x00005607664cf2b7 in runtime.selectgo
   at /usr/lib/golang/src/runtime/select.go:351
2  0x000056076653ec45 in io.(*pipe).read
   at /usr/lib/golang/src/io/pipe.go:57
3  0x000056076653f35a in io.(*PipeReader).Read
   at /usr/lib/golang/src/io/pipe.go:134
4  0x0000560766712c1e in bufio.(*Scanner).Scan
   at /usr/lib/golang/src/bufio/scan.go:219
5  0x00005607669576fd in github.com/sirupsen/logrus.(*Entry).writerScanner
   at /usr/src/debug/osbuild-composer-142-1.20250522git99f5fec.fc42.x86_64/vendor/github.com/sirupsen/logrus/writer.go:86
6  0x00005607669575a8 in github.com/sirupsen/logrus.(*Entry).WriterLevel.gowrap1
   at /usr/src/debug/osbuild-composer-142-1.20250522git99f5fec.fc42.x86_64/vendor/github.com/sirupsen/logrus/writer.go:57
7  0x00005607664f80a1 in runtime.goexit
   at /usr/lib/golang/src/runtime/asm_amd64.s:1700
```

There is a `logrus` logger [method called](https://github.com/sirupsen/logrus/blob/master/writer.go#L57) `Write` and `WriteLevel` which create a new logging entry, create a PIPE and spawn a goroutine that is reading from that PIPE. The caller is expected to close the PIPE writer which was not happening in both `main.go` files.

But that was not the culprit of the problem, a new goroutine was created every single HTTP request. I tested both `go-systemd` library as well as `httprouter` library but they were scaling well and no problems there. Until I realized that both WeldrAPI and CloudAPI use a common logging proxy `EchoLogrusLogger` that had a `Write` method implemented as follows:

```golang
func (l *EchoLogrusLogger) Write(p []byte) (n int, err error) {
	return l.Logger.WithContext(l.Ctx).Writer().Write(p)
}
```

The `Writer()` call creates a new PIPE and a goroutine but is never closed, this method is used from both WeldrAPI (through the standard `log` package) as well as CloudAPI (through Echo logger) at least once per request. The obvious fix and the one we can use if we chose to backport this fix is:

```golang
func (l *EchoLogrusLogger) Write(p []byte) (n int, err error) {
    w := l.Logger.WithContext(l.Ctx).Writer()
    defer w.Close()
    return w.Write(p)
}
```

But this PR aims for better solution that prevents from spawning a new goroutine each `Write` call in the first place. Writer is created only once per logger, that is once for the WeldrAPI since it uses a singleton and once per request for CloudAPI, much better than every single event. Cheers!


cmd/osbuild-composer/main.go
```diff
@@ -80,10 +80,12 @@ func main() {
 	}
 
 	logrus.Info("Loaded configuration:")
-	err = DumpConfig(*config, logrus.StandardLogger().WriterLevel(logrus.InfoLevel))
+	dumpWriter := logrus.StandardLogger().WriterLevel(logrus.DebugLevel)
+	err = DumpConfig(*config, dumpWriter)
```
mvo5, line 84: I guess it does not matter, but would `defer dumpWiter.Close()` make sense here (and in the other places where close is used)? I.e. similar to how there is always a "defer f.Close()" for other resources?

mvo5 (commented):
A bit of a meta-question - AIUI this was an important issue but it looks like there is no test for the fix, is there a way to test for this?

internal/common/echo_logrus.go
```diff
@@ -39,16 +58,27 @@ func toEchoLevel(level logrus.Level) lslog.Lvl {
...
 
+func (l *EchoLogrusLogger) Close() {
+	if l.writer != nil {
+		l.writer.Close()
+		l.writer = nil
+	}
+}
+
+func (l *EchoLogrusLogger) ctxLogger() *logrus.Entry {
+	return l.ctxLogger()
```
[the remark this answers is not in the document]
lzap: Good catch yeah I was doing a last minute improvements and I was planning on storing the logger in the instance directly with context, but `logrus` has such a weird API because `WithContext` does not return a logger but an event. I will remove this helper method, it does not make the code any more clear.

cmd/osbuild-composer/main.go
lzap: I want to close and unblock the goroutine as soon as possible, there is no reason to hold it off until program exits. It is either this, or I can create an anonymous function with `defer` which is IMHO an overkill.

lzap:
> A bit of a meta-question - AIUI this was an important issue but it looks like there is no test for the fix, is there a way to test for this?

Good point, I could add a test that just logs some data. The test would not test the fact we close the PIPE in all places tho. And I could do that if you ask me to, however, logrus is going away, this sprint I have a ticket to finish with logrus>slog refactoring. All logging code will be extracted into the common `logging` library which has extensive tests through and through.

https://github.com/osbuild/logging

In fact I already started the work.

bcl (approved):
Looks good to me!

lzap:
Expect the first logging refactoring this week, I will follow the CRC case and do it in two waves - first I will use a logrus proxy and remove this dependency completely from the project, next I will refactor all log calls to make use of the cleaner `slog` API.

thozza (approved):

lzap:
Rebased, I think this can go in. Tests failures were unrelated.

lzap:
Retrying...

PR #4738, opened 2025-06-02
osbuild-store-dump: use `distrofactor` instead of importing fedora

mvo5:
This commit removes the direct use of `pkg/distro/fedora` and uses `distrofactory` instead. This is the more generic interface, it also helps with PR#1563 as `pkg/distro/fedora` goes away.

As an idle wondering, the `fedoraID` in the tool is fedora-37 and there is no mention in the README what this tool does and no reference from e.g. a gitlab/github so I assume if it compiles its fine, I don't know how to meaningfully test it.


bcl:
Typo in the title `distrofactory` missing y :)

thozza (approved):
LGTM, thanks!

PR #4740, opened 2025-06-03
cloudapi: Disk customization tweaks (adding swap), validation fixes, and tests

achilleas-k:
This PR takes care of a few issues related to the Disk customizations in the Cloud API:
1. Swap partition and logical volume support is now added.
2. The `fs_type` was not marked as required but in the Blueprint schema it is.  It is now also required in the Cloud API.
3. Tests have been added to verify that the schema and blueprint validation behave as expected.

----

This pull request includes:


achilleas-k:
Marked as draft while I explore an alternative.

internal/cloudapi/v2/compose.go
```diff
@@ -1369,9 +1365,9 @@ func convertDiskCustomizations(disk *Disk) (*blueprint.DiskCustomization, error)
 				PartType: common.DerefOrDefault(fs.PartType),
 				MinSize:  fsSize,
 				FilesystemTypedCustomization: blueprint.FilesystemTypedCustomization{
-					Mountpoint: fs.Mountpoint,
```
mvo5, line 1372: Pardon my ignorance here - this is now a pointer because it is no longer marked "required" in the openapi spec so that auto-generated code makes it a pointer?

achilleas-k: Precisely

mvo5 (approved):
This looks good to me, thanks for adding the tests. My review should probably only count +0.75 as I'm not super familiar with openapi and composer.

FTR, I really hope we will soon be in a place where blueprint changes are mostly "automatic" for the downstreams like composer (I think we all share this goal :)

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -2626,11 +2626,10 @@ components:
     FilesystemTyped:
       type: object
       required:
-        - mountpoint
```
bcl, line 2687: Should there be a test for mountpoint added to the non-swap types in the code since it's now possible to pass a non-swap type without setting the mountpoint.

thozza: This case (not the test) is covered by the `bpDisk.Validate()` call when converting the Disk Customizations from the request to BP customization.

So a test case should be added :+1:

achilleas-k: Amended the last commit to add more tests for the mountpoint vs swap cases.

supakeen (approved):

PR #4743, opened 2025-06-05
Update snapshots to 20250605

schutzbot:
Results of the snapshot jobs:
Job(s) succeeded: 179
Job(s) failed: 0

If these are false, rebuild the enumerate cache manually:
Enumerate cache job succeeded: true
Check snapshot succeeded: true

Workflow run: https://github.com/osbuild/rpmrepo/actions/runs/15469185203

---

<details>
<summary>update_schutzfile.py output</summary>

</details>


achilleas-k (approved):

bcl (approved):
Looks good to me, and it will hopefully fix the selinux-policy failures we're seeing with the RHEL 10.1 tests

thozza:
> Looks good to me, and it will hopefully fix the selinux-policy failures we're seeing with the RHEL 10.1 tests

Yes, this is it. I just looked closely at the problem and the cause is that the osbuild version that we pinned in some PRs had 10.1 RPMs built with the `20250605`, so using an older snapshot for SUT host installation causes the issue, because those repos don't have a new enough selinux-policy.

PR #4745, opened 2025-06-10
Stop testing on Fedora 40 - Start testing on Fedora 42

achilleas-k:
SSIA


lzap:
There is quite a bit of `fedora-40` or `f40` in code codebase, not sure how relevant this is. Just double checking.

achilleas-k:
Dropped `test/data/manifests/fedora-40.json` as well.

lzap (dismissed):
Apologies, I used an AI to do that grep. Guess it is not good in grepping codebases :-D There is a conflict now, sorry.

mvo5 (dismissed):

supakeen:
> This PR is as 🟢 as it gets... 😕 Edge CI is known to be failing. And the Installer tests started to fail on c10s and el10.1 - the VM does not boot in time / ssh is not reachable. We need to look at this, but not as part of this PR...

The installer test failures are known (kernel bug).

thozza:
FTR, the kernel bug is https://issues.redhat.com/browse/RHEL-96222

PR #4746, opened 2025-06-11
many: remove jobsite code

croissanne:
This was replaced by `cmd/osbuild-worker-executor`.


kingsleyzissou (approved):

achilleas-k (approved):
🟥🟥🟥🟥🟥🟥🟥

supakeen (approved):

thozza:
There is an issue with the el10 RPMs. Looks like a mismatch of the rpmrepo snapshot version we use for building RPMs and those that we then install the SUT from. I didn't look deeper into the problem.

PR #4750, opened 2025-06-13
cloudapi: Use error 400 for cloudapi delete errors

bcl:
The most common one is going to be trying to delete a compose before it is finished, and that is certainly not worthy of a 500 server error since it is a temporary state.

This pull request includes:


supakeen (approved):
I'm not *that* much of a stickler for status codes and this is fine but I can't help but mention that I feel that `http.StatusConflict` (409) would be more appropriate since if this is 'because a compose is in progress' resubmitting the request at an appropriate time *would* work, while with a 400 resubmitting the same request would *not* work.

```
The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request. 
```

bcl:
Good point. There are likely a number of places where we could expand our use of error codes :) For now let's just stick with what we've been using so far.

supakeen (approved):

PR #4753, opened 2025-06-18
templates/packer: fix installing rpms from copr

croissanne:
There are now 2 colons present, one in the package name, one before the comment.

### DIFF
--- a/templates/packer/ansible/roles/common/tasks/packages.yml
+++ b/templates/packer/ansible/roles/common/tasks/packages.yml
@@ -127,7 +127,7 @@
   shell: |
     dnf copr enable -y @osbuild/osbuild-composer
     COMPOSER_COMMIT_SHORT=$(echo {{ COMPOSER_COMMIT }} | head -c 9)
-    COMPOSER_RPMS=$(dnf search -q --showduplicates osbuild-composer | grep -E "osbuild-composer-worker-[0-9]+.*g$COMPOSER_COMMIT_SHORT.*{{ ansible_architecture }}" | cut -f 1 -d ':')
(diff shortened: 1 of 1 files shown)


templates/packer/ansible/roles/common/tasks/packages.yml
```diff
@@ -127,7 +127,7 @@
   shell: |
     dnf copr enable -y @osbuild/osbuild-composer
     COMPOSER_COMMIT_SHORT=$(echo {{ COMPOSER_COMMIT }} | head -c 9)
-    COMPOSER_RPMS=$(dnf search -q --showduplicates osbuild-composer | grep -E "osbuild-composer-worker-[0-9]+.*g$COMPOSER_COMMIT_SHORT.*{{ ansible_architecture }}" | cut -f 1 -d ':')
+    COMPOSER_RPMS=$(dnf search -q --showduplicates osbuild-composer | grep -E "osbuild-composer-worker-[0-9]+.*g$COMPOSER_COMMIT_SHORT.*{{ ansible_architecture }}" | rev | cut -f 2- -d ':' | rev)
```
thozza: Nitpick: `cut -f 1-2 -d ':'` would also do the job and would be simpler...

thozza (commented):
Thanks for the fix. The expression can be a bit simplified, which I would preder to make it easier to read.

BTW, the first colon is not in the package name, but it is a delimiter of the `epoch` and the `version`.

templates/packer/ansible/roles/common/tasks/packages.yml
croissanne: ye but if they introduce another colon in the rpm name we'd miss it. But point taken.

thozza: That's highly unlikely, because the first colon in the name is actually in the version part when `epoch` is specified. However, I understand that this change is necessary, as the epoch was previously omitted, resulting in only a single colon. It is common to omit the epoch if it is zero.

kingsleyzissou (approved):

PR #4758, opened 2025-06-25
ci: introduce gobump gha

lzap:
Since gobump proven to be useful in `images`, I am proposing it for composer as well. We would drop other dependency bots and only use gobump. It bumps weekly.

This can only be tested after GHA exists and then it can be triggered against `upstream` branch manually to fine-tune it. It is the exact copy from `images` job: https://github.com/osbuild/images/blob/main/.github/workflows/gobump.yml

### DIFF
--- a/.github/workflows/gobump.yml
+++ b/.github/workflows/gobump.yml
@@ -1,0 +1,22 @@
+---
+name: "Updates Go dependencies via gobump"
+
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  schedule:
+    # Every Sunday at 15:00
+    - cron: "0 15 * * 0"
+
+jobs:
+  bump-deps-ubuntu:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):

achilleas-k:
Let's make a reusable GH action out of this at some point instead of having the same block of commands across all the repos we're going to end up using it.

lzap:
You know what, let me try the reusable action to see if that works. I need to get something merged in order to trigger the action manually against a PR tho.

achilleas-k:
> There already is one, however, it could not be used since images needs to be built and tested on Fedora. https://github.com/lzap/gobump/blob/main/action.yml
> 
> I am assuming the same for composer? If that is not the case, I can use the reusable action instead which runs on ubuntu directly.

I think it would be fine to run this action on Ubuntu for both projects.

croissanne (approved):

PR #4759, opened 2025-06-25
[RHEL 8.10] go.mod: bump osbuild/images to v0.40.1 [RHEL-95416]

achilleas-k:
Fixes the size of the ISO rootfs to fit the anaconda tree for the edge-installer.

See osbuild/images#1594

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -34,18 +34,8 @@ jobs:
 - job: copr_build
   trigger: pull_request
   targets: &build_targets
-    - centos-stream-8-aarch64
-    - centos-stream-8-x86_64
-    - centos-stream-9-aarch64
-    - centos-stream-9-x86_64
-    - fedora-all-aarch64
-    - fedora-all-s390x
-    - fedora-all-ppc64le
-    - fedora-all
     - rhel-8-aarch64
     - rhel-8-x86_64
-    - rhel-9-aarch64
-    - rhel-9-x86_64
 - job: copr_build
   trigger: commit
   branch: main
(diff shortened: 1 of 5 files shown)


achilleas-k:
We're going to need more test setup fiddling

mvo5 (dismissed):

thozza (commented):
Getting the CI work on the `rhel-X.Y.0` branches is always soooo painful :sob: 

My random comments/thoughts:

- Are changes to `tools/koji-compose.py` needed when Koji test cases are not run as part of CI? We do not need to test Koji for the 8.10 update at all.
- The same applies to the Cloud API tests (`api.sh`) - we don't need to keep testing them.
- The triggering of nightly tests in `.github/workflows/trigger-gitlab.yml` needs to be removed.
- In general, I found it easier to read to have a separate PR to reduce CI and get it working on 8.10, and then have a new PR that does the required changes. This way, we will verify if the update works correctly, as now you risk losing changes to CI that are necessary due to the images update, among other changes.

achilleas-k:
> * Are changes to `tools/koji-compose.py` needed when Koji test cases are not run as part of CI? We do not need to test Koji for the 8.10 update at all.

Not really no.  But I had to update the Lint job runner, which updated the linter, which caught issues with the old version of the script.  So I decided it's simpler to pull in the new version of the script to make the linter happy rather than disable the check.  If the changes were more substantial I would have disabled it, but it's only a few lines.

> * The same applies to the Cloud API tests (`api.sh`) - we don't need to keep testing them.

True.  Those I'll disable.

> * The triggering of nightly tests in `.github/workflows/trigger-gitlab.yml` needs to be removed.

Yup.  On it.

> * In general, I found it easier to read to have a separate PR to reduce CI and get it working on 8.10, and then have a new PR that does the required changes. This way, we will verify if the update works correctly, as now you risk losing changes to CI that are necessary due to the images update, among other changes.

That's a good idea.  I'll start a new PR for the CI reduction.

PR #4761, opened 2025-06-26
fix: update gobump GHA

lzap:
Updates gobump GHA for this repo.


lzap:
And it works! :-)

https://github.com/osbuild/osbuild-composer/pull/4762

.github/workflows/gobump.yml
```diff
@@ -15,8 +15,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run gobump-deps action
-        uses: lzap/gobump@v1
+        uses: lzap/gobump@main
```
lzap, line 18: I thought GHA uses semantic versioning and will pickup the latest v1.0.9 but it does not work that way. I would need to introduce `v1` branch, which is not worth the effort for such a simple GHA I think.

achilleas-k (approved):

PR #4764, opened 2025-06-27
[RHEL 8.10] Reduce and update tests

achilleas-k:
This PR reduces testing on the rhel-8.10.0 branch to only test things relevant for RHEL 8.10:
- Only run tests on 8.10 GA runners
- Disable any service-related testing
- Update test configurations to match current test setup (terraform configs, runner versions, etc).

This is split off from #4759 to focus on getting tests working before pulling in the new version of osbuild/images.


achilleas-k:
The gitlab trigger action is run from `main`, meaning I can't disable the nightly pipeline trigger which fails for this branch.

croissanne (approved):

thozza (approved):
LGTM

PR #4770, opened 2025-07-08
Consolidate Koji upload code to `osbuild/images` (HMS-8803)

thozza:
Modify `internal/upload/koji` to not depend on any code from `osbuild-composer` and especially any internal code. In addition, do not create any logger in the upload library code, but instead accept it from the library consumer.


achilleas-k (approved):
This is great.  Thank you.
LGTM

PR #4779, opened 2025-07-18
jobqueue: handle escaped null bytes in postgres (HMS-8885)

croissanne:
Postgres doesn't accept `\u0000` in the jsonb datatype. Switch to the
json datatype which is larger and slower, but accepts escaped null
bytes.

As we don't actually query or index the results jsonb directly, the
impact of this should be minimal.

See: https://www.postgresql.org/docs/current/datatype-json.html

---

A more naive alternative can be found in 

```
-- Before altering the type of a column, any view using that column must
-- be dropped.
DROP VIEW ready_jobs;

-- jsonb does not support unicode escaped null bytes, and we don't
-- control much of the output generated by osbuild.
ALTER TABLE jobs
      ALTER COLUMN result TYPE json;

CREATE VIEW ready_jobs AS
SELECT *
FROM jobs
WHERE started_at IS NULL
  AND canceled = FALSE
  AND id NOT IN (
    SELECT job_id
    FROM job_dependencies JOIN jobs ON dependency_id = id
    WHERE finished_at IS NULL
)
ORDER BY queued_at ASC
```

I'm open to opinions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,6 @@ jobs:
         PGPORT: 5432
       run: |
         ./tools/dbtest-prepare-env.sh
-        ./tools/dbtest-run-migrations.sh
     - run: ./tools/dbtest-entrypoint.sh
 
   python-lint:
--- a/cmd/osbuild-composer-dbjobqueue-tests/main_test.go
+++ b/cmd/osbuild-composer-dbjobqueue-tests/main_test.go
@@ -4,33 +4,74 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"os"
+	"os/exec"
+	"strings"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/jackc/pgx/v4"
+	"github.com/stretchr/testify/require"
+
 	"github.com/osbuild/osbuild-composer/pkg/jobqueue"
 	"github.com/osbuild/osbuild-composer/pkg/jobqueue/dbjobqueue"
 
 	"github.com/osbuild/osbuild-composer/internal/jobqueue/jobqueuetest"
 )
 
+func migrate(migration string) error {
+	// migrate
+	gopath, err := exec.Command("go", "env", "GOPATH").Output()
+	if err != nil {
+		return err
+	}
+	tern := fmt.Sprintf("%s/bin/tern", strings.TrimSpace(string(gopath)))
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	if out, err := exec.Command(
+		tern,
+		"migrate",
+		"--conn-string",
+		jobqueuetest.TestDbURL(),
+		"-m",
+		fmt.Sprintf("%s/../../pkg/jobqueue/dbjobqueue/schemas", wd),
+		"-d",
+		migration,
+	).CombinedOutput(); err != nil {
+		fmt.Println("tern output:", string(out))
+		return err
+	}
+	return nil
+}
+
 func TestJobQueueInterface(t *testing.T) {
-	makeJobQueue := func() (jobqueue.JobQueue, func(), error) {
-		// clear db before each run
-		conn, err := pgx.Connect(context.Background(), jobqueuetest.TestDbURL())
+	makeJobQueue := func(migration string, clean bool) (jobqueue.JobQueue, func(), error) {
+		err := migrate(migration)
 		if err != nil {
 			return nil, nil, err
 		}
-		defer conn.Close(context.Background())
-		for _, table := range []string{"job_dependencies", "heartbeats", "jobs"} {
-			_, err = conn.Exec(context.Background(), fmt.Sprintf("DELETE FROM %s", table))
+
+		if clean {
+			conn, err := pgx.Connect(context.Background(), jobqueuetest.TestDbURL())
 			if err != nil {
 				return nil, nil, err
 			}
-		}
-		err = conn.Close(context.Background())
-		if err != nil {
-			return nil, nil, err
+			defer conn.Close(context.Background())
+			for _, table := range []string{"job_dependencies", "heartbeats", "jobs"} {
+				_, err = conn.Exec(context.Background(), fmt.Sprintf("DELETE FROM %s", table))
+				if err != nil {
+					return nil, nil, err
+				}
+			}
+			err = conn.Close(context.Background())
+			if err != nil {
+				return nil, nil, err
+			}
 		}
 
 		q, err := dbjobqueue.New(jobqueuetest.TestDbURL())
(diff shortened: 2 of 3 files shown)


supakeen (approved):
Codewise, sure; but I'm not sure of the implications of the schema change on production :)

croissanne:
> Codewise, sure; but I'm not sure of the implications of the schema change on production :)

This should be fine here, our tables are like 10GB at most, so the type casting + rewriting the table shouldn't take that long. I'll maybe rewrite it with adding a new column + backfill though.

croissanne:
```
---- tern: disable-tx ----

-- disable migration transaction, to allow for batched updates

-- jsonb does not support unicode escaped null bytes, and we don't
-- control much of the output generated by osbuild.
ALTER TABLE jobs ADD COLUMN IF NOT EXISTS new_result json;


-- backfill outside of the transaction to avoid locking the table
DO $$
DECLARE
  remaining_rows INTEGER := 1;
BEGIN
  WHILE remaining_rows > 0 LOOP
    WITH cte_jobs AS (
       SELECT ctid
         FROM jobs
         WHERE new_result IS NULL
         FOR UPDATE
         LIMIT 500
    )
    UPDATE jobs set new_result = result::json
    FROM cte_jobs
    WHERE jobs.ctid = cte_jobs.ctid;

    SELECT COUNT(id)
    FROM jobs
    WHERE result IS NOT NULL AND new_result IS NULL
    INTO remaining_rows;
    RAISE NOTICE 'REMAINING ROWS: %', remaining_rows;
  END LOOP;
END $$;


-- switching over the columns needs to happen in a transaction
BEGIN;
DROP VIEW ready_jobs;
ALTER TABLE jobs DROP COLUMN result;
ALTER TABLE jobs RENAME COLUMN new_result TO result;

CREATE VIEW ready_jobs AS
SELECT *
FROM jobs
WHERE started_at IS NULL
  AND canceled = FALSE
  AND id NOT IN (
    SELECT job_id
    FROM job_dependencies JOIN jobs ON dependency_id = id
    WHERE finished_at IS NULL
)
ORDER BY queued_at ASC;
COMMIT;
```

is an alternative that would work I think

lzap:
Have you considered just fixing the source of these null characters? The database is currently clean of any of these, JSONB should not allow any of these records to be stored. I mean, how valuable are strings with non-readable characters in a column that (I suppose) is only used by humans for debugging?

lzap:
> We can do post-processing in the upper-parts of the stack...

This is actually what I meant, there are just few callsites which insert a new result. The idea was to cleanup all strings there of non-printable characters.

lzap:
> Then I misunderstood what you meant by "source". I had previously suggested a simple replace as well, however, note that `jsonb` types cannot carry many of the escape sequences that are valid JSON; this isn't just about null-bytes but will likely also break on emoji (maybe, not sure) and some other bits and bobs :)

My bad. Anyways, if the result is not machine-processed and it is only meant for humans, I would not mind we simply strip all runes which are not alphanumeric or perhaps printable sub-ASCII 128 (ASCII-7?) characters out completely. That is why I was asking what is the reason for such thing.

But it also comes with a price - we will start storing data which was previously different so we will have a drift in the output.

croissanne:
I think the argument pro `json` over `jsonb` is that currently job results just have to be valid json. And having `\u0000` in the result is still considered valid json. When looking at the intf the comment also states as much.

```
	// Tries to requeue a running job by its ID
	//
	// If the job has reached the maxRetries number of retries already, finish the job instead.
	// `result` must fit the associated job type and must be serializable to JSON.
	// Fills in result, and returns if the job was requeued, or an error.
	RequeueOrFinishJob(id uuid.UUID, maxRetries uint64, result interface{}) (bool, error)
```

croissanne:
```
func (q *DBJobQueue) RequeueOrFinishJob(id uuid.UUID, maxRetries uint64, result interface{}) (bool, error) {
	data, err := json.Marshal(result)
	if err != nil {
		return false, fmt.Errorf("unable to marshal result: %w", err)
	}
	resdata := bytes.ReplaceAll(data, []byte("\\u0000"), []byte{})
```

Is an alternative again then. @lzap @ondrejbudai @supakeen what do you prefer?

lzap:
Let me [review the docs](https://www.postgresql.org/docs/current/datatype-json.html): _The input function for jsonb is stricter: it disallows Unicode escapes for characters that cannot be represented in the database encoding. The jsonb type also rejects \u0000 (because that cannot be represented in PostgreSQL's text type), and it insists that any use of Unicode surrogate pairs to designate characters outside the Unicode Basic Multilingual Plane be correct. Valid Unicode escapes are converted to the equivalent single character for storage; this includes folding surrogate pairs into a single character._

I feel like changing from `jsonb` to `json` is a step back and I find the workaround sufficient. But the docs also mentions it is not only about the null byte, but also about _characters that cannot be represented in the database encoding_. I researched this and it looks like this should not be any issue for UTF database which we I believe use, this is only relevant for different encodings. So we should be good.

For me the workaround is good enough, I believe we do not have this in all builds but this is some weird edge case when someone has this in a username or something (?).

croissanne:
> Let me [review the docs](https://www.postgresql.org/docs/current/datatype-json.html): _The input function for jsonb is stricter: it disallows Unicode escapes for characters that cannot be represented in the database encoding. The jsonb type also rejects \u0000 (because that cannot be represented in PostgreSQL's text type), and it insists that any use of Unicode surrogate pairs to designate characters outside the Unicode Basic Multilingual Plane be correct. Valid Unicode escapes are converted to the equivalent single character for storage; this includes folding surrogate pairs into a single character._
> 
> I feel like changing from `jsonb` to `json` is a step back and I find the workaround sufficient. But the docs also mentions it is not only about the null byte, but also about _characters that cannot be represented in the database encoding_. I researched this and it looks like this should not be any issue for UTF database which we I believe use, this is only relevant for different encodings. So we should be good.
> 
> For me the workaround is good enough, I believe we do not have this in all builds but this is some weird edge case when someone has this in a username or something (?).

It's quite common sadly. A lot of rhel9 images + openscap have null bytes in the logs.

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -43,5 +84,58 @@ func TestJobQueueInterface(t *testing.T) {
...
+	require.Error(t, err)
+
+	tr := jobqueuetest.TestResult{Logs: []byte("{\"blegh\": \"really important stuff!\"}")}
+	_, err = q.RequeueOrFinishJob(id, 0, &tr)
+	require.NoError(t, err)
+
+	require.NoError(t, migrate("last"))
+
+	stop()
+	// continue with a new connection, as the old connection will still fail the check below
```
croissanne: needs a comment on why that is

.github/workflows/tests.yml
```diff
@@ -80,7 +80,6 @@ jobs:
         PGPORT: 5432
       run: |
         ./tools/dbtest-prepare-env.sh
-        ./tools/dbtest-run-migrations.sh
```
lzap, line 83: Just checking was this intentional?

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
```diff
@@ -4,33 +4,74 @@ package main
...
+		"-m",
+		fmt.Sprintf("%s/../../pkg/jobqueue/dbjobqueue/schemas", wd),
+		"-d",
+		migration,
+	).CombinedOutput(); err != nil {
+		fmt.Println("tern output:", string(out))
+		return err
+	}
+	return nil
+}
```
lzap, line 50: Just for the record, it is possible to run migrations directly via Go call, the `tern` library does not advertise this as the author says the API is not yet fixed or stable, but it just [worked fine for provisioning service](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/migrations/migrations.go#L113).

Advantage of this is that when running tests, there will be no such thing as `tern: command not found` error.

lzap (approved):

.github/workflows/tests.yml
croissanne: yes, the test runs the migrations now.

cmd/osbuild-composer-dbjobqueue-tests/main_test.go
croissanne: True we can switch it over at some point. I'd prefer to get this in first though, as the bug has been marked as blocker now.

PR #4781, opened 2025-07-21
deps: bump images to 0.164

supakeen:
We have an issue where `%post` scripts are not being executed due to our incorrect handling of Anaconda modules [1].

[1]: https://github.com/osbuild/bootc-image-builder/issues/968

---

Since Fedora IoT builds through the service we will probably want to release this quickly, if we can. So they can respin with an ISO that works correctly.

However, I think we can't because this `images` version requires a newer version of `osbuild` than is available for RHEL/CentOS (it *is* available on Fedora)?


ondrejbudai (approved):

thozza (commented):
The PR needs to update `Schutzfile`, specifically the pinned ref of `osbuild` for all distros. It needs to be at least the `v156` version for CI to pass.

PR #4783, opened 2025-07-22
test/repositories: use CDN repos for 9.5, 9.6 and 10.0

thozza:
These releases are already GA. For 9.6 and 10.0, we test them in a scheduled GA pipeline, which runs the CI tests on GA runners, which don't have RH VPN access. Previously, some tests in the GA pipeline would fail, because they would use rpmrepo snapshots of nightly composes for these releases, which are behind RH VPN. Change the test repo definitions to use CDN repositories instead, which should fix the GA pipeline.


tkoscieln (approved):
Funky :man_facepalming:  That would propably be my doing, since the GA pipeline does not run on the PRs I didn't notice this.

PR #4808, opened 2025-08-05
cloudapi: depsolve blueprint + image

bcl:
This adds support for depsolving the blueprint and the image type (for the types that support packages) by adding an optional `image_type` to the depsolve request.


internal/cloudapi/v2/depsolve.go
```diff
@@ -56,14 +60,44 @@ func (request *DepsolveRequest) Depsolve(df *distrofactory.Factory, rr *reporegi
...
+		if err != nil {
+			return nil, HTTPError(ErrorUnsupportedImageType)
+		}
+
+		// use the same seed for all images so we get the same IDs
+		bigSeed, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+		if err != nil {
+			return nil, HTTPError(ErrorFailedToGenerateManifestSeed)
+		}
+		manifestSeed := bigSeed.Int64()
```
lzap, line 77: Nitpick: I see this pattern across the codebase, maybe a time to do a follow-up commit that introduces a shared utility function? I would love seeing that, as long as the new place for the code is not in the all-catch `common` package :-D

```
internal/cloudapi/v2/compose.go:1216-   // add the user-defined repositories only to the depsolve job for the
internal/cloudapi/v2/compose.go:1217-   // payload (the packages for the final image)
internal/cloudapi/v2/compose.go:1218-   payloadRepositories := request.GetPayloadRepositories()
internal/cloudapi/v2/compose.go:1219-
internal/cloudapi/v2/compose.go:1220-   // use the same seed for all images so we get the same IDs
internal/cloudapi/v2/compose.go:1221:   bigSeed, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
internal/cloudapi/v2/compose.go:1222-   if err != nil {
internal/cloudapi/v2/compose.go:1223-           return nil, HTTPError(ErrorFailedToGenerateManifestSeed)
internal/cloudapi/v2/compose.go:1224-   }
internal/cloudapi/v2/compose.go:1225-   manifestSeed := bigSeed.Int64()
internal/cloudapi/v2/compose.go:1226-
internal/weldr/api.go:2524-             size = imageType.Size(minSize)
internal/weldr/api.go:2525-     } else {
internal/weldr/api.go:2526-             size = imageType.Size(cr.Size)
internal/weldr/api.go:2527-     }
internal/weldr/api.go:2528-
internal/weldr/api.go:2529:     bigSeed, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
internal/weldr/api.go:2530-     if err != nil {
internal/weldr/api.go:2531-             panic("cannot generate a manifest seed: " + err.Error())
internal/weldr/api.go:2532-     }
internal/weldr/api.go:2533-     seed := bigSeed.Int64()
internal/weldr/api.go:2534-
```

internal/cloudapi/v2/depsolve.go
```diff
@@ -56,14 +60,44 @@ func (request *DepsolveRequest) Depsolve(df *distrofactory.Factory, rr *reporegi
...
+		ibp := blueprint.Convert(bp)
+		manifestSource, _, err := ir.imageType.Manifest(&ibp, ir.imageOptions, ir.repositories, &ir.manifestSeed)
+		if err != nil {
+			return nil, HTTPErrorWithInternal(ErrorFailedToDepsolve, err)
+		}
+
+		packageSet = manifestSource.GetPackageSetChains()
+	} else {
+		// Send the depsolve request to the worker
+		packageSet["os"] = []rpmmd.PackageSet{
```
achilleas-k, line 94: This will work fine for now (and possibly forever), but it does rely on the pipeline name being significant, which I'm not entirely comfortable with.  Relevant discussion here: https://github.com/osbuild/images/pull/1696

Long term solution could be the same, having the `manifestsource` return information about pipeline names (build, payload, etc).

bcl: Right, I also considered combining them all but then that mixes the build set and the anaconda-tree set and that isn't really what most end users care about. I also thought of excluding known pipelines, but that's also not great. So we'll just have to remember to change it. FWIW I think if the user cares about all the details of build, etc. they should just inspect the manifest.

PR #4810, opened 2025-08-06
Many: remove "Image tests" (HMS-9006)

thozza:
The image_tests.sh test case is being run only as part of the RHEL nightly pipeline. This test case evolved over the years. In principle, it can do the following:

 1. Read test cases from /test/data/manifests.
 2. Rebuild the manifest attached to the test case directly by running osbuild.
 3. Compare the image-info report if attached to the test case (NOT done for any test cases in /test/data/manifests).
 5. Boot-test the built image in the respective environment (NOT done for any test cases in /test/data/manifests).

Moreover, these manifests (or test cases) are not updated regularly after being generated for the first time.

All in all, given the fact that the test case rebuilds outdated manifests by directly calling osbuild and does not run any additional
tests, it adds no value compared to the osbuild/images CI. osbuild/images CI always rebuilds the latest image definition version
manifests with osbuild, so at the time the library is updated in osbuild-composer, this scenario has already been tested.

This PR removes:

- `cmd/osbuild-image-tests` - because it was not testing anything that is not already tested elsewhere.
- `cmd/gen-manifests` and friends - because there is no need to generate any manifests in this repository.
- `test/data/manifests` - because these are no longer needed for anything.
- `internal/boot` - because the code was not used to boot test anything, and it depended on the removed `cmd/osbuild-image-tests`.

In addition, some READMEs and SPEC file were updated accordingly.

This PR does not lower the test coverage of our code if we consider `osbuild/images` CI.

/jira-epic HMS-8802

JIRA: [HMS-9006](https://issues.redhat.com/browse/HMS-9006)


supakeen:
In addition; we have tests that build manifests in `osbuild` itself as well?

thozza:
> In addition; we have tests that build manifests in `osbuild` itself as well?

Correct, but those suit a bit different purpose. Although this makes manifest building tested to the extreme :laughing:

achilleas-k (approved):
Very strong contender for Best PR of 2025.  Love this!
The rest of us need to step up our game now.

LGTM!!

thozza:
The packer build is failing due to an issue with the latest `packer-plugin-amazon` release https://github.com/hashicorp/packer-plugin-amazon/issues/583

PR #4813, opened 2025-08-07
Packer: skip packer-plugin-amazon version 1.3.10 due to missing checksum

thozza:
The 1.3.10 release has some issues and fails when used by Packer. See https://github.com/hashicorp/packer-plugin-amazon/issues/586 for more information.


ondrejbudai (approved):
Thanks!

kingsleyzissou (approved):

PR #4834, opened 2025-08-18
internal: use manifest build/payload pipelines (images#1766)

mvo5:
With the new images library changes to move the pipeline
roles out of the image types into the manifest we need
to tweak osbuild-composer to use the new way of getting
the payload and build pipelines.

This commit implements that new method based on the
manifest instead of the image type. See images pr#1766
for the rational of the change.


croissanne:
I will admin merge this. We can address the upgrade failure after itm 26.

PR #4848, opened 2025-09-02
Dockerfile-ubi-packer: do not use quay.io/app-sre/packer image

thozza:
The quay.io/app-sre/packer:latest is outdated and has old Packer version installed. Since we are the only user of it, AppSRE suggested that we build our own image.

Change the containerfile to be built on top of UBI9 (the AppSRE image was based on UBI8) and install a subset of packages that were previously installed in the quay.io/app-sre/packer [0].

Specifically, install:
 - Packer
 - Ansible (since we use it to modify images)
 - awscli (to interact with AWS)

[0] https://github.com/app-sre/container-images/blob/8f0dff56739fd270cbb5d07530ae0b90e23174a6/logrotate/Dockerfile

/jira-epic HMS-9242


templates/packer/worker.pkr.hcl
```diff
@@ -165,6 +165,9 @@ build {
       "-e", "RH_ACTIVATION_KEY=${var.rh_activation_key}",
       "-e", "RH_ORG_ID=${var.rh_org_id}",
       "--tags", "${var.ansible_tags}",
+      # Use legacy SCP protocol, instead of SFTP, to prevent issues like:
+      # "Failed to connect to the host via scp: bash: line 1: /usr/lib/sftp-server: No such file or directory"
+      "--scp-extra-args", "-O",
```
lzap: Can't this be solved just by installing `sftp-server` on the server?

thozza: Well, depending on your definition of "just".

My understanding is that the package would need to be installed on the booted image that Packer is modifying. That would mean using a different provisioner to install it than the Ansible provisioner (because it chokes on gathering facts without it). So it would mean a bigger change.

For the purpose of the Packer build, I'd personally say that falling back to SCP protocol is acceptable.

kingsleyzissou (dismissed):

templates/packer/worker.pkr.hcl
lzap: I have not realized this chicken and egg problem, I was envisioning we just add a package name into a blueprint :-) Okay.

lzap (dismissed):

lzap:
Some tests are failing tho.

test/cases/container-embedding.sh
```diff
@@ -56,8 +56,10 @@ BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
 COMPOSE_START=${TEMPDIR}/compose-start-${IMAGE_KEY}.json
 COMPOSE_INFO=${TEMPDIR}/compose-info-${IMAGE_KEY}.json
 
-FEDORA_IMAGE_DIGEST="sha256:4d76a7480ce1861c95975945633dc9d03807ffb45c64b664ef22e673798d414b"
+# skopeo inspect docker://registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal:latest --format '{{.Digest}}'
+FEDORA_IMAGE_DIGEST="sha256:18e799cd680ad62ddc0c7b1448ecd7f856af96a161b20ed02a86f60e8aa78d74"
 FEDORA_LOCAL_NAME="localhost/fedora-minimal:v1"
+# skopeo inspect docker://registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/manifest-list-test:latest --format '{{.Digest}}'
```
lzap: Some leftover comments perhaps?

thozza:
I had to update the `container-embedding` test case, because @achilleas-k pushed a new multi-arch container image to the docker://registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal:latest tag and effectively changed the manifest digest that was pinned in the test case. The change was made in good faith to fix a problem in images CI, but nobody realized the effect on osbuild-composer CI.

In addition to that, I also bumped the timeout in the `cross-distro` test case, because it is occasionally failing with compose not finishing within 10 minutes.

test/cases/container-embedding.sh
thozza: No, this is so that the next person does not need to figure out how to get the digest...

achilleas-k (approved):
LGTM.  Nice work fixing all this up.

I know this is going to change with the fix for the container-embedding test, but pre-approving as a note for later.

croissanne (approved):

lzap (approved):

achilleas-k (approved):
Thanks again!

thozza:
OMG, I made one more mistake in the container embed test case :facepalm: (The image ID is checked without the `sha256:` prefix, which the variable holding it contains...)

croissanne (approved):

achilleas-k (approved):

PR #4857, opened 2025-09-08
[10.0] internal/blueprint: add JSON tailoring to bp conversion (HMS-9302)

kingsleyzissou:
The blueprint convert function was missing the json tailoring case. This meant that if the json tailoring customization was provided in the blueprint, the customization would get ignored and the tailoring profile would not be applied to the image.


lzap (approved):
Looks clean. Tho I am unsure about the workflow - I see this is being done for EL9/EL10 but I do not see similar patch in `main`. If this was extracted from a patch, please include commit sha that this was extracted from.

A jira issue would be helpful too.

PR #4858, opened 2025-09-09
worker & executor: use vector to forward logs from executor to worker (HMS-9304)

croissanne:
This is a requirement to remove internet access from the executors entirely, yet keep logging.

Mind that this is just the first step, removing internet access would need to be done after the forwarding has been shown to work properly.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
+++ b/templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
@@ -4,7 +4,9 @@ source /tmp/cloud_init_vars
 
 echo "Writing vector config."
 
-REGION=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .region)
+ID_DOC=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/document)
+REGION=$(echo "$ID_DOC" | jq -r .region)
+PRIVATE_IP=$(echo "$ID_DOC" | jq -r .privateIp)
 
 sudo mkdir -p /etc/vector
 sudo tee /etc/vector/vector.yaml > /dev/null << EOF
@@ -13,8 +15,11 @@ sources:
     type: journald
     exclude_units:
       - vector.service
+  executor:
+    type: vector
+    address: ${PRIVATE_IP}:12005
 sinks:
-  out:
+  worker:
     type: aws_cloudwatch_logs
     inputs:
       - journald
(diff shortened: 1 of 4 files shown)


templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
```diff
@@ -4,7 +4,9 @@ source /tmp/cloud_init_vars
 
 echo "Writing vector config."
 
-REGION=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .region)
+ID_DOC=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/document)
+REGION=$(echo "$ID_DOC" | jq -r .region)
+PRIVATE_IP=$(echo "$ID_DOC" | jq -r .privateIp)
```
lzap, line 9: Looking at the AWS docs: _The private IP address of the instance. For IPv4-only and dual-stack instances, this contains the IPv4 address. For IPv6-only instances, this contains the IPv6 address. This field is deprecated; use privateIPv4Address or primaryIPv6Address instead._

Maybe use the newer field name.

templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
```diff
@@ -10,6 +10,7 @@ REGION=$(curl -Ls http://169.254.169.254/latest/dynamic/instance-identity/docume
 HOSTNAME=$(hostname)
 CLOUDWATCH_ENDPOINT="https://logs.$REGION.amazonaws.com"
 OSBUILD_EXECUTOR_CLOUDWATCH_GROUP=${OSBUILD_EXECUTOR_CLOUDWATCH_GROUP:-osbuild-executor-log-group}
+HOST_WORKER_ADDRESS=${HOST_WORKER_ADDRESS:-empty}
```
lzap:
```suggestion
HOST_WORKER_ADDRESS=${HOST_WORKER_ADDRESS:-127.0.0.1}
```

templates/packer/ansible/roles/common/files/worker-initialization-scripts/worker_executor.sh
```diff
@@ -30,6 +31,17 @@ sinks:
     encoding:
       codec: json
 EOF
+
+if [ "$HOST_WORKER_ADDRESS" != "empty" ]; then
```
lzap:
```suggestion
if [ "$HOST_WORKER_ADDRESS" != "127.0.0.1" ]; then
```

Maybe, could be also always created because configuration `127.0.0.1` "should" be the default. At least this would ensure we have the very same configuration on both node types, this could create a slight difference between the two depending on default values.

templates/packer/ansible/roles/common/files/worker-initialization-scripts/vector.sh
croissanne: It must be quite new, as https://github.com/aws/aws-sdk-go-v2/blob/648027edb8aeba036195538174a63cbccaca8c16/feature/ec2/imds/api_op_GetInstanceIdentityDocument.go#L94 doesn't support it yet.

croissanne: Mind if we keep it on the deprecated field until we can switch both the go code and the packer scripts over?

croissanne: https://github.com/aws/aws-sdk-go-v2/issues/3183 opened this

lzap: Sure no biggie, thanks for testing it.

thozza (approved):
LGTM

PR #4859, opened 2025-09-09
gha: split gobump and images

lzap:
This utilizes new features of gobump that allows to split images and non-images dependencies into two separate PRs. It also adds a changelog via GitHub Gist for each PR. I am assuming that the token has permissions to create gists, if not please add this permission for me.


.github/workflows/gobump.yml
```diff
@@ -4,19 +4,33 @@ name: "Updates Go dependencies via gobump"
 on:  # yamllint disable-line rule:truthy
   workflow_dispatch:
   schedule:
-    # Every Sunday at 15:00
-    - cron: "0 15 * * 0"
+    # Every Monday at 6:00 PM (images release at 5:00)
```
mvo5: This is maybe a silly question, but why is this changing from Sun->Mon ?

lzap: Of course sorry, I noticed that we tag `images` on Monday at 5:00 while dep update was executed in the worst possible day, that is the day before. This aligns it as I was trying to explain in the comment (unsuccessfully).

mvo5: Aha, thank you! That makes perfect sense now.

.github/workflows/gobump.yml
```diff
@@ -4,19 +4,33 @@ name: "Updates Go dependencies via gobump"
...
+        run: "git reset --hard HEAD"
+
+      - name: Run gobump-deps action - all other
+        uses: lzap/gobump@main
+        with:
+          go_version: "1.23.9"
+          token: ${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}
+          exec_pr: ./tools/prepare-source.sh
+          exclude: "github.com/osbuild/images"
+          commit_message: "deps: bump all other dependencies"
```
achilleas-k: The commit message will appear on its own when updating dependencies and "all other" wont really make sense.  Perhaps we can change it to something like:
```
deps: update dependencies (excluding osbuild/images)
```

achilleas-k (approved):
Thanks.  LGTM.
Small comment in line, but wont block.

lzap:
Reworded the ugly comment about the schedule.

Reworded both commit messages.

No other changes, thanks.

achilleas-k (approved):

supakeen (approved):

PR #4861, opened 2025-09-11
test: run older CIV tag on nightlies

achilleas-k:
CIV was recently updated [1] to cover changes in the upstream image definitions [2] in Azure images.  The 'pr-456' tag is the latest that does not include those changes.  Older versions of osbuild-composer are being tested in nightlies for 9.7 and 10.1, so it should run with the older image definitions that do not include the changes.

[1] https://github.com/osbuild/cloud-image-val/pull/457
[2] https://github.com/osbuild/images/pull/1814

### DIFF
--- a/test/cases/azure.sh
+++ b/test/cases/azure.sh
@@ -201,6 +201,13 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+elif ! nvrGreaterOrEqual "osbuild-composer" "151"; then
+  # osbuild-composer v151 made changes in the Azure image definitions and CIV
+  # was updated to check for those changes:
+  # https://github.com/osbuild/cloud-image-val/pull/457
+  # This tag does not include those changes, so use it when running against
+  # older versions of composer.
+  TAG="pr-456"
 else
   # If not, get prod container
   TAG="prod"


test/cases/aws.sh
```diff
@@ -218,6 +218,11 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+elif [ "${NIGHTLY:=false}" == "true" ]; then
+  # If running in nightly, use an older CIV container
+  # This tag does not include the changes made in https://github.com/osbuild/cloud-image-val/pull/457
+  # The nightly tests for 9.7 and 10.1 should run with the older image definitions that do not include the changes
+  TAG="rhel-azure-grub-serial-io"
```
thozza: Shouldn't this apply only to the Azure test case?

test/cases/azure.sh
```diff
@@ -201,6 +201,11 @@ greenprint "Pulling cloud-image-val container"
 if [[ "$CI_PROJECT_NAME" =~ "cloud-image-val" ]]; then
   # If running on CIV, get dev container
   TAG=${CI_COMMIT_REF_SLUG}
+elif [ "${NIGHTLY:=false}" == "true" ]; then
+  # If running in nightly, use an older CIV container
+  # This tag does not include the changes made in https://github.com/osbuild/cloud-image-val/pull/457
+  # The nightly tests for 9.7 and 10.1 should run with the older image definitions that do not include the changes
+  TAG="rhel-azure-grub-serial-io"
```
thozza: This will break nightlies again, once the new version of composer lands in RHEL nightly composes. Please check the osbuild-composer version in the condition instead.

test/cases/aws.sh
achilleas-k: Right, yeah, just blindly did all instances of CIV.

test/cases/azure.sh
achilleas-k: Done.  Thanks!

PR #4868, opened 2025-09-17
Weldr API endpoints & store: don't use `rpmmd` package structs for serialization (HMS-9376)

thozza:
Weldr API endpoints and the JSON store, used only by Weldr API, directly use `rpmmd` data structures representing packages for serialization. This means that any changes to the data structures in `osbuild/images` affect Weldr API endpoint responses and the store. This PR decouples Weldr API endpoints and store from `rpmmd` structures.

Note that the `rpmmd.RepoConfig` is still used for serialization. It may make sense to decouple it in the future as well, but I omitted it at this point because it does not block the refactoring I plan to do in `osbuild/images` as part of the SBOM rearchitecture.

/jira-epic HMS-8910

JIRA: [HMS-9376](https://issues.redhat.com/browse/HMS-9376)

### DIFF
--- a/cmd/osbuild-store-dump/main.go
+++ b/cmd/osbuild-store-dump/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/osbuild/images/pkg/sbom"
 	"github.com/osbuild/osbuild-composer/internal/store"
 	"github.com/osbuild/osbuild-composer/internal/target"
+	"github.com/osbuild/osbuild-composer/internal/weldrtypes"
 )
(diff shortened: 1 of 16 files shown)


bcl (approved):
Ack

supakeen (approved):
Sure, though we still have a decently tight coupling through the JSON right?

croissanne (approved):
Neat, I maybe would've made it a subpackage of internal/weldr, but whatever.

PR #4870, opened 2025-09-18
worker/executor: drop cloudwatch logging from executor (HMS-9304)

croissanne:
This drops cloudwatch logging from the executor:
```
    worker: remove cloudwatch group option from aws.ec2 executor
    
    Logs should be forwarded from the executor to the worker.
    
    This facilitates isolating the executor from the internet. Cloudwatch
    doesn't have a well-defined IP range, so it's tricky to update the
    executor's security group in such a way to allow only cloudwatch
    traffic. By forwarding any log traffic from the executor via the worker,
    this problem is eliminated.
    
    Note that this just prepares for the needed security group changes which
    would isolate the executor from the internet.
```

See  #4858.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,11 +143,6 @@ build: $(BUILDDIR)/bin/ build-maintenance
 	go build -o $<osbuild-composer ./cmd/osbuild-composer/
 	go build -o $<osbuild-worker ./cmd/osbuild-worker/
 	go build -o $<osbuild-worker-executor ./cmd/osbuild-worker-executor/
-	go build -o $<osbuild-upload-azure ./cmd/osbuild-upload-azure/
-	go build -o $<osbuild-upload-aws ./cmd/osbuild-upload-aws/
-	go build -o $<osbuild-upload-gcp ./cmd/osbuild-upload-gcp/
-	go build -o $<osbuild-upload-oci ./cmd/osbuild-upload-oci/
-	go build -o $<osbuild-upload-generic-s3 ./cmd/osbuild-upload-generic-s3/
 	go build -o $<osbuild-mock-openid-provider ./cmd/osbuild-mock-openid-provider
 	# also build the test binaries
 	go test -c -tags=integration -o $<osbuild-composer-cli-tests ./cmd/osbuild-composer-cli-tests/main_test.go
--- a/cmd/osbuild-worker/config.go
+++ b/cmd/osbuild-worker/config.go
@@ -68,10 +68,9 @@ type containersConfig struct {
 }
 
 type executorConfig struct {
-	Type            string `toml:"type"`
-	IAMProfile      string `toml:"iam_profile"`
-	KeyName         string `toml:"key_name"`
-	CloudWatchGroup string `toml:"cloudwatch_group"`
+	Type       string `toml:"type"`
+	IAMProfile string `toml:"iam_profile"`
+	KeyName    string `toml:"key_name"`
 }
 
 type repositoryMTLSConfig struct {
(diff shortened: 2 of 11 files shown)


croissanne:
@thozza explicitly asked for your review because you made the cloudwatch group go via userdata. I checked osbuild/ansible-osbuild-worker and I don't think we use this configuration option in the workers? It is present in osbuild/image-builder-terraform however, so we might need to keep the option? I'm not sure.

Makefile
```diff
@@ -143,11 +143,6 @@ build: $(BUILDDIR)/bin/ build-maintenance
 	go build -o $<osbuild-composer ./cmd/osbuild-composer/
 	go build -o $<osbuild-worker ./cmd/osbuild-worker/
 	go build -o $<osbuild-worker-executor ./cmd/osbuild-worker-executor/
-	go build -o $<osbuild-upload-azure ./cmd/osbuild-upload-azure/
-	go build -o $<osbuild-upload-aws ./cmd/osbuild-upload-aws/
-	go build -o $<osbuild-upload-gcp ./cmd/osbuild-upload-gcp/
-	go build -o $<osbuild-upload-oci ./cmd/osbuild-upload-oci/
-	go build -o $<osbuild-upload-generic-s3 ./cmd/osbuild-upload-generic-s3/
 	go build -o $<osbuild-mock-openid-provider ./cmd/osbuild-mock-openid-provider
```
lzap, line 146: Whaaaat surprised this was not recognized by the CI. :-)

cmd/osbuild-worker/config.go
```diff
@@ -68,10 +68,9 @@ type containersConfig struct {
 }
 
 type executorConfig struct {
-	Type            string `toml:"type"`
-	IAMProfile      string `toml:"iam_profile"`
-	KeyName         string `toml:"key_name"`
-	CloudWatchGroup string `toml:"cloudwatch_group"`
+	Type       string `toml:"type"`
+	IAMProfile string `toml:"iam_profile"`
+	KeyName    string `toml:"key_name"`
```
lzap, line 73: Just thinking loud: this will cause TOML parser to fail if and only if TOML is parsed strict, not sure how upgrades will be done but deprecation could be safer.

cmd/osbuild-worker/config_test.go
```diff
@@ -70,16 +70,14 @@ offline_token = "/etc/osbuild-worker/offline_token"
 type = "aws.ec2"
 iam_profile = "osbuild-worker"
 key_name = "osbuild-worker"
-cloudwatch_group = "osbuild-worker"
```
lzap, line 73: Maybe leave it in the test with a comment `# deprecated in HMS-9304` to endure it does not break.

lzap (approved):

cmd/osbuild-worker/config.go
croissanne: I can't think of any location where we update the worker config separately from the worker image / version tbh. Quick pain is fine here I think.

Makefile
croissanne: yeh, i guess nothing uses the make targets. Tbh we could just drop them and tell ppl to use `go build`.

Makefile
```diff
@@ -143,11 +143,6 @@ build: $(BUILDDIR)/bin/ build-maintenance
 	go build -o $<osbuild-composer ./cmd/osbuild-composer/
 	go build -o $<osbuild-worker ./cmd/osbuild-worker/
 	go build -o $<osbuild-worker-executor ./cmd/osbuild-worker-executor/
-	go build -o $<osbuild-upload-azure ./cmd/osbuild-upload-azure/
-	go build -o $<osbuild-upload-aws ./cmd/osbuild-upload-aws/
-	go build -o $<osbuild-upload-gcp ./cmd/osbuild-upload-gcp/
-	go build -o $<osbuild-upload-oci ./cmd/osbuild-upload-oci/
-	go build -o $<osbuild-upload-generic-s3 ./cmd/osbuild-upload-generic-s3/
```
thozza, line 150: :innocent: Thank you!

PR #4873, opened 2025-09-25
go.mod: update to v0.197.0

mvo5:
Update go.mod to 0.197 and update the code. This also includes a tiny commit to pull common.Must() into composer as a helper but I'm happy to do this differently if desired.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -452,7 +452,7 @@ OpenShift_Virtualization:
         INTERNAL_NETWORK: ["true"]
 
 azure.sh:
-  extends: .integration
+  extends: .integration_rhel
   rules:
     # Run only on x86_64
     - !reference [.upstream_rules_x86_64, rules]
@@ -462,7 +462,7 @@ azure.sh:
     SCRIPT: azure.sh
 
 azure.sh_hyperv_gen2:
-  extends: .integration
+  extends: .integration_rhel
   rules:
     # Run only on x86_64
     - !reference [.upstream_rules_x86_64, rules]


internal/cloudapi/v2/depsolve.go
```diff
@@ -86,7 +86,10 @@ func (request *DepsolveRequest) Depsolve(df *distrofactory.Factory, rr *reporegi
 			return nil, HTTPErrorWithInternal(ErrorFailedToDepsolve, err)
 		}
 
-		packageSet = manifestSource.GetPackageSetChains()
+		packageSet, err = manifestSource.GetPackageSetChains()
+		if err != nil {
+			return nil, HTTPErrorWithInternal(ErrorFailedToDepsolve, err)
```
mvo5, line 91: This is a bit of an open question, i.e. if we should create a new ErrorFailedToGetPkSetChains or if we lump it under the existing error, please advise, happy to tweak this.

internal/cloudapi/v2/server.go
```diff
@@ -169,8 +169,12 @@ func (s *Server) enqueueCompose(irs []imageRequest, channel string) (uuid.UUID,
 		return id, HTTPErrorWithInternal(ErrorEnqueueingJob, err)
 	}
 
+	pkgSetChains, err := manifestSource.GetPackageSetChains()
+	if err != nil {
+		return id, HTTPErrorWithInternal(ErrorEnqueueingJob, err)
```
mvo5, line 174: Same question as above here, new error type?

internal/cloudapi/v2/depsolve.go
achilleas-k: I think lumping it with depsolve failure is ok.  It's not strictly a depsolve failure, but it's coming from the depsolver, so I it's fine.

achilleas-k (dismissed):
LGTM

I think the error types are valid, fwiw.

supakeen (approved):

supakeen (approved):

achilleas-k:
> `azure.sh` and `azure.sh_hyperv_gen2` testcases are failing on:
> 
> ```
> >               assert host.file(file_to_check).contains(item), \
>                     f'{item} was expected in {file_to_check}'
> E               AssertionError: earlyprintk=ttyS0 was expected in /proc/cmdline
> E               assert False
> E                +  where False = contains('earlyprintk=ttyS0')
> E                +    where contains = <file /proc/cmdline>.contains
> E                +      where <file /proc/cmdline> = <class 'testinfra.modules.base.GNUFile'>('/proc/cmdline')
> E                +        where <class 'testinfra.modules.base.GNUFile'> = <testinfra.host.Host paramiko://civ-vm-i
> ```
> 
> oci.sh testcases are failing on I have no idea what, it just says 'compose failed'.

Did we change Fedora Azure kernel args recently?  That looks like CIV is expecting to find `earlyprintk=ttyS0` and failing.  But I don't see or remember any changes to the Fedora Azure image types.

achilleas-k (approved):

PR #4875, opened 2025-09-26
internal/worker: delete unneeded compatibility layers (HMS-9455)

thozza:
When we change the worker API, we always add a compatibility layer to handle continuous upgrade of the osbuild-composer instance and composer worker instances in our service deployment.

Some compatibility layers were added years ago; therefore, the initially covered scenario can't happen in our deployment anymore, and we can safely delete the compatibility code.

/jira-epic HMS-8910

JIRA: [HMS-9455](https://issues.redhat.com/browse/HMS-9455)


supakeen (approved):

thozza:
> I have to wonder just how many of these things are still left around in osbuild-composer. I'm probably responsible for a lot of them myself.

Being responsible for backward compatibility is a good thing :) It is just that we usually don't get back to cleaning up things after some time...

PR #4879, opened 2025-09-30
Test/api/gcp: set root partition size to at least 4 GiB

thozza:
It turns out that gce-disk-expand is not able to expand the root partition placed on an LVM LV. The Cloud API test case uses filesystem customizations, which by default convert the partition table to use LVM. It also turns out that the root partition in such case ends up being 3 GiB big. While this is enough for RHEL-9 and RHEL-10, it causes issues for RHEL-8.10 test case, where the root partition ends up having only 20 KiB of free space left, which effectively makes the system unable to do set user SSH keys, etc.

Let's workaround this for now by increasing the size of the root partition and fix it later in osbuild/images by not using LVM for GCE image when using FS customizations.

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -175,7 +175,7 @@ fi
 # this in osbuild, we explicitly set the network backend to 'netavark'.
 # This is relevant only for RHEL-9 / c9s, because Fedora since F40 and el10
 # support only `netavark` backend.
-if [[ $ID == "rhel" && ${VERSION_ID%.*} == "9" ]]; then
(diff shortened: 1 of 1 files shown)


supakeen (approved):
The second part is probably a good topic for a team meeting to be honest. When using the `customizations.filesystem` without setting a size on `/` you tend to always end up with LVM and a root volume that's (often) too small.

croissanne (approved):

thozza:
Unfortunately, I cleaned up some "old" ci-runner images in GCP yesterday from 2023 and 2024, but apparently, we still used those c9s 2024 images :grimacing:  I'll fix it...

supakeen (approved):

croissanne (approved):

PR #4884, opened 2025-10-06
executor: remove internet access entirely (HMS-9304)

croissanne:
cloud/awscloud: revoke internet access from executor
    
    After the vector forwarding (#4858), it should be safe to remove
    internet access from executors again.
    
    By default security groups are created with internet access (egress to
    0.0.0.0/0) being allowed. First remove all egress rules in the
    executor's security group, then add a rule which allows egress to the
    worker host


thozza (approved):
LGTM, thanks for following up on this 🥳

PR #4885, opened 2025-10-06
CloudAPI: fix `/compose/{id}/manifests` endpoint behavior on manifest generation errors (HMS-9501)

thozza:
While porting https://github.com/osbuild/images/pull/1907 to osbuild-composer, I ran into the problem that the CloudAPI endpoint for compose manifests was not checking manifest generation job results and would return a 'nil' value instead of an error if the manifest generation failed.

Moreover, the mock data used by unit tests for depsolved packages was not complete and was lacking basic properties, such as the remote location URL, which is needed to generate an osbuild manifest.

This PR addresses the issues mentioned above, it also adds a unit test for compose manifests Cloud API endpoint and improves unit tests where failure debugging was previously unnecessarily complicated.

/jira-epic HMS-8910

JIRA: [HMS-9501](https://issues.redhat.com/browse/HMS-9501)

### DIFF
--- a/internal/cloudapi/v2/export_test.go
+++ b/internal/cloudapi/v2/export_test.go
@@ -1,0 +1,18 @@
+package v2
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+)
+
+// OverrideSerializeManifestFunc overrides the serializeManifestFunc for testing
+func OverrideSerializeManifestFunc(f func(ctx context.Context, manifestSource *manifest.Manifest, workers *worker.Server, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID uuid.UUID, seed int64)) func() {
+	originalSerializeManifestFunc := serializeManifestFunc
+	serializeManifestFunc = f
+	return func() {
+		serializeManifestFunc = originalSerializeManifestFunc
+	}
+}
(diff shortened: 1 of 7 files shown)


internal/cloudapi/v2/export_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+)
+
+// OverrideSerializeManifestFunc overrides the serializeManifestFunc for testing
+func OverrideSerializeManifestFunc(f func(ctx context.Context, manifestSource *manifest.Manifest, workers *worker.Server, depsolveJobID, containerResolveJobID, ostreeResolveJobID, manifestJobID uuid.UUID, seed int64)) func() {
```
mvo5, line 12: (nitpick) the convention (so far for the few instances we have in imaes and ibcli) is `Mock*(...) (restore func())` (i.e. different prefix and explicit named return func to signal what its about) - its not super important but might be nice to be consistent

internal/cloudapi/v2/server.go
```diff
@@ -694,7 +694,7 @@ func serializeManifest(ctx context.Context, manifestSource *manifest.Manifest, w
 	ms, err := manifestSource.Serialize(depsolveResultsInTheRightFormat, containerSpecs, ostreeCommitSpecs, nil)
 	if err != nil {
 		reason := "Error serializing manifest"
-		jobResult.JobError = clienterrors.New(clienterrors.ErrorManifestGeneration, reason, nil)
+		jobResult.JobError = clienterrors.New(clienterrors.ErrorManifestGeneration, reason, err.Error())
```
mvo5, line 697: :heart:

internal/cloudapi/v2/export_test.go
thozza: Thanks for the suggestion. I'll submit a follow-up, since it will be as much work (rebase, CI and reviews) as updating this PR and getting this in will unblock my other changes that I'm working on.

thozza: https://github.com/osbuild/osbuild-composer/pull/4887

PR #4896, opened 2025-10-20
test/cases/usbi-wsl: delete azure resources in order

croissanne:
Sometimes resources fail to delete, so delete them one by one, in order, without `--no-wait`. Mirrored from the azure api test case, which seems to work better.


thozza (approved):
❤️

supakeen (approved):

PR #4897, opened 2025-10-20
worker: support partial results (HMS-9343)

croissanne:
This allows `/jobs/{token}` to be called multiple times with partial
    results. As an empty result is also allowed, default to finishing the
    job if both result and partial are empty.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-awsec2.go
+++ b/cmd/osbuild-worker/jobimpl-awsec2.go
@@ -28,7 +28,7 @@ func (impl *AWSEC2CopyJobImpl) Run(job worker.Job) error {
 	result := worker.AWSEC2CopyJobResult{}
 
 	defer func() {
-		err := job.Update(&result)
+		err := job.Finish(&result)
 		if err != nil {
 			logWithId.Errorf("Error reporting job result: %v", err)
 		}
@@ -86,7 +86,7 @@ func (impl *AWSEC2ShareJobImpl) Run(job worker.Job) error {
 	result := worker.AWSEC2ShareJobResult{}
 
 	defer func() {
-		err := job.Update(&result)
+		err := job.Finish(&result)
 		if err != nil {
 			logWithId.Errorf("Error reporting job result: %v", err)
 		}
--- a/cmd/osbuild-worker/jobimpl-container-resolve.go
+++ b/cmd/osbuild-worker/jobimpl-container-resolve.go
@@ -52,7 +52,7 @@ func (impl *ContainerResolveJobImpl) Run(job worker.Job) error {
 		}
 	}
 
-	err = job.Update(&result)
+	err = job.Finish(&result)
 	if err != nil {
 		return fmt.Errorf("Error reporting job result: %v", err)
 	}
--- a/cmd/osbuild-worker/jobimpl-depsolve.go
+++ b/cmd/osbuild-worker/jobimpl-depsolve.go
@@ -131,7 +131,7 @@ func (impl *DepsolveJobImpl) Run(job worker.Job) error {
 	var result worker.DepsolveJobResult
 	// ALWAYS return a result
 	defer func() {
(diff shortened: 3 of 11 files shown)


pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -511,6 +515,56 @@ func (q *DBJobQueue) DequeueByID(ctx context.Context, id, workerID uuid.UUID) (u
...
+	defer conn.Release()
+
+	tx, err := conn.Begin(context.Background())
+	if err != nil {
+		return fmt.Errorf("error starting database transaction: %w", err)
+	}
+	defer func() {
+		err = tx.Rollback(context.Background())
+		if err != nil && !errors.Is(err, pgx.ErrTxClosed) {
+			q.logger.Error(err, "Error rolling back retry job transaction", "job_id", id.String())
```
thozza: Copy & paste error - the error message needs updating.

thozza (dismissed):
LGTM besides the inline comment.

pkg/jobqueue/dbjobqueue/dbjobqueue.go
croissanne: woopssss

thozza (approved):

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -511,6 +515,56 @@ func (q *DBJobQueue) DequeueByID(ctx context.Context, id, workerID uuid.UUID) (u
 	return token, dependencies, jobType, args, nil
 }
 
+func (q *DBJobQueue) UpdateJobResult(id uuid.UUID, result interface{}) error {
+	conn, err := q.pool.Acquire(context.Background())
```
lzap, line 519: Can we stop adding `context.Background()` for once? :-) I mean since context are unused it is inevitable but I'd prefer adding this higher up the stack to have less work when adding proper context passing, which is in my TODO. We could have a nicely correlated SQL debug queries here at some point.

internal/worker/server.go
```diff
@@ -817,6 +818,30 @@ func (s *Server) requestJob(ctx context.Context, arch string, jobTypes []string,
...
+
+	err = s.jobs.UpdateJobResult(jobId, partial)
+	if err != nil {
+		switch err {
+		case jobqueue.ErrNotRunning:
+			return ErrJobNotRunning
+		default:
+			return fmt.Errorf("error updating job: %w", err)
+		}
+	}
```
lzap, line 840: This could have been just if statement and with `errors.Is`. But this is fine I guess.

lzap (approved):

internal/worker/server.go
croissanne:
```
	if errors.Is(err, jobqueue.ErrNotRunning) {
		return ErrJobNotRunning
	} else if err != nil {
		return fmt.Errorf("error updating job: %w", err)
	}
```

ig, but the rest of the server code uses switch statements to compare errors. I'd rather keep it that way for consistency, and maybe change all of them to use errors.Is in a cleanup PR?

pkg/jobqueue/dbjobqueue/dbjobqueue.go
croissanne: Point taken, but what's the value of passing along contexts that aren't used? It would make it less consistent with the other functions in the jobqueue currently. Like I think it would make more sense as a change which is done in bulk, with clear descriptions of why etc.

If the context we'd pass along to the jobqueue include cancels, timeouts, etc. they would of course be useful, but we don't do that for queue functions that we expect to return immediately.

Also aren't sql queries already correlated through the fact that it's passing along job_id? Tbh I'd feel weird if the jobqueue queries the context for specific values (unless it just dumps all fields in the context).

PR #4899, opened 2025-10-21
cloudapi: add new dnf customizations [HMS-9335]

achilleas-k:
Add the new DNF customizations to the cloud API.
Add a conversion function from the cloud API DNF customizations to the Blueprint types and use it in both locations where customizations are read and converted.

### DIFF
--- a/internal/cloudapi/v2/compose_test.go
+++ b/internal/cloudapi/v2/compose_test.go
@@ -1274,3 +1274,53 @@ func TestDecodeMinsize(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertDNFCustomizations(t *testing.T) {
+	type testCase struct {
+		in     *DNF
+		expOut *blueprint.DNFCustomization
+	}
+
+	testCases := map[string]testCase{
+		"empty": {},
+		"set_releasever:false": {
+			in: &DNF{
+				Config: &DNFConfig{
+					SetReleasever: common.ToPtr(false),
+				},
+			},
+			expOut: &blueprint.DNFCustomization{
+				Config: &blueprint.DNFConfigCustomization{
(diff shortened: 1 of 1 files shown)


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -2799,6 +2803,18 @@ components:
...
+    DNF:
+      type: object
+      properties:
+        config:
+          $ref: '#/components/schemas/DNFConfig'
+    DNFConfig:
+      type: object
+      properties:
+        set_releasever:
+          type: boolean
```
ondrejbudai, line 2816: Should we also add `default: false` just to be explicit?

internal/cloudapi/v2/compose.go
```diff
@@ -1441,6 +1445,22 @@ func convertDiskCustomizations(disk *Disk) (*blueprint.DiskCustomization, error)
 	return bpDisk, bpDisk.Validate()
 }
 
+func convertDnfCustomizations(dnf *DNF) *blueprint.DNFCustomization {
+	if dnf == nil {
```
ondrejbudai, line 1449:
```suggestion
	if dnf == nil || dnf.Config == nil {
```

and then drop the first `if` later?

internal/cloudapi/v2/openapi.v2.yml
achilleas-k: Done.

internal/cloudapi/v2/compose.go
achilleas-k: I'd rather keep it this way and make sure it doesn't mess with any future expansions of the converter function when we add more sub-objects to DNF (even if we never do).

PR #4900, opened 2025-10-21
gitlab: only update github status at the end of the pipeline(s)

achilleas-k:
Updating the statuses on GitHub at the end of each job causes the statuses to show failure when the jobs are still running.  In osbuild and images we instead have a finish stage that runs the update_github_status.sh script either with the 'finish' argument, if everything succeeded, or with a 'fail' argument, if anything in the pipeline failed.  This makes sure that the statuses on GitHub are only updated at the end of the pipeline runs.


bcl (approved):
Looks good!

PR #4902, opened 2025-10-22
fix worker AMI builds by not running mock as root

ondrejbudai:
**build-rpms.py: use gp3 storage for the ephemeral VM**

gp3 is cheaper and faster.

---

**build-rpms.py: upgrade to RHEL 9.6**

---

**rpmbuild.yml: don't run mock as root**

https://github.com/osbuild/osbuild/commit/a52e256c481c00e61f785f3ad68e29e1ac8a09fb
introduced running osbuild's test in the %check phase of an rpm build.
A lot of osbuild's tests are gated on $UID == 0 (i.e. running as root).
Thus, if $UID != 0, then a lot of tests are intentionally skipped.

RPM building setups (koji, mockbuild.sh in this repository) usually run
mock as a non-root user, thus the tests get skipped. However, this
script actually runs mock as root, and thus these tests don't get
skipped, causing a lot of failures because there are missing tools
in the buildroot, some rootful tests require networking, or they cannot
run in an environment missing certain privileges.

This commit changes the playbook to run tasks as root only when they
require it (subscription-manager, dnf). Mock is no longer run as root,
aligning the rpm building process to what's common in other setups.

---

### DIFF
--- a/tools/appsre-ansible/rpmbuild.yml
+++ b/tools/appsre-ansible/rpmbuild.yml
@@ -1,6 +1,5 @@
 ---
 - name: Build osbuild rpms
-  become: yes
   remote_user: ec2-user
   hosts: rpmbuilder
   gather_facts: no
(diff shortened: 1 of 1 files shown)


croissanne (approved):
very nice, thank you

schuellerf (approved):
thanks!

PR #4903, opened 2025-10-22
Executor with progress updates (HMS-9341)

croissanne:
internal/osbuildexecutor: forward progress to composer
    
    Uses status scanner to monitor osbuild to generate progress
    updates. Progress updates are sent every 30 seconds.
    
    Only one level of subprogress needs to be supported. Because osbuild
    manifests consist of pipelines and stages, more than 2 levels of
    progress are not expected.

---

As a followup this will need something in the cloudapi to actually expose the progress, but this is big enough of a change already, and will need some careful monitoring of the stage workers once deployed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -865,6 +865,19 @@ WorkerExecutor:
     RUNNER: aws/fedora-41-x86_64
     IAM_INSTANCE_PROFILE: worker-executor
 
+WorkerExecutorFailure:
+  stage: test
+  extends: .terraform
+  rules:
+    - !reference [.upstream_rules_all, rules]
+    - !reference [.ga_rules_all, rules]
+  script:
+    - schutzbot/deploy.sh
+    - /usr/libexec/tests/osbuild-composer/worker-executor-crash.sh
+  variables:
+    RUNNER: aws/fedora-41-x86_64
+    IAM_INSTANCE_PROFILE: worker-executor
+
 finish:
   stage: finish
   dependencies: []
--- a/test/cases/worker-executor-crash.sh
+++ b/test/cases/worker-executor-crash.sh
@@ -1,0 +1,233 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+source /usr/libexec/tests/osbuild-composer/shared_lib.sh
+
+
+# Container image used for cloud provider CLI tools
+CONTAINER_IMAGE_CLOUD_TOOLS="quay.io/osbuild/cloud-tools:latest"
+
+# Provision the software under test.
+/usr/libexec/osbuild-composer-test/provision.sh none
+
+TEMPDIR=$(mktemp -d)
+BLUEPRINT_FILE=${TEMPDIR}/blueprint.toml
+COMPOSE_START=${TEMPDIR}/compose-start.json
+COMPOSE_INFO=${TEMPDIR}/compose-info.json
+DESCR_INST=${TEMPDIR}/descr-inst.json
+AUTH_SG=${TEMPDIR}/auth-sgrule.json
+DESCR_SGRULE=${TEMPDIR}/descr-sgrule.json
+KEYPAIR=${TEMPDIR}/keypair.pem
+INSTANCE_ID=$(curl -Ls http://169.254.169.254/latest/meta-data/instance-id)
+WORKER_HOST=$(curl -Ls http://169.254.169.254/latest/meta-data/local-ipv4)
+
+# Check available container runtime
+if type -p podman 2>/dev/null >&2; then
+    CONTAINER_RUNTIME=podman
+elif type -p docker 2>/dev/null >&2; then
+    CONTAINER_RUNTIME=docker
+else
+    echo No container runtime found, install podman or docker.
+    exit 2
+fi
+
+if ! hash aws; then
+    echo "Using 'awscli' from a container"
+    sudo "${CONTAINER_RUNTIME}" pull ${CONTAINER_IMAGE_CLOUD_TOOLS}
+
+    AWS_CMD="sudo ${CONTAINER_RUNTIME} run --rm \
+        -v ${TEMPDIR}:${TEMPDIR}:Z \
+        ${CONTAINER_IMAGE_CLOUD_TOOLS} aws --region $AWS_REGION --output json --color on"
+else
+    echo "Using pre-installed 'aws' from the system"
+    AWS_CMD="aws --region $AWS_REGION --output json --color on"
+fi
+$AWS_CMD --version
+
+subprocessPIDs=()
+function cleanup() {
+    # since this function can be called at any time, ensure that we don't expand unbound variables
+    AWS_CMD="${AWS_CMD:-}"
+
+    if [ -n "$AWS_CMD" ] && [ -f "$KEYPAIR" ]; then
+        $AWS_CMD ec2 delete-key-pair --key-name "key-for-$INSTANCE_ID-executor"
+    fi
+
+    for p in "${subprocessPIDs[@]}"; do
+        sudo pkill -P "$p" || true
+    done
+}
+
+trap cleanup EXIT
+
+$AWS_CMD ec2 create-key-pair --key-name "key-for-$INSTANCE_ID-executor" --query 'KeyMaterial' --output text > "$KEYPAIR"
+chmod 400 "$KEYPAIR"
+$AWS_CMD ec2 describe-key-pairs --key-names "key-for-$INSTANCE_ID-executor"
+
+sudo tee "/etc/osbuild-worker/osbuild-worker.toml" <<EOF
+[osbuild_executor]
+type = "aws.ec2"
+key_name = "key-for-$INSTANCE_ID-executor"
+EOF
+
+sudo systemctl restart osbuild-worker@1.service
+
+# Write a basic blueprint for our image.
+tee "$BLUEPRINT_FILE" > /dev/null << EOF
+name = "bash"
+description = "A base system"
+version = "0.0.1"
+
+[customizations]
+[customizations.services]
+enabled = ["blergh"]
+EOF
+
+sudo composer-cli blueprints push "$BLUEPRINT_FILE"
+
+WORKER_UNIT=$(sudo systemctl list-units | grep -o -E "osbuild.*worker.*\.service")
+sudo journalctl -af -n 1 -u "${WORKER_UNIT}" &
+subprocessPIDs+=( $! )
+
+sudo composer-cli --json compose start bash container | tee "$COMPOSE_START"
+COMPOSE_ID=$(get_build_info ".build_id" "$COMPOSE_START")
+
+EXECUTOR_IP=0
+for _ in {1..60}; do
+    $AWS_CMD ec2 describe-instances --filter "Name=tag:parent,Values=$INSTANCE_ID" > "$DESCR_INST"
+    RESERVATIONS=$(jq -r '.Reservations | length' "$DESCR_INST")
+    if [ "$RESERVATIONS" -gt 0 ]; then
+        EXECUTOR_IP=$(jq -r .Reservations[0].Instances[0].PrivateIpAddress "$DESCR_INST")
+        break
+    fi
+
+    echo "Reservation not ready ret, waiting..."
+    sleep 60
+done
+
+if [ "$EXECUTOR_IP" = 0 ]; then
+    redprint "Unable to find executor host"
+    exit 1
+fi
+
+RDY=0
+for _ in {0..60}; do
+    if ssh-keyscan "$EXECUTOR_IP" > /dev/null 2>&1; then
+        RDY=1
+        break
+    fi
+    sleep 10
+done
+
+if [ "$RDY" = 0 ]; then
+    redprint "Unable to reach executor host $EXECUTOR_IP"
+    exit 1
+fi
+
+greenprint "Setting up executor"
+# the executor should be created with exactly one egress rule (allowing traffic to the worker host)
+SGID=$(jq -r .Reservations[0].Instances[0].SecurityGroups[0].GroupId "$DESCR_INST")
+$AWS_CMD ec2 describe-security-group-rules --filters "Name=group-id,Values=$SGID" > "$DESCR_SGRULE"
+
+EGRESS_TARGET=$(jq -r '.SecurityGroupRules[] | select(.IsEgress).CidrIpv4' "$DESCR_SGRULE")
+if [ "$EGRESS_TARGET" != "$WORKER_HOST/32" ]; then
+    echo executors "$EGRESS_TARGET" is not the expected "$WORKER_HOST/32"
+    exit 1
+fi
+
+# allow the executor to access the internet for the setup:
+$AWS_CMD ec2 authorize-security-group-egress --group-id "$SGID" --protocol tcp --cidr 0.0.0.0/0 --port 1-65535 > "$AUTH_SG"
+SGRULEID=$(jq -r .SecurityGroupRules[0].SecurityGroupRuleId "$AUTH_SG")
+
+GIT_COMMIT="${GIT_COMMIT:-${CI_COMMIT_SHA}}"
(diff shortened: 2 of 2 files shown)


internal/osbuildexecutor/osbuild-executor.go
```diff
@@ -3,9 +3,11 @@ package osbuildexecutor
...
 	"io"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/osbuild/images/pkg/osbuild"
 )
 
 type Executor interface {
-	RunOSBuild(manifest []byte, exports, checkpoints []string, errorWriter io.Writer, opts *osbuild.OSBuildOptions) (*osbuild.Result, error)
+	RunOSBuild(manifest []byte, exports, checkpoints []string, errorWriter io.Writer, logger *logrus.Entry, opts *osbuild.OSBuildOptions) (*osbuild.Result, error)
```
lzap: Please use `logrus.FieldLogger` interface: https://github.com/sirupsen/logrus/blob/cb253f3080f18ec7e55b4c8f15b62fe0a806f130/logrus.go#L142

internal/osbuildexecutor/runner-impl-host.go
```diff
@@ -1,15 +1,55 @@
...
+		if err != nil {
+			return nil, fmt.Errorf(`error parsing osbuild status, please report a bug: %w`, err)
+		}
+		if st == nil {
+			break
+		}
+
+		if st.Progress != nil {
+			// TODO
+			logger.Infof("%s (step %d out of %d)", st.Message, st.Progress.Done, st.Progress.Total)
```
lzap: This should use structured logging, for logrus use `WithFields` or similar method.

croissanne:
> The PR is ready for review but I see no changes, including that `replace` statement. I guess this was just a rebase.

Composer CI has this thing where it blocks execution if the PR is in a draft state (unless you manually click). Maybe we should change that.

thozza:
> > The PR is ready for review but I see no changes, including that `replace` statement. I guess this was just a rebase.
> 
> Composer CI has this thing where it blocks execution if the PR is in a draft state (unless you manually click). Maybe we should change that.

You should use `WIP+test` label if you want tests to run for Draft PR. I'm not saying that we shouldn't reconsider the current state... :roll_eyes:

croissanne:
> > > The PR is ready for review but I see no changes, including that `replace` statement. I guess this was just a rebase.
> > 
> > 
> > Composer CI has this thing where it blocks execution if the PR is in a draft state (unless you manually click). Maybe we should change that.
> 
> You should use `WIP+test` label if you want tests to run for Draft PR. I'm not saying that we shouldn't reconsider the current state... 🙄

ah thanks, had no idea!

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -123,42 +125,90 @@ func writeInputArchive(cacheDir, store string, exports []string, manifestData []
...
+					if err != nil {
+						logger.Errorf("Unable to update job: %s", err.Error())
+					}
+				}()
+			}
+		}
+		if st.Trace != "" {
+			logger.Errorf("%s", st.Trace)
+		}
 	}
```
thozza, line 171: This block is an exact copy of the one in the `host` executor implementation. Would it make sense to extract it into a shared function?

thozza (dismissed):
Thanks, LGTM. I've added one suggestion inline.

internal/osbuildexecutor/runner-impl-aws-ec2.go
croissanne: probably

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -39,63 +41,96 @@ func (wf *writeFlusher) Write(p []byte) (n int, err error) {
...
+		ExtraEnv:    control.Environments,
+		BuildLog:    logf,
+		Stdout:      os.Stdout,
+		Monitor:     osbuild.MonitorJSONSeq,
+		MonitorFile: wPipe,
+	})
 	if err := cmd.Start(); err != nil {
 		return "", err
 	}
+	flusher, ok := output.(http.Flusher)
```
lzap, line 85: You could create a embedded interface `WriterFlusher` and make it part of the contract. This is acceptable, but raises a red flag.

internal/osbuildexecutor/runner-common.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+		if st.Trace != "" {
+			logger.Errorf("%s", st.Trace)
+		}
+	}
+	// avoid a partial job updates after this function has returned
+	wg.Wait()
+	return osbuildStatus.Result()
+}
```
lzap, line 71: The goroutine spawning in a loop is a bit fragile and unnecessary, have you considered a more traditional producer-consumer approach via a channel? There could be simply one goroutine responsible for sending updates, the delay could be enforced with a simple `time.Sleep` and data can be send to the unbuffered channel with `switch` option skipping updates which were not yet picked up. Something like:

```go
updates := make(chan worker.JobResult) // unbuffered channel, sends are non-blocking via select default
var wg sync.WaitGroup

// Single consumer: performs job.Update and enforces throttle via sleep
wg.Add(1)
go func() {
	defer wg.Done()
	for partial := range updates {
		if job != nil {
			if err := job.Update(partial); err != nil {
				logger.Errorf("Unable to update job: %s", err.Error())
			}
		}
		// enforce minimum delay between updates
		time.Sleep(MinTimeBetweenUpdates)
	}
}()

// Producer: reads osbuild status and sends updates (non-blocking) to consumer
for {
	st, err := osbuildStatus.Status()
	if err != nil {
		close(updates)
		return nil, fmt.Errorf(`error parsing osbuild status, please report a bug: %w`, err)
	}
	if st == nil {
		// no more statuses; close channel to stop consumer
		close(updates)
		break
	}

	progress := logrus.Fields{}
	if st.Progress != nil {
		progress["progress-done"] = st.Progress.Done
		progress["progress-total"] = st.Progress.Total
		if st.Progress.SubProgress != nil {
			progress["subprogress-done"] = st.Progress.SubProgress.Done
			progress["subprogress-total"] = st.Progress.SubProgress.Total
		}
	}

	if st.Message != "" {
		logger.WithFields(progress).Infof("OSBuild status: %s", st.Message)
		if job != nil {
			partial := worker.JobResult{
				Progress: &worker.JobProgress{
					Message: st.Message,
				},
			}
			if st.Progress != nil {
				partial.Progress.Done = st.Progress.Done
				partial.Progress.Total = st.Progress.Total
				if st.Progress.SubProgress != nil {
					partial.Progress.SubProgress = &worker.JobProgress{
						Message: st.Progress.SubProgress.Message,
						Done:    st.Progress.SubProgress.Done,
						Total:   st.Progress.SubProgress.Total,
					}
				}
			}

			// non-blocking send: skip this update if consumer is not ready
			select {
			case updates <- partial:
			default:
			}
		}
	}

	if st.Trace != "" {
		logger.Errorf("%s", st.Trace)
	}
}

// wait until consumer processed in-flight update (if any) and exited
wg.Wait()
return osbuildStatus.Result()
```

lzap (changes requested):
The main loop could have been simpler.

cmd/osbuild-worker-executor/handler_build.go
```diff
@@ -39,63 +41,96 @@ func (wf *writeFlusher) Write(p []byte) (n int, err error) {
 	return n, err
 }
 
-func runOsbuild(logger *logrus.Logger, buildDir string, control *controlJSON, output io.Writer) (string, error) {
-	flusher, ok := output.(http.Flusher)
-	if !ok {
-		return "", fmt.Errorf("cannot stream the output")
-	}
-	// stream output over http
```
mvo5, line 47: This is probably an extremly stupid question (sorry!) - why do we need this change? What would happen if we just stream the json-seq in exactly the same way as we used to stream the (previous) text output?

internal/osbuildexecutor/runner-common.go
```diff
@@ -0,0 +1,78 @@
...
+				progress["subprogress-done"] = st.Progress.SubProgress.Done
+				progress["subprogress-total"] = st.Progress.SubProgress.Total
+			}
+		}
+		if st.Message != "" {
+			logger.WithFields(progress).Infof("OSBuild status: %s", st.Message)
+			if job != nil && time.Since(lastUpdated) > MinTimeBetweenUpdates {
+				lastUpdated = time.Now()
+				wg.Add(1)
+				go func() {
```
mvo5: Sorry, another probably extremely stupid question - given that we already only update every 30s, how long does job.Update() usually take? I mean, could we simply do this synchronously (might be worth a tiny comment here or in the commit message why we don't). 

And maybe extracting a small helper function for the update might be nice.

croissanne: Update does make an http request, but I guess it shouldn't take longer than a few hundred ms at the most.

internal/osbuildexecutor/runner-common.go
croissanne: I think writing it like that will result in a deadlock, https://gist.github.com/croissanne/f83f242c1c8447b4ff52deffcef9ef84. Will look at it a bit though.

cmd/osbuild-worker-executor/handler_build.go
croissanne: So in all other cases we use the separate FD to make sure we don't get rogue sprint statements that could upset the scanner. Could not do it here but I do think it makes it a bit more fragile?

internal/osbuildexecutor/runner-common.go
croissanne: It might be worth just doing it flat, just as the buffer is dynamic anyway now.

cmd/osbuild-worker-executor/handler_build.go
mvo5: Right, sorry, what I mean is if we could just use "io.Copy" to do the cpying, something like:
```go
diff --git a/cmd/osbuild-worker-executor/handler_build.go b/cmd/osbuild-worker-executor/handler_build.go
index eb66f03f3..eff8dbf14 100644
--- a/cmd/osbuild-worker-executor/handler_build.go
+++ b/cmd/osbuild-worker-executor/handler_build.go
@@ -2,7 +2,6 @@ package main
 
 import (
        "archive/tar"
-       "bufio"
        "encoding/json"
        "errors"
        "fmt"
@@ -87,26 +86,9 @@ func runOSBuild(logger *logrus.Logger, buildDir string, control *controlJSON, ou
        }
 
        wf := writeFlusher{w: output, flusher: flusher}
-       reader := bufio.NewReader(rPipe)
        wPipe.Close()
-       for {
-               line, err := reader.ReadBytes('\n')
-               if err != nil && err != io.EOF {
-                       return "", fmt.Errorf("cannot read bytes from monitor pipe: %w", err)
-               }
-
-               // handle the case where we have a (final) line without a \n, here we
-               // get a EOF but still have line content to process before we can exit
-               // Note that we have not seen this in practise, its for compat with the
-               // previous bufio.Scanner implementation that had this behavior.
-               if err == io.EOF && len(line) == 0 {
-                       break
-               }
-
-               _, err = wf.Write(line)
-               if err != nil {
-                       return "", fmt.Errorf("cannot write bytes to writeFlusher: %w", err)
-               }
+       if _, err := io.Copy(&wf, rPipe); err != nil && err != io.EOF {
+               return "", err
        }
 
        if err := cmd.Wait(); err != nil {
```
maybe? (I have not tested but I can if you want me to). I.e. I wonder if we need the manual bufio in between or if this is "good enough" ?

croissanne: ok yea that's way cleaner

internal/osbuildexecutor/runner-common.go
lzap: Good point, the channel in the example is not closed, this was an overlook. But I still feel this is easier to read and does not spawn too many goroutines.

internal/osbuildexecutor/runner-impl-host.go
```diff
@@ -1,13 +1,60 @@
...
+	opts.Stdout = &stdoutBuffer
+
+	cmd := osbuild.NewOSBuildCmd(manifest, opts)
+	osbuildStatus := osbuild.NewStatusScanner(rPipe)
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("error starting osbuild: %v", err)
+	}
+	wPipe.Close()
+
+	if err := handleProgress(osbuildStatus, logger, job); err != nil {
```
mvo5, line 39: :heart:

internal/osbuildexecutor/runner-common.go
```diff
@@ -0,0 +1,70 @@
...
+			progress["progress-done"] = st.Progress.Done
+			progress["progress-total"] = st.Progress.Total
+			if st.Progress.SubProgress != nil {
+				progress["subprogress-done"] = st.Progress.SubProgress.Done
+				progress["subprogress-total"] = st.Progress.SubProgress.Total
+			}
+		}
+		if st.Message != "" {
+			logger.WithFields(progress).Infof("OSBuild status: %s", st.Message)
+			if job != nil && time.Since(lastUpdated) > MinTimeBetweenUpdates {
```
mvo5: (super nitpick) maybe a 
``go
			if job != nil && time.Since(lastUpdated) < MinTimeBetweenUpdates {
                   continue
            }
```
here is slightly nicer as the if block is relatively long and following what is part of it and what isn't is not immediately obvious. But maybe I'm overthinking this :)

File: cmd/osbuild-worker-executor/handler_build_test.go [hunk 10]
```diff
@@ -84,12 +84,18 @@ func TestBuildIntegration(t *testing.T) {
...
-	// osbuild is called with --export tree and then the manifest.json
 	restore := main.MockOsbuildBinary(t, fmt.Sprintf(`#!/bin/sh -e
+# make sure the monitor is setup correctly
+>&3 echo '^^{"message": "osbuild-stage-message 1"}'
+>&3 echo '^^{"message": "osbuild-stage-message 2"}'
+>&3 echo '^^{"message": "osbuild-stage-message 3"}'
+
 # echo our inputs for the test to validate
-echo fake-osbuild "$1" "$2" "$3" "$4" "$5" "$6" "$7"
+echo osbuild $@
```
mvo5, line 94: \o/ thank you

internal/osbuildexecutor/runner-common.go
```diff
@@ -36,30 +36,31 @@ func handleProgress(osbuildStatus *osbuild.StatusScanner, logger logrus.FieldLog
...
-					partial.Progress.Total = st.Progress.Total
-					// more than 1 level of subprogress is not expected, just
-					// pipelines and stages.
-					if st.Progress.SubProgress != nil {
-						partial.Progress.SubProgress = &worker.JobProgress{
-							Message: st.Progress.SubProgress.Message,
-							Done:    st.Progress.SubProgress.Done,
-							Total:   st.Progress.SubProgress.Total,
-						}
+			if job == nil || time.Since(lastUpdated) < MinTimeBetweenUpdates {
```
mvo5, line 39: (nitpick) might be nicer to move this change directly two commits earlier when this is indroduced but I have not looked deep, if there is a different reason to do it this way thats fine!

croissanne: oh woops

croissanne: rebased properly this time :)

internal/osbuildexecutor/runner-common.go
```diff
@@ -0,0 +1,71 @@
...
+					}
+				}
+			}
+			err := job.Update(partial)
+			if err != nil {
+				logger.Errorf("Unable to update job: %s", err.Error())
+			}
+		}
+		if st.Trace != "" {
+			logger.Debugf("%s", st.Trace)
```
lzap, line 67: Maybe to use `("OSBuild status: %s", st.Trace)` to match the Info message? Naked logging statements can be confusing.

supakeen: This comment is currently blocking the merge, I'll let you decide if it's important enough to do so :)

lzap (approved):
Sorry I did not mean to hold this off from merging, but since it was rebased (keys, now tests were red again) we need new approvals.

PR #4904, opened 2025-10-22
Add new job type for generating manifests using image-builder-cli [HMS-9049]

achilleas-k:
Add new job type, for generating manifests using image-builder-cli.  Support enabling the functionality using the `IMAGE_BUILDER_EXPERIMENTAL` environment variable.

Currently this doesn't support everything we need.  Resolver options that require credentials and certificates aren't supported because we need a way to pass them to image-builder-cli.  Also, long term, this feature is meant to support building images from bootc containers, which will require pulling the container to inspect it.  We'll have to find a way to share that container with the build job to avoid pulling twice.

~~DRAFT: I should also add an integration test to build an actual image with this option enabled.~~

### DIFF
--- a/test/cases/api/common/common.sh
+++ b/test/cases/api/common/common.sh
@@ -75,16 +75,19 @@ function _instanceCheck() {
         exit 1
     fi
 
-    if [ -n "$OPENSCAP_CUSTOMIZATION_BLOCK" ]; then
-        if ! grep -q "image-builder.insights.compliance-profile-id: pci-dss" <<< "$FACTS"; then
-            echo "System doesn't contain the expected image-builder.insights facts (profile-id)"
-            echo "$FACTS"| grep image-builder
-            exit 1
-        fi
-        if ! grep -q "image-builder.insights.compliance-policy-id: 1af6cced-581c-452c-89cd-33b7bddb816a" <<< "$FACTS"; then
-            echo "System doesn't contain the expected image-builder.insights facts (policy-id)"
-            echo "$FACTS"| grep image-builder
-            exit 1
+    # NOTE: workaround until https://issues.redhat.com/browse/HMS-9822 is resolved (Set OSCAP RHSM facts automatically)
+    if [[ "${IMAGE_BUILDER_EXPERIMENTAL}" == "" ]]; then
+        if [ -n "$OPENSCAP_CUSTOMIZATION_BLOCK" ]; then
+            if ! grep -q "image-builder.insights.compliance-profile-id: pci-dss" <<< "$FACTS"; then
+                echo "System doesn't contain the expected image-builder.insights facts (profile-id)"
+                echo "$FACTS"| grep image-builder
+                exit 1
+            fi
+            if ! grep -q "image-builder.insights.compliance-policy-id: 1af6cced-581c-452c-89cd-33b7bddb816a" <<< "$FACTS"; then
+                echo "System doesn't contain the expected image-builder.insights facts (policy-id)"
+                echo "$FACTS"| grep image-builder
+                exit 1
+            fi
         fi
     fi
 


internal/worker/ibcli-exec.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// TODO: extend to include all options
+	//  - ostree
+	//  - registration
+	//  - bootc
+	//  - seed
+	//  - repositories
+}
+
+func RunImageBuilderManifest(args ImageBuilderArgs, extraEnv []string, errorWriter io.Writer) ([]byte, error) {
```
lzap, line 92: Why to buffer stdout but not stderr? I mean, I would definitely use `io.Writer` for both. Sanne is working on log streaming and not buffering seems like a better fit in here, also more consistent and on top of that allocates less memory possibly.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -0,0 +1,44 @@
...
+
+	result := &worker.ImageBuilderManifestJobResult{
+		Manifest: nil,
+		ManifestInfo: worker.ManifestInfo{
+			OSBuildComposerVersion: common.BuildVersion(),
+		},
+	}
+	// Always update with a result
+	defer func() {
+		err := job.Update(&result)
```
lzap: This error is thrown away, was that intentional?

internal/cloudapi/v2/server.go
```diff
@@ -61,6 +61,12 @@ type Server struct {
 type ServerConfig struct {
 	TenantProviderFields []string
 	JWTEnabled           bool
+
+	// Experimental configuration option. Can only be set through the
+	// IMAGE_BUILDER_EXPERIMENTAL environment variable using:
+	//
+	//     IMAGE_BUILDER_EXPERIMENTAL="image-builder-manifest-generation"
+	ImageBuilderManifestGeneration bool
```
lzap, line 69: Since this is a new type, why this experimental thing? Could it break something?

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Well, it gets logged.  But at this point there's not much else that can be done about it.  This is the same pattern that we have in all the other `Run()` implementations.

https://github.com/osbuild/osbuild-composer/blob/a50f6f965fb42dd4069dc2f7992d971609f1f5b4/cmd/osbuild-worker/jobimpl-depsolve.go#L133-L138

internal/cloudapi/v2/server.go
achilleas-k: It's not fully featured yet (see PR description).  So I'd like for it to only be used in CI for testing for a while.  It's meant to replace the resolve and manifests jobs.

internal/worker/ibcli-exec.go
achilleas-k: The output of the call is a single blob of json (the manifest) that we want to pass on to osbuild.  I think a buffer for the whole thing makes sense in that case, doesn't it?  I don't see a point letting the caller pass a writer for the output, it's only ever a one shot result at the end.

As for stderr, I just went with what we do currently in the osbuild call, which is let the caller decide.

achilleas-k:
Integration test added.  CI runners are having issues though (see #4909).

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -0,0 +1,118 @@
...
+
+			Stages json.RawMessage `json:"stages"`
+		} `json:"pipelines"`
+
+		Sources json.RawMessage `json:"sources"`
+	}
+
+	var pm partialManifest
+	decoder := json.NewDecoder(bytes.NewReader(rawManifest))
+	decoder.DisallowUnknownFields()
```
mvo5: I wonder if we should allow unknown fields here and simplify the partialManifest and just remove the parts we don't need (like Version, Runner, Stages, Sources). I have not checked but as its written it would currently fail for  manifests that declare their exports (cf. https://github.com/osbuild/osbuild/pull/2198)?

It is probably a bit academic so feel free to ignore.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -0,0 +1,118 @@
...
+// to the build pipelines list. The rest are added to the payload pipelines.
+func parseManifestPipelines(rawManifest []byte) (*worker.PipelineNames, error) {
+	if len(rawManifest) == 0 {
+		return nil, fmt.Errorf("manifest is empty")
+	}
+
+	// Partial manifest structure for extracting pipeline names. Non-string
+	// properties we don't care about are parsed as json.RawMessage.
+	type partialManifest struct {
+		Version   string `json:"version"`
```
mvo5, line 88: (super nitpick) Should we double check the version here and error if its anything other than v2 (or is v1 compatible enough)? Probably also a bit accademtic (sorry!)

cmd/osbuild-worker/jobimpl-image-builder-manifest_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"testing"
+
+	main "github.com/osbuild/osbuild-composer/cmd/osbuild-worker"
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseManifestPipelines(t *testing.T) {
```
mvo5, line 11: ❤️

internal/cloudapi/v2/server.go
```diff
@@ -315,6 +321,54 @@ func (s *Server) enqueueCompose(irs []imageRequest, channel string) (uuid.UUID,
...
+	}
+
+	manifestJob := worker.ImageBuilderManifestJob{
+		Args: args,
+
+		// NOTE: image-builder doesn't support setting the rpmmd cache and tries to
+		// read XDG_CACHE_HOME, which fails when running as _osbuild-composer
+		// TODO: make sure we use the same rpmmd cache that's used by the depsolve
+		// job for consistency
+		ExtraEnv: []string{"XDG_CACHE_HOME=/var/cache/osbuild-composer/rpmmd"},
```
mvo5, line 360: Thanks for this comment, lets make sure that we have a --cachdir in ibcli for you :)

test/cases/api.sh
```diff
@@ -903,5 +903,13 @@ TOKEN="$(curl --request POST \
 sudo systemctl restart osbuild-remote-worker@localhost:8700.service
 sudo systemctl is-active --quiet osbuild-remote-worker@localhost:8700.service
 
+if [[ "${IMAGE_BUILDER_EXPERIMENTAL}" != "" ]]; then
+    if echo "${IMAGE_BUILDER_EXPERIMENTAL}" | grep -q "image-builder-manifest-generation=1"; then
+        greenprint "Verifying usage of experimental job type"
+        # verify using log message
```
mvo5, line 910: Smart!

mvo5 (commented):
Very nice! Some quick drive-by musing, very excited about this feature!

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Good idea.  I think we should strictly check for `== "2"`, so we make sure to check this if any other manifest version gets generated.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: I think you're right.  We ignore errors anyway (log them and continue).  If ib-cli somehow returns anything that's not a manifest, we'll get an error even if we deserialise without this option.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Done.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Done.

internal/cloudapi/v2/server.go
mvo5: Would this https://github.com/osbuild/image-builder-cli/pull/358 help you here?

internal/worker/ibcli-exec.go
lzap: I was merely thinking creating the buffer one stack frame earlier but I guess this function will not be heavily shared.

achilleas-k:
Rebased on #4909 to get the new terraform configs and get this thing tested.

internal/worker/ibcli-exec.go
achilleas-k: oh I see.  yeah, it wont.

internal/worker/ibcli-exec_test.go
```diff
@@ -85,11 +87,31 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
...
 				actualCall = append([]string{name}, arg...)
 
+				// The blueprint path is a random temporary directory, so let's
+				// search for it and replace the path in the expected args
+				bpPathIdx := slices.Index(actualCall, "--blueprint") + 1
+				if bpPathIdx > 0 {
+					bpPath := actualCall[bpPathIdx]
+					expCall[bpPathIdx] = bpPath
+
+					// load the blueprint and compare it with the original
```
mvo5: (super nitpick) maybe `// load the blueprint so that we can compare with the original` or move the check here
```go
diff --git a/internal/worker/ibcli-exec_test.go b/internal/worker/ibcli-exec_test.go
index 169aec5cf..39eca5904 100644
--- a/internal/worker/ibcli-exec_test.go
+++ b/internal/worker/ibcli-exec_test.go
@@ -110,6 +110,7 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
                                        bpFileContents, err := io.ReadAll(bpFile)
                                        assert.NoError(err)
                                        assert.NoError(json.Unmarshal(bpFileContents, &onDiskBP))
+                                       assert.Equal(tc.args.Blueprint, onDiskBP)
                                }
 
                                // return a real exec.Command() result so that the output
@@ -122,7 +123,6 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
 
                        assert.Equal(expCall, actualCall)
                        assert.Subset(cmd.Env, tc.extraEnv)
-                       assert.Equal(tc.args.Blueprint, onDiskBP)
                })
        }
 }
```
because when I read this I was looking for the compare but its only a few lines later so I was slightly confused. But feel free to ignore.

internal/cloudapi/v2/server.go
```diff
@@ -315,6 +315,53 @@ func (s *Server) enqueueCompose(irs []imageRequest, channel string) (uuid.UUID,
 	return id, nil
 }
 
+func (s *Server) enqueueComposeIBCLI(irs []imageRequest, channel string) (uuid.UUID, error) {
+	var osbuildJobID uuid.UUID
+	if len(irs) != 1 {
+		return osbuildJobID, HTTPError(ErrorInvalidNumberOfImageBuilds)
```
mvo5: (super nitpick) might be nice to provide a bit more information here, something like:
```go
diff --git a/internal/cloudapi/v2/server.go b/internal/cloudapi/v2/server.go
index 7fd361516..080bfdf48 100644
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -325,7 +325,7 @@ func (s *Server) enqueueComposeIBCLI(irs []imageRequest, channel string) (uuid.U
        logrus.Warnf("using experimental job type: %s", worker.JobTypeImageBuilderManifest)
        var osbuildJobID uuid.UUID
        if len(irs) != 1 {
-               return osbuildJobID, HTTPError(ErrorInvalidNumberOfImageBuilds)
+               return osbuildJobID, HTTPErrorWithInternal(ErrorInvalidNumberOfImageBuilds, fmt.Errorf("expected 1 image request, got %d", len(irs)))
        }
        ir := irs[0]
 
```
but it seems enqueCompse() has the same pattern so feel free to ignore.

internal/cloudapi/v2/server.go
mvo5: Maybe worth a small comment now that we have https://github.com/osbuild/image-builder-cli/pull/358 open that this can be replaced with --rpmmd-cache ? Unless we want to use it this way of course :)

internal/worker/ibcli-exec.go
```diff
@@ -0,0 +1,80 @@
...
+type ImageBuilderArgs struct {
+	Distro    string
+	Arch      string
+	ImageType string
+	Blueprint *blueprint.Blueprint
+
+	// TODO: extend to include all options
+	//  - ostree
+	//  - registration
+	//  - bootc
```
mvo5, line 85: Once this landed (and we added a way to pass a bootc-ref) we would be able to build bootc images in the service, right? That is exciting :)

internal/cloudapi/v2/v2_disk_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func TestComposeDiskCustomizationsValidation(t *testing.T) {
-	srv, _, _, cancel := newV2Server(t, t.TempDir(), false, false)
+	srv, _, _, cancel := newV2Server(t, t.TempDir(), false, false, false)
```
mvo5: This becomes a bit unwieldy, maybe a good time to switch to a `opts *v2ServerOpts` approach? Here is a diff with what we could potentially do here:
[v2serveropts.diff.txt](https://github.com/user-attachments/files/23257283/v2serveropts.diff.txt)

(could be a followup of course)

cmd/osbuild-composer/composer.go
```diff
@@ -173,6 +174,12 @@ func (c *Composer) InitAPI(cert, key string, enableTLS bool, enableMTLS bool, en
 		TenantProviderFields: c.config.Koji.JWTTenantProviderFields,
 	}
 
+	// handle experimental image-builder manifest generation option using the
+	// experimentalflags pkg from osbuild/images.
+	if experimentalflags.Bool("image-builder-manifest-generation") {
```
mvo5, line 179: (super nitpick) I personally would have combined this with the previous commit that added `ImageBuilderManifestGeneration` to the config but up to you how granular you want this.

.gitlab-ci.yml
```diff
@@ -605,6 +605,12 @@ API:
           - aws/rhel-10.0-ga-aarch64
         INTERNAL_NETWORK: ["true"]
 
+      # single test config for experimental feature
```
mvo5, line 593: (super nitpick) I wonder if this and the previous one (or even also the previous-previous one) commits could be combined but very subjective and fine as is.

mvo5 (dismissed):
Thank you! Some small suggestions inline but all could be followups (or ignored, some are really just subjective) I think.

internal/worker/ibcli-exec_test.go
achilleas-k: yeah, that's fine.  In my mind I was just separating "catch the call" from "check the result", but there's no real reason to do it.

internal/cloudapi/v2/server.go
achilleas-k: No we should definitely use the option when we get it.  I think we still haven't decided exactly how we're going to get ib-cli on the workers yet, so we'll need to see when that new option becomes available.  But I'll definitely add the note.

internal/worker/ibcli-exec.go
achilleas-k: Yup, that's the plan.  The thing I want to work on after this is bootc manifest generation and the "passing of the container" from the depsolve worker to the build worker.  Which doesn't seem trivial.

.gitlab-ci.yml
achilleas-k: I think I'm going to add repository handling here and I need to fix up the test to install ib-cli, so I'll do some history rewriting to reduce the commit count, sure.

internal/cloudapi/v2/v2_disk_test.go
achilleas-k: I considered but got lazy because it's just test stuff.  But you're right :)

cmd/osbuild-composer/composer.go
achilleas-k: Done.

.gitlab-ci.yml
achilleas-k: Done.

internal/cloudapi/v2/v2_disk_test.go
achilleas-k: Added.  ty!

internal/worker/ibcli-exec.go
```diff
@@ -0,0 +1,75 @@
...
+
+		if _, err := bpFile.Write(bp); err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to write blueprint: %w", err)
+		}
+		if err := bpFile.Close(); err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to close blueprint file: %w", err)
+		}
+
+		// TODO: catch and log remove errors
+		defer os.Remove(bpFile.Name())
```
thozza: You should ideally defer `Close()` and `Remove()` as a common cleanup action right after calling `CreateTemp()`. Right now, the opened file handle can leak i.e. if `json.Marshal()` returns an error. In such a case, the file handle would not be closed, and the file would not get removed.

cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -458,6 +458,13 @@ func (impl *OSBuildJobImpl) Run(job worker.Job) error {
 
 	// copy pipeline info to the result
 	osbuildJobResult.PipelineNames = jobArgs.PipelineNames
+	if osbuildJobResult.PipelineNames == nil {
+		// jobArgs doesn't provide the pipeline names when the manifest was
+		// generated using image-builder-cli. In this case, the manifest job
+		// result itself should have the pipeline names (under ManifestInfo)
+		// parsed from the manifest itself.
+		osbuildJobResult.PipelineNames = manifestInfo.PipelineNames
+	}
```
lzap, line 467: Out of curiosity, is there a reason to assign first and then check for nil versus checking the `jobArgs.PipelineNames` instead?

lzap (approved):

cmd/osbuild-worker/jobimpl-osbuild.go
achilleas-k: No reason.  Just felt natural at the time.
Do you find it's less readable this way?  I can change it.

internal/worker/export_test.go
```diff
@@ -0,0 +1,13 @@
+package worker
+
+import "os/exec"
+
+// MockExecCommand replaces the exec.Command() wrapper and returns a function
+// that can be called to restore the original.
+func MockExecCommand(mock func(name string, arg ...string) *exec.Cmd) func() {
```
thozza: Nitpick: we tend to name the `restore` function value, to make the purpose of the function self-explanatory (i.e. https://github.com/osbuild/osbuild-composer/blob/c6f40834d38c46c289ad2ff54cac7e735f644e89/internal/cloudapi/v2/export_test.go#L14).

internal/worker/ibcli-exec.go
```diff
@@ -25,6 +25,11 @@ type ImageBuilderArgs struct {
 	//  - repositories
 }
 
+// var wrapper for exec.Command() that can be mocked for testing.
+var execCommand = func(name string, arg ...string) *exec.Cmd {
+	return exec.Command(name, arg...)
+}
```
thozza: Nitpick: You could simply assign `exec.Command` to `execCommand`, without wrapping it in an additional function, no?

internal/worker/ibcli-exec_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			var actualCall []string
+			var cmd *exec.Cmd
+			worker.MockExecCommand(func(name string, arg ...string) *exec.Cmd {
```
thozza: You never defer the restore function, which is also thrown away on each call.

internal/worker/json.go
```diff
@@ -736,3 +736,16 @@ type AWSEC2CopyJobResult struct {
 	Ami    string `json:"ami"`
 	Region string `json:"region"`
 }
+
+// ImageBuilderManifestJob generates a manifest from a build request using
+// image-builder-cli. Includes resolving all content types.
+type ImageBuilderManifestJob struct {
+	// Arguments to the image-builder command line
+	Args ImageBuilderArgs
```
thozza, line 758: I'm sure that you are aware of this, but with this change, the `blueprint.Blueprint` becomes part of the Worker API, so any changes to it will affect the worker. It shouldn't be a problem, since the package is intended for user-facing serialization.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (impl *ImageBuilderManifestJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id().String())
+
+	result := &worker.ImageBuilderManifestJobResult{
+		Manifest: nil,
+		ManifestInfo: worker.ManifestInfo{
+			OSBuildComposerVersion: common.BuildVersion(),
+		},
```
thozza, line 28: We should actually add a new field for `image-builer` and report ibcli version and embedded `osbuild/images` version instead. The goal here is to provide additional context for the generated manifest. Since it is being generated by `ibcli` and not `osbuild-composer`, the composer build version does not add much value.

cmd/osbuild-worker/main.go
```diff
@@ -516,6 +516,10 @@ var run = func() {
 		worker.JobTypeAWSEC2Share: &AWSEC2ShareJobImpl{
 			AWSCreds: awsCredentials,
 		},
+		worker.JobTypeImageBuilderManifest: &ImageBuilderManifestJobImpl{
+			// TODO: needs the same things as the depsolve and container
+			// resolver jobs
+		},
```
thozza: You could add the job to the list above instead, together with the existing depsolve and search jobs, so that it can run in parallel with jobs that actually build images.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -40,5 +43,75 @@ func (impl *ImageBuilderManifestJobImpl) Run(job worker.Job) error {
...
+	result.ManifestInfo.PipelineNames = pipelineNames
+
 	return nil
 }
+
+// Parse the raw manifest into an incomplete representation of the manifest
+// structure in order to retrieve the pipeline names in order. Any pipeline
+// names that appear under the 'build' property of another pipeline are added
+// to the build pipelines list. The rest are added to the payload pipelines.
+func parseManifestPipelines(rawManifest []byte) (*worker.PipelineNames, error) {
```
thozza, line 81: Would it make sense to dump this additional metadata in some form, in addition to the manifest? Maybe together with some of the info that we are then missing when uploading a built image?

Or we could simply add this information to the manifest itself under the `metadata` field -> https://github.com/osbuild/osbuild/blob/64f87a5fd0f7e3c86c78ef2a5cea30bd06dce209/schemas/osbuild2.json#L89-L90

internal/worker/json.go
```diff
@@ -640,6 +640,12 @@ type ManifestInfo struct {
 	// List of relevant modules used by osbuild-composer which
 	// could affect the manifest content.
 	OSBuildComposerDeps []*OSBuildComposerDepModule `json:"osbuild_composer_deps,omitempty"`
+
+	// Ordered list of pipeline names in the manifest, separated into build and
+	// payload pipelines. These are parsed from the manifest itself and copied
+	// to the osbuild job result so it can properly order the osbuild job log
+	// by pipeline execution order.
+	PipelineNames *PipelineNames `json:"pipeline_names,omitempty"`
```
thozza, line 656: It should be possible to simplify the flow and use this field even for manifests generated on the server side by composer. Something to consider for the future...

internal/worker/ibcli-exec.go
```diff
@@ -58,12 +70,37 @@ func RunImageBuilderManifest(args ImageBuilderArgs, extraEnv []string, errorWrit
...
+		if err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to serialize repositories: %w", err)
+		}
+
+		if _, err := reposFile.Write(repos); err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to write repositories: %w", err)
+		}
+		if err := reposFile.Close(); err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to close blueprint file: %w", err)
+		}
```
thozza: ditto: should be ideally deferred to close the file in case of errors.

schutzbot/deploy.sh
```diff
@@ -136,6 +136,28 @@ retry sudo dnf -y install osbuild-tools
...
+
+    cat experimental-override.conf | sudo systemctl edit --stdin osbuild-composer.service
+    sudo systemctl daemon-reload
+    sudo systemctl cat osbuild-composer.service
+
+    if echo "${IMAGE_BUILDER_EXPERIMENTAL}" | grep -q "image-builder-manifest-generation=1"; then
+        # TODO: configure dependency pinning for image-builder like we do for osbuild
+        # https://issues.redhat.com/browse/HMS-9647
+        greenprint "Installing image-builder for experimental manifest generation"
+        sudo dnf -y install image-builder
```
thozza, line 156: We should also figure out a way to set the minimum required `image-builder` version in the SPEC file. The feature obviously relies on specific features in `ibcli`, which were added fairly recently (like the search for repos under `repository/` in the `--data-dir`).

cmd/osbuild-worker/jobimpl-osbuild.go
thozza: You should amend the comment to note the IBCLI manifest job

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Yeah, I think both could be valid.  This was a bit of a workaround to get the PoC up and running, but we shouldn't be doing guesswork in the long run.

schutzbot/deploy.sh
achilleas-k: Yes, definitely.  I held off making any such changes since it's "experimental" and service-only, but we'll need to at least make it an optional dependency at first with a minimum version.

achilleas-k:
> This must have been a lot of work 👏
> 
> Besides my inline comments, I think that it would be good to enable ibcli manifest generation on most of the CloudAPI unit tests. The reason is that at least two of them will break with this change when using ibcli to generate manifests.
> 
> * `/composes/{id}/manifests` - it considers only the `worker.JobTypeManifestIDOnly` job type when searching for the manifest in the osbuild job dependencies, see https://github.com/osbuild/osbuild-composer/blob/a50f6f965fb42dd4069dc2f7992d971609f1f5b4/internal/cloudapi/v2/handler.go#L823
> * `/composes/{id}/metadata` - it relies on the pipeline names being set in the osbuild job arguments, see https://github.com/osbuild/osbuild-composer/blob/a50f6f965fb42dd4069dc2f7992d971609f1f5b4/internal/cloudapi/v2/handler.go#L684

Thank you!  I didn't even consider looking into those.

cmd/osbuild-worker/jobimpl-osbuild.go
lzap: This is okay. It is one way of doing it for sure :-)

internal/worker/json.go
achilleas-k: Yes, and we need to think about this more.  I think we're fine in general, we've been very good about keeping the blueprint stable, but I think this goes hand-in-hand with making ib-cli a versioned dependency.

Maybe it's better if the blueprint is serialised in composer instead and passed on as raw json instead.  WDYT?

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
achilleas-k: Let's leave this for a follow-up, if you don't mind.  I'd like to add `--json` output to the `image-builder --version` command so we can parse this properly instead of dumping the regular output in text or trying to parse it.  It also gives us the images library version that it uses:
```
# image-builder --version
image-builder:
  version: "39"
  commit: unknown
  dependencies:
    images: v0.206.0
    osbuild: "163"
```

cmd/osbuild-worker/main.go
achilleas-k: ah of course!  ty!

internal/worker/json.go
achilleas-k: I like the current state where images defines them itself while it's generating the manifest, but that doesn't carry over to osbuild-composer when using ib-cli.  So maybe we could start using the manifest metadata.  I know @mvo5 and @supakeen were fond of this idea too and I resisted a bit, but now that we're here, I think it's necessary.

achilleas-k:
Marked ready again.  I think I covered everything.

I started working on an integration test that mocks osbuild-depsolve-dnf, osbuild, and image-builder-cli to test the queue without mocking internal functions, only the external commands, but that got a bit too big and I decided to leave it for a separate PR.

> * `/composes/{id}/metadata` - it relies on the pipeline names being set in the osbuild job arguments, see

I think this is fine because we add the pipeline names to the osbuild job results through the ib-manifest job already and it should all work.  I tested it manually, but the way the unit tests now (if I'm reading them all correctly) rely a lot on mocks and I'm not sure if there's a test that covers this directly.

internal/worker/ibcli-exec.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if args.Blueprint != nil {
+		// image-builder can read blueprints in JSON format and uses the extension to detect
+		bpFile, err := os.CreateTemp("", "image-builder-blueprint-*.json")
+		if err != nil {
+			return nil, fmt.Errorf("image-builder manifest: failed to create blueprint file: %w", err)
+		}
+		defer os.Remove(bpFile.Name())
+		defer bpFile.Close()
```
lzap: Why this does work on Linux, closing and then removing is cleaner.

internal/worker/ibcli-exec_test.go
```diff
@@ -85,11 +87,33 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
...
 			assert := assert.New(t)
 
+			expCall := tc.expCall
+
 			var actualCall []string
 			var cmd *exec.Cmd
 			restoreExec := worker.MockExecCommand(func(name string, arg ...string) *exec.Cmd {
 				actualCall = append([]string{name}, arg...)
 
+				// The blueprint path is a random temporary directory, so let's
```
lzap: Temporary file, is it?

internal/cloudapi/v2/v2_test.go
```diff
@@ -138,6 +139,57 @@ func mockDepsolve(t *testing.T, workerServer *worker.Server, wg *sync.WaitGroup,
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			_, token, _, _, _, err := workerServer.RequestJob(ctx, test_distro.TestArchName, []string{worker.JobTypeImageBuilderManifest}, []string{""}, uuid.Nil)
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
lzap, line 156: This can be more simple:

```
if err := ctx.Err(); err != nil {
    return
}
```

internal/worker/ibcli-exec.go
achilleas-k: Isn't that what we're doing?  Closing first, removing second?

internal/worker/ibcli-exec_test.go
achilleas-k: It's under a temporary directory where the repos also get added, so I guess it's more correct to say that "The Blueprint is created in a temporary directory ...".

internal/cloudapi/v2/v2_test.go
achilleas-k: Copied this from the other mock functions.  I'll have a closer look at it and amend.

internal/worker/ibcli-exec.go
lzap: Yes, you are right. I focused on what happens to unclosed files being removed in Linux so hard that I actually missed the order :-)

internal/worker/json.go
thozza: > Maybe it's better if the blueprint is serialised in composer instead and passed on as raw json instead. WDYT?

I think that may be a reasonable option, similar to how we pass manifests around. Then the Worker API does not need to care about the structure, but the `blueprint` library in composer would handle the backward compatibility.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
thozza: I'm fine with a follow-up, but maybe at least a `// TODO:` or `// XXX: ` comment would be appropriate here.

internal/cloudapi/v2/v2_test.go
```diff
@@ -2253,7 +2342,7 @@ func TestDownloadUnknown(t *testing.T) {
 // TestComposeRequestMetadata tests that the original ComposeRequest is included with the
 // metadata response.
 func TestComposeRequestMetadata(t *testing.T) {
-	srv, wrksrv, _, cancel := newV2Server(t, t.TempDir(), false, false)
+	srv, wrksrv, _, cancel := newV2Server(t, t.TempDir(), nil)
```
thozza: You need to test this endpoint also with `ibcli` generated manifests, because it relies on the pipeline names being set in the osbuild job arguments, which is not the case when manifests are not generated on the server side.

thozza (commented):
AFAICT, the `/composes/{id}/metadata` is still not tested with the ibcli manifest job, which affects it directly. The rest looks good from my PoV.

Great work 👏

internal/worker/json.go
thozza: What I meant is that we could be setting this field in the `ManifestInfo` even when generating the manifest on the server side, instead of adding he information directly to the `osbuild` job arguments. This way, we would have a single place for it (in the `ManifestInfo`) and unify it, regardless of how we generated the manifest.

I didn't mean that we should parse it from the manifest when we are generating the manifest on the server side. We should, however add it to the `ManifestInfo`, instead of the `osbuild` job args.

Does this make more sense now?

internal/cloudapi/v2/v2_test.go
thozza: Checking the PR more thoroughly, it seems that the `func (h *apiHandlers) getComposeMetadataImpl(ctx echo.Context, jobId uuid.UUID) error` function was not amended to the `ibcli` generated manifests.

achilleas-k: Right, I wanted to ask about that.  From what I can tell, that gets the metadata from the osbuild job directly, which should have the same data regardless of the manifest job.  I still want to test it, but I didn't see anything that would be affected.  I might be reading it wrong though.

On the topic of the test though, I could add it.  I was passing when I tried it.  I think the problem with that test is that, since the implementations of all the job types are mocked, there are things that could go wrong (e.g. the real manifest job not setting pipeline names while the mocked job does) that wouldn't be caught by the test.

internal/worker/json.go
achilleas-k: I'll change it.

achilleas-k:
Current gitlab CI failure
```
Nov 10 22:31:40 ip-10-31-19-88.us-east-1.aws.redhat.com osbuild-worker[14753]: error: no repositories found in the given paths: [/tmp/image-builder-manifest-478825411]/[{0x562f6a98eb40}]
```

achilleas-k:
Alright, repos should be fixed now.

internal/cloudapi/v2/v2_test.go
thozza: Here is what I mean:

1. When a regular compose is enqueued, the `worker.OSBuildJob.PipelineNames` is set -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/server.go#L305-L308
2. With `enqueueComposeIBCLI`, the `worker.OSBuildJob.PipelineNames` is not set -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/server.go#L371
3. The `getComposeMetadataImpl()` implementation for the `/composes/{id}/metadata` API endpoint iterates over `job.PipelineNames.Payload`, where `job` is the `worker.OSBuildJob`. That struct should have the `job.PipelineNames` set to `nil` when using `enqueueComposeIBCLI` -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/handler.go#L691

While writing the above, I noticed the problem. The unit test doesn't finish the job https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/v2_test.go#L2408 , which means that we return from the endpoint handler too early https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/handler.go#L664-L672 before hitting the `nil` dereference issue. The test should fail if you let the compose finish.

A reasonable thing to do would be to take the `PipelineNames` from the `OSBuildJobResult`, instead of the job arguments. The handler uses the job arguments to then interpret the job result, so it could also use the job result for the `PipelineNames`, which is always set there (Either copied from the job arguments or parsed from the manifest).

Deprecating the `PipelineNames` from the `worker.OSBuildJob` would make the problem go away naturally.

WRT mocking. We could be testing all of these endpoints also as part of `api.sh`.

achilleas-k: > Here is what I mean:
>
> 1. When a regular compose is enqueued, the `worker.OSBuildJob.PipelineNames` is set -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/server.go#L305-L308
> 2. With `enqueueComposeIBCLI`, the `worker.OSBuildJob.PipelineNames` is not set -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/server.go#L371


> 3. The `getComposeMetadataImpl()` implementation for the `/composes/{id}/metadata` API endpoint iterates over `job.PipelineNames.Payload`, where `job` is the `worker.OSBuildJob`. That struct should have the `job.PipelineNames` set to `nil` when using `enqueueComposeIBCLI` -> https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/handler.go#L691
>
> While writing the above, I noticed the problem. The unit test doesn't finish the job https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/v2_test.go#L2408 , which means that we return from the endpoint handler too early https://github.com/achilleas-k/osbuild-composer/blob/450e5582f076efa3f8079adee5a47b1912bed830/internal/cloudapi/v2/handler.go#L664-L672 before hitting the `nil` dereference issue. The test should fail if you let the compose finish.
>

Ah!  Thanks for this.

> A reasonable thing to do would be to take the `PipelineNames` from the `OSBuildJobResult`, instead of the job arguments. The handler uses the job arguments to then interpret the job result, so it could also use the job result for the `PipelineNames`, which is always set there (Either copied from the job arguments or parsed from the manifest).
>

Ok, this explains my confusion!!
For some reason I was convinced it was getting the pipeline names from the osbuild job RESULTS.

> Deprecating the `PipelineNames` from the `worker.OSBuildJob` would make the problem go away naturally.
>

Yeah, I think I'll do that instead.  Have just one avenue for the pipeline names to go through.  Not sure if I should do it here or as a follow-up.

> WRT mocking. We could be testing all of these endpoints also as part of `api.sh`.

I'll do that just to cover all bases.

achilleas-k:
Rebasing on 8f6ecc42637ffb60ba15e58ee2b91b583a68695f.

achilleas-k:
> ```
> Traceback (most recent call last):
>   File "/usr/bin/osbuild", line 33, in <module>
>     sys.exit(load_entry_point('osbuild==163', 'console_scripts', 'osbuild')())
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/osbuild/main_cli.py", line 183, in osbuild_cli
>     manifest.download(object_store, monitor)
>   File "/usr/lib/python3.12/site-packages/osbuild/pipeline.py", line 453, in download
>     raise e
>   File "/usr/lib/python3.12/site-packages/osbuild/pipeline.py", line 448, in download
>     source.download(mgr, store)
>   File "/usr/lib/python3.12/site-packages/osbuild/sources.py", line 39, in download
>     reply = client.call("download", args)
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/osbuild/host.py", line 348, in call
>     ret, _ = self.call_with_fds(method, args)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/osbuild/host.py", line 384, in call_with_fds
>     raise error
> osbuild.host.RemoteError: RuntimeError: Packages/crontabs-1.11^20190603git9e74f2d-6.el10.noarch.rpm: No more mirrors to try - All mirrors were already tried without success,Packages/passt-selinux-0^20250512.g8ec1341-2.el10.noarch.rpm: No more mirrors to try - All mirrors were already tried without success,Packages/passt-0^20250512.g8ec1341-2.el10.x86_64.rpm: No more mirrors to try - All mirrors were already tried without success
>    File "/usr/lib/python3.12/site-packages/osbuild/host.py", line 268, in serve
>     reply, reply_fds = self._handle_message(msg, fds)
>                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/osbuild/host.py", line 301, in _handle_message
>     ret, fds = self.dispatch(name, args, fds)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/osbuild/sources.py", line 105, in dispatch
>     self.fetch_all(args["items"])
>   File "/usr/lib/osbuild/sources/org.osbuild.librepo", line 268, in fetch_all
>     raise RuntimeError(",".join(self.errors))
> time="2025-11-19T12:36:02Z" level=error msg="osbuild job failed: osbuild build failed" channel=local jobId=c01ecb9a-cc5d-43d3-9a8c-80ceaf180db1
> time="2025-11-19T12:36:02Z" level=error msg="failure details : osbuild did not return any output" channel=local jobId=c01ecb9a-cc5d-43d3-9a8c-80ceaf180db1
> Finished job
> ```
> 
> CI runner is definitely on the internal network. Package is available.

This was a fun one: https://github.com/osbuild/osbuild/pull/2253

thozza:
> ```
> System doesn't contain the expected image-builder.osbuild-composer facts
> image-builder.osbuild-composer.api-type: image-builder-cli
> ```

Yeah, at some point, I realized that we will need to make this fact configurable in `ibcli`. For example, we may want to set it to `koji-image-builder` when used in Koji, etc.

internal/worker/ibcli-exec_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"os"
+	"os/exec"
+	"slices"
+	"testing"
+
+	"github.com/osbuild/osbuild-composer/internal/worker"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRunImageBuilderManifestCall(t *testing.T) {
```
mvo5, line 21: :heart:

internal/worker/ibcli-exec_test.go
```diff
@@ -0,0 +1,144 @@
...
+				// search for it and replace the path in the expected args.
+				// Also, load the blueprint contents to compare them with the
+				// original from the test case.
+				var onDiskBP json.RawMessage
+				bpPathIdx := slices.Index(actualCall, "--blueprint") + 1
+				if bpPathIdx > 0 {
+					bpPath := actualCall[bpPathIdx]
+					expCall[bpPathIdx] = bpPath
+
+					bpFile, err := os.Open(bpPath)
```
mvo5, line 244: (nitpick) I think this could be slightly short by using os.ReadFile():
```go
					bpFileContents, err := os.ReadFile(bpPath)
```

internal/cloudapi/v2/server.go
mvo5: https://github.com/osbuild/image-builder-cli/pull/358 is merged but I guess its fine to do this as a followup, XDG_CACHE_HOME is a fine way of doing this too.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (impl *ImageBuilderManifestJobImpl) Run(job worker.Job) error {
+	logWithId := logrus.WithField("jobId", job.Id().String())
+
+	result := &worker.ImageBuilderManifestJobResult{
+		Manifest: nil,
+		ManifestInfo: worker.ManifestInfo{
+			OSBuildComposerVersion: common.BuildVersion(),
+			// TODO: add image-builder version fields when we get
```
mvo5, line 26: I have good news:
```yaml
$ image-builder --version|yq
image-builder:
  version: "42"
  commit: unknown
  dependencies:
    images: v0.223.0
    osbuild: "167"
```
i.e. "image-buidler --version" is already machine readable yaml :)

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
```diff
@@ -42,5 +45,75 @@ func (impl *ImageBuilderManifestJobImpl) Run(job worker.Job) error {
...
+	result.ManifestInfo.PipelineNames = pipelineNames
+
 	return nil
 }
+
+// Parse the raw manifest into an incomplete representation of the manifest
+// structure in order to retrieve the pipeline names in order. Any pipeline
+// names that appear under the 'build' property of another pipeline are added
+// to the build pipelines list. The rest are added to the payload pipelines.
+func parseManifestPipelines(rawManifest []byte) (*worker.PipelineNames, error) {
```
mvo5, line 81: Sorry, I only noticed this now - I wonder it might be an option to use `image-builder describe-image` here? It also supports showing the pipelines:
```yaml
$ image-builder describe-image qcow2
No distro name specified, selecting "fedora-43" based on host, use --distro to override
@WARNING - the output format is not stable yet and may change
distro: fedora-43
type: server-qcow2
arch: x86_64
os_version: "43"
bootmode: hybrid
partition_type: gpt
default_filename: disk.qcow2
build_pipelines:
  - build
payload_pipelines:
  - os
  - image
  - qcow2
...
```
or
```console
$ image-builder describe-image qcow2|tail -n+2|yq .payload_pipelines[]
No distro name specified, selecting "fedora-43" based on host, use --distro to override
os
image
qcow2
```
and its valid yaml (when cutting line 1), we could also add an env like IMAGE_BUILDER_EXPERIMENTAL=describe-image-no-warnings  to make the warning go away. 


But of course also fine to keep this as is and think about it at some later point :)

internal/cloudapi/v2/v2_test.go
```diff
@@ -234,7 +234,17 @@ func mockSearch(t *testing.T, workerServer *worker.Server, wg *sync.WaitGroup, f
 	return cancel
 }
 
-func newV2Server(t *testing.T, dir string, enableJWT bool, fail bool, ibManifest bool) (*v2.Server, *worker.Server, jobqueue.JobQueue, context.CancelFunc) {
+type v2ServerOpts struct {
```
mvo5, line 289: :heart:

internal/worker/ibcli-exec.go
```diff
@@ -7,20 +7,82 @@ import (
...
 	"os"
 	"os/exec"
+	"path/filepath"
+
+	"github.com/osbuild/images/pkg/rpmmd"
 )
 
+// TODO: remove this and the conversion functions below when
+// https://issues.redhat.com/browse/HMS-9746 is resolved
+type repository struct {
```
mvo5, line 18: This is no longer needed, right? With https://github.com/osbuild/images/pull/2058 merged we can also use "baseurl" []string here? but maybe followup material?

internal/worker/ibcli-exec_test.go
```diff
@@ -128,14 +189,40 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
...
+				datadirIdx := slices.Index(actualCall, "--data-dir") + 1
+				if datadirIdx > 0 {
+					datadir := actualCall[datadirIdx]
+
+					// the datadir in the expected call is just a * glob, so
+					// there's no point checking it for now
+					expCall[datadirIdx] = datadir
+
+					reposPath := filepath.Join(datadir, "repositories", fmt.Sprintf("%s.json", tc.args.Distro))
+					reposFile, err := os.Open(reposPath)
```
mvo5, line 273: (nitpick) same coment as above, `reposFileContents, err := os.ReadFile(reposPath)` here would be probably be slightly nicer (shorter)

internal/worker/ibcli-exec.go
```diff
@@ -206,3 +213,32 @@ func handleRepositories(args ImageBuilderArgs, pardir string) ([]string, error)
...
+	subFile, err := os.Create(filepath.Join(pardir, "subscription.json"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create subscription file: %w", err)
+	}
+	defer subFile.Close()
+
+	// registration file for image-builder must be a map with 'redhat' as a
+	// top-level key and 'subscription' below it
+	subMap := map[string]map[string]subscription.ImageOptions{"redhat": {"subscription": *args.Subscription}}
+	sub, err := json.Marshal(subMap)
```
lzap, line 234: I wonder if we should have a type in `images` imported both by the CLI and from here. This is okay for now.

internal/worker/ibcli-exec_test.go
```diff
@@ -27,7 +27,7 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
 
 	testCases := map[string]testCase{
 		"empty": {
-			expCall: []string{"image-builder", "manifest", "--distro", "", "--arch", "", "--", ""}, // TODO: make this an error
+			expCall: []string{"image-builder", "manifest", "--use-librepo=false", "--distro", "", "--arch", "", "--", ""}, // TODO: make this an error
```
lzap, line 32: Possibly a leftover?

lzap (approved):
`>>> ALL TESTS GREEN <<<`

internal/worker/ibcli-exec.go
```diff
@@ -93,6 +93,7 @@ func RunImageBuilderManifest(args ImageBuilderArgs, extraEnv []string, errorWrit
 
 	clArgs := []string{
 		"manifest",
+		"--use-librepo=false", // TODO: remove once https://github.com/osbuild/osbuild/pull/2253 arrives
```
mvo5, line 98: (nitpick) this was merged 3 weeks ago so we might be able to remove it now (but fine for followup material)

internal/worker/ibcli-exec.go
```diff
@@ -206,3 +213,32 @@ func handleRepositories(args ImageBuilderArgs, pardir string) ([]string, error)
...
+
+	// registration file for image-builder must be a map with 'redhat' as a
+	// top-level key and 'subscription' below it
+	subMap := map[string]map[string]subscription.ImageOptions{"redhat": {"subscription": *args.Subscription}}
+	sub, err := json.Marshal(subMap)
+	if err != nil {
+		return nil, fmt.Errorf("image-builder manifest: failed to serialize subscription options: %w", err)
+	}
+
+	if _, err := subFile.Write(sub); err != nil {
```
mvo5, line 239: Hm, could this just be `if _, err := os.WriteFile(filepath.Join(pardir, "subscription.json"), sub, 0600); err != nil {` and we remove the os.Create() above? It seems if we use os.Create() we might want to create a json encoder to stream directly into the file. But I personally like this and would just move to os.WriteFile() for slightly shorter code.

internal/worker/ibcli-exec_test.go
```diff
@@ -232,6 +288,34 @@ func TestRunImageBuilderManifestCall(t *testing.T) {
...
+
+					// we can't predict the temporary directory name, but we
+					// can make sure it matches the glob
+					g, err := glob.Compile(expPath)
+					assert.NoError(err)
+					assert.True(g.Match(subPath))
+
+					expCall[subPathIdx] = subPath
+
+					subFile, err := os.Open(subPath)
```
mvo5, line 309: Same comment about os.ReadFile(subPath) here

mvo5 (approved):
Thank you! This looks very nice and is a big step forward. I have some ideas/suggestions/questions but all can be followups I think :)

internal/worker/ibcli-exec_test.go
achilleas-k: Which part?  Disabling librepo is necessary until I add version pinning for image-builder and we know exactly which version we're running.

achilleas-k:
Regarding all the comments about workarounds to things already fixed in IB: The one test we have runs on 10.1 and installs IB through the repos.  I left version pinning for testing for a followup (https://issues.redhat.com/browse/HMS-9647).
We also need to make IB a dependency of osbuild-composer (in the `.spec` file) with a minimum version for the features and fixes we rely on (https://issues.redhat.com/browse/HMS-9659).  So for now, this PR, I need to keep the workarounds there.  I didn't want to have to figure out version pinning for testing here as well as it was already quite long.  We might want to add RPM builds in IB's CI like we do with osbuild and osbuild-composer so we can test against specific versions when we need to.

internal/worker/ibcli-exec.go
achilleas-k: Yeah, we should standardise on the `redhat` top level key there.

cmd/osbuild-worker/jobimpl-image-builder-manifest.go
thozza: I'm not against it, but there is a downside, even if we specify the minimum required version of `image-builder` in `osbuild-composer.spec`.

If we were to make a backwards incompatible change in the output, the minimum required version would not prevent us from accidentally updating only `ibcli`, while not updating `osbuild-composer` and breaking the pipelines parsing.

I'm not sure which approach is worse, the current one or parsing `describe-image` output...

internal/cloudapi/v2/v2_test.go
```diff
@@ -2322,18 +2427,20 @@ func TestComposeRequestMetadata(t *testing.T) {
 		"status": "pending"
 	}`, jobId, jobId))
 
-	// metadata response should include the ComposeRequest even when build is not done
-	test.TestRoute(t, srv.Handler("/api/image-builder-composer/v2"), false, "GET", fmt.Sprintf("/api/image-builder-composer/v2/composes/%v/metadata", jobId), ``, http.StatusOK,
-		fmt.Sprintf(`{
+			// metadata response should include the ComposeRequest even when build is not done
```
thozza, line 2430: Nitpick: it would be still good to test the endpoint with the compose being done for the reasons I mentioned in https://github.com/osbuild/osbuild-composer/pull/4904#discussion_r2519430439. But not a blocker from my PoV, because the original issue that would manifest in such a scenario has been fixed.

thozza (approved):
LGTM, let's get this in... 👏

internal/worker/ibcli-exec_test.go
lzap: I meant the TODO but I see it was there before.

Issue #4427, opened 2024-10-23
Build fails because of out of sync mirrors


jbenc:
See https://github.com/osbuild/osbuild-composer/issues/1031 (I can't reopen that one).

The build is failing every single time for me. `journalctl -u osbuild-worker@1` shows:

[osbuild-worker.log](https://github.com/user-attachments/files/17487743/osbuild-worker.log)

Those two mirrors that it picked on random(?) are out of sync and do not have the newest packages.

jbenc:
This is on Fedora 40.

bcl:
The metadata should get refreshed on restart of the system. If that fails you can remove it manually by deleting the `rpmmd` cache directories under `/var/cache/osbuild-worker` and `/var/cache/osbuild-compose`

teodor-pripoae:
I fixed it temporary by hardcoding the mirror in `/usr/share/osbuild-composer/repositories/fedora-40.json` and removing rpmmd cache.

PR #4911, opened 2025-10-31
cloudapi,weldr: enable librepo by default (and update tests)

mvo5:
We could start by only enabling on weldr, it seems
the service very rarely deals with issue about
out-of-sync mirrors.

---

mocks: make BaseDepsolveResult() return a DepsolveResult

The BaseDepsolveResult was only returning the packages part of
the DepsolveResult. With librepo we need the repos as well so
that packages can be linked to repos. This commit returns the
full DepsolveResult with the Packages/Repos linked now.

---

weldr: enable librepo by default (and update tests)

This commit enables librepo by default in the weldr api. This
should help with on-prem out of sync mirror issues like:
https://github.com/osbuild/osbuild-composer/issues/4427

---

cloudapi: enable librepo by default (and update tests)

This commit enables librepo by default in the cloudapi.

We have librepo in osbuild/images/ibcli/bib for a while now, c.f.
https://github.com/osbuild/image-builder-cli/pull/51
https://github.com/osbuild/bootc-image-builder/pull/786 (enable)
https://github.com/osbuild/bootc-image-builder/pull/840 (make default)

It is not configurable right now, I am not sure there is a reason
to keep libcurl around for manifest generation as its more fragile
but if it is desired I can look into that too. Having it configurable
would allow us to revert to the previous org.osbuild.curl in case
anything is wrong with the new downloader (and arguably it was
only excercised by a relatively small number of people compared
to how much the service is building).

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -8,7 +8,7 @@
 %bcond_with relax_requires
 
 # The minimum required osbuild version
-%global min_osbuild_version 163
+%global min_osbuild_version 166
 
 %global goipath         github.com/osbuild/osbuild-composer
 


thozza:
> This commit enables librepo by default in the cloudapi. This should help with on-prem out of sync mirror issues like: #4427

AFAIK, it won't, because CloudAPI is not used on-prem by default. Weldr API needs to use it to help with on-prem issues.

thozza (approved):
LGTM, thanks

Schutzfile
```diff
@@ -2,7 +2,7 @@
   "fedora-41": {
     "dependencies": {
       "osbuild": {
-        "commit": "a52e256c481c00e61f785f3ad68e29e1ac8a09fb"
+        "commit": "b1e74a309d5c2c088959c6e0322739fe20969912"
```
achilleas-k, line 5: We should also bump the mimimum osbuild version in the spec file, right?

achilleas-k (dismissed):

Schutzfile
mvo5: Thanks, excellent catch! Updated.

achilleas-k (dismissed):

PR #4913, opened 2025-11-03
cloudapi/v2: add netinst image type (HMS-9566)

tkoscieln:
Add `network-installer` image type to Cloud API.


internal/cloudapi/v2/handler.go
```diff
@@ -224,6 +224,8 @@ func imageTypeFromApiImageType(it ImageTypes, arch distro.Arch) string {
 		return "live-installer"
 	case ImageTypesMinimalRaw:
 		return "minimal-raw"
+	case ImageTypesNetinst:
+		return "netinst"
```
thozza: Nitpick: we tend to use `-installer` suffix for all installer image types, should we use `net-installer` instead?

croissanne: honestly we should probably add this to ibcli as well

tkoscieln: Possibly, not sure? I used the definition in [images](https://github.com/osbuild/images/blob/main/data/distrodefs/rhel-10/imagetypes.yaml#L2142) as a reference and since images are our source of truth, I assumed the naming should mirror what is in that repository? But yeah, I wondered why the net installer does not have that suffix just like the other installer types.

mvo5: Yeah, this is a good point, I opened https://github.com/osbuild/images/pull/1986 with a quick suggestion but open for other ideas about the naming of course.

tkoscieln: Based on [osbuild/images#1986 ](https://github.com/osbuild/images/pull/1986) I updated the image type to `network-installer`

PR #4921, opened 2025-11-17
templates/packer: add required tags to images (HMS-9609)

croissanne:
ssia


lzap (approved):

croissanne:
> Cannot we just hard-code the cost center? It's just a random number tbh. 😇

I wasn't sure to what extent this can be public, and figured why risk it. It's not set as a secret either btw, just an export statement before the test script in app-intf.

PR #4932, opened 2025-11-20
renovate: config for batch updates every Sunday

achilleas-k:
I currently don't have access to the konflux application to disable renovate / mintmaker on this repo, like I did with bootc-image-builder.  Until we do, let's at least make the PRs less noisy by batching them into once a week.

Add a renovate config to do all Go dependency updates in one batch every Sunday.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "packageRules": [
+    {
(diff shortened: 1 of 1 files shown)


croissanne (changes requested):
oh I'd just disable it entirely, see https://github.com/osbuild/image-builder-crc/blob/main/renovate.json

we handle it with gobump right?

achilleas-k:
> how about #4935?

Yes

PR #4935, opened 2025-11-20
renovate.json: ignore gomod updates and set other limits

croissanne:
The konflux bots are very noisy if not restrained.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "github>konflux-ci/mintmaker//config/renovate/renovate.json",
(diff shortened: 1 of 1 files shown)


achilleas-k:
Should we move it under `.github/`?  I don't like polluting the root too much (not that it's particularly clean right now, but let's do what we can).

https://docs.renovatebot.com/configuration-options/

croissanne:
> Should we move it under `.github/`? I don't like polluting the root too much (not that it's particularly clean right now, but let's do what we can).
> 
> https://docs.renovatebot.com/configuration-options/

done!

PR #4943, opened 2025-11-24
test/api/azure.sh cloud cleaning fixes

ondrejbudai:
**test/api/azure.sh: fix public IP address clean up**

The P in IP should be uppercase. :/

---

**test/api/azure.sh: delete vnets too**

The comment mentioned that vnets are re-used between tests but that's
clearly not true as we use unique names for them. Let's clean them up
as well so we don't have to rely on cloud-cleaner as a fallback.

---


croissanne (dismissed):
wow

supakeen (dismissed):
aCK!

ondrejbudai:
ugh, I had to convince shellcheck that using sed is fine.

lzap (approved):

lzap:
There are some relevant (?) failures.

PR #4958, opened 2025-12-12
worker: support separate set of s3 credentials (HMS-160)

croissanne:
worker: support separate credentials for S3 uploads
    
    Currently we use iam instance profiles for everything, these are useful
    because no separate set of credentials need to be maintained. However,
    they have a limitation when it comes to generating presigned urls for
    objects in s3, instance profiles only allow for urls that are valid for
    6 hours. Using a separate set of credentials just for s3 allows for
    generating urls that are valid for 7 days.

### DIFF
--- a/templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_aws_creds.sh
+++ b/templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_aws_creds.sh
@@ -15,24 +15,45 @@ bucket = "${WORKER_CONFIG_AWS_BUCKET:-}"
 EOF
 
 AWS_ACCOUNT_IMAGE_BUILDER_ARN=${AWS_ACCOUNT_IMAGE_BUILDER_ARN:-}
-if [[ -z "$AWS_ACCOUNT_IMAGE_BUILDER_ARN" ]]; then
-  echo "AWS_ACCOUNT_IMAGE_BUILDER_ARN not defined, skipping."
-  exit 0
-fi
+if [[ -n "$AWS_ACCOUNT_IMAGE_BUILDER_ARN" ]]; then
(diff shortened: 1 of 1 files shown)


cmd/osbuild-worker/config_test.go
```diff
@@ -47,6 +47,7 @@ upload_threads = 8
 
 [aws]
 credentials = "/etc/osbuild-worker/aws-creds"
+s3_credentials = "/etc/osbuild-worker-aws-s3-creds"
```
thozza: Nitpick: I assume that it was supposed to be `/etc/osbuild-worker/aws-s3-creds`

templates/packer/ansible/roles/common/files/worker-initialization-scripts/get_aws_creds.sh
```diff
@@ -25,14 +25,30 @@ fi
...
+
+
+if [[ "$S3_ACCESS_KEY_ID" != null ]; then
+  sudo tee /etc/osbuild-worker/aws_s3_credentials.toml > /dev/null << EOF
+[default]
+aws_access_key_id = "$S3_ACCESS_KEY_ID"
+aws_secret_access_key = "$S3_SECRET_ACCESS_KEY"
 EOF
+  sudo tee -a /etc/osbuild-worker/osbuild-worker.toml > /dev/null << EOF
+s3_credentials = "/etc/osbuild-worker/aws_s3_credentials.toml"
```
thozza: Nitpick: It may be reasonable to assign the config path to a variable instead of having its literal value twice in the script. It would be less prone to typos.

thozza (commented):
Looks good, but there is one typo in a unit test that should make it fail AFAICT

thozza (approved):
LGTM

PR #4975, opened 2026-01-05
Move workers to rhel 10 (HMS-9987)

croissanne:
rpm on rhel 10 supports importing post quantum cryptographic keys,
    whilst rpm on rhel 9 will fail when importing a PQC key. This is a
    problem specifically when creating the buildroot, which uses the host's
    rpm binary. This is mitigated by moving the workers to rhel 10.


achilleas-k (dismissed):

lzap (dismissed):
Tried to grep `rhel-9` in the repo to find leftovers... was not a great idea :-)

Let's go and fix if we missed something.

kingsleyzissou (approved):

PR #4977, opened 2026-01-09
worker: add job to generate bootc manifests [HMS-10006]

croissanne:
This adds a terrible hack, but this could be improved further by having the bootc manifest job generate the targets as well. The real solution is to change images so the bootc distro no longer needs to start the conatiner in order to expose image type data.


achilleas-k:
Well.  It worked!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1173,6 +1173,9 @@ components:
         distribution:
           type: string
           example: 'rhel-8'
+        bootc_ref:
+          type: string
+          example: "quay.io/whatever/..."
         image_request:
```
lzap, line 1175: Maybe add full and valid ref? I could confirm if it accepts reference including tag `quay.io/org/container:latest`

test/cases/api-bootc.sh
```diff
@@ -0,0 +1,201 @@
...
+}
+
+WORKDIR=$(mktemp -d)
+REQ="${WORKDIR}/compose_request.json"
+ARCH=$(uname -m)
+
+cat > "$REQ" << EOF
+{
+  "bootc": {
+    "reference": "quay.io/centos-bootc/centos-bootc:stream9"
```
achilleas-k, line 174: I'd like to start mirroring and/or pinning this with an update job like we do for repos and other containers.  I'm about to set it up for testing bootc-based builds in osbuild/images.  Just making a note to return here and change this when we do.

achilleas-k (approved):
For all our talk of dirty hacks, this is quite clean and compartmentalised.  A joy to review.  Ty!

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1163,16 +1163,18 @@ components:
...
               type: string
-
     ComposeRequest:
       additionalProperties: false
-      required:
-        - distribution
       not: {required: ['customizations', 'blueprint']}
+      # oneOf:
+      #   - required: [distribution]
+      #   - required: [bootc_ref]
```
thozza: Why is this commented out? A comment would be helpful.

internal/cloudapi/v2/server.go
```diff
@@ -478,6 +478,61 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
...
+	}
+
+	// TODO: Hardcoding this is obviously a very bad hack, but currently this image type
+	// information isn't retrievable from images without running the bootc container.
+	if imageTypeFromApiImageType(ir.ImageType) == "qcow2" {
+		tgts[0].ImageName = "disk.qcow2"
+		tgts[0].OsbuildArtifact.ExportFilename = "disk.qcow2"
+		tgts[0].OsbuildArtifact.ExportName = "qcow2"
+	}
+	id, err := s.workers.EnqueueOSBuildAsDependency("x86_64", &worker.OSBuildJob{
```
thozza: Shouldn't this use `ir.Architecture` instead of hardcoded `x86_64` value? :thinking:

internal/cloudapi/v2/server.go
```diff
@@ -478,6 +478,61 @@ func (s *Server) enqueueKojiCompose(taskID uint64, server, name, version, releas
...
+		target.NewWorkerServerTarget(),
+	}
+
+	// TODO: Hardcoding this is obviously a very bad hack, but currently this image type
+	// information isn't retrievable from images without running the bootc container.
+	if imageTypeFromApiImageType(ir.ImageType) == "qcow2" {
+		tgts[0].ImageName = "disk.qcow2"
+		tgts[0].OsbuildArtifact.ExportFilename = "disk.qcow2"
+		tgts[0].OsbuildArtifact.ExportName = "qcow2"
+	}
```
thozza, line 527: What if the requested image type is not `qcow2`? Should it produce an error?

internal/cloudapi/v2/v2_test.go
```diff
@@ -2582,3 +2582,45 @@ func TestComposeIBManifest(t *testing.T) {
...
+	// get the osbuild job (as root job) and check that its dependency is an ImageBuilderManifestJob
+	rootJobs, err := queue.AllRootJobIDs(context.Background())
+	require.NoError(t, err)
+	require.Len(t, rootJobs, 1)
+
+	osbuildJobID := rootJobs[0]
+	osbuildJobType, argsJSON, deps, _, err := queue.Job(osbuildJobID)
+	osbuildJobTypeSplit := strings.Split(osbuildJobType, ":")
+	require.NoError(t, err)
+	require.Equal(t, worker.JobTypeOSBuild, osbuildJobTypeSplit[0])
```
thozza, line 2616: You should also verify that `osbuildJobTypeSplit[1] == test_distro.TestArch3Name`

test/cases/api-bootc.sh
```diff
@@ -0,0 +1,201 @@
...
+  },
+  "image_request": {
+    "architecture": "$ARCH",
+    "image_type": "guest-image",
+    "repositories": $(jq ".\"$ARCH\"" /usr/share/tests/osbuild-composer/repositories/centos-stream-9.json),
+    "upload_options": {
+      "region": "${AWS_REGION}"
+    }
```
thozza: Is this needed at all, since the upload to AWS is not supported?

internal/cloudapi/v2/v2_test.go
```diff
@@ -2582,3 +2582,45 @@ func TestComposeIBManifest(t *testing.T) {
...
+			"image_type": "guest-image",
+			"upload_options": {}
+		}
+	}`, test_distro.TestArch3Name), http.StatusCreated, `
+	{
+		"href": "/api/image-builder-composer/v2/compose",
+		"kind": "ComposeId"
+	}`, "id")
+
+	// get the osbuild job (as root job) and check that its dependency is an ImageBuilderManifestJob
```
thozza: Typo: `ImageBuilderManifestJob` should really be `JobTypeBootcManifest`

cmd/osbuild-worker/jobimpl-bootc-manifest.go
```diff
@@ -0,0 +1,143 @@
...
+				nil,
+			)
+		}
+		err := job.Finish(result)
+		if err != nil {
+			logWithId.Errorf("Error reporting job result: %v", err)
+		}
+	}()
+
+	logWithId.Infof("Generating bootc manifest (image ref: %s)", args.Ref)
```
thozza, line 68: I think that a reasonably extensive comment explaining the bootc manifest generation flow would be helpful here.

internal/cloudapi/v2/openapi.v2.yml
croissanne: ergh it doesn't work, let me just remove it. Can't figure out a way to have it in the spec with the advanced condition between customiztaions and blueprint already being there.

internal/cloudapi/v2/server.go
croissanne: yes :dog:

internal/cloudapi/v2/server.go
croissanne: yes let's do that

thozza (approved):
Thanks for the fixes!

PR #4991, opened 2026-01-26
cloudapi/v2: add distro metadata endpoint (HMS-9643)

tkoscieln:
Adds an endpoint for getting distro metadata - /distributions/{distro}. Returns all metadata distro has defined in osbuild/images. Includes arch metadata and corresponding image types metadata. Supports filtering by image type and by architecture and allows multiple values for both of these options.

Example output from  `/distributions/rhel-9.6?architecture=x86_64&image_type=ami`: 
```
{
  "architectures": {
    "aarch64": {
      "image_types": {
        "ami": {
          "aliases": [
            "aws"
          ],
          "base_partition_table": {
            "partitions": [
              {
                "fs_type": "vfat",
                "label": "ESP",
                "minsize": "209715200",
                "mountpoint": "/boot/efi"
              },
              {
                "fs_type": "xfs",
                "label": "boot",
                "minsize": "1073741824",
                "mountpoint": "/boot"
              },
              {
                "fs_type": "xfs",
                "label": "root",
                "minsize": "2147483648",
                "mountpoint": "/"
              }
            ],
            "type": "gpt"
          },
          "boot_mode": "uefi",
          "default_size": 10737418240,
          "exports": [
            "image"
          ],
          "filename": "image.raw",
          "iso_label": null,
          "mime_type": "application/octet-stream",
          "name": "ami",
          "ostree_ref": "",
          "partition_type": "gpt",
          "payload_package_sets": [
            "blueprint"
          ],
          "required_blueprint_options": null,
          "supported_blueprint_options": [
            "distro",
            "packages",
            "modules",
            "groups",
            "enabled_modules",
            "containers",
            "customizations.cacerts",
            "customizations.directories",
            "customizations.disk",
            "customizations.dnf",
            "customizations.files",
            "customizations.filesystem",
            "customizations.partitioning_mode",
            "customizations.fips",
            "customizations.firewall",
            "customizations.user",
            "customizations.sshkey",
            "customizations.group",
            "customizations.hostname",
            "customizations.kernel",
            "customizations.locale",
            "customizations.openscap",
            "customizations.repositories",
            "customizations.rpm",
            "customizations.services",
            "customizations.timezone",
            "customizations.rhsm",
            "name",
            "version",
            "description"
          ]
        }
      },
      "name": "aarch64"
    }
  },
  "codename": "",
  "module_platform_id": "platform:el9",
  "name": "rhel-9.6",
  "os_version": "9.6",
  "ostree_ref": "rhel/9/%s/edge",
  "product": "Red Hat Enterprise Linux",
  "releasever": "9"
}
```

### DIFF
--- a/internal/cloudapi/v2/errors.go
+++ b/internal/cloudapi/v2/errors.go
@@ -83,6 +83,7 @@ const (
 	ErrorArtifactNotFound                         ServiceErrorCode = 1022
 	ErrorDeletingJob                              ServiceErrorCode = 1023
 	ErrorDeletingArtifacts                        ServiceErrorCode = 1024
+	ErrorGettingImageTypes                        ServiceErrorCode = 1025
 
 	// Errors contained within this file
 	ErrorUnspecified          ServiceErrorCode = 10000
@@ -173,6 +174,7 @@ func getServiceErrors() serviceErrors {
 		serviceError{ErrorGettingComposeList, http.StatusInternalServerError, "Unable to get list of composes"},
 		serviceError{ErrorDeletingJob, http.StatusBadRequest, "Unable to delete job"},
 		serviceError{ErrorDeletingArtifacts, http.StatusInternalServerError, "Unable to delete job artifacts"},
+		serviceError{ErrorGettingImageTypes, http.StatusInternalServerError, "Unable to get list of image types"},
 
 		serviceError{ErrorUnspecified, http.StatusInternalServerError, "Unspecified internal error "},
 		serviceError{ErrorNotHTTPError, http.StatusInternalServerError, "Error is not an instance of HTTPError"},
--- a/internal/cloudapi/v2/handler.go
+++ b/internal/cloudapi/v2/handler.go
@@ -4,6 +4,7 @@ package v2
 import (
 	"cmp"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -18,7 +19,9 @@ import (
 	"github.com/labstack/echo/v4"
 
 	"github.com/osbuild/blueprint/pkg/blueprint"
+	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/distro/defs"
 	"github.com/osbuild/images/pkg/manifest"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/rpmmd"
@@ -1573,6 +1576,217 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
 	}
 
 	return ctx.JSON(http.StatusOK, distros)
+}
+
+// GetDistribution returns details for a specific distribution
+// with optional filtering by image type and architecture
+func (h *apiHandlers) GetDistribution(ctx echo.Context, distroName string, params GetDistributionParams) error {
+	d := h.server.distros.GetDistro(distroName)
+	if d == nil {
+		return HTTPError(ErrorUnsupportedDistribution)
+	}
+
+	architectures := make(map[string]ArchitectureInfo)
+
+	for _, archName := range d.ListArches() {
+		if params.Architecture != nil && !slices.Contains(*params.Architecture, archName) {
+			continue
+		}
+
+		imageTypesInfo := make(map[string]ImageTypeInfo)
+
+		imageTypes, err := distro.GetImageTypes(d, archName)
+		if err != nil {
+			return HTTPErrorWithInternal(ErrorGettingImageTypes, err)
+		}
+
+		for _, imageType := range imageTypes {
+			if params.ImageType != nil && !slices.Contains(*params.ImageType, imageType.Name()) {
+				continue
+			}
+
+			var isoLabel *string
+			if label, err := imageType.ISOLabel(); err == nil {
+				isoLabel = &label
+			}
+
+			var basePartTable *Disk
+			partitionTable, err := imageType.BasePartitionTable()
+			if err != nil {
+				// Ignore errors about missing partition tables - still a valid image type
+				if errors.Is(err, defs.ErrNoPartitionTableForImgType) ||
+					errors.Is(err, defs.ErrNoPartitionTableForArch) {
+					continue
+				}
+				return HTTPErrorWithInternal(ErrorGettingImageTypes, err)
+			}
+			if partitionTable != nil {
+				basePartTable, err = partitionTableToDisk(partitionTable)
+				if err != nil {
+					return HTTPErrorWithInternal(ErrorJSONMarshallingError, err)
+				}
+			}
+
+			var partitionType ImageTypeInfoPartitionType
+			ptType := imageType.PartitionType()
+			if ptType != disk.PT_NONE {
+				partitionType = ImageTypeInfoPartitionType(ptType.String())
+			}
+
+			imageTypesInfo[imageType.Name()] = ImageTypeInfo{
+				Name:                      imageType.Name(),
+				Aliases:                   common.ToPtr(imageType.Aliases()),
+				Filename:                  common.ToPtr(imageType.Filename()),
+				MimeType:                  common.ToPtr(imageType.MIMEType()),
+				OstreeRef:                 common.ToPtr(imageType.OSTreeRef()),
+				IsoLabel:                  isoLabel,
+				DefaultSize:               common.ToPtr(imageType.Size(0)),
+				PartitionType:             common.ToPtr(partitionType),
+				BasePartitionTable:        basePartTable,
+				BootMode:                  common.ToPtr(ImageTypeInfoBootMode(imageType.BootMode().String())),
+				PayloadPackageSets:        common.ToPtr(imageType.PayloadPackageSets()),
+				Exports:                   common.ToPtr(imageType.Exports()),
+				RequiredBlueprintOptions:  common.ToPtr(imageType.RequiredBlueprintOptions()),
+				SupportedBlueprintOptions: common.ToPtr(imageType.SupportedBlueprintOptions()),
+			}
+		}
+
+		if len(imageTypesInfo) > 0 {
+			architectures[archName] = ArchitectureInfo{
+				Name:       archName,
+				ImageTypes: &imageTypesInfo,
+			}
+		}
+	}
+
+	return ctx.JSON(http.StatusOK, DistributionDetails{
+		Id:               d.Name(),
+		Kind:             "DistributionDetails",
+		Href:             fmt.Sprintf("/api/image-builder-composer/v2/distributions/%s", d.Name()),
+		Name:             d.Name(),
+		Codename:         common.ToPtr(d.Codename()),
+		Releasever:       common.ToPtr(d.Releasever()),
+		OsVersion:        common.ToPtr(d.OsVersion()),
+		ModulePlatformId: common.ToPtr(d.ModulePlatformID()),
+		Product:          common.ToPtr(d.Product()),
+		OstreeRef:        common.ToPtr(d.OSTreeRef()),
+		Architectures:    &architectures,
+	})
+}
+
+// partitionTableToDisk maps a disk.PartitionTable to a Disk
+func partitionTableToDisk(partitionTable *disk.PartitionTable) (*Disk, error) {
+	var partitions []Partition
+	for _, partition := range partitionTable.Partitions {
+		switch pt := partition.Payload.(type) {
+		case *disk.Filesystem:
+			if pt != nil {
+				fs := FilesystemTyped{
+					FsType: FilesystemTypedFsType(pt.Type),
+				}
+
+				if partition.Size > 0 {
+					fs.Minsize = common.ToPtr(Minsize(strconv.FormatUint(partition.Size.Uint64(), 10)))
+				}
+
+				if pt.Mountpoint != "" {
+					fs.Mountpoint = &pt.Mountpoint
+				}
+
+				if pt.Label != "" {
+					fs.Label = &pt.Label
+				}
+
+				var result Partition
+				if err := result.FromFilesystemTyped(fs); err != nil {
+					return nil, err
+				}
+
+				partitions = append(partitions, result)
+			}
+		case *disk.LVMVolumeGroup:
+			if pt != nil {
+				vg := VolumeGroup{
+					Type: VolumeGroupType(Lvm),
+					Name: &pt.Name,
+				}
+
+				if partition.Size > 0 {
+					vg.Minsize = common.ToPtr(Minsize(strconv.FormatUint(partition.Size.Uint64(), 10)))
+				}
+
+				var lvs []LogicalVolume
+				for _, lv := range pt.LogicalVolumes {
+					apiLv := LogicalVolume{
+						Name: &lv.Name,
+					}
+
+					if lv.Size > 0 {
+						size := Minsize(fmt.Sprintf("%d", lv.Size))
+						apiLv.Minsize = &size
+					}
+
+					if fs, ok := lv.Payload.(*disk.Filesystem); ok && fs != nil {
+						apiLv.FsType = LogicalVolumeFsType(fs.Type)
+						if fs.Mountpoint != "" {
+							apiLv.Mountpoint = &fs.Mountpoint
+						}
+						if fs.Label != "" {
+							apiLv.Label = &fs.Label
+						}
+					}
+
+					lvs = append(lvs, apiLv)
+				}
+
+				vg.LogicalVolumes = lvs
+				var result Partition
+				if err := result.FromVolumeGroup(vg); err != nil {
+					return nil, err
+				}
+
+				partitions = append(partitions, result)
+			}
+		case *disk.Btrfs:
+			if pt != nil {
+				btrfs := BtrfsVolume{
+					Type: Btrfs,
+				}
+
+				if partition.Size > 0 {
+					btrfs.Minsize = common.ToPtr(Minsize(strconv.FormatUint(partition.Size.Uint64(), 10)))
+				}
+
+				var subvols []BtrfsSubvolume
+				for _, sv := range pt.Subvolumes {
+					subvol := BtrfsSubvolume{
+						Mountpoint: sv.Mountpoint,
+						Name:       sv.Name,
+					}
+					subvols = append(subvols, subvol)
+				}
+
+				btrfs.Subvolumes = subvols
+				var result Partition
+				if err := result.FromBtrfsVolume(btrfs); err != nil {
+					return nil, err
+				}
+
+				partitions = append(partitions, result)
+			}
+		}
+	}
+	var ptType *DiskType
+
+	if partitionTable.Type != disk.PT_NONE {
+		t := DiskType(partitionTable.Type.String())
+		ptType = &t
+	}
+
+	return &Disk{
+		Type:       ptType,
+		Partitions: partitions,
+	}, nil
 }
 
 // GetComposeDownload downloads a compose artifact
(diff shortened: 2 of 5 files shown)


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -657,6 +657,58 @@ paths:
...
+      parameters:
+        - in: path
+          name: distro
+          schema:
+            type: string
+            example: 'rhel-9.6'
+          required: true
+          description: Name of the distribution
+        - in: query
+          name: image_type
```
tkoscieln, line 675: Not sure what to do here - do we need validation for both the `image_type` and `architecture` options? It is possible that user will search for image type or arch that do not exist on the distro or they will simply misspell. Current behaviour just omits the incorrect value from the response, essentially ignoring it.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,180 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+		}
+
+		imageTypesInfo := make(map[string]ImageTypeInfo)
+
+		imageTypes, err := distro.GetImageTypes(d, archName)
+		if err != nil {
+			return err
+		}
+		for _, imageType := range imageTypes {
+			// Filter by image type if specified
```
lzap: I guess this is obvious from the code this could have been dropped.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,180 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			}
+		}
+
+		if len(imageTypesInfo) > 0 {
+			architectures[archName] = ArchitectureInfo{
+				Name:       archName,
+				ImageTypes: &imageTypesInfo,
+			}
+		}
+	}
```
lzap, line 1660: That is a lot of code for such endpoint, I am not sure about this. Someone else need to take a look.

internal/cloudapi/v2/handler.go
tkoscieln: Dropped

internal/cloudapi/v2/handler.go
tkoscieln: This got unfortunately bloated by the need to map the partitions to the API format. There is already a helper function `convertDiskCustomizations` doing something similar, but that one maps from the API Disk format to Disk customization in Blueprint format. I tried to find another way to do the mapping but was unable to come up with anything else than mapping it manually like this.

lzap: Keep it, what I mean was I am not too confident to say anything about this, composer is the codebase I has the least experience with so far. So maybe this should go into `images`, let's see what others have to say about this @thozza @croissanne

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+								size := Minsize(fmt.Sprintf("%d", partition.Size))
+								vg.Minsize = &size
+							}
+							var lvs []LogicalVolume
+							for _, lv := range pt.LogicalVolumes {
+								apiLv := LogicalVolume{
+									Name: &lv.Name,
+								}
+								if lv.Size > 0 {
+									size := Minsize(fmt.Sprintf("%d", partition.Size))
```
kingsleyzissou: Should this be `lv.size`?

```
size := Minsize(fmt.Sprintf("%d", lv.Size))
```

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+	for _, archName := range d.ListArches() {
+		if params.Architecture != nil && !slices.Contains(*params.Architecture, archName) {
+			continue
+		}
+
+		imageTypesInfo := make(map[string]ImageTypeInfo)
+
+		imageTypes, err := distro.GetImageTypes(d, archName)
+		if err != nil {
+			return err
```
kingsleyzissou: Nitpick: in some places we return `HTTPError` and in this case we return the `err` directly. It might make sense to wrap this in `HTTPError` too.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+								fs.Minsize = &size
+							}
+							if pt.Mountpoint != "" {
+								fs.Mountpoint = &pt.Mountpoint
+							}
+							if pt.Label != "" {
+								fs.Label = &pt.Label
+							}
+							var result Partition
+							_ = result.FromFilesystemTyped(fs)
```
kingsleyzissou: What happens if there is an error? Maybe we should handle this? Or maybe log the error?

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+									}
+									if lv.Payload.(*disk.Filesystem).Label != "" {
+										apiLv.Label = &lv.Payload.(*disk.Filesystem).Label
+									}
+								}
+								lvs = append(lvs, apiLv)
+							}
+							vg.LogicalVolumes = lvs
+							var result Partition
+							_ = result.FromVolumeGroup(vg)
```
kingsleyzissou: We're discarding the error here too

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+									subvol := BtrfsSubvolume{
+										Mountpoint: sv.Mountpoint,
+										Name:       sv.Name,
+									}
+									subvols = append(subvols, subvol)
+								}
+							}
+							btrfs.Subvolumes = subvols
+							var result Partition
+							_ = result.FromBtrfsVolume(btrfs)
```
kingsleyzissou: Same as above

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,179 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+							var lvs []LogicalVolume
+							for _, lv := range pt.LogicalVolumes {
+								apiLv := LogicalVolume{
+									Name: &lv.Name,
+								}
+								if lv.Size > 0 {
+									size := Minsize(fmt.Sprintf("%d", partition.Size))
+									apiLv.Minsize = &size
+								}
+								if lv.Payload != nil && lv.Payload.(*disk.Filesystem) != nil {
```
kingsleyzissou: We're checking `lv.Payload`multiple times here, could simplify this to:
```
if fs, ok := lv.Payload.(*disk.Filesystem); ok && fs != nil {
    apiLv.FsType = LogicalVolumeFsType(fs.Type)
    if fs.Mountpoint != "" ...
}
```

internal/cloudapi/v2/handler.go
tkoscieln: Added HTTPErrors wrappers for all of these Marshalling errors

internal/cloudapi/v2/handler.go
tkoscieln: Added dedicated internal HTTP error wrapper for this one

internal/cloudapi/v2/openapi.v2.yml
lzap: I think just omitting from results is fine for this use case, this is just an informative endpoint after all.

If you really want to let the user know that some combination is incorrect, return such information in the result. You can structure result as a hash, or add additional information like list of error messages produced during processing. Or even just a single error that can have all the possible wrapped errors in Go style.

Another alternative is to explicitly always return an error if there is a combination that is not possible, that sounds too limiting tho.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			continue
+		}
+
+		imageTypesInfo := make(map[string]ImageTypeInfo)
+
+		imageTypes, err := distro.GetImageTypes(d, archName)
+		if err != nil {
+			return HTTPErrorWithInternal(ErrorGettingImageTypes, err)
+		}
+		for _, imageType := range imageTypes {
```
lzap, line 1603: Nitpick: missing newline.

Now I see the whole code is quite compact, I mean, feel free to ignore. I generally prefer more newlines in general.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			}
+			var isoLabel *string
+			if label, err := imageType.ISOLabel(); err == nil {
+				isoLabel = &label
+			}
+
+			var basePartTable *Disk
+			partitionTable, err := imageType.BasePartitionTable()
+			if err == nil && partitionTable != nil {
+				partitions := make([]Partition, 0)
```
lzap: Nitpick: `var parititons []Parititon` can be used with `append` it will make the allocation automatically. Tho, I am not sure if it was your intention to have empty slice in the result or `nil`. I think `nil` might be actually more appropriate.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+					switch pt := partition.Payload.(type) {
+					case *disk.Filesystem:
+						if pt != nil {
+							fs := FilesystemTyped{
+								FsType: FilesystemTypedFsType(pt.Type),
+							}
+							if partition.Size > 0 {
+								size := Minsize(fmt.Sprintf("%d", partition.Size))
+								fs.Minsize = &size
+							}
```
lzap: This pattern is common in this patch and could be optimized. It looks like `paritition.Size` is integer based custom type, it has a helper method that returns `unit64`. The target type is `string`-based.

I think creating a helper method in `datasizes` that returns the size in bytes formatted in string would be the most appropriate. Then the code could be just:

```go
if partition.Size > 0 {
 fs.Minsize = common.Ptr(Minsize(partition.Size.String()))
}
```

Make sure to properly document the newly-created String method and the fact it always returns in bytes. Theoretically, you can create a function that returns it in human-readable form but that is a fair stretch, bytes should be ok I guess?

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+								}
+								if lv.Size > 0 {
+									size := Minsize(fmt.Sprintf("%d", lv.Size))
+									apiLv.Minsize = &size
+								}
+								if fs, ok := lv.Payload.(*disk.Filesystem); ok && fs != nil {
+									apiLv.FsType = LogicalVolumeFsType(lv.Payload.(*disk.Filesystem).Type)
+									if lv.Payload.(*disk.Filesystem).Mountpoint != "" {
+										apiLv.Mountpoint = &lv.Payload.(*disk.Filesystem).Mountpoint
+									}
```
lzap: The type was already asserted this can be something like:

```
apiLv.FsType = LogicalVolumeFsType(fs.Type)
if fs.Mountpoint != "" {
	apiLv.Mountpoint = &fs.Mountpoint
}
```

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+						if pt != nil {
+							btrfs := BtrfsVolume{
+								Type: Btrfs,
+							}
+							if partition.Size > 0 {
+								size := Minsize(fmt.Sprintf("%d", partition.Size))
+								btrfs.Minsize = &size
+							}
+							var subvols []BtrfsSubvolume
+							if len(pt.Subvolumes) > 0 {
```
lzap: You do not need to check for lengths before iterating any slice, just iterate it it will not be called when it is empty.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,188 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			if params.ImageType != nil && !slices.Contains(*params.ImageType, imageType.Name()) {
+				continue
+			}
+			var isoLabel *string
+			if label, err := imageType.ISOLabel(); err == nil {
+				isoLabel = &label
+			}
+
+			var basePartTable *Disk
+			partitionTable, err := imageType.BasePartitionTable()
```
croissanne, line 1614: Maybe split this one out in a helper, would make it a lot more readable

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -834,6 +886,139 @@ components:
           items:
             $ref: '#/components/schemas/BlueprintRepository'
 
+    DistributionDetails:
+      type: object
```
croissanne: Could these do with `additionalProperties: false` ? Always nice to enforce that (even though we're not consistent with it).

internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -834,6 +886,139 @@ components:
           items:
             $ref: '#/components/schemas/BlueprintRepository'
 
+    DistributionDetails:
```
croissanne, line 889: I know we're not always consistent with it, but technically in the guidelines for the openshift api top-level objects are supposed to include the fields defined in ObjectReference.

So:
```
DistributionDetails:
      - $ref: '#/components/schemas/ObjectReference'
      - type object
        ...
```

Sadly they're not public, I'll send them to you.

internal/cloudapi/v2/errors.go
```diff
@@ -56,6 +56,7 @@ const (
 	ErrorDistroOrBootcNotBoth         ServiceErrorCode = 43
 	ErrorDistroAndBootcMissing        ServiceErrorCode = 44
 	ErrorDistroMissing                ServiceErrorCode = 45
+	ErrorDistroNotFound               ServiceErrorCode = 46
```
croissanne: Could just reuse `ErrorUnsupportedDistribution` here I  think. It's what we return in cases like these (instead of 404, which ig is a bit weird maybe). If we change these to 404, let's do it in a separate PR or something?

internal/cloudapi/v2/handler.go
tkoscieln: I am not sure, depends on what we expect as consumers from the API? Partitions don't have `omitempty` in API spec, so if there are no partitions the current implementation would return empty field, which seems more reasonable to me than returning `nil` because it explicitely states that there are no partitions unlike `nil`? :thinking:

internal/cloudapi/v2/openapi.v2.yml
tkoscieln: Added to DistributionDetails, ArchitectureInfo and ImageTypeInfo. Not sure if it's needed in the latter two though :thinking:

internal/cloudapi/v2/openapi.v2.yml
tkoscieln: Done

internal/cloudapi/v2/handler.go
tkoscieln: This requires change in images and will be done in follow up.

internal/cloudapi/v2/handler.go
tkoscieln: Added some newlines in effort to make the code more readable in general, especially in the split helper

internal/cloudapi/v2/handler.go
tkoscieln: Ok, changed it to `nil`, so now the `base_partition_table` will be omitted if there are no partitions from the response. Adjusted tests accordingly.

internal/cloudapi/v2/handler.go
croissanne: I think the error needs to be checked here too, no? Currently if the error here != nil it's just ignored.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,216 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			}
+
+			var isoLabel *string
+			if label, err := imageType.ISOLabel(); err == nil {
+				isoLabel = &label
+			}
+
+			var basePartTable *Disk
+			partitionTable, err := imageType.BasePartitionTable()
+			if err == nil && partitionTable != nil {
```
croissanne: Drop `err == nil` and just check the error first.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,216 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+					fs.Mountpoint = &pt.Mountpoint
+				}
+
+				if pt.Label != "" {
+					fs.Label = &pt.Label
+				}
+
+				var result Partition
+				err := result.FromFilesystemTyped(fs)
+
```
croissanne: nitpick, weird newline here

also, not strictly needed but you could write this as
```
if err := result.FromFilesystemTyped(fd); err != nil{
 return nil, err
}
partitions = append...
```

matter of preference tho

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1576,216 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+		switch pt := partition.Payload.(type) {
+		case *disk.Filesystem:
+			if pt != nil {
+				fs := FilesystemTyped{
+					FsType: FilesystemTypedFsType(pt.Type),
+				}
+
+				if partition.Size > 0 {
+					size := Minsize(fmt.Sprintf("%d", partition.Size))
+					fs.Minsize = &size
```
lzap: Since this is used at several places and implementing a helper would require PR in `images`, I suggest you shorten this to just (pseudo-code I haven't tested):

```
fs.Minsize = common.Ptr(Minsize(strconv.Itoa(partition.Size.Uint64())))
```

I think this might work if not ignore.

lzap (dismissed):

internal/cloudapi/v2/handler.go
tkoscieln: `fs.Minsize = common.ToPtr(Minsize(strconv.FormatUint(partition.Size.Uint64(), 10)))` this worked fine

internal/cloudapi/v2/handler.go
tkoscieln: Rewritten

internal/cloudapi/v2/handler.go
tkoscieln: Since BasePartitionTable can raise two specific errors for no partitions, which should be still valid image type, I've added check for those two specific errors to be ignored and rest gets raised as HTTP Error.

internal/cloudapi/v2/handler.go
```diff
@@ -1575,6 +1578,217 @@ func (h *apiHandlers) GetDistributionList(ctx echo.Context) error {
...
+			}
+
+			var basePartTable *Disk
+			partitionTable, err := imageType.BasePartitionTable()
+			if err != nil {
+				// Ignore errors about missing partition tables - still a valid image type
+				if errors.Is(err, defs.ErrNoPartitionTableForImgType) ||
+					errors.Is(err, defs.ErrNoPartitionTableForArch) {
+					continue
+				}
```
croissanne, line 1620: tiiiiiiiiiiiniest nitpick, this can be outside (and before) the err != nil check to reduce nesting with 1 level. Not a big deal tho.

PR #4992, opened 2026-01-27
cloudapi: support pxe images for bootc composes (HMS-10100)

croissanne:
ssia

### DIFF
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -522,6 +522,10 @@ func (s *Server) enqueueBootcCompose(request ComposeRequest, channel string) (uu
 		tgts[0].ImageName = "disk.qcow2"
 		tgts[0].OsbuildArtifact.ExportFilename = "disk.qcow2"
 		tgts[0].OsbuildArtifact.ExportName = "qcow2"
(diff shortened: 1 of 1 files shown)


test/cases/api-bootc.sh
```diff
@@ -197,3 +197,29 @@ curl \
     --cert /etc/osbuild-composer/client-crt.pem \
     "https://localhost/api/image-builder-composer/v2/composes/$COMPOSE_ID"
 test "$UPLOAD_STATUS" = "success"
+
+
+cat > "$REQ" << EOF
+{
+  "bootc": {
+    "reference": "quay.io/centos-bootc/centos-bootc:stream9"
```
supakeen: container must contain dracut-live

croissanne: Is there any image that's currently available that has it pre-installed? Otherwise I might just drop the test.

supakeen: No; you would need to create a (derived) container yourself. Note that this only matters if you *actually* want to build it.

croissanne: iiiii'll just ignore it then :3

PR #4994, opened 2026-01-29
[RHEL-9.2] update logrus

diaasami:


diaasami:
Changes were done downstream instead to reduce effort maintaining CI

PR #4995, opened 2026-01-29
[RHEL-8.6] update logrus

diaasami:


diaasami:
Changes were done downstream instead

PR #4996, opened 2026-01-30
cloudapi: Add optional blueprint-id UUID field (HMS-3881)

ochosi:
In order to be able to bake the blueprint UUID into the image as an RHSM fact, we need to extend composer's API.

This PR builds on https://github.com/osbuild/images/pull/2163

### DIFF
--- a/internal/cloudapi/v2/compose_test.go
+++ b/internal/cloudapi/v2/compose_test.go
@@ -5,6 +5,8 @@ import (
 	"io/fs"
 	"testing"
 
+	"github.com/google/uuid"
+	openapi_types "github.com/oapi-codegen/runtime/types"
 	"github.com/osbuild/blueprint/pkg/blueprint"
 	repos "github.com/osbuild/images/data/repositories"
 	"github.com/osbuild/images/pkg/customizations/subscription"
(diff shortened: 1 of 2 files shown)


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1374,6 +1374,12 @@ components:
           $ref: '#/components/schemas/Koji'
         blueprint:
           $ref: '#/components/schemas/Blueprint'
+        blueprint_id:
```
supakeen, line 1377: Minor question but why isn't this part of `#/components/schemas/Blueprint`? There's probably a reason for that but what is it?

ochosi: @supakeen  This is my conceptual rationale (and others may disagree): The blueprint schema describes the 'what' of a blueprint, whereas the blueprint_id is an external identifier that image-builder-crc creates when the blueprint is stored.
Differently put: I didn't want to inject a service-specific provenance data point into the blueprint schema.

ochosi: @achilleas-k @croissanne  Does my thinking here my sense to you?

supakeen: It at least makes sense to me 🙂

PR #5004, opened 2026-02-09
cloudapi/v2: allow upload types for pxe & network installer (HMS-10178)

kingsleyzissou:
We need to enable the `aws.s3` upload type for the hosted frontend and the `local` upload type for the on-prem frontend.

JIRA: [HMS-10178](https://issues.redhat.com/browse/HMS-10178)


lzap (approved):
One observation, `pxe-tar-xz` is mentioned as `pxe:xxx` that is way too generic name I think. One can PXE boot pretty much anything, not sure when/where/why we have chosen such a generic name.

PR #5006, opened 2026-02-10
Add Transactions and Modules fields to depsolve job result (HMS-9090)

thozza:
## Overview

This PR prepares osbuild-composer for migration to a newer osbuild/images version that uses `Transactions` instead of `Packages` for multi-stage RPM installation. The changes add new fields (`Transactions`, `Modules`, `Solver`) to `DepsolveJobResult` and introduce DTOs to decouple from library types.

**Bug fix**: The `Modules` field from `depsolvednf.DepsolveResult` was previously discarded by the worker after depsolving, causing modularity information to be lost and modules not being configured properly on built images. This PR captures and preserves the modularity data from depsolving.

**Migration path**: This PR must be merged and released first, then all infrastructure (composer and workers) must be upgraded to this version. This ensures all workers populate the `Transactions` field. Only after this intermediate step can osbuild-composer be updated to the new osbuild/images version that removes `Packages`, requires `Transactions` and moves to multi-stage RPM installation.

## Architectural Changes

- **DTO pattern**: Introduced `DepsolvedRepoConfig` and `DepsolvedModuleSpec` DTOs instead of using `rpmmd.*` types directly. This decouples the worker REST API from osbuild/images library internals, preventing library changes from breaking the API contract.
- **Centralized conversion**: Added `ToDepsolvednfResult()` helper method that handles all field mapping in one place, making future field additions easier to maintain.
- **Backward compatibility by design**: Both `PackageSpecs` and `Transactions` are populated during the transition period. The server iterates over `PackageSpecs` but passes `Transactions` through when available.

## Key Changes

- Add `DepsolvedRepoConfig` and `DepsolvedModuleSpec` DTO types with conversion helpers
- Add `Transactions`, `Modules`, and `Solver` fields to `DepsolveJobResult`
- Add `ToDepsolvednfResult()` helper and use it in cloudapi server
- Refactor `depsolve()` to return `*DepsolveJobResult` directly (cleaner 1-value return vs 7-value)
- Preserve nil semantics in list conversion functions (nil vs empty slice)
- Extend cloudapi test mocks with all new depsolve result fields

## Breaking Changes

This PR is fully backward compatible. Workers populate both `PackageSpecs` (for old servers) and `Transactions` (for new servers). The server continues to iterate over `PackageSpecs` but includes `Transactions` in the result passed to manifest serialization.

## Testing

- Added JSON roundtrip tests for all new DTO types (`DepsolvedRepoConfig`, `DepsolvedModuleSpec`)
[...]

PR #5008, opened 2026-02-10
test/api/common: verify that the nodejs:20 module is enabled [HMS-10242]

achilleas-k:
We already check that the module is *installed*, but we should also verify that the module is enabled.

NOTE: This test is expected to fail now until #5006 is merged and osbuild is updated to include https://github.com/osbuild/osbuild/commit/3cfa599bce8cd767f52ea6354eaa5f634fa37f93

### DIFF
--- a/test/cases/api/aws.sh
+++ b/test/cases/api/aws.sh
@@ -210,9 +210,9 @@ function verify() {
   echo "ARCH is $ARCH"
 
   if [ "$ARCH" = "aarch64" ]; then
-    INST_TYPE="t4g.micro"
+    INST_TYPE="t4g.small"
   elif [ "$ARCH" = "x86_64" ]; then
-    INST_TYPE="t2.micro"
+    INST_TYPE="t2.small"
   else
     echo "Unsupported architecture ❌"
     exit 1


achilleas-k:
Huh, ok: https://github.com/osbuild/osbuild-composer/blob/6aa7c0d934e2e26090ee9fc1981cb13338a06f09/test/cases/api.sh#L388-L395

I suppose we could disable it and the test would still be valid.

lzap (approved):

achilleas-k:
Rebasing this after #5012 is merged should make everything pass.

lzap:
> Which I swear I haven't forgotten about :)

Me neither, I want to finish logging refactoring first and than this will be me next side project.

lzap:
> we would have a converter from the CloudAPI BP to the UBP

I was hoping once we implement UBP in images, then cloud API just adopts UBP in v3, or whatever is the next version.

lzap (approved):
I was not aware of `<<<` bashism, nice.

achilleas-k:
Rebased on both required changes.  Let's see how this goes!

achilleas-k:
> I was not aware of `<<<` bashism, nice.

I tend to prefer it over `echo | grep` even though that's more common because it more closely matches how one would grep (or `jq`, etc) a file, in terms of the order of options and arguments.

lzap (approved):

achilleas-k:
Added an extra fail check for the `dnf module list` call.

bcl (approved):
Good catch!

achilleas-k:
I'll see if I can reproduce the Azure failure locally and poke around.

lzap (dismissed):

test/cases/api/common/common.sh
```diff
@@ -283,3 +279,35 @@ function verify_cacert_customization {
     exit 1
   fi
 }
+
+
+function verify_modules_customization {
+  local _ssh="$1"
+  if jq -e .customizations.enabled_modules <<< "${REQUEST_FILE}"; then
```
ondrejbudai: Isn't `$REQUEST_FILE` just a file? I think you want `<` here.

achilleas-k: yup, thanks!  I shouldn't rush -_-

PR #5013, opened 2026-02-11
deps: set Go version to 1.24.12 explicitly

lzap:
It is better to be explicit so GHA go action does not need to
re-download Go toolbox version again. Also, let's upgrade linter as
well.

---

Also ugprades pgx to v5.

Rebased on top of https://github.com/osbuild/osbuild-composer/pull/5026

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -2,6 +2,7 @@
 
 set -eu
 
+# 1.24.12
 GO_MINOR_VERSION="1.24"
 GO_VERSION="${GO_MINOR_VERSION}.12"
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,7 +70,7 @@ jobs:
     - uses: actions/checkout@v5
     - uses: actions/setup-go@v6
       with:
-        go-version: "1.24"
+        go-version: "1.24.12"
     - env:
         PGUSER: postgres
         PGPASSWORD: foobar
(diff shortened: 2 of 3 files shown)


tools/prepare-source.sh
```diff
@@ -2,8 +2,8 @@
 
 set -eu
 
-GO_MINOR_VERSION="1.23"
-GO_VERSION="${GO_MINOR_VERSION}.9"
+GO_MINOR_VERSION="1.24"
+GO_VERSION="1.24.12"
```
lzap: This change will make easier to search for Go version when doing search and replace during update.

thozza: I disagree, since you will still need to manually update the patch part of the version. So you can't just simply search and replace anyway.

lzap: From the git history, I can tell that we did more major updates than minor recently. But I already anticipated this so I am going to rebase a change that should please both sides :-D

```
# Go 1.24.12 (keep the comment up-to-date for easier search)
GO_MINOR_VERSION="1.24"
GO_VERSION="${GO_MINOR_VERSION}.12"
```

achilleas-k: Was this done?

lzap: Ah damn I did not push, well, since you already created https://github.com/osbuild/osbuild-composer/pull/5026 let's merge that one and I can rebase on top.

PR #5015, opened 2026-02-12
go.mod: update osbuild/images to v0.238.0 for multi-stage RPM installation support (HMS-9090)

thozza:
Update osbuild/images to v0.238.0, which introduces multi-stage RPM installation to fix building images with RPMs that use `%pretrans` scriptlets (HMS-9090). The new library version removes the `Packages` field from `depsolvednf.DepsolveResult`, replacing it with `Transactions` (`TransactionList`), and adds a Repo *RepoConfig pointer to rpmmd.Package that is validated during manifest serialization. This PR adapts all consumers across composer to the new API while maintaining full backward compatibility for mixed-version deployments.

## Architectural Changes

`ToDepsolvednfResult()` now uses `Transactions` as the primary source for converting worker results to the library type. When `Transactions` is absent (old worker), it falls back to wrapping `PackageSpecs` as a single-element `TransactionList`. This enables gradual infrastructure upgrades without coordination:

- **New server + old worker:** fallback wraps `PackageSpecs` into a single transaction — builds work as before
- **New server + new worker:** real multi-stage transactions are used — `%pretrans` fix is active
- **Old server + new worker:** old server reads `PackageSpecs` (still populated by new worker) — builds work as before

`PackageSpecs` in `DepsolveJobResult` is kept and populated from `Transactions.AllPackages()` for this backward compatibility, with a TODO to remove it after ~2 releases.

The DTO-to-rpmmd package conversion now populates the `Repo` pointer on every package, mirroring the approach used by the depsolvednf v2 parser: repos are converted first into a `repoID → *RepoConfig` map, which is then passed through `ToRPMMD()` / `ToRPMMDList()` / `DepsolvedTransactionsToRPMMD()`. All three helpers now return errors for empty or unknown `RepoID`, catching depsolve bugs early rather than during manifest serialization.

## Key Changes

- Vendor osbuild/images v0.238.0 with multi-stage RPM installation and `TransactionList` API
- Switch all `DepsolveResult.Packages` usages to `Transactions.AllPackages()` across worker, weldr API, and store-dump
- Rework `ToDepsolvednfResult()` to return an error and iterate `Transactions` as primary with `PackageSpecs` fallback
- Populate `Repo *RepoConfig` pointer on every converted package via a repo map built from `RepoConfig`s
- Make `ToRPMMD()`, `ToRPMMDList()`, and `DepsolvedTransactionsToRPMMD()` return errors on empty or unknown `RepoID`
- Worker `depsolve()` populates `PackageSpecs` from `Transactions.AllPackages()` for backward compat
- Update mock data ordering to match NEVRA-sorted output of `AllPackages()`

## Breaking Changes

This PR is fully backward compatible. `PackageSpecs` continues to be populated by the worker and consumed by older servers. `ToDepsolvednfResult()` handles results from older workers that lack `Transactions`.

## Testing

Unit tests updated across worker, weldr, and client packages due to depsolve result being ordered by Full NEVRA. `TestDepsolveJobResultToDepsolvednfResult` covers the normal path (multiple transactions, optional fields like SBOM/modules/solver, multiple pipelines) and the backward-compatible fallback path (PackageSpecs-only input).


croissanne (dismissed):
yup!

croissanne (approved):

PR #5017, opened 2026-02-12
test/rhel-upgrade: bump target versions and refactor repo override

thozza:
Update upgrade test target versions to RHEL 9.8 and RHEL 10.2.

Refactor repository override generation in upgrade_verify.sh to use VERSION_ID dynamically instead of maintaining separate hardcoded JSON blocks for each RHEL version. GPG keys are now defined as variables and selected via case statement based on the running version.

Previously, the GPG signing keys were swapped, meaning that el9 repos specified keys used by el10 and the other way around. This previously didn't cause any issues, because the repo configs had "check_gpg" disabled. The keys are now fixed.


thozza:
I was brave enough to enable GPG check on the upgrade test repos, which was probably too optimistic :sob:

lzap (approved):

achilleas-k (approved):

PR #5021, opened 2026-02-12
test: enable cloud-image-validator [HMS-9868]

achilleas-k:
This reverts commit 23594710227e4d60631b0aa4ed4957503df390d7.

---

We disabled CIV a while ago because it was failing on some cloud configuration issue.  Let's re-enable it and fix any issues it has now that we have a bit more capacity.

### DIFF
--- a/test/cases/aws.sh
+++ b/test/cases/aws.sh
@@ -205,80 +205,77 @@ if nvrGreaterOrEqual "osbuild-composer" "83"; then
     esac
 fi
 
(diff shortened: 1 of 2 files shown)


achilleas-k:
Failures seem to be only for 10.2.  I'll update the CIV to match.

achilleas-k:
Closing this as Wont Do.  CIV is going through a big redesign right now and we're rethinking our cloud image testing more globally.  We'll come around to this again as part of that work.

PR #5087, opened 2026-04-09
Make bootc container source type configurable (HMS-10262)

thozza:
The `enqueueBootcCompose` function hardcodes `UseRemoteContainerSource: false`, which works for on-prem deployments where containers live in local podman storage but will break the service deployment that needs `org.osbuild.skopeo` (remote pull). Add a `[bootc]` config section with a `use_remote_container_source` toggle - settable via TOML or the `BOOTC_USE_REMOTE_CONTAINER_SOURCE` environment variable - and wire it through `ServerConfig` into the compose pipeline so each deployment gets the right container source type.

## Architectural Changes

Follow the existing per-section config pattern (`[koji]`, `[worker]`, `[weldr_api]`) to add a new `[bootc]` TOML section. A per-request API option was considered but deferred - on-prem and service deployments never mix modes today, and a server-side default would still be required, so the simpler approach avoids unnecessary API surface. The env-var support also required teaching `loadConfigFromEnv` to handle `bool` fields, which were previously skipped.

## Key Changes

- Add `BootcConfig` struct with `use_remote_container_source` field, configurable via TOML and env var
- Extend `loadConfigFromEnv` to support `bool` fields with `env` tags (previously a no-op) and extract a `lookupEnvTagValue` helper to reduce duplication
- Add `BootcUseRemoteContainerSource` to `v2.ServerConfig` and replace the hardcoded `false` in `enqueueBootcCompose`
- Set `BOOTC_USE_REMOTE_CONTAINER_SOURCE=true` in the OpenShift deployment template for the service environment
- Extend unit tests (`TestComposeBootc`, `TestHandleBootcPreManifest_HappyPath`) to cover both local and remote container source variants
- Extend `api-bootc.sh` functional test to validate manifest sources and run both variants in CI via a GitLab matrix

## Breaking Changes

This PR is fully backward compatible. The default when the option is omitted is `false`, preserving existing on-prem behavior to use local container sources.

## Testing

Unit tests verify config loading (default value, TOML parsing, and env-var override), propagation of the flag through `ServerConfig` into `BootcPreManifestJob` args, and correct container source `Local` property in the pre-manifest result for both `true` and `false` variants. The `api-bootc.sh` functional test was extended to inspect the compose manifest for the expected source type (`org.osbuild.skopeo` vs `org.osbuild.containers-storage`), and the GitLab CI matrix now exercises both configurations.

### DIFF
--- a/templates/openshift/composer.yml
+++ b/templates/openshift/composer.yml
@@ -128,6 +128,8 @@ objects:
             value: ${DISTRO_ALIASES}
           - name: CHANNEL
             value: ${CHANNEL}
+          - name: BOOTC_USE_REMOTE_CONTAINER_SOURCE
+            value: "${BOOTC_USE_REMOTE_CONTAINER_SOURCE}"
           ports:
           - name: composer-api
             protocol: TCP
@@ -350,3 +352,10 @@ parameters:
       Channel where this pod is deployed.
       This is appended to the logs. Usually something like
       "local", "staging" or "production".
+  - name: BOOTC_USE_REMOTE_CONTAINER_SOURCE
+    value: "true"
+    description: >
+      Whether to use remote container source for bootc composes.
+      This is used to control the container source type in the compose manifest.
+      For the service deployment, this should be true, because one can't expect
+      the container to be present in local podman storage.
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -163,6 +163,7 @@ API-bootc:
     matrix:
       - RUNNER:
           - aws/centos-stream-9-x86_64
+        BOOTC_USE_REMOTE_CONTAINER_SOURCE: ["true", "false"]
 
 Packer:
   stage: test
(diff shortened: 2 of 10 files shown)


croissanne (approved):

lzap (approved):

templates/openshift/composer.yml
```diff
@@ -128,6 +128,8 @@ objects:
             value: ${DISTRO_ALIASES}
           - name: CHANNEL
             value: ${CHANNEL}
+          - name: BOOTC_USE_REMOTE_CONTAINER_SOURCE
```
croissanne, line 131: This won't work actually, needs to be added under the configmap.

croissanne (changes requested):
Noticed a small error in the openshift template changes.

templates/openshift/composer.yml
thozza: It will, because the config option is passed via ENV variable to composer, which will read it from there. There is no need to add it to the config file, although it is definitely possible to use this approach instead of the ENV.

croissanne: oh ic, you added an env tag, nvm ya. thanks!

croissanne (approved):

PR #5092, opened 2026-04-13
Allow any valid upload target for bootc composes, not just local (HMS-10507)

thozza:
Bootc composes hardcoded a local upload target at enqueue time, making them unusable in the hosted service where cloud upload targets (e.g., AWS S3) are required. This PR moves upload target construction into the `BootcPreManifest` server-side job — where the `distro.ImageType` is available after resolving bootc container info — and validates targets against `targetSupportMap()` at enqueue time, consistent with the non-bootc flow. This is a required prerequisite for enabling bootc composes in the service.

While this change technically enables building bootc-based qcow2 images with an AWS S3 upload target, that is not the goal here. The focus is on refactoring the upload target handling while keeping on-prem bootc composes working. An end-to-end test verifying the full service flow will follow as a next step.

## Architectural Changes

Target construction is deferred from enqueue time to the `BootcPreManifest` job because `distro.ImageType` is only available after resolving bootc container info, and the worker must not derive image-type-dependent data (version skew risk). The OSBuild job graph gains a dependency on `BootcPreManifest` and reads targets from dynargs via `PreManifestDynArgsIdx`, following the existing container-resolve pattern.

## Key Changes

- Add `BootcUploadTarget` struct and extend `BootcPreManifestJob`/`Result` to carry target specs and constructed targets across the job boundary
- Construct complete targets in `handleBootcPreManifest` using the same `getTarget()` code path as non-bootc composes
- Add `PreManifestDynArgsIdx` to `OSBuildJob`; worker reads targets from dynargs with mutual-exclusivity guard
- Replace hardcoded local target in `enqueueBootcCompose` with `targetSupportMap()` validation and target normalization
- Remove `*ComposeRequest` parameter from `getTarget()`/`GetTargets()` — derive container image name from `distro.ImageType` instead
- Update `api-bootc.sh` to use explicit `upload_targets` and add image download verification

## Breaking Changes

This PR is fully backward compatible. The `UploadOptions` field continues to work — it is normalized to a default target type at enqueue time, same as non-bootc.

## Testing

- Extended `TestComposeBootc` with table-driven cases: aws.s3 and local via `upload_targets`, default via `upload_options`, invalid target rejection, and missing target rejection
- Verified job graph wiring: OSBuild depends on ManifestByID + BootcPreManifest, static targets empty, dynargs index set
- Extended pre-manifest happy-path tests with aws.s3, local, and multi-target cases asserting target names, artifact metadata, and target-specific options
- Added pre-manifest error case for invalid upload target options JSON
- Updated `api-bootc.sh` to use explicit local target and verify image download

### DIFF
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -577,28 +577,60 @@ func (s *Server) enqueueBootcCompose(request ComposeRequest, channel string) (uu
 		return uuid.Nil, err
 	}
 
-	// Only local targets are supported
-	if ir.UploadTargets != nil {
-		for _, ut := range *ir.UploadTargets {
-			if ut.Type != UploadTypesLocal {
-				return uuid.Nil, HTTPError(ErrorInvalidUploadTarget)
-			}
-		}
-	}
-	tgts := []*target.Target{
-		target.NewWorkerServerTarget(),
-	}
-
 	imageTypeName := imageTypeFromApiImageType(ir.ImageType)
 
-	// TODO: Hardcoding this is obviously a very bad hack, but currently this image type
-	// information isn't retrievable from images without running the bootc container.
-	if imageTypeName == "qcow2" {
-		tgts[0].ImageName = "disk.qcow2"
-		tgts[0].OsbuildArtifact.ExportFilename = "disk.qcow2"
-		tgts[0].OsbuildArtifact.ExportName = "qcow2"
-	} else {
+	// TODO: Only qcow2 (guest-image) is supported for bootc composes for now.
+	if imageTypeName != "qcow2" {
 		return uuid.Nil, HTTPErrorWithDetails(ErrorUnsupportedImageType, nil, "only qcow2 (guest-image) is supported for bootc composes")
+	}
+
+	// Validate and normalize upload targets — consistent with non-bootc flow.
+	// Error if no targets provided.
+	if ir.UploadOptions == nil && (ir.UploadTargets == nil || len(*ir.UploadTargets) == 0) {
+		return uuid.Nil, HTTPError(ErrorJSONUnMarshallingError)
+	}
+
+	tsm := targetSupportMap()
+	var uploadTargetSpecs []worker.BootcUploadTarget
+
+	if ir.UploadTargets != nil {
+		for _, ut := range *ir.UploadTargets {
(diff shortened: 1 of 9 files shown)


internal/cloudapi/v2/server.go
```diff
@@ -577,28 +577,60 @@ func (s *Server) enqueueBootcCompose(request ComposeRequest, channel string) (uu
...
+	if ir.UploadOptions != nil {
+		defTargetType, err := getDefaultTarget(ir.ImageType)
+		if err != nil {
+			return uuid.Nil, err
+		}
+		if !tsm[defTargetType][ir.ImageType] {
+			return uuid.Nil, HTTPErrorWithDetails(
+				ErrorInvalidUploadTarget, nil,
+				fmt.Sprintf("invalid default upload target type %q for image type %q", defTargetType, ir.ImageType),
+			)
```
croissanne, line 624: Is this an "ErrorInvalidUploadTarget" / 400? Isn't this a programming error? Like if the default target for an image type is wrong then it should be a 500, no?

croissanne (changes requested):
One comment, otherwise lgtm

internal/cloudapi/v2/server.go
thozza: You are correct, but I've chosen this error code for the sake of consistency with the non-bootc flow, which uses this one in the same situation.

The most relevant part is in the `ImageRequest.GetTargets()`, where `getTarget()` is called on a `defTargetType`. `getTargets()` has a `switch` statement, which returns `ErrorInvalidUploadTarget` for unknown target name.

https://github.com/osbuild/osbuild-composer/blob/a58545b6c422588ad2e72ccb2783cbd4f9ea5031/internal/cloudapi/v2/imagerequest.go#L396-L401

https://github.com/osbuild/osbuild-composer/blob/a58545b6c422588ad2e72ccb2783cbd4f9ea5031/internal/cloudapi/v2/imagerequest.go#L434-L436

I can add a comment about it into the code if you think that it would be helpful. Or I can change it, but then arguably, the error returned by `getTarget()` should be changed as well when called on the default target type.

croissanne: oh if it's consistent with the non-bootc flow, then let's leave it for now.

croissanne (approved):
thank you

lzap (approved):

PR #5095, opened 2026-04-14
Update for RHEL 9.9 and 10.3 nightlies

jlsherrill:
- Code updates for osbuild/images v0.255
- Update the [osbuild/images](https://github.com/osbuild/images) version in go.mod to the version that contains the new rhel-X.Y repo definitions.
- Update the ref in schutzbot/terraform to point to the [osbuild/gitlab-ci-terraform](https://github.com/osbuild/gitlab-ci-terraform) commit that contains the new nightly runners.
- Add rhel-X.Y snapshot repo overrides to Schutzfile snapshots for the new distribution.
- Update the [osbuild/osbuild](https://github.com/osbuild/osbuild) ref for all distros in the Schutzfile, to point to the commit from main branch that has the rhel-X.Y RPM builds in our S3 CI cache.
- Add new rhel-X.Y-nightly CI runners to .gitlab-ci.yml job which builds RPMs and also to jobs that run tests on RHEL.
- Update the rhel-X distro aliases defined in the default configuration by changing the values returned by GetDefaultConfig() in cmd/osbuild-composer/config.go. Specifically, make sure that rhel-X is an alias to rhel-X.Y.


.gitlab-ci.yml
```diff
@@ -92,8 +92,12 @@ init:
   RUNNER:
```
jlsherrill, line 92: not sure if all of these are necessary, but followed existing nightlies

thozza: It is until 9.8/10.2 goes GA

Schutzfile
```diff
@@ -132,34 +139,34 @@
...
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-baseos-n9.9-20260405"
           },
           {
             "title": "RHEL-9-RPMREPO-NIGHTLY-AppStream",
             "name": "appstream",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-appstream-n9.8-20260405"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-appstream-n9.9-20260405"
           },
           {
             "title": "RHEL-9-RPMREPO-NIGHTLY-CRB",
```
thozza: This is wrong, you need to keep repo overrides for both nightlies (9.8 and 9.9) to keep them working. Otherwise, 9.8 will try to use CDN, which will fail.

Schutzfile
```diff
@@ -192,34 +206,34 @@
...
             "title": "RHEL-10-RPMREPO-NIGHTLY-BaseOS",
             "name": "baseos",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el10/el10-x86_64-baseos-n10.2-20260405"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el10/el10-x86_64-baseos-n10.3-20260405"
           },
           {
             "title": "RHEL-10-RPMREPO-NIGHTLY-AppStream",
             "name": "appstream",
-            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el10/el10-x86_64-appstream-n10.2-20260405"
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el10/el10-x86_64-appstream-n10.3-20260405"
```
thozza: ditto keeping repo overrides for 10.2 nightly

thozza (changes requested):
Thanks.

There is an issue with repo overrides for 9.8 and 10.2 in the Schutzfile, but the rest looks good.

croissanne:
Nitckpick on the format of the commit messages: https://osbuild.org/docs/developer-guide/general/workflow/#preferred-commit-message-format

Schutzfile
jlsherrill: ahah! thanks for the explanation

jlsherrill:
updated!

thozza (dismissed):
Thanks.

osbuild-composer.spec
```diff
@@ -8,7 +8,7 @@
 %bcond_with relax_requires
 
 # The minimum required osbuild version
-%global min_osbuild_version 175
+%global min_osbuild_version 178
```
thozza, line 11: you need to also update pinned osbuild ref in the `Schutzfile` for all distros, so that this version is available during the provisioning.

Schutzfile
```diff
@@ -9,7 +9,7 @@
   "fedora-42": {
     "dependencies": {
       "osbuild": {
-        "commit": "89601f77042d28dcd7a4b192c901d94ffea8d251"
+        "commit": "eb2d3de47cebc09d5bc68e215f14efa0e5941594"
```
thozza: https://github.com/osbuild/osbuild/commit/eb2d3de47cebc09d5bc68e215f14efa0e5941594 is not new enough. While it satisfies the minimum osbuild version dependency, the commit is from before you added the 9.9 / 10.3 workers to osbuild CI, so there are no RPM builds of osbuild for these releases. That's the reason why CI is failing on those releases in this PR.

jlsherrill:
This is now 🍏 check the last few commits for the changes that fixed things

thozza (dismissed):

achilleas-k (approved):

thozza (approved):

PR #5096, opened 2026-04-14
spec: update composer-cli dependency to weldr-client

yselkowitz:
This was renamed long ago, and the Obsoletes/Provides for the old name was recently dropped:

https://github.com/osbuild/weldr-client/commit/349392c16fbac610e3cb71b09f5dcdf3929c8b6e

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -21,7 +21,7 @@ A service for building customized OS artifacts, such as VM images and OSTree
 commits, that uses osbuild under the hood. Besides building images for local
 usage, it can also upload images directly to cloud.
 
-It is compatible with composer-cli and cockpit-composer clients.
+It is compatible with weldr-client and cockpit-composer clients.
 }
 
 Name:           osbuild-composer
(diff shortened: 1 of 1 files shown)


brlane-rht (commented):
I'm slightly opposed to changing this. `weldr-client` is technically the library and `composer-cli` is the cmdline tool provided by `weldr-client`. It's unlikely to happen, but leaving it like this allows it to be replaced by something else providing `composer-cli`.

yselkowitz:
Leaving this as is makes the tests subpackage uninstallable.

supakeen:
> Leaving this as is makes the tests subpackage uninstallable.

Would it also be fixed by re-adding the (probably erroneously dropped) `Provides` in `weldr-client`? That would probably make @bcl happy and the `-tests` package at the same time? 🙂

yselkowitz:
> Would it also be fixed by re-adding the (probably erroneously dropped) `Provides` in `weldr-client`? That would probably make @bcl happy and the `-tests` package at the same time? 🙂

That would also fix it.

brlane-rht:
Ah! I didn't realize that had gotten lost in the change and when I looked at this yesterday using dnf repoquery I must have been looking at the wrong version :(
I've restored the Obsolete/Provides lines that got lost and started a rawhide build as well as an upstream PR at https://github.com/osbuild/weldr-client/pull/189

PR #5098, opened 2026-04-16
Add end-to-end functional test for bootc composes in a service deployment (HMS-10008)

thozza:
No existing test covers the full production-like bootc compose pipeline: Cloud API with JWT auth, private container registry, AWS EC2 executor, cloud upload target, and image verification — all in one flow. This PR adds `api-bootc-service.sh`, a new test driver that exercises this complete pipeline, along with supporting infrastructure: a reusable executor setup helper, a bootc guest-image handler for S3 uploads, and bootc container ref pinning in the Schutzfile.

## Architectural Changes

- **Handler pattern** matching the existing `api.sh` approach — pluggable image-type verification via sourced scripts in `api/bootc/`. The guest-image handler (`guest.s3.sh`) sources the non-bootc `aws.s3.sh` handler and overrides only `checkEnv()`, `createReqFile()`, and `verify()`, minimizing duplication.
- **Executor setup extracted to shared helper** (`api/common/executor.sh`) — provides `setupExecutorKeypair()`, `waitForExecutorInstance()`, `verifyExecutorNetworkIsolation()`, `provisionExecutor()`, `startExecutor()`, and `cleanupExecutor()` so both this test and `worker-executor.sh` can share provisioning logic eventually (this is out of scope for this PR).
- **Container registry auth via `REGISTRY_AUTH_FILE` systemd drop-in** instead of a `[containers]` config section — simpler and covers all container operations uniformly (BootcInfoResolveJob, ContainerResolveJob, osbuild).
- **Container source type verification extracted** to `api/common/bootc.sh` (`verifyContainerSourceType()`) for reuse across `api-bootc.sh` and the new service test.

## Key Changes

- Add `test/cases/api-bootc-service.sh` — standalone test driver for the full bootc service compose pipeline (JWT, private registry, executor, S3 upload, image verification)
- Add `test/cases/api/bootc/guest.s3.sh` — bootc-specific handler for `guest-image` with `aws.s3` upload target, reusing S3 verification/cleanup from the existing `aws.s3.sh` handler
- Add `test/cases/api/common/executor.sh` — reusable AWS EC2 executor setup/teardown helper extracted from the `worker-executor.sh` pattern
- Add `test/cases/api/common/bootc.sh` — shared `verifyContainerSourceType()` helper, refactored out of `api-bootc.sh`
- Extend `Schutzfile` with `bootc` container ref mappings (structured as image-type → arch → ref-type) for `rhel-10.1`
- Add `API-bootc-service` CI job in `.gitlab-ci.yml` targeting `aws/rhel-10.1-ga-x86_64` with the `worker-executor` IAM profile

## Testing

- New `api-bootc-service.sh` test exercises the full production-like pipeline: JWT-authenticated Cloud API compose with a private bootc container ref, built on an AWS EC2 executor, uploaded to S3, and verified offline with `osbuild-image-info`
- Pre- and post-compose checks validate that the bootc container ref transitions from absent to present in local storage (confirming BootcInfoResolveJob pulled it correctly)
- Executor network isolation is verified — the security group only allows egress to the worker host, with temporary internet access granted only during provisioning and revoked afterward
- Compose manifest is validated to confirm the remote container source (`org.osbuild.skopeo`) is used when `use_remote_container_source = true`

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -222,6 +222,13 @@
     "dependencies": {
       "osbuild": {
         "commit": "d36f11818f6f7b301ba392c6f7378eb908c1f943"
+      },
+      "bootc": {
+        "guest-image": {
+          "x86_64": {
+            "base": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2:latest"
+          }
+        }
       }
     }
   },
--- a/test/cases/api-bootc-service.sh
+++ b/test/cases/api-bootc-service.sh
@@ -1,0 +1,441 @@
+#!/usr/bin/bash
+
+#
+# End-to-end functional test for bootc composes in a service deployment.
+# Tests the full pipeline: Cloud API + JWT + private registry + AWS EC2
+# executor + cloud upload target + image verification.
+#
+# Usage: api-bootc-service.sh <image-type>
+#   e.g.: api-bootc-service.sh guest-image
+#
+
+set -euo pipefail
+
+source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+source /usr/libexec/tests/osbuild-composer/shared_lib.sh
+source /usr/libexec/tests/osbuild-composer/api/common/image-types.sh
+source /usr/libexec/tests/osbuild-composer/api/common/bootc.sh
+source /usr/libexec/tests/osbuild-composer/api/common/executor.sh
+
+if (( $# != 1 )); then
+    redprint "Usage: $0 <image-type>"
+    redprint "Supported image types: ${IMAGE_TYPE_GUEST}"
+    exit 1
+fi
+
+IMAGE_TYPE="$1"
+
+# Select handler based on image type
+case "${IMAGE_TYPE}" in
+    "$IMAGE_TYPE_GUEST")
+        source /usr/libexec/tests/osbuild-composer/api/bootc/guest.s3.sh
+        ;;
+    *)
+        redprint "Unknown image type: ${IMAGE_TYPE}"
+        exit 1
+        ;;
+esac
+
+ARTIFACTS="${ARTIFACTS:-/tmp/artifacts}"
+
+# Container image used for cloud provider CLI tools (used by sourced handlers)
+export CONTAINER_IMAGE_CLOUD_TOOLS="quay.io/osbuild/cloud-tools:latest"
+
+# This test relies on podman-specific behavior (authfile + image exists checks).
+if ! type -p podman 2>/dev/null >&2; then
+    redprint "podman is required for api-bootc-service.sh."
+    exit 2
+fi
+CONTAINER_RUNTIME=podman
+
+#
+# Resolve bootc container ref from Schutzfile
+#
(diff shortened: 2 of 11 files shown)


Schutzfile
```diff
@@ -176,6 +176,13 @@
     "dependencies": {
       "osbuild": {
         "commit": "89601f77042d28dcd7a4b192c901d94ffea8d251"
+      },
+      "bootc": {
+        "guest-image": {
+          "x86_64": {
+            "base": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2:latest"
```
achilleas-k, line 229: Do we want to pin this or are we fine with `:latest` because we control the builds?

test/cases/api/common/executor.sh
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/bash
+
+#
+# This module contains functions for setting up and cleaning up the executor.
+#
+# It expects the following variables to be set by the caller:
```
brlane-rht, line 6: That's a nice comprehensive list, maybe it should validate that all of that is true?

thozza: Well, adding a standalone function to do the validation seems too excessive for this type of code and doing it at the source time is hard, because some of the variables can be and are defined after the script being sourced. The truth is that these need to be defined before the functions that use them are called.

I must admit that I'm not sure what would be the best solution here.

achilleas-k (dismissed):
Really nice.  LGTM.

Schutzfile
thozza: I think that we should pin a specific build once we actually boot-test the container images from the `bootc-foundry` repo.

But we are not there yet, so for the purpose of adding the test in its current form, using the `latest` tag seemed like a reasonable compromise. We can easily start pinning the ref at any point, which was my motivation for adding this mechanism in the first place.

thozza:
I had to rebase and force-push due to conflict in `Schutzfile` that was a result of https://github.com/osbuild/osbuild-composer/pull/5095

test/cases/api-bootc-service.sh
```diff
@@ -0,0 +1,441 @@
...
+
+IMAGE_TYPE="$1"
+
+# Select handler based on image type
+case "${IMAGE_TYPE}" in
+    "$IMAGE_TYPE_GUEST")
+        source /usr/libexec/tests/osbuild-composer/api/bootc/guest.s3.sh
+        ;;
+    *)
+        echo "Unknown image type: ${IMAGE_TYPE}"
```
brlane-rht: Could you use redprint for the errors here? It sure makes it easier to find problems when skimming logs.

thozza: I changed also a few other places to use `redprint` instead of `echo`

Schutzfile
achilleas-k: SGTM

achilleas-k (approved):
Rebase and redprints 👍

brlane-rht (approved):
Looks good!

PR #5099, opened 2026-04-17
SPEC: add explicit Requires on podman for worker subpackage

thozza:
The worker handles jobs, such as the BootcInfoResolve, which implicitly depend on the presence of the podman CLI. Let's add an explicit dependency to the worker package, to ensure that podman is present on the system after osbuild-composer-worker package installation.


lzap (approved):

brlane-rht (approved):
Ideally we'd be pulling a list of requires from images, since that's what is actually requiring podman, but since it is a vendored module that's not simple :)

PR #5102, opened 2026-04-17
Enable all bootc-supported disk image types for bootc composes, not just qcow2 (HMS-10508)

thozza:
The bootc compose flow previously rejected every image type except `qcow2` (guest-image), even though the underlying bootc distro already supports several disk formats. This PR replaces the hardcoded check with a dynamic validation helper that queries the bootc distro's image type registry, so any disk image type recognized by the bootc distro is automatically accepted. In practice this unlocks the Cloud API image types that overlap with bootc disk formats: `aws` (ami), `azure` (vhd), `gcp` (gce), `vsphere` (vmdk), `vsphere-ova` (ova), and `pxe-tar-xz` - installer types remain out of scope.

## Architectural Changes

Introduce `bootcSupportedImageType` in a new `bootc.go` file. Rather than maintaining a static allowlist of image types in composer (which would drift from the bootc distro definitions in `osbuild/images`), the helper constructs a lightweight dummy `BootcDistro` via `generic.NewBootc` and probes it for the requested image type. This keeps composer in sync with supported types automatically. A TODO notes that a dedicated upstream helper (e.g. `generic.BootcSupportedImageTypes`) would be cleaner and should be pursued as a follow-up.

## Key Changes

- Add `bootcSupportedImageType` helper that validates image types against the bootc distro at runtime
- Replace the hardcoded `imageTypeName != "qcow2"` guard in `enqueueBootcCompose` with a call to the new helper
- Parameterize `TestComposeBootc` to exercise `aws` (ami) and `vsphere` (vmdk) alongside the existing `guest-image` (qcow2) cases
- Extend pre-manifest happy-path tests with `ami`/AWS and `vhd`/Azure upload target scenarios
- Update the `api-bootc.sh` integration test to loop over `guest-image` and `aws`, verifying the full compose-build-download path for both
- Switch bootc-related tests from `test_distro.TestArch3Name` to literal `"x86_64"` since the bootc flow builds its own distro and needs a real architecture

## Breaking Changes

This PR is fully backward compatible. All previously supported requests (`guest-image`/qcow2) continue to work. Additional image types that were previously rejected with a 400 error are now accepted if supported by the bootc distro.

## Testing

- Added `TestBootcSupportedImageType` covering all expected supported types (guest-image, aws, azure, gcp, vsphere, vsphere-ova, pxe-tar-xz on x86_64; guest-image and aws on aarch64), unsupported types (minimal-raw, image-installer, ec2), and invalid architecture input
- Parameterized `TestComposeBootc` and pre-manifest happy-path tests to verify the full enqueue and manifest-generation flow for non-qcow2 image types with matching upload targets
- Extended the `api-bootc.sh` integration test to compose both `guest-image` and `aws` with local upload targets, providing end-to-end CI coverage for the newly enabled formats

### DIFF
--- a/internal/cloudapi/v2/server.go
+++ b/internal/cloudapi/v2/server.go
@@ -579,9 +579,8 @@ func (s *Server) enqueueBootcCompose(request ComposeRequest, channel string) (uu
 
 	imageTypeName := imageTypeFromApiImageType(ir.ImageType)
 
-	// TODO: Only qcow2 (guest-image) is supported for bootc composes for now.
-	if imageTypeName != "qcow2" {
-		return uuid.Nil, HTTPErrorWithDetails(ErrorUnsupportedImageType, nil, "only qcow2 (guest-image) is supported for bootc composes")
+	if err := bootcSupportedImageType(ir.Architecture, imageTypeName); err != nil {
+		return uuid.Nil, err
 	}
 
 	// Validate and normalize upload targets — consistent with non-bootc flow.
--- a/internal/cloudapi/v2/bootc.go
+++ b/internal/cloudapi/v2/bootc.go
@@ -1,0 +1,76 @@
+package v2
+
+import (
+	"fmt"
+
+	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/bib/osinfo"
+	"github.com/osbuild/images/pkg/bootc"
+	"github.com/osbuild/images/pkg/distro/generic"
+)
+
+// bootcSupportedImageType checks whether the given image type name is supported
+// for bootc composes on the given architecture by constructing a dummy bootc
+// distro and querying its image type registry.
+//
+// NOTE: This constructs a dummy bootc distro with placeholder values solely to
+// query the set of supported image types from the bootc-generic YAML definitions
+// in osbuild/images. The image type names and their availability are determined
+// by the YAML and do not depend on the actual container metadata. The dummy
+// values for Imgref, DefaultRootFs, Size, and OSInfo are required by the
+// NewBootc constructor validation but are not used for image type listing.
+//
+// TODO: Consider adding a dedicated helper to the osbuild/images library
+// (e.g. generic.BootcSupportedImageTypes) that returns the list of supported
+// image type names without requiring a full bootc.Info. This would eliminate
+// the need for dummy values and make the contract less fragile.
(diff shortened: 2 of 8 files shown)


brlane-rht (approved):
Looks good

internal/cloudapi/v2/server.go
```diff
@@ -579,9 +579,8 @@ func (s *Server) enqueueBootcCompose(request ComposeRequest, channel string) (uu
 
 	imageTypeName := imageTypeFromApiImageType(ir.ImageType)
 
-	// TODO: Only qcow2 (guest-image) is supported for bootc composes for now.
-	if imageTypeName != "qcow2" {
-		return uuid.Nil, HTTPErrorWithDetails(ErrorUnsupportedImageType, nil, "only qcow2 (guest-image) is supported for bootc composes")
+	if err := bootcSupportedImageType(ir.Architecture, imageTypeName); err != nil {
+		return uuid.Nil, err
```
achilleas-k, line 583: The ergonomics of this function are a bit weird.  I would have expected something like this to return a boolean and _maybe_ an error in addition to distinguish failures (failed to init distro) from the plain "nope" response to question of arch+imagetype support.

achilleas-k (approved):
Very nice.  Thanks.  Tiny comment inline, but wont block.

LGTM!

internal/cloudapi/v2/server.go
thozza: That's fair and valid point. I did consider `(bool, error)` but ended up here intentionally due to these reasons:

- There's no state where `false` is returned with a `nil` error - an error is the "nope." A bool would be redundant and would introduce a representable, but meaningless state (`false`, `nil`) or (`true`, `error`).
- This is an unexported, single-caller helper extracted mainly for testability; otherwise it'd be inlined in `enqueueBootcCompose()`. The caller just passes the error through, so the signature is optimized for that one call site, not for general reuse.
- The dummy distro construction failing is effectively a programming error guarded by unit tests. It should never happen at runtime, so distinguishing it from "unsupported image type" at the caller level doesn't add much value.
- Returning dedicated error types and then making the caller `switch/errors.As` just to re-map them back to the same HTTP errors already produced inside the helper would mean additional code to achieve identical behavior.

That said, I'd be happy to revisit if you see a concrete scenario where the caller would need to branch on failure mode.

PR #5108, opened 2026-04-22
Add optional container image cleanup to bootc info resolve job and enable it for the service (HMS-10446)

thozza:
SaaS workers accumulate container images over time as a side effect of resolving bootc container info via `podman run`. These images remain in local storage indefinitely, eventually exhausting disk space on long-lived workers. This PR introduces a config-driven, best-effort cleanup mechanism that removes pulled container images after each BootcInfoResolve job completes, gated behind a new `[bootc_info_resolve]` TOML config section so on-prem behavior is unchanged by default.

## Architectural Changes

- Cleanup uses per-ref `defer` statements inside the resolution loop, ensuring images are removed even on mid-loop errors or panics, with LIFO ordering relative to the existing panic-recovery defer.
- Image removal is best-effort: failures are logged as warnings but never fail the job, since the resolution itself already succeeded.

## Key Changes

- Add `[bootc_info_resolve]` config section with a `cleanup_images` flag (default `false`) to `workerConfig`
- Enable `cleanup_images = true` in the SaaS deployment template (`osbuild-worker.toml`)
- Update the `api-bootc-service` E2E test to enable cleanup and verify images are removed post-compose

## Breaking Changes

This PR is fully backward compatible. The new config section is optional and cleanup defaults to `false`, preserving existing on-prem behavior.

## Testing

- Add four unit test cases covering: cleanup on successful resolution, partial cleanup on mid-loop failure, cleanup disabled (no-op), and cleanup failure being non-fatal to the job
- Update the `api-bootc-service` E2E test to enable cleanup and flip the post-compose assertion from "image exists" to "image does not exist"

### DIFF
--- a/cmd/osbuild-worker/config.go
+++ b/cmd/osbuild-worker/config.go
@@ -82,6 +82,14 @@ type repositoryMTLSConfig struct {
 	Proxy          string `toml:"proxy"`
 }
 
+type bootcInfoResolveConfig struct {
+	// NOTE: When enabled, the worker removes container images from local
+	// storage after resolving bootc info. This prevents storage buildup
+	// in SaaS deployments where the worker processes many different
+	// container refs over time.
+	CleanupImages bool `toml:"cleanup_images"`
+}
+
 type workerConfig struct {
 	Composer       *composerConfig             `toml:"composer"`
 	Koji           map[string]kojiServerConfig `toml:"koji"`
@@ -96,8 +104,9 @@ type workerConfig struct {
 	BasePath string `toml:"base_path"`
 	DNFJson  string `toml:"dnf-json"`
 	// default value: &{ Type: host }
-	OSBuildExecutor      *executorConfig       `toml:"osbuild_executor"`
-	RepositoryMTLSConfig *repositoryMTLSConfig `toml:"repository_mtls"`
+	OSBuildExecutor      *executorConfig         `toml:"osbuild_executor"`
+	RepositoryMTLSConfig *repositoryMTLSConfig   `toml:"repository_mtls"`
+	BootcInfoResolve     *bootcInfoResolveConfig `toml:"bootc_info_resolve"`
 	// something like "production" or "staging" to be added to logging
 	DeploymentChannel string `toml:"deployment_channel"`
 	// clean store between runs, this should only be used with workers running on AWS within an ASG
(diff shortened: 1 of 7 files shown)


lzap:
> * Image removal is best-effort: failures are logged as warnings but never fail the job, since the resolution itself already succeeded.

This is a security problem, inability to clean up must stop the worker. This was discussed in https://github.com/osbuild/osbuild-composer/pull/5039

thozza:
> > * Image removal is best-effort: failures are logged as warnings but never fail the job, since the resolution itself already succeeded.
> 
> This is a security problem, inability to clean up must stop the worker. This was discussed in #5039

I think the security concern from #5039 doesn't apply here - the threat models are different.

In #5039, the osbuild **store** can contain artifacts from a prior image build. If it's not cleaned, a subsequent build can access those artifacts - a malicious RPM scriptlet could read sources that belong to a different image build. That's a cross-build data leakage vector, so failure to clean the store is rightly treated as critical.

Container images in podman local storage don't have the same exposure. The BootcInfoResolve job resolves a specific image ref from its job args via podman - it can't discover or access other images in local storage. The images are also not mounted into the osbuild build sandbox, so there's no path for a build job to reach them. A leftover image from a prior job is just wasted disk space, not a data leakage risk.

The real security-sensitive part of this flow is the `podman exec` we run against binaries inside the container during inspection. But that's an existing concern (and orthogonal to cleanup) - if we ever allow untrusted 3rd-party container refs, we'd need to run the entire inspection on an isolated executor, same as we do for osbuild. Cleanup wouldn't change that threat model either way.

So I'd keep cleanup as best-effort: warn and move on. The worst case is a worker gradually filling disk, which is an operational issue (and detectable via monitoring), not a security one.

lzap (approved):
This is good. Going forward, we could just delete tags and leverage podman's `image prune` command that removes all unreachable commits. This "garbage collection" would only happen when some storage limit is met.

thozza:
> This is good. Going forward, we could just delete tags and leverage podman's `image prune` command that removes all unreachable commits. This "garbage collection" would only happen when some storage limit is met.

Yes, this solution is the bare minimum, so that we have at least some minimal cleanup. It can be later turned off and we can add an external cleanup mechanism or rework it.

kingsleyzissou (approved):

PR #5109, opened 2026-04-22
cicd: split gobump into two jobs

lzap:
When called serially, there can be leftovers in vendor/ dir that might
be accidentally commited. Running this in two separate and isolated jobs
is better.


avitova (approved):
Interesting. Let's try then. 🤔

thozza (approved):

PR #5110, opened 2026-04-22
weldr: Use the job id instead of creating a new one

brlane-rht:
When a weldr api compose is started it stores information about the build in the weldr store, using a new UUID that is different than the job id returned by the compose. This leads to weldr-client showing two different UUIDs for the same compose, which is confusing.

This change switches to using the same job id UUID for the weldr store as it uses for the compose job. This will allow weldr-client to associate the information and combine it when displaying the composes.

This will have no impact on any other part of the system. The weldr store is only used by the internal store module and the weldr api.

Related: HMS-10332

### DIFF
--- a/cmd/osbuild-store-dump/main.go
+++ b/cmd/osbuild-store-dump/main.go
@@ -181,7 +181,6 @@ func main() {
 		[]*target.Target{
 			awsTarget,
 		},
-		id1,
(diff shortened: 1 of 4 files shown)


thozza (approved):
Makes sense. FWIW, this is consistent with the UUID returned by Cloud API compose endpoint, which is also the leaf job ID (osbuild job ID or koji-finalize job ID).

lzap (approved):

PR #5112, opened 2026-04-23
Enable bootc installer ISO builds via Cloud API (HMS-10568)

thozza:
Add support for building bootable container ISOs through the Cloud API v2. The vendored `osbuild/images` library already supports the `bootc-generic-iso` image type - which creates ISOs directly from a bootable container via GRUB2/squashfs, no anaconda - but it was unreachable from the API. This patch set adds a `bootable-container-iso` image type and an optional `iso_payload_reference` field to the `Bootc` schema, wires them through the existing bootc compose pipeline, and adds unit and e2e test coverage.

## Architectural Changes

- **API name decoupled from internal name**: The API exposes `bootable-container-iso` which maps to `bootc-generic-iso` internally, keeping the images library naming out of the public surface (especially the "bootc" part).
- **No job graph changes**: The existing bootc pipeline (BootcInfoResolve → BootcPreManifest → ContainerResolve → ManifestByID → OSBuild) handles the optional payload container without modification. The payload ref flows through `ImageOptions.Bootc.InstallerPayloadRef` and the manifest's container source extraction picks it up automatically.
- **Early rejection with dedicated errors**: Two new error codes (46, 47) catch misuse at the API boundary - `iso_payload_reference` on non-ISO types and `bootable-container-iso` in distribution-based composes - instead of letting requests fail deep in distro lookup or manifest generation.

## Key Changes

- Add `iso_payload_reference` optional field to the `Bootc` OpenAPI schema and `bootable-container-iso` to the `ImageTypes` enum
- Add error codes 46 (`ErrorIsoPayloadReferenceForbidden`) and 47 (`ErrorBootcOnlyImageType`) for clear validation messages
- Map `bootable-container-iso` → `bootc-generic-iso` in the handler, register it in the S3 upload target support map, and reject it in the distribution-based compose path
- Validate `iso_payload_reference` and pass it through to `ImageOptions.Bootc.InstallerPayloadRef` in `enqueueBootcCompose`
- Add `HasImageType` helper on `ComposeRequest` to check image types across singular/plural request shapes
- Add e2e test handler (`container-iso.s3.sh`) that builds an ISO, downloads from S3, and validates the ISO 9660 format

## Breaking Changes

## Testing

- Unit tests validate the full compose request lifecycle: ISO with and without payload (201), `iso_payload_reference` on non-ISO type (error 46), `bootable-container-iso` with distribution (error 47), and payload ref propagation to the pre-manifest job
- Pre-manifest test exercises `bootc-generic-iso` manifest generation end-to-end, including `KernelInfo` fixture setup needed by ISO pipelines and pipeline shape assertions (`build` + `os-tree` with 1-2 container sources depending on payload)
- E2e handler downloads the built ISO from S3 and verifies it is a valid ISO 9660 image; payload ref resolution is non-fatal so the test works with or without a configured payload container


internal/cloudapi/v2/openapi.v2.yml
```diff
@@ -1388,6 +1388,13 @@ components:
         reference:
           type: string
           example: "quay.io/centos-bootc/centos-bootc:stream9"
+        installer_payload_ref:
+          type: string
+          description: |
+            Bootc container image reference for the OS to be installed by the
+            installer. Required when image_type is image-installer. Must not be
+            set for non-installer image types.
+          example: "quay.io/centos-bootc/centos-bootc:stream9"
```
achilleas-k: Getting ahead of some of the discussion that might come up here.

This is consistent with the options in ib-cli.  I think some people find it confusing that the main OS container goes here when building an ISO.  I can understand where that confusion comes from, but I maintain that the way we define and use them is more technically correct and preferable.

At the end of the day, the cloud API isn't user-facing so perhaps it's not important what we do here.  But we're all fans of consistency so it should reflect whatever we're doing in the user-facing parts.

We should be very clear about this when describing the API properties, both in the spec and in the code.  I think there is a good chance that the confusion over the meaning of these properties can lead to problems down the line when refactoring or extending the code on either side of the API (i.e. in composer itself, or in the clients, CRC and cockpit-ib).

thozza: I agree with your comment, but I'm unsure if there is any action for me in this PR or in a followup in the `image-builder-crc` repo to enable bootc installer images. :thinking:

Is there any? Or is a proper field description, like the one used here, good enough?

achilleas-k: Sorry, no.  I should have mentioned: There's no action needed in this PR.

We could expand on the description of the new field and maybe add a description to `reference` that described different scenarios, but I don't think it's necessary.

supakeen:
> Remap image-installer → bootc-installer in enqueueBootcCompose, validate the payload ref, and pass it through BootcImageOptions

Please use `bootc-generic-iso` instead of `bootc-installer`. The Containerfiles I added to `bootc-foundry` are assuming that image type and it's the only image type we'll offer in the future. As far your preference on the API side; I don't particularly mind that much but do note that the `bootc-generic-iso` isn't a 'payload' installer per se. It can be whatever is in the container; a live ISO, a network installer, an offline installer, etc.

`bootc-installer` and `anaconda-iso` are (to be) deprecated image types that will be part of the (not yet existing) migration guide.

thozza:
I've reworked the PR to use `bootc-generic-iso` image type. It is now exposed as `bootable-container-iso` image type at the Cloud API level, since we discussed that we may not want to include "bootc" in the name. I think that this name correctly expresses the versatility of the image type.

I've also added the optional `iso_payload_reference` field to the `ComposeRequest.Bootc` object. The Image Mode for Lightspeed service MVP needs to build the ISO with the payload container embedded in the image. However, making the field optional does not restrict us from supporting other variants in the future.

Keeping as Draft until we have a working derived installer container image built in the bootc-foundry repo.

croissanne (approved):

internal/cloudapi/v2/compose.go
```diff
@@ -1476,3 +1476,17 @@ func decodeMinsize(size *Minsize) (uint64, error) {
...
 }
+
+func (request *ComposeRequest) HasImageType(imageType ImageTypes) bool {
+	if request.ImageRequests != nil {
+		for _, ir := range *request.ImageRequests {
+			if ir.ImageType == imageType {
+				return true
+			}
+		}
+	}
```
lzap, line 1487: Could use `slices.ContainsFunc` for probably one or two less lines but this is good.

lzap (approved):

internal/cloudapi/v2/compose.go
thozza: Yeah, I'm still getting used to using these functions instead of loop. I'll submit a followup.

thozza: https://github.com/osbuild/osbuild-composer/pull/5124

PR #5116, opened 2026-04-30
cicd: exec make build for each bump

lzap:
Gobump currently fails: https://github.com/osbuild/osbuild-composer/actions/workflows/gobump.yml

The reason is incompatibility between OpenAPI libraries: https://github.com/oapi-codegen/oapi-codegen/pull/2295

Symptoms are that the project will not build and this only happens at the very and of the bump process when `prepare-source.sh` is called. Then something fails at this point, gobump just exists and never files a PR.

```
+ go generate ./cmd/... ./internal/... ./pkg/...
# github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen
../../../vendor/github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen/schema.go:909:13: invalid operation: v == element.Ref (mismatched types openapi3.MappingRef and string)
internal/cloudapi/v2/handler.go:1: running "go": exit status 1
# github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen
../../../vendor/github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen/schema.go:909:13: invalid operation: v == element.Ref (mismatched types openapi3.MappingRef and string)
internal/worker/api/api.go:1: running "go": exit status 1
```

This aims to fix or improve this situation. The change will now build the project for each and every bumped direct dependency. This should prevent such situations and automatically exclude the offending library from the update trying 5 times to find latest stable release that was previously working.

I am pushing a branch directly to `osbuild` org so I can trigger this manually and test it.

Edit: Previously this was also parametrization of include/exclude but I dropped this from the PR.

### DIFF
--- a/.github/workflows/gobump.yml
+++ b/.github/workflows/gobump.yml
@@ -19,6 +19,8 @@ jobs:
         with:
           go_version: "1.24.12"
           token: ${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}
+          exec: ./tools/prepare-source.sh
+          exec2: "make build"
           exec_pr: ./tools/prepare-source.sh
           include: "github.com/osbuild/images github.com/osbuild/blueprint"
           commit_message: "deps: bump osbuild/images dependency"
@@ -34,6 +36,8 @@ jobs:
         with:
           go_version: "1.24.12"
           token: ${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}
+          exec: ./tools/prepare-source.sh
+          exec2: "make build"
           exec_pr: ./tools/prepare-source.sh
           exclude: "github.com/osbuild/images,github.com/osbuild/blueprint"
           commit_message: "deps: update dependencies (w/o osbuild/images)"


lzap:
Gobump is now passing: https://github.com/osbuild/osbuild-composer/actions/runs/25155760175

(manually triggered against this branch)

.github/workflows/gobump.yml
```diff
@@ -3,6 +3,19 @@ name: "Updates Go dependencies via gobump"
 
 on:  # yamllint disable-line rule:truthy
   workflow_dispatch:
+    inputs:
+      include:
+        description: >
+          Module paths for the images-only bump job (space-separated).
+        required: false
+        type: string
+        default: "github.com/osbuild/images github.com/osbuild/blueprint"
```
brlane-rht: Why are includes space separated and excludes comma separated?

brlane-rht (dismissed):
Looks good!

.github/workflows/gobump.yml
lzap: Cos the underlaying `gobump` utility uses arguments for one, command line option for the other.

I wish it was not the case for the action but it is all over the place already. :-D

supakeen (dismissed):

lzap:
It looks like your GitLab CI job failed primarily because it couldn't find and install the specific version of the osbuild-composer-worker package requested by your test script.

Here is a breakdown of the critical issues found in the log:

1. The "Smoking Gun": Package Version Mismatch
The script explicitly tried to verify the installation of version 169 of the worker, but version 167 was actually installed.

The Command: rpm -q osbuild-composer-worker-169

The Result: package osbuild-composer-worker-169 is not installed

Why it happened: Earlier in the logs, the script searched for a repository containing version 169. It failed to find a valid repo for that specific commit (795effd...) and automatically fell back to a parent commit (3f0933e...). That parent commit only contained version 167, which was then installed. Since the test strictly requires 169, the job exited with an error.

2. Forbidden Access (403 Errors)

During the environment preparation, the runner attempted to download the gitlab-runner RPM from an S3 bucket and was repeatedly denied:
[FAILED] gitlab-runner_amd64.rpm: Status code: 403 for https://s3.dualstack.us-east-1.amazonaws.com/...
While the job eventually moved past this (likely using a cached version or a fallback), it indicates a potential issue with the permissions or the URL path in your deployment scripts.

croissanne:
Note that oapi-codegen had another release, including kin-openapi v0.135, so it should work now. but this is still good to have

lzap:
Ok I dropped the parametrization from the PR, leaving just the extra build command in between gobump updates. This will prevent such problems when a dependency update cannot be built.

croissanne (approved):

lzap:
Run against the current branch: https://github.com/osbuild/osbuild-composer/actions/runs/25746063399

achilleas-k (approved):

PR #5125, opened 2026-05-06
cloudapi/v2: add default target for bootc iso

croissanne:
The default upload target was missing for the bootable-container-iso
    type. For requests coming from image builder crc, the default target is
    always used. Thus, this resulted in an "unsupported image type" error.
    
    Also add error details to the getDefaultTarget function in case the
    image type is unsupported.

### DIFF
--- a/internal/cloudapi/v2/imagerequest.go
+++ b/internal/cloudapi/v2/imagerequest.go
@@ -249,6 +249,8 @@ func getDefaultTarget(imageType ImageTypes) (UploadTypes, error) {
 		fallthrough
 	case ImageTypesVsphereOva:
 		fallthrough
+	case ImageTypesBootableContainerIso:
+		fallthrough
 	case ImageTypesWsl:
 		return UploadTypesAwsS3, nil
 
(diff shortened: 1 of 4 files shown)


internal/cloudapi/v2/imagerequest.go
```diff
@@ -281,7 +283,7 @@ func getDefaultTarget(imageType ImageTypes) (UploadTypes, error) {
 		return UploadTypesOciObjectstorage, nil
 
 	default:
-		return "", HTTPError(ErrorUnsupportedImageType)
+		return "", HTTPErrorWithDetails(ErrorUnsupportedImageType, nil, fmt.Sprintf("image type %s does not have a default target", imageType))
```
achilleas-k, line 286: ❤️

achilleas-k (approved):

lzap (approved):

internal/cloudapi/v2/imagerequest.go
lzap: Yeah was quite confusing.

supakeen (approved):

thozza (approved):
Thanks!

PR #5132, opened 2026-05-11
test: Extract failure/retry/oauth tests from api.sh

ochosi:
The `api.sh` test script runs ~50 times per pipeline (once per image-type/runner combination in the CI matrix). It contains three sub-tests that exercise infrastructure logic completely orthogonal to the image type being built.
This makes parsing the logs for the actual failure harder, because e.g. the compose failure path will always print errors to the command-line output.

- **Compose failure path** — submits a bogus package, waits for `failure` status (depsolve-level, no cloud interaction)
- **Job retry** — stops the worker mid-build, asserts DB `retries = 1` (pure job-queue logic)
- **OAuth2/JWT** — reconfigures to JWT mode, tests Bearer token behavior on public vs protected endpoints

Running these 50 times tests the same code path 50 times.

### Solution

1. **Extract shared helpers** into `test/cases/api/common/composer-db.sh` — provides `setup_db`, `teardown_db`, `write_tls_composer_config`, `sendCompose`, `waitForState`, and `collectMetrics`. Both `api.sh` and the new script source this.

2. **Create `test/cases/api-unit.sh`** — a combined test script that runs all three sub-tests sequentially in one job. It prints a pass/fail/skip summary table at the end for easy log reading.

3. **Remove the sub-tests from `api.sh`** — the Prometheus metrics check stays (it's just 3 lines around the compose that must happen anyway).

4. **Add a single `API-unit` CI job** on `aws/rhel-10.1-ga-x86_64` — one job replaces ~50 redundant runs.

### Test summary output

The new script tracks results and prints a summary at exit:

```
======================================================================
  TEST SUMMARY
======================================================================
  Compose failure path (bad package)                 [PASS]
  Module hotfixes failure path                       [SKIP]
  Job retry (worker crash)                           [PASS]
  OAuth2/JWT authentication                          [PASS]
======================================================================
  RESULT: ALL TESTS PASSED
======================================================================
```

### DIFF
--- a/test/cases/api-unit.sh
+++ b/test/cases/api-unit.sh
@@ -1,0 +1,295 @@
+#!/usr/bin/bash
+
+#
+# Combined unit tests for the Cloud API v2 that do NOT depend on a specific
+# image type or cloud provider. Covers:
+#   1. Compose failure path (non-existent package -> "failure" status)
+#   2. Worker crash / job retry (DB retries column)
+#   3. OAuth2/JWT authentication (mock OpenID provider, Bearer tokens)
+#
+# These tests were extracted from api.sh because they exercise infrastructure
+# logic that is orthogonal to the image-type matrix, eliminating ~50 redundant
+# runs per pipeline.
+#
+
+set -uo pipefail
+
+source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+source /usr/libexec/tests/osbuild-composer/shared_lib.sh
+
+# ---------------------------------------------------------------------------
+# Test result tracking
+# ---------------------------------------------------------------------------
+declare -a TEST_NAMES=()
+declare -a TEST_RESULTS=()
+
+function record_result() {
+    TEST_NAMES+=("$1")
+    TEST_RESULTS+=("$2")
+}
+
+function print_summary() {
+    echo ""
+    echo "======================================================================"
+    echo "  TEST SUMMARY"
+    echo "======================================================================"
+    local failed=0
+    for i in "${!TEST_NAMES[@]}"; do
+        local status="${TEST_RESULTS[$i]}"
+        local name="${TEST_NAMES[$i]}"
+        if [ "$status" = "PASS" ]; then
+            printf "  %-50s [PASS]\n" "$name"
+        elif [ "$status" = "SKIP" ]; then
+            printf "  %-50s [SKIP]\n" "$name"
+        else
+            printf "  %-50s [FAIL]\n" "$name"
+            failed=1
+        fi
+    done
+    echo "======================================================================"
+    if [ "$failed" = 1 ]; then
+        echo "  RESULT: SOME TESTS FAILED"
+    else
+        echo "  RESULT: ALL TESTS PASSED"
+    fi
+    echo "======================================================================"
+    echo ""
+    return $failed
+}
+
+# ---------------------------------------------------------------------------
+# Provision and setup (failures here are fatal -- no point running tests)
+# ---------------------------------------------------------------------------
+set -e
+
+/usr/libexec/osbuild-composer-test/provision.sh
+
+source /usr/libexec/tests/osbuild-composer/api/common/composer-db.sh
+source /usr/libexec/tests/osbuild-composer/api/common/common.sh
+
+greenprint "Setting up PostgreSQL database"
+setup_db
+
+greenprint "Writing TLS composer config"
+write_tls_composer_config
+
+sudo systemctl restart osbuild-composer
+
+WORKDIR=$(mktemp -d)
+function cleanups() {
+    set +eu
+    print_summary
+    FINAL_RC=$?
+
+    greenprint "Cleaning up"
+    /usr/libexec/osbuild-composer-test/run-mock-auth-servers.sh stop || true
+    teardown_db
+    sudo rm -rf "$WORKDIR"
+
+    exit $FINAL_RC
+}
+trap cleanups EXIT
+
+ARCH=$(uname -m)
+DISTRO="$ID-${VERSION_ID}"
+REQUEST_FILE="${WORKDIR}/compose_request.json"
+
+cat > "$REQUEST_FILE" << EOF
+{
+  "distribution": "$DISTRO",
+  "customizations": {
+    "packages": [
+      "bash"
+    ]
+  },
+  "image_request": {
+    "architecture": "$ARCH",
+    "image_type": "guest-image",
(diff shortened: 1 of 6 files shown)


test/cases/api/common/composer-db.sh
```diff
@@ -0,0 +1,148 @@
...
+ca = "/etc/osbuild-composer/ca-crt.pem"
+pg_host = "localhost"
+pg_port = "5432"
+pg_database = "osbuildcomposer"
+pg_user = "postgres"
+pg_password = "foobar"
+pg_ssl_mode = "disable"
+pg_max_conns = 10
+EOF
+}
```
thozza: I'd expect more focused config snippet that configures only DB parts, not also authentication.

test/cases/api/common/composer-db.sh
```diff
@@ -0,0 +1,148 @@
...
+                 https://localhost/api/image-builder-composer/v2/compose)
+
+    if [ "$HTTPSTATUS" != "201" ]; then
+        redprint "Sending compose request failed:"
+        cat "$OUTPUT"
+        return 1
+    fi
+
+    COMPOSE_ID=$(jq -r '.id' "$OUTPUT")
+}
```
thozza: This helper should not be in this file at all. It has nothing to do with the DB setup. `test/cases/api/common/common.sh` seems like the natural place for things like this.

Also deduplicate https://github.com/osbuild/osbuild-composer/blob/aaf4eb9999df0c8af2a7d51f25061d20245c7541/test/cases/api-bootc-service.sh#L322-L344

test/cases/api/common/composer-db.sh
```diff
@@ -0,0 +1,148 @@
...
+                echo "API output: $OUTPUT"
+                return 1
+                ;;
+        esac
+
+        sleep 30
+    done
+
+    export UPLOAD_STATUS UPLOAD_TYPE UPLOAD_OPTIONS UPLOAD_STATUSES
+}
```
thozza: Ditto, this should not be here. Ideally also deduplicate https://github.com/osbuild/osbuild-composer/blob/aaf4eb9999df0c8af2a7d51f25061d20245c7541/test/cases/api-bootc-service.sh#L346-L400

test/cases/api-unit.sh
```diff
@@ -0,0 +1,369 @@
...
+    -rsaPem /etc/osbuild-composer/client-key.pem \
+    -cert /etc/osbuild-composer/composer-crt.pem \
+    -key /etc/osbuild-composer/composer-key.pem &
+KILL_PIDS+=("$!")
+
+sudo /usr/libexec/osbuild-composer-test/osbuild-mock-openid-provider \
+    -a localhost:8081 \
+    -rsaPubPem /etc/osbuild-composer/client-crt.pem \
+    -rsaPem /etc/osbuild-composer/client-key.pem &
+KILL_PIDS+=("$!")
```
thozza: Please use https://github.com/osbuild/osbuild-composer/blob/main/tools/run-mock-auth-servers.sh instead

test/cases/api-unit.sh
```diff
@@ -0,0 +1,369 @@
...
+
+if [ "$RETRY_PASSED" = true ]; then
+    record_result "Job retry (worker crash)" "PASS"
+else
+    record_result "Job retry (worker crash)" "FAIL"
+fi
+
+# ============================================================================
+# Section 3: OAuth2/JWT
+# ============================================================================
```
thozza, line 212: We should not be really testing anything else than OAuth2/JWT AFAICT, because the Cloud API in the Service is not used with anything else.

Also, it would be good if we could just use https://github.com/osbuild/osbuild-composer/blob/main/tools/provision.sh everywhere and do the JWT setup properly there for all tests.

thozza (dismissed):
Thanks.

I think that this makes a lot of sense, but the extracted common helpers are all mixed up in a single file that claims to be a DB helper, while there are some that are not. There is a room to deduplicate some more code with these common helpers and some code in the added api-unit.sh can be replaced by existing helpers.

test/cases/api-unit.sh
ochosi: Thanks for the hint - that makes sense. Done.

test/cases/api-unit.sh
ochosi: Neat!  Shoud've seen this. Done.

test/cases/api/common/composer-db.sh
ochosi: True! I'll separate it out

ochosi:
> Thanks.
> 
> I think that this makes a lot of sense, but the extracted common helpers are all mixed up in a single file that claims to be a DB helper, while there are some that are not. There is a room to deduplicate some more code with these common helpers and some code in the added api-unit.sh can be replaced by existing helpers.

All the feedback is now addressed - thanks for the review!

brlane-rht (dismissed):
Looks good!

test/cases/api-unit.sh
```diff
@@ -0,0 +1,305 @@
...
+greenprint "====== Section 3: OAuth2/JWT ======"
+
+OAUTH_PASSED=true
+
+greenprint "Reconfiguring composer for JWT authentication"
+sudo cp -a /usr/share/tests/osbuild-composer/composer/osbuild-composer-jwt.toml \
+    /etc/osbuild-composer/osbuild-composer.toml
+sudo tee -a /etc/osbuild-composer/osbuild-composer.toml > /dev/null <<EOF
```
supakeen: This appends to the file which will go 'OK' because `[worker]` is the last section but it's not very explicit and feels fragile? In the previous code we wrote the whole file so the ambiguity wasn't there. It'll work so not a blocker but I'm not super fond of it either.

supakeen (dismissed):
Thanks, are the similar calls in `api-bootc.sh`  and `regression-old-worker-new-composer.sh` also going to be using the `common.sh` implementation in the future?

ochosi:
> Thanks, are the similar calls in `api-bootc.sh` and `regression-old-worker-new-composer.sh` also going to be using the `common.sh` implementation in the future?

@supakeen  I was considering to deduplicate those as well. I'm ok with doing so in this PR or a separate follow-up, whichever you prefer!

test/cases/api-unit.sh
ochosi: Thanks for the review - that makes sense. It's fixed now and there's no more append, but the file instead gets created as a whole.

supakeen (approved):

achilleas-k (approved):

PR #5135, opened 2026-05-13
.tekton/osbuild-composer: add COMMIT as a build argument

croissanne:
In the build-container step, add COMMIT as a build argument. This argument is used to set the build version.

This fixes the 'devel' version in all konflux-built containers.


achilleas-k (approved):
TY

lzap (approved):

PR #5136, opened 2026-05-13
test/cases/api: fix race condition in worker registration

croissanne:
This fixes an issue where the worker would register with osbuild-composer before the latter was restarted to switch from jsondb to the psql container as a database. When the worker would then try to pick up a job, it would fail to do so because the worker was not known to composer, because it was registered against jsondb.

Restarting the worker together with composer solves this, as it forces the worker to register again, and this time against the postgres db.

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -144,7 +144,8 @@ pg_ssl_mode = "disable"
 pg_max_conns = 10
 EOF
 
-sudo systemctl restart osbuild-composer
+# Restart the worker as well to make sure it is registered against the psql DB
+sudo systemctl restart osbuild-composer osbuild-remote-worker@localhost:8700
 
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):

achilleas-k:
Weird that we never had this problem before.

TYSM!

croissanne:
> Weird that we never had this problem before.
> 
> TYSM!

it should actually be fine, like it should just restart the worker and dequeue again. Maybe something changed where one of the services isn't restarted automatically? idk

achilleas-k (dismissed):

brlane-rht (dismissed):
Ack!

lzap (dismissed):

achilleas-k (approved):

lzap (approved):

achilleas-k:
Only failure is an openshift virt test (see #5139).
Admin-merging.

PR #5137, opened 2026-05-13
gitlabci: Drop the Container Embedding test

ochosi:
This test is actually for a very specific customization and is therefore covered in the images library. It conceptually predates the split between composer and the image definitions and should therefore now go.


achilleas-k (approved):

lzap (approved):

achilleas-k:
The only failure is openshift virt, which is now disabled in `main` (#5139).

PR #5138, opened 2026-05-13
test/repositories: fix 10.1 repo configs

achilleas-k:
In 1ac98e8b00f9a9875204a45bc7fbe415e111e663 I changed the rpmrepo URLs to CDN for GA RHEL versions, but I also accidentally enabled RHSM for repos that were kept as rpmrepo (GCE and SAPHANA).

We never use the CDN for these and always keep them on rpmrepo.


brlane-rht (dismissed):
Looks good!

lzap (dismissed):
I really think we should get rid of those million copies of PGP keys which make these JSONs totally unreadable.

https://github.com/osbuild/images/pull/2098

achilleas-k:
> I really think we should get rid of those million copies of PGP keys which make these JSONs totally unreadable.
> 
> [osbuild/images#2098](https://github.com/osbuild/images/pull/2098)

You're gonna love this: I think I need to update the 10.1 GPG keys.

lzap (approved):
Really do not understand your pushback in the PR I made, hardcoding the most common GPG keys in the codebase and having some reference string and a custom JSON loader would do it... :-D

achilleas-k:
> Really do not understand your pushback in the PR I made, hardcoding the most common GPG keys in the codebase and having some reference string and a custom JSON loader would do it... :-D

I'm not entirely against it, but I think it's not a pure improvement.  Also when it was opened originally, there were a lot of moving parts in related places that made me afraid it would cause issues with external things that assume the structure we have and we weren't in a place where we could deal with all the side effects at the time.

There's a _simplicity_ of sorts to having a self contained repo config that you can move around, read and write without needing any special rules (like a custom loader).  There's documentation in places that tells users they can copy the repo configs to override them that wouldn't work if the configs relied on paths that are no longer valid.  When the PR was opened, we didn't have time (I personally didn't have the capacity) to think about and resolve all of that.

There's also other related annoyances.  It's annoying that we sometimes use `gpgkey (string)` and `gpgkeys (array)` in different places (osbuild-composer vs images).  It's also a bit ugly that these configs kinda have the structure of `.repo` files but not exactly (ignoring for as second the json vs conf format difference).

I get the annoyance with the huge strings.  You're definitely right that it's stupid, hard to read, hard to diff and review.  But it's a core piece of config that spans our whole stack (images, osbuild-composer, image-builder-cli, image-builder-crc, rpmrepo and its automations) and if we're going to make any changes to it, even if it's an optional or backwards compatible change that doesn't require changing everywhere all at once, I like to solve such problems by taking the time to consider all the related problems we might want to solve, designing a change, and doing it correctly wherever it's needed, rather than reactively patching issues as they crop up.

ochosi (approved):

PR #5144, opened 2026-05-19
gitlabci: Reduce API PR tests to 6 targets

ochosi:
# Reduce infra load on pull requests
In order to reduce the load on our PR test infrastructure this change drops several image and RPM builds from the matrix, **reducing the job count by 52 jobs**.
Note that these tests serve as end-to-end integration tests for composer, not for image building per se, which is fully covered in `osbuild/images`. Also, this PR only affects the testing we do on pull requests, there are no changes to nightly/GA pipelines.

## Reduced build matrix
The changes proposed reduce the job count from 53 to 6 targets.
The remaining image builds are:
- 4 public clouds (GC, Azure, AWS, OCI)
- qcow2
- installer/iso

The qcow2 image-info had to be disabled for the package customization. For some reason, this was also commented out in the VMWare tests, so this was never tested and as it seems, the test doesn't work. (I assume the postgresql package installed fine, there are no errors, but the image-info check on the built image fails.)

## Fewer RPM builds
Also, 5 RPM jobs are being dropped, bringing the previous total of 20 down to 15. While these jobs run in parallel, they are sometimes flaky due to brittle infrastructure and tend to block PR pipelines early. So reducing the number of parallel RPM builds should help.
They could probably be even further reduced by using a different runner for the packer job: the RHEL 10.1 GA aarch64 RPM is built only for this part of the job.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -521,29 +521,32 @@ cross-distro.sh:
 
 .API_TESTS: &API_TESTS
   IMAGE_TYPE:
+    - guest-image
     - aws
     - azure
-    - edge-commit
     - gcp
-    - vsphere
-    - edge-commit generic.s3
-    - edge-container
     - oci
+    - image-installer
 
 API:
   stage: test
   extends: .terraform
   rules:
-    - !reference [.upstream_and_ga_rules_all, rules]
     - !reference [.ga_rules_all, rules]
-    # note: cloud API is not supported for on-prem installations so
-    # don't run this test case for nightly trees
   script:
     - schutzbot/deploy.sh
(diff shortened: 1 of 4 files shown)


achilleas-k:
I had https://github.com/osbuild/osbuild-composer/pull/5172 prepped for dropping RHEL 8 testing but we can keep this one instead.

ochosi:
> I had #5172 prepped for dropping RHEL 8 testing but we can keep this one instead.

They may not massively conflict. I'd merge them all one after another...

brlane-rht:
Last commit says `WIP` at the bottom of it. Looks ok other than that.

ondrejbudai (approved):
Lgtm!

ondrejbudai (approved):

PR #5145, opened 2026-05-21
[9.8_10.2].gitlab-ci.yml: only run tests relevant for 9.8 and 10.2 ga (RHEL-179248)

croissanne:
ssia


thozza (changes requested):
EDIT: Sorry, I should have look at the target branch :facepalm:

thozza (dismissed):

ondrejbudai:
The base tests are failing on quite a cryptic error:

```
Command failed (exit status 1): ERROR: ModulesError: msg: DNF error occurred: InternalError: osbuild-depsolve-dnf output was empty: depsolve failed with exit code -1
```

I'm pretty sure it's because the depsolve process doesn't actually exit but gets killed. I opened a PR in images to fix this error message: https://github.com/osbuild/images/pull/2365

I think it gets OOM killed because we are now listing all packages from the CDN repos that are much larger than our snapshots (because they contain all package versions). My initial idea was to use a larger runner but the base tests run also on aarch64 for which we currently don't have large runners. We have a precedence of using swap for tests with big memory spikes though, so that's why I'm trying. ;)

We should at least consider backporting https://github.com/osbuild/osbuild/pull/2456 to 9.8/10.2, I wonder how many OOMs users might get against the CDN repos.

ondrejbudai:
So, adding a 4GiB swap fixed 10.2 but not 9.6, it still OOMed , so I yolo'd it and tried https://github.com/osbuild/osbuild/pull/2456. This once again fixed only 10.2, 9.6 started to failing on a different depsolve error.

I then switched my focus on remediating some CVEs. Unfortunately, for some deps we need Go 1.25. Fortunately, the target distributions already have Go 1.26 so that's not an issue.

achilleas-k (approved):
yay

PR #5146, opened 2026-05-21
Fix CI by updating everything

achilleas-k:
- Gitlab runners
- osbuild commit
- Test repo configs


thozza:
```
- nothing provides selinux-policy-targeted >= 42.1.20 needed by osbuild-selinux-182-1.20190819gita5d1c15.el10.x86_64 from osbuild
```

The runner needs to use newer repositories for package installation.

croissanne:
I think we might need to switch some tests over from 10.1 to 10.2; 10.1 using the GA repositories cannot support osbuild 182 it seems? So we could just try 10.2? Not sure :/ 

I can't see how we can get 'newer' repositories for 10.1 ga either


edit: of course the executor isn't subscribed, it's fine now

ochosi (approved):

croissanne (approved):

lzap (approved):

PR #5152, opened 2026-05-25
test: make color print functions print all args

ondrejbudai:
base_tests.sh currently doesn't print a test result report:

```
[2026-05-22T17:29:39+00:00] 😃 Passed tests:
[2026-05-22T17:29:39+00:00] ☹ Failed tests:
```

This is because 4c7b3dd25 changed the report to use the colored print functions instead of plain echo. However, the colored print functions print only the first arg instead of all of them like echo does. Thus, the reporting broke because it passes multiple args to the print functions.

Fix this by changing the color printing functions to print all args, roughly matching echo's behavior.

### DIFF
--- a/test/cases/shared_lib.sh
+++ b/test/cases/shared_lib.sh
@@ -35,15 +35,15 @@ function get_build_info() {
 
 # Colorful timestamped output.
 function greenprint {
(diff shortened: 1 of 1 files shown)


thozza (approved):
Thanks 😄

lzap (approved):

ochosi (approved):

PR #5153, opened 2026-05-25
bump Go to 1.25

ondrejbudai:
RHEL 9 and 10 are on 1.26, the oldest supported Fedora (42) is on 1.25 so we are safe to bump.


pkg/splunk_logger/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/osbuild/osbuild-composer/pkg/splunk_logger
 
 go 1.23
 
-toolchain go1.24.12
+toolchain go1.25.10
```
lzap: You are bumping the toolchain, but not the actual Go version, see line 3.

lzap (changes requested):
Other than that, ok.

pkg/splunk_logger/go.mod
ondrejbudai: Thanks!

brlane-rht (approved):
Looks good

brlane-rht:
Not sure what's going on here, but it seems like `gitlab-runner artifacts-uploader` dropped support for `-timeout` -- except I can't figure out where gitlab-runner is called from :/

lzap:
Broken CI: https://github.com/osbuild/osbuild-composer/pull/5146

lzap (approved):
Rebased let's go.

PR #5157, opened 2026-05-26
Ensure the host executor logs failures in the journal (HMS-10713)

croissanne:
This addresses the issue of missing error messages/logs for the host executor. If osbuild returns a valid result, it is interpreted by the worker, which then generates proper job errors, journal messages and compose logs.

The ec2 executor can be addressed separately, it currently always fails with a (go) error (even if osbuild itself returns a valid result), due to the executor writing invalid json and the monitor failing to parse it (https://github.com/osbuild/images/blob/main/pkg/osbuild/monitor.go#L114). It should get a similar treatment.

### DIFF
--- a/internal/osbuildexecutor/runner-impl-host.go
+++ b/internal/osbuildexecutor/runner-impl-host.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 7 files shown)


croissanne:
I'd like to backport this one too to 9.8 and 10.2 as soon as the CI is up and running there

internal/osbuildexecutor/runner-impl-host.go
```diff
@@ -40,14 +40,16 @@ func (he *hostExecutor) RunOSBuild(manifest []byte, logger logrus.FieldLogger, j
 		return nil, fmt.Errorf("unable to construct osbuild result: %w", err)
 	}
 
-	if err := cmd.Wait(); err != nil {
-		return nil, fmt.Errorf("error running osbuild: %w", err)
+	// In case the exit code is 1 or 2, osbuild ran and returns a (failed) result
+	if err, exitcode := cmd.Wait(), cmd.ProcessState.ExitCode(); err != nil && exitcode != 1 && exitcode != 2 {
+		return nil, fmt.Errorf("osbuild failed: %w, %s", err, stdoutBuffer.String())
```
thozza: This is OK for now, but it may make sense to make this implementation and https://github.com/osbuild/images/blob/d2988d6612a442f005fd09f53fdc104c4c3b81f3/pkg/osbuild/osbuild-exec.go#L160-L178 consistent.

thozza (dismissed):
Thanks!

internal/osbuildexecutor/runner-impl-host.go
croissanne: I updated it to be more similar, with a small change in case jsonoutput is false and no error was returned. Added some tests for that case, though it's kinda just theoretical because all calls in composer are done with jsonoutput set to true.

lzap (dismissed):
Typo in git commit 3/4 "istinction"

achilleas-k (dismissed):
Just reviewed the backport version of this, so skimmed here too.  LGTM.

croissanne:
> Just reviewed the backport version of this, so skimmed here too. LGTM.

i ducked up the test slightly, here and there. amended it now

croissanne:
ok I think it's finally ready ™

edit: ok 1 more push

achilleas-k (approved):

thozza (approved):

PR #5158, opened 2026-05-26
ci_details: Fix repolist

brlane-rht:
yum is now dnf and the repolist command does not have a '-v' option. Use 'dnf repolist' which will list the enabled repos.


achilleas-k (approved):
`yum` is never going to die, is it.

lzap (approved):
Easier to type, as simple as that.

PR #5163, opened 2026-05-27
cloudapi/v2: Fix no partition table handling and add aliases to filtering (HMS-10672)

tkoscieln:
I've messed this up back in February in the original change [28d3dcb](https://github.com/osbuild/osbuild-composer/commit/28d3dcbde069cd7a88453f09be827d49307bed64). The code was doing the exact opposite of what the comment was suggesting it's supposed to do. When collecting distribution data, there was a bug that ommits image types without partition tables. As a result, the entire image type was dropped (such as WSL) from the distro metadata. Fix this by inverting the logic of error handling.

Also allows for aliases in filtering image types by name. Implementing it here will allow us not use any "translator" in frontend that uses the aliases when picking targets.


brlane-rht (approved):
Looks good!

lzap (approved):

PR #5164, opened 2026-05-27
[rhel-9.8_10.2] Ensure the host executor logs failures in the journal  (RHEL-179244)

croissanne:
This addresses the issue of missing error messages/logs for the host executor. If osbuild returns a valid result, it is interpreted by the worker, which then generates proper job errors, journal messages and compose logs.

The ec2 executor can be addressed separately, it currently always fails with a (go) error (even if osbuild itself returns a valid result), due to the executor writing invalid json and the monitor failing to parse it (https://github.com/osbuild/images/blob/main/pkg/osbuild/monitor.go#L114). It should get a similar treatment.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -240,6 +240,8 @@ func (impl *OSBuildJobImpl) getOCI(tcp oci.ClientParams) (oci.Client, error) {
 
 func validateResult(result *worker.OSBuildJobResult, jobID string) {
 	logWithId := logrus.WithField("jobId", jobID)
+
+	// in case of failures JobError is expected to be set
 	if result.JobError != nil {
 		logWithId.Errorf("osbuild job failed: %s", result.JobError.Reason)
 		if result.JobError.Details != nil {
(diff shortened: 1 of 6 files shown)


cmd/osbuild-worker/jobimpl-osbuild.go
```diff
@@ -240,24 +240,24 @@ func (impl *OSBuildJobImpl) getOCI(tcp oci.ClientParams) (oci.Client, error) {
...
+	// bug in case JobError is nil and the job is marked as failed
+	if result.OSBuildOutput == nil || !result.OSBuildOutput.Success || !result.Success {
+		reason := "osbuild job failed without errors, this is a bug"
+		logWithId.Error(reason)
 		result.JobError = clienterrors.New(clienterrors.ErrorBuildJob, reason, nil)
 		return
-	} else {
-		logWithId.Infof("osbuild job succeeded")
 	}
-	result.Success = true
```
achilleas-k, line 260: Is setting this not necessary?

achilleas-k (dismissed):
Neat.  Looking forward to having better error logs <3

cmd/osbuild-worker/jobimpl-osbuild.go
croissanne: No, we always run osbuild with `--json` and thus it will always be set. Rather if the result isn't successful and we get no error or anything, it's a bug (which is now reflected in the above condition `if result.OSBuildOutput == nil || !result.OSBuildOutput.Success || !result.Success {`)

croissanne: OH wait this is the top-level result, fuck we might need to. hang on good catch

croissanne: oh wait, it's always set at the end
```

	targetErrors := osbuildJobResult.TargetErrors()
	if len(targetErrors) != 0 {
		osbuildJobResult.JobError = clienterrors.New(clienterrors.ErrorTargetError, "at least one target failed", targetErrors)
	} else {
		osbuildJobResult.Success = true
		osbuildJobResult.UploadStatus = "success"
	}
```

croissanne: So in this case there is no JobError on the JobResult AND osbuild itself failed or didn't run or the overall job wasn't marked as successful (meaning it returned early or there were target errors).
So this would be a bug we shouldn't hide.

achilleas-k: Thanks!  Sorry for the confusion :)

achilleas-k (approved):

thozza (approved):

PR #5170, opened 2026-05-29
packer: Bump AWS instance type to c6*.xlarge

ochosi:
We currently see builds slowing down and then stalling until they time out, possibly due to EBS IOPS burst credit exhaustion or memory pressure.


lzap (dismissed):

lzap:
Observation: we tend to not use any swap and it can help quite a bit with helping to survive memory pressure. A script that creates 1GB swap file and mounts it (or use ephemeral storage when available) could help so the system can swap unused pages.

achilleas-k (dismissed):

lzap (dismissed):

ochosi:
Closed in favor of https://github.com/osbuild/osbuild-composer/pull/5161 since it was a regression in the packer aws module.

PR #5174, opened 2026-06-06
test: deduplicate sendCompose/waitForState in bootc and regression tests

ochosi:
As promised in #5132 here's a follow-up to deduplicate some of the code that is shared across some of our shellscript tests.

Both `api-bootc.sh` and `regression-old-worker-new-composer.sh` carried local copies of `sendCompose()` and `waitForState()` that duplicated the shared implementations in `api/common/common.sh`.

Make the API base URL configurable via `COMPOSER_API_BASE` (defaulting to https://localhost) so the regression test can point at its container's :8080 port, then migrate both tests to source `common.sh` and drop their local definitions.

Additionally migrate `api-bootc.sh` to use `composer-db.sh`'s `setup_db` / `teardown_db` and `common.sh`'s `write_tls_composer_config`, eliminating further inline duplication of DB lifecycle and TOML config boilerplate.

### DIFF
--- a/test/cases/api-bootc.sh
+++ b/test/cases/api-bootc.sh
@@ -5,18 +5,10 @@ set -euo pipefail
 source /usr/libexec/osbuild-composer-test/set-env-variables.sh
 source /usr/libexec/tests/osbuild-composer/shared_lib.sh
 source /usr/libexec/tests/osbuild-composer/api/common/bootc.sh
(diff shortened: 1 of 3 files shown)


supakeen (approved):
Thank you; this is a nice diff!

brlane-rht (approved):
Looks good!

PR #5178, opened 2026-06-09
Consolidate generic S3 tests into single efficient job

thozza:
Merge three separate generic S3 integration test CI jobs (HTTP, HTTPS secure, HTTPS insecure) into a single self-contained script that runs all variants sequentially. This eliminates redundant provisioning and image builds across 12 CI job slots, leveraging osbuild caching so only the first build is full and subsequent ones are near-instant. Additionally, switch to plain AWS runners (no nested virt) since the test no longer boots VMs, and restrict the runner matrix to platforms where on-premise Weldr actually ships.

## Architectural Changes

Replace the multi-script layered approach (thin `test/cases/` wrappers calling shared `tools/` scripts calling `s3_test.sh` calling `libvirt_test.sh`) with a single flat script organized as functions. Each test variant (HTTP, HTTPS secure, HTTPS insecure) is a self-contained function that manages its own MinIO lifecycle. Verification switches from booting a VM via libvirt to a lightweight `qemu-img check`, since the goal is validating the upload landed correctly in S3, not testing VM boot behavior.

## Key Changes

- Consolidate `generic_s3_http.sh`, `generic_s3_https_secure.sh`, and `generic_s3_https_insecure.sh` into a single `generic_s3.sh` with function-per-variant structure
- Replace libvirt boot verification with `qemu-img info` + `qemu-img check` on the downloaded artifact
- Switch generic_s3 job to plain AWS runners (no nested virtualization required)
- Reduce runner matrix to CentOS Stream 9/10, Fedora 44, and RHEL 9.9/10.3 nightly only — drop RHEL GA since generic S3 is not service-supported
- Remove now-unused `tools/generic_s3_test.sh` and `tools/generic_s3_https_test.sh`
- Inline `aws_s3.sh` job definition since the shared `.generic_s3` template is removed

## Breaking Changes

## Testing

- All three S3 upload code paths (HTTP, HTTPS with CA verification, HTTPS with skip verification) remain exercised in a single job
- Image validity is confirmed via `qemu-img check` instead of full VM boot, keeping verification meaningful while cutting runtime significantly
- Blueprint is pushed once and osbuild caching ensures builds 2 and 3 hit the cache (same blueprint + same qcow2 target)


achilleas-k (dismissed):

lzap (approved):

test/cases/generic_s3.sh
```diff
@@ -17,101 +31,296 @@ if type -p podman 2>/dev/null >&2; then
...
+sudo "${CONTAINER_RUNTIME}" pull "${CONTAINER_MINIO_SERVER}"
+sudo "${CONTAINER_RUNTIME}" pull "${CONTAINER_IMAGE_CLOUD_TOOLS}"
 
-# Set up temporary files.
-MINIO_CONFIG_DIR=${TEMPDIR}/minio-config
-MINIO_PROVIDER_CONFIG=${TEMPDIR}/minio.toml
+# Generate TLS certs for HTTPS variants (reused across both HTTPS tests)
+# Certs must be outside /tmp because osbuild-worker has PrivateTmp=true
+pushd "${TEMPDIR}"
+curl -L -o certgen "https://github.com/minio/certgen/releases/download/${CERTGEN_VERSION}/certgen-linux-amd64"
```
brlane-rht, line 62: Pulling a script and injecting it in the middle of CI code doesn't seem safe these days.

test/cases/generic_s3.sh
```diff
@@ -1,14 +1,28 @@
...
 
 source /usr/libexec/osbuild-composer-test/set-env-variables.sh
+source /usr/libexec/tests/osbuild-composer/shared_lib.sh
 
-set -euo pipefail
-
-# Container images for MinIO Server and Client
-CONTAINER_MINIO_CLIENT="quay.io/minio/mc:latest"
+# Container images
 CONTAINER_MINIO_SERVER="quay.io/minio/minio:latest"
```
brlane-rht, line 8: Is this a Red Hat maintained project?

brlane-rht (changes requested):
I'm unsure about the safety of the minio project (probably ok, assuming quay.io is managing and scanning it).
I'm opposed to pulling scripts from github projects and running them. There should be a safer way to generate the fake certs (eg. use openssl directly).

thozza:
Thanks for the review. I don't see myself spending more time on this in the near future...

lzap:
I did not like downloading and executing binaries either, but this was already in the codebase. This PR just made the behavior more stable. :-(

Edit: Ah MiniIO was there, certgen was not. Hmm.

thozza:
> I did not like downloading and executing binaries either, but this was already in the codebase. This PR just made the behavior more stable. :-(
> 
> Edit: Ah MiniIO was there, certgen was not. Hmm.

All of it was there before, including the certgen binary, see https://github.com/osbuild/osbuild-composer/pull/5178/changes#diff-06aa8a16666ecd1f004246eefbd8dcbfa8b9ec6a890ef5d6991af54450257039L25 (check the deleted `tools/generic_s3_https_test.sh` file)

lzap:
Yeah I swear I saw it, I was checking it during my review but I could not find it the second time, thanks. Well, it was a nice improvement :(

PR #5179, opened 2026-06-12
spec: drop the tests subpackage from EL11

ondrejbudai:
We don't intend to ship osbuild-composer in EL11. The first step was to remove weldr-client but the tests subpackage is still pulling it as a dep. Let's remove the tests package from osbuild-composer in EL11 to start chipping away the unwanted components from it.

See https://github.com/fedora-eln/content-resolver-input/pull/1606

### DIFF
--- a/osbuild-composer.spec
+++ b/osbuild-composer.spec
@@ -109,7 +109,7 @@ export LDFLAGS="${LDFLAGS} -X 'github.com/osbuild/osbuild-composer/internal/comm
 
 make man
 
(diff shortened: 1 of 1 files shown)


supakeen (approved):

lzap (approved):

avitova:
Rebasing, so that we see if this is still green. But it should since this only removes tests. So let's mege this, if we did not change our mind...?

brlane-rht (approved):
Ack!

PR #5183, opened 2026-06-16
Cloud API: expose build_reference for bootc (HMS-10836)

thozza:
Add an optional `build_reference` field to the Cloud API's `Bootc` schema, enabling API consumers to specify a separate build container image. The internal plumbing for separate build containers already existed (`BuildInfoIdx`, `SetBuildContainer`) but was unreachable from the public API. This closes that gap while preserving full backward compatibility when the field is omitted.

This is a preparation, so that the `koji-osbuild` plugin can gain bootc support and keep the CLI and Hub plugin schema consistent with the `koji-image-builder` version (which exposes also the build ref). Also this may be needed eventually, so let's be on the safe side and wire up the functionality and expose it via the API.

## Architectural Changes

The design leverages the existing `BootcInfoResolveModeBuild` resolve mode and `BuildInfoIdx` mechanism already present in the worker job graph. When `build_reference` is set, a second lightweight resolve spec (collecting only `Imgref`, `ImageID`, and `Arch`) is appended to the `BootcInfoResolve` job, and `BootcPreManifestJob.BuildInfoIdx` is set to point to it. No new error codes, validation logic, or downstream worker changes are needed.

## Key Changes

- Add `build_reference` optional field to `Bootc` in `openapi.v2.yml` and regenerate Go types
- Wire `build_reference` into the compose job graph: append a build-mode resolve spec and set `BuildInfoIdx` on the pre-manifest job

## Breaking Changes

## Testing

- Add a `with_build_reference` test case to `TestComposeBootc` verifying the job graph gets 2 resolve specs (base full + build build-mode) and `BuildInfoIdx == 1`
- Add a `separate_build_container` case to `TestHandleBootcPreManifest_HappyPath` confirming `SetBuildContainer` flows through to a third pipeline spec in the manifest source output


brlane-rht:
Did you write these commit messages or did Claude? They're reading more like a description of what code is changing and less like the normal style of describing why the change was made. Calling out all the function and variable names in the message is unnecessary when the diffs are so short.

lzap (approved):
I usually do not read the implementation details in the PR description as these are present in the individual commits. Only when something does not make sense, I return to it. But yeah, these can be verbose. I do not mind, just not read them?

This is good, the "index" thing is something I do not like in general but this was ACKed already and I do not see any better reasonable solution.

thozza:
> Did you write these commit messages or did Claude? They're reading more like a description of what code is changing and less like the normal style of describing why the change was made. Calling out all the function and variable names in the message is unnecessary when the diffs are so short.

Yes, I did use AI to compose these commit messages, but I've also read them and OK with them. Commit messages what is done and why, with the exception of the last two commits that are modifying only unit tests and there is not much WHY to explain beyond "extend tests to cover the changes". I'll try to keep them more concise.

achilleas-k (approved):

PR #5185, opened 2026-06-19
Update import paths for image-builder [HMS-10853]

achilleas-k:
Extra changes needed:
- cloud API unit tests were using old Fedora versions that are no longer supported.  Updated to Fedora 44.
- Small change in result handling in jobimpl-osbuild.go in uploadToS3() due to change in AWS library.

PR #5192, opened 2026-06-24
test/data/repositories/rhel-9.9: disable gpg check (HMS-10929)

croissanne:
Crypto-policies broke pqrpm by not allowing ML-DSA signatures. Temporarily disable gpg checks until this situation is resolved.


brlane-rht (dismissed):
Ack. Argh. Ugh :)

achilleas-k (dismissed):
What Brian said.

achilleas-k:
Will just disabling the checks solve the issue?  I think (might be wrong) that having gpg keys in the repo configs will still add the key options in the rpm stage and it will try to import them, it just wont use them to verify the packages.  So the key import will still fail.  That is **if** I'm right.

achilleas-k:
Rebasing for #5185.

achilleas-k:
> Will just disabling the checks solve the issue? I think (might be wrong) that having gpg keys in the repo configs will still add the key options in the rpm stage and it will try to import them, it just wont use them to verify the packages. So the key import will still fail. That is **if** I'm right.

Oh, I was wrong: https://github.com/osbuild/image-builder/pull/2203

Nice.

achilleas-k:
The upgrade test is still failing: https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/jobs/15018857691

It looks like it sets up repos after upgrading: https://github.com/osbuild/osbuild-composer/blob/5926c2af4bc2c20770c75de9744b8446289797d6/test/data/rhel-upgrade/upgrade_verify.sh#L76-L99

ochosi (approved):
LGTM as a temporary change

avitova (approved):

PR #5206, opened 2026-07-22
executor: return failed osbuild result form prepare sources (RHEL-212871)

croissanne:
In case of an `exec.ExitError`, RunOSBuild does not return a go error, but a valid `osbuild.Result` with `success` set to false.

By returning this valid result from the ec2 executor in case source preparation fails, the relevant error is surfaced.

### DIFF
--- a/internal/osbuildexecutor/runner-impl-aws-ec2.go
+++ b/internal/osbuildexecutor/runner-impl-aws-ec2.go
@@ -31,15 +31,14 @@ type awsEC2Executor struct {
 	tmpDir     string
 }
 
-func prepareSources(manifest []byte, logger logrus.FieldLogger, opts *osbuild.OSBuildOptions) error {
+func prepareSources(manifest []byte, logger logrus.FieldLogger, opts *osbuild.OSBuildOptions) (*osbuild.Result, error) {
 	hostExecutor := NewHostExecutor()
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):

internal/osbuildexecutor/runner-impl-aws-ec2.go
```diff
@@ -265,10 +264,13 @@ func extractOutputArchive(outputDirectory, outputTar string) error {
 }
 
 func (ec2e *awsEC2Executor) RunOSBuild(manifest []byte, logger logrus.FieldLogger, job worker.Job, opts *osbuild.OSBuildOptions) (*osbuild.Result, error) {
-	err := prepareSources(manifest, logger, opts)
+	result, err := prepareSources(manifest, logger, opts)
```
avitova: result is defined twice, I think this would bring up few errors

avitova (dismissed):
nice nice

lzap (dismissed):

croissanne:
Needs https://github.com/osbuild/osbuild-composer/pull/5207 first

brlane-rht (approved):
Seems like a great idea!

avitova (approved):

PR #5208, opened 2026-07-23
cloudapi/v2: support subscriptions for image mode composes (HMS-11080)

lucasgarfield:
Image mode composes submitted through the cloud API silently dropped
any subscription customization: the `subscription` block in the request was
parsed but never reached manifest generation, so the resulting image would
boot unregistered. Package mode composes have supported subscriptions for a
long time. This wires the same support into the image mode path.


brlane-rht (dismissed):
Nice!

avitova:
Sorry, I had to merge a fix for those koji failures that are happening on Schutzbot. Could you rebase? I know there are conflicts:(

avitova (approved):
Very nice, good catch!

brlane-rht (approved):
Ack!

PR #5216, opened 2026-08-04
composer: fall back to /etc/yum.repos.d/ when no repo configs are found

kayscheerer:
When no osbuild-specific JSON/YAML repository config files exist under
/etc/osbuild-composer/repositories or /usr/share/osbuild-composer/repositories,
the composer previously returned a NoReposLoadedError (fatal unless
ignore_missing_repos=true), making it impossible to use on systems whose
only repository definitions live in /etc/yum.repos.d/ — for example,
Satellite-managed RHEL machines using certificate-based subscriptions.

In my opinion, the repositories already configured on the system should be the defaults.
My VMs are all managed by a RedHat satellite and cannot reach this default "cdn..." repositories. However, you can't even remove them from the GUI, which is quite painful. I tried X different hacks to get rid of those "cdn" repos because the image building will not work if there is any unreachable repository in the list (or atleast the short error message that it cannot find the metadata in those repos make me believe that).
I found in the documentation of [weldr](https://github.com/osbuild/weldr-client#package-sources) that it should pick it up from the already configured repositories. I would prefer that generally. 

Two remarks: Instead of just complaining about a feature you don't have I wanted to propose a possible code solution up front. I haven't worked with Go so far so I would really value your input if the code is not up to your standards. I will check again to fix the checks if they break. 

Edit:
image
image
My current issue in some screenshots. These default repositories seem to come from `/usr/share/osbuild-composer/repositories/`. If I understood correctly, you should add your own repositories to `/etc/osbuild-composer/repositories`. In that case, the current solution is just broken if you can't reach these defaults.
My current solution is to overwrite the default files with ansible. That also requires restarting the systemd service and generally it seems ugly to me. I'm also open to have other solutions that work but so far I couldn't find any.
Part of the ansible code I currently have to maintain for this:
```
- name: Copy Satellite repository definitions for all RHEL 9 variants
  ansible.builtin.copy:
    src: rhel-9.json
    dest: "/etc/osbuild-composer/repositories/{{ item }}"
    owner: root
    group: root
    mode: "0644"
  loop:
    - rhel-9.json
    - rhel-9.0.json
    - rhel-9.2.json
    - rhel-9.4.json
    - rhel-9.6.json
    - rhel-9.7.json
    - rhel-9.8.json
  notify: Restart osbuild-composer

- name: Flush handlers so composer restarts before we verify sources
  ansible.builtin.meta: flush_handlers

- name: Wait for osbuild-composer to be ready to accept requests
  ansible.builtin.command:
    cmd: composer-cli status show
  register: composer_ready
  until: composer_ready.rc == 0
  retries: 15
  delay: 2
  changed_when: false 
...
```

### Suggested code change:

Add repoRegistryFromYumReposD() which reads all *.repo files from
/etc/yum.repos.d/, parses the standard DNF/yum INI format using
gopkg.in/ini.v1 (already a transitive dependency via the rhsm package),
and returns a RepoRegistry keyed to the auto-detected host distro and arch.
The function:
  - skips sections with enabled=0
  - includes sections with no enabled key (DNF default)
  - maps all relevant fields: baseurl, gpgkey, gpgcheck, metalink,
    mirrorlist, sslcacert, sslclientkey, sslclientcert, sslverify,
    metadata_expire
  - correctly handles multiple whitespace-separated values for baseurl
    and gpgkey (backslash line continuation, joined by ini.v1)

The fallback is inserted into the NoReposLoadedError branch in
NewComposer(). IgnoreMissingRepos still applies if both the JSON/YAML
configs and the yum.repos.d/ fallback yield nothing.

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -12,19 +12,25 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	logrus "github.com/sirupsen/logrus"
+	"gopkg.in/ini.v1"
 
 	"github.com/osbuild/osbuild-composer/pkg/jobqueue"
 	"github.com/osbuild/osbuild-composer/pkg/jobqueue/dbjobqueue"
 
+	"github.com/osbuild/image-builder/pkg/arch"
 	"github.com/osbuild/image-builder/pkg/depsolvednf"
+	"github.com/osbuild/image-builder/pkg/distro"
 	"github.com/osbuild/image-builder/pkg/distrofactory"
 	"github.com/osbuild/image-builder/pkg/experimentalflags"
 	"github.com/osbuild/image-builder/pkg/reporegistry"
+	"github.com/osbuild/image-builder/pkg/rpmmd"
 	"github.com/osbuild/osbuild-composer/internal/auth"
 	"github.com/osbuild/osbuild-composer/internal/cloudapi"
 	v2 "github.com/osbuild/osbuild-composer/internal/cloudapi/v2"
@@ -83,12 +89,18 @@ func NewComposer(config *ComposerConfigFile, stateDir, cacheDir string) (*Compos
 	case nil:
 		// fine
 	case *reporegistry.NoReposLoadedError:
-		if !c.config.IgnoreMissingRepos {
-			return nil, fmt.Errorf("error loading repository definitions: %w", err)
+		// No osbuild-specific repo config files found; try to build the registry
+		// from the system's /etc/yum.repos.d/ so that the host's configured
+		// repositories are used automatically.
+		logrus.Info("No repository config files found in configured paths, falling back to /etc/yum.repos.d/")
+		c.repos, err = repoRegistryFromYumReposD("/etc/yum.repos.d")
+		if err != nil {
+			if !c.config.IgnoreMissingRepos {
+				return nil, fmt.Errorf("error loading repository definitions from /etc/yum.repos.d/: %w", err)
+			}
+			logrus.Infof("Failed to load repositories from /etc/yum.repos.d/: %v", err)
+			logrus.Info("ignore_missing_repos enabled: continuing")
 		}
-		// running without repositories is allowed: log message and continue
-		logrus.Info(err.Error())
-		logrus.Info("ignore_missing_repos enabled: continuing")
 	default:
 		return nil, fmt.Errorf("error loading repository definitions: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


supakeen:
To set some expectations: this PR is bound to generate a bunch of discussion and some team members are on vacations since it's Europe summer time so it might take a bit for people to get to reading this.

For me this would be more appealing if it was gated behind a command line argument or configuration option.

---

I do generally like being able to generate repository configs from `.repo` files as a separate feature; perhaps that would be better suited in `github.com/osbuild/image-builder` (the library and `image-builder` executable that's used by `osbuild-composer`) to then use those functions here.

kayscheerer:
> For me this would be more appealing if it was gated behind a command line argument or configuration option.

That would also be fine with me. Would you rather have a boolean e.g. "--mirror-dnf-repos" or on a repo basis like --mirror-dnf-repo="satellite-appstream"  where the name is the ID used for the dnf repository `[satellite-appstream]`?

kayscheerer:
fyi Updated the PR description to explain my problem with some more details

supakeen:
> > For me this would be more appealing if it was gated behind a command line argument or configuration option.
> 
> That would also be fine with me. Would you rather have a boolean e.g. "--mirror-dnf-repos" or on a repo basis like --mirror-dnf-repo="satellite-appstream" where the name is the ID used for the dnf repository `[satellite-appstream]`?

I'd need to think about it a bit, perhaps `--force-repo-host` (as we have multiple `--force-repo`-style options. Alternatively it could just work through `--force-repo-dir` *if* we allow parsing `.repo` files as well as `.yaml/.json`), don't go implementing any of this right now it really needs some thought of what would make sense to do 🙂

achilleas-k:
Not using host repositories was a conscious choice made a long time ago, before my time, but it's one I agree with and support fully.  There's a lot of messy repo configs out there that can cause build failures or affect a build implicitly in ways that are hard to detect.  Using host resources and configurations in general is something we avoid on principle where possible.

Echoing what Simon said above though, supporting repo configs in `.repo` format would be very useful and something I'd like to see.  Using those files over the built-in ones would need to be an explicit choice by the user.  Implicit fallbacks like this are generally a bad idea, in my experience.

kayscheerer:
As I mentioned in the PR description I'm currently overwriting the .json files with Ansible. The UI shows the repositories correctly but I run into follow up issues. 
It would be nice to have a "test" button to test whether a configured repo works or not. 
I think this is actually a bug. 
I opened a different PR with explanations why the current solution breaks for me. If I'm using it wrong or configured it wrong it would be nice to get some help there (then this PR here can also wait for more discussions)
https://github.com/osbuild/osbuild-composer/pull/5217

supakeen:
Not stale, people have been busy; I'll mention this PR today to some people to see if they have time to take a look.

kayscheerer:
Just another note from my side:
I think the current solution you already have implemented works quite well. 
The only thing which is kind of weird today is that you have to force getting rid of the cdn urls. 
If the cdn redhat repos do not work out of the box you have to go down a rabbit hole to find out what exactly you need to do.

croissanne:
For VMs in a satellite group, overwriting the repositories in /etc is required, and it's not possible via cockpit-composer indeed. The solution you're using there with ansible, using repositories overwrites, is kinda all we have with composer at the moment.

I wouldn't be against an explicit `--use-host-repos` option, with the caveat that the build (artifact) would be unsupported (the option itself would be supported). My point is that all our testing happens from the baseos/appstream repos, host repo configs could include all sorts of things.

kayscheerer:
Closing this PR. 
I currently also see the need to specify the repositories for the build explicitly. Using the host repos would actually not work and would make it difficult to specify where an RPM should come from. Generally, I still think it could be easier to remove the default cdn repos. Currently, it needs knowledge about the code base to know which file to overwrite

PR #5218, opened 2026-08-04
worker: preserve container spec ordering after resolution

ondrejbudai:
container.Resolver.Finish() sorts results by digest so it's not guaranteed that it keeps the input order. However, osbuild-composer mistakenly started relying on keeping the order:

For bootc-generic-iso with an iso_payload_reference, the os-tree pipeline carries two containers. If the payload container's digest sorts before the source container's, they end up in reversed positions, causing the payload to be deployed as the OS tree and vice versa.

Fix this by switching to the new-ish Resolve() method of the resolver that does resolving synchronously. This allows us to neatly condense the 2 loops into 1 while making sure that the order is not changed.

Kudos to @lucasgarfield for discovering the bug, and to @supakeen for finding out that the container refs get somewhere swapped.

### DIFF
--- a/cmd/osbuild-worker/jobimpl-container-resolve.go
+++ b/cmd/osbuild-worker/jobimpl-container-resolve.go
@@ -77,10 +77,11 @@ func (impl *ContainerResolveJobImpl) Run(job worker.Job) error {
 }
 
 // resolveContainers resolves container specs grouped by pipeline name.
+// Each container is resolved individually with Resolve() to preserve
+// positional ordering.
 func resolveContainers(arch, authFilePath string, pipelineSpecs map[string][]worker.ContainerSpec) (map[string][]worker.ContainerSpec, error) {
-	// NOTE: container.Resolver.Finish() sorts results by digest, destroying Add() ordering.
-	// Use a separate resolver per pipeline because, positional slicing
-	// across pipelines would not work, due to the sorting by digest.
+	resolver := container.NewResolver(arch)
+	resolver.AuthFilePath = authFilePath
 
 	result := make(map[string][]worker.ContainerSpec, len(pipelineSpecs))
 	for name, specs := range pipelineSpecs {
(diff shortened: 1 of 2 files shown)


achilleas-k (dismissed):
Oh!  I see now.

LGTM

ondrejbudai:
Well, the former approach wasn't gonna fly because the resolver normalizes the `Name` field so we cannot use it to preserve the ordering.

When I started digging deeper, I noticed we now have a handy [`ResolveAll` method](https://github.com/osbuild/image-builder/blob/448fa78f42758641120d6993373c9a103d322ecd/pkg/container/resolver.go#L120) that does exactly what we need. So we just need some shuffling between different struct formats and calling this method to keep the order.

achilleas-k (dismissed):
Very good!

cmd/osbuild-worker/jobimpl-container-resolve.go
```diff
@@ -66,30 +66,31 @@ func (impl *ContainerResolveJobImpl) Run(job worker.Job) error {
...
-			result.JobError = clienterrors.New(clienterrors.ErrorContainerResolution, err.Error(), nil)
-			return fmt.Errorf("Error resolving containers for pipeline %q: %v", name, err)
-		}
+	// Use ResolveAll to ensure that the order of the specs is preserved.
+	// This is now necessary because some pipelines depend on the order
+	// of the specs. E.g. bootc-generic-iso
+	resolved, err := resolver.ResolveAll(sources)
+	if err != nil {
+		result.JobError = clienterrors.New(clienterrors.ErrorContainerResolution, err.Error(), nil)
+		return fmt.Errorf("Error resolving containers: %v", err)
```
supakeen: Why drop the pipeline name here? Perhaps we should inline what `ResolveAll` does instead or is it not that important to know where we failed?

Something like:

```
  result.PipelineSpecs = make(map[string][]worker.ContainerSpec, len(sources))
  for name, ss := range sources {
      pipelineResult := make([]worker.ContainerSpec, len(ss))
      for i, s := range ss {
          res, err := resolver.Resolve(s)
          if err != nil {
              result.JobError = clienterrors.New(clienterrors.ErrorContainerResolution, err.Error(), nil)
              return fmt.Errorf("Error resolving containers for pipeline %q: %v", name, err)
          }
          pipelineResult[i] = worker.ContainerSpecFromVendorSpec(res)
      }
      result.PipelineSpecs[name] = pipelineResult
  }
```

Would keep the pipeline in the error I think (not tested).

supakeen (dismissed):
Looks OK, small comment about losing the pipeline name.

However, it would be really nice to have a test that asserts the invariant introduced here so it doesn't get optimized away again in the future. I don't know how much work that would be if it's a lot then let's not bother and address the bug first.

ondrejbudai:
You are right, @supakeen, `ResolveAll()` looks tempting but it forces us to drop the pipeline name, and it forces us to do a weird double-loop dance for converting the structs.

I decided to use a simple `Resolve` instead that allows us to have just a single loop and much more readable code. Moreover, I extracted the resolution code into a separate method so it can be tested using a mock resolver. Now the ordering requirement for the resolving job is actually expressed in the codebase which I really like. :)

cmd/osbuild-worker/jobimpl-container-resolve.go
ondrejbudai: Done.

supakeen (approved):

achilleas-k (approved):

achilleas-k:
Needs a code owner approval from Tomas, but he's out.  Merging anyway.

PR #5222, opened 2026-08-12
.tekton: disable git-clone symlink check

avitova:
The git-clone task fails on centos-stream-* symlinks. Set enableSymlinkCheck=false on the clone-repository task in all Konflux pipelines.

We are getting
```
time="2026-08-12T00:35:59Z" level=error msg="Broken symlink found: /var/workdir/source/repositories/centos-stream-10.json"
time="2026-08-12T00:35:59Z" level=error msg="Broken symlink found: /var/workdir/source/repositories/centos-stream-9.json"
```
on #5186, trying to fix it, although there will likely be more failures after we merge this.

### DIFF
--- a/.tekton/osbuild-composer-maintenance-pull-request.yaml
+++ b/.tekton/osbuild-composer-maintenance-pull-request.yaml
@@ -143,6 +143,8 @@ spec:
         value: $(params.output-image).git
       - name: ociArtifactExpiresAfter
         value: $(params.image-expires-after)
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):
Thanks.

I think we wont be able to disable it at some point, meaning it'll become mandatory.  I'm more inclined to start including repo configs as separate files again instead of embedding them if we have to.

brlane-rht (approved):
Ack

PR #5223, opened 2026-08-12
weldr: remove repository metadata preload

achilleas-k:
The metadata preload was done so that the repository metadata caches are ready and hot for when the first request comes in.  However, in cases where a lot of distro repositories are configured (such as when running on a newer distro that supports building all older ones, or when testing), preloading metadata from multiple distros can take very long and is mostly useless.

This is the cause of a lot of delays in starting up osbuild-composer in testing.  For example, the first depsolve tests take over 10 minutes to complete in CI because of this.

The original intent of the preloading was not bad.  The idea was that between when osbuild-composer starts and the first compose (or depsolve) is requested, it would have enough time to download the repo metadata and be ready to start building without blocking on the first request by the user.  But I think in some cases this causes more issues than it solves.

HMS-10800

This issue was also noted as a potential issue in disconnected or firewalled environments, which might raise concerns about software needlessly "phoning home" on startup.

RHEL-10640

### DIFF
--- a/cmd/osbuild-composer/composer.go
+++ b/cmd/osbuild-composer/composer.go
@@ -158,9 +158,6 @@ func (c *Composer) InitWeldr(weldrListener net.Listener, distrosImageTypeDenylis
 		return err
 	}
 	c.weldrListener = weldrListener
-
-	// Preload the Metadata for all the supported distros
-	c.weldr.PreloadMetadata()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


brlane-rht (approved):
Ack.

supakeen (approved):

lzap (approved):
Yes, 100x yes.

PR #5227, opened 2026-08-21
test/cases: remove enabled services in test case

avitova:
Test cases for azure and openshift virtualization use enabled services: sshd, cloud-init, cloud-init-llocal, cloud-config and cloud-final. These should be enabled by defaut, so there is no need to add these manually to the blueprint.


achilleas-k (approved):

brlane-rht (approved):
Looks good!

PR #5232, opened 2026-08-24
schutzbot/mockbuild: build CI RPMs with nocheck

achilleas-k:
In CI we build RPMs first for each distro we're going to test and then each job installs those RPMs to run the tests.  Running the tests on rpm build as well is redundant and makes the build take a very long time.

Building RPMs with 'nocheck' will speed up the RPM step of each pipeline, speeding up the whole GitLab CI workflow, with no test coverage reduction.

### DIFF
--- a/schutzbot/mockbuild.sh
+++ b/schutzbot/mockbuild.sh
@@ -181,6 +181,7 @@ mock -r "$MOCK_CONFIG" \
     --define "commit ${COMMIT}" \
     --config-opts=cleanup_on_failure=False \
     --config-opts=cleanup_on_success=True \
+    --nocheck \
     --with=tests \
     ${RELAX_REQUIRES:+"$RELAX_REQUIRES"} \
     --resultdir "$REPO_DIR" \


supakeen (approved):
Do we want to do this elsewhere as well?

achilleas-k:
> Do we want to do this elsewhere as well?

In other projects, or elsewhere in this project?

In other projects, the tests don't take as long.  From my regular building of osbuild RPMs, for example, the difference in building with or without checks isn't as big.  Still, we could.

For elsewhere in this project: Maybe copr?  Though that's the only place where we do any upstream testing of s390x and ppc64le.

supakeen:
> > Do we want to do this elsewhere as well?
> 
> In other projects, or elsewhere in this project?
> 
> In other projects, the tests don't take as long. From my regular building of osbuild RPMs, for example, the difference in building with or without checks isn't as big. Still, we could.
> 
> For elsewhere in this project: Maybe copr? Though that's the only place where we do any upstream testing of s390x and ppc64le.

I mostly meant in `image-builder`, or `osbuild` but I guess we run less in `%check` there.

achilleas-k:
> I mostly meant in `image-builder`, or `osbuild` but I guess we run less in `%check` there.

And we don't build RPMs in GitLab CI in image-builder.

avitova (approved):

PR #5233, opened 2026-08-25
cloudapi/v2: apply user AWS tags when registering AMIs

kbilwane:
## Remarks

When Image Builder publishes an AMI, callers can already choose the AMI name (`snapshot_name`), but any tags they send are dropped at registration. Composer only applied its own internal tags, so hosted Image Builder could not honor customer tagging on upload.

This PR threads optional user tags into `Register` so they land on the AMI and snapshot, and still sets `composer-api=<key>` so maintenance can find the image after the name is no longer composer-generated.

That is the composer half of [RHEL-123580](https://issues.redhat.com/browse/RHEL-123580), a customer-reported request open since **23 October 2025** (~10 months). Until this ships, users have to tag AMIs after the compose finishes. It should merge **before** [image-builder-crc#1883](https://github.com/osbuild/image-builder-crc/pull/1883); CRC sending `tags` against current composer OpenAPI will fail validation.

## Summary
- Accept optional AWS tags on `AWSEC2UploadOptions` and pass them through the AWS target into `Register`.
- Keep a `composer-api` tag so service maintenance can still find images when `snapshot_name` / AMI Name is user-chosen.

Related: https://issues.redhat.com/browse/RHEL-123580

## Test plan
- [ ] `TestAWSTargetPreservesSnapshotNameAndTags` passes
- [ ] AWS compose with `snapshot_name` and `tags` registers the AMI with those tags

[RHEL-123580]: https://redhat.atlassian.net/browse/RHEL-123580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/cloud/awscloud/maintenance.go
+++ b/internal/cloud/awscloud/maintenance.go
@@ -10,9 +10,9 @@ import (
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
 
-// For service maintenance images are discovered by the "Name:composer-api-*" tag filter. Currently
-// all image names in the service are generated, so they're guaranteed to be unique as well. If
-// users are ever allowed to name their images, an extra tag should be added.
+// For service maintenance images are discovered by the "composer-api" tag.
+// User-chosen AMI names are stored in the Name tag; composer-api remains the
+// stable identifier for cleanup.
 func (a *AWS) DescribeImagesByTag(tagKey, tagValue string) ([]ec2types.Image, error) {
 	imgs, err := a.ec2.DescribeImages(
 		context.Background(),
--- a/internal/cloudapi/v2/imagerequest.go
+++ b/internal/cloudapi/v2/imagerequest.go
@@ -42,11 +42,22 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
 		amiBootMode = common.ToPtr(string(ec2types.BootModeValuesLegacyBios))
 	}
 
+	var tags []target.AWSTag
+	if awsUploadOptions.Tags != nil {
+		for _, tag := range *awsUploadOptions.Tags {
+			tags = append(tags, target.AWSTag{Key: tag.Key, Value: tag.Value})
+		}
+	}
+	// Keep a stable composer-api tag so maintenance can find images even when
+	// snapshot_name / AMI Name is a user-chosen value.
+	tags = append(tags, target.AWSTag{Key: "composer-api", Value: key})
+
 	t := target.NewAWSTarget(&target.AWSTargetOptions{
 		Region:            awsUploadOptions.Region,
 		Key:               key,
 		ShareWithAccounts: awsUploadOptions.ShareWithAccounts,
 		BootMode:          amiBootMode,
+		Tags:              tags,
 	})
 	if awsUploadOptions.SnapshotName != nil {
 		t.ImageName = *awsUploadOptions.SnapshotName
--- a/cmd/osbuild-worker/jobimpl-osbuild.go
+++ b/cmd/osbuild-worker/jobimpl-osbuild.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/osbuild/image-builder/pkg/arch"
+	ibawscloud "github.com/osbuild/image-builder/pkg/cloud/awscloud"
 	"github.com/osbuild/image-builder/pkg/cloud/azure"
 	"github.com/osbuild/image-builder/pkg/cloud/gcp"
 	"github.com/osbuild/image-builder/pkg/container"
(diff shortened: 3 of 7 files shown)


internal/cloud/awscloud/maintenance.go
```diff
@@ -10,9 +10,9 @@ import (
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
 
-// For service maintenance images are discovered by the "Name:composer-api-*" tag filter. Currently
-// all image names in the service are generated, so they're guaranteed to be unique as well. If
-// users are ever allowed to name their images, an extra tag should be added.
+// For service maintenance images are discovered by the "composer-api" tag.
+// User-chosen AMI names are stored in the Name tag; composer-api remains the
+// stable identifier for cleanup.
```
croissanne, line 15: This updates the comment but not the actual caller in `cmd/osbuild-service-maintenance/aws.go`.


Also I think we'd need a short transition period here where we list images by both the Name tag and the `composer-api` tag.

internal/cloudapi/v2/imagerequest.go
```diff
@@ -42,11 +42,22 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
...
+
 	t := target.NewAWSTarget(&target.AWSTargetOptions{
 		Region:            awsUploadOptions.Region,
 		Key:               key,
 		ShareWithAccounts: awsUploadOptions.ShareWithAccounts,
 		BootMode:          amiBootMode,
+		Tags:              tags,
 	})
 	if awsUploadOptions.SnapshotName != nil {
 		t.ImageName = *awsUploadOptions.SnapshotName
```
croissanne, line 63: We do need to ensure uniqueness for service images. That said, maybe we don't enforce it here, osbuild/image-builder-crc#1883 is probable a more suitable place for it.

internal/cloudapi/v2/imagerequest.go
```diff
@@ -42,11 +42,22 @@ func newAWSTarget(options UploadOptions, imageType distro.ImageType) (*target.Ta
...
 
+	var tags []target.AWSTag
+	if awsUploadOptions.Tags != nil {
+		for _, tag := range *awsUploadOptions.Tags {
+			tags = append(tags, target.AWSTag{Key: tag.Key, Value: tag.Value})
+		}
+	}
+	// Keep a stable composer-api tag so maintenance can find images even when
+	// snapshot_name / AMI Name is a user-chosen value.
+	tags = append(tags, target.AWSTag{Key: "composer-api", Value: key})
```
croissanne, line 53: @thozza just to make sure, is it fine if we end up adding this for all brew images as well? I assume they post-process the images a bit + tags are lost when moving images across accounts but still.

croissanne (changes requested):
Thank you, mostly LGTM. The maintenance code needs an update though.

kbilwane:
**Remarks**

Callers can already choose the AMI name (`snapshot_name`), but user tags were dropped at registration. This PR applies those tags on the AMI and snapshot, and still sets `composer-api=<key>` so maintenance can find the image when the name is user-chosen.

This is the composer half of [RHEL-123580](https://issues.redhat.com/browse/RHEL-123580) (open since 23 October 2025, ~10 months). It should merge before https://github.com/osbuild/image-builder-crc/pull/1883.

[RHEL-123580]: https://redhat.atlassian.net/browse/RHEL-123580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

internal/cloud/awscloud/maintenance.go
kbilwane: Updated cmd/osbuild-service-maintenance/aws.go to list images by both Name=composer-api-* (old images) and the composer-api tag (new images), then dedupe by AMI id.

internal/cloudapi/v2/imagerequest.go
kbilwane: Agreed — leaving uniqueness to the hosted service in https://github.com/osbuild/image-builder-crc/pull/1883

cmd/osbuild-service-maintenance/aws.go
```diff
@@ -55,11 +55,30 @@ func AWSCleanup(maxConcurrentRequests int, dryRun bool, accessKeyID, accessKey s
...
-		}
+		byName, err := a.DescribeImagesByTag("Name", "composer-api-*")
+if err != nil {
+	log.Printf("Unable to describe images by Name for region %s: %v", region, err)
+	continue
+}
+byTag, err := a.DescribeImagesByTag("composer-api", "*")
+if err != nil {
+	log.Printf("Unable to describe images by composer-api tag for region %s: %v", region, err)
+	continue
```
croissanne: You can run `tools/prepare-source.sh` which will run `go fmt` etc.

croissanne (changes requested):
Thanks, lgtm. Needs a `go fmt` though.

Also sorry for the late review, I was out last week.

internal/cloudapi/v2/imagerequest.go
kbilwane: Ran gofmt on the maintenance cleanup and imagerequest changes.

PR #5245, opened 2026-09-08
Add compose_id to the logs (HMS-11272)

schuellerf:
This is part 1 of my journey to add `compose_id` to logs where available.
Main change of this PR is to create the `compose_id` _before_ the other related jobs are created so we can save it in the database and reference for all other jobs.

This should help a lot when looking at logs.


schuellerf:
as `splunk_logger` changed, the tests with `go mod vendor` are supposed to break? 🤔
(until this PR is merged)

```
 type SplunkEvent struct {
-	Message   string `json:"message"`
-	Ident     string `json:"ident"`
-	Host      string `json:"host"`
-	ComposeID string `json:"compose_id,omitempty"`
-	JobID     string `json:"job_id,omitempty"`
+	Message string `json:"message"`
+	Ident   string `json:"ident"`
+	Host    string `json:"host"`
 }
```

pkg/jobqueue/jobqueue.go
```diff
@@ -149,3 +156,20 @@ type Worker struct {
...
+
+// Child stamps a job as belonging to composeID without pinning the job UUID.
+func Child(composeID uuid.UUID) EnqueueParams {
+	return EnqueueParams{ComposeID: composeID}
+}
+
+// Root pins the job UUID to composeID so the graph root is the compose id.
+func Root(composeID uuid.UUID) EnqueueParams {
+	return EnqueueParams{ID: composeID, ComposeID: composeID}
+}
```
croissanne, line 175: i think the names `Child` and `Root` are actually confusing here. If only because the KojiFinalize or OSBuild jobs are the leaf nodes, and the dependencies are the  roots.

I'd just drop the helper functions and create the Params inline when queueing jobs.

pkg/jobqueue/dbjobqueue/schemas/010_jobs_compose_id.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE jobs ADD COLUMN IF NOT EXISTS compose_id uuid;
```
croissanne, line 1: The name here makes the queue a bit less generic imo, I wonder if we could call it someting like `group_id`?

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -260,6 +264,24 @@ func NewWithConfig(url string, config Config) (*DBJobQueue, error) {
 	return q, nil
 }
 
+func (q *DBJobQueue) ComposeID(id uuid.UUID) (uuid.UUID, error) {
```
croissanne, line 267: Same here, I think like `GroupID` would be a bit more explicit.

croissanne (commented):
Thank you! This makes sense, but some of the naming threw me off at first. I think we can keep the job queue generic and be bit more explicit at the same time.

cmd/osbuild-worker/main.go
```diff
@@ -143,9 +155,10 @@ func RequestAndRunJob(client *worker.Client, acceptedJobTypes []string, jobImpls
 		return err
 	}
 
+	log := jobLog(job)
```
supakeen, line 158: Does this rename the `jobId` field to `job_id` which is then applied on all the lines below which have a big warning on them to "DO NOT EDIT", or does it duplicate the field under both names?

pkg/jobqueue/dbjobqueue/dbjobqueue.go
```diff
@@ -260,6 +264,32 @@ func NewWithConfig(url string, config Config) (*DBJobQueue, error) {
 	return q, nil
 }
 
+func (q *DBJobQueue) jobLogArgs(id uuid.UUID, args ...string) []string {
+	cid, err := q.ComposeID(id)
```
supakeen, line 268: This gets a second connection from the database (in `ComposeID`) and I think this is executed a lot; can't we pass in the connection to use instead? Right now it'd double the amount of connections taken from the pool?

internal/cloudapi/v2/server.go
supakeen, line 792: Is it `jobId` or `job_id`?

internal/cloudapi/v2/handler.go
```diff
@@ -1297,7 +1298,23 @@ func (h *apiHandlers) postCloneComposeImpl(ctx echo.Context, jobId uuid.UUID) er
...
+			return HTTPError(ErrorUnknownUploadTarget)
+		}
+		if len(awsT.ShareWithAccounts) > 0 {
+			shares = append(shares, awsT.ShareWithAccounts...)
+		}
+		if img.ShareWithAccounts != nil && len(*img.ShareWithAccounts) > 0 {
+			shares = append(shares, (*img.ShareWithAccounts)...)
+		}
+		willShare := len(shares) > 0
+		composeID := uuid.New()
```
supakeen, line 1315: This isn't used in all paths so maybe move it into where it's being used?

internal/cloudapi/v2/handler.go
```diff
@@ -1297,7 +1298,23 @@ func (h *apiHandlers) postCloneComposeImpl(ctx echo.Context, jobId uuid.UUID) er
...
 
+		if img.Region != options.Region {
 			// Check dependents if we need to do a copyjob
 			foundDep := false
 			for _, d := range osbuildInfo.Dependents {
```
supakeen, line 1320: Ok this might b e very wrong but what if an image has been previously shared or cloned doesn't this find the old job instead and then gets confused about which copy job to use so you get mixed compose IDs?

supakeen (commented):
Not very well known around these parts of the codebase but some remarks.

pkg/jobqueue/dbjobqueue/schemas/010_jobs_compose_id.sql
schuellerf: For consistency we would need to propagate the name then? So in the log systems (splunk, sumo, cloud watch…) there would be the field `group_id` which might be confusing what this is - not sure if we really need a generic name here…
What would be a possible use where this wouldn't be _the_ `compose_id` but a generic group?

PR #5252, opened 2026-09-21
osbuild-worker: retry launching secure instance on removal (HMS-11010)

schuellerf:
Sometimes AWS removes our secure instance. This does not happen very often so we retry once in only this case.


supakeen (commented):
If we want to know if an EC2 instance got killed why don't we ask AWS instead of doing a whole bunch of "try to figure out if the instance is gone"?

schuellerf:
> If we want to know if an EC2 instance got killed why don't we ask AWS instead of doing a whole bunch of "try to figure out if the instance is gone"?

As far as I learned, there would only be an [interface](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html) on the _executor_ we would need to implement in addition - which would be more effort and we would need to hand this information the worker anyway who would need to implement the retry (on this new signal).
But I'll set this to draft anyway as we might pivot towards using ondemand instances for executors and this problems gets less relevant…

supakeen:
> > If we want to know if an EC2 instance got killed why don't we ask AWS instead of doing a whole bunch of "try to figure out if the instance is gone"?
> 
> As far as I learned, there would only be an [interface](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html) on the _executor_ we would need to implement in addition - which would be more effort and we would need to hand this information the worker anyway who would need to implement the retry (on this new signal). But I'll set this to draft anyway as we might pivot towards using ondemand instances for executors and this problems gets less relevant…

As far as I understand AWS we can do (from the worker):

1. Any error.
2. Retry the POST (except if the error occurred during streaming I guess, no point to retry and just check if the instance still exists immediately).
3. Call `DescribeInstance` on the instance ID of the instance that might be gone.
4. Check if the value is `InstanceNotFound` or its state is `InstanceStateNameTerminated`, `InstanceStateNameShuttingDown`.

If any of the latter three are true then the instance is gone and we don't need to bother with trying to figure out why itw as gone and can just recreate immediately?

This to me feels better than trying to parse the various error messages we could get when connecting to the machine or losing connection in the middle.

It would get rid of all the heuristics that are in here, the functions to detect those heuristics, all the errno and string matching stuff, and it would be more resilient to false positives.

*But* this all assumes the AWS API is ... somewhat capable and updates fast.

Something like this on `secureInstanceClient`:

```go
func (c *awsCloudClient) IsInstanceGone(ctx context.Context, instanceID string) (bool, error) {    
    out, err := c.ec2.DescribeInstances(ctx, &ec2.DescribeInstancesInput{    
        InstanceIds: []string{instanceID},    
    })    
    if err != nil {    
        // Instance not found = gone    
        var notFound *ec2types.InstanceNotFoundException    
        if errors.As(err, &notFound) {    
            return true, nil    
        }    
        return false, err    
    }    
    for _, r := range out.Reservations {    
        for _, inst := range r.Instances {    
            switch inst.State.Name {    
            case ec2types.InstanceStateNameTerminated,    
                 ec2types.InstanceStateNameShuttingDown:    
                return true, nil    
            }    
        }    
    }    
    return false, nil    
} 
```

PR #5254, opened 2026-09-21
Add component AWS tags executor (HMS-11010)

schuellerf:
Add AWS tags for "executor" and "worker" for better cost-tracking.


internal/cloud/awscloud/secure-instance_test.go
```diff
@@ -218,3 +218,106 @@ func TestDoCreateFleetRetry(t *testing.T) {
...
+	for _, spec := range input.TagSpecifications {
+		if spec.ResourceType == resourceType {
+			return tagMap(spec.Tags)
+		}
+	}
+	t.Fatalf("no TagSpecification for %s", resourceType)
+	return nil
+}
+
+func TestSecureInstanceTags(t *testing.T) {
```
croissanne, line 242: Could we write these tests as a table (driven) test? https://go.dev/wiki/TableDrivenTests I think it should be possible to have a single table driven test for all tagging stuff. 

For each test case just check if the `SecureInstanceTags` fn returns the expected values + that actually running the secure instance passes along the correct tags.

croissanne (commented):
Nice, thank you! LGTM but I think we should rewrite the test as a table driven test. We tend to write any new tests that way.

internal/cloud/awscloud/secure-instance_test.go
schuellerf: Thanks for the hint - that's nicer for sure, please check

croissanne (commented):
Nice! LGTM, could you squash the test change into the first commit? Then it should be good to go

schuellerf:
sure thing 🗜️ - done

croissanne (approved):

achilleas-k (approved):

